PR #1760, opened 2018-04-09
Fix hasSubterm calls for higher-order

ajreynol:
Need to consider operators when checking hasSubterm.

This is to avoid, e.g. infinite loops:

f : Int -> Int
g : Int -> Int
f = (g (f a))

we cannot eliminate f -> \ x. ((g (f a)) x) since its body contains f.


src/expr/node.h
```diff
@@ -1539,6 +1539,10 @@ bool NodeTemplate<ref_count>::hasSubterm(NodeTemplate<false> t, bool strict) con
 
   for (unsigned i = 0; i < toProcess.size(); ++ i) {
     TNode current = toProcess[i];
+    if (current.hasOperator() && current.getOperator() == t)
```
HanielB: Add the HO flag since this test only makes sense to HO and both ``hasOperator`` and ``getOperator`` do switches

src/theory/quantifiers/term_util.cpp
```diff
@@ -690,43 +693,31 @@ Node TermUtil::ensureType( Node n, TypeNode tn ) {
...
-      visited[n] = true;
-      for( unsigned i=0; i<n.getNumChildren(); i++ ){
-        if( containsTerms2( n[i], t, visited ) ){
-          return true;
-        }
+    }
+    visited[n] = true;
+    if (n.hasOperator())
+    {
+      if (containsTerms2(n.getOperator(), t, visited))
```
HanielB: merge ``if``

ajreynol: i dont like merging ifs if things become sensitive to evaluation order.  Here, if !n.hasOperator(), then calling containsTerms2(n.getOperator()...) will crash.

src/expr/node.h
ajreynol: I think the precedent here is to always check.  See:
https://github.com/CVC4/CVC4/blob/master/src/expr/node.cpp#L129
Basically, node is a generic utility that we probably don't want to depend on an theory/uf option.

src/theory/quantifiers/term_util.cpp
HanielB: but if ``!n.hasOperator()`` the function would not be called. I don't think separating the ``if`` helps any more than writing a single one

ajreynol: If they are merged:

if( n.hasOperator() && containsTerms2(n.getOperator(), t, visited))

if the second conjunct is evaluated first, the function is called and CVC4 may crash.  Generally, I don't like relying on the evaluation order of && being strict.

HanielB: I think it's fair to assume that && will always be evaluated strictly in this order.

ajreynol: I disagree, I have been bitten by this in the past.  Plus I think the code is better documented if these are kept separate: this is saying that the developer wants to enforce the evaluation order.

PR #1762, opened 2018-04-09
Fixes for free variables in assertions

ajreynol:
This fixes two cases of free variables in assertions:

(1) --sygus-inference mistakenly used a bound variable for an annotation instead of a skolem

(2) CHOICE and LAMBDA can now occur in assertions, in particular CHOICE is used for expanding the definitions of inverse transcendental functions, LAMBDA is allowed in the higher-order smt2 parser.  This PR updates a number of places where we were not treating these as quantifiers.

### DIFF
--- a/src/smt/term_formula_removal.cpp
+++ b/src/smt/term_formula_removal.cpp
@@ -215,8 +215,11 @@ Node RemoveTermFormulas::run(TNode node, std::vector<Node>& output,
     // The representation is now the skolem
     return skolem;
   }
-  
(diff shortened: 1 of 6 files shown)


ajreynol:
Btw, I removed a virtual function ppDontRewriteSubterm, this appeared to be only for quantifiers. I think this is more intuitive if the check on https://github.com/CVC4/CVC4/pull/1762/files#diff-e116885c6724373f396534aa14ca6068R1062 checked all quantifier types (which includes CHOICE and LAMBDA).  

A separate question is how we could have some generic infastructure for some "isQuant(Kind k)".

PR #1773, opened 2018-04-13
Fix get-unsat-core detection in regression script

4tXJ7f:
Due to a typo in the regression script, the (get-unsat-core) command was
not recognized (the script was looking for (get-unsat-cores)), so it
tried to run a regression script containing (get-unsat-core) even when
CVC4 was compiled without proof support. This commit fixes the typo and
additionally changes the script to look for the pattern in the benchmark
file instead of the metadata file (i.e. the .expect file if it exists or
the benchmark file otherwise). The second fix is not strictly necessary
for our current regressions but might have lead to surprising outcomes
in the future.

### DIFF
--- a/test/regress/run_regression.py
+++ b/test/regress/run_regression.py
@@ -143,13 +143,7 @@ def run_regression(proof, dump, wrapper, cvc4_binary, benchmark_path):
     metadata_lines = None
     with open(metadata_filename, 'r') as metadata_file:
         metadata_lines = metadata_file.readlines()
-
-    benchmark_content = None
-    if metadata_filename == benchmark_path:
-        benchmark_content = ''.join(metadata_lines)
-    else:
-        with open(benchmark_path, 'r') as benchmark_file:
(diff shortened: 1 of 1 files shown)


test/regress/run_regression.py
```diff
@@ -187,8 +193,8 @@ def run_regression(proof, dump, wrapper, cvc4_binary, benchmark_path):
             # been set explicitly, the benchmark is invalid.
             sys.exit('Cannot determine status of "{}"'.format(benchmark_path))
 
-    if not proof and ('(get-unsat-cores)' in metadata_content
-                      or '(get-unsat-assumptions)' in metadata_content):
+    if not proof and ('(get-unsat-core)' in benchmark_content
```
timothy-king: Maybe submit just this and open a second PR for the other issue?

4tXJ7f: Updated to be fix only.

PR #1774, opened 2018-04-13
Disable split for negative contains.

ajreynol:


4tXJ7f:
For the record, removing this split gets us +143 sat, +0 unsat, -7 sat, -0 unsat with the standard rewriter on the CAV 2017 benchmarks.

PR #1777, opened 2018-04-14
Skolemize candidate rewrite rule checks

ajreynol:
Avoid assertion failures when using --sygus-rr-synth-check.

This also cleans up the utilities for computing free variables in term_util.


ajreynol:
Regarding a single pass: this can't be done, we're applying the Skolemization to terms that are enumerated during search.
Yes, I do agree we need a global solution for collecting.  Eventually, term_util will be split up and moved to more appropriate places, there is an issue for that (https://github.com/CVC4/CVC4/issues/1216)

PR #1783, opened 2018-04-16
Make 256 the default cardinality for strings

### LINKED ISSUE #1781: Cardinality of string elements
According to http://cvc4.cs.stanford.edu/wiki/Strings#Alphabet CVC4 strings are composed of alphabets with 256 elements. The attached benchmark tries to generate all of them by enumerating the solutions returned by CVC4. Alas, it only generates 128 elements, those with ASCII codes 65 through 192.

I suspect this is a bug, but wanted to double check. Run the attached script with `cvc4 --incremental`, using a freshly built cvc4 using brew.

[b.smt2.txt](https://github.com/CVC4/CVC4/files/1913878/b.smt2.txt)

ajreynol:
Fixes #1781.

Previously the default was 128, which didnt match the documentation on the webpage.

### DIFF
--- a/src/theory/strings/theory_strings.cpp
+++ b/src/theory/strings/theory_strings.cpp
@@ -143,7 +143,11 @@ TheoryStrings::TheoryStrings(context::Context* c, context::UserContext* u,
   d_true = NodeManager::currentNM()->mkConst( true );
   d_false = NodeManager::currentNM()->mkConst( false );
 
-  d_card_size = 128;
+  d_card_size = 256;
+  if (options::stdPrintASCII())
+  {
+    d_card_size = 128;
+  }
 }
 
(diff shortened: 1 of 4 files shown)


src/theory/strings/theory_strings.cpp
```diff
@@ -143,7 +143,11 @@ TheoryStrings::TheoryStrings(context::Context* c, context::UserContext* u,
   d_true = NodeManager::currentNM()->mkConst( true );
   d_false = NodeManager::currentNM()->mkConst( false );
 
-  d_card_size = 128;
+  d_card_size = 256;
```
4tXJ7f: Nit: could just write this as `d_card_size = options::stdPrintASCII() ? 256 : 128;`.

PR #1784, opened 2018-04-16
 Reenable filtering based on ordering in sygus sampler

ajreynol:
This code was disabled in https://github.com/CVC4/CVC4/commit/9dcaaeba4880a8f145df00289ff1b092a7e3dd47#diff-6c28bdfd922fb6c528a0092ba705ee98
since it was (mistakenly) thought to be subsumed by filtering based on matching.

Update: fixed the filtering so that we allow strange rewrites where both sides contain a unique free variable, e.g.:
(candidate-rewrite (* 0 x) (* 0 y))


ajreynol:
I've updated the PR so that each of the individual filtering techniques can be enabled/disabled.  To get accurate stats with exact filtering, you should also use --no-sygus-rr-synth-filter-order.

src/theory/quantifiers/sygus_sampler.cpp
```diff
@@ -707,36 +721,62 @@ Node SygusSamplerExt::registerTerm(Node n, bool forceKeep)
...
-    Trace("sse-match") << "SSE check matches : " << n << " [rhs = " << eq_n
-                      << "]..." << std::endl;
-    if (!d_match_trie.getMatches(bn, &d_ssenm))
+    bool nor = isOrdered(bn);
+    bool eqor = isOrdered(beq_n);
+    Trace("sygus-synth-rr-debug") << "Ordered? : " << nor << " " << eqor
+                                  << std::endl;
+    if (eqor || nor)
+    {
+      // if only one is ordered, then the ordered one's variables cannot be
```
4tXJ7f: Let's say the order is `[ x, y ]`, couldn't one be ordered `x` and be a subset of the unordered term `y + x`?

src/theory/quantifiers/sygus_sampler.cpp
```diff
@@ -432,6 +433,16 @@ bool SygusSampler::containsFreeVariables(Node a, Node b)
         {
           return false;
         }
+        else if (strict)
+        {
+          if (fv_found.size() + 1 == fvs.size())
+          {
+            return false;
+          }
+          Assert(std::find(fv_found.begin(), fv_found.end(), cur)
```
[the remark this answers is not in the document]
ajreynol: Yes, but it should only be visited once in this function, I've added a comment to indicate this.

src/theory/quantifiers/sygus_sampler.cpp
ajreynol: The comment is stating our criteria (it's not stating what is/isn't possible).  I'll update this comment too to make it clear.

PR #1786, opened 2018-04-16
Add timeout (option) to regression script

4tXJ7f:
This commit adds the option to run regressions with a timeout using the
`TEST_TIMEOUT` environment variable. The default timeout is set to 10
minutes. This should make it less likely that our nightly builds hang
and makes it easier to sort out slow tests.

Default timeout tested with regression level 2 on a debug build with
proofs.

### DIFF
--- a/test/regress/run_regression.py
+++ b/test/regress/run_regression.py
@@ -41,13 +41,13 @@ def run_process(args, cwd, timeout, s_input=None):
 
     out = ''
     err = ''
+    exit_status = 124
     timer = threading.Timer(timeout, lambda p: p.kill(), [proc])
     try:
         timer.start()
         out, err = proc.communicate(input=s_input)
         exit_status = proc.returncode
     finally:
-        exit_status = 124
         timer.cancel()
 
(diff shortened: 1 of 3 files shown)


test/regress/run_regression.py
```diff
@@ -24,8 +25,30 @@
...
+        stdin=subprocess.PIPE,
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE)
+
+    timer = threading.Timer(timeout, lambda p: p.kill(), [proc])
+    try:
+        timer.start()
+        out, err = proc.communicate(input=s_input)
+    finally:
+        timer.cancel()
```
timothy-king: You might want to indicate that an error status has occurred on time outs. (Maybe this is already taken care of. In which case, you might want to update the documentation.)

4tXJ7f: That's a good point. I changed the code to explicitly initialize the variables and return 124 if the process times out (that seems to be the return code when you use the `timeout` util in Linux.

PR #1793, opened 2018-04-20
Refactor IteRemoval preprocessing pass

4tXJ7f:
This commit refactors the IteRemoval pass to follow the new format.
In addition to moving the code, this entails the following changes:

- The timer for the ITE removal is now called differently (the default
  timer of PreprocessingPass is used) and measures just the
  preprocessing pass without applySubstitutions(). It also measures the
  time used by both invocations of the ITE removal pass.
- Debug output will be slightly different because we now just rely on
  the default functionality of PreprocessingPass.
- PreprocessingPassContext now has a spendResource() method and takes a
  ResourceManager* in its constructor.
- d_iteRemover has been moved from SmtEnginePrivate to SmtEngine (and
  made an std::unique_ptr) to make it accessible by
  PreprocessingPassContext.
- AssertionPipeline now owns the d_iteSkolemMap, which makes it
  accessible by preprocessing passes. The idea behind this is that the
  preprocessing passes collect information in AssertionPipeline and
  d_iteSkolemMap fits that pattern.


src/smt/smt_engine.cpp
```diff
@@ -3860,7 +3839,8 @@ Result SmtEngine::check() {
        (not d_logic.isQuantified() &&
         d_logic.isPure(THEORY_ARITH) && d_logic.isLinear() && !d_logic.isDifferenceLogic() &&  !d_logic.areIntegersUsed()
         )){
-      if(d_private->d_iteSkolemMap.empty()){
+      if (d_private->d_assertions.getIteSkolemMap().empty())
```
yoni206: Do you think `d_assertions` of `SmtEnginePrivate` will be used in other places? If not, maybe it is a good idea to keep its `d_assertions` private, and have a public `getIteSkolemMap()` whose body is `return d_assertions.getIteSkolemMap()`?

yoni206:
Is this pass tested in the current regressions?

src/smt/smt_engine.h
```diff
@@ -152,6 +154,9 @@ class CVC4_PUBLIC SmtEngine {
   std::map< Node, std::vector< Node > > d_fmfRecFunctionsConcrete;
   NodeList* d_fmfRecFunctionsDefined;
 
+  /** Instance of the ITE remover */
+  std::unique_ptr<RemoveTermFormulas> d_iteRemover;
```
timothy-king: Can this go into the SMTEnginePrivate? That would avoid the forward decl.

4tXJ7f: There is an issue with that: I need access to `d_iteRemover` in my preprocessing pass. If it is in SmtEnginePrivate, I can't do that (without including `smt_engine.cpp` or adding a getter for d_iteRemover to SmtEngine but then we need the forward declaration again). I guess one solution would be to have a separate SmtEnginePrivate header that we could include when needed (but this should probably be a separate PR). What do you think about that?

timothy-king: > I need access to d_iteRemover in my preprocessing pass.
Can you point out where this is needed and explain why it is needed? (I won't be very quick to respond so the first opportunity to discuss this might be next week.)

src/smt/smt_engine.cpp
4tXJ7f: This is a good point, I made `d_assertions` private again.

src/smt/smt_engine.h
4tXJ7f: I've moved `d_iteRemover` back to `SmtEnginePrivate`. The code now passes a pointer to the ITE remover to the `PreprocessingPassContext`, which is cleaner.

4tXJ7f:
I have addressed the comments, PTAL.

timothy-king:
This addresses my concerns.

src/smt/smt_engine.cpp
```diff
@@ -4382,22 +4362,15 @@ void SmtEnginePrivate::processAssertions() {
   }
   Debug("smt") << " d_assertions     : " << d_assertions.size() << endl;
 
-  Trace("smt-proc") << "SmtEnginePrivate::processAssertions() : pre-ite-removal" << endl;
-  dumpAssertions("pre-ite-removal", d_assertions);
   {
-    Chat() << "removing term ITEs..." << endl;
-    TimerStat::CodeTimer codeTimer(d_smt.d_stats->d_iteRemovalTime);
```
HanielB: Why are you removing the time stats?

4tXJ7f: Because the PreprocessingPass class automatically provides a timer for each pass, no need to do it explicitly.

HanielB: Ah, that's good to know!

4tXJ7f:
> Is this pass tested in the current regressions?

Yes, this pass is applied a great deal in the regressions.

Issue #1535, opened 2018-01-25
Investigate Swig 3 compatibility


4tXJ7f:
We have issues with Swig 3, see #1533 for an example (I seem to remember that there were other issues as well).

4tXJ7f:
It looks like Swig 3 is not a problem (anymore?). We know of at least one person that is using Swig 3 (for Python bindings) successfully in production. I am closing this issue. If other, concrete issues appear in the future, we can open separate issues.

PR #1794, opened 2018-04-20
Delete functions instead of using CVC4_UNDEFINED

4tXJ7f:
C++11 supports explicitly deleting functions that should not be used
(explictly or implictly), e.g. copy or assignment constructors. We were
previously using the CVC4_UNDEFINED macro that used a compiler-specific
attribute. The C++11 feature should be more portable.

@timothy-king: I'm adding you as a reviewer because this could be of interest to you.


timothy-king:
You have already brought up SWIG. It is my biggest concern with making this type of change. Please make sure to additionally test building the java examples as well as building from scratch before checking this in.

4tXJ7f:
Alright, so it turns out that CVC4 builds with Swig 3. I am not sure what happened to the earlier issues that we were seeing (#1535). After updating Travis to use Swig 3, the tests pass now and we know about at least one person that is using Swig 3 successfully in production. I think that we can move forward with this PR.

timothy-king:
What is the status of swig 3 on some of the common linux distributions? Debian stable being the likely slow track. (Also mac?) We also might choose to not care about supporting slower moving platforms for being able to compile language bindings at this point. If we need to make that choice, I would rope in @barrettcw .

mpreiner:
Debian stable has swig3, which was introduced with Debian stretch last year.

4tXJ7f:
@barretcw gave his approval just now. macOS (homebrew) does not even provide Swig 2 anymore, we had to provide our own formula for that, so Swig 3 makes things easier.

PR #1799, opened 2018-04-20
Allow metadata lines in test files to have leading spaces

yoni206:
Currently, lines like `;Expect sat` instead of `; Expect sat` cause problems. This PR fixes it.


yoni206:
@4tXJ7f I use lstrip() and not strip() because the trailing "\n"s are counted on.

PR #1811, opened 2018-04-23
Equality resolution in the extended rewriter

ajreynol:
60.5 -> 58.5 % redundancy for depth 6 crci
82.2 -> 81.2 % redundancy for depth 7 crci


src/theory/quantifiers/extended_rewrite.cpp
```diff
@@ -718,6 +725,84 @@ Node ExtendedRewriter::extendedRewriteBcp(
...
+            {
+              if (bcp_kinds.empty())
+              {
+                ccs = ccs.substitute(
+                    vars.begin(), vars.end(), subs.begin(), subs.end());
+              }
+              else
+              {
+                std::map<Node, Node> assign;
+                assign[vars[0]] = subs[0];
```
4tXJ7f: Why only the first substitution?

src/theory/quantifiers/extended_rewrite.cpp
```diff
@@ -718,6 +725,84 @@ Node ExtendedRewriter::extendedRewriteBcp(
...
+  NodeManager* nm = NodeManager::currentNM();
+  Kind nk = n.getKind();
+  bool gpol = (nk == andk);
+  for (unsigned i = 0, nchild = n.getNumChildren(); i < nchild; i++)
+  {
+    Node lit = n[i];
+    if (lit.getKind() == eqk)
+    {
+      Node eq;
+      if (gpol == isXor)
```
4tXJ7f: I don't quite understand why it is ok to turn `(= a b) V ...` into `(= (not a) b) V ...` (I understand the case where you have `(xor a b) /\ ...` and turn it into `(= (not a) b)`).

ajreynol: This isn't turning `(= a b) V ...` into `(= (not a) b) V ...`.
Instead, "eq" here is the equality that can be used as a substitution.  So for example, for `(not a) = b V ( b ^ a )`, this can be seen as `a = b => ( b ^ a )`, which we can use substitution to get `(not a) = b V ( b ^ a ){ a -> b }` which simplifies `(not a) = b V b`.  (which then by BCP simplifies to `a V b`).

4tXJ7f: That makes sense.

PR #1814, opened 2018-04-25
Refactor nonlinear check

ajreynol:
Makes it so that there are "waiting lemmas" that are flushed after a call to check-model.

This is in preparation for more improvements to transcendental functions.


src/theory/arith/nonlinear_extension.cpp
```diff
@@ -1615,6 +1643,125 @@ void NonlinearExtension::getCurrentPiBounds( std::vector< Node >& lemmas ) {
...
+                                                bool isLower,
+                                                unsigned prec) const
+{
+  Assert(c.isConst());
+  Rational cr = c.getConst<Rational>();
+
+  unsigned lower = 0;
+  unsigned upper = pow(10, prec);
+
+  Rational den = Rational(upper);
```
HanielB: Shouldn't you follow the note at 

https://github.com/CVC4/CVC4/blob/6cdf6fb87a48c04b72722382f43b7fded0b97778/src/util/rational_cln_imp.h#L51

which says that you should avoid creating Rationals from integers with this constructor?

src/theory/arith/nonlinear_extension.cpp
```diff
@@ -1615,6 +1643,125 @@ void NonlinearExtension::getCurrentPiBounds( std::vector< Node >& lemmas ) {
...
+      if (esign == 1 && !isLower)
+      {
+        curr_r = Rational(curr + 1) / den;
+      }
+      else if (esign == -1 && isLower)
+      {
+        curr_r = Rational(curr - 1) / den;
+      }
+      curr_r = curr_r * pow_ten;
+      cret = NodeManager::currentNM()->mkConst(csign == 1 ? curr_r : -curr_r);
```
HanielB: cache NodeManager

PR #1816, opened 2018-04-26
Fix subgoal generation context

ajreynol:
Fixes a use-after-free issue.


ajreynol:
Also cleans up mem leaks.

PR #1819, opened 2018-04-27
New module for synthesizing functions in a data-driven SyGuS approach

HanielB:
This PR establishes the module's interface, initialization and communication channels with the core utility which will generate the actual solutions and is yet to be implemented.


src/theory/quantifiers/sygus/sygus_unif_rl.cpp
```diff
@@ -0,0 +1,51 @@
+/*********************                                                        */
+/*! \file sygus_unif_io.cpp
```
ajreynol: rl

src/theory/quantifiers/sygus/cegis_unif.h
```diff
@@ -0,0 +1,157 @@
...
+   * list of enumerators being used to build solutions for candidate by the
+   * above utility.
+   */
+  std::vector<Node> d_enums;
+  /** map from enumerators to active guards */
+  std::map<Node, Node> d_enum_to_active_guard;
+  /* list of learned refinement lemmas */
+  std::vector<Node> d_refinement_lemmas;
+  /**
+  * This is called on the refinement lemma and will replace the argumets of the
```
ajreynol: argument

src/theory/quantifiers/sygus/cegis_unif.cpp
```diff
@@ -0,0 +1,293 @@
...
+      }
+    }
+  }
+  /* put in cache */
+  Trace("cegis-unif-purify") << "... caching [" << n << "] = " << nb << "\n";
+  cache[n] = nb;
+  return nb;
+}
+
+Node CegisUnif::purifyLemma(
```
ajreynol: This function is a bit hard to follow. 
Can you merge this function and the one above and add a flag "ensureConst" that ensures that the return value is a constant?

src/theory/quantifiers/sygus/cegis_unif.cpp
```diff
@@ -0,0 +1,293 @@
...
+void CegisUnif::registerRefinementLemma(const std::vector<Node>& vars,
+                                        Node lem,
+                                        std::vector<Node>& lems)
+{
+  Node plem;
+  std::vector<Node> model_guards;
+  std::unordered_map<Node, Node, NodeHashFunction> cache;
+  Trace("cegis-unif") << "Registering lemma at CegisUnif : " << lem << " with model value " << d_parent->getModelValue(lem) << "\n";
+  /* Make the purified lemma which will guide the unification utility. */
+  plem = purifyLemma(lem, model_guards, cache);
```
ajreynol: I think we need to rewrite this Node.
Try purifying f( 1 + f( 0 ) ) = 0, this would be f( 0 ) != c V f( 1 + c ) = 0 for some constant c, but 1+c should be rewritten.
It would suffice to call plem = Rewriter::rewrite( plem ); here.

src/theory/quantifiers/sygus/cegis_unif.cpp
```diff
@@ -0,0 +1,252 @@
...
+                            bool ensureConst,
+                            std::vector<Node>& model_guards,
+                            BoolNodePairMap& cache)
+{
+  unsigned i, size = n.getNumChildren();
+  Kind k = n.getKind();
+  bool fapp = k == APPLY_UF && size > 0 && n[0] == d_candidate;
+  Trace("cegis-unif-purify") << "PurifyLemma : " << n << "\n";
+  /* top level applications of function-to-synthsize should never be removed */
+  if (ensureConst || !fapp)
```
ajreynol: I'm not sure why we are not caching if !ensureConst  && fapp.  I think this should be cached as well.

src/theory/quantifiers/sygus/cegis_unif.cpp
```diff
@@ -0,0 +1,252 @@
...
+    nb = NodeManager::currentNM()->mkNode(k, children);
+    Trace("cegis-unif-purify") << "PurifyFuncApp : transformed " << n
+                               << " into " << nb << "\n";
+  }
+  else
+  {
+    nb = n;
+  }
+  /* get model value if d_candidate applied over constants */
+  if (ensureConst && fapp)
```
ajreynol: I think this should just be ensureConst.  Try running it on f( 1+f(0) ).  Here, when we recurse to f(0), we should return the model value of f(0), which is a child of +, which isn't a fapp.

src/theory/quantifiers/sygus/cegis_unif.cpp
```diff
@@ -0,0 +1,252 @@
...
+  {
+    Trace("cegis-unif-purify") << "PurifyFuncApp : candidate app " << nb
+                               << "\n";
+    for (i = 1; i < size; ++i)
```
ajreynol: Remove this.  ensureConst means the entire return value is a constant. (it doesn't depend on children already being constant).

src/theory/quantifiers/sygus/cegis_unif.cpp
```diff
@@ -0,0 +1,252 @@
...
+        Trace("cegis-unif-purify") << "PurifyFuncApp : adding model eq\n";
+        model_guards.push_back(
+            NodeManager::currentNM()->mkNode(EQUAL, nv, nb).negate());
+        nb = nv;
+      }
+    }
+  }
+  /* top level applications of function-to-synthesize should not interfere with
+     internal applications being purified */
+  if (ensureConst || !fapp)
```
ajreynol: Like the check at the top, I think this check can be removed.

We should then assert that:
Assert( !ensureConst || nb.isConst() );

src/theory/quantifiers/sygus/cegis_unif.cpp
HanielB: This is what happens already. Without the ``fapp`` guard we would, besides replacing ``1+f(0)`` by ``1+[f(0)]^M", also replace ``1+[f(0)]^M`` by ``[1+[f(0)]^M]^M``, which we do not need to do, since the rewriter will do it and this would generate a bogus model equality.

ajreynol: Ah, right.  I think once we have the assertion Assert( !ensureConst || nb.isConst() ); at the bottom this will be clear.

src/theory/quantifiers/sygus/cegis_unif.cpp
HanielB: I removed the now spurious tests but I used a different assert. It would fail e.g. for ``(+ 1 2)``. What we need to guarantee is ``Assert(!ensureConst || !fapp || nb.isConst())``

src/theory/quantifiers/sygus/cegis_unif.cpp
```diff
@@ -0,0 +1,234 @@
...
+    {
+      Trace("cegis-unif-purify") << "PurifyLemma : adding model eq\n";
+      model_guards.push_back(
+          NodeManager::currentNM()->mkNode(EQUAL, nv, nb).negate());
+      nb = nv;
+    }
+  }
+  /* every non-top level application of function-to-synthesize must be reduced
+     to a concrete constant */
+  Assert(!ensureConst || !fapp || nb.isConst());
```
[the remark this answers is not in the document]
HanielB: I agree we should rewrite ``nb`` however if I remove ``!fapp`` from the assertion then I have to test than ``nb == d_candidate`` because we do not rewrite the deep embedding of the function into the function (its model value), and we want to keep the deep embedding.

PR #1821, opened 2018-04-27
Add support for str.code

ajreynol:
Adds support for a "str.code" operator in preparation for the new strings standard.  This is a variant of "char.code" for strings, which returns -1 when its string argument does not have length 1.

Supporting str.code requires:
- A lemma schema for enforcing that str.code( s ) is in the proper range for each instance
- Tracking which equivalence classes have str.code applications and modifying the model construction procedure to assign the proper strings for equivalence classes that have str.code.
- Introducing the constraint x = c ^ str.code( x ) = code(c) for all constants c of length one that occur in the context when at least one other str.code application occurs in the context, where x is the proxy variable for c.
- A lemma schema to ensure that str.code is injective.

### DIFF
--- a/src/printer/smt2/smt2_printer.cpp
+++ b/src/printer/smt2/smt2_printer.cpp
@@ -1061,13 +1061,6 @@ static string smtKindString(Kind k, Variant v)
   case kind::STRING_STRREPL: return "str.replace" ;
   case kind::STRING_PREFIX: return "str.prefixof" ;
   case kind::STRING_SUFFIX: return "str.suffixof" ;
-<<<<<<< HEAD
-  case kind::STRING_ITOS: return "int.to.str" ;
-  case kind::STRING_STOI: return "str.to.int" ;
-  case kind::STRING_CODE: return "str.code";
-  case kind::STRING_IN_REGEXP: return "str.in.re";
-  case kind::STRING_TO_REGEXP: return "str.to.re";
-=======
   case kind::STRING_ITOS:
     return v == smt2_6_1_variant ? "str.from-int" : "int.to.str";
   case kind::STRING_STOI:
@@ -1076,7 +1069,6 @@ static string smtKindString(Kind k, Variant v)
     return v == smt2_6_1_variant ? "str.in-re" : "str.in.re";
   case kind::STRING_TO_REGEXP:
     return v == smt2_6_1_variant ? "str.to-re" : "str.to.re";
->>>>>>> 884ad1a946ad6a04664ef97121ce1cebb5513d40
   case kind::REGEXP_CONCAT: return "re.++";
   case kind::REGEXP_UNION: return "re.union";
   case kind::REGEXP_INTER: return "re.inter";


ajreynol:
Btw, regarding annotations of benchmarks, generally annotating the benchmarks with COMMAND-LINE/EXPECT is the exception to the rule. Most regressions (including those I added here) only annotate the status with :status sat/unsat.

src/theory/strings/theory_strings_rewriter.h
```diff
@@ -109,6 +109,12 @@ class TheoryStringsRewriter {
   * Returns the rewritten form of node.
   */
   static Node rewritePrefixSuffix(Node node);
+  /** rewrite str.code
+   * This is the entry point for post-rewriting terms n of the form
+   *   str.code( t )
+   * Returns the rewritten form of node.
```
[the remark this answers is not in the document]
ajreynol: I think this would be too verbose for documentation in the h file. The rewriters for each of the string functions are very complex and subject to change.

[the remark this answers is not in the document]
4tXJ7f: I agree that it is not necessary to document this in the header but to keep that documentation in the body of the function.

src/theory/strings/theory_strings.cpp
```diff
@@ -423,7 +427,9 @@ int TheoryStrings::getReduction( int effort, Node n, Node& nr ) {
           //we've reduced this n
           Trace("strings-extf-debug") << "  resolve extf : " << n << " based on positive contain reduction." << std::endl;
           return 1;
-        }else{
+        }
+        else if (n.getKind() != kind::STRING_CODE)
+        {
           // for STRING_SUBSTR, STRING_STRCTN with pol=-1,
```
4tXJ7f: Could we add an assertion for the expected kinds here?

src/util/regexp.h
```diff
@@ -28,21 +28,57 @@
...
+  /** get the internal unsigned for ASCII code c. */
+  static unsigned convertCodeToUnsignedInt(unsigned c)
+  {
+    return (c < start_code() ? c + num_codes() : c) - start_code();
+  }
+  /** get the ASCII code number that internal unsigned i corresponds to. */
+  static unsigned convertUnsignedIntToCode(unsigned i)
+  {
+    unsigned ii = i + start_code();
+    return ii >= num_codes() ? ii - num_codes() : ii;
```
4tXJ7f: Suggestion: Could just say `return (ii + start_code()) % num_codes();`

src/util/regexp.h
```diff
@@ -28,21 +28,57 @@
...
     return (c >= ' ' && c <= '~');  // isprint( (int)c );
   }
+  /** get the internal unsigned for ASCII code c. */
+  static unsigned convertCodeToUnsignedInt(unsigned c)
+  {
+    return (c < start_code() ? c + num_codes() : c) - start_code();
+  }
+  /** get the ASCII code number that internal unsigned i corresponds to. */
+  static unsigned convertUnsignedIntToCode(unsigned i)
+  {
```
4tXJ7f: `Assert(i < num_codes());` might be useful here.

test/regress/regress1/strings/str-code-sat.smt2
```diff
@@ -0,0 +1,24 @@
+(set-logic QF_S)
```
4tXJ7f: Nit: shouldn't this be `QF_SLIA` since there are integer additions/comparisons?

test/regress/regress1/strings/str-code-unsat-2.smt2
```diff
@@ -0,0 +1,6 @@
+(set-logic QF_S)
```
4tXJ7f: Nit: shouldn't this be `QF_SLIA` since there are integer comparisons?

test/regress/regress1/strings/str-code-unsat-3.smt2
```diff
@@ -0,0 +1,21 @@
+(set-logic QF_S)
```
4tXJ7f: Nit: shouldn't this be `QF_SLIA` since there are integer additions/comparisons?

test/regress/regress1/strings/str-code-unsat.smt2
```diff
@@ -0,0 +1,21 @@
+(set-logic QF_S)
```
4tXJ7f: Nit: shouldn't this be `QF_SLIA` since there are integer additions/comparisons?

src/theory/strings/theory_strings.cpp
```diff
@@ -3344,10 +3513,25 @@ void TheoryStrings::registerTerm( Node n, int effort ) {
...
+        Node code_range = nm->mkNode(
+            kind::AND,
+            nm->mkNode(kind::GEQ, n, d_zero),
+            nm->mkNode(
+                kind::LT, n, nm->mkConst(Rational(CVC4::String::num_codes()))));
+        Node lem = nm->mkNode(kind::ITE, code_len, code_range, code_eq_neg1);
+        Trace("strings-lemma") << "Strings::Lemma CODE : " << lem << std::endl;
+        Trace("strings-assert") << "(assert " << lem << ")" << std::endl;
+        d_out->lemma(lem);
       }
```
4tXJ7f: Maybe keep an else case with `AlwaysAssert(false, "String Terms/str.code only in registerTerm.");` (or Unreachable)?

src/theory/strings/theory_strings.cpp
```diff
@@ -2013,64 +2073,165 @@ void TheoryStrings::checkNormalForms(){
...
+    }
+    if (hasProcessed())
+    {
+      return;
+    }
+    // now, ensure that str.code is injective
+    std::vector<Node> cmps;
+    cmps.insert(cmps.end(), nconst_codes.begin(), nconst_codes.end());
+    cmps.insert(cmps.end(), const_codes.begin(), const_codes.end());
+    std::reverse(cmps.begin(), cmps.end());
```
4tXJ7f: Nit: Instead of reversing the vector, you can get them from the original vectors in reverse order:
```
std::vector<Node> cmps;
cmps.insert(cmps.end(), const_codes.rbegin(), const_codes.rend());
cmps.insert(cmps.end(), nconst_codes.rbegin(), nconst_codes.rend());
```
Also, you could merge the first two statements to:
```
std::vector<Node> cmps(const_codes.rbegin(), const_codes.rend());
```

test/regress/regress1/strings/str-code-sat.smt2
ajreynol: Yes, good catch.

src/theory/strings/theory_strings.cpp
ajreynol: No this isn't necessary anymore, I've changed the policy for registerTerm now, everything gets passed to this method but for some terms we don't do anything.  I'll add a comment documenting registerTerm.

PR #1825, opened 2018-04-27
Make factoring inference more aggressive

### LINKED ISSUE #1752: CVC4 answers unsat for nonlinear formula but unknown when combined with UFs
Hi all,

I've been checking two formulas, F1 with only nonlinear:
```
(declare-fun c () Int)
(declare-fun a () Int)
(declare-fun b () Int)
(assert (> a 0))
(assert (= c (* a b)))
(assert (not (= b (div c a))))
(check-sat)
```
and F2 with nonlinear and UF:
```
(declare-fun c (Int) Int)
(declare-fun a (Int) Int)
(declare-fun b (Int) Int)
(assert (> (a 0) 0))
(assert (= (c 0) (* (a 0) (b 0))))
(assert (not (= (b 0) (div (c 0) (a 0)))))
(check-sat)
```
CVC4 quickly answers unsat for F1, but unknown for F2.
Is that expected? I didn't think using these UFs would make it so much harder.

ajreynol:
Makes it so that if we have a polynomial x*c + x*t + s > 0, we factor to infer k = c + t ^ k*x + s > 0. Previously we required that c was not a constant.

This fixes #1752.


ajreynol:
This is +40-7 on QF_NRA: https://www.starexec.org/starexec/secure/details/job.jsp?id=27812

For QF_NIA it is +276-378: https://www.starexec.org/starexec/secure/details/job.jsp?id=27815
I'll investigate this a bit more.

PR #1836, opened 2018-05-01
Make CegisUnif default to Cegis when no unif used

HanielB:
CegisUnif now inherits from Cegis and performs regular CEGIS when none of its functions-to-synthesize are being solved through unification techniques.

This PR is in preparation for the combination of both approaches to be done in the future.


src/theory/quantifiers/sygus/cegis_unif.cpp
```diff
@@ -83,6 +116,69 @@ bool CegisUnif::constructCandidates(const std::vector<Node>& enums,
                                     std::vector<Node>& candidate_values,
                                     std::vector<Node>& lems)
 {
+  NodeManager* nm = NodeManager::currentNM();
+  if (d_no_unif)
+  {
+    candidate_values.insert(
+        candidate_values.end(), enum_values.begin(), enum_values.end());
+    if (options::sygusDirectEval())
```
ajreynol: You should generally never copy and paste code.  The point of making this class inherit from cegis is so that this code can be encapsulated in a function that is called in both places.

src/theory/quantifiers/sygus/cegis_unif.cpp
```diff
@@ -36,43 +36,76 @@ bool CegisUnif::initialize(Node n,
...
-                      << "...\n";
-  d_sygus_unif.initialize(d_qe, candidates, d_enums, lemmas);
-  Assert(!d_enums.empty());
-  Trace("cegis-unif") << "Initialize " << d_enums.size() << " enumerators for "
-                      << d_candidate << "...\n";
-  /* initialize the enumerators */
-  for (const Node& e : d_enums)
+  /* Init UNIF util */
+  /* TODO combine with enumeration manager */
+  if (false)
```
ajreynol: Remove this.  You should make this code do the right thing (if necessary, modify the SygusUnifRL utility to not allocate things you can't handle).

src/theory/quantifiers/sygus/cegis_unif.cpp
```diff
@@ -36,43 +36,76 @@ bool CegisUnif::initialize(Node n,
...
   }
   return true;
 }
 
 void CegisUnif::getTermList(const std::vector<Node>& candidates,
-                            std::vector<Node>& terms)
+                            std::vector<Node>& enums)
 {
-  Assert(candidates.size() == 1);
+  if (d_no_unif)
```
ajreynol: This should not be a special case.

src/theory/quantifiers/sygus/cegis_unif.cpp
```diff
@@ -83,6 +116,69 @@ bool CegisUnif::constructCandidates(const std::vector<Node>& enums,
                                     std::vector<Node>& candidate_values,
                                     std::vector<Node>& lems)
 {
+  NodeManager* nm = NodeManager::currentNM();
+  if (d_no_unif)
```
ajreynol: Again, this should not be a special case.

src/theory/quantifiers/sygus/cegis_unif.cpp
```diff
@@ -153,14 +246,34 @@ Node CegisUnif::purifyLemma(Node n,
...
     if (i == 0 && fapp)
     {
-      children.push_back(n[0]);
+      /* Non unif functions are kept as is */
+      if (!d_sygus_unif.usingUnif(n[0]))
+      {
+        children.push_back(n[0]);
+        continue;
+      }
+      std::map<Node, Node>::const_iterator it = d_app_to_purified.find(n);
```
ajreynol: I think you need to check cache based on the node *after* it has been purified.  Instead, we are caching based on n which is not necessary in the form f( e, c1...cn ).

src/theory/quantifiers/sygus/cegis_unif.cpp
ajreynol: This should add each candidate c to enums if c is not a unificiation candidate

ajreynol: We should additionally add to enums all evaluation points.

src/theory/quantifiers/sygus/cegis.h
```diff
@@ -94,6 +94,18 @@ class Cegis : public SygusModule
...
                                 const std::vector<Node>& vals,
                                 std::vector<Node>& lems);
+
+  /** evaluates candidate values on current refinement lemmas
+   *
+   * Returns true if refinement lemmas are added after evaluation, false
+   * otherwise.
+   *
+   * Also eagerly unfolds evaluation functions in a heuristic manner, which is
+   * useful when evaluation functions have been axiomatized (e.g. for boolean
```
ajreynol: Remove "when evaluation functions have been axiomatized".

src/theory/quantifiers/sygus/cegis_unif.cpp
```diff
@@ -25,9 +26,10 @@ namespace theory {
 namespace quantifiers {
 
 CegisUnif::CegisUnif(QuantifiersEngine* qe, CegConjecture* p)
-    : SygusModule(qe, p)
+    : Cegis(qe, p), d_sygus_unif(p)
 {
   d_tds = d_qe->getTermDatabaseSygus();
+  d_enum_to_active_guard.clear();
```
ajreynol: This line is unnecessary.

src/theory/quantifiers/sygus/cegis_unif.cpp
```diff
@@ -36,43 +38,54 @@ bool CegisUnif::initialize(Node n,
                            std::vector<Node>& lemmas)
 {
   Trace("cegis-unif") << "Initialize CegisUnif : " << n << std::endl;
-  Assert(candidates.size() > 0);
-  if (candidates.size() > 1)
+  /* For regular CEGIS */
```
ajreynol: This is also copy/pasted code right? This should be encapsulated somewhere.

ajreynol: actually, you can just remove this code, this is for cegis-sample, which is likely incompatible with cegisUnif.

src/theory/quantifiers/sygus/cegis_unif.cpp
```diff
@@ -36,43 +38,54 @@ bool CegisUnif::initialize(Node n,
...
+      continue;
+    }
+    Valuation& valuation = d_qe->getValuation();
+    for (const Node& e : d_enums)
     {
-      terms.push_back(e);
+      Assert(d_enum_to_active_guard.find(e) != d_enum_to_active_guard.end());
+      Node g = d_enum_to_active_guard[e];
+      /* Get whether the active guard for this enumerator is if so, then there
+         may
```
ajreynol: nit: fix the layout here.

src/theory/quantifiers/sygus/cegis_unif.cpp
```diff
@@ -83,16 +96,21 @@ bool CegisUnif::constructCandidates(const std::vector<Node>& enums,
...
+    Trace("cegis-unif-lemma") << "Added eval lemmas\n";
     return false;
   }
   unsigned min_term_size = 0;
   std::vector<unsigned> enum_consider;
+  NodeManager* nm = NodeManager::currentNM();
   Trace("cegis-unif-enum") << "Register new enumerated values :\n";
   for (unsigned i = 0, size = enums.size(); i < size; ++i)
   {
+    if (std::find(d_enums.begin(), d_enums.end(), enums[i]) == d_enums.end())
```
ajreynol: Maybe document why we are doing this. It may also be helpful to rename d_enums -> d_cond_enums.

src/theory/quantifiers/sygus/sygus_unif_rl.cpp
```diff
@@ -52,14 +63,139 @@ void SygusUnifRl::notifyEnumeration(Node e, Node v, std::vector<Node>& lemmas)
...
+  Node nv = n;
+  /* get model value of non-top level applications of candidates */
+  if (ensureConst && fapp)
+  {
+    nv = d_parent->getModelValue(n);
+    Assert(n != nv);
+    Trace("sygus-unif-rl-purify") << "PurifyLemma : model value for " << n
+                                  << " is " << nv << "\n";
+  }
+  Assert(!createModelEq || nb != nv);
```
ajreynol: There is a syntax error here, nb is not defined.

src/theory/quantifiers/sygus/sygus_unif_rl.cpp
```diff
@@ -52,14 +63,139 @@ void SygusUnifRl::notifyEnumeration(Node e, Node v, std::vector<Node>& lemmas)
...
+    if (it == d_app_to_purified.end())
+    {
+      /* Build purified head with fresh skolem and recreate node */
+      std::stringstream ss;
+      ss << nb[0] << "_" << d_purified_count[nb[0]]++;
+      Node new_f = nm->mkSkolem(ss.str(), nb[0].getType());
+      /* Add new enumerator to candidate */
+      d_cand_to_pt_enum[nb[0]].push_back(new_f);
+      /* collect children and rebulid node */
+      children.clear();
```
ajreynol: i think these 10 lines can be replaced by children[1] = new_f;

src/theory/quantifiers/sygus/sygus_unif_rl.cpp
```diff
@@ -99,14 +235,14 @@ void SygusUnifRl::collectPoints(Node n)
 
 void SygusUnifRl::initializeConstructSol()
 {
-  if (d_hasRLemmas && d_rlemmas != d_prev_rlemmas)
+  if (d_rlemmas != d_prev_rlemmas)
```
ajreynol: I don't think we need this code any more. You can also remove collectPoints above.

src/theory/quantifiers/sygus/cegis_unif.h
```diff
@@ -19,57 +19,47 @@
...
-   *
-   * The module takes ownership of a conjecture when it contains a single
-   * function-to-synthesize
-  */
+  /** initialize this class */
   bool initialize(Node n,
                   const std::vector<Node>& candidates,
                   std::vector<Node>& lemmas) override;
-  /** adds the candidate itself to enums */
+  /** adds the candidates themselves to enums */
```
ajreynol: Update this comment.

src/theory/quantifiers/sygus/sygus_unif_rl.cpp
```diff
@@ -52,95 +63,161 @@ void SygusUnifRl::notifyEnumeration(Node e, Node v, std::vector<Node>& lemmas)
...
-        }
-        Trace("sygus-unif-rl-sol") << ")\n";
-      }
-    }
+    model_guards.push_back(plem);
+    plem = NodeManager::currentNM()->mkNode(OR, model_guards);
   }
+  plem = Rewriter::rewrite(plem);
+  Trace("sygus-unif-rl-lemma") << "Purified lemma : " << plem << "\n";
+  d_prev_rlemmas = d_rlemmas;
```
ajreynol: These lines are also unecessary now (remove d_prev_rlemmas and d_rlemmas).

src/theory/quantifiers/sygus/sygus_unif_rl.cpp
```diff
@@ -52,95 +63,161 @@ void SygusUnifRl::notifyEnumeration(Node e, Node v, std::vector<Node>& lemmas)
...
+  {
+    if (i == 0 && fapp)
     {
+      children.push_back(n[i]);
       continue;
     }
-    visited.insert(cur);
-    unsigned size = cur.getNumChildren();
-    if (cur.getKind() == APPLY_UF && size > 0)
+    Node child = purifyLemma(n[i], ensureConst || fapp, model_guards, cache);
```
ajreynol: I think we technically could make the recursive call be ```!nu_fapp && ( ensureConst || fapp )``` where ```nu_fapp``` is true if we are an evaluation function but it is not a unification one.

PR #1837, opened 2018-05-01
Cegis unif enumerator manager

ajreynol:
This is in preparation for a CegisUnif approach which will try to find minimally diverse models for the set of evaluation points.

### DIFF
--- a/src/theory/quantifiers/sygus/cegis_unif.cpp
+++ b/src/theory/quantifiers/sygus/cegis_unif.cpp
@@ -235,7 +235,8 @@ void CegisUnif::registerRefinementLemma(const std::vector<Node>& vars,
       NodeManager::currentNM()->mkNode(OR, d_parent->getGuard().negate(), lem));
 }
 
-CegisUnifEnumManager::CegisUnifEnumManager(QuantifiersEngine* qe) : d_qe(qe), d_curr_cfun_val( qe->getSatContext(), 0 )
+CegisUnifEnumManager::CegisUnifEnumManager(QuantifiersEngine* qe, CegConjecture* parent) : d_qe(qe), d_parent(parent),
+d_curr_cfun_val( qe->getSatContext(), 0 )
 {
   
 }
@@ -325,7 +326,7 @@ Node CegisUnifEnumManager::getOrMkLiteral( unsigned n )
       Node c = ci.first;
       Node eu = nm->mkSkolem( "eu", c.getType() );
       ci.second.d_enums.push_back( eu );
-      // register it TODO
+      d_tds->registerEnumerator(e, d_candidate, d_parent);
     }
     
     return new_lit;
@@ -346,6 +347,7 @@ void CegisUnifEnumManager::registerEvalPtAtCostFunValue( Node c, Node ei, Node l
     disj.push_back( ei.eqNode( itc->second.d_enums[i] ) );
   }
   Node lem = NodeManager::currentNM()->mkNode( OR, disj );
+  d_qe->getOutputChannel().lemma(lem);
 }
   
 }  // namespace quantifiers
--- a/src/theory/quantifiers/sygus/cegis_unif.cpp
+++ b/src/theory/quantifiers/sygus/cegis_unif.cpp
@@ -301,7 +301,7 @@ void CegisUnifEnumManager::incrementNumEnumerators()
   if (itc == d_guq_lit.end())
   {
     // allocate the new literal
-    NodeManager * nm = NodeManager::currentNM();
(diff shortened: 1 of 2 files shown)


src/theory/quantifiers/sygus/cegis_unif.h
```diff
@@ -152,6 +152,95 @@ class CegisUnif : public SygusModule
 
 }; /* class CegisUnif */
 
+/** Cegis Unif Enumeration Manager
+ *
+ * This function enforces a decision heuristic that limits the number of
```
HanielB: function --> class

src/theory/quantifiers/sygus/cegis_unif.h
```diff
@@ -152,6 +152,95 @@ class CegisUnif : public SygusModule
...
+   public:
+    TypeInfo() {}
+    /** candidates for this type */
+    std::vector<Node> d_candidates;
+    /** the set of enumerators we have allocated for this candidate */
+    std::vector<Node> d_enums;
+    /** the set of evaluation points of this type */
+    std::vector<Node> d_eval_points;
+  };
+  std::map<TypeNode, TypeInfo> d_ce_info;
```
HanielB: doc

src/theory/quantifiers/sygus/cegis_unif.h
```diff
@@ -152,6 +152,95 @@ class CegisUnif : public SygusModule
...
+  /** literals of the form G_uq_n for each n */
+  std::map<unsigned, Node> d_guq_lit;
+  /**
+   * The minimal n such that G_uq_n is asserted positively in the
+   * current SAT context.
+   */
+  context::CDO<unsigned> d_curr_guq_val;
+  /** increment the number of enumerators */
+  void incrementNumEnumerators();
+  /** get current cost fun literal */
```
HanielB: describe cost fun here or elsewhere

src/theory/quantifiers/sygus/cegis_unif.h
```diff
@@ -152,6 +152,95 @@ class CegisUnif : public SygusModule
...
+   * synthesis-by-unification approach for c.
+   */
+  void initialize(std::vector<Node>& cs);
+  /** register evaluation point for candidate
+   *
+   * This notifies this class that eis is a set of evaluation points for
+   * candidate c, where c should be a candidate that was passed to initialize
+   * in the vector cs.
+   */
+  void registerEvalPts(std::vector<Node>& eis, Node c);
```
HanielB: clarify that adding evaluation points triggers the creation of the lemmas constraining their values according to the current cost function

src/theory/quantifiers/sygus/cegis_unif.cpp
```diff
@@ -235,6 +235,133 @@ void CegisUnif::registerRefinementLemma(const std::vector<Node>& vars,
...
+}
+
+Node CegisUnifEnumManager::getOrMkLiteral(unsigned n)
+{
+  std::map<unsigned, Node>::iterator itc = d_guq_lit.find(n);
+  if (itc == d_guq_lit.end())
+  {
+    NodeManager* nm = NodeManager::currentNM();
+    // initialize it
+    Node new_lit = Rewriter::rewrite(nm->mkSkolem("G_cost", nm->booleanType()));
```
HanielB: why do this rewrite?

src/theory/quantifiers/sygus/cegis_unif.cpp
```diff
@@ -235,6 +235,133 @@ void CegisUnif::registerRefinementLemma(const std::vector<Node>& vars,
...
+    new_lit = d_qe->getValuation().ensureLiteral(new_lit);
+    AlwaysAssert(!new_lit.isNull());
+    d_qe->getOutputChannel().requirePhase(new_lit, true);
+    d_guq_lit[n] = new_lit;
+    // allocate an enumerator for each candidate
+    for (std::pair<const TypeNode, TypeInfo>& ci : d_ce_info)
```
HanielB: I think it makes more sense for this cade to be ``incrementNumEnumerators`` than here

src/theory/quantifiers/sygus/cegis_unif.cpp
ajreynol: the requirePhase function below requires that new_lit is in rewritten form.  This rewrite likely will never do anything, but it is for paranoid reasons.

src/theory/quantifiers/sygus/cegis_unif.cpp
ajreynol: I agree, I've reorganized things a bit to make this clear.

PR #1838, opened 2018-05-01
Add support for incremental eager bit-blasting.

mpreiner:
Eager incremental solving is achieved via solving under assumptions. As soon as incremental and eager bit-blasting is enabled, assertions are passed to the SAT solver as assumptions rather than assertions. This requires the eager SAT solver to support incremental solving, which is currently only supported by CryptoMiniSat.


mpreiner:
@4tXJ7f: I finally added a regression test. PTAL

aniemetz:
LGTM

PR #1847, opened 2018-05-02
Add the symmetry breaker module

PaulMeng:
A detailed description of the symmetry breaking constraints can be found [here](https://bitbucket.org/cvc4/relations/src/master/SymmetryBreaking/).

For a symmetry group of integer, real and bit-vectors variables {v1, ..., vn}, we generate ordered constraints: {v1 <= v2, ..., v{n-1} <= vn}.

For a symmetry group of variables of other types {v1, ..., vn}, we generate the following two kinds of constraints.
1. Triplet constraints
v_i = v_j => v_i = v_{j-1} for all 0 <= i < j-1 < j < n
2. Segment constraints
v_i = v_j => (v_0 = v_1 OR \ldots OR v_{i-1} = v_{i}) for all 1 <= i < j < n


src/preprocessing/passes/symmetry_breaker.h
```diff
@@ -0,0 +1,67 @@
...
+   * */
+  Node generateSymBkConstraints(std::vector<std::vector<Node> >& parts);
+
+ private:
+  /** True and false constant nodes */
+  Node d_trueNode;
+  Node d_falseNode;
+
+  /**
+   * Get the order kind for node depending on the type of node
```
ajreynol: Can you document the intuition for this?

src/preprocessing/passes/symmetry_breaker.h
```diff
@@ -0,0 +1,67 @@
...
+    d_falseNode = NodeManager::currentNM()->mkConst<bool>(false);
+  }
+
+  /**
+   * Destructor
+   * */
+  ~SymmetryBreaker() {}
+
+  /**
+   * This is to generate symmetry breaking constraints for partitions in parts.
```
ajreynol: Document what parts is (with regards to its properties of specifying a symmetry).

src/preprocessing/passes/symmetry_breaker.h
```diff
@@ -0,0 +1,67 @@
...
+   * The symmetry breaking constraints SB returned by this function conjuncted
+   * with the original assertions SB ^ C is equisatisfiable to the C.
+   *
+   * */
```
ajreynol: nit: change this to */

src/preprocessing/passes/symmetry_breaker.h
```diff
@@ -0,0 +1,67 @@
...
+#include <map>
+#include <string>
+#include <vector>
+#include "expr/node.h"
+
+namespace CVC4 {
+namespace preprocessing {
+namespace passes {
+
+class SymmetryBreaker
```
ajreynol: Document this class. Give a few sentence overview of its purpose.

ajreynol: In fact, you should put the description of this PR in a comment here.

ajreynol:
This is still support code.

src/preprocessing/passes/symmetry_breaker.cpp
```diff
@@ -0,0 +1,116 @@
...
+
+              for (unsigned int k = 0; k < i; ++k)
+              {
+                rhs = NodeManager::currentNM()->mkNode(
+                    kind::OR,
+                    rhs,
+                    NodeManager::currentNM()->mkNode(kd, part[k], part[k + 1]));
+              }
+              // Generate segment constraints
+              // v_i = v_j => (v_0 = v_1 OR \ldots OR v_{i-1} = v_{i}) for all 1
```
ajreynol: I don't think this comment is accurate (I'm not sure about the code).
The segment constraints we discussed were:
```v_i = v_j => v_i = v_i-(j-i) OR v_i = v_{i-1} for i<j``` where ```v_i = v_i-(j-i)``` is false if i-(i-j)<0.
Right?

Comment (ajreynol) [reply src/preprocessing/passes/symmetry_breaker.cpp]: Also, I would add the intutition of the two constraints in a sentence each:
A "triplet" constraints ensure that equivalence classes over v_1...v_n are in consecutive segments. The "segment" constraints ensure that the length of segments occur in descending order.  I would also rename these in the comments as "consecutive" and "length order" constraints respectively.

File: src/preprocessing/passes/symmetry_breaker.h [hunk 6]
```diff
@@ -0,0 +1,87 @@
...
+   * Destructor
+   * */
+  ~SymmetryBreaker() {}
+
+  /**
+   * This is to generate symmetry breaking constraints for partitions in parts.
+   * The symmetry breaking constraints SB returned by this function conjuncted
+   * with the original assertions SB ^ C is equisatisfiable to the C.
+   * */
+  Node generateSymBkConstraints(std::vector<std::vector<Node> >& parts);
```
ajreynol: Also, make this take ```const std::vector<std::vector<Node> >& parts```.  (You are not modifying parts.

src/preprocessing/passes/symmetry_breaker.cpp
PaulMeng: Fixed.

src/preprocessing/passes/symmetry_breaker.cpp
```diff
@@ -0,0 +1,119 @@
...
+      Kind kd = getOrderKind(part[0]);
+
+      if (kd != kind::EQUAL)
+      {
+        for (unsigned int i = 0; i < part.size() - 1; ++i)
+        {
+          // Generate less than or equal to constraints: part[i] <= part[i+1]
+          Node constraint =
+              NodeManager::currentNM()->mkNode(kd, part[i], part[i + 1]);
+          sbConstraint = NodeManager::currentNM()->mkNode(
```
ajreynol: Can you add constraint to a vector and make a single constraint at the end? It is inefficient to make all these intermediate nodes, same with below.

PaulMeng: Sure. 
But do I need to check if the vector constraints is empty or not before calling mkNode(kind::AND, constraints)? Or mkNode will take care of the empty case and also the single element?

ajreynol: Unfortunately, no, this is done in various places in the code, see: https://github.com/ajreynol/CVC4/blob/impRealIntParsePrint/src/theory/datatypes/theory_datatypes.cpp#L2185
for an example.  We're planning to make a utility for this at some point, but you should just write these 6 lines in your code for now.

src/options/smt_options.toml
```diff
@@ -342,6 +342,14 @@ header = "options/smt_options.h"
   type       = "bool"
   default    = "false"
   help       = "calculate sort inference of input problem, convert the input based on monotonic sorts"
+  
+[[option]]
+  name       = "symmetryBreaker"
+  category   = "regular"
+  long       = "symmetry-break"
```
4tXJ7f: It might be more consistent to call the option `--symmetry-breaker`.

src/options/smt_options.toml
```diff
@@ -342,6 +342,14 @@ header = "options/smt_options.h"
...
+  type       = "bool"
+  default    = "false"
+  help       = "generate symmetry breaking constraints after the symmetry detection and append them to the original assertions"
```
4tXJ7f: This option should probably automatically enable `--symmetry-detect`, too. Also, do we need those to be separate options?

src/preprocessing/passes/symmetry_breaker.h
```diff
@@ -0,0 +1,91 @@
...
+ ** Copyright (c) 2009-2018 by the authors listed in the file AUTHORS
+ ** in the top-level source directory) and their institutional affiliations.
+ ** All rights reserved.  See the file COPYING in the top-level source
+ ** directory for licensing information.\endverbatim
+ **
+ ** \brief Symmetry breaker for theories
+ **/
+
+#ifndef __CVC4__PREPROCESSING_PASSES_SYMMETRY_BREAKER_H_
+#define __CVC4__PREPROCESSING_PASSES_SYMMETRY_BREAKER_H_
```
4tXJ7f: `__CVC4__PREPROCESSING__PASSES__SYMMETRY_BREAKER_H_` (double underscores between namespaces).

src/preprocessing/passes/symmetry_breaker.cpp
```diff
@@ -0,0 +1,124 @@
...
+    {
+      Kind kd = getOrderKind(part[0]);
+
+      if (kd != kind::EQUAL)
+      {
+        for (unsigned int i = 0; i < part.size() - 1; ++i)
+        {
+          // Generate less than or equal to constraints: part[i] <= part[i+1]
+          Node constraint =
+              NodeManager::currentNM()->mkNode(kd, part[i], part[i + 1]);
```
4tXJ7f: Suggestion: a common pattern that we use is that we have a `NodeManager* nm = NodeManager::currentNM();` at the beginning of a function that uses the node manager a couple of times.

src/options/smt_options.toml
ajreynol: I agree we don't need separate options for this.

src/options/smt_options.toml
ajreynol: Yes, this is a better name.

ajreynol:
Regarding the regressions, Paul or I will do a separate PR that makes the symmetry breaking stuff into a preprocessing pass and enables it on the existing regressions in regress1/sym/.

src/options/smt_options.toml
PaulMeng: I changed the name to --symmetry-breaker-exp for now because "--symmetry-breaker" has been taken.

src/options/smt_options.toml
PaulMeng: I removed --symmetry-detect and only use symmetry-breaker-exp to enable the symmetry breaking.

src/preprocessing/passes/symmetry_breaker.h
PaulMeng: Done

src/preprocessing/passes/symmetry_breaker.cpp
PaulMeng: Done

PR #1860, opened 2018-05-03
Refactor bv-abstraction preprocessing pass.

mpreiner:


src/smt/smt_engine.cpp
```diff
@@ -4035,7 +4018,7 @@ void SmtEnginePrivate::processAssertions() {
   if ( options::bvAbstraction() &&
       !options::incrementalSolving()) {
     dumpAssertions("pre-bv-abstraction", d_assertions);
```
4tXJ7f: The PreprocessingPass class automatically adds support for dumping before and after the pass, so those are not needed.

src/preprocessing/passes/bv_abstraction.cpp
```diff
@@ -0,0 +1,67 @@
...
+    : PreprocessingPass(preprocContext, "bv-abstraction"){};
+
+PreprocessingPassResult BvAbstraction::applyInternal(
+    AssertionPipeline* assertionsToPreprocess)
+{
+  std::vector<Node> new_assertions;
+  std::vector<Node> assertions(assertionsToPreprocess->begin(),
+                               assertionsToPreprocess->end());
+  TheoryEngine* te = d_preprocContext->getTheoryEngine();
+  bv::TheoryBV* bv_theory = (bv::TheoryBV*)te->theoryOf(THEORY_BV);
```
4tXJ7f: Use a `static_cast` for this conversion.

src/preprocessing/passes/bv_abstraction.cpp
```diff
@@ -0,0 +1,67 @@
...
+  for (unsigned i = 0, size = assertionsToPreprocess->size(); i < size; ++i)
+  {
+    assertionsToPreprocess->replace(i, Rewriter::rewrite(new_assertions[i]));
+  }
+  // If we are using the lazy solver and the abstraction applies, then UF
+  // symbols were introduced.
+  if (options::bitblastMode() == bv::BITBLAST_MODE_LAZY && changed)
+  {
+    LogicRequest req(*d_preprocContext->getSmt());
+    req.widenLogic(theory::THEORY_UF);
```
4tXJ7f: Actually, I think that it might be a good idea to add a method `PreprocessingPassContext::widenLogic()` that does this. With that, we can capture the instances where we add another theory in a preprocessing pass more explicitly and can hopefully add some feature in the future to intelligently detect issues with such a request.

src/preprocessing/passes/bv_abstraction.cpp
```diff
@@ -0,0 +1,67 @@
...
+ **     Bit-vectors in Satisfiability Modulo Theories
+ **     https://cs.nyu.edu/media/publications/hadarean_liana.pdf
+ **/
+
+#include "preprocessing/passes/bv_abstraction.h"
+
+#include <vector>
+
+#include "options/bv_options.h"
+#include "theory/bv/abstraction.h"
```
4tXJ7f: Is this include needed?

PR #1861, opened 2018-05-03
Initialize cegis unif strategy

ajreynol:
Initialized cegis unif strategy by recognizing ITE loops in the strategy tree of the function-to-synthesize. Currently, we only check whether there is a top-level strategy where this is possible.

This PR also improves various functions in SygusUnifStrat (it decouples staticLearnRedundantOps/finishInit/debugPrint).


src/theory/quantifiers/sygus/sygus_unif_rl.cpp
```diff
@@ -237,6 +272,83 @@ bool SygusUnifRl::usingUnif(Node f)
...
+      for (unsigned c = 1; c <= 2; c++)
+      {
+        std::pair<Node, NodeRole> child = etis->d_cenum[c];
+        if (child.first != e || child.second != nrole)
+        {
+          success = false;
+          break;
+        }
+      }
+      if (success)
```
HanielB: remove the boolean variable and test for c == 3

PR #1871, opened 2018-05-03
Move Lazy trie datastructure to its own file

HanielB:
Preparation for further developing CegisUnif

### DIFF
--- a/src/theory/quantifiers/lazy_trie.cpp
+++ b/src/theory/quantifiers/lazy_trie.cpp
@@ -1,5 +1,5 @@
 /*********************                                                        */
-/*! \file sygus_sampler.cpp
+/*! \file lazy_trie.cpp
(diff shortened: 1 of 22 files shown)


src/theory/quantifiers/lazy_trie.cpp
```diff
@@ -0,0 +1,65 @@
+/*********************                                                        */
+/*! \file sygus_sampler.cpp
```
ajreynol: lazy_trie.

src/theory/quantifiers/lazy_trie.h
```diff
@@ -0,0 +1,105 @@
+/*********************                                                        */
+/*! \file sygus_sampler.h
```
ajreynol: lazy_trie

PR #1812, opened 2018-04-23
Sygus repair constants

ajreynol:
Implements techniques for repairing constants in cegis.

The option is disabled by default for now.

PR #1873, opened 2018-05-04
Infrastructure for CEGQI handled status 

ajreynol:
Moves several functions to CegInstantiation, makes them static, and introduces enumeration types for them.


ajreynol:
This is a prerequisite for #1812.

src/theory/quantifiers/cegqi/ceg_instantiator.h
```diff
@@ -186,6 +186,24 @@ enum CegInstPhase
 
 std::ostream& operator<<(std::ostream& os, CegInstPhase phase);
 
+/**
+ * The handled status of a sort/term/quantified formula, indicating whether
+ * counterexample-guided instantiation handles it.
+ */
+enum CegHandledStatus
```
HanielB: Clarify at a high level what is the difference between each of these states. What separates e.g. UNHANDLED from PARTIALLY_UNHANDLED?

src/theory/quantifiers/cegqi/ceg_instantiator.h
```diff
@@ -297,14 +315,49 @@ class CegInstantiator {
...
+  //------------------------------------ static queries
   /** Is k a kind for which counterexample-guided instantiation is possible?
    *
-  * This typically corresponds to kinds that correspond to operators that
-  * have total interpretations and are a part of the signature of
-  * satisfaction complete theories (see Reynolds et al., CAV 2015).
-  */
-  static bool isCbqiKind(Kind k);
-
+   * If this method returns -1, then we prohibit CBQI for terms involving this
```
HanielB: Use the actual enum values

src/theory/quantifiers/cegqi/ceg_instantiator.cpp
```diff
@@ -127,21 +141,219 @@ bool CegInstantiator::isEligible( Node n ) {
...
+        }
+      }
+    }
+  }
+  else if (tn.isSort())
+  {
+    QuantEPR* qepr = qe != nullptr ? qe->getQuantEPR() : nullptr;
+    if (qepr != nullptr)
+    {
+      if (qepr->isEPR(tn))
```
HanielB: join with above ``if``

src/theory/quantifiers/cegqi/ceg_instantiator.h
ajreynol: it depends on the context, I've added further documentation above isCbqiQuant.

HanielB: I think this would be more clear if you did not use "HANDLED" for "EXCLUSIVE" and "PARTIALLY_HANDLED" to "SHARED WITH OTHER INST TECHNIQUES." 

Also it's still not clear, looking only at the documentation, what UNHANDLED_UNCONDITIONAL means.

PR #1875, opened 2018-05-04
Remove special case for record selector printing

### LINKED ISSUE #1867: Index out of bounds when dumping APPLY_SELECTOR_TOTAL
```
$ builds/bin/cvc4 --dump raw-benchmark test/regress/regress0/bug586.cvc
OPTION "incremental" false;
OPTION "logic" "ALL";
DATATYPE
  role = r1 | r2 | r3
END;
emptyRoleSet : SET OF role;
ASSERT emptyRoleSet = {} :: SET OF role;
d : ARRAY role OF [# pos:SET OF role, neg:SET OF role #];
ASSERT d[r3]Assertion failure.
void CVC4::printer::cvc::CvcPrinter::toStream(std::ostream&, CVC4::TNode, int, bool, bool) const
../../../../src/printer/cvc/cvc_printer.cpp:412:
  sindex>=0
```

ajreynol:
Fixes #1867.


ajreynol:
It's a different issue, but I'll see if I can address it as well.

ajreynol:
It's not easy to fix, let's merge this.

Issue #1005, opened 2017-08-19
CVC4 does not correctly distinguish between integer and real constants (Bugzilla #792)


mpreiner:
**Imported from Bugzilla**
Reporter: *Clark Barrett*
Status: *NEW*
Severity: *minor*
Assigned to: *Clark Barrett*
Component: *SMT-LIB support*
Milestone: *---*
Version: *master*
Platform: *PC*
OS: *Linux*

Original attachment names and IDs:
- _[20160805-Int_vs_Real_constants.smt2](http://cvc4.cs.stanford.edu/bugzilla_attachments/427_20160805-Int_vs_Real_constants.smt2)_ (File that exhibits incorrect syntax accepted by CVC4, ID 427)

On 2017-04-11 10:13:54 -0700, Clark Barrett wrote:
> Created attachment 427
> File that exhibits incorrect syntax accepted by CVC4
> 
> According to SMT\-LIB in logics that mix integers and reals, 1 is an integer while 1.0 is a real.  CVC4 does not distinguish between these and thus it accepts input which is not strictly correct according to the standard.  An example is attached.

PR #1879, opened 2018-05-04
Improvements in parsing and printing related to mixed int/real

### LINKED ISSUE #1852: Assertion failure when dumping datatypes regression
```
$ builds/bin/cvc4 --preprocess-only --lang=smt2.5 --dump raw-benchmark test/regress/regress0/bug541.smt2
(set-option :incremental false)
(set-logic ALL_SUPPORTED)
(declare-datatypes () ((__cvc4_tuple (__cvc4_tuple_ctor))
))
(declare-sort T1 0)
(declare-sort T2 0)
(declare-datatypes () ((Pair (mk-pair (first T1) (second T2)))
))
CVC4 was terminated by the C++ runtime.
Perhaps an exception was thrown during stack unwinding.  (Don't do that.)
terminate called after throwing an instance of 'CVC4::AssertionException'
  what():  Assertion failure
void CVC4::expr::NodeValue::dec()
../../../../src/expr/node_value.h:442:
  NodeManager::currentNM() != __null
No current NodeManager on destruction of NodeValue: maybe a public CVC4 interface function is missing a NodeManagerScope ?
Aborted (core dumped)
```

ajreynol:
This eliminates some hacks for dealing with Int/Real.
- Eliminates the use of "to_real" to cast decimals like "2.0" that happen to be Int. We now replace these by (/ 2 1) instead of (to_real 2), which has the advantage of being smt-lib compliant for all theories, including QF_LRA.
- Eliminates the use of a hack to use "type ascriptions" when returning values from a get-value command. Instead, we use division with 1 when necessary.  This affects the output of a few regressions, but we remain smt-lib compliant.
- Addresses a bug with printing arbitrary type ascriptions for smt2 terms. This partially addresses #1852.
- Updates our printing of negative rationals to be (/ (- n) m) instead of (- (/ n m)), which is consistent with the smt lib standard for real values (http://smtlib.cs.uiowa.edu/theories-Reals.shtml).


ajreynol:
Possibly fixes #1005.

src/smt/command.cpp
```diff
@@ -1410,10 +1410,8 @@ void GetValueCommand::invoke(SmtEngine* smtEngine)
       {
         // Need to wrap in special marker so that output printers know this
```
4tXJ7f: It is not a special marker anymore but just a standard division.

src/printer/smt2/smt2_printer.cpp
```diff
@@ -1707,30 +1694,29 @@ static void toStream(std::ostream& out, const DefineFunctionRecCommand* c)
...
+    {
+      out << ")";
+    }
   }else{
     out << "(/ ";
     if(neg) {
+      out << "(- ";
       Rational abs_r = (-r);
       out << abs_r.getNumerator();
       if(decimal) { out << ".0"; }
```
4tXJ7f: Do we need this? Isn't a division always of type real?

src/printer/smt2/smt2_printer.cpp
```diff
@@ -1707,30 +1694,29 @@ static void toStream(std::ostream& out, const DefineFunctionRecCommand* c)
...
+  // Notice that we print (/ (- 5) 3) instead of (- (/ 5 3)),
+  // the former is compliant with real values in the smt lib standard.
   if(r.isIntegral()) {
     if (neg) {
+      out << "(- ";
       out << (-r);
     }else {
       out << r;
     }
     if (decimal) { out << ".0"; }
```
4tXJ7f: Can this case actually occur?

src/printer/smt2/smt2_printer.cpp
```diff
@@ -322,30 +322,42 @@ void Smt2Printer::toStream(std::ostream& out,
...
+  if (n.getKind() == kind::APPLY_TYPE_ASCRIPTION)
   {
-    if (n.getType() != force_nt)
+    force_nt = TypeNode::fromType(
+        n.getOperator().getConst<AscriptionType>().getType());
+    type_asc_arg = n[0];
+  }
+  else if (!force_nt.isNull() && n.getType() != force_nt)
+  {
+    type_asc_arg = n;
```
4tXJ7f: Shouldn't this be `type_asc_arg = force_nt`?

src/printer/smt2/smt2_printer.cpp
```diff
@@ -322,30 +322,42 @@ void Smt2Printer::toStream(std::ostream& out,
...
-        Node nn = NodeManager::currentNM()->mkNode(
-            kind::APPLY_TYPE_ASCRIPTION,
-            NodeManager::currentNM()->mkConst(
-                AscriptionType(force_nt.toType())),
-            n);
-        toStream(out, nn, toDepth, types, TypeNode::null());
-      }
-      return;
+      out << "(" << smtKindString(
+                        force_nt.isInteger() ? kind::TO_INTEGER : kind::TO_REAL,
```
4tXJ7f: Isn't it problematic that we output a to_int/to_real here? Shouldn't we use the division trick?

ajreynol: The issue with ``/`` is that it is not allowed when n is not a constant, when the logic is linear.
It always works if we are NIA, NRA, or NIRA, and it always works if n is constant.

The issue with ``to_real`` as mentioned is that it is only allowed in the mixed Int/Real theory.
It always works if we are LIRA, or NIRA.

I think it may be better to do the division trick for now.

src/printer/smt2/smt2_printer.cpp
ajreynol: No, force_nt is a type, type_asc_arg is a term, the latter corresponds to the "argument of the type ascription", see comments above this line.

src/printer/smt2/smt2_printer.cpp
ajreynol: You're right its not necessary.

src/printer/smt2/smt2_printer.cpp
ajreynol: I think it can occur but its not necessary.

4tXJ7f: Maybe we can remove the `decimal` argument?

src/printer/smt2/smt2_printer.cpp
4tXJ7f: You are right, I misread, sorry.

PR #1884, opened 2018-05-07
Infrastructure for approximations in model output

ajreynol:
This adds infrastructure to bookkeep cases when our model construction is approximate.  The motivating use case for this feature is trancendental/irrational-valued functions.  This has been requested by GE.

As a simple example, for input:
```
(set-logic QF_NRA)
(declare-fun x () Real)
(assert (= (* x x) 2))
(assert (> x 1.414))
(check-sat)
```
After a forthcoming merge of this branch:  https://github.com/ajreynol/CVC4/tree/checkModelVts, CVC4 outputs:
```
sat
(model
(approximation (and (>= x (/ 741455 524288)) (>= (* (- 1.0) x) (- (/ 1482917 1048576)))))
(define-fun x () Real (/ 741455 524288))
)
```

This PR:
- Adds a method "recordApproximation" in TheoryModel.
- Prints approximations in the smt2 printer.
- Fixes two minor related issues: proper printing of PI in smt2, and disabling debugCheckModel in debug mode if --check-model is not set.


ajreynol:
@barrettcw, you might be interested in this.

src/theory/theory_engine.cpp
```diff
@@ -645,7 +645,7 @@ void TheoryEngine::check(Theory::Effort effort) {
         {
           // if we are incomplete, there is no guarantee on the model.
           // thus, we do not check the model here. (related to #1693)
-          if (!d_incomplete)
+          if (!d_incomplete && options::checkModels())
```
HanielB: change the above comment to reflect the new behavior

src/theory/theory_model.cpp
```diff
@@ -93,6 +95,19 @@ bool TheoryModel::getHeapModel( Expr& h, Expr& neq ) const {
   }
 }
 
+bool TheoryModel::hasApproximations() const { return !d_approx_list.empty(); }
+
+std::vector<std::pair<Expr, Expr> > TheoryModel::getApproximations() const
+{
+  std::vector<std::pair<Expr, Expr> > approx;
+  for (unsigned i = 0, size = d_approx_list.size(); i < size; i++)
```
HanielB: foreach

PR #1888, opened 2018-05-08
Generalize symmetry detection for 1 symmetry variable mapped to n input variables 

ajreynol:
This generalizes the symmetry detection algorithm to merge children of commutative operators whose symmetries are automorphisms over {x1...xn} for any n.  This generalizes the current code which was capable of merging automorphisms over singleton sets {x1}.


ajreynol:
This is waiting on #1890.

ajreynol:
After this PR, CVC4_REGRESSION_ARGS=--symmetry-breaker-exp passes regress1.

src/preprocessing/passes/symmetry_detect.cpp
```diff
@@ -21,22 +21,483 @@ using namespace std;
...
+    {
+      if (d_base_vars.find(v) == d_base_vars.end() && v != merge_var)
+      {
+        if (merge_var.isNull() && curr_merge_var.isNull())
+        {
+          curr_merge_var = v;
+        }
+        else
+        {
+          // cannot include
```
PaulMeng: Here, we cannot include, is it because the current partition p have 2 or more new variables in its partition?

src/preprocessing/passes/symmetry_detect.cpp
```diff
@@ -21,22 +21,483 @@ using namespace std;
...
+        Trace("sym-dt-debug2")
+            << "...cannot include (not enough room for new merge var "
+            << num_v_max << "<" << d_num_new_indices_needed << ")\n";
+        include_success = false;
+      }
+    }
+    else if (!include_success && !merge_var.isNull())
+    {
+      // decrement
+      num_merge_var_max--;
```
PaulMeng: If we reach here, we do not find a candidate and also include failed before and we had a candidate "merge_var" (not null) to merge in last round. But why decrement "num_merge_var_mar"? Is it because in last success round, we add the last successful index to "new_indices" and set "num_merge_var_max" to its max "num_v_max" but did not decrement "num_merge_var_max"?

src/preprocessing/passes/symmetry_detect.cpp
```diff
@@ -313,147 +801,162 @@ void SymmetryDetect::collectChildren(Node node, vector<Node>& children)
...
-  {
-    Trace("sym-dt") << "[sym-dt] Partition: " << sub_vars_it->first << " -> {";
+  // initialize partition merger class
+  PartitionMerger pm;
+  pm.initialize(k, partitions, indices);
 
-    for (vector<Node>::iterator part_it = (sub_vars_it->second).begin();
-         part_it != (sub_vars_it->second).end();
-         ++part_it)
+  for (unsigned index : indices)
```
PaulMeng: Should we also categorize the "indices" here by types so that we only compare indices of the same type and reduce the amount of comparisons?

src/preprocessing/passes/symmetry_detect.cpp
ajreynol: Yes, recall in this method, we are trying to find a set of partitions that suffice to add exactly one new variable to the region. This case holds when adding the current region would cause 2+ variables to be added, hence, we cannot add this region.

src/preprocessing/passes/symmetry_detect.cpp
ajreynol: This case runs when we have selected a candidate merge_var to add, the current region has merge_var, but fails to meet the criteria to be included.  

The value of "num_merge_var_max" is an upper bound on the number of regions that we might choose to include in the merge that contain merge_var. This is an optimization to realize when the current merge_var is infeasible. 

In this case, we are choosing not to merge a region that contains merge_var, thus the value of num_merge_var_max decreases by one.

src/preprocessing/passes/symmetry_detect.cpp
ajreynol: Yes, good catch.  I think the code does not currently do this.  This indeed would be a good optimization. After merging this PR, I will add this.

src/preprocessing/passes/symmetry_detect.h
```diff
@@ -25,6 +25,176 @@
...
+  std::vector<Node> d_variables;
+
+  /** The mapping from a node to its children */
+  std::map<Node, PartitionTrie> d_children;
+
+  /** Add variable v to the trie, indexed by
+    * parts[0].var_to_subvar[v]....parts[n].var_to_subvar[v]. */
+  Node addNode(Node v, std::vector<Partition>& parts);
+
+  /** Get all the new regions of a partition and store in part */
```
HanielB: elaborate

src/preprocessing/passes/symmetry_detect.cpp
```diff
@@ -21,22 +21,483 @@ using namespace std;
...
+  bool merged = false;
+  bool success = false;
+  unsigned base_choose = d_base_vars.size() - 1;
+  unsigned base_occurs_req = d_base_vars.size();
+  do
+  {
+    Trace("sym-dt-debug") << "   base variables must occur " << base_occurs_req
+                          << " times." << std::endl;
+    // check if all the base_vars occur at least the required number of
+    // times
```
HanielB: make this one line, it fits

src/preprocessing/passes/symmetry_detect.cpp
```diff
@@ -21,22 +21,483 @@ using namespace std;
...
+      Trace("sym-dt-debug") << "   ...merged: " << merge_var << std::endl;
+      Assert(!merge_var.isNull());
+      merged = true;
+      success = true;
+      // update the number of new indicies needed
+      if (base_choose > 0)
+      {
+        d_num_new_indices_needed +=
+            nChoosek(d_base_vars.size(), base_choose - 1);
+        // base_occurs_req =   TODO
```
HanielB: details

src/preprocessing/passes/symmetry_detect.cpp
```diff
@@ -21,22 +21,483 @@ using namespace std;
...
+      new_indices.pop_back();
+      // if we included with the merge var, no use trying not including
+      if (curr_merge_var.isNull() && !merge_var.isNull())
+      {
+        Trace("sym-dt-debug2") << "...fail (failed merge var)\n";
+        return false;
+      }
+    }
+  }
+  // if we haven't chosen a merge variable, we may be out of time TODO
```
HanielB: out of time?

src/preprocessing/passes/symmetry_detect.cpp
```diff
@@ -21,22 +21,483 @@ using namespace std;
...
+  bool isComm = theory::quantifiers::TermUtil::isComm(k);
+  if (isComm)
+  {
+    std::sort(children.begin(), children.end());
+  }
+  Node sn;
+  for (const Node& sc : children)
+  {
+    Assert(!sc.isNull());
+    if (sn.isNull())
```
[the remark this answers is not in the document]
ajreynol: This is sn not sc?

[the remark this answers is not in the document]
HanielB: oh, sorry about that

src/preprocessing/passes/symmetry_detect.cpp
```diff
@@ -21,22 +21,483 @@ using namespace std;
 namespace CVC4 {
 namespace preprocessing {
 namespace passes {
+namespace symbreak {
 
-SymmetryDetect::Partition SymmetryDetect::detect(const vector<Node>& assertions)
+/** returns n choose k, that is, n!/(k! * (n-k)!) */
```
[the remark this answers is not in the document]
ajreynol: n!/(k! * (n-k)!) is n factorial divided by k factorial times (n-k) factorial?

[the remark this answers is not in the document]
HanielB: aaaah, of course! I did not know that was how this was said in English (n choose k), so I was trying to parse with the wrong context :)

PR #1890, opened 2018-05-09
Make symmetry-breaker-exp into a preprocessing pass

ajreynol:
Also fixes options on the sym regressions, which currently is breaking regress1.

### DIFF
--- a/src/smt/smt_engine.cpp
+++ b/src/smt/smt_engine.cpp
@@ -2598,10 +2598,10 @@ void SmtEnginePrivate::finishInit() {
       new IntToBV(d_preprocessingPassContext.get()));
   std::unique_ptr<PseudoBooleanProcessor> pbProc(
       new PseudoBooleanProcessor(d_preprocessingPassContext.get()));
-  std::unique_ptr<SymBreakerPass> sbProc(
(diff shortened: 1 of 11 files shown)


src/smt/smt_engine.cpp
```diff
@@ -2598,6 +2598,8 @@ void SmtEnginePrivate::finishInit() {
       new IntToBV(d_preprocessingPassContext.get()));
   std::unique_ptr<PseudoBooleanProcessor> pbProc(
       new PseudoBooleanProcessor(d_preprocessingPassContext.get()));
+  std::unique_ptr<SymBreakerPass> sbProc(
```
HanielB: fix order

src/smt/smt_engine.cpp
```diff
@@ -2610,6 +2612,7 @@ void SmtEnginePrivate::finishInit() {
   d_preprocessingPassRegistry.registerPass("int-to-bv", std::move(intToBV));
   d_preprocessingPassRegistry.registerPass("pseudo-boolean-processor",
                                            std::move(pbProc));
+  d_preprocessingPassRegistry.registerPass("sym-break", std::move(sbProc));
```
HanielB: ditto

PR #1891, opened 2018-05-09
Better option names for PBE

ajreynol:


src/theory/datatypes/datatypes_sygus.cpp
```diff
@@ -662,8 +662,12 @@ bool SygusSymBreakNew::registerSearchValue( Node a, Node n, Node nv, unsigned d,
         // class
         // is it equivalent under examples?
         Node bvr_equiv;
-        if (aconj->getPbe()->hasExamples(a)) {
-          bvr_equiv = aconj->getPbe()->addSearchVal(tn, a, bvr);
+        if (options::sygusSymBreakPbe())
+        {
+          if (aconj->getPbe()->hasExamples(a))
```
HanielB: can you join these ``if`` or ``aconj->getPbe()`` would fail if the option is not set?

ajreynol: i'd get a segfault since aconj->getPbe() would be null.

PR #1895, opened 2018-05-10
Floating point theory solver based on SymFPU

martin-cs:
With this patch CVC4 is (as far as we can tell) the fastest, most correct and most complete floating-point solver.  The patch set is quite large and complex and is best reviewed a patch at a time.  It does:

1. Includes a copy of SymFPU with an appropriate licence (alternatively a script can pull this from the SymFPU repository).
2. Uses a literal SymFPU back-end (wrapping CVC4::BitVector) to implement the arbitrary precision floating-point type (CVC4::FloatingPoint).
3. Adds new kinds for the components of a word-blasted float.
4. Uses a symbolic SymFPU back-end (wrapping Node) to build equivalent expression in terms of bit-vectors.
5. Links equalities and classification functions at the model level so that the equality and bit-vector engines can work together.
6. Implements conversions to and from real via lazy conversion into ordering statements.

( Paging @aniemetz @mpreiner and @4tXJ7f for reviewing assistance, I know there are some style issues, which I will work on. )


src/theory/fp/theory_fp.cpp
```diff
@@ -309,31 +309,37 @@ Node TheoryFp::toRealUF(Node node) {
...
-Node TheoryFp::expandDefinition(LogicRequest &lr, Node node) {
-  Trace("fp-expandDefinition")
-      << "TheoryFp::expandDefinition(): " << node << std::endl;
-
+void TheoryFp::enableUF(LogicRequest &lr) {
   if (!this->d_expansionRequested) {
     lr.widenLogic(
         THEORY_UF);  // Needed for conversions to/from real and min/max
-    lr.widenLogic(THEORY_BV);
+    //lr.widenLogic(THEORY_BV); // This has to be done when the logic is set
```
aniemetz: Remove?

src/util/floatingpoint.cpp
```diff
@@ -14,9 +14,50 @@
...
+#include "symfpu/core/sqrt.h"
+#include "symfpu/core/convert.h"
+#include "symfpu/core/remainder.h"
+#include "symfpu/core/ite.h"
+#include "symfpu/utils/properties.h"
+#include "symfpu/utils/numberOfRoundingModes.h"
+
+namespace symfpu {
+
+#define CVC4LITITEDFN(T) template <>					\
```
aniemetz: Use `CVC4_XXX` for macros, e.g., `CVC4_LIT_ITE_DFN`.

src/util/floatingpoint.cpp
```diff
@@ -14,9 +14,50 @@
...
+  struct ite<::CVC4::symfpuLiteral::prop, T> {				\
+    static const T & iteOp (const ::CVC4::symfpuLiteral::prop &cond,	\
+			    const T &l,					\
+			    const T &r) {				\
+      if (cond) {							\
+	return l;							\
+      } else {								\
+	return r;							\
```
aniemetz: `return cond ? l : r;`

src/util/floatingpoint.cpp
```diff
@@ -32,20 +73,307 @@ FloatingPointSize::FloatingPointSize (const FloatingPointSize &old) : e(old.e),
...
+    template <bool isSigned>
+    prop wrappedBitVector<isSigned>::isAllZeros() const {return (*this == wrappedBitVector<isSigned>::zero(this->getWidth()));}
+
+
+    template <bool isSigned>
+    wrappedBitVector<isSigned> wrappedBitVector<isSigned>::maxValue (const bwt &w) {
+      if (isSigned) {
+	BitVector base(w-1, 0U);
+	return wrappedBitVector<true>((~base).zeroExtend(1));
+      } else {
```
aniemetz: `else` obsolete since then-branch returns.

src/util/floatingpoint.cpp
```diff
@@ -32,20 +73,307 @@ FloatingPointSize::FloatingPointSize (const FloatingPointSize &old) : e(old.e),
...
+	return wrappedBitVector<false>::allOnes(w);
+      }
+    }
+
+    /*
+    template <>
+    wrappedBitVector<false> wrappedBitVector<false>::maxValue (const bwt &w) {
+      return wrappedBitVector<false>::allOnes(w);
+    }
+    */
```
aniemetz: Remove?

src/util/floatingpoint.h
```diff
@@ -115,30 +117,164 @@ namespace CVC4 {
...
+    template <> struct signedToLiteralType< true> {
+      typedef int literalType;
+    };
+    template <> struct signedToLiteralType<false> {
+      typedef  unsigned int literalType;
+    };
+
+
+    template <bool isSigned>
+    class wrappedBitVector : public BitVector {
```
aniemetz: Reorder according to the style guidelines:
https://github.com/CVC4/CVC4/wiki/Code-Style-Guidelines#names-and-order-of-includes

src/theory/fp/theory_fp_rewriter.cpp
```diff
@@ -35,6 +35,7 @@
 
 #include "base/cvc4_assert.h"
 #include "theory/fp/theory_fp_rewriter.h"
+#include "theory/fp/fp_converter.h"
```
aniemetz: Order includes alphabetically, this applies to all header files.

src/theory/fp/theory_fp_rewriter.cpp
```diff
@@ -788,6 +789,95 @@ namespace constantFold {
...
+      result = arg0.getLiteral().inf;
+      break;
+    case kind::FLOATINGPOINT_COMPONENT_ZERO :
+      result = arg0.getLiteral().zero;
+      break;
+    case kind::FLOATINGPOINT_COMPONENT_SIGN :
+      result = arg0.getLiteral().sign;
+      break;
+    default :
+      Unreachable("Unknown kind used in constantoldComponentFlag");
```
aniemetz: Is this supposed to be "constantoldComponentFlag" or is it a typo (and it should read "constantFoldComponentFlag") or should this be "componentFlag"?

src/theory/fp/theory_fp.cpp
```diff
@@ -411,7 +419,13 @@ void TheoryFp::convertAndEquateTerm(TNode node) {
 #endif
 
     } else {
-      Assert((node.getKind() == kind::EQUAL));
+      // Component bits should not be altered.
+      // These are the only bits that should be allowed through
```
aniemetz: Is this a typo? Else, what does this sentence mean?

src/theory/fp/fp_converter.cpp
```diff
@@ -12,25 +12,1029 @@
...
+#include "symfpu/core/compare.h"
+#include "symfpu/core/classify.h"
+#include "symfpu/core/convert.h"
+#include "symfpu/core/remainder.h"
+
+
+namespace symfpu {
+  using namespace ::CVC4::theory::fp::symfpuSymbolic;
+
+#define CVC4SYMITEDFN(T) template <>					\
```
aniemetz: Again, use `CVC4_XXX`.

src/theory/fp/fp_converter.cpp
```diff
@@ -12,25 +12,1029 @@
...
+      ::CVC4::Node cond = _cond.getNode();				\
+      ::CVC4::Node l = _l.getNode();					\
+      ::CVC4::Node r = _r.getNode();					\
+									\
+      /* Handle some common symfpu idioms */				\
+      if (cond.isConst()) {						\
+	if (cond == nm->mkConst(::CVC4::BitVector(1U, 1U))) {		\
+	  return l;							\
+	} else {							\
+	  return r;							\
```
aniemetz: Again, you could use the shorthand `... ? ... : ... ;` here.

src/theory/fp/fp_converter.cpp
```diff
@@ -12,25 +12,1029 @@
...
+			      nm->mkNode(::CVC4::kind::BITVECTOR_AND,	\
+					 nm->mkNode(::CVC4::kind::BITVECTOR_NOT, cond), \
+					 r[0]),				\
+			      r[1],					\
+			      l);					\
+	  }								\
+	  								\
+	}								\
+      }									\
+      return T(::CVC4::NodeManager::currentNM()->mkNode(::CVC4::kind::BITVECTOR_ITE, cond, l, r)); \
```
aniemetz: `return T(nm->mkNode(::CVC4::kind::BITVECTOR_ITE, cond, l, r));`

src/theory/fp/fp_converter.cpp
```diff
@@ -12,25 +12,1029 @@
...
+
+    d_additionalAssertions.push_back(tmp.valid(fpt(current.getType())).getNode());
+
+    return tmp;
+  }
+
+
+
+  // Non-convertible things should only be added to the stack at the very start, thus...
+  #define PASSTHROUGH   Assert(workStack.empty())
```
aniemetz: Macros should start with `CVC4_`, see https://github.com/CVC4/CVC4/wiki/Code-Style-Guidelines#naming-macros.

src/theory/fp/fp_converter.cpp
```diff
@@ -12,25 +12,1029 @@
...
+	      case roundTowardNegative : r.insert(current, traits::RTN()); break;
+	      case roundTowardZero : r.insert(current, traits::RTZ()); break;
+	      default :	Unreachable("Unknown rounding mode"); break;
+	      }
+
+	    } else {
+	      /******** Variables ********/
+	      //rm tmp(symfpu::NONDET);
+	      //rm tmp(buildRoundingModeUFApp(current));
+	      rm tmp(NodeManager::currentNM()->mkNode(kind::ROUNDINGMODE_BITBLAST, current));
```
aniemetz: Avoid calling `currentNM()` in each loop iteration.

src/theory/fp/fp_converter.cpp
```diff
@@ -12,25 +12,1029 @@
...
+	      }
+	      break;
+
+	    case kind::FLOATINGPOINT_MIN_TOTAL :
+	    case kind::FLOATINGPOINT_MAX_TOTAL :
+	      {
+		fpMap::const_iterator arg1(f.find(current[0]));
+		fpMap::const_iterator arg2(f.find(current[1]));
+		// Should not need to recurse down this argument
+		//ubvMap::const_iterator arg3(u.find(current[2]));
```
aniemetz: Remove?

src/theory/fp/fp_converter.cpp
```diff
@@ -12,25 +12,1029 @@
...
+    return result;
+  }
+
+
+  Node FpConverter::getValue (Valuation &val, TNode var) {
+    Assert(Theory::isLeafOf(var, THEORY_FP));
+
+    TypeNode t(var.getType());
+
+#ifdef REMOVE
```
aniemetz: Is this a leftover for debugging purposes? If it can be removed, remove. This applies to all blocks guarded with `#ifdef REMOVE`.

src/theory/fp/theory_fp.cpp
martin-cs: Reworked so it is a comment making clear why THEORY_BV is not set in the obvious place (i.e. here).

src/util/floatingpoint.cpp
martin-cs: Done.

src/util/floatingpoint.cpp
martin-cs: Done.

src/util/floatingpoint.cpp
martin-cs: It is and as isSigned is a template it should compile away correctly.  BUT... if it's OK with you, I'd like to keep it as, to me, it makes it clear that it's two cases.  Maybe I'm showing the first signs of coming down with a case of functional programming but ... it just seems a bit nicer?

src/util/floatingpoint.cpp
martin-cs: Gone!

src/util/floatingpoint.h
martin-cs: Headers re-ordered.  (I'm not sure the connection to this line though...)

src/theory/fp/theory_fp_rewriter.cpp
martin-cs: Fixed.

src/theory/fp/theory_fp.cpp
martin-cs: Looking at it again, the extra cases are not needed so they can be removed, along with the comment.

src/theory/fp/fp_converter.cpp
martin-cs: Done.

src/theory/fp/fp_converter.cpp
martin-cs: and so I have!

src/theory/fp/fp_converter.cpp
martin-cs: I would love to say this would teach me to not just copy and paste but to update as well but ... I don't think it will :-\

src/theory/fp/fp_converter.cpp
martin-cs: Done.

src/theory/fp/theory_fp_type_rules.h
```diff
@@ -613,6 +616,117 @@ class FloatingPointToRealTotalTypeRule {
 };
 
 
+class FloatingPointComponentBit {
+public :
+  inline static TypeNode computeType(NodeManager* nodeManager, TNode n, bool check)
```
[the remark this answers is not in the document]
martin-cs: Will apply globally.

src/theory/fp/theory_fp_rewriter.cpp
martin-cs: Will apply globally.

martin-cs:
Have fixed all of the specific issues and rebased on to https://github.com/CVC4/CVC4/pull/1905.  Next I will fix the general issues.

martin-cs:
A work-in-progress update, most but not all general issues are fixed.

src/theory/fp/kinds
```diff
@@ -283,4 +283,28 @@ operator FLOATINGPOINT_TO_REAL_TOTAL 2 "floating-point to real (defined for all
 typerule FLOATINGPOINT_TO_REAL_TOTAL   ::CVC4::theory::fp::FloatingPointToRealTotalTypeRule
 
 
+
+operator FLOATINGPOINT_COMPONENT_NAN 1 "NaN component of bit-blasted of a floating-point number"
```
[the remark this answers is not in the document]
martin-cs: Fixed.

src/theory/fp/kinds
```diff
@@ -283,4 +283,28 @@ operator FLOATINGPOINT_TO_REAL_TOTAL 2 "floating-point to real (defined for all
...
+typerule FLOATINGPOINT_COMPONENT_SIGN   ::CVC4::theory::fp::FloatingPointComponentBit
+
+operator FLOATINGPOINT_COMPONENT_EXPONENT 1 "Exponent component of bit-blasted of a floating-point number"
+typerule FLOATINGPOINT_COMPONENT_EXPONENT   ::CVC4::theory::fp::FloatingPointComponentExponent
+
+operator FLOATINGPOINT_COMPONENT_SIGNIFICAND 1 "Significand component of bit-blasted of a floating-point number"
+typerule FLOATINGPOINT_COMPONENT_SIGNIFICAND   ::CVC4::theory::fp::FloatingPointComponentSignificand
+
+
+operator ROUNDINGMODE_BITBLAST 1 "The bit-vector for a non-deterministic rounding mode"
```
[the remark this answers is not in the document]
martin-cs: Hmmm... I'd prefer not if that's OK.  Rounding Modes are their own kind and there are other kinds with ROUNDINGMODE in their name.

src/theory/fp/fp_converter.cpp
```diff
@@ -12,23 +12,1155 @@
  ** \brief Conversion of floating-point operations to bit-vectors using symfpu.
  **/
 
+// theory.h Only needed for the leaf test
```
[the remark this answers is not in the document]
martin-cs: Done.

src/theory/fp/fp_converter.cpp
```diff
@@ -12,23 +12,1155 @@
...
 #include "theory/fp/fp_converter.h"
+#include "theory/theory.h"
```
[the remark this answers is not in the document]
martin-cs: Not sure I follow.  I think this is the correct header ordering according to https://github.com/CVC4/CVC4/wiki/Code-Style-Guidelines#names-and-order-of-includes

src/theory/fp/fp_converter.cpp
```diff
@@ -12,23 +12,1155 @@
...
+#include "symfpu/core/ite.h"
+#include "symfpu/core/multiply.h"
+#include "symfpu/core/packing.h"
+#include "symfpu/core/remainder.h"
+#include "symfpu/core/sign.h"
+#include "symfpu/core/sqrt.h"
+#include "symfpu/utils/numberOfRoundingModes.h"
+#include "symfpu/utils/properties.h"
+
+namespace symfpu {
```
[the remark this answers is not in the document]
martin-cs: Yes but...

The if-then-else construct for symfpu is ... kinda more of a pain than I had intended.  Template specialisation in C++ seems like a great idea until you try to actually use it.  One of the problems is that the specialisation (which needs both symfpu headers and CVC4 headers) has to be done in the symfpu namespace...

src/theory/fp/fp_converter.cpp
```diff
@@ -12,23 +12,1155 @@
...
+using namespace ::CVC4::theory::fp::symfpuSymbolic;
+
+#define CVC4_SYM_ITE_DFN(T)                                                \
```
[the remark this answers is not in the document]
martin-cs: ... and then there is this problem.  It is a macro because we need four copies of exactly the same code.  It would be great to do it with a template but that causes problems with template specialisation.  If you want I can explain at greater length but, at the moment, this is the best of a bad set of options.

src/theory/fp/fp_converter.h
```diff
@@ -35,7 +51,656 @@ typedef PairHashFunction<TypeNode, TypeNode, TypeNodeHashFunction,
...
+  }
+
+  inline symbolicBitVector<isSigned> signExtendRightShift(
+      const symbolicBitVector<isSigned> &op) const
+  {
+    return symbolicBitVector<isSigned>(::CVC4::NodeManager::currentNM()->mkNode(
+        ::CVC4::kind::BITVECTOR_ASHR, *this, op));
+  }
+
+  /*** Modular opertaions ***/
```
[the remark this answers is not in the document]
martin-cs: Fixed.

martin-cs:
Another update.  I think this should address all of @4tXJ7f 's specific comments and the following general issues from @aniemetz and @mpreiner : 

1. Have run clang-format on all patches.
2. Have removed and / or actioned most of the todos.  There are a couple left in theory_fp_rewriter.cpp which are essentially questions about what to do if symfpu isn't used / other tidying up.  There are several in theory_fp.cpp from the final patch which implements lazy generation of float <-> real conversions.  I'm happy to open an issue for these as currently there are some Zeno-like issues with this.
3. I've tried to improve the documentation in the headers.
4. Implementations have been moved from .h to .cpp files.  The one exception is theory_fp_type_rules.h which has all methods implemented in the .h file (and no .cpp file), refactoring this seems like a separate task.
5. All header includes should now be in the correct order.
6. I've added CVC4_USE_SYMFPU around the parts of the code that use it.  Without this the solver will break if you try to create / solve problems using float-point numbers.  However this is no less broken than before so is a net improvement.

If you have more reviews / comments / requests please let me know.

martin-cs:
@florianschanda has run this version through his test suite and reports no additional errors.

florianschanda:
In addition, I have now tried to integrate this in our tools and it makes a large, positive impact for our test-suite and real-world users :) We can't wait...!

aniemetz:
@martin-cs can you open an issue for bullet 4 of your list (refactor theory_fp_type_rules.h) so that we don't forget about it and keep track?

martin-cs:
@aina https://github.com/CVC4/CVC4/issues/1913

src/util/floatingpoint.cpp
aniemetz: Well, it may be a question of personal preferences. I actually find it to be clearer to not have the else branch here, since it makes it more obvious that the else branch is actually the default case.
IMO it is not nicer but uglier, but as I said, that might be a matter of preferences.

src/theory/fp/fp_converter.cpp
aniemetz: :-)

src/theory/fp/kinds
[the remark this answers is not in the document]
aniemetz: Hm, this is a tough one. Rounding modes are for FP only, so it would actually be nice to have it named accordingly, but I see your point...

src/util/floatingpoint.cpp
martin-cs: I think we may have to agree to disagree on that one :-)

src/theory/fp/kinds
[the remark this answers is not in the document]
martin-cs: The kind in question is, literally, the bit-blasting of a rounding mode variable, which is why it seems, at least to me, better.

aniemetz:
@martin-cs  Also, can you open issues for the remaining todos and match them with the corresponding issue number (TODO blablabla #issue)?

martin-cs:
@aina : I've opened https://github.com/CVC4/CVC4/issues/1914 and added it to the TODOs.

src/theory/fp/theory_fp_rewriter.cpp
```diff
@@ -787,6 +788,118 @@ namespace constantFold {
...
+                           NodeManager::currentNM()->mkConst(res));
+  }
+
+  RewriteResponse componentExponent(TNode node, bool)
+  {
+    Assert(node.getKind() == kind::FLOATINGPOINT_COMPONENT_EXPONENT);
+
+    FloatingPoint arg0(node[0].getConst<FloatingPoint>());
+
+    // \todo Add a proper interface for this sort of thing to FloatingPoint
```
aniemetz: What about this one?

src/theory/fp/theory_fp_rewriter.cpp
```diff
@@ -787,6 +788,118 @@ namespace constantFold {
...
+  RewriteResponse roundingModeBitBlast(TNode node, bool)
+  {
+    Assert(node.getKind() == kind::ROUNDINGMODE_BITBLAST);
+
+    RoundingMode arg0(node[0].getConst<RoundingMode>());
+    BitVector value;
+
+#ifdef CVC4_USE_SYMFPU
+    /* \todo fix the numbering of rounding modes so this doesn't need
+     * to call symfpu at all and remove the dependency on fp_converter.h */
```
aniemetz: And this one?

src/theory/fp/theory_fp_rewriter.cpp
martin-cs: Ooops!  These ... it's hard to figure out what to do.  They are guarded by CVC4_USE_SYMFPU and the todo is ... it's not clear if they should be in with "avoid horrible death if floating-point things are used without CVC4_USE_SYMFPU" and I'm not sure how possible that is or whether it would just wind up duplicating effort.

I'll open an issue.

martin-cs: https://github.com/CVC4/CVC4/issues/1915 Done.

mpreiner:
Do not merge yet, I'll make the symfpu dependency optional again since we can now compile CVC4 without symfpu.

PR #1905, opened 2018-05-11
Add contrib/get-symfpu for downloading symfpu.

mpreiner:
For now we make symfpu a required dependency that gets downloaded during ./configure (if symfpu-CVC4 does not exist in the CVC4 base directory). Currently, it is not possible to compile CVC4 (after #1895) without symfpu.

This needs to be merged before merging #1895.


config/symfpu.m4
```diff
@@ -0,0 +1,46 @@
...
+    ),
+    SYMFPU_HOME="$withval",
+    [ if test -z "$SYMFPU_HOME" && ! test -e "$ac_abs_confdir/symfpu-CVC4/symfpu/core"; then
+        AC_MSG_FAILURE([must give --with-symfpu-dir=PATH, define environment variable SYMFPU_HOME, or use contrib/get-symfpu to setup symfpu for CVC4!])
+      fi
+    ]
+  )
+
+  # Check if symfpu was installed via contrib/get-symfpu
+  AC_MSG_CHECKING([whether symfpu was already installed via contrib/get-symfpu])
```
4tXJ7f: Nit: Comment and message are slightly wrong. This also checks whether SYMFPU_HOME or --with-symfpu-dir have been set (also the "already" can be removed).

config/symfpu.m4
```diff
@@ -0,0 +1,46 @@
...
+  if test -z "$SYMFPU_HOME" && test -e "$ac_abs_confdir/symfpu-CVC4/symfpu/core"; then
+    SYMFPU_HOME="$ac_abs_confdir/symfpu-CVC4"
+    AC_MSG_RESULT([yes, $SYMFPU_HOME])
+    have_symfpu_headers=1
+  else
+    AC_MSG_RESULT([no])
+  fi
+else
+  AC_MSG_RESULT([no, user didn't request symfpu])
+  with_symfpu=no
```
4tXJ7f: Nit: I don't think that this variable is used afterwards, so maybe it is not necessary to set it here.

mpreiner: It is used in configure.ac.

PR #1909, opened 2018-05-11
Fixing Mac compile errors

dddejan:


examples/hashsmt/sha1_inversion.cpp
```diff
@@ -54,19 +54,14 @@ int main(int argc, char* argv[]) {
     ofstream output(argv[2]);
     output << expr::ExprSetDepth(-1) << language::SetLanguage(language::output::LANG_SMTLIB_V2);
     output << SetBenchmarkLogicCommand("QF_BV") << endl;
-    output << SetBenchmarkStatusCommand(SMT_SATISFIABLE) << endl;
+
+    hashsmt::Word message(msg.size()*8, "x");
```
4tXJ7f: If I am reading this correctly, then `msg` is now unused and it is just always all `x`s of the same size as `argv[1]`. Is this intentional?

examples/hashsmt/sha1_inversion.cpp
```diff
@@ -54,19 +54,14 @@ int main(int argc, char* argv[]) {
...
 
     // Make the variables the size of the string
     hashsmt::cvc4_uchar8 *cvc4input = new hashsmt::cvc4_uchar8[msgSize];
```
4tXJ7f: Suggestion: This could also just be a vector, like that we don't need the `delete[]` at all.

.settings/language.settings.xml
```diff
@@ -1,9 +1,28 @@
 <?xml version="1.0" encoding="UTF-8" standalone="no"?>
 <project>
-	<configuration id="cdt.managedbuild.toolchain.gnu.base.1461790692" name="Default">
-		<extension point="org.eclipse.cdt.core.LanguageSettingsProvider">
-			<provider copy-of="extension" id="org.eclipse.cdt.ui.UserLanguageSettingsProvider"/>
-			<provider-reference id="org.eclipse.cdt.managedbuilder.core.MBSLanguageSettingsProvider" ref="shared-provider"/>
-		</extension>
-	</configuration>
+
```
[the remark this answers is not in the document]
4tXJ7f: Actually @dddejan, since you are an eclipse user, it would be great to have your input for how valuable it is to have those files under version control. [There seems to be some disagreement about that.](https://stackoverflow.com/questions/1551008/why-does-eclipse-create-a-settings-directory?utm_medium=organic&utm_source=google_rich_qa&utm_campaign=google_rich_qa)

dddejan:
I just wanted to fix the compile problem. You guys can figure it out, the pull request process is too much work for such small change.

ajreynol:
I am refactoring the code that led to the compile errors and will fix the issues there. This PR may not be necessary.

ajreynol:
Btw, sorry about the issues, we'll try not to break this sort of thing in the future.

4tXJ7f:
@dddejan sorry about that. I'll take this over if it is still necessary after @ajreynol's fixes (and I have WiFi again).

dddejan:
Btw, travis has mac builds, not a bad idea to add those.

https://docs.travis-ci.com/user/reference/osx/

4tXJ7f:
@dddejan Yeah, we are actually using them for our [homebrew tap](https://github.com/CVC4/homebrew-cvc4). The problem with the mac builds is that (in my experience) there is often quite a big backlog and the machines seem to be fairly slow. The current plan is to create scripts to do nightly builds of homebrew bottles, such that we test the mac builds more regularly and users don't have to compile CVC4 themselves. This is on my todo list.

PR #1918, opened 2018-05-14
Synthesize candidate-rewrites from standard inputs

ajreynol:
This PR adds a new preprocessing pass --synth-rr-prep which registers all subterms of the input into a candidate rewrite database per type.
This has the effect of printing messages of the form "(candidate-rewrite t1 t2)" where t1 and t2 are subterms of the input, are equivalent based on the criterion used by our sampler, and are not identified as equal by the current rewriter.  The extended rewriter can be tested using the additional flag --synth-rr-prep-ext-rew.


src/preprocessing/passes/synth_rew_rules.cpp
```diff
@@ -0,0 +1,147 @@
...
+  std::vector<Node>& assertions = assertionsToPreprocess->ref();
+
+  // compute the variables we will be sampling
+  std::vector<Node> vars;
+  unsigned nsamples = options::sygusSamples();
+
+  Options& nodeManagerOptions = NodeManager::currentNM()->getOptions();
+
+  // attribute to mark processed terms
+  SynthRrComputedAttribute srrca;
```
4tXJ7f: Suggestion: could use a map for consistency with `visited` and since the attribute is not used anywhere else.

src/preprocessing/passes/synth_rew_rules.cpp
```diff
@@ -0,0 +1,147 @@
...
+struct SynthRrComputedAttributeId {};
+typedef expr::Attribute<SynthRrComputedAttributeId, bool> SynthRrComputedAttribute;
+  
+SynthRewRulesPass::SynthRewRulesPass(PreprocessingPassContext* preprocContext)
+    : PreprocessingPass(preprocContext, "synth-rr"){};
+
+PreprocessingPassResult SynthRewRulesPass::applyInternal(
+    AssertionPipeline* assertionsToPreprocess)
+{
+  Trace("synth-rr-pass") << "Synthesize rewrite rules from assertions..."
```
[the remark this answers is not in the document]
ajreynol: Good to know, I'll keep this one though.

src/preprocessing/passes/synth_rew_rules.cpp
```diff
@@ -0,0 +1,147 @@
...
+
+using namespace std;
+
+namespace CVC4 {
+namespace preprocessing {
+namespace passes {
+
+//attribute for whether we have computed rewrite rules for a given term
+struct SynthRrComputedAttributeId {};
+typedef expr::Attribute<SynthRrComputedAttributeId, bool> SynthRrComputedAttribute;
```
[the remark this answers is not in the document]
ajreynol: Are we making this a strict guideline? Note all the expr::Attributte declarations currently use "typedef".

src/preprocessing/passes/synth_rew_rules.cpp
ajreynol: Actually, it's important that this is a global attribute.  The issue is that in the subcall to the separate SmtEngine, we need to remember that this attribute has been set, which we wouldn't get from a local map. This is to avoid a infinite loop, see issue #1924.

4tXJ7f: Hmm, can you add a comment for that? This is pretty subtle. Also, I should probably submit my patch for having different options for the subsolver at some point...

src/preprocessing/passes/synth_rew_rules.cpp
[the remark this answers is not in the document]
4tXJ7f: It's not mandatory, no. It is easier to read (IMHO) and more powerful though.

src/preprocessing/passes/synth_rew_rules.cpp
ajreynol: Yeah that would be great.  I'll comment.

PR #1919, opened 2018-05-14
Fix file extension

cdonovick:


cdonovick:
@ajreynol I am assuming this should be added to Makefile.tests

ajreynol:
@cdonovick the benchmark should not be enabled. The benchmark is actually an example of a benchmark we can't handle in the sygus solver. I haven't looked into have a good way of incorporating this into the regressions.

cdonovick:
@4tXJ7f How do I "add to the disabled tests"

PR #1930, opened 2018-05-16
Option to force return values of Bool functions to be constant in CegisUnif

HanielB:
Also removes wrong distinctness criterion of enumerators and guards addition of ITE constructor to Boolean grammars with the sygusUnif option.


src/theory/quantifiers/sygus/sygus_unif_strat.cpp
```diff
@@ -793,11 +839,20 @@ void SygusUnifStrategy::staticLearnRedundantOps(
   // do not use recursive Boolean connectives for conditions of ITEs
   if (nrole == role_ite_condition)
   {
+    TypeNode sygus_tn = TypeNode::fromType(dt.getSygusType());
     for (unsigned j = 0, size = dt.getNumConstructors(); j < size; j++)
     {
       Node op = Node::fromExpr(dt[j].getSygusOp());
       Trace("sygus-strat-slearn")
           << "...for ite condition, look at operator : " << op << std::endl;
+      if (op.isConst() && dt[j].getNumArgs() == 0)
```
ajreynol: You need to check whether the sygus types match. In particular the branches of the ITEs should match the overall type.  The code below also needs fixing. We also need to ensure that we don't overconstrain the condition.

PR #1951, opened 2018-05-21
Make sygus infer find function definitions

ajreynol:
This makes us able to solve standard inputs with sygus, plug in the solutions as define-fun and answer SAT.


src/theory/quantifiers/sygus_inference.cpp
```diff
@@ -204,17 +219,70 @@ bool SygusInference::simplify(std::vector<Node>& assertions)
...
+                           << std::endl;
+      final_ff.push_back(ff);
+      final_ff_sol.push_back(it->second);
+      master_smte->defineFunction(ff.toExpr(), args, it->second[1]);
     }
     else
     {
-      assertions[i] = truen;
+      // all synthesis solutions should correspond to a variable we introduced
+      Assert(false);
```
HanielB: why not just assert that ``itffv != ff_var_to_ff.end()`` before the ``if``?

src/theory/quantifiers/sygus_inference.cpp
```diff
@@ -204,17 +219,70 @@ bool SygusInference::simplify(std::vector<Node>& assertions)
...
+    }
+  }
+
+  // apply substitution to everything, should result in SAT
+  Node truen = nm->mkConst(true);
```
HanielB: this is not used

Issue #1954, opened 2018-05-22
--check-proofs fails but --check-unsat-cores --check-proofs succeeds


4tXJ7f:
For one of our regression tests, `--check-proofs` alone fails while the combination `--check-unsat-cores --check-proofs` succeeds (see output below). The issue might be that there is a condition somewhere that only checks for unsat cores being enabled instead of also checking if proof generation is enabled.

```
$ builds/bin/cvc4 --check-proofs test/regress/regress0/bv/core/slice-15.smt
/tmp/cvc4_USj3xA, line 1923, column 163:
The type expected for a symbol does not match the symbol's type.
1. The symbol: bb.l66
2. The expected type: (th_holds (not (and (iff (bitof BVSKOLEM$$_1 0) (bitof BVSKOLEM$$_1 1)) (and (iff (bitof BVSKOLEM$$_1 1) (bitof BVSKOLEM$$_1 2)) (and (iff (bitof BVSKOLEM$$_1 2) (bitof BVSKOLEM$$_1 3)) (and (iff (bitof BVSKOLEM$$_1 3) (bitof BVSKOLEM$$_1 4)) (and (iff (bitof BVSKOLEM$$_1 4) (bitof BVSKOLEM$$_1 5)) (and (iff (bitof BVSKOLEM$$_1 5) (bitof BVSKOLEM$$_1 6)) (and (iff (bitof BVSKOLEM$$_1 6) (bitof BVSKOLEM$$_1 7)) (and (iff (bitof BVSKOLEM$$_1 7) (bitof BVSKOLEM$$_1 8)) (and (iff (bitof BVSKOLEM$$_1 8) (bitof BVSKOLEM$$_1 9)) (and (iff (bitof BVSKOLEM$$_1 9) (bitof BVSKOLEM$$_1 10)) (and (iff (bitof BVSKOLEM$$_1 10) (bitof BVSKOLEM$$_1 11)) (and (iff (bitof BVSKOLEM$$_1 11) (bitof BVSKOLEM$$_1 12)) (and (iff (bitof BVSKOLEM$$_1 12) (bitof BVSKOLEM$$_1 13)) (and (iff (bitof BVSKOLEM$$_1 13) (bitof BVSKOLEM$$_1 14)) (iff (bitof BVSKOLEM$$_1 14) true)))))))))))))))))
[...]

Issue #1953, opened 2018-05-22
CVC4 produces/complains about satisfiable unsat core


4tXJ7f:
```
$ builds/bin/cvc4 --check-unsat-cores test/regress/regress0/uf/cnf_abc.smt2
(error "Internal error detected
void CVC4::SmtEngine::checkUnsatCore()
../../../../src/smt/smt_engine.cpp:5190:
SmtEngine::checkUnsatCore(): produced core was satisfiable."
```

4tXJ7f:
Note: `--check-proofs` succeeds (and the combination `--check-proofs --check-unsat-cores` also succeeds, which may indicate that CVC4 is not checking unsat cores when both options are enabled).

4tXJ7f:
My speculation above is not true. It looks like we are generating a different unsat core with proofs enabled (with 18 assertions instead of 14).

4tXJ7f:
Fixed by #1958.

PR #1955, opened 2018-05-22
Check unsat cores in regressions also without LFSC

4tXJ7f:
When moving the LFSC checker out of the CVC4 repository in commit
dfa69ff98a14fcc0f2387e59a0c9994ef440e7d0, we disabled checking unsat
cores when CVC4 was compiled without LFSC due to the complexity of the
regression script. This commit changes the new(-ish) Python regression
script to check unsat cores even when CVC4 was compiled without LFSC.
This is done by having two separate flags, --with-lfsc and
--enable-proof, for the regression script that mirror the configuration
flags. The regression script performs unsat cores and proofs checks
based on those flags.

Additionally, this commit changes the regression script to check proofs
and unsat cores in two independent runs. Testing them in a single run
masked issues #1953 and #1954.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1052,9 +1052,11 @@ AC_ARG_WITH([cxxtest-dir],
 
 TESTS_ENVIRONMENT=
 RUN_REGRESSION_ARGS=
-# FIXME currently, unsat core regressions are disabled without LFSC
 if test "$with_lfsc" = yes; then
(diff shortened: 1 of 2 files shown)


4tXJ7f:
Note that regressions will fail due to the open issues mentioned above.

aniemetz:
So this PR should only be merged in when the open issues are resolved?

4tXJ7f:
Yeah, I already submitted two PRs to fix the issues #1957 (awaiting review) and #1958 (merged).

4tXJ7f:
One more fix in #1976. Now it should hopefully work out :)

PR #1958, opened 2018-05-22
Disable UF symmetry breaker for unsat cores

4tXJ7f:
Currently, --check-unsat-cores fails for one of our regressions. It
turns out that the issue is due to preprocessing done by the symmetry
breaker.  The symmetry breaker does not add dependencies and thus the
generated unsat core is incomplete/wrong. This commit disables the
symmetry breaker when we are generating unsat cores (it was previously
only disabled when generating proofs). Fixes issue #1953.


ajreynol:
Perhaps clarify in the title of this PR that this is the old implementation of the UF symmetry breaker (not Paul's one).  Regardless I can review this.

PR #1968, opened 2018-05-23
Fix some GCC 6.0 warnings in the floating-point theory.

martin-cs:


4tXJ7f:
Those should already be fixed in #1959 :)

martin-cs:
I am a bad person because I don't read my CVC4 change e-mails as well as I should :-\

PR #1976, opened 2018-05-24
Disable unconstrainedSimp pass when proofs enabled

4tXJ7f:
Currently, we may end up enabling the unconstrainedSimp pass when proofs
are enabled because the check that this pass is disabled happens before
a check for automatically enabling it. This lead to issues when running
for example on regress0/aufbv/bug00.smt with --check-proofs. This commit
moves the code for automatically enabling the pass to only be run if
proofs and unsat cores are disabled. Note: This commit is mostly a
simple code move but formatting in setDefaults was a bit off, so there
are a lot of formatting changes.


4tXJ7f:
Hint: "Diff settings" > "Hide whitespace changes" should be helpful for reviewing this PR :)

PR #1983, opened 2018-05-25
Reenable repair const

ajreynol:
This reenables the previous scheme for sygus modules that don't support repair, and enables the option by default.

### DIFF
--- a/src/theory/quantifiers/sygus/sygus_module.h
+++ b/src/theory/quantifiers/sygus/sygus_module.h
@@ -124,9 +124,8 @@ class SygusModule
   }
   /**
    * Are we trying to repair constants in candidate solutions?
-   * If a module returns false for this and the option sygusRepairConst
-   * is true, we use a default scheme for trying to repair constants in
-   * candidate solutions in CegConjecture.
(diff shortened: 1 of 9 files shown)


src/theory/quantifiers/sygus/sygus_module.h
```diff
@@ -122,6 +122,13 @@ class SygusModule
   {
     return Node::null();
   }
+  /**
+   * Are we trying to repair constants in candidate solutions?
+   * If a module returns false for this and the option sygusRepairConst
+   * is true, we use a default scheme for trying to repair constants in
+   * candidate solutions in CegConjecture.
+   */
+  virtual bool usingRepairConst() { return false; }
```
HanielB: Instead of ``usingRepairConst`` I think the test should be ``usingSymbolicConst``. The actual question being answered here is "are we handling candidate solutions with symbolic constants?"

ajreynol: Actually, in terms of this PR, the important question from CegConjecture is "did you try to repair the candidate solution you gave me?"

PR #2004, opened 2018-05-29
 Make user's SMT2 version override file version

### LINKED ISSUE #1969: Smt-lib-version is overwritten by smt2 commands even when user provides input language
If one sets the smt2-lib language with a variant, e.g. "--lang=smt2.6.1", then this is overwritten by the smt2 command:

```(set-info :smt-lib-version 2.6)```

This issue is that the Parser does not have access to options, so it cannot tell whether or not the input language was set by user, hence we always overwrite.

### PR DESCRIPTION
**Pull request by 4tXJ7f**
Fixes #1969. With this commit, the language specified by the user takes
priority over the version specified in the file itself.

Note: This PR depends on #2003.

### CODE REVIEW

File: test/regress/regress0/lang_opts.smt2 [hunk 1]
```diff
@@ -0,0 +1,7 @@
+; Check that the language set in the command line options has higher priority
+; than the language specified in the input file.
+;
+; COMMAND-LINE: --lang=smt2.5
+; EXPECT: "LANG_SMTLIB_V2_5"
+(set-info :smt-lib-version 2.6)
```
ajreynol: Can you add another test here for --lang=smt2.6.1?

PR #2023, opened 2018-05-30
Enable cegqi (with model values) for floating point by default

ajreynol:


4tXJ7f:
Hmm, unfortunately, it looks like this causes problems (this is on BVFP/20170501-Heizmann-UltimateAutomizer/bary_diverge_true-unreach-call.c_1264.smt2 but it is not the only benchmark where this happens):

```
(error "Unreachable code reached
CVC4::theory::RewriteResponse CVC4::theory::fp::rewrite::notFP(CVC4::TNode, bool)
../../../../src/theory/fp/theory_fp_rewriter.cpp:58
non floating-point kind (272) in floating point rewrite?")
```

PR #2051, opened 2018-06-04
Move assertion in arrays

### LINKED ISSUE #2045: Assertion failure in preRegisterTermInternal
Flags: `--decision=internal --full-saturate-quant`
AUFBVDTLIA/20172804-Barrett/rec-fun-ijcar2016/wo-wdf/mutant_wo_fd_vcs_wdf/2.5-cvc4-AbsFun.scala-4.smt2-2.smt2

Error:
```
(error "Assertion failure.
void CVC4::theory::arrays::TheoryArrays::preRegisterTermInternal(CVC4::TNode)
../../../../src/theory/arrays/theory_arrays.cpp:653:

  d_isPreRegistered.find(node) != d_isPreRegistered.end()
")
```

ajreynol:
It appears as though array's preRegisterInternal( t ) can make a recursive call to preRegisterInternal( t ) for the same term t.  On the recursive call, the method terminates since t has been added to the equality engine. However, we get an assertion failure saying that we haven't preregistered t. However, we are in the process of pre-registering it. To avoid the assertion failure, this PR adds t to the preRegister cache (which is only used in debug) *before* recursive calls to preRegisterInternal.

This fixes #2045.

### DIFF
--- a/src/theory/arrays/theory_arrays.cpp
+++ b/src/theory/arrays/theory_arrays.cpp
@@ -668,6 +668,7 @@ void TheoryArrays::preRegisterTermInternal(TNode node)
     {
       d_equalityEngine.addTerm(node);
     }
+    Assert((d_isPreRegistered.insert(node), true));
 
     if (options::arraysLazyRIntro1() && !options::arraysWeakEquivalence()) {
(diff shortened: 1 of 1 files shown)


ajreynol:
This is @barrettcw's code.

Issue #2036, opened 2018-06-01
Do not use built-in transcendentals by default


4tXJ7f:

4tXJ7f:
Fixed in #2052.

PR #2052, opened 2018-06-05
Only enable transcendentals if logic is N[I]RAT

4tXJ7f:
With this commit, we only enable transcendentals if the logic is set to
NRAT or NIRAT (or ALL). It also changes the SMT2 parser to emit an error
message if the user tries to shadow symbols defined in the theory.

### DIFF
--- a/src/parser/smt2/smt2.cpp
+++ b/src/parser/smt2/smt2.cpp
@@ -51,11 +51,13 @@ void Smt2::addArithmeticOperators() {
   Parser::addOperator(kind::LEQ);
   Parser::addOperator(kind::GT);
   Parser::addOperator(kind::GEQ);
+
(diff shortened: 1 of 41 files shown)


4tXJ7f:
This PR addresses #2036.

src/parser/smt2/smt2.cpp
```diff
@@ -51,7 +51,10 @@ void Smt2::addArithmeticOperators() {
   Parser::addOperator(kind::LEQ);
   Parser::addOperator(kind::GT);
   Parser::addOperator(kind::GEQ);
-  
+}
```
ajreynol: I think POW is actually a non-transcendental function (it was there before we supported transcendental functions).  It's not standard.  However, it probably makes sense to have this as a non-transcendental function.

PR #2059, opened 2018-06-08
Reset decisions at SAT level after solving

### LINKED ISSUE #1990: Failing model check
For the following example, CVC4's --check-model fails:
```
(set-logic SAT)
(declare-fun v1 () Bool)
(declare-fun v2 () Bool)
(declare-fun v3 () Bool)
(declare-fun v4 () Bool)
(declare-fun v5 () Bool)
(check-sat)
(assert (or v1 true))
(assert (or v2 v3))
(push)
(check-sat)
(assert false)
(push)
(check-sat)
(pop)
(push)
(check-sat)
```

Output:
```
sat
sat
unsat
(error "Internal error detected
void CVC4::SmtEngine::checkModel(bool)
../../../../src/smt/smt_engine.cpp:5455:
SmtEngine::checkModel(): ERRORS SATISFYING ASSERTIONS WITH MODEL:
assertion:     false
simplifies to: false
expected `true'.
Run with `--check-models -v' for additional diagnostics.")
```

4tXJ7f:
Some quick background: CVC4 has two primary contexts: the assertion
context (which corresponds to the push/pops issued by the user) and the
decision/SAT context (which corresponds to the push/pops done by the
user and each decision made by MiniSat.

Before 2cc0e2c6a691fb6d30ed8879832b49bc1f277d77, we had an additonal
push/pop when doing the actual solving. With these removed, it could
happen that we get a wrong result when doing incremental solving:

```
...
; check-sat call returns sat, the decision level in the SAT solver is
; non-zero at this point
(check-sat)
; Solver::addClause_() determines that we cannot add the clause to the
; SAT solver directly because the decision level is non-zero (i.e. the
; clause is treated like a theory lemma), thus it is added to the lemmas
; list.
(assert false)
; The solver stores some information about the current state, including
; the value of the "ok" flag in Solver, which indicates whether the
; current constraints are unsatisfiable. Note that "ok" is true at this
; point.
(push)
; Now, in Solver::updateLemmas(), we add clauses from the lemmas list.
; The problem is that empty clauses (which "false" is after removing "false"
; literals) and unit clauses are not added like normal clauses. Instead,
; the empty clause, essentially just results in the "ok" flag being set to
; false (unit clauses are added to the decision trail).
(check-sat)
; Here, the solver restores the information stored during
; (push), including the "ok" flag, which is now true again.
(pop)
; At this point, the solver has "forgotten" about (assert false) since
; the "ok" flag is back to true and it answers sat.
(check-sat)
```

There are multiple ways to look at the problem and to fix it:

- One could argue that an input assertion should always be added
directly to the clauses in the SAT solver, i.e. the solver should
always be at decision level 0 when we are adding input clauses. The
advantage of this is that it is relatively easy to implement,
corresponds to what the original MiniSat code does (it calls
Solver::cancelUntil(0) after solving), and is no worse than what we had
before commit 2cc0e2c6a691fb6d30ed8879832b49bc1f277d77 (push/pop do a
strict superset of what resetting the decision at the current assertion
level does). The disadvantage is that we might throw away some useful work.

- One could argue that is fine that (assert false) is treated like a
theory lemma. The advantage being that it might result in more efficient
solving and the disadvantage being that it is much trickier to
implement.

Unfortunately, it is not quite clear from the code what the original
intention was but we decided to implement the first solution. This
commit exposes a method in MiniSat to reset the decisions at the current
assertion level. We call this method from SmtEngine after solving.
Resetting the decisions directly after solving while we are still in
MiniSat does not work because it causes issues with datastructures in
the SMT solver that use the decision context (e.g.
TheoryEngine::d_incomplete seems to be reset too early, resulting in us
answering sat instead of unknown).

Fixes #1990.

### DIFF
--- a/src/prop/minisat/core/Solver.cc
+++ b/src/prop/minisat/core/Solver.cc
@@ -344,7 +344,7 @@ bool Solver::addClause_(vec<Lit>& ps, bool removable, ClauseId& id)
     // Fit to size
     ps.shrink(i - j);
 
-    // If we are in solve or decision level > 0
+    // If we are in solve_ or propagate
     if (minisat_busy)
     {
       Debug("pf::sat") << "Add clause adding a new lemma: ";
--- a/src/prop/minisat/core/Solver.cc
+++ b/src/prop/minisat/core/Solver.cc
@@ -345,7 +345,8 @@ bool Solver::addClause_(vec<Lit>& ps, bool removable, ClauseId& id)
     ps.shrink(i - j);
 
     // If we are in solve or decision level > 0
-    if (minisat_busy || decisionLevel() > 0) {
+    if (minisat_busy)
+    {
       Debug("pf::sat") << "Add clause adding a new lemma: ";
       for (int k = 0; k < ps.size(); ++k) {
         Debug("pf::sat") << ps[k] << " ";
@@ -369,6 +370,8 @@ bool Solver::addClause_(vec<Lit>& ps, bool removable, ClauseId& id)
       // Debug("cores") << "lemma push " << proof_id << " " << (proof_id & 0xffffffff) << std::endl;
       // lemmas_proof_id.push(proof_id);
     } else {
+      assert(decisionLevel() == 0);
+
       // If all false, we're in conflict
       if (ps.size() == falseLiteralsCount) {
         if(PROOF_ON()) {
@@ -534,9 +537,7 @@ void Solver::cancelUntil(int level) {
     }
 }
 
-void Solver::popTrail() {
-  cancelUntil(0);
-}
+void Solver::resetTrail() { cancelUntil(0); }
 
 //=================================================================================================
 // Major methods:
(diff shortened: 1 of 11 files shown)


4tXJ7f:
Note: regress2 passes but I am doing additional tests, thus the "do not merge" label.

4tXJ7f:
With this patch, we actually get slightly more solutions (2493086 vs 2484241) when running on all incremental benchmarks but there are definitely improvements that can be done to our incremental solving. It is also worth noting that we were running with the additional push/pop last year, so this should still be a net improvement albeit a small one. I didn't see any errors when running on the incremental benchmarks with this patch.

src/prop/minisat/core/Solver.cc
```diff
@@ -345,7 +345,8 @@ bool Solver::addClause_(vec<Lit>& ps, bool removable, ClauseId& id)
     ps.shrink(i - j);
 
     // If we are in solve or decision level > 0
-    if (minisat_busy || decisionLevel() > 0) {
+    if (minisat_busy)
```
ajreynol: update the comment above?

PR #2061, opened 2018-06-08
Disable BV-abstraction in the competition script.

mpreiner:
Since we have several issues with the BV-abstraction code we'll disable it in the 2018 competition run script. Net gain if enabled is +3-1. We need to revisit the BV-abstraction code and thoroughly test it.


ajreynol:
Do we need to disable in QF_NIA as well?

ajreynol:
Does the +3-1 include QF_NIA?

mpreiner:
It's +3-1 on QF_BV. BV-abstraction is a pass for the mcm family in QF_BV. I don't think it will make a big difference in QF_NIA.

PR #2064, opened 2018-06-11
Fix equality conflicts reported by FP

ajreynol:
This fixes a soundness issue in FP where it would report unsound unit conflicts when a conflict was found by propagation from the equality engine of FP.

This changes FP to mimic the behavior of other theories, which set a flag to indicate that there is a conflict but do not report the propagating (dis)equality as a standalone conflict. Here are examples of what other theories do in this case:
https://github.com/CVC4/CVC4/blob/master/src/theory/datatypes/theory_datatypes.cpp#L677
https://github.com/CVC4/CVC4/blob/master/src/theory/sets/theory_sets_private.cpp#L2099
https://github.com/CVC4/CVC4/blob/master/src/theory/uf/theory_uf.cpp#L286

### DIFF
--- a/src/theory/fp/theory_fp.cpp
+++ b/src/theory/fp/theory_fp.cpp
@@ -907,8 +907,9 @@ bool TheoryFp::handlePropagation(TNode node) {
 
   bool stat = d_out->propagate(node);
 
-  if (!stat) handleConflict(node);
-
+  if (!stat) {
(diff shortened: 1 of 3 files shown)


martin-cs:
Thanks @ajreynol .

ajreynol:
No problem @martin-cs, happy we found this :)

PR #2066, opened 2018-06-11
Fix simple regexp consume

### LINKED ISSUE #2060: Unreachable code in TheoryStringsRewriter::testConstStringInRegExp()
Flags: `-m --strings-exp`

Error:

```
(error "Unreachable code reached.; static bool CVC4::theory::strings::TheoryStringsRewriter::testConstStringInRegExp(CVC4::String&, nsigned int, CVC4::TNode)
../../../../src/theory/strings/theory_strings_rewriter.cpp:1019
")
```

Instance:
```
(set-logic QF_S)

(declare-const action String)
(declare-const example_key String)

(assert (str.in.re action (re.++
                           (str.to.re "foobar:ab")
                           (re.* re.allchar)
                   )))

(declare-const action_1 String)
(declare-const action_2 String)

(assert (and
         (= action (str.++ action_1 example_key action_2))
         (= action_1 "foobar:a")
         ))

(check-sat)
(get-model)
```

ajreynol:
This fixes #2060.

### DIFF
--- a/src/theory/strings/theory_strings_rewriter.h
+++ b/src/theory/strings/theory_strings_rewriter.h
@@ -29,6 +29,33 @@ namespace strings {
 
 class TheoryStringsRewriter {
 private:
+  /** simple regular expression consume 
+   * 
+   * This method is called when we are rewriting a membership of the form
+   *   s1 ++ ... ++ sn in r1 ++ ... ++ rm
+   * We have that mchildren consists of the strings s1...sn, and children
+   * consists of the regular expressions r1...rm. 
+   * 
+   * This method tries to strip off parts of the concatenation terms. It updates
+   * the vectors such that the resulting vectors are such that the membersip
+   * mchildren[n'...n''] in children[m'...m''] is equivalent to the input
+   * membership. The argument dir indicates the direction to consider, where
+   * 0 means strip off the front, 1 off the back, and < 0 off of both.
+   * 
+   * If this method returns the false node, then we have inferred that the input
+   * membership is equivalent to false.
+   * 
+   * For example, given input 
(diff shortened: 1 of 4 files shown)


src/theory/strings/theory_strings_rewriter.h
```diff
@@ -28,7 +28,34 @@ namespace theory {
...
+   * membership. The argument dir indicates the direction to consider, where
+   * 0 means strip off the front, 1 off the back, and < 0 off of both.
+   *
+   * If this method returns the false node, then we have inferred that the input
+   * membership is equivalent to false.
+   *
+   * For example, given input
+   *   mchildren = { "ab", x }, children = { "a", ("cd")* } and dir = 0,
+   * this method updates:
+   *   mchildren = { "b", x }, children = { ("cd")* }
```
4tXJ7f: What does the method return in this case?

src/theory/strings/theory_strings_rewriter.h
```diff
@@ -28,7 +28,34 @@ namespace theory {
...
+   * the vectors such that the resulting vectors are such that the membersip
+   * mchildren[n'...n''] in children[m'...m''] is equivalent to the input
+   * membership. The argument dir indicates the direction to consider, where
+   * 0 means strip off the front, 1 off the back, and < 0 off of both.
+   *
+   * If this method returns the false node, then we have inferred that the input
+   * membership is equivalent to false.
+   *
+   * For example, given input
+   *   mchildren = { "ab", x }, children = { "a", ("cd")* } and dir = 0,
```
4tXJ7f: Shouldn't it be [["a"]] and [["cd"]] on the right-hand side?

ajreynol: Yes, I've fixed.

PR #2090, opened 2018-06-26
sygusComp2018: Add evaluator

4tXJ7f:
This commit adds the Evaluator class that can be used to quickly
evaluate terms under a given substitution without going through the
rewriter. This has been shown to lead to significant performance
improvements on SyGuS PBE problems with a large number of inputs because
candidate solutions are evaluated on those inputs. With this commit, the
evaluator only gets called from the SyGuS solver but there are
potentially other places in the code that could profit from it.


ajreynol:
This is PR 1/18 from the sygus comp 2018 branch (https://github.com/ajreynol/CVC4/tree/sygusComp2018-2).

src/theory/evaluator.h
```diff
@@ -0,0 +1,103 @@
+/*********************                                                        */
+/*! \file evaluator.h
+ ** \verbatim
+ ** Top contributors (to current version):
+ **   Andres Noetzli
+ ** This file is part of the CVC4 project.
+ ** Copyright (c) 2009-2017 by the authors listed in the file AUTHORS
```
ajreynol: 2018

src/theory/evaluator.h
```diff
@@ -0,0 +1,103 @@
...
+namespace CVC4 {
+namespace theory {
+
+/**
+ * Struct that holds the result of an evaluation. The actual value is stored in
+ * a union to avoid the overhead of a class hierarchy with virtual methods.
+ */
+struct EvalResult
+{
+  // Describes which type of result is being stored
```
ajreynol: I tend to always use /** */ for comments above members in a class. Not sure if there is a coding policy.

src/theory/evaluator.h
```diff
@@ -0,0 +1,103 @@
...
+  enum
+  {
+    BOOL,
+    BITVECTOR,
+    RATIONAL,
+    STRING,
+    INVALID
+  } d_tag;
+
+  // Stores the actual result
```
ajreynol: Ditto.

src/theory/evaluator.h
```diff
@@ -0,0 +1,103 @@
...
+  ~EvalResult();
+
+  /**
+   * Converts the result to a Node. If the result is not valid, this function
+   * returns the null node.
+   */
+  Node toNode() const;
+};
+
+class Evaluator
```
ajreynol: Maybe a quick sentence or two about the purpose of this class in a comment here.

src/theory/evaluator.cpp
```diff
@@ -0,0 +1,591 @@
+/*********************                                                        */
+/*! \file evaluator.cpp
+ ** \verbatim
+ ** Top contributors (to current version):
+ **   Andres Noetzli
+ ** This file is part of the CVC4 project.
+ ** Copyright (c) 2009-2017 by the authors listed in the file AUTHORS
```
ajreynol: 2018

4tXJ7f:
I have addressed the comments and added a couple of additional comments in the source file.

PR #2091, opened 2018-06-26
 Disable uf symmetry breaker in incremental mode

### LINKED ISSUE #2065: UF symmetry breaker should be disabled in incremental mode

ajreynol:
Fixes #2065.


ajreynol:
Its a performance problem.  I dont think there is a case where we'd emit an error, since if the user wants to try setting UF symmetry breaker to true in incremental, they are allowed to (we guard with wasSetByUser()). But this PR makes it so that its off by default.

PR #2096, opened 2018-06-26
Fix assertion for relational triggers

### LINKED ISSUE #2030: Assertion failure in InstMatchGenerator
We get the following assertion failure on AUFNIRA/FFT/z3.667166.smt2 when running with `--trigger-sel=max --full-saturate-quant` (currently minimizing the benchmark):

```
(error "Assertion failure.
virtual int CVC4::theory::inst::InstMatchGenerator::getNextMatch(CVC4::Node, CVC4::theory::inst::InstMatch&, CVC4::theory::QuantifiersEngin
e*, CVC4::theory::inst::Trigger*)
../../../../src/theory/quantifiers/ematching/inst_match_generator.cpp:424:

  d_eq_class.isNull()
")
```

ajreynol:
Fixes #2030.


4tXJ7f:
What about the other assertion metioned in the issue? Is that something different?

ajreynol:
Yes, I'm not sure why those are there.  And I believe those are fixed in #2099 (will comment on #2099 if not).

PR #2104, opened 2018-06-26
 sygusComp2018: optimization for invariance test

ajreynol:
This is PR 6/18 from the sygus comp 2018 branch (https://github.com/ajreynol/CVC4/tree/sygusComp2018-2).

This does a simple miniscoping step in invariance test, based on the following reasoning:

Checking whether A1^...^An evaluates to true is the same as checking for each i=1...n that Ai evaluates to true. The latter has the advantage that we can stop as soon as a single disjunct does not evaluate to true.


src/theory/quantifiers/sygus/sygus_invariance.h
```diff
@@ -97,30 +97,42 @@ class SygusInvarianceTest
...
-  /** the formula we are evaluating */
-  Node d_conj;
+  /** the formulas we are evaluating */
+  std::vector<Node> d_terms;
   /** the variable */
   TNode d_var;
   /** the result of the evaluation */
   Node d_result;
+  /** the parent kind we are checking, undefined if size(d_terms) is 1. */
+  Kind d_kind;
```
4tXJ7f: It looks like this member variable is actually not needed. It is not read after `init()` as far as I can tell.

ajreynol: I'd prefer having it around, since it makes the data of the class coherent, i.e. notice that it is used in the comment describing the invariant of the class on line 105.

4tXJ7f: Fair enough.

PR #2120, opened 2018-06-28
Refactor ApplySubsts preprocessing pass.

aniemetz:


aniemetz:
Addressed all comments.

PR #2130, opened 2018-07-02
New C++ API: Implementation of Kind maps.

aniemetz:


src/api/cvc4cpp.cpp
```diff
@@ -90,5 +91,571 @@ std::ostream& operator<<(std::ostream& out, const Result& r)
...
+    {REGEXP_LOOP, CVC4::Kind::REGEXP_LOOP},
+    {REGEXP_EMPTY, CVC4::Kind::REGEXP_EMPTY},
+    {REGEXP_SIGMA, CVC4::Kind::REGEXP_SIGMA},
+    /* Quantifiers --------------------------------------------------------- */
+    {FORALL, CVC4::Kind::FORALL},
+    {EXISTS, CVC4::Kind::EXISTS},
+    {LAST_KIND, CVC4::Kind::LAST_KIND},
+};
+
+std::unordered_map<CVC4::Kind, Kind, CVC4::kind::KindHashFunction>
```
mpreiner: Make static.

src/api/cvc4cpp.cpp
```diff
@@ -90,5 +91,571 @@ std::ostream& operator<<(std::ostream& out, const Result& r)
   return out;
 }
 
+/* -------------------------------------------------------------------------- */
+/* Kind                                                                       */
+/* -------------------------------------------------------------------------- */
+
+static std::unordered_map<Kind, CVC4::Kind, KindHashFunction> s_kinds{
```
mpreiner: Add a comment about mapping external to internal kinds?

src/api/cvc4cpp.cpp
```diff
@@ -90,5 +91,571 @@ std::ostream& operator<<(std::ostream& out, const Result& r)
...
+        {CVC4::Kind::SORT_TYPE, INTERNAL_KIND},
+        {CVC4::Kind::TESTER_TYPE, INTERNAL_KIND},
+        {CVC4::Kind::TYPE_CONSTANT, INTERNAL_KIND},
+        {CVC4::Kind::FUNCTION_TYPE, INTERNAL_KIND},
+        {CVC4::Kind::SEXPR_TYPE, INTERNAL_KIND},
+        {CVC4::Kind::BITVECTOR_TYPE, INTERNAL_KIND},
+        {CVC4::Kind::FLOATINGPOINT_TYPE, INTERNAL_KIND},
+        /* ----------------------------------------------------------------- */
+        {CVC4::Kind::LAST_KIND, LAST_KIND},
+    };
```
mpreiner: Should we maybe add a function that converts internal to external kinds, which maps to `INERNAL_KIND` by default if the external kind is not in the map? This would get rid of all the `INTERNAL_KIND` mappings and does not force us to add internal only kinds to the map.

src/api/cvc4cpp.cpp
4tXJ7f: Can be `const`.

src/api/cvc4cpp.cpp
4tXJ7f: Can be `const static`.

src/api/cvc4cpp.cpp
4tXJ7f: Yeah, I was thinking about that, too. I guess it is a matter of performance at the end of the day.

src/api/cvc4cpp.cpp
mpreiner: And also add comment?

PR #2131, opened 2018-07-02
New C++ API: Implementation of Term.

aniemetz:
Note: Disabled code is disabled because the implementation of Sort and the Kind maps are not yet merged in.

### DIFF
--- a/src/api/cvc4cpp.cpp
+++ b/src/api/cvc4cpp.cpp
@@ -844,17 +844,15 @@ bool Term::operator!=(const Term& t) const
   return *d_expr != *t.d_expr;
 }
 
-#if 0
 Kind Term::getKind() const
 {
-  return s_kinds_internal[d_expr->getKind()];
+  return intToExtKind(d_expr->getKind());
 }
 
 Sort Term::getSort() const
 {
   return Sort(d_expr->getType());
 }
-#endif
 
 bool Term::isNull() const
 {


src/api/cvc4cpp.cpp
```diff
@@ -16,7 +16,11 @@
 
 #include "api/cvc4cpp.h"
 
+#include "base/cvc4_assert.h"
+#include "expr/expr.h"
+#include "expr/expr_manager.h"
 #include "util/result.h"
+#include "util/utility.h"
```
4tXJ7f: One of the Travis builds fails because std::find_if is defined in `<algorithm>` and that include is missing here.

aniemetz: It is actually missing in util/utility.h, so I added it there.

src/api/cvc4cpp.cpp
```diff
@@ -268,5 +273,229 @@ size_t SortHashFunction::operator()(const Sort& s) const {
...
+  if (it.d_iterator != nullptr)
+  {
+    d_iterator = new CVC4::Expr::const_iterator(
+        *static_cast<CVC4::Expr::const_iterator*>(it.d_iterator));
+  }
+}
+
+Term::const_iterator& Term::const_iterator::operator=(const const_iterator& it)
+{
+  if (d_iterator != nullptr)
```
4tXJ7f: Suggestion: `delete` of a `nullptr` is fine, so these checks can be removed.

PR #2132, opened 2018-07-02
New C++ API: Implementation of OpTerm.

aniemetz:
Note: Disabled code is disabled because the implementation of Sort and the Kind maps are not yet merged in.


4tXJ7f:
I marked this as "do not merge" because some things need to be enabled after other PRs are merged.

PR #2136, opened 2018-07-03
New C++ API: Implementation of datatype declaration classes.

aniemetz:


src/api/cvc4cpp.cpp
```diff
@@ -268,5 +270,117 @@ size_t SortHashFunction::operator()(const Sort& s) const {
...
+{
+  CVC4::Type t = *stor.d_sort.d_type;
+  if (t.isNull())
+  {
+    d_ctor->addArg(stor.d_name, DatatypeSelfType());
+  }
+  else
+  {
+    d_ctor->addArg(stor.d_name, t);
+  }
```
mpreiner: That could be a one liner: `d_ctor->addArg(stor.d_name, t.isNull() ? DatatypeSelfType() : t);`

aniemetz: Actually no, since t has type CVC4::Type and CVC4::DatatypeSelfType is not a CVC4::Type.

PR #2142, opened 2018-07-04
New C++ API: Implementation of datatype classes.

aniemetz:


src/expr/datatype.cpp
```diff
@@ -834,6 +839,12 @@ void DatatypeConstructor::setSygus(Expr op,
   d_sygus_pc = spc;
 }
 
+const std::vector<DatatypeConstructorArg>* DatatypeConstructor::getSelectors()
```
ajreynol: I would call this getArgs() and not getSelectors().

aniemetz: Done

PR #2143, opened 2018-07-04
New C++ API: Implementation of Solver class: Sort handling.

aniemetz:


src/api/cvc4cpp.cpp
```diff
@@ -1204,5 +1209,159 @@ std::ostream& operator<<(std::ostream& out,
   return out;
 }
 
+/* -------------------------------------------------------------------------- */
+/* Solver                                                                     */
+/* -------------------------------------------------------------------------- */
+
+Solver::Solver(Options* opts)
+{
+  d_opts = std::unique_ptr<Options>(new Options());
```
4tXJ7f: The order of data members in Solver is wrong: `d_opts` should be declared declared before `d_exprMgr` and before `d_smtEngine` otherwise we might run into troubles because `d_opts` gets destroyed before `d_smtEngine` and `d_exprMgr`. The initialization of `d_opts` can also be put in the initializer list.

src/api/cvc4cpp.cpp
```diff
@@ -1204,5 +1209,159 @@ std::ostream& operator<<(std::ostream& out,
...
+  // CHECK:
+  // for (unsigned i = 0; i < sorts.size(); ++ i)
+  //   !sorts[i].isFunctionLike()
+  // else "function-like types in tuples not allowed"
+  std::vector<Type> types = sortVectorToTypes(sorts);
+  return d_exprMgr->mkTupleType(types);
+}
+
+std::vector<Type> Solver::sortVectorToTypes(
+    const std::vector<Sort>& vector) const
```
4tXJ7f: Nit: Rename `vector` to `sorts`?

aniemetz: Done

4tXJ7f:
Marking as "do not merge" since @aniemetz has to decide in which order and when to merge these API PRs.

PR #2144, opened 2018-07-04
New C++ API: Implementation of Solver class: Term handling.

aniemetz:


src/api/cvc4cpp.cpp
```diff
@@ -1204,5 +1204,233 @@ std::ostream& operator<<(std::ostream& out,
...
+  // const unsigned n = children.size() - (mk == kind::metakind::PARAMETERIZED ?
+  // 1 : 0); n < minArity(kind) || n > maxArity(kind) else "Exprs with kind %s
+  // must have at least %u children and "
+  //      "at most %u children (the one under construction has %u)"
+  std::vector<Expr> echildren = termVectorToExprs(children);
+  return d_exprMgr->mkExpr(*opTerm.d_expr, echildren);
+}
+
+std::vector<Expr> Solver::termVectorToExprs(
+    const std::vector<Term>& vector) const
```
4tXJ7f: Nit: rename `vector` to `exprs`?

src/api/cvc4cpp.cpp
```diff
@@ -1204,5 +1204,233 @@ std::ostream& operator<<(std::ostream& out,
...
+  //      "to make variables and constants, see mkVar(), mkBoundVar(), "
+  //      "and mkConst()."
+  // CHECK:
+  // const unsigned n = children.size() - (mk == kind::metakind::PARAMETERIZED ?
+  // 1 : 0); n < minArity(kind) || n > maxArity(kind) else "Exprs with kind %s
+  // must have at least %u children and "
+  //      "at most %u children (the one under construction has %u)"
+  std::vector<Expr> echildren = termVectorToExprs(children);
+  CVC4::Kind k = extToIntKind(kind);
+  return kind::isAssociative(k) ? d_exprMgr->mkAssociative(k, echildren)
```
4tXJ7f: `isAssociative` is currently incomplete: https://github.com/CVC4/CVC4/blob/14b9dbaa0c9e8dce52d1a28595dc1cc80756abed/src/expr/kind_template.cpp#L39-L54

Are there plans to address this? I guess it is not super critical since `mkAssociative` just seems to break up operators with lots of children but we should have a plan going forward at least.

4tXJ7f:
Marking as "do not merge" since @aniemetz has to decide in which order and when to merge these API PRs.

Issue #2137, opened 2018-07-03
satisfiable unsat core


akgperson:
The error given when running input below it:
```
(error "Internal error detected
void CVC4::SmtEngine::checkUnsatCore()
../../../../src/smt/smt_engine.cpp:5292:

  
SmtEngine::checkUnsatCore(): produced core was satisfiable.")


(set-logic ALL)
(set-option :incremental true)
(declare-const v0 Bool)
(declare-const v1 Bool)
(declare-const v2 Bool)
(declare-const v3 Bool)
(declare-const i2 Int)
(declare-const i4 Int)
(declare-const i5 Int)
(declare-const i9 Int)
(declare-const i11 Int)
(declare-const i12 Int)
(declare-const r0 Real)
(declare-const v4 Bool)
(check-sat)
(declare-const i14 Int)
(assert v1)
(declare-const v5 Bool)
(assert (or v1 v0 (< 65736062 r0) v2 (< 65736062 r0) v1 v5 v0))
(declare-const v6 Bool)
(assert (< 65736062 r0))
(push 1)
(declare-const v7 Bool)
(declare-const v8 Bool)
(declare-const v9 Bool)
(declare-const r1 Real)
(declare-const v10 Bool)
(assert (not v3))
(assert (=> (or (>= 917.81677 489856 65736062) (and v0 (> 489856 65736062 (- r0)) v5 v3 v5 v5 v1 v3 v0 (or v1 v0 (< 65736062 r0) v2 (< 65736062 r0) v1 v5 v0) v6) (< 917.81677 r0 (- r0)) (> 489856 65736062 (- r0))) (<= i11 i14)))
(declare-const v11 Bool)
(declare-const v12 Bool)
(assert (= v8 (or (>= 917.81677 489856 65736062) (and v0 (> 489856 65736062 (- r0)) v5 v3 v5 v5 v1 v3 v0 (or v1 v0 (< 65736062 r0) v2 (< 65736062 r0) v1 v5 v0) v6) (< 917.81677 r0 (- r0)) (> 489856 65736062 (- r0))) (<= i11 i14) v7 (< i11 47) (>= 917.81677 489856 65736062)))
(assert v1)
(check-sat)
(pop 1)
(declare-const r2 Real)
(declare-const i15 Int)
(declare-const r3 Real)
(assert (xor v4 (< 917.81677 r0 (- r0)) (> 489856 65736062 (- r0))))
(declare-const i16 Int)
(declare-const r4 Real)
[...]

PR #2147, opened 2018-07-05
Properly cleanup assertion stack in CnfProof

4tXJ7f:
Fixes issue #2137. CnfProof has a stack of assertions that are being
converted to clauses. Previously, it could happen that while an
assertion was being added, TheoryProxy::explainPropagation() would be
called from Solver::reason() and push an assertion to the stack that was
then not removed. This lead to a clause id of the assertion being
associated with the explanation instead, which in turn could lead to a
wrong unsat core. This commit identifies two cases where
TheoryProxy::explainPropagation() is called without cleaning up the
assertion stack afterwards. It also adds an assertion that the assertion
stack must be empty when we are getting the unsat core.

### DIFF
--- a/src/prop/minisat/core/Solver.cc
+++ b/src/prop/minisat/core/Solver.cc
@@ -236,33 +236,34 @@ void Solver::resizeVars(int newSize) {
 }
 
 CRef Solver::reason(Var x) {
+  Debug("pf::sat") << "Solver::reason(" << x << ")" << std::endl;
 
-    // If we already have a reason, just return it
-    if (vardata[x].reason != CRef_Lazy) return vardata[x].reason;
+  // If we already have a reason, just return it
+  if (vardata[x].reason != CRef_Lazy) return vardata[x].reason;
 
-    // What's the literal we are trying to explain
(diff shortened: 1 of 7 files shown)


src/prop/minisat/core/Solver.cc
```diff
@@ -321,10 +322,15 @@ CRef Solver::reason(Var x) {
...
-           ProofManager::getCnfProof()->registerConvertedClause(id, true);
-           // no need to pop current assertion as this is not converted to cnf
-           );
+    PROOF(ClauseId id = ProofManager::getSatProof()->registerClause(
+              real_reason, THEORY_LEMMA);
+          ProofManager::getCnfProof()->registerConvertedClause(id, true);
+          // explainPropagation() pushes the explanation on the assertion stack
+          // in CnfProof, so we need to pop it here. This is important because
+          // reason() may be called indirectly while adding a clause, which can
+          // lead to a wrongassertion being associated with the clause being
```
mpreiner: ...wrong assertion...

4tXJ7f: What, you don't know what a wrongassertion is? :P

PR #2149, opened 2018-07-05
Change datastructure of CDList to std::vector

4tXJ7f:
Before, we were storing elements in a CDList by allocating raw buffers.
When reallocating space, we were moving the elements using memcpy
without invoking the copy or move constructors, which lead to warnings
with GCC 8.1 (and could be problematic, depending on the elements stored
in the list). This commit changes the underlying storage to use
std::vector. This implies a couple of changes: The destructor of the
elements in CDList is now always called, the elements must be
CopyAssignable (which does not work with classes that have const data
members), and some specialization is needed for CDList<bool> because
std::vector<bool> is (possibly) specialized in the STL to be a vector of
bits and doesn't support taking the address of an element. After this
change, the constructor of CDList now takes a `callCleanup` parameter
instead of the `callDestructor` to determine whether the cleanup
function should be called or not when truncating the list (we can
consider removing this parameter altogether in the future since the
default cleanup function is a noop). I've checked the usages of
CDList/CDQueue (which is based on CDList), and this should have no
effect.


timothy-king:
This is a rather core datastructure and some of the trickier invariants of the system rely on its behavior? What performance tests have you done for this change? Have you tested this under ASAN/MSAN/etc.?

src/context/cdlist.cpp
```diff
@@ -0,0 +1,55 @@
...
+ ** about std::vector<bool>.
+ **/
+
+#include "context/cdlist.h"
+
+namespace CVC4 {
+namespace context {
+
+template <>
+void CDList<bool, DefaultCleanUp<bool> >::truncateList(const size_t size)
```
timothy-king: Are there any CDList<bool>'s existing already? If not we could simply ban that expansion for now  with a static_assert and a comment.

src/context/cdlist.h
```diff
@@ -327,111 +134,184 @@ class CDList : public ContextObj {
    * and end() will have the same value (NULL).
    */
   class const_iterator {
```
timothy-king: Is using `using const_iterator = std::vector<T>::const_iterator` now an option?

src/context/cdlist.h
```diff
@@ -327,111 +134,184 @@ class CDList : public ContextObj {
...
+      {
+        --d_size;
+        d_cleanUp(&(d_list[d_size]));
+      }
     }
+    else
+    {
+      d_size = size;
+    }
+    d_list.erase(d_list.begin() + d_size, d_list.end());
```
timothy-king: Wouldn't resize() be faster?

src/context/cdlist.h
```diff
@@ -327,111 +134,184 @@ class CDList : public ContextObj {
...
+   */
+  void truncateList(const size_t size)
+  {
+    Assert(size <= d_size);
+    if (d_callCleanup)
+    {
+      while (d_size != size)
+      {
+        --d_size;
+        d_cleanUp(&(d_list[d_size]));
```
timothy-king: This is going to change the order in which the destructors are called on these objects. Though the clean-up order will be the same as the previous destruction order.

How did you check all of the instances that have a destructor/cleanup function for equivalence?

4tXJ7f:
You are correct that this is quite a fundamental change. So far, I've manually inspected the usages of CDList/CDQueue to check that we shouldn't be changing any semantics. I should and will run some performance tests/ASAN as well though.

4tXJ7f:
Alright, I was able to reproduce the issue with Ubuntu 14.04 (gcc 4.8.4).

aniemetz:
@4tXJ7f what's the status of this?

4tXJ7f:
I've updated the PR and changed the implementation s.t. a specialization for `std::vector<bool>` is not required anymore. When running with UBSan and ASan, there is one benchmark where an assertion fails but the assertion does not seem to be directly related to `CDList`. Investigating...

4tXJ7f:
Overall performance (tested on 33983 benchmarks from all logics using the benchmarking script and a 1200s timeout):

```
                                  status  total  solved    sat  unsat   best  timeout  memout  error  uniq  disagr  time_cpu    memory
config                                                                                                                                
2022-10-03-performance1000-vector     ok  33983   30201  11627  18574  23366     3644       1      0    14       0 5427719.8 6377726.7
2022-10-03-performance1000-main       ok  33983   30198  11620  18578  23401     3647       1      0    11       0 5428330.2 6379924.9
```

scatter_plot

mpreiner (approved):

PR #2154, opened 2018-07-05
Feature/fp rewrite improvement

martin-cs:
Improvements to the floating-point rewriter thanks to Andy's rewrite rule generator.

### DIFF
--- a/src/theory/fp/theory_fp_rewriter.cpp
+++ b/src/theory/fp/theory_fp_rewriter.cpp
@@ -255,6 +255,34 @@ namespace rewrite {
     } 
   }
 
+  RewriteResponse compactRemainder (TNode node, bool isPreRewrite) {
+    Assert(node.getKind() == kind::FLOATINGPOINT_REM);
+    Assert(!isPreRewrite);  // status assumes parts have been rewritten
+
+    Node working = node;
+
+    // (fp.rem (fp.rem X Y) Y) == (fp.rem X Y)
+    if (working[0].getKind() == kind::FLOATINGPOINT_REM && // short-cut matters!
+	working[0][1] == working[1]) {
+      working = working[0];
+    }
+
+    // Sign of the RHS does not matter
+    if (working[1].getKind() == kind::FLOATINGPOINT_NEG ||
+	working[1].getKind() == kind::FLOATINGPOINT_ABS) {
+      working[1] = working[1][0];
+    }
+
+    // Lift negation out of the LHS so it can be cancelled out
+    if (working[0].getKind() == kind::FLOATINGPOINT_NEG) {
+      NodeManager * nm = NodeManager::currentNM();
+      working = nm->mkNode(kind::FLOATINGPOINT_NEG,
+			   nm->mkNode(kind::FLOATINGPOINT_REM, working[0][0], working[1]));
+    }
+
+    return RewriteResponse(REWRITE_DONE, working);
+  }
+
 }; /* CVC4::theory::fp::rewrite */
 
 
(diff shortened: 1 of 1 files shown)


src/theory/fp/theory_fp_rewriter.cpp
```diff
@@ -69,7 +68,17 @@ namespace rewrite {
...
+    }
+
+    return RewriteResponse(REWRITE_DONE, node);
+  }
+
+  RewriteResponse compactAbs (TNode node, bool) {
+    Assert(node.getKind() == kind::FLOATINGPOINT_ABS);
+    if (node[0].getKind() == kind::FLOATINGPOINT_NEG ||
+	node[0].getKind() == kind::FLOATINGPOINT_ABS) {
+      return RewriteResponse(REWRITE_AGAIN, node[0][0]);
```
4tXJ7f: We could apply a similar rewrite in other cases such as removing all signs in a multiplication. E.g. `abs(-a * b) --> abs(a * b)`. This might be for another PR though.

src/theory/fp/theory_fp_rewriter.cpp
```diff
@@ -246,6 +255,34 @@ namespace rewrite {
...
+
+    // (fp.rem (fp.rem X Y) Y) == (fp.rem X Y)
+    if (working[0].getKind() == kind::FLOATINGPOINT_REM && // short-cut matters!
+	working[0][1] == working[1]) {
+      working = working[0];
+    }
+
+    // Sign of the RHS does not matter
+    if (working[1].getKind() == kind::FLOATINGPOINT_NEG ||
+	working[1].getKind() == kind::FLOATINGPOINT_ABS) {
```
4tXJ7f: Again, we could drop `FLOATINGPOINT_NEG`s in a multiplication (maybe a helper function for that would be good). Again, something for another PR.

src/theory/fp/theory_fp_rewriter.cpp
```diff
@@ -246,6 +255,34 @@ namespace rewrite {
...
+      working[1] = working[1][0];
+    }
+
+    // Lift negation out of the LHS so it can be cancelled out
+    if (working[0].getKind() == kind::FLOATINGPOINT_NEG) {
```
4tXJ7f: If we have `-x rem y --> -(x rem y)` shouldn't this rewrite also work for `abs(x) rem y --> abs(x rem y)`?

src/theory/fp/theory_fp_rewriter.cpp
martin-cs: Yeah... there are a lot of sign related rewrite but I think I want a better way of approaching them so it will be a different PR.

src/theory/fp/theory_fp_rewriter.cpp
martin-cs: It's on the TODO list.

src/theory/fp/theory_fp_rewriter.cpp
martin-cs: Yeah... so you'd think that and I thought that but the computer says no:

```
(set-logic QF_FP)
(set-info :status sat)
(declare-fun x () Float16)
(declare-fun y () Float16)
(declare-fun abs_rem () Float16)
(assert (= abs_rem (fp.abs (fp.rem x y))))
(declare-fun rem_abs () Float16)
(assert (= rem_abs (fp.rem (fp.abs x) y)))
(assert (not (= abs_rem rem_abs)))
(check-sat)
(get-model)
```

```
sat
(model
(define-fun x () (_ FloatingPoint 5 11) (fp #b1 #b01111 #b0000000000))
(define-fun y () (_ FloatingPoint 5 11) (fp #b1 #b01111 #b1000000000))
(define-fun abs_rem () (_ FloatingPoint 5 11) (fp #b0 #b01110 #b0000000000))
(define-fun rem_abs () (_ FloatingPoint 5 11) (fp #b1 #b01110 #b0000000000))
)
```

...  and now you're going to ask me why ...

x = -1.0
y = -1.5
abs_rem = abs(rem(-1.0, -1.5)) = abs(-1.0 - -1.5*1) = abs(0.5) = 0.5
rem_abs = rem(abs(-1.0), -1.5) = rem(1.0, -1.5) = 1.0 - 1.5*1 = -0.5

Floating point is weird!

martin-cs: BUT (fp.abs (fp.rem (fp.abs x) y)) = (fp.abs (fp.rem x y))) LOL

4tXJ7f: Hahahaha, I was considering to check it but then I thought "no way that this doesn't hold" and here we are...

martin-cs: This is why the rewrite rule synthesis is so valuable.  I've had a few "That must work; how could it not?" moments and also a few "There is no way those could be equal, surely it will fail when ...".

PR #2155, opened 2018-07-06
Add more sophisticated floating-point sampler

4tXJ7f:
This commit adds a floating-point sampler inspired by PyMPF [0]. It also
creates a new Sampler class that can be used for creating random values
of different sorts (currently BV and FP values).

[0] https://github.com/florianschanda/PyMPF


4tXJ7f:
@martin-cs This might be of interest to you.

src/theory/quantifiers/sygus_sampler.cpp
```diff
@@ -490,21 +491,13 @@ Node SygusSampler::getRandomValue(TypeNode tn)
...
-  else if (tn.isFloatingPoint() )
+  else if (tn.isFloatingPoint())
   {
-    // extremely naive uniform generation of floating points
     unsigned e = tn.getFloatingPointExponentSize();
     unsigned s = tn.getFloatingPointSignificandSize();
-    TypeNode bvt = nm->mkBitVectorType(e+s);
-    Node bvc = getRandomValue(bvt);
-    return nm->mkConst(FloatingPoint(e,s,bvc.getConst<BitVector>()));
+    return nm->mkConst(Sampler::pickFpBiased(e, s));
```
ajreynol: Can we add an option to toggle FpUniform and FpBiased here? This will be useful for evaluation, to show how much we gain with FpBiased.

PR #2164, opened 2018-07-13
New C++ API: Implementation of Solver class: OpTerm handling.

aniemetz:


src/api/cvc4cpp.cpp
```diff
@@ -2213,5 +2213,120 @@ std::vector<Expr> Solver::termVectorToExprs(
...
+{
+  // CHECK:
+  // kind == RECORD_UPDATE_OP
+  return d_exprMgr->mkConst(CVC4::RecordUpdate(arg));
+}
+
+OpTerm Solver::mkOpTerm(Kind kind, uint32_t arg)
+{
+  OpTerm res;
+  if (kind == DIVISIBLE_OP)
```
4tXJ7f: Suggestion: use a switch statement (more consistent with other code, warnings about duplicate cases, etc.).

src/api/cvc4cpp.cpp
```diff
@@ -2213,5 +2213,120 @@ std::vector<Expr> Solver::termVectorToExprs(
...
+  }
+  else if (kind == INT_TO_BITVECTOR_OP)
+  {
+    res = d_exprMgr->mkConst(CVC4::IntToBitVector(arg));
+  }
+  else if (kind == FLOATINGPOINT_TO_UBV_OP)
+  {
+    res = d_exprMgr->mkConst(CVC4::FloatingPointToUBV(arg));
+  }
+  else if (kind == FLOATINGPOINT_TO_UBV_TOTAL_OP)
```
4tXJ7f: This is somewhat orthogonal to this PR but we should check if users are supposed to be able to create `FLOATINGPOINT_TO_UBV_TOTAL_OP` and `FLOATINGPOINT_TO_SBV_TOTAL_OP` operators.

src/api/cvc4cpp.cpp
```diff
@@ -2213,5 +2213,120 @@ std::vector<Expr> Solver::termVectorToExprs(
...
+  return res;
+}
+
+OpTerm Solver::mkOpTerm(Kind kind, uint32_t arg1, uint32_t arg2)
+{
+  OpTerm res;
+  if (kind == BITVECTOR_EXTRACT_OP)
+  {
+    res = d_exprMgr->mkConst(CVC4::BitVectorExtract(arg1, arg2));
+    std::cout << res << std::endl;
```
4tXJ7f: This line should probably be removed (looks like it was for debugging).

src/api/cvc4cpp.cpp
```diff
@@ -2213,5 +2213,120 @@ std::vector<Expr> Solver::termVectorToExprs(
...
+  }
+  // CHECK: kind valid?
+  Assert(!res.isNull());
+  return res;
+}
+
+OpTerm Solver::mkOpTerm(Kind kind, uint32_t arg1, uint32_t arg2)
+{
+  OpTerm res;
+  if (kind == BITVECTOR_EXTRACT_OP)
```
4tXJ7f: Suggestion: change to switch statement.

src/api/cvc4cpp.cpp
aniemetz: According to @martin-cs those are internal. I'll remove in a separate PR.

PR #2169, opened 2018-07-13
Avoid ambiguous overloads in BitVector

4tXJ7f:
`long` is a 32-bit integer on Windows. CVC4's BitVector class had a
constructor for `unsigned int` and `unsigned long`, which lead to issues
with the new CVC4 C++ API because the two constructors were ambiguous
overloads. This commit changes the constructors to use `uint32_t` and
`uint64_t`, which are plattform independent and more explicit (mirroring
the API).


4tXJ7f:
I've added the comments, PTAL. I also removed the `static_cast`s in favor of writing `0u` in fp_converter.cpp because it feels weird to cast an integer literal.

PR #2171, opened 2018-07-14
Add floating-point support in evaluator

4tXJ7f:
Currently, the operations implemented by the FloatingPoint class are
fairly slow, so it is not always beneficial to use the Evaluator as
opposed to the rewriter because the rewriter does more aggressive
caching. Thus, the evaluator is disabled by default for floating-point
logics. Note that this commit also adds a check for the --sygus-eval-opt
flag in SygusSampler::evaluate() (previously, it was always used in that
method).

### DIFF
--- a/src/smt/smt_engine.cpp
+++ b/src/smt/smt_engine.cpp
@@ -1312,6 +1312,14 @@ void SmtEngine::setDefaults() {
         options::cbqi.set(true);
       }
     }
+    if (d_logic.isTheoryEnabled(THEORY_FP)
+        && !options::sygusEvalOpt.wasSetByUser())
+    {
+      Notice() << "SmtEngine: turning off evaluator for floating-point logics "
+                  "for better performance"
+               << endl;
+      options::sygusEvalOpt.set(false);
+    }
   }
 
   if (options::bitblastMode() == theory::bv::BITBLAST_MODE_EAGER)
--- a/src/theory/evaluator.cpp
+++ b/src/theory/evaluator.cpp
@@ -17,6 +17,7 @@
 #include "theory/evaluator.h"
 
 #include "theory/bv/theory_bv_utils.h"
+#include "theory/rewriter.h"
 #include "theory/theory.h"
 #include "util/integer.h"
 
@@ -33,6 +34,11 @@ EvalResult::EvalResult(const EvalResult& other)
       new (&d_bv) BitVector;
       d_bv = other.d_bv;
       break;
+    case ROUNDINGMODE:
+      new (&d_rm) RoundingMode;
+      d_rm = other.d_rm;
+      break;
+    case FLOATINGPOINT: new (&d_fp) FloatingPoint(other.d_fp); break;
     case RATIONAL:
       new (&d_rat) Rational;
       d_rat = other.d_rat;
@@ -57,6 +63,11 @@ EvalResult& EvalResult::operator=(const EvalResult& other)
         new (&d_bv) BitVector;
         d_bv = other.d_bv;
         break;
+      case ROUNDINGMODE:
+        new (&d_rm) RoundingMode;
+        d_rm = other.d_rm;
+        break;
+      case FLOATINGPOINT: new (&d_fp) FloatingPoint(other.d_fp); break;
       case RATIONAL:
         new (&d_rat) Rational;
         d_rat = other.d_rat;
@@ -80,6 +91,16 @@ EvalResult::~EvalResult()
       d_bv.~BitVector();
       break;
     }
+    case ROUNDINGMODE:
+    {
+      d_rm.~RoundingMode();
+      break;
+    }
+    case FLOATINGPOINT:
+    {
+      d_fp.~FloatingPoint();
+      break;
+    }
     case RATIONAL:
     {
       d_rat.~Rational();
@@ -89,9 +110,9 @@ EvalResult::~EvalResult()
     {
       d_str.~String();
       break;
-
-      default: break;
     }
+
+    default: break;
   }
 }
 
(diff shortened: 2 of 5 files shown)


src/theory/quantifiers/sygus/term_database_sygus.cpp
```diff
@@ -1572,11 +1572,6 @@ Node TermDbSygus::evaluateBuiltin(TypeNode tn,
     }
     if (!res.isNull())
     {
-      Assert(res
```
ajreynol: Why is this assertion removed?

src/smt/smt_engine.cpp
```diff
@@ -1312,6 +1312,14 @@ void SmtEngine::setDefaults() {
         options::cbqi.set(true);
       }
     }
+    if (d_logic.isTheoryEnabled(THEORY_FP)
```
ajreynol: Perhaps we should check that this is *pure* FP?  My concern is that many times we use the "ALL" logic and hence by default this would disable evaluation.

src/theory/quantifiers/sygus/term_database_sygus.cpp
4tXJ7f: The assertion is now in the evaluator itself, so that we don't have to have it at every call site: https://github.com/CVC4/CVC4/pull/2171/files#diff-09e6de4f37c686bc6c75e235aa393660R148

src/smt/smt_engine.cpp
4tXJ7f: Good point, will do.

src/theory/evaluator.cpp
```diff
@@ -524,6 +552,460 @@ EvalResult Evaluator::evalInternal(TNode n,
...
+        }
+
+        case kind::FLOATINGPOINT_EQ:
+        case kind::FLOATINGPOINT_GEQ:
+        case kind::FLOATINGPOINT_GT:
+        {
+          // These kinds should have been removed by earlier rewrites. In
+          // debug, we would like to know about this. In production, we can
+          // just return an invalid result and fall back to the full rewriter.
+          Assert(false);
```
martin-cs: This is true as things stand but this does make me a little bit nervous.

src/theory/evaluator.cpp
```diff
@@ -524,6 +552,460 @@ EvalResult Evaluator::evalInternal(TNode n,
...
+          const RoundingMode& rm = results[currNode[0]].d_rm;
+          const FloatingPoint& lhs = results[currNode[1]].d_fp;
+          const FloatingPoint& rhs = results[currNode[2]].d_fp;
+          results[currNode] = EvalResult(lhs.plus(rm, rhs));
+          break;
+        }
+
+        case kind::FLOATINGPOINT_SUB:
+        {
+          const RoundingMode& rm = results[currNode[0]].d_rm;
```
martin-cs: Sub is rewritten out.

src/theory/evaluator.cpp
```diff
@@ -524,6 +552,460 @@ EvalResult Evaluator::evalInternal(TNode n,
...
+          const FloatingPoint& val = results[currNode[1]].d_fp;
+          results[currNode] = EvalResult(val.rti(rm));
+          break;
+        }
+
+        case kind::FLOATINGPOINT_MIN:
+        {
+          const FloatingPoint& lhs = results[currNode[0]].d_fp;
```
martin-cs: min and max are rewritten to min and max total but ... I think that might be in expandDefintion rather than the rewriter.

src/theory/evaluator.cpp
```diff
@@ -524,6 +552,460 @@ EvalResult Evaluator::evalInternal(TNode n,
...
+          const BitVector& val = results[currNode[1]].d_bv;
+
+          results[currNode] =
+              EvalResult(FloatingPoint(param.t, rm, val, false));
+          break;
+        }
+
+        case kind::FLOATINGPOINT_TO_UBV:
+        {
+          TNode op = currNode.getOperator();
```
martin-cs: Also converted to _TOTAL version, as are other conversions from FP.

src/theory/evaluator.cpp
```diff
@@ -524,6 +552,460 @@ EvalResult Evaluator::evalInternal(TNode n,
...
+        {
+          const FloatingPoint& val = results[currNode[0]].d_fp;
+          const Rational& partial = results[currNode[1]].d_rat;
+
+          results[currNode] = EvalResult(val.convertToRationalTotal(partial));
+          break;
+        }
+
+#ifdef CVC4_USE_SYMFPU
+        case kind::FLOATINGPOINT_COMPONENT_NAN:
```
martin-cs: Not sure if you need to eval these?

src/theory/evaluator.h
```diff
@@ -88,6 +96,9 @@ class Evaluator
    * `args` and the corresponding values `vals`. The function returns a null
    * node if there is a subterm that is not constant under the substitution or
    * if an operator is not supported by the evaluator.
+   *
+   * Note: The evaluator expects that `n` has been rewritten to the point where
+   * it does not contain non-essential operators (e.g. FLOATINGPOINT_EQ).
    */
```
martin-cs: Ah; useful.  Could we Assert that rewriting doesn't change it?

4tXJ7f: Good point, I will add an assert.

src/theory/evaluator.cpp
4tXJ7f: Probably not, I was just trying to be as complete as possible.

src/theory/evaluator.cpp
4tXJ7f: Good to know, thanks!

src/theory/evaluator.cpp
4tXJ7f: Good to know, thanks!

src/theory/evaluator.cpp
4tXJ7f: Good to know, thanks!

src/theory/evaluator.cpp
4tXJ7f: Whenever we use the evaluator, we have a fallback to the normal rewriter: if the evaluator detects that it can't handle something, it returns a null node and the rewriter can give it a try. That means that this should be safe to do but we also get an alert that something changed when we are running with debug.

src/theory/quantifiers/sygus/term_database_sygus.cpp
ajreynol: Sounds good

PR #2172, opened 2018-07-14
Make conjecture generator's uf term enumeration safer

ajreynol:
This ensures that the function getEnumerateUfTerm does not have out of bounds accesses. This is an alternative fix to the one brought up in #2158.


src/theory/quantifiers/conjecture_generator.cpp
```diff
@@ -1087,25 +1093,41 @@ void ConjectureGenerator::getEnumerateUfTerm( Node n, unsigned num, std::vector<
...
+          if (index == vec.size())
+          {
+            // no more indices to iterate, we increment limit and reset below
             success = false;
           }
         }
       }
+      else
+      {
+        Assert(false);
```
4tXJ7f: Can you add a comment about why this Assertion failure is non-critical?

PR #2178, opened 2018-07-17
 sygusComp2018: pbe multi-enumerator fairness option

ajreynol:
This is PR 15/18 from the sygus comp 2018 branch (https://github.com/ajreynol/CVC4/tree/sygusComp2018-2).


src/theory/quantifiers/sygus/sygus_pbe.cpp
```diff
@@ -410,19 +410,29 @@ bool CegConjecturePbe::constructCandidates(const std::vector<Node>& enums,
...
       unsigned sz = d_tds->getSygusTermSize( enum_values[i] );
+      szs.push_back(sz);
       if( i==0 || sz<min_term_size ){
-        enum_consider.clear();
         min_term_size = sz;
-        enum_consider.push_back( i );
-      }else if( sz==min_term_size ){
+      }
+    }
+    unsigned allowDiff = options::sygusPbeMultiFair() ? 0 : 1;
```
HanielB: Why is the else case like that? Which behavior is being modeled and why?

ajreynol: Which else case? Do you mean what is the meaning of allowDiff?

HanielB: The case in which allowDiff is assigned to 1. I did not quite understand why "1", since it seems this only allows, when we are not "being fair", that only terms which are only one size bigger than the minimum can be added.

src/options/quantifiers_options.toml
```diff
@@ -1066,12 +1066,12 @@ header = "options/quantifiers_options.h"
...
 
 [[option]]
-  name       = "sygusPbeMultiFair"
+  name       = "sygusPbeMultiFairDiff"
   category   = "regular"
-  long       = "sygus-pbe-multi-fair"
-  type       = "bool"
-  default    = "true"
-  help       = "when using multiple enumerators, ensure that we only register value of minimial term size"
+  long       = "sygus-pbe-multi-fair-diff=N"
```
HanielB: I'd also like to have an option where N is "infnity", i.e. we don't really enforce any kind of fairness. What do you think?

ajreynol: Yeah you're right, I'll add the previous flag and this integer.

PR #2179, opened 2018-07-18
Optimizations and fixes for computing whether a type is finite

ajreynol:
A CAV 2018 paper by Hu and D'Antoni found that CVC4's sygus solver is extremely slow for sygus benchmarks with large grammars. For these benchmarks, the overwhelming bottleneck is computing whether or not a type is finite, which takes CVC4 several minutes to do, even for moderately large grammars. The reason is because the exact cardinality of the type was being computed.

This PR optimizes the computation of whether a type is finite. In particular, whenever possible, we avoid computing an exact cardinality for types.

This also fixes a few incorrect cases in isInterpretedFinite:  for finite model finding, arrays and functions with consistuent type with cardinality one were not treated as finite.

It also fixes an even more core issue in the cardinality utility, whose function isOne was wrong. This corrects existing issues for finite model finding for corecursive single datatypes.

@barrettcw , you may be interested in this pull request.


src/expr/type_node.cpp
```diff
@@ -83,46 +83,81 @@ bool TypeNode::isInterpretedFinite()
...
+      else
+      {
+        // If the consistuent type of the array has cardinality one, then the
+        // array type has cardinality one, independent of the index type.
+        isInterpretedFinite = tnc.getCardinality().isOne();
+      }
+    }
+    else if (isSet())
+    {
+      isInterpretedFinite = getSetElementType().isInterpretedFinite();
```
barrettcw: A set can have arbitrarily many elements, so isn't its cardinality infinite?

barrettcw: Ah - you are right.  Sorry about that!

kbansal:
Sorry for dropping-in from nowhere, but those fixes sound major enough to deserve a test file or two? :)

ajreynol:
Hi @kbansal, I added a sygus benchmark that goes from timeout -> solved easily by this PR.

FYI, the existing regressions already test this code quite a bit (this code is run on essentially every regression).  In fact, the regression regress0/ho/finite-fun-ext.smt2 caught one of the bugs that led to one of the fixes in this PR.

src/theory/quantifiers/term_enumeration.cpp
```diff
@@ -109,15 +109,18 @@ bool TermEnumeration::mayComplete(TypeNode tn)
...
-      Node oth = NodeManager::currentNM()->mkConst(Rational(1000));
-      Node eq = NodeManager::currentNM()->mkNode(LEQ, card, oth);
-      eq = Rewriter::rewrite(eq);
-      mc = eq.isConst() && eq.getConst<bool>();
+      Cardinality c = tn.getCardinality();
+      if( !c.isLargeFinite() )
+      {
+        Node card = NodeManager::currentNM()->mkConst(
+            Rational(c.getFiniteCardinality()));
+        Node oth = NodeManager::currentNM()->mkConst(Rational(1000));
```
HanielB: what about doing an option instead of this hardcoded limit? It'd be cool to be able to use FMF on bitvectors of bitwidth larger than 9.

src/theory/datatypes/datatypes_sygus.cpp
```diff
@@ -568,7 +568,9 @@ Node SygusSymBreakNew::getSimpleSymBreakPred(TypeNode tn,
           unsigned c1 = deq_child[0][i];
           unsigned c2 = deq_child[1][i];
           TypeNode tnc = children[c1].getType();
-          if (tnc == children[c2].getType() && !tnc.getCardinality().isOne())
+          // we may only apply this symmetry breaking scheme (which introduces
+          // disequalities) if the types are infinite.
+          if (tnc == children[c2].getType() && !tnc.isInterpretedFinite())
```
HanielB: you are strengthening the restriction here, right? Did you test if this leads to any loss of performance? (I imagine it must be hard to assess since `isOne` was wrong before)

src/expr/type_node.cpp
```diff
@@ -83,46 +83,81 @@ bool TypeNode::isInterpretedFinite()
...
-              break;
-            }
+            isInterpretedFinite = false;
+            break;
           }
         }
+        if (!isInterpretedFinite)
+        {
+          // similar to arrays, functions are finite if their range type
+          // has cardinality one, regardless of the arguments.
```
HanielB: then why not move this block to before checking the arguments and only check them if range is not of size 1?

src/theory/datatypes/datatypes_sygus.cpp
ajreynol: I'm making this so that the symmetry breaking kicks in less often (only for infinite types). Looking back at this code, I'm not convinced that it is sound if the types are finite, regardless of being cardinality one. So this makes the code safer.

src/expr/type_node.cpp
ajreynol: Because it is more expensive to compute the cardinality of type. This calls getCardinality() only if necessary.

ajreynol: A further note here: we could implement a dedicated function TypeNode::isCardinalityOne, but I didnt want the changes in this PR to become too large :)

HanielB: It'd be nice to open an issue then

PR #2187, opened 2018-07-20
 sygusComp2018: Improvements to CEGIS loop

ajreynol:
This is PR 17/18 from the sygus comp 2018 branch (https://github.com/ajreynol/CVC4/tree/sygusComp2018-2).

This adds the key optimization of "verification subcalls", and minor cleanup to the CEGIS loop.


src/theory/quantifiers/sygus/ce_guided_conjecture.cpp
```diff
@@ -392,22 +401,52 @@ void CegConjecture::doCheck(std::vector<Node>& lems)
...
+      }
+      else if (r.asSatisfiabilityResult().isSat() == Result::UNSAT)
+      {
+        // if the result in the subcall was unsatisfiable, we avoid
+        // rechecking, hence we drop "query" from the verification lemma
+        lem = d_quant.negate();
+      }
+      // In the rare case that the subcall is unknown, we add the verification
+      // lemma in the main solver. This should only happen if the quantifier
+      // free logic is undecidable.
```
HanielB: could we add some assertions here to be sure this does not happen in the decidable fragments?

ajreynol: I don't think there is a clear way to write this. We'd have to track an exact description of what is/isn't decidable over time.

PR #2194, opened 2018-07-23
Provide access methods to get heap/nil when using separation logic

ayveejay:
When using CVC4 via its API, there does not exist the capability to access the constructed `heap`/`nil` expression directly.

This branch provides a rudimentary getter for both the `heap` and `nil` expression, such that these constructs are accessible programatically via the SWIG API.

Is there a way to return multiple Expr nodes (e.g., using a `std::pair`) via SWIG? This would make the API nicer.

Are there any tests that should be written for introducing this kind of functionality?


aniemetz:
This PR does not conform to our style guidelines (see https://github.com/CVC4/CVC4/wiki/Code-Style-Guidelines). For future PRs, please refer to this guidelines.

ayveejay:
Hi @aniemetz -- you're totally correct, sorry about that. I've opened-up #2203 to correct this mistake!

4tXJ7f:
Thanks for your work @ayveejay. If you find time, could you add a check to throw a `RecoverableModalException` if separation logic is not being used (`d_logic.isTheoryEnabled()` should help there)? Also, it would be great to have a small system test for this (see [this](https://github.com/CVC4/CVC4/blob/master/test/system/boilerplate.cpp) for an example) to have a small demonstration of how this should be used and to test that we don't break it in the future.

ayveejay:
@4tXJ7f all changes actioned -- please see #2229.

PR #2195, opened 2018-07-23
Avoid explicit dependancy on Python 3

ayveejay:
Currently the `mkoptions.py` script explicitly depends on Python 3 -- it can actually be modified to run under Python 2, with some use of the `future`s package.

This PR simply introduces enough of futures to get `mkoptions.py` working via Python 2.

### DIFF
--- a/src/options/mkoptions.py
+++ b/src/options/mkoptions.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python3
+#!/usr/bin/env python
 """
     Generate option handling code and documentation in one pass. The generated
     files are only written to the destination file if the contents of the file


mpreiner:
Which version of Python 2 had issues with the script? The script should actually already work with Python 2.7.

ayveejay:
@mpreiner yep, you're correct! The only issue is related to the `#!` that explicitly calls `python3`.

I've updated the PR.

ayveejay:
Btw, I am happy to do this in a fresh branch (if you're also happy) to avoid the "commit noise".

mpreiner:
No worries, we can use this PR since we will do a squash commit anyways.

PR #2199, opened 2018-07-24
 New C++ API: First batch of commands (SMT-LIB and non-SMT-LIB).

aniemetz:


src/api/cvc4cpp.cpp
```diff
@@ -2248,6 +2248,97 @@ OpTerm Solver::mkOpTerm(Kind kind, uint32_t arg1, uint32_t arg2)
   return res;
 }
 
+/* Non-SMT-LIB commands                                                       */
+/* -------------------------------------------------------------------------- */
+
+Term Solver::simplify(const Term& t)
```
mpreiner: Missing space after `simplify`.

PR #2200, opened 2018-07-24
Improvements to sets + cardinality + quantifiers

ajreynol:
This PR includes several improvements to sets + cardinality + quantifiers, including:
- The use of choice functions to represent canonical instantiations for bounded set quantification,
- Simplifications to the minimization modes for fmf-bound (some combinations led to instantiations with free variables),
- Improve handling of cardinality in the sets solver:  (1) track which element types have cardinality and only build the cardinality graph (in terms of Bansal et al IJCAR 2016) for those types, and (2) only set incomplete if cardinality is applied to unhandled operators like universe/relations. This allows combinations of relations/universe and cardinality provided that the two are not explicitly combined.

### DIFF
--- a/src/options/options_handler.cpp
+++ b/src/options/options_handler.cpp
@@ -582,23 +582,6 @@ depth \n\
 \n\
 ";
 
-const std::string OptionsHandler::s_fmfBoundMinModeModeHelp = "\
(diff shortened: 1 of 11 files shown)


ajreynol:
I've addressed the review.

PR #2201, opened 2018-07-24
New C++ API: Second batch of commands (SMT-LIB).

aniemetz:


src/api/cvc4cpp.cpp
```diff
@@ -2353,5 +2353,234 @@ Sort Solver::declareDatatype(
   return mkDatatypeSort(dtdecl);
 }
 
+/**
+ *  ( declare-fun <symbol> () <sort> )
+ */
+Term Solver::declareFun (const std::string& symbol,
+                         Sort sort) const
```
4tXJ7f: Semi-related to this PR: we could make `Sort` and `Term` const references in these methods (to avoid the overhead of `std::shared_ptr`).

src/api/cvc4cpp.cpp
```diff
@@ -2353,5 +2353,234 @@ Sort Solver::declareDatatype(
...
+  // - bound_vars matches sort of funs
+  // - exprs matches sort of funs
+  // CHECK:
+  // CHECK:
+  // for bv in bound_vars (for v in bv): is bound var
+  std::vector<Expr> efuns = termVectorToExprs(funs);
+  std::vector<std::vector<Expr>> ebound_vars;
+  for (const auto& v : bound_vars)
+  {
+    ebound_vars.push_back (termVectorToExprs(v));
```
4tXJ7f: Apply clang-format.

src/api/cvc4cpp.cpp
aniemetz: Would that apply to more than only the methods in this PR?

4tXJ7f: Yes, this would apply to other methods as well. We have discussed this offline and we came to the conclusion that it is not urgent for now. We can revisit it once we are using the new API.

PR #2202, opened 2018-07-24
Extended rewriter for bit-vectors

ajreynol:
Includes many of the rewrites from this spreadsheet https://docs.google.com/spreadsheets/d/1nbZ6dWv8e4UVySjIdSyeDFPhkBAbh3z4SfxsbJ9L1VI/edit#gid=0

These rewrites are used in the extended rewriter, which is used by default for sygus solving only.


ajreynol:
Decided to break this up into multiple PRs, closing this.

PR #2204, opened 2018-07-25
 Move reg exp rewrites from prerewrite to postrewrite

ajreynol:
This also cleans some of the regular expression rewrite functions, for instance, the functions for re.inter and re.union are merged.

### DIFF
--- a/src/theory/strings/theory_strings_rewriter.cpp
+++ b/src/theory/strings/theory_strings_rewriter.cpp
@@ -547,86 +547,70 @@ Node TheoryStringsRewriter::rewriteStarRegExp(TNode node)
   return retNode;
 }
 
-Node TheoryStringsRewriter::rewriteOrRegExp(TNode node)
+Node TheoryStringsRewriter::rewriteAndOrRegExp(TNode node)
 {
-  Assert( node.getKind() == kind::REGEXP_UNION );
+  Kind nk = node.getKind();
+  Assert(nk == REGEXP_UNION || nk == REGEXP_INTER);
   Trace("strings-prerewrite")
-      << "Strings::rewriteOrRegExp start " << node << std::endl;
-  Node retNode = node;
+      << "Strings::rewriteAndOrRegExp start " << node << std::endl;
   std::vector<Node> node_vec;
-  bool allflag = false;
-  for(unsigned i=0; i<node.getNumChildren(); ++i) {
-    if(node[i].getKind() == kind::REGEXP_UNION) {
-      Node tmpNode = node[i];
-      for (unsigned int j = 0; j < tmpNode.getNumChildren(); ++j)
+  for (const Node& ni : node)
+  {
+    if (ni.getKind() == nk)
+    {
+      for (const Node& nic : ni)
       {
-        if (std::find(node_vec.begin(), node_vec.end(), tmpNode[j])
-            == node_vec.end())
+        if (std::find(node_vec.begin(), node_vec.end(), nic) == node_vec.end())
         {
-          if(std::find(node_vec.begin(), node_vec.end(), tmpNode[j]) == node_vec.end()) {
-            node_vec.push_back(tmpNode[j]);
-          }
+          node_vec.push_back(nic);
         }
       }
-    } else if(node[i].getKind() == kind::REGEXP_EMPTY) {
-      // can be removed
-    } else if(node[i].getKind() == kind::REGEXP_STAR && node[i][0].getKind() == kind::REGEXP_SIGMA) {
-      allflag = true;
-      retNode = node[i];
-      break;
-    } else {
-      if(std::find(node_vec.begin(), node_vec.end(), node[i]) == node_vec.end()) {
-        node_vec.push_back( node[i] );
-      }
     }
-  }
-  if(!allflag) {
-    std::vector< Node > nvec;
-    retNode = node_vec.size() == 0 ? NodeManager::currentNM()->mkNode( kind::REGEXP_EMPTY, nvec ) :
-          node_vec.size() == 1 ? node_vec[0] : NodeManager::currentNM()->mkNode(kind::REGEXP_UNION, node_vec);
-  }
-  Trace("strings-prerewrite")
-      << "Strings::rewriteOrRegExp end " << retNode << std::endl;
-  return retNode;
-}
-
-Node TheoryStringsRewriter::rewriteAndRegExp(TNode node)
-{
-  Assert( node.getKind() == kind::REGEXP_INTER );
-  Trace("strings-prerewrite")
-      << "Strings::rewriteAndRegExp start " << node << std::endl;
-  Node retNode = node;
-  std::vector<Node> node_vec;
-  //Node allNode = Node::null();
-  for(unsigned i=0; i<node.getNumChildren(); ++i) {
-    if(node[i].getKind() == kind::REGEXP_INTER) {
-      for (const Node& nc : node[i])
+    else if (ni.getKind() == REGEXP_EMPTY)
+    {
+      if (nk == REGEXP_INTER)
       {
-        if (std::find(node_vec.begin(), node_vec.end(), nc) == node_vec.end())
-        {
-          node_vec.push_back(nc);
-        }
+        return returnRewrite(node, ni, "re.and-empty");
       }
-    } else if(node[i].getKind() == kind::REGEXP_EMPTY) {
-      retNode = node[i];
-      break;
-    } else if(node[i].getKind() == kind::REGEXP_STAR && node[i][0].getKind() == kind::REGEXP_SIGMA) {
-      // can be removed
-    } else {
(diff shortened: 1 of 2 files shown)


src/theory/strings/theory_strings_rewriter.cpp
```diff
@@ -385,11 +385,12 @@ Node TheoryStringsRewriter::rewriteConcat(Node node)
   return retNode;
 }
 
-Node TheoryStringsRewriter::prerewriteConcatRegExp( TNode node ) {
+Node TheoryStringsRewriter::rewriteConcatRegExp(TNode node)
+{
   Assert( node.getKind() == kind::REGEXP_CONCAT );
   NodeManager* nm = NodeManager::currentNM();
   Trace("strings-prerewrite")
```
4tXJ7f: Also update the trace tag?

src/theory/strings/theory_strings_rewriter.cpp
```diff
@@ -498,98 +499,186 @@ Node TheoryStringsRewriter::prerewriteConcatRegExp( TNode node ) {
   return node;
 }
 
-Node TheoryStringsRewriter::prerewriteOrRegExp(TNode node) {
-  Assert( node.getKind() == kind::REGEXP_UNION );
-  Trace("strings-prerewrite") << "Strings::prerewriteOrRegExp start " << node << std::endl;
+Node TheoryStringsRewriter::rewriteStarRegExp(TNode node)
+{
```
4tXJ7f: Add an assert for the node kind.

src/theory/strings/theory_strings_rewriter.cpp
```diff
@@ -498,98 +499,186 @@ Node TheoryStringsRewriter::prerewriteConcatRegExp( TNode node ) {
...
 
-Node TheoryStringsRewriter::prerewriteAndRegExp(TNode node) {
-  Assert( node.getKind() == kind::REGEXP_INTER );
-  Trace("strings-prerewrite") << "Strings::prerewriteOrRegExp start " << node << std::endl;
-  Node retNode = node;
+Node TheoryStringsRewriter::rewriteAndOrRegExp(TNode node)
+{
+  Kind nk = node.getKind();
+  Assert(nk == REGEXP_UNION || nk == REGEXP_INTER);
+  Trace("strings-prerewrite")
```
4tXJ7f: Update trace tag?

src/theory/strings/theory_strings_rewriter.cpp
```diff
@@ -984,6 +1073,40 @@ RewriteResponse TheoryStringsRewriter::postRewrite(TNode node) {
   {
     retNode = rewriteStringCode(node);
   }
+  else if (node.getKind() == REGEXP_CONCAT)
```
4tXJ7f: Can we make this whole else-if chain into a switch statement? It also seems rather inefficient to use `getKind()` in every condition.

src/theory/strings/theory_strings_rewriter.cpp
```diff
@@ -984,6 +1073,40 @@ RewriteResponse TheoryStringsRewriter::postRewrite(TNode node) {
...
+    retNode = rewriteAndOrRegExp(node);
+  }
+  else if (node.getKind() == REGEXP_STAR)
+  {
+    retNode = rewriteStarRegExp(node);
+  }
+  else if (node.getKind() == REGEXP_PLUS)
+  {
+    retNode =
+        nm->mkNode(REGEXP_CONCAT, node[0], nm->mkNode(REGEXP_STAR, node[0]));
```
4tXJ7f: Since we are only doing this post-rewrite, we might miss rewrites of the `nm->mkNode(REGEXP_STAR, node[0])` part. We'd either to return REWRITE_AGAIN or do this as a prerewrite.

src/theory/strings/theory_strings_rewriter.cpp
```diff
@@ -498,98 +499,192 @@ Node TheoryStringsRewriter::prerewriteConcatRegExp( TNode node ) {
...
+    return returnRewrite(node, node[0], "re-star-empty-string");
+  }
+  else if (node[0].getKind() == REGEXP_EMPTY)
+  {
+    // (empty)* ---> ""
+    retNode = nm->mkNode(STRING_TO_REGEXP, nm->mkConst(String("")));
+    return returnRewrite(node, retNode, "re-star-empty");
+  }
+  else if (node[0].getKind() == REGEXP_UNION)
+  {
```
4tXJ7f: Add comment for this rewrite as well.

src/theory/strings/theory_strings_rewriter.cpp
```diff
@@ -498,98 +499,192 @@ Node TheoryStringsRewriter::prerewriteConcatRegExp( TNode node ) {
...
+  if (retNode != node)
+  {
+    // flattening and removing children, based on loop above
+    return returnRewrite(node, retNode, "re.andor-flatten");
+  }
+  return node;
+}
+
+Node TheoryStringsRewriter::rewriteLoopRegExp(TNode node)
+{
```
4tXJ7f: Add assertion for node kind.

src/theory/strings/theory_strings_rewriter.cpp
```diff
@@ -498,98 +499,192 @@ Node TheoryStringsRewriter::prerewriteConcatRegExp( TNode node ) {
...
+  Node retNode = node;
+  Node r = node[0];
+  if (r.getKind() == REGEXP_STAR)
+  {
+    return r;
```
4tXJ7f: Use `returnRewrite()`.

src/theory/strings/theory_strings_rewriter.cpp
```diff
@@ -498,98 +499,192 @@ Node TheoryStringsRewriter::prerewriteConcatRegExp( TNode node ) {
...
+  }
+  TNode n1 = node[1];
+  if (!n1.isConst())
+  {
+    throw LogicException("re.loop contains non-constant integer (1).");
```
4tXJ7f: Ugh, why isn't this an assertion? This should be caught in the type checker. It also feels like we should maybe treat `re.loop` as an indexed operator instead?

src/theory/strings/theory_strings_rewriter.cpp
```diff
@@ -498,98 +499,192 @@ Node TheoryStringsRewriter::prerewriteConcatRegExp( TNode node ) {
...
+  AlwaysAssert(rz <= n1.getConst<Rational>(),
+               "Negative integer in string REGEXP_LOOP (1)");
+  Assert(n1.getConst<Rational>() <= RMAXINT,
+         "Exceeded LONG_MAX in string REGEXP_LOOP (1)");
+  unsigned l = n1.getConst<Rational>().getNumerator().toUnsignedInt();
+  std::vector<Node> vec_nodes;
+  for (unsigned i = 0; i < l; i++)
+  {
+    vec_nodes.push_back(r);
+  }
```
4tXJ7f: Nit: loop can be replaced by: `std::vector<Node> vec_nodes(l, r);`

src/theory/strings/theory_strings_rewriter.cpp
```diff
@@ -498,98 +499,192 @@ Node TheoryStringsRewriter::prerewriteConcatRegExp( TNode node ) {
...
+                  ? rest
+                  : vec_nodes.size() == 1
+                        ? nm->mkNode(REGEXP_CONCAT, r, rest)
+                        : nm->mkNode(REGEXP_CONCAT,
+                                     nm->mkNode(REGEXP_CONCAT, vec_nodes),
+                                     rest);
+  }
+  Trace("strings-lp") << "Strings::lp " << node << " => " << retNode
+                      << std::endl;
   return retNode;
```
4tXJ7f: Use `returnRewrite`.

src/theory/strings/theory_strings_rewriter.cpp
ajreynol: I'll cache getKind().

src/theory/strings/theory_strings_rewriter.cpp
ajreynol: I think we are calling REWRITE_AGAIN_FULL actually on https://github.com/CVC4/CVC4/blob/master/src/theory/strings/theory_strings_rewriter.cpp#L992, so this should be fine.

src/theory/strings/theory_strings_rewriter.cpp
ajreynol: Yes I agree on both points. I'll turn this into an assertion for now.

src/theory/strings/theory_strings_rewriter.cpp
4tXJ7f: Hm, ok. Seems, rather inefficient to call it over-and-over again but I guess it is fine for now.

PR #2213, opened 2018-07-26
Fix issues with printing parametric datatypes in smt2

### LINKED ISSUE #1852: Assertion failure when dumping datatypes regression

### LINKED ISSUE #1870: Illegal argument error when parsing dumped SMT2.5 DT benchmark
```
$ builds/bin/cvc4 --preprocess-only --lang=smt2.5 --dump raw-benchmark -qq test/regress/regress0/datatypes/dt-param-card4-bool-sat.smt2 > dt.smt2
$ builds/bin/cvc4 --lang=smt2.5 dt.smt2
(error "Parse Error: dt.smt2:10.35: Illegal argument detected
CVC4::Type CVC4::SymbolTable::Implementation::lookupType(const string&, const std::vector<CVC4::Type>&) const
  `params' is a bad argument; expected p.first.size() == params.size() to hold
type constructor arity is wrong: `Pair' requires 0 parameters but was provided 2")
```

ajreynol:
This includes:
- Property type casting arguments of parametric constructors,
- Not printing out spurious declare-sort for intermediate parametric sorts and unresolved type constructors,
- Printing parameters properly for both the 2.5 and 2.6 forms for datatypes.

This fixes #1852 and fixes #1870 (those benchmarks can now be dumped/re-parsed).


src/printer/smt2/smt2_printer.cpp
```diff
@@ -1907,15 +1919,74 @@ static void toStream(std::ostream& out,
...
+      {
+        out << (j > 0 ? " " : "") << d.getParameter(j);
+      }
+    }
+    else
+    {
+      out << std::endl;
+      out << "ERROR: datatypes in each block must have identical parameter "
+             "lists.";
+      out << std::endl;
```
4tXJ7f: Instead of printing this to `out`, we should probably use `Error()` s.t. the error ends up on stderr instead of stdout (or in a file). We could also add an `AlwaysAssert(false)` to exit if this point is reached.

src/printer/smt2/smt2_printer.cpp
```diff
@@ -1907,15 +1919,74 @@ static void toStream(std::ostream& out,
...
+        out << ")";
+      }
     }
     out << ")";
   }
   else
   {
-    out << " () (";
+    out << " (";
+    // can only print if all datatypes in this block have the same parameters
```
4tXJ7f: Nit: Missing period at the end of this line (makes the comment confusing to parse).

src/printer/smt2/smt2_printer.cpp
```diff
@@ -1907,15 +1919,74 @@ static void toStream(std::ostream& out,
...
+    // different parameter lists.
+    bool success = true;
+    const Datatype& d = datatypes[0].getDatatype();
+    unsigned nparam = d.getNumParameters();
+    for (unsigned j = 1, ndt = datatypes.size(); j < ndt; j++)
+    {
+      const Datatype& dj = datatypes[j].getDatatype();
+      if (dj.getNumParameters() != nparam)
+      {
+        success = false;
```
4tXJ7f: Nit: Why not just break here and remove lines 1970-1973 (similar to the break statement on line 1966)?

ajreynol: Because lines 1970-1973 also handle the case where the else branch set success to false.

src/printer/smt2/smt2_printer.cpp
ajreynol: I'm following the convention here: https://github.com/CVC4/CVC4/blob/master/src/printer/smt2/smt2_printer.cpp#L1251

Let me know if you think this should change.  (There are advantages I suppose to printing on out, since it ensures that the user is informed of what went wrong in their printing).

src/printer/smt2/smt2_printer.cpp
4tXJ7f: But that one `break`s anyway :)

ajreynol: Right, so a break here is redundant :)

4tXJ7f: I don't know if it was the lack of caffeine or something else but I misread the code and missed the fact that the other `break` was for the inner loop.

src/printer/smt2/smt2_printer.cpp
4tXJ7f: I still think that errors belong on stderr but it looks like this is out of scope for this PR.

PR #2219, opened 2018-07-26
Fix issues with bv2nat

ajreynol:
This includes:
- A missing case in the smt2 printer,
- A bug in an inference of int2bv applied to bv2nat where the types are different.

### DIFF
--- a/src/theory/bv/theory_bv.cpp
+++ b/src/theory/bv/theory_bv.cpp
@@ -494,6 +494,9 @@ bool TheoryBV::doExtfInferences(std::vector<Node>& terms)
             Assert(ee->areEqual(parent[0], n));
             lem = nm->mkNode(kind::IMPLIES, parent[0].eqNode(n), lem);
           }
+          // this handles inferences of the form, e.g.:
+          //   ((_ int2bv 32) (bv2nat x)) == x (if their types are equivalent)
+          //   (bv2nat ((_ int2bv 32) x)) == x + k*2^32 for some k
           Trace("bv-extf-lemma")
               << "BV extf lemma (collapse) : " << lem << std::endl;
           d_out->lemma(lem);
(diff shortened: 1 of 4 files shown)


src/theory/bv/theory_bv.cpp
```diff
@@ -475,28 +475,36 @@ bool TheoryBV::doExtfInferences(std::vector<Node>& terms)
...
-          Assert(ee->areEqual(parent[0], n));
-          lem = nm->mkNode(kind::IMPLIES, parent[0].eqNode(n), lem);
+          if (parent[0] != n)
+          {
+            Assert(ee->areEqual(parent[0], n));
+            lem = nm->mkNode(kind::IMPLIES, parent[0].eqNode(n), lem);
+          }
+          // this handles inferences of the form, e.g.:
+          //   ((_ int2bv 32) (bv2nat x)) == x (if their types are equivalent)
+          //   (bv2nat ((_ int2bv 32) x)) == x + k*2^32 for some k
```
aniemetz: This is not restricted to 32 bit right? Maybe use `w` instead?

src/theory/bv/theory_bv.cpp
```diff
@@ -475,28 +475,36 @@ bool TheoryBV::doExtfInferences(std::vector<Node>& terms)
...
-              "int_bv_cong", t.getType(), "for int2bv/bv2nat congruence");
-          t = nm->mkNode(kind::PLUS, t, nm->mkNode(kind::MULT, coeff, k));
-        }
-        Node lem = parent.eqNode(t);
+          if (n.getKind() == kind::INT_TO_BITVECTOR)
+          {
+            Assert(t.getType().isInteger());
+            // congruent modulo 2^( bv width )
+            unsigned bvs = n.getType().getBitVectorSize();
+            Node coeff = nm->mkConst(Rational(Integer(1).multiplyByPow2(bvs)));
```
aniemetz: We have a utility function for this value: https://github.com/CVC4/CVC4/blob/master/src/theory/bv/theory_bv_utils.h#L93

aniemetz: Oh wait, this is not a bit-vector constant. Nevermind.

PR #2225, opened 2018-07-27
Fix option handler for lazy/bv-sat-solver combinations.

mpreiner:
Further, unifies all *limitHandler and *limitPerHandler to limitHandler.

### DIFF
--- a/src/options/options_handler.cpp
+++ b/src/options/options_handler.cpp
@@ -53,6 +53,28 @@
 namespace CVC4 {
 namespace options {
 
+// helper functions
+namespace {
+
+void throwLazyBBUnsupported(theory::bv::SatSolverMode m)
+{
+  std::string sat_solver;
+  if (m == theory::bv::SAT_SOLVER_CADICAL)
+  {
+    sat_solver = "CaDiCaL";
+  }
+  else
+  {
+    Assert(m == theory::bv::SAT_SOLVER_CRYPTOMINISAT);
+    sat_solver = "CryptoMiniSat";
+  }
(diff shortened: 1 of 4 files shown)


src/options/options_handler.cpp
```diff
@@ -1087,28 +1073,13 @@ theory::bv::SatSolverMode OptionsHandler::stringToSatSolver(std::string option,
   if(optarg == "minisat") {
     return theory::bv::SAT_SOLVER_MINISAT;
   } else if(optarg == "cryptominisat") {
-    
-    if (options::incrementalSolving() &&
```
4tXJ7f: Is CryptoMiniSat now supported with incremental solving?

src/options/options_handler.cpp
```diff
@@ -1178,14 +1147,12 @@ theory::bv::BitblastMode OptionsHandler::stringToBitblastMode(
     if (!options::bitvectorAlgebraicSolver.wasSetByUser()) {
       options::bitvectorAlgebraicSolver.set(true);
     }
+    if (options::bvSatSolver() != theory::bv::SAT_SOLVER_MINISAT)
+    {
+      throwLazyBBUnsupported(options::bvSatSolver());
+    }
     return theory::bv::BITBLAST_MODE_LAZY;
   } else if(optarg == "eager") {
-
```
4tXJ7f: Is this PR supposed to be merged after #1838?

mpreiner: Oh, I totally forgot that incremental eager is not yet merged in :-). Yes this should come after #1838 is merged.

PR #2228, opened 2018-07-27
Require argument description for non-{bool,void} options.

mpreiner:
Not adding the argument description is now an error.

Further, adds missing argument descriptions.


src/options/smt_options.toml
```diff
@@ -634,7 +634,7 @@ header = "options/smt_options.h"
 [[option]]
   name       = "solveRealAsInt"
   category   = "undocumented"
-  long       = "solve-real-as-int"
+  long       = "solve-real-as-int=N"
```
4tXJ7f: This is bool :)

mpreiner: I added a check for bool/void options that aborts if `=...` is specified.

PR #2229, opened 2018-07-28
Improvements and tests for the API around separation logic

ayveejay:
As part of #2194, @4tXJ7f asked if it was possible to make some tweaks around the separation logic "API". This PR actions those changes.

Four main modifications:

- Introduces a system that validating that, when not using `THEORY_SEP`, that it is not possible to obtain the separation logic heap/nil (`validate_exception()`)
- Introduces a system test demonstrating how to use the separation logic theory, and then how to use the "getters" to obtain and interrogate the heap/nil expressions (`validate_getters()`)
- Refactors the original getters to avoid duplicate code
- Add a check as part of the getters to ensure that `THEORY_SEP` is in use

### DIFF
--- a/src/smt/smt_engine.cpp
+++ b/src/smt/smt_engine.cpp
@@ -5280,13 +5280,13 @@ std::pair<Expr, Expr> SmtEngine::getSepHeapAndNilExpr(void)
     return std::make_pair(heap, nil);
   }
   InternalError(
-      "SmtEngine::getHeapExpr(): failed to obtain heap/nil expressions from "
+      "SmtEngine::getSepHeapExpr(): failed to obtain heap/nil expressions from "
       "theory model.");
 }
 
-Expr SmtEngine::getHeapExpr() { return getSepHeapAndNilExpr().first; }
+Expr SmtEngine::getSepHeapExpr() { return getSepHeapAndNilExpr().first; }
 
-Expr SmtEngine::getNilExpr() { return getSepHeapAndNilExpr().second; }
+Expr SmtEngine::getSepNilExpr() { return getSepHeapAndNilExpr().second; }
 
 void SmtEngine::checkUnsatCore() {
   Assert(options::unsatCores(), "cannot check unsat core if unsat cores are turned off");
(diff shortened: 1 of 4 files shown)


ayveejay:
I would like to add a test for the exception logic. Please do not merge yet.

ayveejay:
Also, and I'm totally easy, would we could add a parameter to `getSeparationLogicHeapOrNilExprs` (notice: renamed `And` to `Or`) such that the param controls which gets returned -- just a thought!

I'd really like to avoid duplicating this code (which is what I had originally) into the individual getters for both `Heap` and `Nil` -- I'm just open to the way which we do it (`std::pair` seemed best to me, but maybe this goes against some of the CVC4 "style").

timothy-king:
> Quick query: would you prefer it if we renamed getHeapExpr to be getSeparationLogicHeapExpr (similarly for Nil)?
I am okay with both names. I just want to avoid things like "SepLog". But I'll let others weigh in on this too. @aniemetz thoughts/preferences?

> std::pair seemed best to me, but maybe this goes against some of the CVC4 "style"
The std::pair seems fine to me.

aniemetz:
I used `mkSepNil` to create a separation logic nil expression in the new API (still experimental): https://github.com/CVC4/CVC4/blob/master/src/api/cvc4cpp.h#L1756
I wanted to avoid introducing long function names (since that is a pain in general IMO). `getSeparationLogicHeapOrNilExprs` is quite a mouthful IMO, I'd also be ok with `getSepHeapAndNilExpr`. This should be an And, right? Since you get both as a pair?
Similarly, `getSepHeapExpr` and `getSepNilExpr`.

ayveejay:
I suggested to change it from `And` to be `Or` such that you could do something like `getSeparationLogicHeapOrNilExprs(true /* heap */)` and avoid having the function return `std::pair`. I think @timothy-king and I are both happy with the idea of the function returning a `std:pair`, so we can probably ignore this alternative suggestion.

 I am happy to change it to `getSepHeapAndNilExpr` (over the current `getSeparationLogicHeapAndNilExprs`); I can also change it to be `getSepHeapExpr`/`getSepNilExpr`.

aniemetz:
@timothy-king you OK with that?

ayveejay:
I made the changes ... I can always back them out if @timothy-king prefers the old name.

timothy-king:
I would mildly prefer `getSeparationHeapAndNilExprs` over `getSepHeapAndNilExpr`. But I'm okay with the current proposal too. I'll leave the choice up to you.

PR #2232, opened 2018-07-30
 Proposal for adding map utility functions to CVC4.

timothy-king:


src/theory/quantifiers/fmf/bounded_integers.cpp
```diff
@@ -451,11 +453,14 @@ void BoundedIntegers::checkOwnership(Node f)
           success = true;
           //set Attributes on literals
           for( unsigned b=0; b<2; b++ ){
-            if( bound_lit_map[b].find( v )!=bound_lit_map[b].end() ){
-              Assert( bound_lit_pol_map[b].find( v )!=bound_lit_pol_map[b].end() );
+            if (Node* bound_lit_b_v = FindOrNull(bound_lit_map[b], v))
```
timothy-king: This is an example usage of these functions that addresses CID 1470212.

ajreynol:
For the change to bounded_integers, I would prefer:

```
          for( unsigned b=0; b<2; b++ ){          
            std::map< Node, Node >& blitMapB = bound_lit_map[b];
            std::map< Node, bool >& blitPolMapB = bound_lit_pol_map[b];
            if( boundLitMapB.find( v )!=boundLitMapB.end() ){
              Assert( blitPolMapB.find( v )!=blitPolMapB.end() );
              BoundIntLitAttribute bila;
              boundLitMapB[v].setAttribute( bila, blitPolMapB[v] ? 1 : 0 );
            }else{
              Assert( it->second!=BOUND_INT_RANGE );
            }
          }
```

Would this be enough to avoid the coverity warnings?

timothy-king:
@ajreynol To answer your question, that is close to being equivalent to the original. The side effect of inserting into blitPolMapB needs to be moved into the if condition to make it equivalent. If this is the semantics it should have, this is fine. There is also an additional lookup in this version when assertions are compiled in on the blitPolMapB compared to the version proposed in this CL.

But your preference does get at the heart of what this CL is taking about. I am essentially proposing 4 shorthands for using the STL style maps in a more declarative fashion for very common boilerplate [and maybe more in the future]. This is largely a stylistic and readability choice for the team. If the team does not believe that this is ultimately more readable, this should not go forward.

mpreiner:
@ajreynol Your proposed version is a lot less readable. Why exactly do you prefer your solution? I like how the proposed helper functions make the code more readable, cleaner and makes coverity complain less.

ajreynol:
@timothy-king, for me, I prefer having the code use as little custom defined functions as possible. The reason is when I am debugging I want fewer places to have to look in the code and fewer things to have to remember what they do. Thus I tend to prefer being more verbose to being elegant.

This issue is sort of orthogonal to the fact that many places in the code may have side effects based on map lookups. The one in the function you point out is likely due to the fact that I wrote this code hastily. I am in favor of cleaning these up, but for the parts of my code, I would prefer using the more standard style.

timothy-king:
@ajreynol Adding these will increase the team's collective cognitive load to remember what these 3 functions do. The hope would be that they are sufficiently simple and self descriptive so that the benefit in readability will decrease the cost of maintaining the code to the team and will outweigh the costs of making everyone remember these and maintaining the extra library. Where folks come down on readability though is subjective. I suggest discussing this at the weekly meeting.

src/base/map_util.h
```diff
@@ -0,0 +1,94 @@
...
+ **
+ ** Currently implemented classes of functions:
+ ** - ContainsKey
+ **   Returns true if a map contains a key. (Also Supports std::set and
+ **   std::unordered_set.)
+ ** - FindOr*
+ **   Finds an data element mapped to by the map. Variants include FindOrNull
+ **   and FindOrDie.
+ **
+ ** Potential future classes of functions:
```
barrettcw: Is there supposed to be a list of future functionalities here?

src/context/cdhashmap.h
```diff
@@ -101,13 +101,26 @@ class CDOhash_map : public ContextObj {
...
+  // The type of the <Key, Data> pair mapped by this class.
+  //
+  // Implementation:
+  // The data and key visible to users of CDHashMap are only visible through
+  // const references. Thus the type of dereferencing a
+  // CDHashMap<Key, Data>::iterator.second should be value_type::second_type
+  // should be a `const Data&`. (Otherwise, access operations would need to
+  // makeCurrent() to get a Data&, which is an unacceptable performance hit.)
+  // To allow for the desired updating in other context, we store a
+  // std::pair<const Key, const Data> and break the const encapsulation when
```
barrettcw: contexts

src/base/map_util.h
timothy-king: Added the two candidates I am the most eager to get to.

src/context/cdhashmap.h
timothy-king: I went with "scenarios" as "contexts" is potentially confusing here.

timothy-king:
Removed the changes from bounded_integers.cpp as discussed in the meeting. Adding a comment to describe the intention of the Assert statement.

src/context/cdhashmap.h
```diff
@@ -101,13 +101,26 @@ class CDOhash_map : public ContextObj {
...
  public:
-  // The type of the <Key, Data> value mapped to by this class.
-  using value_type = std::pair<Key, Data>;
+  // The type of the <Key, Data> pair mapped by this class.
+  //
+  // Implementation:
+  // The data and key visible to users of CDHashMap are only visible through
+  // const references. Thus the type of dereferencing a
+  // CDHashMap<Key, Data>::iterator.second should be value_type::second_type
+  // should be a `const Data&`. (Otherwise, access operations would need to
```
aniemetz: Sentence doesn't quite parse, remove either `should be value_type::second_type` or `should be a const Data&.` ?

src/theory/quantifiers/fmf/bounded_integers.cpp
```diff
@@ -451,11 +452,19 @@ void BoundedIntegers::checkOwnership(Node f)
...
           for( unsigned b=0; b<2; b++ ){
-            if( bound_lit_map[b].find( v )!=bound_lit_map[b].end() ){
-              Assert( bound_lit_pol_map[b].find( v )!=bound_lit_pol_map[b].end() );
+            if (bound_lit_map[b].find(v) != bound_lit_map[b].end())
+            {
+              // This assertion may fail. We intentionally do not enable this in
+              // production as it is considered safe for this to fail. We fail
+              // in debug mode to have this instances raised to our attention.
+              Assert(bound_lit_pol_map[b].find(v)
+                     != bound_lit_pol_map[b].end());
```
aniemetz: This would be a candidate for the future "warning" (or whatever we will name it) macro, right? (Referring to our discussion on Wednesday, here.) Any ideas how we could keep track of newly introduced ones that are still Asserts since we don't have the macro in place yet?

src/context/cdhashmap.h
timothy-king: PTAL

src/theory/quantifiers/fmf/bounded_integers.cpp
timothy-king: I added WARNING_CANDIDATE as a comment. This should make tracking easy via grep.

src/base/map_util.h
```diff
@@ -0,0 +1,96 @@
...
+
+#include "base/cvc4_check.h"
+
+namespace CVC4 {
+
+// Returns true if the `map` contains the `key`.
+//
+// Supports sets as well.
+template <class M, class Key>
+bool ContainsKey(const M& map, const Key& key)
```
4tXJ7f: Nit: I think that we are lacking an explicit guideline on this right now but we generally use lower camelCase for function names.

src/base/map_util.h
```diff
@@ -0,0 +1,96 @@
...
+ **
+ ** \brief Utility functions for dealing with maps in a mostly uniform fashion.
+ **
+ ** Utility functions for dealing with maps and related classed in a mostly
+ ** uniform fashion. These are stylistically encouraged (but not required) in
+ ** new code. Supports:
+ ** - std::map
+ ** - std::unordered_map
+ ** - CVC4::context::CDHashmap
+ ** - CVC4::context::CDInsertHashmap
```
4tXJ7f: Also supports set types, no?

src/base/map_util.h
```diff
@@ -0,0 +1,96 @@
...
+ **   and FindOrDie.
+ **
+ ** Potential future classes of functions:
+ ** - InsertOrUpdate
+ ** - InsertIfNotPresent
+ **/
+
+#include "cvc4_private.h"
+
+#ifndef __CVC4__MAP_UTIL_H
```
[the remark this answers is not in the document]
timothy-king: Done.

src/base/map_util.h
timothy-king: The style guide is consistent with using mixed case in examples for functions: [FooFactory and FooConsumer](https://github.com/CVC4/CVC4/wiki/Code-Style-Guidelines#ownership-and-smart-pointers) and [Function1 and Function2](https://github.com/CVC4/CVC4/wiki/Code-Style-Guidelines#nonmember-static-member-and-global-functions). FWIW the [google style guide](https://google.github.io/styleguide/cppguide.html#Function_Names) suggests mixed case for functions. My personal preference also goes against this for free functions. (Methods are more ambiguous and I think this is what you are talking about.) In absence of a guideline to the contrary, I think we should just move forward with this as is. If you think renaming this is an improvement, I recommend doing this in a follow up.

src/base/map_util.h
timothy-king: Clarified.

src/theory/quantifiers/fmf/bounded_integers.cpp
```diff
@@ -451,11 +452,20 @@ void BoundedIntegers::checkOwnership(Node f)
...
           //set Attributes on literals
           for( unsigned b=0; b<2; b++ ){
-            if( bound_lit_map[b].find( v )!=bound_lit_map[b].end() ){
-              Assert( bound_lit_pol_map[b].find( v )!=bound_lit_pol_map[b].end() );
+            if (bound_lit_map[b].find(v) != bound_lit_map[b].end())
+            {
+              // WARNING_CANDIDATE:
+              // This assertion may fail. We intentionally do not enable this in
+              // production as it is considered safe for this to fail. We fail
+              // in debug mode to have this instances raised to our attention.
```
barrettcw: Nit: "instances" -> "instance"

PR #2236, opened 2018-07-31
Make candidate rewrite match filtering handle polymorphic operators properly

### LINKED ISSUE #1923: MatchTrie should handle type polymorphism correctly

ajreynol:
Currently, the discrimination tree index used for candidate rewrite rule filtering based on matching does not properly distinguish polymorphic operators, which leads to type errors. This makes the index handle them correctly.

Fixes #1923.

### DIFF
--- a/src/theory/quantifiers/candidate_rewrite_filter.cpp
+++ b/src/theory/quantifiers/candidate_rewrite_filter.cpp
@@ -219,15 +219,12 @@ void CandidateRewriteFilter::initialize(SygusSampler* ss,
   // initialize members of this class
   d_match_trie.clear();
   d_pairs.clear();
-  if (options::sygusRewSynthFilterCong())
-  {
-    // initialize the dynamic rewriter
-    std::stringstream ss;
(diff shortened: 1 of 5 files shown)


src/theory/quantifiers/candidate_rewrite_filter.cpp
```diff
@@ -379,18 +384,29 @@ bool CandidateRewriteFilter::notify(Node s,
     for (unsigned i = 0, size = vars.size(); i < size; i++)
     {
       Trace("crf-match") << "    " << vars[i] << " -> " << subs[i] << std::endl;
-      // TODO (#1923) ensure that we use an internal representation to
-      // ensure polymorphism is handled correctly
-      Assert(vars[i].getType().isComparableTo(subs[i].getType()));
     }
   }
+#ifdef CVC4_ASSERTIONS
```
4tXJ7f: Note to self: we should have a way to check this without resorting to `#ifdef`s.

PR #2238, opened 2018-07-31
Fixes for sygus inference

ajreynol:
This includes:
- Enabling sygus-specific options in SmtEngine::setDefaults,
- Disabling a variant of miniscoping (triggered by many chc-comp18 benchmarks),
- Treating free constants as functions to synthesize


src/smt/smt_engine.cpp
```diff
@@ -2050,6 +2052,10 @@ void SmtEngine::setDefaults() {
     if( !options::miniscopeQuantFreeVar.wasSetByUser() ){
       options::miniscopeQuantFreeVar.set( false );
     }
+    if (!options::quantSplit.wasSetByUser())
+    {
+      options::quantSplit.set(false);
```
HanielB: did you evaluate the impact of removing this is default with CEGQI?

src/theory/quantifiers/sygus_inference.cpp
```diff
@@ -160,7 +169,7 @@ bool SygusInference::simplify(std::vector<Node>& assertions)
   }
 
   // if no free function symbols, there is no use changing into SyGuS
-  if (free_functions.empty())
+  if (free_functions_uf.empty())
```
HanielB: please add an option to represent this behavior of only inferring when non-constant free functions exist

src/theory/quantifiers/sygus_inference.cpp
```diff
@@ -252,16 +261,21 @@ bool SygusInference::simplify(std::vector<Node>& assertions)
     if (itffv != ff_var_to_ff.end())
     {
       Node ff = itffv->second;
+      Expr body = it->second;
       std::vector<Expr> args;
-      for (const Node& v : lambda[0])
+      if (lambda.getKind() == LAMBDA)
```
HanielB: add a comment to say this is because of the constant functions

src/smt/smt_engine.cpp
ajreynol: This CEGQI is "sygus CEGQI". So if this ever kicked in before, we'd segfault :)

src/theory/quantifiers/sygus_inference.cpp
ajreynol: Actually i think the default behavior should be to always do this. I think if the user doesnt want sygus inference to kick in in this case, they should just disable the original option. I'll remove this check.

PR #2253, opened 2018-08-02
Adding less consistency lemmas in bv-ackermann preprocessing pass

yoni206:
The bv-ackermannization preprocessing pass adds consistency lemmas for each occurrence of a function application in the input formula.
However, sometimes not all occurrences are needed.
For example, if the input formula is: `f(f(x))=g(f(x))` then no consistency lemma is needed, because a concrete value for f(x) is not needed for a solution.

This PR changes bv-ackermann, so that it adds consistency lemmas only for function applications that appear at "top-level", that is, not as sub-terms of another function application.
Consistency lemmas are now added while collecting the functions and the arguments.

### DIFF
--- a/src/preprocessing/passes/bv_ackermann.cpp
+++ b/src/preprocessing/passes/bv_ackermann.cpp
@@ -49,8 +49,7 @@ void addLemmaForPair(TNode args1,
   if (args1.getKind() == kind::APPLY_UF)
   {
     Assert(args1.getOperator() == func);
-    Assert(args2.getKind() == kind::APPLY_UF
-                 && args2.getOperator() == func);
+    Assert(args2.getKind() == kind::APPLY_UF && args2.getOperator() == func);
     Assert(args1.getNumChildren() == args2.getNumChildren());
 
     std::vector<Node> eqs(args1.getNumChildren());
@@ -90,11 +89,10 @@ void storeFunctionAndAddLemmas(TNode func,
   if (set.find(term) == set.end())
   {
     TypeNode tn = term.getType();
-    Node skolem = nm->mkSkolem(
-        "BVSKOLEM$$",
-        tn,
-        "is a variable created by the ackermannization "
-        "preprocessing pass for theory BV");
+    Node skolem = nm->mkSkolem("BVSKOLEM$$",
+                               tn,
+                               "is a variable created by the ackermannization "
+                               "preprocessing pass for theory BV");
     for (const auto& t : set)
     {
       addLemmaForPair(t, term, func, assertions, nm);
@@ -107,16 +105,17 @@ void storeFunctionAndAddLemmas(TNode func,
      * (otherwise, no lemma is generated).
      * Therefore, we defer this for term in case it is the first element in the
      * set*/
-    if (set.size() == 2) 
+    if (set.size() == 2)
     {
       for (TNode elem : set)
       {
         vec->insert(vec->end(), elem.begin(), elem.end());
       }
-    } else if (set.size() > 2)
+    }
+    else if (set.size() > 2)
     {
       vec->insert(vec->end(), term.begin(), term.end());
-    } 
+    }
   }
 }
 
@@ -136,7 +135,7 @@ void collectFunctionsAndLemmas(FunctionToArgsMap& fun_to_args,
                                std::vector<TNode>* vec,
                                AssertionPipeline* assertions)
 {
-  TNodeSet seen;       
+  TNodeSet seen;
   NodeManager* nm = NodeManager::currentNM();
   TNode term;
   while (!vec->empty())
@@ -145,33 +144,33 @@ void collectFunctionsAndLemmas(FunctionToArgsMap& fun_to_args,
     vec->pop_back();
     if (seen.find(term) == seen.end())
     {
-        TNode func;
-        if (term.getKind() == kind::APPLY_UF)
+      TNode func;
+      if (term.getKind() == kind::APPLY_UF)
+      {
+        storeFunctionAndAddLemmas(term.getOperator(),
+                                  term,
+                                  fun_to_args,
+                                  fun_to_skolem,
+                                  assertions,
+                                  nm,
+                                  vec);
+      }
+      else if (term.getKind() == kind::SELECT)
+      {
+        storeFunctionAndAddLemmas(
+            term[0], term, fun_to_args, fun_to_skolem, assertions, nm, vec);
+      }
+      else
+      {
+        AlwaysAssert(
+            term.getKind() != kind::STORE,
+            "Cannot use eager bitblasting on QF_ABV formula with stores");
+        /* add children to the vector, so that they are processed later */
+        for (TNode n : term)
         {
-          storeFunctionAndAddLemmas(term.getOperator(),
-                                    term,
-                                    fun_to_args,
-                                    fun_to_skolem,
-                                    assertions,
-                                    nm,
-                                    vec);
+          vec->push_back(n);
         }
-        else if (term.getKind() == kind::SELECT)
-        {
-          storeFunctionAndAddLemmas(
-              term[0], term, fun_to_args, fun_to_skolem, assertions, nm, vec);
-        }
-        else
-        {
-          AlwaysAssert(
-              term.getKind() != kind::STORE,
-              "Cannot use eager bitblasting on QF_ABV formula with stores");
-          /* add children to the vector, so that they are processed later */
-          for (TNode n : term)
-          {
-            vec->push_back(n);
-          }
-        }
+      }
       seen.insert(term);
     }
   }
--- a/src/preprocessing/passes/bv_ackermann.cpp
+++ b/src/preprocessing/passes/bv_ackermann.cpp
@@ -46,8 +46,7 @@ void addLemmaForPair(TNode args1,
                      TNode args2,
                      const TNode func,
                      AssertionPipeline* assertionsToPreprocess,
-                     NodeManager* nm,
-                     std::stack<TNode>* stack)
+                     NodeManager* nm)
 {
   Node args_eq;
 
@@ -102,7 +101,7 @@ void storeFunctionAndAddLemmas(TNode func,
         "preprocessing pass for theory BV");
     for (const auto& t : set)
     {
-      addLemmaForPair(t, term, func, assertions, nm, stack);
+      addLemmaForPair(t, term, func, assertions, nm);
     }
     set.insert(term);
     fun_to_skolem.addSubstitution(term, skolem);
--- a/src/preprocessing/passes/bv_ackermann.cpp
+++ b/src/preprocessing/passes/bv_ackermann.cpp
@@ -79,7 +79,6 @@ void addLemmaForPair(TNode args1,
   assertionsToPreprocess->push_back(lemma);
   /* add the constraint to the stack so that
   * collectFunctionsAndLemmas will process it as well. */
-  stack->push(lemma);
 }
 
 void storeFunctionAndAddLemmas(TNode func,
@@ -109,6 +108,10 @@ void storeFunctionAndAddLemmas(TNode func,
     }
     set.insert(term);
     fun_to_skolem.addSubstitution(term, skolem);
+    for (TNode arg : term)
+    {
+      stack->push(arg);
+    }
   }
 }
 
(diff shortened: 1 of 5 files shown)


src/preprocessing/passes/bv_ackermann.cpp
```diff
@@ -41,11 +42,54 @@ namespace passes {
...
+  * collectFunctionsAndLemmas will process it as well. */
+  stack->push(lemma);
+}
+
+void storeFunctionAndAddLemmas(TNode func,
+                               TNode term,
+                               FunctionToArgsMap& fun_to_args,
+                               SubstitutionMap& fun_to_skolem,
+                               AssertionPipeline* assertions,
+                               NodeManager* nm,
```
barrettcw: Since you're passing in the node manager, why not use it below (instead of calling NodeManager::currentNM())?

src/preprocessing/passes/bv_ackermann.cpp
```diff
@@ -54,42 +98,69 @@ void storeFunction(
...
 }
 
-void collectFunctionSymbols(
-    TNode term,
-    FunctionToArgsMap& fun_to_args,
-    SubstitutionMap& fun_to_skolem,
-    std::unordered_set<TNode, TNodeHashFunction>& seen)
+/* We only add top-level applications of functions.
+ * For example: when we see "f(g(x))", we do not add g is a function and x as a
+ * parameter.
```
barrettcw: g as a function

src/preprocessing/passes/bv_ackermann.cpp
```diff
@@ -41,11 +42,54 @@ namespace passes {
...
+                     const TNode func,
+                     AssertionPipeline* assertionsToPreprocess,
+                     NodeManager* nm,
+                     std::stack<TNode>* stack)
+{
+  Node args_eq;
+
+  if (args1.getKind() == kind::APPLY_UF)
+  {
+    AlwaysAssert(args1.getKind() == kind::APPLY_UF
```
barrettcw: Let's get rid of the first part of this AlwaysAssert as it is the same as the if-condition

barrettcw:
What's the reason for the latest commit?  Did you find a bug with the previous version?

yoni206:
yes, I've entered an infinite loop because I pushed previously processed terms into the stack.

src/preprocessing/passes/bv_ackermann.cpp
```diff
@@ -79,7 +79,6 @@ void addLemmaForPair(TNode args1,
   assertionsToPreprocess->push_back(lemma);
   /* add the constraint to the stack so that
   * collectFunctionsAndLemmas will process it as well. */
-  stack->push(lemma);
 }
```
barrettcw: Also, stack is no longer needed as an argument to this function.

src/preprocessing/passes/bv_ackermann.cpp
```diff
@@ -54,42 +94,82 @@ void storeFunction(
...
         "preprocessing pass for theory BV");
+    for (const auto& t : set)
+    {
+      addLemmaForPair(t, term, func, assertions, nm, stack);
+    }
+    set.insert(term);
     fun_to_skolem.addSubstitution(term, skolem);
+    /* add the arguments to the stack so that
+     * collectFunctionsAndLemmas will process them as well. */
+    for (TNode arg : term)
```
barrettcw: A slightly tricky optimization can be added: if set has only one term in it, we don't need to add its arguments to the stack.  Of course, then, as soon as it has two terms, we have to add args from both terms to the stack.

src/preprocessing/passes/bv_ackermann.cpp
```diff
@@ -54,42 +94,82 @@ void storeFunction(
...
+          AlwaysAssert(
+              term.getKind() != kind::STORE,
+              "Cannot use eager bitblasting on QF_ABV formula with stores");
+          /* add childrens to the stack, so that they are processed later */
+          for (const TNode& n : term)
+          {
+            stack->push(n);
+          }
+        }
+    seen.insert(term);
```
barrettcw: Spacing seems to be off here.

src/preprocessing/passes/bv_ackermann.cpp
```diff
@@ -41,11 +42,57 @@ namespace passes {
...
+                     NodeManager* nm)
+{
+  Node args_eq;
+
+  if (args1.getKind() == kind::APPLY_UF)
+  {
+    AlwaysAssert(args1.getOperator() == func);
+    AlwaysAssert(args2.getKind() == kind::APPLY_UF
+                 && args2.getOperator() == func);
+    AlwaysAssert(args1.getNumChildren() == args2.getNumChildren());
```
aniemetz: Why are these AlwaysAssert?
Note that AlwaysAssert is always enabled (as thte name implies), i.e., not only for debug but also for production builds. An assertion should only be an AlwaysAssert, if it is a critical runtime error (abort conditions such as ID overflows, mem out, ...). It seems to me that these should be simple Asserts.

src/preprocessing/passes/bv_ackermann.cpp
```diff
@@ -41,11 +42,57 @@ namespace passes {
...
+      eqs[i] = nm->mkNode(kind::EQUAL, args1[i], args2[i]);
+    }
+    args_eq = bv::utils::mkAnd(eqs);
+  }
+  else
+  {
+    AlwaysAssert(args1.getKind() == kind::SELECT && args1[0] == func);
+    AlwaysAssert(args2.getKind() == kind::SELECT && args2[0] == func);
+    AlwaysAssert(args1.getNumChildren() == 2);
+    AlwaysAssert(args2.getNumChildren() == 2);
```
aniemetz: Ditto

src/preprocessing/passes/bv_ackermann.cpp
```diff
@@ -41,11 +42,57 @@ namespace passes {
...
+    args_eq = nm->mkNode(kind::EQUAL, args1[1], args2[1]);
+  }
+  Node func_eq = nm->mkNode(kind::EQUAL, args1, args2);
+  Node lemma = nm->mkNode(kind::IMPLIES, args_eq, func_eq);
+  assertionsToPreprocess->push_back(lemma);
+}
+
+void addChildrenToStack(TNode node, std::stack<TNode>* stack)
+{
+  for (TNode arg: node)
```
aniemetz: Use `for (const TNode& arg : node)`

src/preprocessing/passes/bv_ackermann.cpp
```diff
@@ -54,42 +93,82 @@ void storeFunction(
   NodeSet& set = fun_to_args[func];
```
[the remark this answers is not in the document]
aniemetz: `func_to_args` is a `FunctionToArgsMap`, which is defined as `std::unordered_map<Node, NodeSet, NodeHashFunction>`. However, all nodes are inserted as a mapping of TNode -> Set(TNode). It seems to me that this could rather be a map from TNode to TNodeSet. But if you change this, double check if all inserted nodes are indeed TNodes, else you'll run into trouble.

src/preprocessing/passes/bv_ackermann.cpp
```diff
@@ -54,42 +101,91 @@ void storeFunction(
...
     fun_to_skolem.addSubstitution(term, skolem);
+    set.insert(term);
+    /* Add the arguments of term (newest element in set) to the stack, so that
+     * collectFunctionsAndLemmas will process them as well. 
+     * This is only needed if the set has at least two elements 
+     * (otherwise, no lemma is generated).
+     * Therefore, we defer this for term in case it is the first element in the set*/
+    if (set.size() == 2) 
+    {
+      for (Node elem : set)
```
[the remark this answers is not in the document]
aniemetz: See my reply above. In any case, you should use a const ref here: `for (const Node& elem : set)`.

src/preprocessing/passes/bv_ackermann.cpp
```diff
@@ -80,7 +78,7 @@ void addLemmaForPair(TNode args1,
 
 void addChildrenToStack(TNode node, std::stack<TNode>* stack)
 {
-  for (TNode arg: node)
+  for (const TNode& arg: node)
```
barrettcw: Not sure why you would use a const TNode&: TNode's do not incur any overhead, but TNode& adds another layer of indirection.

src/preprocessing/passes/bv_ackermann.cpp
```diff
@@ -120,7 +118,7 @@ void storeFunctionAndAddLemmas(TNode func,
      * Therefore, we defer this for term in case it is the first element in the set*/
     if (set.size() == 2) 
     {
-      for (Node elem : set)
+      for (const TNode& elem : set)
```
barrettcw: ditto

src/preprocessing/passes/bv_ackermann.cpp
aniemetz: `TNode arg` and `const TNode&` should compile to the exact same code:
https://godbolt.org/g/97P4b9

Since this is best practice for Nodes, for consistency I'd rather have that pattern used wherever possible. Maybe we should add that to the guidelines?

barrettcw:
If you've confirmed this, then either one will be OK with me.

-Clark

src/preprocessing/passes/bv_ackermann.cpp
```diff
@@ -41,55 +40,150 @@ namespace passes {
...
+        {
+          storeFunctionAndAddLemmas(
+              term[0], term, fun_to_args, fun_to_skolem, assertions, nm, stack);
+        }
+        else
+        {
+          AlwaysAssert(
+              term.getKind() != kind::STORE,
+              "Cannot use eager bitblasting on QF_ABV formula with stores");
+          /* add childrens to the stack, so that they are processed later */
```
aniemetz: `childrens` --> `children`

aniemetz:
@barrettcw not sure if this is a question, but the link I pasted confirms this.

barrettcw:
OK - I see.  Sounds good.

-Clark

src/preprocessing/passes/bv_ackermann.cpp
```diff
@@ -41,55 +40,150 @@ namespace passes {
...
+    Assert(args1.getNumChildren() == 2);
+    Assert(args2.getNumChildren() == 2);
+    args_eq = nm->mkNode(kind::EQUAL, args1[1], args2[1]);
+  }
+  Node func_eq = nm->mkNode(kind::EQUAL, args1, args2);
+  Node lemma = nm->mkNode(kind::IMPLIES, args_eq, func_eq);
+  assertionsToPreprocess->push_back(lemma);
+}
+
+void addChildrenToStack(TNode node, std::stack<TNode>* stack)
```
mpreiner: I would prefer you use a `std::vector` instead of `std::stack`. `std::stack` is a container adapter that uses by default a `std::deque` as container, which is slower than `std::vector`. If you want to use a vector container you need to declare `std::stack<TNode, std::vector>` .

src/preprocessing/passes/bv_ackermann.cpp
```diff
@@ -41,55 +40,150 @@ namespace passes {
...
+    /* Add the arguments of term (newest element in set) to the stack, so that
+     * collectFunctionsAndLemmas will process them as well. 
+     * This is only needed if the set has at least two elements 
+     * (otherwise, no lemma is generated).
+     * Therefore, we defer this for term in case it is the first element in the set*/
+    if (set.size() == 2) 
+    {
+      for (const TNode& elem : set)
+      {
+        addChildrenToStack(elem, stack);
```
mpreiner: With std::vector as stack you can use a one liner: `stack.insert(stack.end(), elem.begin(), elem.end());`

src/preprocessing/passes/bv_ackermann.cpp
```diff
@@ -41,55 +40,150 @@ namespace passes {
...
+      }
+    } else if (set.size() > 2)
+    {
+      addChildrenToStack(term, stack);
```
mpreiner: Ditto.

src/preprocessing/passes/bv_ackermann.h
```diff
@@ -26,16 +26,17 @@
...
-
 #include <unordered_map>
 
 namespace CVC4 {
 namespace preprocessing {
 namespace passes {
 
-typedef std::unordered_map<Node, NodeSet, NodeHashFunction> FunctionToArgsMap;
+typedef std::unordered_set<TNode, TNodeHashFunction> TNodeSet;
+typedef std::unordered_map<TNode, TNodeSet, TNodeHashFunction> FunctionToArgsMap;
```
mpreiner: Prefer using `using` over `typedef`: `using TNodeSet = std::unordered_set<TNode, TNodeHashFunction>`

4tXJ7f:
Marking "do not merge" for clang-format.

Issue #2186, opened 2018-07-20
Support standard SL inputs


ajreynol:
An SL standard was created this year for SL comp. CVC4 does not yet support this standard.

PR #2257, opened 2018-08-02
Parse standard separation logic inputs

ajreynol:
Updates the previous regressions (in an ad-hoc format) to the new format.

This is work towards #2186 (we still don't support SL logic strings).


ajreynol:
Details on the format can be found here: https://www.irif.fr/~sighirea/sl-comp/18/

PR #2259, opened 2018-08-02
Add rewrites for BV ITE and BV COMP with const condition/child.

aniemetz:


martin-cs:
I might be misreading things but I think the BITVECTOR_ITE(C,X,X) ~~> X case is also potentially useful.

PR #2261, opened 2018-08-02
 Fixes and improvements for single invocation inference

ajreynol:
Corrects a few corner cases in the module that infers whether a property is single invocation. For example, this includes when we have 0-argument function to synthesize, and we can eliminate that function from a disjunct in a sygus conjecture, e.g.:
`exists f:Int. f != 0`

It also makes it so that we can infer single invocation for synthesis conjecture with multiple functions to synthesize that have distinct return types but identical argument types. For example:
`exists f: Int -> Int, g : Int -> Bool. forall x. f(x)=0 <=> g(x)`

These issues have come up in Horn clause benchmarks.


src/theory/quantifiers/single_inv_partition.cpp
```diff
@@ -187,7 +205,17 @@ bool SingleInvocationPartition::init(std::vector<Node>& funcs,
       std::vector<Node> si_subs;
       Trace("si-prt") << "Process conjunct : " << conj[i] << std::endl;
       // do DER on conjunct
-      Node cr = TermUtil::getQuantSimplify(conj[i]);
+      // Must avoid eliminating the first-order input functions. We use a
+      // substitution to avoid this.
```
HanielB: please say which simplification simplification would be eliminating that functions that you are avoiding by substituting with the skolems

Issue #2197, opened 2018-07-23
Handle indexed symbols in sygus


ajreynol:
The sygus variant of the smt2 parser does not accept indexed symbols in sygus grammars. We should add this.

ajreynol:
This is now subsumed by #2863.

PR #2266, opened 2018-08-02
Simplify and improve the sygus parser

ajreynol:
Currently, there is code duplication for parsing constants in sygus grammars, e.g.:
https://github.com/CVC4/CVC4/blob/master/src/parser/smt2/Smt2.g#L1048
https://github.com/CVC4/CVC4/blob/master/src/parser/smt2/Smt2.g#L2304

This PR removes this duplication by introducing a new ANTLR grammar "termAtomic" in Smt2.g.

As a result of this PR, we now parse sygus grammars with any constant we otherwise parse. This addresses known issues where CVC4 rejects grammars with floating point constants.

It also removes some unnecessary code for converting builtin kinds to their total versions.

This is work towards #2197.  We still do not support sygus grammars with non-trivial *composite* terms, such as applications of indexed function symbols.


aniemetz:
Labelled "do not merge' so that @ajreynol can fix the formatting.

PR #2267, opened 2018-08-03
Fix printing statistics in case of signals.

mpreiner:
Adds a newline after each statistic.

### DIFF
--- a/src/util/statistics.cpp
+++ b/src/util/statistics.cpp
@@ -126,6 +126,7 @@ void StatisticsBase::safeFlushInformation(int fd) const {
       safe_print(fd, s_regDelim);
     }
     s->safeFlushStat(fd);
+    safe_print(fd, "\n");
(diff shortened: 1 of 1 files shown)


src/util/statistics.cpp
```diff
@@ -126,6 +126,7 @@ void StatisticsBase::safeFlushInformation(int fd) const {
       safe_print(fd, s_regDelim);
     }
     s->safeFlushStat(fd);
+    safe_print(fd, "\n");
   }
   safe_print(fd, "\n");
```
4tXJ7f: Nit: we can probably remove this one now.

mpreiner: I kept it to match the newline behaviour of the non-safe printing.

mpreiner: I removed it now for both versions.

PR #2282, opened 2018-08-07
 Wait to do sygus qe preprocess until full effort check 

ajreynol:
Doing sygus qe preprocess involves invoking an external SMT engine. The commit made this happen during preregistration, which apparently breaks some invariant in NodeVisitor for not being "re-entrant" (https://github.com/CVC4/CVC4/blob/master/src/smt_util/node_visitor.h#L47).

This PR makes this step happen a bit later (when QuantifiersEngine is called to check at full effort), which avoid the assertion failure.  Notice that the code was simply reorganized in a minor way although the diff looks strange.

This fixes the nightlies.

### DIFF
--- a/src/theory/quantifiers/sygus/ce_guided_instantiation.h
+++ b/src/theory/quantifiers/sygus/ce_guided_instantiation.h
@@ -35,7 +35,8 @@ private:
   bool d_last_inst_si;
   /** the conjecture we are waiting to assign */
   Node d_waiting_conj;
-private:
+
+ private:
   /** assign quantified formula q as the conjecture
    *
    * This method returns true if q was successfully assigned as the synthesis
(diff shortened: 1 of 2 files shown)


src/theory/quantifiers/sygus/ce_guided_instantiation.h
```diff
@@ -33,7 +33,20 @@ class CegInstantiation : public QuantifiersModule
...
+  /** the conjecture we are waiting to assign */
+  Node d_waiting_conj;
+
+ private:
+  /** assign quantified formula q as the conjecture
+   *
+   * This method returns true if q was successfully assigned as the synthesis
+   * conjecture considered by this class. This method may return false, for
+   * instance, if this class determines that it would rather rewrite q to
+   * an equivalent form r (in which case this method returns the lemma
```
HanielB: close this parenthesis

PR #2283, opened 2018-08-08
Require Swig 3

4tXJ7f:
Removes some hacks due to Swig 2's incomplete C++11 support and adds
checks for version 3 at configuration time as well as in swig.h


mpreiner:
It seems that `base/tls.h` is included in several files, e.g., `src/main/main.h`

4tXJ7f:
Yup, I just realized after pushing :)

src/base/cvc4_assert.cpp
```diff
@@ -20,14 +20,13 @@
 
 #include "base/cvc4_assert.h"
 #include "base/output.h"
-#include "base/tls.h"
 
 using namespace std;
 
 namespace CVC4 {
 
 #ifdef CVC4_DEBUG
```
timothy-king: Remove the ifdef too

PR #2292, opened 2018-08-10
Add contrib/get-gmp script.

mpreiner:


contrib/get-gmp
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+
```
timothy-king: You might want to discourage most folks from using this unless they know that they need to. The version installed by their distribution  is probably the version that they want to use. All that is needed is a comment here.

mpreiner: Done. Is that enough to discourage people? :-)

contrib/get-gmp
```diff
@@ -0,0 +1,38 @@
...
+# distribution does not ship with static GMP libraries (e.g., Arch Linux) and
+# you want to build CVC4 statically.
+# In most of the cases the GMP version installed on your system is the one you
+# want and should use.
+#
+
+source "$(dirname "$0")/get-script-header.sh"
+
+if [ -z "${BUILD_TYPE}" ]; then
+  BUILD_TYPE="--disable-shared --enable-static"
```
4tXJ7f: Nit: why use a different pattern than for `$HOST` and `$GMPVERSION` to set the default here? :)

mpreiner: Because by default we want a static GMP library if we only call contrib/get-cmp?

PR #2293, opened 2018-08-10
Do not use static initialization in CxxTest runner

4tXJ7f:
The static initialization in the CxxTest runner was causing problems
when having `std::unique_ptr`s in test classes. When the ExprManager's
deconstructor is called, we count on certain static objects to be around
(e.g.
https://github.com/CVC4/CVC4/blob/0a02fd2b69c0c0f454fc33d8028b24f4fcf431de/src/expr/attribute_internals.h#L508).
If the ExprManager is (indirectly) owned by a `std::unique_ptr` in a
static class, however, there are no such guarantees as the destruction
order of static objects is not defined. This commit adds a flag for
CxxTest to not use static initialization in the test runner, which
solves the issue.  Additionally, the commit fixes a warning about a
missing virtual deconstructor in ParserBlack that came up after using
the new flags.

This fixes an issue reported in the nightly builds.

### DIFF
--- a/test/unit/Makefile.am
+++ b/test/unit/Makefile.am
@@ -150,7 +150,7 @@ $(UNIT_TESTS:%=@am__quote@./@DEPDIR@/%.Plo@am__quote@): %.Plo:
 
 $(UNIT_TESTS:%=@abs_builddir@/%.cpp): @abs_builddir@/%.cpp: %.h
 	$(AM_V_at)$(MKDIR_P) `dirname "$@"`
-	$(AM_V_GEN)$(CXXTESTGEN) --have-eh --have-std --error-printer -o "$@" "$<"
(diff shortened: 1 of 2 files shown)


aniemetz:
Note that this is also fixed if the solver object is explicitly destroyed in tearDown(). Previously, the ExprManager was deleted in tearDown(), maybe that's why we didn't have that issue?

aniemetz:
@4tXJ7f thanks for helping to find a fix for this nasty issue!

4tXJ7f:
Yes, that's exactly the case. Deleting the `ExprManager` in `tearDown()` avoided the issue because at that time all the static objects were still alive. Note that the generated code by CxxTest still contains a bunch of static variables, even with the flag but at least it fixes the ASAN issue until we have time to move away from CxxTest.

PR #2297, opened 2018-08-12
Removing support for T* and const T* attributes.

timothy-king:
These are unused.


timothy-king:
It may be possible to deprecate and remove cleanup strategies from Attributes once this is in.

test/unit/expr/attribute_black.h
```diff
@@ -214,6 +134,4 @@ class AttributeBlack : public CxxTest::TestSuite {
     delete node;
   }
 
-};
-
-int AttributeBlack::MyData::count = 0;
+};
```
barrettcw: nit: add newline

timothy-king: Done

PR #2304, opened 2018-08-14
Consolidating the Node::substitute() functions into one implementation.

timothy-king:
This PR consolidates the implementations of Node::substitute() down to one workhorse based around std::unordered_map.

I have gone with the approach of:
* keeping old behaviors and contracts (orders are kept when relevant, keep using TNode even if dangerous, keep substitute_internal public as their are existing users)
* using c++11 features and map_util.h to simplify the code.

I will need help with doing a before/after performance comparison.

A follow-up PR will move the Node::substitute() implementations into its own class. These do not need to be on Node/TNode.

### DIFF
--- a/src/expr/node.h
+++ b/src/expr/node.h
@@ -254,7 +254,7 @@ class NodeTemplate {
    * return values and the NodeBuilders on the stack [potentially
    * transitively].)
    */
-  Node substitute_internal(
+  Node substituteInternal(
       std::unordered_map<TNode, TNode, TNodeHashFunction>& cache) const;
 
  public:
@@ -1351,11 +1351,11 @@ NodeTemplate<ref_count>::substitute(TNode node, TNode replacement) const {
   }
   std::unordered_map<TNode, TNode, TNodeHashFunction> cache{
       {node, replacement}};
-  return substitute_internal(cache);
+  return substituteInternal(cache);
 }
 
 template <bool ref_count>
-Node NodeTemplate<ref_count>::substitute_internal(
+Node NodeTemplate<ref_count>::substituteInternal(
     std::unordered_map<TNode, TNode, TNodeHashFunction>& cache) const
 {
   if (TNode* in_cache = FindOrNull(cache, *this))
@@ -1375,7 +1375,7 @@ Node NodeTemplate<ref_count>::substitute_internal(
     }
     else
     {
-      nb << getOperator().substitute_internal(cache);
+      nb << getOperator().substituteInternal(cache);
     }
   }
   for (const_iterator i = begin(), iend = end(); i != iend; ++i)
@@ -1386,11 +1386,13 @@ Node NodeTemplate<ref_count>::substitute_internal(
     }
     else
     {
-      nb << (*i).substitute_internal(cache);
+      nb << (*i).substituteInternal(cache);
     }
   }
   Node n = nb;
-  AlwaysAssert(cache.insert(std::make_pair(*this, n)).second);
+  CVC4_UNUSED const bool inserted =
+      cache.insert(std::make_pair(*this, n)).second;
+  Assert(inserted);
   return n;
 }
 
@@ -1409,7 +1411,7 @@ NodeTemplate<ref_count>::substitute(Iterator1 nodesBegin,
     cache.insert(std::make_pair(*nodesBegin, *replacementsBegin));
   }
   Assert(nodesBegin == nodesEnd && replacementsBegin == replacementsEnd);
-  return substitute_internal(cache);
+  return substituteInternal(cache);
 }
 
 template <bool ref_count>
@@ -1424,7 +1426,7 @@ NodeTemplate<ref_count>::substitute(Iterator substitutionsBegin,
     cache.insert(std::make_pair((*substitutionsBegin).first,
                                 (*substitutionsBegin).second));
   }
-  return substitute_internal(cache);
+  return substituteInternal(cache);
 }
 
 template <bool ref_count>
--- a/src/expr/node.h
+++ b/src/expr/node.h
@@ -36,12 +36,13 @@
 #include "base/configuration.h"
 #include "base/cvc4_assert.h"
 #include "base/exception.h"
+#include "base/map_util.h"
 #include "base/output.h"
-#include "expr/type.h"
-#include "expr/kind.h"
-#include "expr/metakind.h"
 #include "expr/expr.h"
 #include "expr/expr_iomanip.h"
+#include "expr/kind.h"
+#include "expr/metakind.h"
+#include "expr/type.h"
 #include "options/language.h"
 #include "options/set_language.h"
 #include "util/hash.h"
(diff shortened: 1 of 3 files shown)


src/expr/node.h
```diff
@@ -1354,51 +1349,48 @@ NodeTemplate<ref_count>::substitute(TNode node, TNode replacement) const {
...
+    {
+      nb << (*i).substitute_internal(cache);
     }
   }
-
-  // put in cache
   Node n = nb;
-  Assert(node != n);
-  cache[*this] = n;
+  AlwaysAssert(cache.insert(std::make_pair(*this, n)).second);
```
ajreynol: We shouldn't have any AlwaysAssert in this function (the performance of substitute is a bottleneck for some applications like quantifiers).  I'd recommend making this Assert.

timothy-king: The side effect needs to always happen so 
```
Assert(cache.insert(std::make_pair(*this, n)).second);
```
does not make sense. Would you prefer something like:
```
CVC4_UNUSED const bool inserted = cache.insert(std::make_pair(*this, n)).second;
Assert(inserted);
```

I do not understand the motivation of the comment though. Is the claim that the untaken branch for the assertion has a measurable performance impact? Is it the exceptions? (This can be switched to a CVC4_CHECK. Would that be acceptable?) I am somewhat doubtful this can be distinguished from noise in practice. (Maybe it can on microbenchmarks?)

ajreynol: Yes, the original code is a bit confusing, I misinterpreted the fact that what is inside the AlwaysAssert is what is doing the work.

Typically, I expect that what is inside an Assert(...) or AlwaysAssert(...) is side effect free (not sure if there is a strict policy on this).  

The replaced code you give above is what I would expect.

mpreiner: I would also prefer keeping Assert(...) and AlwaysAssert(...) side effect free.

ajreynol:
Marking do not merge to wait for cluster runs.

mpreiner:
Ok here are the results from the cluster. I ran both version on all quantified benchmarks since the quantifiers module uses substitute() a lot.

Note that the new version has two errors with `(error "std::bad_alloc")`. In one of those cases the base version also runs into a memory out. In two other cases `new` is significantly faster than `base`.

Here is the difference on the benchmarks:

```
                                                                                                     DIRECTORY |  2018-08-08-node-substitute-base |   2018-08-08-node-substitute-new |                                                                                                                                        
                                                                                                     BENCHMARK |  STAT  RES EXIT CPU[s] SPACE[MB] |  STAT  RES EXIT CPU[s] SPACE[MB] |                                                                                                                                        
                                                             AUFLIA/20170829-Rodin-smt1630225104785523677.smt2 |    to    -    2  120.1     186.3 |    ok   20    0  113.2     194.1 |                                                                                                                                        
                                                             AUFLIA/20170829-Rodin-smt4311391195430732252.smt2 |    ok   30    0  115.3      61.2 |    to    -    2 120.09      61.3 |                                                                                                                                        
                                                   AUFNIRA/nasa-vc_normalize_subst-thruster_symm_0001.fof.smt2 |    to    -    2 120.01     104.5 |    ok   20    0   9.29      51.7 |                                                                                                                                        
                                                                      LRA/scholl-smt08-RNDPRE-RNDPRE_4_60.smt2 |    ok   10    0  110.1      96.9 |    to    -    2  120.1      83.4 |                                                                                                                                        
       UF/20170428-Barrett-cdt-cade2015-nada-afp-coinductive_list-x2015_09_10_16_56_00_260_1430662.smt_in.smt2 |    ok   30    0  110.2     126.9 |    to    -    2  120.1     127.1 |                                                                                                                                        
 UF/20170428-Barrett-cdt-cade2015-nada-afp-sorted_list_operations-x2015_09_10_16_47_21_859_1047027.smt_in.smt2 |    ok   20    0 110.11     136.3 |    to    -    2 120.01     131.7 |                                                                                                                                        
                                                    UF/sledgehammer-Fundamental_Theorem_Algebra-uf.815756.smt2 |    to    -    2 120.18    3028.7 |    ee    -    0  76.15    2114.0 |                                                                                                                                        
                                                                      UF/sledgehammer-NS_Shared-uf.866499.smt2 |    ok   20    0  115.7      69.7 |    to    -    2 120.02      69.6 |                                                                                                                                        
                UFDT/cdt-cade2015-data-afp-sorted_list_operations-x2015_09_10_16_47_49_549_1075143.smt_in.smt2 |    ok   20    0  119.8     151.1 |    to    -    2  120.1     151.1 |                                                                                                                                        
                                   UFDT/rec-fun-ijcar2016-wo-wdf-isa_nun_bench_v4_enc-nunchaku2513265.nun.smt2 |    mo    -    3 119.61    8211.9 |    to    -    2  120.3    7571.7 |                                                                                                                                        
                                                                        UFDTLIA/hipspec-nichomachus-goal6.smt2 |    mo    -    3  115.0    8214.3 |    ee    -    0  69.89    3336.0 |                                                                                                                                        
                                    UFLIA/grasshopper-uninstantiated-delete_postcondition_of_delete_54_12.smt2 |    ok   20    0  91.37      92.9 |    to    -    2  120.1     119.1 |                                                                                                                                        
                                                                   UFLIA/sledgehammer-Hoare-smtlib.794454.smt2 |    to    -    2  120.1      66.0 |    ok   20    0   2.29      42.6 |                                                                                                                                        
                                                  UFNIA/lahiri-cav09-storm-queries-mqueue_example_2_2_2_1.smt2 |    ok   20    0 115.59     422.1 |    to    -    2 120.09     422.1 |                                                                                                                                        
                                                  UFNIA/lahiri-cav09-storm-queries-mqueue_example_2_3_4_2.smt2 |    ok   20    0  119.1     447.4 |    to    -    2 120.01     445.3 |                                                                                                                                        
                                             UFNIA/lahiri-cav09-storm-queries-serial_read_example_2_1_8_2.smt2 |    ok   20    0  115.8    7157.7 |    to    -    2  120.3    7158.0 |                                                                                                                                        
                                                UFNIA/lahiri-cav09-storm-queries-serial_write_example_1_2.smt2 |    to    -    2 120.29    7146.1 |    ok   20    0 113.59    7139.7 |                                                                                                                                        
                                             UFNIA/lahiri-cav09-storm-queries-usbsamp_bug_example_2_2_4_3.smt2 |    to    -    2 120.31    7728.2 |    ok   20    0  115.0    7728.2 |

```

ajreynol:
I expect the major differences in run time in new for those 2 benchmarks is because somehow a different search was triggered.

So you are showing the only benchmarks where there was a different result?  This is over 50k benchmarks or so?

Another thing I'm curious about is whether there is any consistent gain/loss in performance on benchmarks both configs solve. Maybe a scatter plot?  But perhaps just the total time on the common set of benchmarks they both solve is enough.

ajreynol:
Actually, can you just send me a csv of the runtimes? :)

ajreynol:
Unfortunately it looks like quantifiers gets 3.4% slower on average with the new code.  There are 350 benchmarks in the csv from Mathias where we are >1 second slower, and 121 where we are >1 second faster.

ajreynol:
Actually looking back at the code, it seems that it makes sense that this code will introduce a performance hit.  The common use case of substitute is using std::vector< Node >::iterator's.  Thus reconstructing the std::unordered_map is probably not a good idea (plus std::unordered_map in my experience tend to be fairly slow).

ajreynol:
An alternative would be to reduce the std::unordered_map implementation to the Iterator's version.

timothy-king:
Interesting. Fixing this up might require more cycles than I have at the moment given the current constraints. It might just make more sense to move this to node_algorithm.h/cpp as is for now. I might take another crack at simplifying this later. Closing this PR.

PR #2311, opened 2018-08-15
Move node algorithms to separate file

4tXJ7f:
This commit moves node algorithms to separate files
(`node_algorithm.{h,cpp}`). The node header is one of the files that
gets included in most places in the CVC4 code base and it is fairly big,
which negatively affects our compile time.  The goal of this commit is
to build a foundation for separating the representation of a node from
common algorithms performed on it. The algorithms file only needs to be
included by the files that use those algorithms (generally source files)
which is a significantly smaller number than the number of files that
include `node.h` (directly or indirectly).

Note: this commit currently does not move the `Node::substitute()`
method because of #2304.

@taking This might be of interest to you.


src/expr/node_algorithm.h
```diff
@@ -0,0 +1,130 @@
...
+
+ private:
+  TNode d_node;
+  Kind d_kind;
+};
+
+/**
+ * Check if the node has a subterm t.
+ */
+bool hasSubterm(TNode n, TNode t, bool strict = false);
```
timothy-king: Document `n` and `strict` for the function.

src/expr/node_algorithm.h
```diff
@@ -0,0 +1,130 @@
...
+namespace CVC4 {
+namespace expr {
+
+/**
+ * A class that pretends iterating over a node of a certain kind. If the node
+ * matches the kind, the class iterates over the children, otherwise the
+ * iterator just points to the node. This is useful if you want to pretend to
+ * iterate over a "unary" PLUS, for instance, since unary PLUSes don't exist.
+ */
+class KindedIterator
```
timothy-king: I suggest keeping KindedIterator unchanged & unmoved in this PR.

Instead go forward with the two algorithms, which have plenty of usages, and send a different PR to remove KindedIterator entirely. It appears to have exactly 1 usage. Just simulate what this class is providing by doing this there. Unless you are planning on adding a lot of future usages, in which case add it to its own file.

Another option for this PR is to move this into src/prop/prop_engine.cpp.

mpreiner: I suggest to remove the kinded_iterator and change the loop in `src/prop/prop_engine.cpp`. If the iterator was not really used until now, I doubt there will be more future usages.

4tXJ7f: I've actually come across other use cases. Here's [one example](https://github.com/CVC4/CVC4/blob/master/src/theory/strings/theory_strings_rewriter.cpp#L2072-L2097). I think the main reason for not using it was that people weren't aware of its existence (at least that's what the reason was for me when I wrote this code). I do agree that we can exclude it from this PR, however.

4tXJ7f: We could put it in separate files (e.g. `node_iterator.{h,cpp}`) where we collect helpers for iterating over nodes (e.g. an iterator that traverses in depth-first order and an iterator that flattens while traversing).

4tXJ7f: I have removed the changes to the kinded iterator from the changeset.

4tXJ7f:
@aniemetz I've slightly expanded the formatting to cover small blocks around the changes. PTAL.

PR #2313, opened 2018-08-15
Remove unused tuple classes

4tXJ7f:
Since we are using C++11, we can replace the triple and quad classes
with std::tuple.


aniemetz:
Labelled "do not merge" so that @4tXJ7f gets a chance to incorporate review comments.

4tXJ7f:
I've addressed the comments.

PR #2314, opened 2018-08-15
Add constexpr annotations to help coverity understand constant ...

timothy-king:
variable guards.


src/theory/quantifiers/sygus/ce_guided_single_inv_sol.cpp
```diff
@@ -702,7 +702,7 @@ Node CegConjectureSingleInvSol::reconstructSolution(Node sol,
             // if it is not already reconstructed
             if( d_reconstruct.find( itt->second )==d_reconstruct.end() ){
               Trace("csi-rcons") << "...reconstructed " << ns << " for term " << nr << std::endl;
-              bool do_check = true;//getPathToRoot( itt->second );
+              constexpr bool do_check = true;//getPathToRoot( itt->second );
```
ajreynol: I would be fine if you wanted to eliminate this variable and simplify the code accordingly.

timothy-king: Done.

src/theory/bv/theory_bv.cpp
```diff
@@ -861,9 +861,10 @@ bool TheoryBV::storePropagation(TNode literal, SubTheory subtheory)
   }
 
   // Propagate differs depending on the subtheory
-  // * bitblaster needs to be left alone until it's done, otherwise it doesn't know how to explain
+  // * bitblaster needs to be left alone until it's done, otherwise it doesn't
+  //   know how to explain
   // * equality engine can propagate eagerly
-  bool ok = true;
+  constexpr bool ok = true;
```
aniemetz: Something doesn't quite make sense here. `ok` is always true, right? So the branch on line 870 is never taken. Shouldn't line 869 be `ok = d_out->propagate(literal)`?

aniemetz: You can only use `constexpr` for `ok` if this is always true, right? If I address this issue in a future PR I will have to drop `constexpr` again? Maybe we should fix the underlying issue first. Is this PR urgent?

timothy-king: > You can only use constexpr for ok if this is always true, right?

Correct.

> If I address this issue in a future PR I will have to drop constexpr again?

This seems like a trivial cost, and if I add a comment it falls to a few seconds to resolve by deleting it and the comment.

> Maybe we should fix the underlying issue first.

I suggest going forward as it is an improvement over the current state.

> Is this PR urgent?

No.

src/theory/quantifiers/sygus/ce_guided_single_inv_sol.cpp
```diff
@@ -704,19 +704,18 @@ Node CegConjectureSingleInvSol::reconstructSolution(Node sol,
             // if it is not already reconstructed
             if( d_reconstruct.find( itt->second )==d_reconstruct.end() ){
```
aniemetz: Can you clang-format this whole if-block?

src/theory/bv/theory_bv.cpp
aniemetz: I just wanted to make sure that this is what we have to do when we address this problem.
Yes, add a comment. Maybe open an issue and add the issue number. You can assign me.

PR #2317, opened 2018-08-15
global-negate preprocessing pass

yoni206:
This PR refactors the global-negate preprocessing pass to the new format.


src/preprocessing/passes/global_negate.cpp
```diff
@@ -12,21 +12,23 @@
...
 using namespace CVC4::kind;
+using namespace CVC4::theory;
+namespace {
 
-namespace CVC4 {
-namespace theory {
-namespace quantifiers {
-
-void GlobalNegate::simplify(std::vector<Node>& assertions)
+Node simplify(std::vector<Node>& assertions, NodeManager* nm)
```
ajreynol: Can you make this a private member function of GlobalNegate?

src/preprocessing/passes/global_negate.cpp
```diff
@@ -12,21 +12,23 @@
...
+#include "expr/node.h"
 #include "theory/rewriter.h"
 
+namespace CVC4 {
+namespace preprocessing {
+namespace passes {
+
 using namespace std;
 using namespace CVC4::kind;
+using namespace CVC4::theory;
```
ajreynol: Move these above `namespace CVC4 {`

src/preprocessing/passes/global_negate.cpp
```diff
@@ -12,21 +12,23 @@
...
+namespace {
```
ajreynol: not sure what this `namespace {` is for?

src/preprocessing/passes/global_negate.cpp
```diff
@@ -90,21 +92,30 @@ void GlobalNegate::simplify(std::vector<Node>& assertions)
...
+{
+  if (assertionsToPreprocess->size() > 0)
   {
-    if (i == 0)
-    {
-      assertions[i] = body;
-    }
-    else
+    NodeManager* nm = NodeManager::currentNM();
+    assertionsToPreprocess->replace(
```
ajreynol: I'm not sure that we should use `replace` here, or follow the original behavior of replacing assertions by `true`.  I actually think its more appropriate to call:
```
Node ret = simplify(assertionsToPreprocess->ref(), nm);
assertionsToPreprocess->clear();
assertionsToPreprocess->push_back(ret);
```

Perhaps @4tXJ7f can comment?

4tXJ7f: As far as I know, we have to keep the number of assertions the same or larger, otherwise a bunch of other things will break (I don't remember what concretely but one issue is for example that we'd have to update `d_realAssertionsEnd`). In the future, it would be better to be able to actually remove assertions, yes.

ajreynol:
Ok, then we will I suppose say that the first asssertion depends on every
other assertion.

ajreynol:
Although notice that unsat cores should definitely be disabled when this
pass is run.

yoni206:
I think it is:
https://github.com/CVC4/CVC4/blob/ee4004505fa7f086872880d2d693c0608af29050/src/smt/smt_engine.cpp#L1494

PR #2321, opened 2018-08-15
Initializing bool members in command.cpp.

timothy-king:
I initialized all of these to false as that is what I guessed the intent was. That decision needs reviews in 3 cases.

### DIFF
--- a/src/smt/command.cpp
+++ b/src/smt/command.cpp
@@ -385,7 +385,7 @@ std::string PopCommand::getCommandName() const { return "pop"; }
 /* class CheckSatCommand                                                      */
 /* -------------------------------------------------------------------------- */
 
-CheckSatCommand::CheckSatCommand() : d_expr() {}
+CheckSatCommand::CheckSatCommand() : d_expr(), d_inUnsatCore() {}
 CheckSatCommand::CheckSatCommand(const Expr& expr, bool inUnsatCore)
     : d_expr(expr), d_inUnsatCore(inUnsatCore)
 {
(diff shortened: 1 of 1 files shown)


src/smt/command.cpp
```diff
@@ -1778,7 +1779,10 @@ std::string GetSynthSolutionCommand::getCommandName() const
 /* class GetQuantifierEliminationCommand                                      */
 /* -------------------------------------------------------------------------- */
 
-GetQuantifierEliminationCommand::GetQuantifierEliminationCommand() : d_expr() {}
+GetQuantifierEliminationCommand::GetQuantifierEliminationCommand()
+    : d_expr(), d_doFull()
```
ajreynol: d_doFull should default to true (sorry I missed this).

src/smt/command.cpp
```diff
@@ -385,7 +385,7 @@ std::string PopCommand::getCommandName() const { return "pop"; }
 /* class CheckSatCommand                                                      */
 /* -------------------------------------------------------------------------- */
 
-CheckSatCommand::CheckSatCommand() : d_expr() {}
+CheckSatCommand::CheckSatCommand() : d_expr(), d_inUnsatCore() {}
```
aniemetz: Seems like d_inUnsatCore is not used in CheckSatCommand (and in CheckSatAssumingCommand). I can submit a PR to remove it.

timothy-king: I went ahead and removed these.

PR #2323, opened 2018-08-16
 Eliminate partial operators in sygus grammar normalization

ajreynol:
This corrects a bug that was introduced in #2266 (the hack removed there was necessary).

This ensures that we handle operators like bvudiv, bvsdiv, bvurem, div, rem, / properly in sygus.

This also enables total semantics for BV div-by-zero for sygus.


src/smt/smt_engine.cpp
```diff
@@ -1294,8 +1294,12 @@ void SmtEngine::setDefaults() {
   // Language-based defaults
   if (!options::bitvectorDivByZeroConst.wasSetByUser())
   {
+    // Bitvector-divide-by-zero changed semantics in SMT LIB 2.6, thus we
+    // set this option if the input format in SMT LIB 2.6. We also set this
```
HanielB: in SMT --> is SMT?

PR #2326, opened 2018-08-16
Minor improvements to the interface for sygus sampler

ajreynol:
This is 1/3 PRs for merging the sygus query generator branch https://github.com/ajreynol/CVC4/tree/sygusQueryGen.


src/theory/quantifiers/sygus/cegis.cpp
```diff
@@ -527,8 +527,9 @@ bool Cegis::sampleAddRefinementLemma(const std::vector<Node>& candidates,
         // mark this as a CEGIS point (no longer sampled)
         d_cegis_sample_refine.insert(i);
         std::vector<Node> vars;
+        d_cegis_sampler.getVariables(vars);
```
4tXJ7f: It looks like `vars` is not used. Can we remove it?

src/theory/datatypes/datatypes_sygus.cpp
```diff
@@ -895,8 +895,9 @@ Node SygusSymBreakNew::registerSearchValue(
               Trace("sygus-rr-debug")
                   << "; unsound: are not equivalent for : " << std::endl;
               std::vector<Node> vars;
+              its->second.getVariables(vars);
```
4tXJ7f: It looks like `vars` is only used for the assertion check. Move it into an `if (Configuration::isAssertionBuild())`?

ajreynol: It's also used in the printing within the for loop: https://github.com/CVC4/CVC4/pull/2326/files#diff-b5f4e26d011c1b54d669904177e990efL903.

src/theory/quantifiers/sygus/cegis.cpp
ajreynol: Yes good catch.

ajreynol:
Anything more for this one?

src/theory/datatypes/datatypes_sygus.cpp
4tXJ7f: Ah yes, but then why not just move that function call into an if statement guarded by `Configuration::isAssertionBuild() || Trace.on("sygus-rr-debug")`? Like that we don't incur the overhead of copying the variables when we are running in production without tracing.

ajreynol: Actually, this output should be part of the exception on line 921 (I'm surprised its not already). I'll refactor.

Issue #1178, opened 2017-10-01
Implement SyGuS grammars for missing types


ajreynol:
We currently do not support default grammars for types like Real and BitVector.

We should extend:
https://github.com/CVC4/CVC4/blob/master/src/theory/quantifiers/term_database_sygus.cpp#L2743
to do this.

This is required for Lustre invariant synthesis benchmarks for Kind2.

ajreynol:
I think we can close this for now. If a new specific type becomes required, let's open a new issue.

ajreynol:
We should add support for Array.

ajreynol:
Arrays were supported in #2685, strings were supported in #3148.

PR #2328, opened 2018-08-17
Adding support for bitvector SyGuS problems without grammars

HanielB:
This partially addresses #1178 by generating grammars for bitvectors


HanielB:
I asked @mpreiner about which operators we can exclude, so `sdiv srem ashr sub` could all be removed. But I'm not sure if we should be more conservative and just keep them.

PR #2334, opened 2018-08-17
 Make sygus inference a preprocessing pass

ajreynol:


src/preprocessing/passes/sygus_inference.h
```diff
@@ -33,25 +36,36 @@ namespace quantifiers {
...
-   * interpretation in the solution found by a separate call to an SMT engine
-   * and returns true, or leaves the assertions unmodified and returns false.
+   * interpretation in a sygus solution, or leaves the assertions unmodified.
+   */
+  PreprocessingPassResult applyInternal(
+      AssertionPipeline* assertionsToPreprocess) override;
+  /** solve sygus
+   *
+   * Returns true if we can recast the input problem assertions as a sygus
+   * problem as successfully solve it using a separate call to an SMT engine.
```
HanielB: as --> and?

src/preprocessing/passes/sygus_inference.cpp
```diff
@@ -259,44 +307,19 @@ bool SygusInference::simplify(std::vector<Node>& assertions)
...
-                                         final_ff_sol.begin(),
-                                         final_ff_sol.end());
-    if (curr != prev)
+    else
     {
-      curr = Rewriter::rewrite(curr);
-      Trace("sygus-infer-debug")
-          << "...rewrote " << prev << " to " << curr << std::endl;
-      assertions[i] = curr;
+      Assert(false);
```
HanielB: You don't need this or the `else` since you already have an assertion for getting into the `if`

Issue #2305, opened 2018-08-14
Hard Kind2-BMC problems for CVC4


daniel-larraz:
Hello,

Attached you can find a set of [SMT scripts](https://github.com/CVC4/CVC4/files/2288492/BMC.zip) to be solved in incremental mode that Z3 and Yices 2 can solve in seconds, but CVC4 can not solve in minutes.

Could you please take a look at them to check if there is something you can do?

Thank you,
Daniel

ajreynol:
Thanks for the benchmarks.

For internal discussion:

As an initial step, I've profiled and found that arithmetic's propagate method is consistently 40% of the run time on these benchmarks.  Using the option `--new-prop`, this decreases to 8%. We should aim to make this option default.

daniel-larraz:
If you take a look at the logic of the SMT problems that I uploaded, you can see it does not include any restrictions at all (it is the predefined symbol ALL). Although this is not a big deal for Z3 and Yices 2, it makes a big difference for CVC4. In fact, I found that providing a specific logic for these problems is enough for solving all the original Lustre models with a time limit of 2 minutes per problem. However, even doing that, the total time that it takes to solve all the problems using CVC4 is much longer than the total time that it takes when using Z3 or Yices 2.

So after you improved the code, I performed the following tests:

First, I ran Kind 2 with CVC4 again over the original 41 Lustre models using your suggested flags (--decision=internal --ext-rew-prep --ext-rew-prep-agg) and the "ALL" logic. Kind 2 could find a counterexample for 31 of the 41 models using a time limit of 2 minutes per problem. That is a considerable improvement with respect to the original situation, but there were still 10 problems that could not be solved with this setting.

Then I repeated the experiment, but providing a specific logic this time. In this case, Kind 2 could solve the 41 problems in a time comparable to the one that it takes when using Z3. I extended the experiment to the whole set of benchmarks, and I got similar results (same number of solved problems in comparable amount of time). In other words, the combination of using a specific logic and the suggested command-line options made CVC4 competitive with respect to Z3.

Since providing a specific logic is not a problem in our context, I do not consider the current performance of CVC4 a major issue anymore. However, notice there is room for improvement. First, by reducing the time differences of solving the problems when the logic is provided or not. But secondly, by improving the solving times even when the logic is provided. Although CVC4 times are comparable to Z3 times, Yices 2 outperforms Z3 and CVC4 for the SMT problems generated by Kind 2.

[...]

PR #2335, opened 2018-08-17
Makes the new row propagation system default

HanielB:
By running on all SMT-LIB benchmarks which involve any kind of arithmetic logic (a total of 97285 benchmarks) we have the following results, with 300s timeout:

**new-prop** vs **default** is +2386 / - 2298, solving in total 68870 benchmarks against 68782.

On the attached plots we can see clear trends favoring new-prop:

[cactus_comp-default_vs_new-prop.csv.pdf](https://github.com/CVC4/CVC4/files/2298204/cactus_comp-default_vs_new-prop.csv.pdf)
[scatter_comp-default_vs_new-prop.csv.pdf](https://github.com/CVC4/CVC4/files/2298205/scatter_comp-default_vs_new-prop.csv.pdf)

A breakdown per logic of gains/losses:

```
| Logic   | Gain | Loss | Net |
|---------+------+------+-----|
| LRA     |   15 |   15 |   0 |
| UFLIA   |    5 |    4 |   1 |
| UFNIA   |    8 |    7 |   1 |
| QF_IDL  |   71 |   25 |  46 |
| QF_LIA  |   63 |   64 |  -1 |
| QF_LRA  |   74 |    2 |  72 |
| QF_NIA  | 1937 | 1973 | -36 |
| QF_NRA  |  206 |  203 |   3 |
| QF_ALIA |    0 |    1 |  -1 |
| UFNRA   |    5 |    1 |   4 |
| AUFLIA  |    1 |    3 |  -2 |
| AUFNIA  |    1 |    0 |   1 |
|---------+------+------+-----|
| Total   | 2386 | 2298 |  88 |
|---------+------+------+-----|
```

The biggest variations are on QF_NRA and QF_NIA, but that seems to be expected given the heuristic nature of our solving methods in these logics.

The data is available here: 
[data.zip](https://github.com/CVC4/CVC4/files/2298722/data.zip)

Using this option by default is work towards addressing #2305


ajreynol:
@timothy-king , you'll be interested in this.

ajreynol:
I think the PR is still failing regressions

HanielB:
Fixed it. It was another nl benchmark

HanielB:
I did a debug run on the same benchmarks also with 300s timeout and there were two assertion failures, but these are about a known issue already fixed in PR #2099

timothy-king:
@ajreynol My only real contribution to make is that similar tests should be done on incremental benchmarks too.

HanielB:
I'll evaluate on incremental benchmarks as well, thanks @timothy-king

HanielB:
Of the 6122 incremental benchmarks of SMT-LIB that have any kind of arithmetic logic both configurations solve the same 6076 benchmarks. There are minimal solving time differences, and on average **new-prop** is 0.1% slower.  A debug run had no errors.

timothy-king:
No errors on debug runs and performance difference is noise. This seems fine.

PR #2340, opened 2018-08-20
Generalize CandidateRewriteDatabase to ExprMiner

ajreynol:
This is in preparation for mining possibly multiple kinds of information simulataneously (queries and rewrites) from the sygus enumeration.

This makes a virtual base class ExprMiner, which CandidateRewriteDatabase inherits from. It also introduces a ExprMinerManager, which will be used to manage the interaction between multiple expression miners.

This is 2/3 PRs for merging the sygus query generator branch https://github.com/ajreynol/CVC4/tree/sygusQueryGen.


src/theory/quantifiers/candidate_rewrite_database.cpp
```diff
@@ -36,48 +36,44 @@ CandidateRewriteDatabase::CandidateRewriteDatabase()
...
 
-void CandidateRewriteDatabase::initializeSygus(QuantifiersEngine* qe,
+void CandidateRewriteDatabase::initializeSygus(const std::vector<Node>& vars,
+                                               QuantifiersEngine* qe,
                                                Node f,
-                                               unsigned nsamples,
-                                               bool useSygusType)
+                                               SygusSampler* ss)
 {
+  Assert( ss!=nullptr );
```
4tXJ7f: Apply clang-format.

src/theory/quantifiers/expr_miner_manager.h
```diff
@@ -0,0 +1,103 @@
...
+ ** \verbatim
+ ** Top contributors (to current version):
+ **   Andrew Reynolds
+ ** This file is part of the CVC4 project.
+ ** Copyright (c) 2009-2018 by the authors listed in the file AUTHORS
+ ** in the top-level source directory) and their institutional affiliations.
+ ** All rights reserved.  See the file COPYING in the top-level source
+ ** directory for licensing information.\endverbatim
+ **
+ ** \brief expr_miner_manager
```
4tXJ7f: Add short comment.

src/theory/quantifiers/expr_miner_manager.h
```diff
@@ -0,0 +1,103 @@
...
+
+#include "cvc4_private.h"
+
+#ifndef __CVC4__THEORY__QUANTIFIERS__EXPR_MINER_MANAGER_H
+#define __CVC4__THEORY__QUANTIFIERS__EXPR_MINER_MANAGER_H
+
+#include "theory/quantifiers/candidate_rewrite_database.h"
+#include "theory/quantifiers/extended_rewrite.h"
+#include "theory/quantifiers/sygus_sampler.h"
+#include "theory/quantifiers_engine.h"
```
4tXJ7f: Include `expr/node.h`.

src/theory/quantifiers/expr_miner_manager.h
```diff
@@ -0,0 +1,103 @@
...
+  void initializeSygus(QuantifiersEngine* qe,
+                       Node f,
+                       unsigned nsamples,
+                       bool useSygusType);
+  /** enable rewrite rule synthesis (--sygus-rr-synth) */
+  void enableRewriteRuleSynth();
+  /** add term
+   *
+   * Expression miners may print information on the output stream out, for
+   * instance, candidate-rewrites.
```
4tXJ7f: Describe return value.

src/theory/quantifiers/expr_miner_manager.h
```diff
@@ -0,0 +1,103 @@
...
+  bool addTerm(Node sol, std::ostream& out);
+  /**
+   * Same as above, but the argument rew_print is set to true if a rewrite rule
+   * was printed on the output stream out.
+   */
+  bool addTerm(Node sol, std::ostream& out, bool& rew_print);
+
+ private:
+  /** whether we are doing rewrite synthesis */
+  bool d_do_rew_synth;
```
4tXJ7f: [Change to camelCase for member variables.](https://github.com/CVC4/CVC4/wiki/Code-Style-Guidelines#members)

src/theory/quantifiers/expr_miner_manager.cpp
```diff
@@ -0,0 +1,101 @@
...
+ **
+ ** \brief Implementation of expr_miner_manager
+ **/
+
+#include "theory/quantifiers/expr_miner_manager.h"
+
+#include "theory/quantifiers_engine.h"
+
+using namespace std;
+using namespace CVC4::kind;
```
4tXJ7f: It doesn't seem like those `using` statements are used.

src/theory/quantifiers/expr_miner.h
```diff
@@ -0,0 +1,96 @@
...
+ **
+ ** \brief expr_miner
+ **/
+
+#include "cvc4_private.h"
+
+#ifndef __CVC4__THEORY__QUANTIFIERS__EXPRESSION_MINER_H
+#define __CVC4__THEORY__QUANTIFIERS__EXPRESSION_MINER_H
+
+#include <map>
```
4tXJ7f: Add `#include <vector>` and `#include <memory>`

src/theory/quantifiers/expr_miner_manager.h
```diff
@@ -0,0 +1,103 @@
...
+  /** the sygus function passed to initializeSygus, if any */
+  Node d_sygus_fun;
+  /** whether we are using sygus types */
+  bool d_use_sygus_type;
+  /** pointer to the quantifiers engine, used if d_use_sygus is true */
+  QuantifiersEngine* d_qe;
+  /** the sygus term database of d_qe */
+  TermDbSygus* d_tds;
+  /** candidate rewrite database */
+  CandidateRewriteDatabase d_crd;
```
4tXJ7f: Shouldn't this be an `ExprMiner*` or a list of `ExprMiner`s?

src/theory/quantifiers/expr_miner.h
```diff
@@ -0,0 +1,96 @@
...
+class ExprMiner
+{
+ public:
+  ExprMiner() : d_sampler(nullptr) {}
+  virtual ~ExprMiner() {}
+  /** initialize
+   *
+   *
+   */
+  virtual void initialize(const std::vector<Node>& vars,
```
4tXJ7f: Does every `ExprMiner` need a list of variables? If yes, can the list of variables differ between different `ExprMiner`s?

src/theory/quantifiers/expr_miner_manager.h
```diff
@@ -0,0 +1,103 @@
...
+#include "theory/quantifiers_engine.h"
+
+namespace CVC4 {
+namespace theory {
+namespace quantifiers {
+
+/** ExpressionMinerManager
+ *
+ * This class manages a set of expression miners. It provides a common place
+ * to register expression so that multiple mining algorithms can be run in
```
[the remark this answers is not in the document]
ajreynol: No, this should say "expressions".

src/theory/quantifiers/sygus/ce_guided_conjecture.h
```diff
@@ -267,12 +266,12 @@ class CegConjecture {
...
   Node d_nsg_guard;
   //-------------------------------- end non-syntax guided (deprecated)
-  /** candidate rewrite objects for each program variable
+  /** expression miner managers for each program variable
    *
    * This is used for the sygusRewSynth() option to synthesize new candidate
    * rewrite rules.
    */
-  std::map<Node, CandidateRewriteDatabase> d_crrdb;
+  std::map<Node, ExpressionMinerManager> d_exprm;
```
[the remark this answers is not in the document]
ajreynol: No, the stream of expressions generated by multiple functions-to-synthesize are independent, thus they should have independent expression miner managers.

src/theory/quantifiers/expr_miner.h
```diff
@@ -0,0 +1,96 @@
+/*********************                                                        */
+/*! \file expr_miner.h
```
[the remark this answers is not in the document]
ajreynol: Yes, I think eventually we should move some of this code out of quantifiers. How about I open an issue for this?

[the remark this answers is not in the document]
ajreynol: Opened https://github.com/CVC4/CVC4/issues/2442

src/theory/quantifiers/candidate_rewrite_database.cpp
```diff
@@ -252,19 +190,23 @@ bool CandidateRewriteDatabase::addTerm(Node sol,
...
-        out << "(" << (verified ? "" : "candidate-") << "rewrite ";
-        if (d_using_sygus)
-        {
-          Printer* p = Printer::getPrinter(options::outputLanguage());
-          p->toStreamSygus(out, sol);
-          out << " ";
-          p->toStreamSygus(out, eq_sol);
-        }
-        else
+        if (!d_silent)
```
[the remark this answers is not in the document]
ajreynol: This isn't related to verbosity. There may be cases where we want to use the candidate rewrite database as a utility without it printing the rewrites it finds. 

Concretely, the use case is when we are using sygus to enumerate queries, and we use the candidate rewrite database to filter out expressions, but we are not interested in the rewrites themselves.  In this case, we want the rewrites computed but not printed, regardless of the verbosity.

src/theory/quantifiers/expr_miner_manager.h
ajreynol: At the moment, no. The idea is that this class is going to manage the interaction between expression miners, which is unique to the expression miners in question. For example, if query generation is enabled, then rewrite rule generation is used to filter query generation. This interaction is something that the individual miners should not have to worry about (so it can't be virtual-ized).

src/theory/quantifiers/expr_miner.h
ajreynol: The expression miners for a given expression miner manager all have the same variable list.

ajreynol: I'm adding this to the documentation of expr_miner_manager.

src/theory/quantifiers/expr_miner.h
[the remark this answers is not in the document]
4tXJ7f: Yeah, I think that this is a good idea.

src/theory/quantifiers/expr_miner.h
4tXJ7f: In that case, we should store the list of variables in the ExprMinerManager and then just a pointer/reference to the list in the ExprMiners to avoid copying the list.

src/theory/quantifiers/expr_miner_manager.h
4tXJ7f: Hmm, the reason why I am asking is that we are currently not really using the fact that `CandidateRewriteDatabase` is an `ExprMiner`. If we are planning to treat all the `ExprMiner`s as specialized objects, then maybe we should consider not even having an `ExprMiner` super class and just move `initializeChecker` et al. to some kind of utils class. Maybe we can reconsider after we have a couple of `ExprMiner`s.

src/theory/quantifiers/candidate_rewrite_database.cpp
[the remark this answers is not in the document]
4tXJ7f: Ok, that makes sense.

src/theory/quantifiers/sygus/ce_guided_conjecture.h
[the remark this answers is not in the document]
4tXJ7f: I see. I guess it makes sense that different streams get different expression miner managers. Could you add a comment about this?

src/theory/quantifiers/expr_miner_manager.h
ajreynol: Well, there are still common data members: the variables and sampler.

src/theory/quantifiers/expr_miner.h
4tXJ7f: They would depend on someone else owning the vector for them, which wouldn't be the end of the world. Dereferencing the pointer to the vector is a trade off to an extent, yes. You avoid copying a vector and you consume less memory at the cost of an occasional memory dereference. For now, it's fine.

src/theory/quantifiers/expr_miner_manager.h
4tXJ7f: An alternative design would be to package the variables and the sampler into a class that is then a member variable of the different `ExprMiner`s. What's better is rather subjective.

PR #2344, opened 2018-08-21
Add tests that enumerate and verify rewrite rules

4tXJ7f:
The regression script now avoids running with `--check-synth-sol`
when `--sygus-rr` is used.


ajreynol:
Can you update the description of the PR (still talks about exit status 2).
I'm wondering if we should decrease the time and make these in regress1.

4tXJ7f:
I've added an expect to check that "Maximum term size ... exceeded" is outputted as expected. I've also made the regressions take less time and moved them to regress1. With a production build they all take around 1s (some slightly more).

PR #2347, opened 2018-08-21
Addressing 'Invalid iterator comparison' warnings from coverity.

timothy-king:
The main cause seems to be having map operator[] on both sides of comparisons.


ajreynol:
I addressed these in #2349.

timothy-king:
Resolved by #2349.

PR #2354, opened 2018-08-22
More regressions that increase coverage

ajreynol:


mpreiner:
What's the time required to solve each regressions test?

ajreynol:
regress1/quantifiers/lra-vts-inf.smt2 is about 0.2 seconds in production, the others are extremely quick (<.01 second).

PR #2357, opened 2018-08-22
 Fixing some coverity warnings

ajreynol:
@timothy-king , I'm addressing the lowest hanging fruit coverity issues in this PR.


src/theory/quantifiers/ematching/trigger.cpp
```diff
@@ -859,12 +860,15 @@ Node Trigger::getInversion( Node n, Node x ) {
...
       }else{
-        Assert( cindex==-1 );
+        Assert(!cindexSet);
         cindex = i;
+        cindexSet = true;
       }
     }
-    Assert( cindex!=-1 );
-    return getInversion( n[cindex], x );
+    if (cindexSet)
```
4tXJ7f: The fix has a slightly different semantics than before (previously not setting the cindex was considered an error). Is this on purpose?

ajreynol: Yes, this is a case that we should explicitly guard the segfault (it is in code that is very infrequently run).

4tXJ7f: Yes, we should definitely not segfault. What I meant to ask is if we do the right thing if `cindexSet` is false (i.e. if returning a null node is the correct course of action). It does sound like it is though.

PR #2360, opened 2018-08-22
Refactor ITE simplification preprocessing pass.

aniemetz:
Note: This moves theory/ite_utilities.(cpp|h) to preprocessing/util.


src/preprocessing/preprocessing_pass_context.h
```diff
@@ -59,6 +60,7 @@ class PreprocessingPassContext
 
   /** Instance of the ITE remover */
   RemoveTermFormulas* d_iteRemover;
+
```
mpreiner: Is there any reason for adding a newline here?

aniemetz: No, that's a left over, I'll remove.

PR #2362, opened 2018-08-23
Add missing overrides in unit tests

4tXJ7f:
Note: There are still warnings because of the code generated by CxxTest but at least less. Also, I did not clang-format the changes for the review because it changes a lot of formatting.


test/unit/util/rational_white.h
```diff
@@ -35,7 +35,7 @@ class RationalWhite : public CxxTest::TestSuite {
 
   void testCompareAgainstZero(){
     Rational q(0);
-    TS_ASSERT_THROWS_NOTHING(q == 0;);
+    TS_ASSERT_THROWS_NOTHING(q = 0;);
```
aniemetz: Why did you change this?

4tXJ7f: Huh, good catch. Not sure how that happened.

PR #2365, opened 2018-08-23
Fix printing of total bitvectors operators for SMT-LIB 2.6

HanielB:


ajreynol:
Just a note here: this change is justified since smt lib 2.6 does not distinguish between total and partial division. This makes it so that we are able to print/re-parse (rewritten) bv formulas.

PR #2372, opened 2018-08-24
 Fix more simple coverity warnings

ajreynol:
Includes some simplifications and documentation.

@timothy-king , you may be interested in this.


src/theory/sets/theory_sets_private.cpp
```diff
@@ -37,25 +37,26 @@ namespace sets {
...
-  d_external(external),
-  d_notify(*this),
-  d_equalityEngine(d_notify, c, "theory::sets::ee", true),
-  d_conflict(c)
+                                     context::UserContext* u)
+    : d_members(c),
+      d_deq(c),
+      d_deq_processed(u),
+      d_keep(c),
+      d_full_check_incomplete(false),
```
4tXJ7f: What about `d_sentLemma` and `d_addedFact`?

PR #2380, opened 2018-08-26
Fix unsigned integer type issues in strings

ajreynol:
This fixes the nightlies, which were broken by computing a rewrite for substr( _, _, N ) where max(uint32) < N < max(long), which broken on cln builds. 

This PR ensures we always compare overflows in strings against max(uint32).

### DIFF
--- a/src/theory/strings/regexp_operation.cpp
+++ b/src/theory/strings/regexp_operation.cpp
@@ -34,7 +34,7 @@ RegExpOpr::RegExpOpr()
                                                      std::vector<Node>{})),
       d_zero(NodeManager::currentNM()->mkConst(::CVC4::Rational(0))),
       d_one(NodeManager::currentNM()->mkConst(::CVC4::Rational(1))),
-      RMAXINT(LONG_MAX),
+      d_rMaxInt(UINT32_MAX),
       d_sigma(NodeManager::currentNM()->mkNode(kind::REGEXP_SIGMA,
                                                std::vector<Node>{})),
       d_sigma_star(NodeManager::currentNM()->mkNode(kind::REGEXP_STAR, d_sigma))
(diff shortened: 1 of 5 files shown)


src/theory/strings/regexp_operation.cpp
```diff
@@ -34,7 +34,7 @@ RegExpOpr::RegExpOpr()
                                                      std::vector<Node>{})),
       d_zero(NodeManager::currentNM()->mkConst(::CVC4::Rational(0))),
       d_one(NodeManager::currentNM()->mkConst(::CVC4::Rational(1))),
-      RMAXINT(LONG_MAX),
+      d_rMaxInt(UINT32_MAX),
```
4tXJ7f: Nit: `numeric_limits<uint32_t>::max()` is preferred in C++.

src/theory/strings/theory_strings.cpp
```diff
@@ -103,7 +103,7 @@ TheoryStrings::TheoryStrings(context::Context* c,
                              Valuation valuation,
                              const LogicInfo& logicInfo)
     : Theory(THEORY_STRINGS, c, u, out, valuation, logicInfo),
-      RMAXINT(LONG_MAX),
+      d_rMaxInt(UINT32_MAX),
```
4tXJ7f: Ditto

src/theory/strings/theory_strings_rewriter.cpp
```diff
@@ -653,13 +653,13 @@ Node TheoryStringsRewriter::rewriteLoopRegExp(TNode node)
   }
   TNode n1 = node[1];
   NodeManager* nm = NodeManager::currentNM();
-  CVC4::Rational RMAXINT(LONG_MAX);
+  CVC4::Rational rMaxInt(UINT32_MAX);
```
4tXJ7f: Maybe we should define the maximum value in one place and then refer to it from everywhere?

src/theory/strings/theory_strings_rewriter.cpp
```diff
@@ -1216,9 +1216,9 @@ Node TheoryStringsRewriter::rewriteSubstr(Node node)
     if (node[1].isConst() && node[2].isConst())
     {
       CVC4::String s = node[0].getConst<String>();
-      CVC4::Rational RMAXINT(LONG_MAX);
-      unsigned start;
-      if (node[1].getConst<Rational>() > RMAXINT)
+      CVC4::Rational rMaxInt(UINT32_MAX);
```
4tXJ7f: Ditto.

src/theory/strings/theory_strings_rewriter.cpp
```diff
@@ -2641,10 +2641,10 @@ bool TheoryStringsRewriter::stripSymbolicLength(std::vector<Node>& n1,
               // we can remove part of the constant
               // lower bound minus the length of a concrete string is negative,
               // hence lowerBound cannot be larger than long max
-              Assert(lbr < Rational(LONG_MAX));
+              Assert(lbr < Rational(UINT32_MAX));
```
4tXJ7f: Ditto.

PR #2384, opened 2018-08-27
New C++ API: Fix isDefinedKind() to not be ambigious with respect to …

aniemetz:
…underlying type.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,9 +19,6 @@ dist: trusty
 
 env:
  global:
-  # The next declaration is the encrypted COVERITY_SCAN_TOKEN, created
-  #   via the "travis encrypt" command using the project repo's public key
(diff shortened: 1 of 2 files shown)


timothy-king:
I do not think the fix is going to be quick. Maybe assign someone else too.

1. I suggest wide scale performance testing for this change.
2. This is defined to fit in 10 bits elsewhere:
#define __CVC4__EXPR__NODE_VALUE__NBITS__KIND 10
https://github.com/CVC4/CVC4/blob/master/src/expr/metakind_template.h#L122
3. This is also confusingly a uint64_t in another location.
  uint64_t d_kind      : NBITS_KIND;
https://github.com/CVC4/CVC4/blame/master/src/expr/node_value.h#L95
It is not clear the d_kind can be changed to signed without more investigation.

These probably need to be reconciled.

aniemetz:
This change does not affect d_kind or the internal Kind implementation at all. It only concerns the API Kinds (the external kinds). I don't see how 1-3 applies here.

timothy-king:
I misinterpreted which Kind was being updated. Seems fine then.

aniemetz:
I can add a comment that the size of the underlying type of Kind depends on __CVC4__EXPR__NODE_VALUE__NBITS__KIND (which is currently defined as 10 bits).

PR #2385, opened 2018-08-27
Use std:unique_ptr instead of raw pointers in theory/bv.

mpreiner:
This should also fix CIDs 1465687, 1465695, 1465696, and 1465701.

### DIFF
--- a/src/theory/bv/bv_quick_check.cpp
+++ b/src/theory/bv/bv_quick_check.cpp
@@ -145,7 +145,7 @@ bool BVQuickCheck::collectModelInfo(theory::TheoryModel* model, bool fullModel)
 
 BVQuickCheck::~BVQuickCheck() {
   clearSolver();
-  delete d_bitblaster;
+  d_bitblaster.reset();
 }
 
 QuickXPlain::QuickXPlain(const std::string& name, BVQuickCheck* solver, unsigned long budget)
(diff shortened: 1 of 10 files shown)


src/theory/bv/bv_subtheory_algebraic.h
```diff
@@ -181,9 +181,9 @@ class AlgebraicSolver : public SubtheorySolver {
   uint64_t d_numSolved;
   uint64_t d_numCalls;
 
-  context::Context* d_ctx;
-  QuickXPlain* d_quickXplain;   /**< separate quickXplain module as it can reuse the current SAT solver */
-  
+  /**< separate quickXplain module as it can reuse the current SAT solver */
```
4tXJ7f: Nit: remove `<`.

src/theory/bv/bv_quick_check.cpp
```diff
@@ -145,7 +145,7 @@ bool BVQuickCheck::collectModelInfo(theory::TheoryModel* model, bool fullModel)
 
 BVQuickCheck::~BVQuickCheck() {
   clearSolver();
-  delete d_bitblaster;
+  d_bitblaster.reset();
```
4tXJ7f: Why explicitly delete this here? If the `reset()` is necessary, maybe comment on why otherwise just omit it and let the `std::unique_ptr` take care of it.

mpreiner: Yes, you are right. I thought this is actually in a reset method... Will remove it.

src/theory/bv/bv_subtheory_algebraic.h
mpreiner: Done.

mpreiner:
ASAN build is OK and didn't find any issues.

PR #2390, opened 2018-08-27
Solve equalities between Boolean variables in presolve.

ajreynol:
This makes it so that decision-level-0-propagated equalities between Boolean variables X = Y are solved so that X -> Y becomes a model substitution.

This is work towards #2305.

This also fixes an issue where TheoryBool was distinguishing between VARIABLE nodes instead of using isVar(), meaning it was treating input variables and skolem variables differently (likely unintentionally).

### DIFF
--- a/src/theory/booleans/theory_bool.cpp
+++ b/src/theory/booleans/theory_bool.cpp
@@ -40,20 +40,20 @@ Theory::PPAssertStatus TheoryBool::ppAssert(TNode in, SubstitutionMap& outSubsti
 
   // Add the substitution from the variable to its value
   if (in.getKind() == kind::NOT) {
-    if (in[0].getKind() == kind::VARIABLE) {
+    if (in[0].isVar())
+    {
       outSubstitutions.addSubstitution(in[0], NodeManager::currentNM()->mkConst<bool>(false));
-    } else {
-      return PP_ASSERT_STATUS_UNSOLVED;
+      return PP_ASSERT_STATUS_SOLVED;
     }
   } else {
-    if (in.getKind() == kind::VARIABLE) {
+    if (in.isVar())
+    {
       outSubstitutions.addSubstitution(in, NodeManager::currentNM()->mkConst<bool>(true));
-    } else {
-      return PP_ASSERT_STATUS_UNSOLVED;
+      return PP_ASSERT_STATUS_SOLVED;
     }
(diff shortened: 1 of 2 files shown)


src/theory/booleans/theory_bool.cpp
```diff
@@ -40,20 +40,20 @@ Theory::PPAssertStatus TheoryBool::ppAssert(TNode in, SubstitutionMap& outSubsti
 
   // Add the substitution from the variable to its value
   if (in.getKind() == kind::NOT) {
-    if (in[0].getKind() == kind::VARIABLE) {
+    if (in[0].isVar())
+    {
       outSubstitutions.addSubstitution(in[0], NodeManager::currentNM()->mkConst<bool>(false));
-    } else {
-      return PP_ASSERT_STATUS_UNSOLVED;
+      return PP_ASSERT_STATUS_SOLVED;
```
aniemetz: Do we not have to return PP_ASSERT_STATUS_UNSOLVED in the else case here?

src/theory/booleans/theory_bool.cpp
```diff
@@ -40,20 +40,20 @@ Theory::PPAssertStatus TheoryBool::ppAssert(TNode in, SubstitutionMap& outSubsti
...
     }
   } else {
-    if (in.getKind() == kind::VARIABLE) {
+    if (in.isVar())
+    {
       outSubstitutions.addSubstitution(in, NodeManager::currentNM()->mkConst<bool>(true));
-    } else {
-      return PP_ASSERT_STATUS_UNSOLVED;
+      return PP_ASSERT_STATUS_SOLVED;
```
aniemetz: Ditto

src/theory/booleans/theory_bool.cpp
ajreynol: The previous code was returning PP_ASSERT_STATUS_UNSOLVED when in/in[0] *is not* a variable, and fell through to return PP_ASSERT_STATUS_SOLVED.  Since the new code makes it so that there is another case to handle (equalities between variables), the code was flipped so that we return PP_ASSERT_STATUS_SOLVED if in/in[0] *is* a variable, and fall through to call the default method otherwise.

ajreynol:
@barrettcw , you may be interested in this PR.

PR #2393, opened 2018-08-27
Fix sort inference for quantified variables of interpreted types

ajreynol:


test/regress/regress0/fmf/sort-infer-typed-082718.smt2
```diff
@@ -0,0 +1,7 @@
+; COMMAND-LINE: --sort-inference --finite-model-find --no-check-unsat-cores
+; EXPECT: unsat
+(set-option :incremental false)
```
HanielB: is this necessary?

PR #2396, opened 2018-08-28
Refactor how assertions are added to decision engine

4tXJ7f:
Before refactoring the preprocessing passes, we were using three
arguments to add assertions to the decision engine. Now all that
information lives in the AssertionPipeline. This commit moves the
AssertionPipeline to its own file and changes the `addAssertions()`
methods related to the decision engine to take an AssertionPipeline as
an argument instead of three separate ones. Additionally, the
TheoryEngine now uses an AssertionPipeline for lemmas. Finally,
the commit moves substitutions-related information to the preprocessing
pass context. The reason for the move is that the preprocessing
pass context is a better place for information that persists
across multiple check-sat calls whereas the AssertionPipeline
is limited to information about the assertions currently being
processed.


src/preprocessing/assertion_pipeline.h
```diff
@@ -0,0 +1,100 @@
...
+  IteSkolemMap d_iteSkolemMap;
+
+  /** Size of d_nodes when preprocessing starts */
+  size_t d_realAssertionsEnd;
+}; /* class AssertionPipeline */
+
+}  // namespace preprocessing
+}  // namespace CVC4
+
+#endif /* __CVC4__PREPROCESSING__PREPROCESSING_PASS_H */
```
mpreiner: `__CVC4__PREPROCESSING__ASSERTION_PIPELINE_H`

src/decision/decision_engine.h
```diff
@@ -173,21 +174,13 @@ class DecisionEngine {
...
 
-  /* add a set of assertions */
-  void addAssertions(const vector<Node> &assertions);
-
   /**
    * add a set of assertions, while providing a mapping between skolem
    * variables and corresponding assertions. It is assumed that all
    * assertions after assertionEnd were generated by ite
    * removal. Hence, iteSkolemMap maps into only these.
    */
```
mpreiner: This comment needs to be updated since the mapping is now implicitly given by the assertion pipeline.

src/preprocessing/assertion_pipeline.h
```diff
@@ -0,0 +1,100 @@
...
+  {
+    d_nodes.clear();
+    d_realAssertionsEnd = 0;
+  }
+
+  Node& operator[](size_t i) { return d_nodes[i]; }
+  const Node& operator[](size_t i) const { return d_nodes[i]; }
+  void push_back(Node n) { d_nodes.push_back(n); }
+
+  std::vector<Node>& ref() { return d_nodes; }
```
mpreiner: Is it actually a good idea to return a non-const reference of `d_nodes` since it should always be modified via the assertion pipeline methods.

src/preprocessing/passes/apply_substs.cpp
```diff
@@ -60,7 +60,7 @@ PreprocessingPassResult ApplySubsts::applyInternal(
       assertionsToPreprocess->replace(
           i,
           theory::Rewriter::rewrite(
-              assertionsToPreprocess->getTopLevelSubstitutions().apply(
+              d_preprocContext->getTopLevelSubstitutions().apply(
```
mpreiner: Store `d_preprocContext->getTopLevelSubstitutions()`  in a variable before the loop and access it here.

src/smt/smt_engine.cpp
```diff
@@ -3918,7 +3922,7 @@ void SmtEnginePrivate::processAssertions() {
     // proper data structure.
 
     // Placeholder for storing substitutions
-    d_assertions.getSubstitutionsIndex() = d_assertions.size();
+    d_preprocessingPassContext->getSubstitutionsIndex() = d_assertions.size();
```
mpreiner: That is not really nice to read. Can't we add a `setSubstitutionIndex(...)` method?

src/preprocessing/assertion_pipeline.h
4tXJ7f: That would unfortunately imply a lot of changes in a lot of places. I think that it may be better to address this in a separate PR (but I agree that this should be done).

mpreiner: Yes I agree this is too much for this PR. Should we create an issue for that?

src/preprocessing/assertion_pipeline.h
4tXJ7f: Done

src/preprocessing/assertion_pipeline.h
4tXJ7f: Done (#2404)

src/smt/smt_engine.cpp
4tXJ7f: Done

src/preprocessing/passes/apply_substs.cpp
4tXJ7f: Done

src/decision/decision_engine.h
4tXJ7f: Done. I've updated the comment and moved the details to assertion_pipeline.h.

4tXJ7f:
I've addressed the comments. PTAL

PR #2400, opened 2018-08-28
Add regular expression elimination module

ajreynol:
This adds option --re-elim (disabled by default), which eliminates regular expression memberships during ppRewrite in strings.


src/theory/strings/regexp_elim.cpp
```diff
@@ -0,0 +1,441 @@
...
+  }
+  bool success = true;
+  std::vector<Node> char_constraints;
+  Node index = nm->mkBoundVar(nm->integerType());
+  Node substr_ch = nm->mkNode(STRING_SUBSTR, x, index, d_one);
+  substr_ch = Rewriter::rewrite(substr_ch);
+  // handle the case where it is purely characters
+  for (const Node& r : disj)
+  {
+    Assert(r.getKind() != REGEXP_SIGMA);
```
4tXJ7f: We should probably also assert that the `REGEXP_UNION` is flattened, i.e. `r.getKind() != REGEXP_UNION`.

src/theory/strings/regexp_elim.cpp
```diff
@@ -0,0 +1,441 @@
...
+    // e.g.
+    //   x in (re.* (re.union "A" "B" )) --->
+    //     forall k. 0<=k<len(x) => (substr(x,k,1) = "A" OR substr(x,k,1)="B")
+    return returnElim(atom, res, "star-char");
+  }
+  // otherwise, for stars of constant length these are periodic
+  if (disj.size() == 1)
+  {
+    Node r = disj[0];
+    if (r.getKind() == STRING_TO_REGEXP)
```
4tXJ7f: We can generalize this rewrite to also apply in the case `x in ("A" sigma)*` where we'd assert that every even letter is "A" and the string is of even length (i.e. the regexp within the star can be a regex concat of sigmas and constant strings and the sigmas just count towards the length constraints and each constant string adds another substring constraint).

src/theory/strings/regexp_elim.cpp
```diff
@@ -0,0 +1,441 @@
...
+  NodeManager* nm = NodeManager::currentNM();
+  Node x = atom[0];
+  Node lenx = nm->mkNode(STRING_LENGTH, x);
+  Node re = atom[1];
+  // memberships of the form x in re.++ * s1 * ... * sn *, where * are
+  // any number of repetitions (exact or indefinite) of re.allchar.
+  Trace("re-elim-debug") << "Try re concat with gaps " << atom << std::endl;
+  std::vector<Node> children;
+  TheoryStringsRewriter::getConcat(re, children);
+  bool success = true;
```
4tXJ7f: Can we have multiple `success` variables with different, more descriptive names? E.g. the first one could be `onlySigmasAndConsts` or something like that.

src/theory/strings/regexp_elim.cpp
```diff
@@ -0,0 +1,441 @@
...
+          non_greedy_find_vars.push_back(k);
+          prev_end = nm->mkNode(PLUS, prev_end, k);
+        }
+        Node curr = nm->mkNode(STRING_STRIDOF, x, sc, prev_end);
+        Node idofFind = curr.eqNode(d_neg_one).negate();
+        conj.push_back(idofFind);
+        prev_end = nm->mkNode(PLUS, curr, lensc);
+      }
+      // if applicable, process the last gap
+      if (i == (size - 1))
```
4tXJ7f: Suggestion: could move this outside of the loop and have the loop only go up to `sep_children.size() - 1`.

src/theory/strings/regexp_elim.cpp
```diff
@@ -0,0 +1,441 @@
...
+    {
+      Node sc = sep_children[i];
+      if (gap_minsize[i] > 0)
+      {
+        // the gap to this child is at least gap_minsize[i]
+        prev_end =
+            nm->mkNode(PLUS, prev_end, nm->mkConst(Rational(gap_minsize[i])));
+      }
+      Node lensc = nm->mkNode(STRING_LENGTH, sc);
+      if (gap_exact[i])
```
4tXJ7f: Shouldn't this condition be `gap_exact[i] && i + 1 != size`? Otherwise we add two `STRING_SUBSTR` constraints for the last index.

src/theory/strings/regexp_elim.cpp
```diff
@@ -0,0 +1,441 @@
...
+      Assert(!conj.empty());
+      Node res = conj.size() == 1 ? conj[0] : nm->mkNode(AND, conj);
+      // process the non-greedy find variables
+      if (!non_greedy_find_vars.empty())
+      {
+        std::vector<Node> children;
+        for (const Node& v : non_greedy_find_vars)
+        {
+          Node bound = nm->mkNode(
+              AND, nm->mkNode(LEQ, d_zero, v), nm->mkNode(LT, v, lenx));
```
4tXJ7f: Shouldn't the upper bound be `lenx - gap_minsize_end` or something along the lines? We want to make sure that the last of the sep_children cannot overlap with the second to last of the sep_children.

src/theory/strings/regexp_elim.cpp
```diff
@@ -0,0 +1,441 @@
...
+  Node sStartIndex = d_zero;
+  Node sLength = lenx;
+  std::vector<Node> sConstraints;
+  std::vector<Node> rexpElimChildren;
+  for (unsigned r = 0; r < 2; r++)
+  {
+    unsigned index = r == 0 ? 0 : nchildren - 1;
+    Node c = children[index];
+    if (c.getKind() == STRING_TO_REGEXP)
+    {
```
4tXJ7f: Assert that `children[1]` or `children[n - 2]` (depending on the direction) cannot be `STRING_TO_REGEXP` since they should have been merged

src/theory/strings/regexp_elim.cpp
```diff
@@ -0,0 +1,441 @@
...
+      for (unsigned i = 1; i < (nchildren - 1); i++)
+      {
+        rexpElimChildren.push_back(children[i]);
+      }
+    }
+    if (c.getKind() != STRING_TO_REGEXP)
+    {
+      rexpElimChildren.push_back(c);
+    }
+  }
```
4tXJ7f: Can we assert that `rexpElimChildren.size() + sConstraints.size() == nchildren`?

src/theory/strings/regexp_elim.cpp
```diff
@@ -0,0 +1,441 @@
...
+
+  if (!options::regExpElimAgg())
+  {
+    return Node::null();
+  }
+  // only aggressive rewrites below here
+
+  // if the first or last child is constant string, split
```
4tXJ7f: A better description than "split" would be helpful here. Also, why is this an aggressive rewrite? This one seems quite harmless.

src/theory/strings/regexp_elim.cpp
```diff
@@ -0,0 +1,441 @@
...
+    sConstraints.push_back(nm->mkNode(STRING_IN_REGEXP, ss, regElim));
+    Node ret = nm->mkNode(AND, sConstraints);
+    // e.g.
+    // x in re.++( "A", R ) ---> substr(x,0,1)="A" ^ substr(x,1,len(x)-1) in R
+    return returnElim(atom, ret, "concat-splice");
+  }
+  Assert(nchildren > 1);
+  for (unsigned i = 0; i < nchildren; i++)
+  {
+    if (children[i].getKind() == STRING_TO_REGEXP)
```
4tXJ7f: I think that this can be generalized to a sequence of `STRING_TO_REGEXP` and `REGEXP_SIGMA`, actually. Each `STRING_TO_REGEXP` becomes a substr constraint whereas each `REGEXP_SIGMA` contributes to `lens` below.

src/theory/strings/regexp_elim.cpp
```diff
@@ -0,0 +1,441 @@
...
+  // the first gap is initially strict zero
+  gap_minsize.push_back(0);
+  gap_exact.push_back(true);
+  unsigned nchildren = children.size();
+  for (unsigned i = 0; i < nchildren; i++)
+  {
+    Node c = children[i];
+    Trace("re-elim-debug") << "  " << c << std::endl;
+    success = false;
+    if (c.getKind() == STRING_TO_REGEXP)
```
4tXJ7f: It seems like we should also be able to deal with `REGEXP_RANGE` here, no? Ranges would just add a constraint that if you extract that char, it matches the regexp range.

src/theory/strings/regexp_elim.cpp
```diff
@@ -0,0 +1,441 @@
...
+        Node rps = TheoryStringsRewriter::mkConcat(REGEXP_CONCAT, rsuffix);
+        Node ks = nm->mkNode(PLUS, k, lens);
+        Node substrSuffix = nm->mkNode(
+            STRING_IN_REGEXP,
+            nm->mkNode(STRING_SUBSTR, x, ks, nm->mkNode(MINUS, lenx, ks)),
+            rps);
+        echildren.push_back(substrSuffix);
+      }
+      Node body = nm->mkNode(AND, echildren);
+      if (!k.isNull())
```
4tXJ7f: `k` is never null, is it? We probably need a Boolean that tracks whether we introduced a bound variable.

src/theory/strings/regexp_elim.cpp
```diff
@@ -0,0 +1,441 @@
...
+                       nm->mkNode(LEQ, d_zero, k),
+                       nm->mkNode(LT, k, nm->mkNode(MINUS, lenx, lens)));
+        echildren.push_back(bound);
+      }
+      Node substrEq = nm->mkNode(STRING_SUBSTR, x, k, lens).eqNode(s);
+      echildren.push_back(substrEq);
+      if (i > 0)
+      {
+        std::vector<Node> rprefix;
+        rprefix.insert(rprefix.end(), children.begin(), children.begin() + i);
```
4tXJ7f: Nit: the previous statements can be merged to `std::vector<Node> rprefix(children.begin(), children.begin() + i);`

src/theory/strings/regexp_elim.cpp
```diff
@@ -0,0 +1,441 @@
...
+        Node rpn = TheoryStringsRewriter::mkConcat(REGEXP_CONCAT, rprefix);
+        Node substrPrefix = nm->mkNode(
+            STRING_IN_REGEXP, nm->mkNode(STRING_SUBSTR, x, d_zero, k), rpn);
+        echildren.push_back(substrPrefix);
+      }
+      if (i + 1 < nchildren)
+      {
+        std::vector<Node> rsuffix;
+        rsuffix.insert(rsuffix.end(), children.begin() + i + 1, children.end());
```
4tXJ7f: Nit: the previous statements can be merged to `std::vector<Node> rprefix(children.begin() + i + 1, children.end());`

src/theory/strings/regexp_elim.cpp
```diff
@@ -0,0 +1,441 @@
...
+  {
+    for (const Node& r : re[0])
+    {
+      disj.push_back(r);
+    }
+  }
+  else
+  {
+    disj.push_back(re[0]);
+  }
```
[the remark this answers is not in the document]
ajreynol: I'm not sure I understand this syntax. How does this work if re is not kind `REGEXP_UNION`?

src/theory/strings/regexp_elim.cpp
```diff
@@ -0,0 +1,441 @@
...
+  if (disj.size() == 1)
+  {
+    Node r = disj[0];
+    if (r.getKind() == STRING_TO_REGEXP)
+    {
+      Node s = r[0];
+      if (s.isConst())
+      {
+        Node lens = nm->mkNode(STRING_LENGTH, s);
+        Assert(lens.isConst());
```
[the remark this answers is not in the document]
ajreynol: Yes, good catch.

src/theory/strings/regexp_elim.cpp
ajreynol: Yep, we can add this later (out of scope for this PR though).

src/theory/strings/regexp_elim.cpp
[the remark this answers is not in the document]
4tXJ7f: This is a [kinded_iterator](https://github.com/CVC4/CVC4/blob/master/src/expr/node.h#L721-L732). It iterates over the children of the node if the node matches the kind and otherwise just iterates over the node itself.

src/theory/strings/regexp_elim.cpp
ajreynol: Abstractly, each of the non-greedy find variables k just means "I'm going to skip k indices in the string at this point". Based on the way the constraints are constructed, there is no remaining obligations on k at this point.  For example:

`x in re.++( re.*( re.allchar ), "A", re.allchar, "B", re.*( re.allchar ) )` gets converted to:

`indexof( x, "A", k ) != -1 ^ substr( x, indexof( x, "A", k )+2, 1 ) = "B"`, with non-greedy find variable k. At this point we just need to bound k, which can be bound to `0 <= k < len( x )`.

I'll add this as an example.

src/theory/strings/regexp_elim.cpp
ajreynol: I will refactor some of this, I believe there is a bug.

src/theory/strings/regexp_elim.cpp
ajreynol: The reason this could be bad is that confuses the regexp solver.  Say x in R1 and x in R2, the regexp solver has the chance to show unsat by showing R1 intersect R2 is empty. If we change x in R1 to substr( x, n1, n2 ) in R1' ^ ..., then we don't.

src/theory/strings/regexp_elim.cpp
ajreynol: I don't think this is critical to this function.

src/theory/strings/regexp_elim.cpp
ajreynol: This will be refactored.

src/theory/strings/regexp_elim.cpp
ajreynol: This would definitely need to be a "aggressive" rewrite. We maybe could do this (although out of scope for this PR).

src/theory/strings/regexp_elim.cpp
4tXJ7f: Thanks for the explanation, that makes sense.

src/theory/strings/regexp_elim.cpp
4tXJ7f: It is not critical but it would be nice to make it explicit that this rewrite does not apply multiple times in a row.

src/theory/strings/regexp_elim.cpp
4tXJ7f: The previous iteration of the code had an issue (I think) where the rewritten constraints could match more strings than the original constraints if there was an overlap between the last two `sep_children`. The new code resolves that.

src/theory/strings/regexp_elim.cpp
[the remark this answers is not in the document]
ajreynol: Interesting. I think I'd prefer keeping this code since perhaps there would be other ways to infer a regular expression is disjunctive.

PR #2411, opened 2018-08-31
Refactor ceg conjecture initialization

ajreynol:
This fixes a bug where if you use --sygus-qe-preproc and the conjecture was not processed by QE, the first call to check would be in a state that was not properly initialized.

It also moves the "feasiable guard" of sygus out of the single invocation solver and into the main solver.

### DIFF
--- a/src/theory/quantifiers/sygus/ce_guided_conjecture.cpp
+++ b/src/theory/quantifiers/sygus/ce_guided_conjecture.cpp
@@ -168,7 +168,7 @@ void CegConjecture::assign( Node q ) {
   d_feasible_guard = Rewriter::rewrite(d_feasible_guard);
   d_feasible_guard = d_qe->getValuation().ensureLiteral(d_feasible_guard);
   AlwaysAssert(!d_feasible_guard.isNull());
-  // this must be called, both to ensure that the feasible guard is not
+  // this must be called, both to ensure that the feasible guard is
   // decided on with true polariy, but also to ensure that output channel
   // has been used on this call to check.
   d_qe->getOutputChannel().requirePhase(d_feasible_guard, true);
--- a/src/theory/quantifiers/sygus/ce_guided_conjecture.cpp
+++ b/src/theory/quantifiers/sygus/ce_guided_conjecture.cpp
@@ -70,6 +70,7 @@ void CegConjecture::assign( Node q ) {
   Assert( q.getKind()==FORALL );
   Trace("cegqi") << "CegConjecture : assign : " << q << std::endl;
   d_quant = q;
+  NodeManager* nm = NodeManager::currentNM();
 
   // pre-simplify the quantified formula based on the process utility
   d_simp_quant = d_ceg_proc->preSimplify(d_quant);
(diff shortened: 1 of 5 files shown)


src/theory/quantifiers/sygus/ce_guided_conjecture.cpp
```diff
@@ -163,10 +164,19 @@ void CegConjecture::assign( Node q ) {
   }
   
   // initialize the guard
-  if (d_ceg_si->getGuard().isNull())
+  d_feasible_guard = nm->mkSkolem("G", nm->booleanType());
+  d_feasible_guard = Rewriter::rewrite(d_feasible_guard);
```
HanielB: what is the reason again to rewrite fresh boolean constants?

src/theory/quantifiers/sygus/ce_guided_conjecture.cpp
```diff
@@ -163,10 +164,19 @@ void CegConjecture::assign( Node q ) {
...
+  d_feasible_guard = d_qe->getValuation().ensureLiteral(d_feasible_guard);
+  AlwaysAssert(!d_feasible_guard.isNull());
+  // this must be called, both to ensure that the feasible guard is not
```
HanielB: is not --> is

src/theory/quantifiers/sygus/ce_guided_conjecture.cpp
```diff
@@ -196,20 +205,25 @@ bool CegConjecture::isSingleInvocation() const {
...
-    if( d_qe->getValuation().hasSatValue( getGuard(), value ) ) {
-      if( !value ){
-        Trace("cegqi-engine-debug") << "Conjecture is infeasible." << std::endl;
-        return false;
-      }
-    }else{
-      Assert( false );
+  }
+  bool value;
+  Assert(!d_feasible_guard.isNull());
```
HanielB: Why you used `AlwaysAssert` after creating but now you use `Assert`?

src/theory/quantifiers/sygus/ce_guided_single_inv.cpp
```diff
@@ -65,51 +65,53 @@ CegConjectureSingleInv::~CegConjectureSingleInv() {
...
+{
+  Assert(!g.isNull());
+  Assert(!d_single_inv.isNull());
+  // make for new var/sk
+  d_single_inv_var.clear();
+  d_single_inv_sk.clear();
+  Node inst;
+  if (d_single_inv.getKind() == FORALL)
+  {
+    for (unsigned i = 0; i < d_single_inv[0].getNumChildren(); i++)
```
HanielB: cache size

src/theory/quantifiers/sygus/ce_guided_single_inv.cpp
```diff
@@ -65,51 +65,53 @@ CegConjectureSingleInv::~CegConjectureSingleInv() {
...
+    {
+      std::stringstream ss;
+      ss << "k_" << d_single_inv[0][i];
+      Node k = NodeManager::currentNM()->mkSkolem(
```
HanielB: cache nodemanager

src/theory/quantifiers/sygus/ce_guided_conjecture.cpp
ajreynol: needsCheck is called frequently, this is for performance.

src/theory/quantifiers/sygus/ce_guided_conjecture.cpp
ajreynol: OutputChannel's ensureLiteral has the invariant that its input must be in rewritten form, otherwise we crash. While rewriting a Boolean skolem here will almost assuredly do nothing, this is a sanity check.

HanielB: ok

src/theory/quantifiers/sygus/ce_guided_conjecture.cpp
HanielB: ok

PR #2412, opened 2018-08-31
Allows SAT checks of repair const to have different options

HanielB:
Before the new SMT engine used in `sygus_repair_const` needed to have the same options as the original SMT engine. This commit applies the same strategy used in `candidate_rewrite_database` of creating a new expression manager, which can have different options, and keeping a correspondence with the old one. 

The main use of this new feature is to allow timeouts to be used in the SAT checks done by `sygus_repair_const`.

This commit also makes the SAT checks use the default miniscoping (which is disabled when sygus is used) and introduces an option to use the extended rewriter in preprocessing in the subqueries.

(I'm not sure why the diff looks so weird. One reason could be that I inverted the polarity of the `if` in the end of `repairSolution`, so that it returns `false` when we cannot build the solution, rather than build the solution inside the block.)


src/options/quantifiers_options.toml
```diff
@@ -1000,6 +1000,21 @@ header = "options/quantifiers_options.h"
...
   help       = "use approach to repair constants in sygus candidate solutions"
 
+[[option]]
+  name       = "sygusRepairConstTimeout"
+  category   = "regular"
+  long       = "sygus-repair-const-timeout=N"
+  type       = "unsigned long"
+  help       = "timeout (in milliseconds) for the satisfiability check to repair constants in sygus candidate solutions"
+
+[[option]]
```
ajreynol: I wouldn't add this option (otherwise we could want to duplicate every option for the repair-const case). Instead, you can use --ext-rew-prep as a command line parameter, which should be passed automatically to the subcall.

src/theory/quantifiers/sygus/sygus_repair_const.cpp
```diff
@@ -206,68 +207,107 @@ bool SygusRepairConst::repairSolution(const std::vector<Node>& candidates,
                                   << std::endl;
       return false;
     }
+  }
 
-    // do miniscoping explicitly
-    if (fo_body[1].getKind() == AND)
+  Trace("cegqi-engine") << "Repairing previous solution..." << std::endl;
+  // make the satisfiability query
+  bool needExport = true;
```
ajreynol: Can you encapsulate this block of code so that we don't pollute this utility?  I actually did this on this branch: https://github.com/CVC4/CVC4/pull/2340/files.

HanielB:
I've addressed the review.

PR #2423, opened 2018-09-04
Squash implementation of counterexample-guided instantiation

ajreynol:
Currently, InstStrategyCbqi is a base class for a module that does counterexample-guided instantiation. Previously, there were two independent implementations of the technique, however, the old one was eliminated a few years ago, meaning InstStrategyCegqi is the sole class that inherits from it. This squashes the two classes. It also documents some of the moved fields and makes it fit the coding guidelines for public/protected/private blocks.


src/theory/quantifiers/cegqi/inst_strategy_cbqi.cpp
```diff
@@ -49,13 +61,33 @@ InstStrategyCbqi::InstStrategyCbqi(QuantifiersEngine* qe)
 //, d_added_inst( qe->getUserContext() )
 {
   d_qid_count = 0;
+  d_out = new CegqiOutputInstStrategy(this);
```
mpreiner: Use a `std::unique_ptr` here?

src/theory/quantifiers/cegqi/inst_strategy_cbqi.cpp
```diff
@@ -482,17 +532,41 @@ Node InstStrategyCbqi::doNestedQERec( Node q, Node n, std::map< Node, Node >& vi
...
-void InstStrategyCbqi::registerCounterexampleLemma( Node q, Node lem ){
-  Trace("cbqi-debug") << "Counterexample lemma  : " << lem << std::endl;
-  d_quantEngine->addLemma( lem, false );
+void InstStrategyCegqi::registerCounterexampleLemma(Node q, Node lem)
+{
+  // must register with the instantiator
+  // must explicitly remove ITEs so that we record dependencies
+  std::vector<Node> ce_vars;
+  for (unsigned i = 0;
+       i < d_quantEngine->getTermUtil()->getNumInstantiationConstants(q);
```
mpreiner: Cache size.

src/theory/quantifiers/cegqi/inst_strategy_cbqi.cpp
```diff
@@ -482,17 +532,41 @@ Node InstStrategyCbqi::doNestedQERec( Node q, Node n, std::map< Node, Node >& vi
...
+       i++)
+  {
+    ce_vars.push_back(
+        d_quantEngine->getTermUtil()->getInstantiationConstant(q, i));
+  }
+  std::vector<Node> lems;
+  lems.push_back(lem);
+  CegInstantiator* cinst = getInstantiator(q);
+  cinst->registerCounterexampleLemma(lems, ce_vars);
+  for (unsigned i = 0; i < lems.size(); i++)
```
mpreiner: Ditto

PR #2427, opened 2018-09-04
 Extended rewriter for string equalities

ajreynol:
This also includes a few new rewrites for regular expressions.

With this PR, our performance improves approximately 5x for regular expression enumeration.


src/theory/quantifiers/extended_rewrite.cpp
```diff
@@ -1661,6 +1666,155 @@ Node ExtendedRewriter::extendedRewriteArith(Node ret)
...
+                if (vec[j] != hchar)
+                {
+                  isHomogeneous = false;
+                  break;
+                }
+              }
+            }
+            if (isHomogeneous)
+            {
+              std::sort(c[1 - i].begin(), c[1 - i].end());
```
4tXJ7f: Nit: maybe use `std::is_sorted()` to check if sorting is going to do anything. It seems like this should be slightly cheaper than sorting always and comparing later.

src/theory/quantifiers/extended_rewrite.cpp
```diff
@@ -1661,6 +1666,155 @@ Node ExtendedRewriter::extendedRewriteArith(Node ret)
...
+              Node ss = strings::TheoryStringsRewriter::mkConcat(STRING_CONCAT,
+                                                                 c[1 - i]);
+              if (ss != ret[1 - i])
+              {
+                // e.g. "AA" = x ++ y ---> "AA" = y ++ x if y < x
```
4tXJ7f: Note: we can probably slightly strengthen the rewrite by also looking at the constants on the rhs. If any constant contains a letter that is not "A", we can rewrite to false. Otherwise, we can remove all "A"s from the rhs and the same amount from the lhs. Maybe out-of-scope for this PR.

ajreynol: Some of this is already covered by the multi-set reasoning in the strings rewriter for recognizing conflicts, but yeah, removing "A"s from both sides isnt covered and would help.

PR #2431, opened 2018-09-05
 More extended rewrites for strings equality

ajreynol:
Also enables rewrites for equalities with homogenous constants, which was accidentally disabled previously.

### DIFF
--- a/src/theory/quantifiers/extended_rewrite.cpp
+++ b/src/theory/quantifiers/extended_rewrite.cpp
@@ -1729,7 +1729,7 @@ Node ExtendedRewriter::extendedRewriteStrings(Node ret)
         debugExtendedRewrite(ret, new_ret, "string-eq-unify");
         return new_ret;
       }
-      
+
       // ------- using the contains rewriter to reduce equalities
       Node tcontains[2];
       bool tcontainsOneTrue = false;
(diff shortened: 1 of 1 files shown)


src/theory/quantifiers/extended_rewrite.cpp
```diff
@@ -1793,15 +1797,75 @@ Node ExtendedRewriter::extendedRewriteStrings(Node ret)
                 }
               }
             }
-            if (isHomogeneous && !std::is_sorted(c[1-i].begin(),c[1-i].end()))
+            if (isHomogeneous)
             {
+              std::sort(c[1 - i].begin(), c[1 - i].end());
```
4tXJ7f: Nit: instead of having `c[1 - i]` and `c[i]` everywhere maybe assign them to variables.

src/theory/quantifiers/extended_rewrite.cpp
```diff
@@ -1793,15 +1797,75 @@ Node ExtendedRewriter::extendedRewriteStrings(Node ret)
...
+                {
+                  // process the constant : either we have a conflict, or we
+                  // drop an equal number of constants on the LHS
+                  std::vector<unsigned> vecj =
+                      c[1 - i][j].getConst<String>().getVec();
+                  for (unsigned k = 0, sizev = vecj.size(); k < sizev; k++)
+                  {
+                    bool conflict = false;
+                    if (vec.empty())
+                    {
```
4tXJ7f: Can you maybe annotate each of these cases with an example?

src/theory/quantifiers/extended_rewrite.cpp
ajreynol: I think I'll keep them like this (I dont want an overabundance of variable names).

PR #2433, opened 2018-09-05
 Eliminate select over store in quantifier bodies

ajreynol:


4tXJ7f:
Marking it "do not merge" because of @HanielB's comment (it sounds like @ajreynol needs to merge this).

ajreynol:
+3-1 on quantified SMT LIB with a 15 second timeout, virtually no difference (all 4 were close to solved in 15 sec).

PR #2434, opened 2018-09-05
 Make isClosedEnumerable a member of TypeNode

ajreynol:
This allows mayComplete in TermEnumeration to be a static member, which is required for a new functionality.

### DIFF
--- a/src/expr/type_node.cpp
+++ b/src/expr/type_node.cpp
@@ -197,9 +197,9 @@ bool TypeNode::isClosedEnumerable()
       setAttribute(IsClosedEnumerableComputedAttr(), true);
       TypeNode tn = *this;
       const Datatype& dt = getDatatype();
-      for (unsigned i = 0; i < dt.getNumConstructors(); i++)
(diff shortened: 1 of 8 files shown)


src/expr/type_node.cpp
```diff
@@ -165,6 +165,62 @@ bool TypeNode::isInterpretedFinite()
...
+      ret = getSetElementType().isClosedEnumerable();
+    }
+    else if (isDatatype())
+    {
+      // avoid infinite loops: initially set to true
+      setAttribute(IsClosedEnumerableAttr(), ret);
+      setAttribute(IsClosedEnumerableComputedAttr(), true);
+      TypeNode tn = *this;
+      const Datatype& dt = getDatatype();
+      for (unsigned i = 0; i < dt.getNumConstructors(); i++)
```
HanielB: cache size

src/expr/type_node.cpp
```diff
@@ -165,6 +165,62 @@ bool TypeNode::isInterpretedFinite()
...
+      {
+        for (unsigned j = 0; j < dt[i].getNumArgs(); j++)
```
HanielB: ditto

PR #2440, opened 2018-09-07
Using a single condition enumerator in sygus-unif

HanielB:
This commit allows the use of unification techniques in which the search space for conditions in explored independently of the search space for return values (i.e. there is a single condition enumerator for which we always ask new values, similar to the PBE case).

In comparison with asking the ground solver to come up with a minimal number of condition values to resolve all my separation conflicts:

- *main advantage*: can quickly enumerate relevant condition values for solving the problem
- *main disadvantage*: there are many "spurious" values (as in not useful for actual solutions) that get in the way of "good" values when we build solutions from the lazy trie.

A follow-up PR will introduce an information-gain heuristic for building solutions, which ultimately greatly outperforms the other flavor of sygus-unif.

There is also small improvements for trace messages.

### DIFF
--- a/src/options/quantifiers_options.toml
+++ b/src/options/quantifiers_options.toml
@@ -989,7 +989,7 @@ header = "options/quantifiers_options.h"
   long       = "sygus-unif-cond-indpendent-no-repeat-sol"
   type       = "bool"
   default    = "true"
-  help       = "Don't try repeated solutions"
+  help       = "Do not try repeated solutions when using independent synthesis of conditions in unification-based function synthesis"
 
 [[option]]
   name       = "sygusBoolIteReturnConst"
--- a/src/options/quantifiers_options.toml
+++ b/src/options/quantifiers_options.toml
@@ -741,7 +741,7 @@ header = "options/quantifiers_options.h"
   read_only  = true
   help       = "optimization, skip instances based on possibly irrelevant portions of quantified formulas"
 
-### Rewrite rules options 
+### Rewrite rules options
 
 [[option]]
   name       = "quantRewriteRules"
@@ -761,7 +761,7 @@ header = "options/quantifiers_options.h"
   read_only  = true
   help       = "add one instance of rewrite rule per round"
 
-### Induction options 
+### Induction options
 
 [[option]]
   name       = "quantInduction"
@@ -1180,7 +1180,7 @@ header = "options/quantifiers_options.h"
   type       = "bool"
   default    = "true"
   help       = "Minimize synthesis solutions"
-  
+
 [[option]]
   name       = "sygusEvalOpt"
   category   = "regular"
@@ -1196,7 +1196,7 @@ header = "options/quantifiers_options.h"
   type       = "bool"
   default    = "false"
   help       = "static inference techniques for computing whether arguments of functions-to-synthesize are relevant"
-  
+
 # Internal uses of sygus
 
 [[option]]
--- a/src/options/quantifiers_options.toml
+++ b/src/options/quantifiers_options.toml
@@ -976,6 +976,22 @@ header = "options/quantifiers_options.h"
   help       = "Unification-based function synthesis"
 
 [[option]]
+  name       = "sygusUnifCondIndependent"
+  category   = "regular"
+  long       = "sygus-unif-cond-independent"
+  type       = "bool"
+  default    = "false"
+  help       = "Synthesize conditions indepedently from return values (may generate bigger solutions)"
+
+[[option]]
+  name       = "sygusUnifCondIndNoRepeatSol"
+  category   = "regular"
+  long       = "sygus-unif-cond-indpendent-no-repeat-sol"
+  type       = "bool"
+  default    = "true"
+  help       = "Don't try repeated solutions"
+
+[[option]]
   name       = "sygusBoolIteReturnConst"
   category   = "regular"
   long       = "sygus-bool-ite-return-const"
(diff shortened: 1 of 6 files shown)


src/theory/quantifiers/sygus/cegis.cpp
```diff
@@ -166,7 +166,17 @@ bool Cegis::constructCandidates(const std::vector<Node>& enums,
     Trace("cegis") << "  Enumerators :\n";
     for (unsigned i = 0, size = enums.size(); i < size; ++i)
     {
-      Trace("cegis") << "    " << enums[i] << " -> ";
+      bool isUnit = false;
```
ajreynol: isUnit is just for tracing purposes? If so, can you guard this is Trace.isOn("cegis")?

src/theory/quantifiers/sygus/cegis_unif.cpp
```diff
@@ -135,6 +138,22 @@ bool CegisUnif::processConstructCandidates(const std::vector<Node>& enums,
         // get the current unification enumerators
         d_u_enum_manager.getEnumeratorsForStrategyPt(
             e, unif_enums[index][e], index);
+        if (index == 1 && options::sygusUnifCondIndependent())
+        {
+          Assert(unif_enums[index][e].size() == 1);
+          Node eu = unif_enums[index][e][0];
+          Assert(d_u_enum_manager.d_enum_to_active_guard.find(eu)
+                 != d_u_enum_manager.d_enum_to_active_guard.end());
+          Node g = d_u_enum_manager.d_enum_to_active_guard[eu];
```
ajreynol: Can you put this logic in the EnumManager?  Just take this code and make it a function in that class and call it here.  This will be needed for the decision manager refactoring.

src/theory/quantifiers/sygus/cegis_unif.cpp
```diff
@@ -207,6 +228,18 @@ bool CegisUnif::processConstructCandidates(const std::vector<Node>& enums,
     {
       d_sygus_unif.setConditions(
           e, cost_lit, unif_enums[1][e], unif_values[1][e]);
+      // if condition enumerator had value, exclude this value
+      if (options::sygusUnifCondIndependent() && !unif_enums[1][e].empty())
+      {
+        Node eu = unif_enums[1][e][0];
+        Assert(d_u_enum_manager.d_enum_to_active_guard.find(eu)
+               != d_u_enum_manager.d_enum_to_active_guard.end());
+        Node g = d_u_enum_manager.d_enum_to_active_guard[eu];
```
ajreynol: Same thing here, make/use a function in EnumManager for getting the active guard for eu.

src/theory/quantifiers/sygus/sygus_unif_rl.cpp
```diff
@@ -547,6 +560,50 @@ Node SygusUnifRl::DecisionTreeInfo::buildSol(Node cons,
   unsigned c_counter = 0;
   // do we need to resolve a separation conflict?
   bool needs_sep_resolve = false;
+  if (options::sygusUnifCondIndependent())
```
ajreynol: Does this block of code belong here?  Maybe we should have buildSol call two independent functions (buildSolUnif and buildSolUnifIndependent or something).  This would allow us to document the methods independently and would be easier to further modularize later.

src/options/quantifiers_options.toml
```diff
@@ -975,6 +975,22 @@ header = "options/quantifiers_options.h"
...
+  default    = "false"
+  help       = "Synthesize conditions indepedently from return values (may generate bigger solutions)"
+
+[[option]]
+  name       = "sygusUnifCondIndNoRepeatSol"
+  category   = "regular"
+  long       = "sygus-unif-cond-indpendent-no-repeat-sol"
+  type       = "bool"
+  default    = "true"
+  help       = "Don't try repeated solutions"
```
ajreynol: Make this more specific.

src/theory/quantifiers/sygus/cegis.cpp
HanielB: This code block in already inside a `Trace.isOn("cegis")` block

HanielB:
I've addressed the review, PTAL

src/theory/quantifiers/sygus/sygus_unif_rl.h
```diff
@@ -231,6 +244,8 @@ class SygusUnifRl : public SygusUnif
                        const std::vector<Node>& conds);
 
    private:
+    /** built solutions */
```
ajreynol: Can you document this a bit more?  When is this added to? What solutions do you mean?

src/theory/quantifiers/sygus/cegis_unif.cpp
```diff
@@ -329,6 +362,33 @@ void CegisUnifEnumManager::initialize(
...
   incrementNumEnumerators();
+  // create single condition enumerator for each decision tree strategy
+  if (options::sygusUnifCondIndependent())
+  {
+    // allocate a condition enumerator for each candidate
+    for (std::pair<const Node, StrategyPtInfo>& ci : d_ce_info)
+    {
+      Node ceu = nm->mkSkolem("cu", ci.second.d_ce_type);
+      // instantiate template for removing redundant operators
+      if (!ci.second.d_sbt_lemma_tmpl[1].first.isNull())
```
ajreynol: It seems that some of this is copy/pasted code, can we make this a method that initializes an enumerator with the proper symmetry breaking set up?

ajreynol:
There is a syntax error in debug, can you fix?

PR #2441, opened 2018-09-07
Add (str.replace (str.replace y w y) y z) rewrite

4tXJ7f:
This commit adds the following rewrite:

```
(str.replace (str.replace y w y) y z) -->
  (str.replace (str.replace y w z) y z)
if (str.len w) >= (str.len z) and w != z
```

This is a generalization of several rewrite-candidates that we cannot
prove with a 30s timeout (term size limit 2):

```
(candidate-rewrite (str.replace (str.replace x "B" "A") x "A") (str.replace (str.replace x "B" x) x "A"))
(candidate-rewrite (str.replace (str.replace x "A" "B") x "B") (str.replace (str.replace x "A" x) x "B"))
(candidate-rewrite (str.replace (str.replace x "A" "") x "") (str.replace (str.replace x "A" x) x ""))
(candidate-rewrite (str.replace (str.replace x "B" "") x "") (str.replace (str.replace x "B" x) x ""))
```

The justification for the rewrite is given in the comments in the code.
The commit also adds a regression test that tests some pitfalls with
similar terms.

### DIFF
--- a/src/theory/strings/theory_strings_rewriter.cpp
+++ b/src/theory/strings/theory_strings_rewriter.cpp
@@ -2203,6 +2203,57 @@ Node TheoryStringsRewriter::rewriteReplace( Node node ) {
       return returnRewrite(node, res, "repl-subst-idx");
     }
   }
+
+  if (node[0].getKind() == STRING_STRREPL)
+  {
+    Node x = node[0];
+    Node y = node[1];
+    Node z = node[2];
+    if (x[0] == x[2] && x[0] == y)
+    {
+      // (str.replace (str.replace y w y) y z) -->
+      //   (str.replace (str.replace y w z) y z)
+      // if (str.len w) >= (str.len z) and w != z
+      //
+      // Reasoning: There are two cases: (1) w does not appear in y and (2) w
+      // does appear in y.
+      //
+      // Case (1): In this case, the reasoning is trivial. The
+      // inner replace does not do anything, so we can just replace its third
+      // argument with any string.
+      //
+      // Case (2): After the inner replace, we are guaranteed to have a string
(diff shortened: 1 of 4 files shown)


4tXJ7f:
Note: this PR passes regress2 and I've checked that we now can prove the rewrite-candidates affected by this and I've also run the rewrite verification for a while without it reporting an unsound rewrite.

src/theory/strings/theory_strings_rewriter.cpp
```diff
@@ -2203,6 +2203,57 @@ Node TheoryStringsRewriter::rewriteReplace( Node node ) {
...
   }
+
+  if (node[0].getKind() == STRING_STRREPL)
+  {
+    Node x = node[0];
+    Node y = node[1];
+    Node z = node[2];
+    if (x[0] == x[2] && x[0] == y)
+    {
+      // (str.replace (str.replace y w y) y z) -->
```
ajreynol: I have some hesitations about this current rewrite in terms of termination. 
I'm wondering here if both of these rewrite to `(str.replace y w z)`?
It seems that the side conditions ensure both allow us to simplify to a single replace, which would be great.

4tXJ7f: Yeah, they unfortunately don't quite rewrite to `(str.replace y w z)`. My intuition for why the right-hand side is slightly better is because you have at most one of the two replaces doing anything. On the left-hand side, you may have both replaces that change the string. It is maybe not the strongest argument but that's the best reason that I could come up with when deciding on an implementation.

ajreynol: Ok, I think this is fine for now.

PR #2445, opened 2018-09-10
Set NodeManager to nullptr when  exporting vars

4tXJ7f:
PR #2409 assumed that temporarily setting the NodeManager to nullptr
when creating variables is not needed anymore. However, this made our
portfolio build fail. This commit reintroduces the temporary NodeManager
change while creating variables.


4tXJ7f:
Good point, I just checked. The `mkNullaryOperator` is actually exported when we run the regressions with the portfolio build but it seems to work out fine, so we probably don't need to add that again.

PR #2451, opened 2018-09-11
Uses information gain heuristic for building better solutions from DTs

HanielB:
Leads to significant improvements on invariant synthesis benchmarks from Lustre verification in comparison with both CEGIS and the flavor of CegisUnif that encodes the search for minimal solutions in the SMT solver.

Currently we use the basic information gain heuristic that is agnostic to implication points. Since we have a full assignment when attempting to build solutions this is already very good, but still open to further improvements as well as relatively sensitive to "noise" in the data.


src/theory/quantifiers/sygus/sygus_unif_rl.h
```diff
@@ -300,6 +299,52 @@ class SygusUnifRl : public SygusUnif
...
+      void recomputeSolHeuristically(std::map<Node, Node>& hd_mv);
+      /** recursively select (best) conditions to split poinst
+       *
+       * At each call picks the best condition according to information gain and
+       * splits the set of points accordingly, then recurses on them.
+       *
+       * The base case is a set being fully classified (i.e. all points have the
+       * same value)
+       */
+      void buildDt(std::vector<Node>& pts,
```
ajreynol: Is this the right name for this function? It seems that this is specific to the heuristic.

src/theory/quantifiers/sygus/sygus_unif_rl.h
```diff
@@ -300,6 +299,52 @@ class SygusUnifRl : public SygusUnif
...
+      /** true and false nodes */
+      Node d_true;
+      Node d_false;
+      /** cache of conditions evaluations on heads */
+      std::map<std::pair<Node, Node>, Node> d_eval_cond_hd;
+
+      /** rebuild decision tree using information gain heuristic
+       *
+       * In a scenario in which the decision tree potentially contains more
+       * conditions than necessary, it is benefetial to rebuild it in a way that
```
ajreynol: beneficial

src/theory/quantifiers/sygus/sygus_unif_rl.h
```diff
@@ -300,6 +299,52 @@ class SygusUnifRl : public SygusUnif
...
+                   std::vector<Node> conds,
+                   std::map<Node, Node>& hd_mv,
+                   int ind);
+      /** computes the Shannon entropy of a set of points
+       *
+       * The entropy depends on how many positive and negative points are in the
+       * set and in their distribution. The polarity of the points (evaluation
+       * heads) is queried from hd_mv
+       */
+      double getEntropy(const std::vector<Node>& pts,
```
ajreynol: Explain ind argument.

src/theory/quantifiers/sygus/sygus_unif_rl.h
ajreynol: Also, explain the arguments of this method in a formal way.

src/theory/quantifiers/sygus/sygus_unif_rl.cpp
```diff
@@ -902,10 +908,160 @@ Node SygusUnifRl::DecisionTreeInfo::PointSeparator::extractSol(
...
+}
+
+double SygusUnifRl::DecisionTreeInfo::PointSeparator::getEntropy(
+    const std::vector<Node>& hds, std::map<Node, Node>& hd_mv, int ind)
+{
+  double p = 0, n = 0;
+  // get number of points evaluated positively and negatively with feature
+  for (const Node& e : hds)
+  {
+    if (d_dt->d_unif->d_tds->sygusToBuiltin(hd_mv[e]) == d_true)
```
ajreynol: Cache the d_tds.

src/theory/quantifiers/sygus/sygus_unif_rl.h
```diff
@@ -300,6 +300,63 @@ class SygusUnifRl : public SygusUnif
...
+      /** rebuild decision tree using information gain heuristic
+       *
+       * In a scenario in which the decision tree potentially contains more
+       * conditions than necessary, it is beneficial to rebuild it in a way that
+       * "better" conditions occurr closer to the top.
+       *
+       * The information gain heuristic selects conditions that lead to a
+       * greater reduction of the Shannon entropy in the set of points
+       */
+      void recomputeSolHeuristically(std::map<Node, Node>& hd_mv);
```
ajreynol: It seems that PointSeparator is a low level utility, shouldn't this function be at the same level as buildSolution in DecisionTreeInfo?

ajreynol: The same goes for buildDtbuildDtInfoGain.

src/theory/quantifiers/sygus/sygus_unif_rl.h
```diff
@@ -296,10 +356,14 @@ class SygusUnifRl : public SygusUnif
       LazyTrieMulti d_trie;
       /** extracts solution from decision tree built */
       Node extractSol(Node cons, std::map<Node, Node>& hd_mv);
+      /** computes the result of applying cond on the respective point of hd */
```
ajreynol: Can you explain this in a bit more detail?

src/theory/quantifiers/sygus/sygus_unif_rl.h
```diff
@@ -296,10 +356,14 @@ class SygusUnifRl : public SygusUnif
...
+      Node computeCond(Node cond, Node hd);
 
      private:
       /** reference to parent unif util */
       DecisionTreeInfo* d_dt;
+      /** cache of conditions evaluations on heads */
```
ajreynol: Ditto.  Say if d_eval_cond_head[<s,t>] = u, then s,t,u are such that ....

src/theory/quantifiers/sygus/sygus_unif_rl.h
```diff
@@ -296,10 +356,26 @@ class SygusUnifRl : public SygusUnif
...
+      Node computeCond(Node cond, Node hd);
 
      private:
       /** reference to parent unif util */
       DecisionTreeInfo* d_dt;
+      /** cache of conditions evaluations on heads
+       *
+       * If for example cond is (\lambda xy. x < y) and hd is an evaluation head
+       * in point (hd 0 1), then after invoking computeCond(cond, hd) this map
+       * will contain d_eval_cond_hd[<c, hd>] = true, since
```
ajreynol: You mean c -> cond?

PR #2452, opened 2018-09-11
Make README an org file

ericlesaquiles:
i find it somewhat pleasant to have a prettier readme

### DIFF
--- a/README.org
+++ b/README.org
@@ -33,22 +33,22 @@ for help getting started using CVC4.
 
 We are always happy to hear feedback from our users:
 
-* if you need help with using CVC4, please refer to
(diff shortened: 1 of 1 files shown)


4tXJ7f:
Thank you for you suggestion. It is true that some nicer formatting would make the README more pleasant to read and we are actually considering to move to markdown. The problem with org files is that not every editor has native support for them and we would like to avoid editor-specific formats.

ericlesaquiles:
Personally, I find an org file more organized, even in an ordinary text editor (it is just plain text anyway). But I understand your point and thank you for your response

PR #2453, opened 2018-09-11
Fix compiler warning in line_buffer.cpp.

aniemetz:


aniemetz:
This IS backwards ;-)

PR #2454, opened 2018-09-11
New C++ API: Try to fix (false positive) Coverity warnings.

aniemetz:


4tXJ7f:
Hmm, unfortunately, this didn't seem to help :( I guess we will have to go the route of declaring Kind to be `int32_t` but then we have to be careful.

PR #2455, opened 2018-09-11
 New C++ API: Add checks for Terms/OpTerms.

aniemetz:


src/api/cvc4cpp.cpp
```diff
@@ -2342,9 +2145,13 @@ OpTerm Solver::mkOpTerm(Kind kind, uint32_t arg)
       res = d_exprMgr->mkConst(CVC4::TupleUpdate(arg));
       break;
     default:
-      // CHECK: kind valid?
-      Assert(!res.isNull());
+      PrettyCheckArgument(true,
```
4tXJ7f: This should be `false` instead of `true`, no?

src/api/cvc4cpp.cpp
```diff
@@ -2383,9 +2190,13 @@ OpTerm Solver::mkOpTerm(Kind kind, uint32_t arg1, uint32_t arg2)
       res = d_exprMgr->mkConst(CVC4::FloatingPointToFPGeneric(arg1, arg2));
       break;
     default:
-      // CHECK: kind valid?
-      Assert(!res.isNull());
+      PrettyCheckArgument(true,
```
4tXJ7f: Ditto.

src/api/cvc4cpp.h
```diff
@@ -2412,6 +2412,10 @@ class CVC4_PUBLIC Solver
   std::vector<Type> sortVectorToTypes(const std::vector<Sort>& vector) const;
   /* Helper to convert a vector of sorts to internal types. */
   std::vector<Expr> termVectorToExprs(const std::vector<Term>& vector) const;
+  /* Helper to check for API misuse in mkTerm functions. */
+  void check_mk_op_term(OpTerm opTerm, uint32_t nchildren) const;
```
4tXJ7f: Nit: camelCase would be more consistent.

src/api/cvc4cpp.cpp
```diff
@@ -630,6 +632,16 @@ CVC4::Kind extToIntKind(Kind k)
   }
   return it->second;
 }
+
+uint32_t minArity(Kind k)
+{
+  return CVC4::ExprManager::minArity(extToIntKind(k));
```
4tXJ7f: Coverity might object to that again, I am not sure (because it thinks that there could be negative values).

src/api/cvc4cpp.cpp
```diff
@@ -2010,31 +1917,72 @@ Term Solver::mkConst(Kind kind, uint32_t arg1, uint32_t arg2, Term arg3) const
...
 /* Create terms                                                               */
 /* -------------------------------------------------------------------------- */
 
+void Solver::check_mk_term(Kind kind, uint32_t nchildren) const
+{
+  PrettyCheckArgument(isDefinedKind(kind),
+                      kind,
+                      "Invalid kind '%s'",
+                      kindToString(kind).c_str());
+  const CVC4::kind::MetaKind mk = kind::metaKindOf(extToIntKind(kind));
```
4tXJ7f: Coverity may complain here.

src/api/cvc4cpp.cpp
```diff
@@ -630,6 +632,16 @@ CVC4::Kind extToIntKind(Kind k)
...
 }
+
+uint32_t minArity(Kind k)
+{
+  return CVC4::ExprManager::minArity(extToIntKind(k));
+}
+
+uint32_t maxArity(Kind k)
+{
+  return CVC4::ExprManager::maxArity(extToIntKind(k));
```
4tXJ7f: Same here.

src/api/cvc4cpp.cpp
```diff
@@ -2010,31 +1917,72 @@ Term Solver::mkConst(Kind kind, uint32_t arg1, uint32_t arg2, Term arg3) const
...
 
+void Solver::check_mk_term(Kind kind, uint32_t nchildren) const
+{
+  PrettyCheckArgument(isDefinedKind(kind),
+                      kind,
+                      "Invalid kind '%s'",
+                      kindToString(kind).c_str());
+  const CVC4::kind::MetaKind mk = kind::metaKindOf(extToIntKind(kind));
+  PrettyCheckArgument(
+      mk == kind::metakind::PARAMETERIZED || mk == kind::metakind::OPERATOR,
```
4tXJ7f: Shouldn't this be `mk != kind::metakind::PARAMETERIZED && mk != kind::metakind::OPERATOR`?

src/api/cvc4cpp.cpp
```diff
@@ -2010,31 +1917,72 @@ Term Solver::mkConst(Kind kind, uint32_t arg1, uint32_t arg2, Term arg3) const
...
+
+void Solver::check_mk_op_term(OpTerm opTerm, uint32_t nchildren) const
+{
+  const Kind kind = opTerm.getKind();
+  const CVC4::Kind int_kind = extToIntKind(kind);
+  const CVC4::Kind int_op_kind =
+      NodeManager::operatorToKind(opTerm.d_expr->getNode());
+  PrettyCheckArgument(int_kind != kind::BUILTIN
+                          && CVC4::kind::metaKindOf(int_op_kind)
+                                 != kind::metakind::PARAMETERIZED,
```
4tXJ7f: Shouldn't this be `int_kind == kind::BUILTIN || CVC4::kind::metaKindOf(int_op_kind) == kind::metakind::PARAMETERIZED`?

src/api/cvc4cpp.cpp
```diff
@@ -2010,31 +1917,72 @@ Term Solver::mkConst(Kind kind, uint32_t arg1, uint32_t arg2, Term arg3) const
...
+                      opTerm,
+                      "This term constructor is for parameterized kinds only");
+  check_mk_term(kind, nchildren);
```
4tXJ7f: This probably won't work out because `kind` is parametrized and `check_mk_term` does not accept parametrized kinds?

src/api/cvc4cpp.cpp
aniemetz: No. See https://github.com/CVC4/CVC4/blob/master/src/expr/expr_manager_template.cpp#L172.

src/api/cvc4cpp.cpp
aniemetz: Indeed. PrettyCheckArgument is so confusing since it expects the condition where it should *not* fail rather than the condition where it should fail. Using PrettyCheckArgument is temporary anyhow, but I might write a new macro for these checks (that also doesn't use printf) before continuing to add these checks just because this is annoying and we have to do it anyways.

src/api/cvc4cpp.cpp
aniemetz: You're right, I will have to duplicate the checks.

src/api/cvc4cpp.cpp
aniemetz: Yes. Again, confusing....

src/api/cvc4cpp.h
aniemetz: I agree, that's a leftover since it first was a macro that I wanted to convert to a 'static' function....

src/api/cvc4cpp.cpp
4tXJ7f: Right, that makes sense. I got confused by the original comment (because of the `else` part): https://github.com/CVC4/CVC4/pull/2455/files#diff-d4eff0da8e72b4f2d088f43ce4ecf899L2069 .

src/api/cvc4cpp.cpp
4tXJ7f: That sounds good. I think that stating the condition in the positive is ok since it mirrors an `Assert()` but the name of the macro should reflect that (with `PrettyCheckArgument` the naming is a bit ambiguous and I had to look up which way it goes).

aniemetz:
Migrated this to the new API check macros, addressed everything (but might have missed something that Coverity will complain about, but let's wait for it to actually complain). @4tXJ7f PTAL.

src/api/cvc4cpp.cpp
```diff
@@ -2019,31 +1934,79 @@ Term Solver::mkConst(Kind kind, uint32_t arg1, uint32_t arg2, Term arg3) const
...
+          || CVC4::kind::metaKindOf(int_op_kind)
+                 == kind::metakind::PARAMETERIZED,
+      opTerm,
+      "This term constructor is for parameterized kinds only");
+  if (nchildren)
+  {
+    uint32_t min_arity = ExprManager::minArity(int_op_kind);
+    uint32_t max_arity = ExprManager::maxArity(int_op_kind);
+    CVC4_API_KIND_CHECK_EXPECTED(
+        nchildren >=  min_arity && nchildren <= max_arity,
```
4tXJ7f: Micro nit: there are two spaces after `>=`. Maybe clang-format didn't catch that.

PR #2457, opened 2018-09-12
Add Skolem cache for strings, refactor length registration

ajreynol:
This PR is in preparation for doing more aggressive caching of skolems (for example, in the strings preprocessor).

It refactors sendLengthLemma -> registerLength, which unifies how length lemmas for skolems and other string terms are handled.


src/theory/strings/theory_strings.cpp
```diff
@@ -2624,47 +2626,55 @@ void TheoryStrings::processNEqc( std::vector< std::vector< Node > > &normal_form
...
+  ssi << pinfer[use_index].d_id;
+  sendInference(pinfer[use_index].d_ant,
+                pinfer[use_index].d_antn,
+                pinfer[use_index].d_conc,
+                ssi.str().c_str(),
+                pinfer[use_index].sendAsLemma());
+  // Register the new skolems from this inference. We register them here
+  // (lazily), since the code above has now decided to use the inference
+  // at use_index that involves them.
+  for (std::pair<const LengthStatus, std::vector<Node> > sks :
```
4tXJ7f: Nit: `const std::pair<const LengthStatus, std::vector<Node> >&`

src/theory/strings/theory_strings.cpp
```diff
@@ -2624,47 +2626,55 @@ void TheoryStrings::processNEqc( std::vector< std::vector< Node > > &normal_form
...
+       pinfer[use_index].d_new_skolem)
+  {
+    for (const Node& n : sks.second)
+    {
+      registerLength(n, sks.first);
```
4tXJ7f: Should we assert that `sks.first` is either `LENGTH_SPLIT` or `LENGTH_GEQ_ONE` (that's the only two cases that the original code dealt with)?

src/theory/strings/theory_strings.cpp
```diff
@@ -3842,49 +3923,6 @@ Node TheoryStrings::mkLength( Node t ) {
...
-    return it->second;
-  }
-}
-
-//isLenSplit: -1-ignore, 0-no restriction, 1-greater than one, 2-one
-Node TheoryStrings::mkSkolemS( const char *c, int isLenSplit ) {
-  Node n = NodeManager::currentNM()->mkSkolem( c, NodeManager::currentNM()->stringType(), "string sko" );
-  d_all_skolems.insert(n);
-  d_length_lemma_terms_cache.insert( n );
-  ++(d_statistics.d_new_skolems);
```
4tXJ7f: Either update this statistic in the new code or remove it.

src/theory/strings/theory_strings.cpp
ajreynol: No, the new code is more robust because it is agnostic to the LengthStatus here. We may add other LengthStatus for new skolems, there is no need to restrict them here.

4tXJ7f: Great!

src/theory/strings/theory_strings.cpp
4tXJ7f: Also `const` :)

PR #2459, opened 2018-09-12
Improvements and fixes for symmetry detection and breaking

ajreynol:
This fixes a few open issues with symmetry detection algorithm. It also extends the algorithm to do:
- Alpha equivalence to recognize symmetries between quantified formulas,
- A technique to recognize a subset of variables in two terms are symmetric, e.g. from x in A ^ x in B, we find A and B are interchangeable by treating x as a fixed symbol,
- Symmetry breaking for maximal subterms instead of variables.


src/preprocessing/passes/symmetry_breaker.cpp
```diff
@@ -101,23 +104,29 @@ Node SymmetryBreaker::generateSymBkConstraints(const vector<vector<Node>>& parts
...
-    return kind::BITVECTOR_ULE;
+    return LEQ;
   }
-  else
+  else if (tn.isBitVector())
   {
-    return kind::EQUAL;
+    return BITVECTOR_ULE;
   }
+  if (tn.isFirstClass())
```
HanielB: maybe use `else` as in the above examples to be consistent? Or you could remove the `else` above

src/preprocessing/passes/symmetry_breaker.cpp
```diff
@@ -128,27 +137,40 @@ PreprocessingPassResult SymBreakerPass::applyInternal(
 {
   Trace("sym-break-pass") << "Apply symmetry breaker pass..." << std::endl;
   // detect symmetries
-  std::vector<std::vector<Node>> part;
+  std::vector<std::vector<Node>> sterms;
   symbreak::SymmetryDetect symd;
-  symd.getPartition(part, assertionsToPreprocess->ref());
+  symd.computeTerms(sterms, assertionsToPreprocess->ref());
+  if (!sterms.empty())
```
HanielB: Is this supposed to be a placeholder for code to be added later? Otherwise inline this `if` in the trace command

src/preprocessing/passes/symmetry_breaker.cpp
```diff
@@ -128,27 +137,40 @@ PreprocessingPassResult SymBreakerPass::applyInternal(
...
+  {
+    Trace("sb-constraint") << "; found symmetry" << std::endl;
+  }
   if (Trace.isOn("sym-break-pass"))
   {
-    Trace("sym-break-pass") << "Detected symmetry partition:" << std::endl;
-    for (const std::vector<Node>& p : part)
+    if (sterms.empty())
```
HanielB: this `if-then-else` can also easily be embedded in the trace below. And you will not get in the `for` anyway when `sterms` is empty

src/preprocessing/passes/symmetry_detect.h
```diff
@@ -65,10 +66,18 @@ class Partition
    * { w-> { x, y } }
    */
   std::map<Node, std::vector<Node> > d_subvar_to_vars;
+  /** add variable */
```
HanielB: detail arguments

src/preprocessing/passes/symmetry_detect.h
```diff
@@ -65,10 +66,18 @@ class Partition
...
+  void addVariable(Node sv, Node v);
+  /** remove variable */
+  void removeVariable(Node sv);
```
HanielB: it seems that `removeVariable` is not the "inverse" of `addVariable`. Clarify behavior

src/preprocessing/passes/symmetry_detect.h
```diff
@@ -176,61 +186,39 @@ class PartitionMerger
...
    * Destructor
    * */
   ~SymmetryDetect() {}
 
   /** Get the final partition after symmetry detection.
-   *  If a vector in parts contains two variables x and y,
-   *  then assertions and assertions { x -> y, y -> x } are
+   *
+   *  If a vector in sterms contains two variables x and y,
+   *  then assertions and assertions { x -> y, x -> y } are
```
HanielB: is the substitution right?

src/preprocessing/passes/symmetry_detect.h
```diff
@@ -275,6 +266,21 @@ class SymmetryDetect
    * */
   void collectChildren(Node node, std::vector<Node>& children);
 
+  /** Compute alpha equivalent terms
+   *
+   * This constructs sterm_to_indices such that if sterm_to_indices[t]
```
HanielB: Explain this also in a more abstract way

src/preprocessing/passes/symmetry_detect.cpp
```diff
@@ -475,11 +470,7 @@ Node SymmetryDetect::getSymBreakVariable(TypeNode tn, unsigned index)
   {
     std::stringstream ss;
     ss << "_sym_bk_" << tn << "_" << (it->second.size() + 1);
-    Node fresh_var =
-        NodeManager::currentNM()->mkSkolem(ss.str(),
-                                           tn,
-                                           "symmetry breaking variable",
-                                           NodeManager::SKOLEM_EXACT_NAME);
+    Node fresh_var = NodeManager::currentNM()->mkBoundVar(ss.str(), tn);
```
HanielB: cache `NodeManager` outside loop

src/preprocessing/passes/symmetry_detect.cpp
```diff
@@ -567,14 +648,18 @@ Partition SymmetryDetect::findPartitions(Node node)
...
-  vector<Partition> partitions;
-
   // Get all children of node
   Trace("sym-dt-debug") << "[sym-dt] collectChildren for: " << node
-                        << " with operator " << node.getKind() << endl;
+                        << " with kind " << node.getKind() << endl;
+  // Children of node
+  std::vector<Node> children;
+  bool operatorChild = false;
+  if (node.getKind() == kind::APPLY_UF)
```
HanielB: remove `kind::`

src/preprocessing/passes/symmetry_detect.cpp
```diff
@@ -746,7 +894,7 @@ Partition SymmetryDetect::findPartitions(Node node)
   }
   else
   {
-    if (node.getMetaKind() == kind::metakind::PARAMETERIZED)
+    if (node.getMetaKind() == kind::metakind::PARAMETERIZED && !operatorChild)
```
HanielB: remove `kind::`

src/preprocessing/passes/symmetry_detect.cpp
```diff
@@ -512,6 +504,98 @@ void SymmetryDetect::getPartition(vector<vector<Node> >& parts,
...
+        }
+      }
+    }
+  }
+}
+
+/**
+ * We build the partition trie indexed by
+ * parts[0].var_to_subvar[v]....parts[n].var_to_subvar[v]. The leaves of a
+ * partition trie is the new regions of a partition
```
HanielB: is --> are

src/preprocessing/passes/symmetry_detect.cpp
```diff
@@ -512,6 +504,98 @@ void SymmetryDetect::getPartition(vector<vector<Node> >& parts,
...
+        sterms.push_back(sp.second);
+      }
+      else
+      {
+        unsigned tsymId = tsids[tsids.size() - 1];
+        Trace("sym-dt-tsym-cons") << "...use tsym id " << tsymId << std::endl;
+        // get the substitution
+        std::vector<Node> vars;
+        std::vector<Node> subs;
+        for (const Node v : sp.second)
```
HanielB: const Node&

src/preprocessing/passes/symmetry_detect.cpp
```diff
@@ -676,67 +751,140 @@ Partition SymmetryDetect::findPartitions(Node node)
...
+  // for each variable
+  for (const Node& n : all_vars)
+  {
+    Trace("sym-dt-debug") << "[sym-dt] Add a variable {" << n
+                          << "} to the partition trie, #partitions = "
+                          << active_indices.size() << "..." << endl;
+    std::vector<Node> subvars;
+    std::vector<unsigned> useVarInd;
+    Node useVar;
+    for (const unsigned& i : active_indices)
```
HanielB: just use `unsigned`?

src/preprocessing/passes/symmetry_detect.cpp
```diff
@@ -676,67 +751,140 @@ Partition SymmetryDetect::findPartitions(Node node)
...
+  if (!isAssocComm)
   {
-    Node an = pt.addNode(n, active_partitions);
-    // if this is a new node, allocate
-    if (an == n)
+    Assert(active_indices.size() == children.size());
+    // order matters, and there is no chance we merged children
+    schildren.resize(children.size());
+  }
+  for (const unsigned& i : active_indices)
```
HanielB: ditto

src/preprocessing/passes/symmetry_breaker.cpp
ajreynol: This isn't mutually exclusive with the others, this is a type class whereas the others are types.

src/preprocessing/passes/symmetry_detect.h
ajreynol: Will fix.

PR #2468, opened 2018-09-15
Refactor preprocessing pass registration

4tXJ7f:
This commit refactors how preprocessing pass registration works,
inspired by LLVM's approach [0]. The basic idea is that every
preprocessing pass declares a static variable of type `RegisterPass` in
its source file that registers the pass with the
`PreprocessingPassRegistry` when starting the program. The registry is a
singleton that keeps track of all the available passes and allows other
code to create instances of the passes (note: previously the registry
itself was owning the passes but this is no longer the case). One of the
advantages of this solution is that we have a list of available passes
directly at the beginning of the program, which is useful for example
when parsing options.

As a side effect, this commit also fixes the SortInference pass, which
was expecting arguments other than the preprocessing pass context in its
constructor.

This commit is required for fixing dumping pre/post preprocessing
passes. It is also the ground work for allowing the user to specify a
preprocessing pipeline using command-line arguments.

[0] https://llvm.org/docs/WritingAnLLVMPass.html


src/preprocessing/preprocessing_pass_registry.h
```diff
@@ -29,32 +29,102 @@
 namespace CVC4 {
 namespace preprocessing {
 
+class PreprocessingPassContext;
+
+/**
+ * The PreprocessingPassRegistry keeps track of the available preprocessing
+ * passes and how to create instances of those passes. This class is intended
+ * to be used as a singleton and can be shared between threads (assuming that
```
aniemetz: Is this really true? We have passes that maintain pass-specific data that is at the same time specific to the current SmtEngine, e.g., https://github.com/CVC4/CVC4/blob/master/src/preprocessing/passes/non_clausal_simp.h#L50.

4tXJ7f: In that case, you'd create multiple instances of a pass. Each SmtEngine gets its own instance.

Issue #2473, opened 2018-09-17
AssertionPipeline: Split assertions into three vectors


4tXJ7f:
There are three types of assertions that we store in `AssertionPipeline`: user assertions, substitutions (only in incremental mode), and assertions generated by ITE removal. These should be split into three different vectors instead of using "magical" indices.

Issue #1255, opened 2017-10-17
Substitutions in incremental mode are handled with a hack


barrettcw:
In incremental mode, new substitutions obtained after the top level assertions have been processed are stuffed into a "special" assertion in the assertion list.  This has bit us a few times.  This should be refactored and handled in a cleaner way.l

PR #2474, opened 2018-09-17
Make substitution index context-independent

4tXJ7f:
When we do solving in incremental mode, we store substitutions at a
special index in our list of assertions. Previously, we used a
context-dependent variable for that. However, this is not needed since
the list of assertions just consists of the assertions currently being
processed, which are independent of the assertions seen so far. This
commit changes the index to be an ordinary integer and moves it to the
AssertionPipeline. Additionally, it abstracts access to the index in
preparation for splitting AssertionPipeline into three vectors (see
issue #2473).

@barrettcw I am assigning this to you since you were the one who opened #1255. Please let me know if this is a step in the right direction.


mpreiner:
@barrettcw @4tXJ7f what's the status of this PR?

PR #2476, opened 2018-09-17
Decision strategy: incorporate UF with cardinality constraints

ajreynol:
Most of the LOC in these changes are because the code for option --uf-ss-totality was split to different locations.

### DIFF
--- a/src/theory/uf/theory_uf_strong_solver.cpp
+++ b/src/theory/uf/theory_uf_strong_solver.cpp
@@ -441,32 +441,51 @@ void Region::debugPrint( const char* c, bool incClique ) {
   }
 }
 
-SortModel::SortModel( Node n,
-                      context::Context* c,
-                      context::UserContext* u,
-                      StrongSolverTheoryUF* thss )
-  : d_type( n.getType() )
-  , d_thss( thss )
-  , d_regions_index( c, 0 )
-  , d_regions_map( c )
-  , d_split_score( c )
-  , d_disequalities_index( c, 0 )
-  , d_reps( c, 0 )
-  , d_conflict( c, false )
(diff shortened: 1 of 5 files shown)


src/theory/uf/theory_uf_strong_solver.cpp
```diff
@@ -1357,16 +1258,62 @@ int SortModel::getNumRegions(){
...
+    d_totality_terms[0][c - 1] = var;
+  }
+  else
+  {
+    d_totality_terms[0].push_back(var);
+  }
+  // must be distinct from all other cardinality terms
+  for (int i = 1, size = d_totality_terms[0].size(); i < size; i++)
+  {
+    Node lem = nm->mkNode(NOT, var.eqNode(d_totality_terms[0][i - 1]));
```
HanielB: nit: replace the explicit `mkNode(NOT...` with `.notNode`?

src/theory/uf/theory_uf_strong_solver.cpp
```diff
@@ -1357,16 +1258,62 @@ int SortModel::getNumRegions(){
...
+    Trace("uf-ss-lemma") << "Totality distinctness lemma : " << lem
+                         << std::endl;
+    d_thss->getOutputChannel().lemma(lem);
+  }
+  // must send totality axioms for each existing term
+  for (NodeIntMap::iterator it = d_regions_map.begin();
```
HanielB: you're using iterators because you don't trust foreach with context-dependent data-structures?

ajreynol: Yes, we haven't tested those as far as I know.

PR #2486, opened 2018-09-18
New C++ API: Introduce new macro and exception for API checks.

aniemetz:
This introduces a new macro `CVC4_API_CHECK` for guards against API misuse.

Note that the destructor of stream `CVC4ApiExceptionStream` has to be explicitly set to `noexcept(false)`. This is due to the fact that in C++11, destructors default to noexcept(true), which triggers a call to std::terminate() if the destructor throws an exception.

### DIFF
--- a/src/api/cvc4cpp.cpp
+++ b/src/api/cvc4cpp.cpp
@@ -17,6 +17,7 @@
 #include "api/cvc4cpp.h"
 
 #include "base/cvc4_assert.h"
+#include "base/cvc4_check.h"
 #include "expr/expr.h"
 #include "expr/expr_manager.h"
 #include "expr/kind.h"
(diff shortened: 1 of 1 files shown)


src/api/cvc4cpp.cpp
```diff
@@ -34,6 +35,46 @@
...
 
+namespace {
+/* -------------------------------------------------------------------------- */
+/* API guard helpers                                                          */
+/* -------------------------------------------------------------------------- */
+
+class CVC4_PUBLIC CVC4ApiException : public Exception
+{
+ public:
+  CVC4ApiException(std::stringstream& stream) { d_msg = stream.str(); }
```
4tXJ7f: I think that this can be `const std::stringstream&`.

src/api/cvc4cpp.cpp
```diff
@@ -34,6 +35,46 @@
 namespace CVC4 {
 namespace api {
 
+namespace {
+/* -------------------------------------------------------------------------- */
+/* API guard helpers                                                          */
+/* -------------------------------------------------------------------------- */
+
+class CVC4_PUBLIC CVC4ApiException : public Exception
```
4tXJ7f: We probably need to define this exception in the header s.t. people can catch it. I don't think that it necessarily needs to inherit from `CVC4::Exception` but can directly inherit from `std::exception`.

src/api/cvc4cpp.cpp
```diff
@@ -34,6 +35,46 @@
...
+   * a destructor that throws an exception and in C++11 all destructors
+   * default to noexcept(true) (else this triggers a call to std::terminate). */
+  CVC4_NO_RETURN ~CVC4ApiExceptionStream() noexcept(false)
+  {
+    throw CVC4ApiException(d_stream);
+  }
+
+  std::ostream& ostream() { return d_stream; }
+
+  CVC4ApiExceptionStream& operator<<(const std::string& msg)
```
4tXJ7f: This method is not needed (I think).

src/api/cvc4cpp.cpp
```diff
@@ -1813,57 +1854,50 @@ Term Solver::mkConst(RoundingMode rm) const
 
 Term Solver::mkConst(Kind kind, Sort arg) const
 {
-  PrettyCheckArgument(kind == EMPTYSET,
-                      kind,
-                      "Invalid kind '%s', expected EMPTY_SET",
-                      kindToString(kind).c_str());
+  CVC4_API_CHECK(kind == EMPTYSET)
+      << "Invalid kind '" << kindToString(kind).c_str()
```
4tXJ7f: No need to call `c_str()` anymore (applies to all checks).

src/api/cvc4cpp.cpp
aniemetz: That's what I'm currently doing, this PR will change quite a bit, just FYI.

PR #2491, opened 2018-09-18
More aggressive caching of string skolems.

ajreynol:
This includes avoiding skolems for substr when remainders are guaranteed to be empty, and unifying the string skolems for indexof and replace.

### DIFF
--- a/src/theory/strings/theory_strings_preprocess.cpp
+++ b/src/theory/strings/theory_strings_preprocess.cpp
@@ -22,6 +22,7 @@
 #include "options/strings_options.h"
 #include "proof/proof_manager.h"
 #include "smt/logic_exception.h"
+#include "theory/strings/theory_strings_rewriter.h"
 
 using namespace CVC4;
 using namespace CVC4::kind;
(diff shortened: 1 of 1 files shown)


src/theory/strings/theory_strings_preprocess.cpp
```diff
@@ -97,62 +102,63 @@ Node StringsPreprocess::simplify( Node t, std::vector< Node > &new_nodes ) {
...
 
     // substr( x, n, len( x ) - n )
-    Node st = nm->mkNode(
-        kind::STRING_SUBSTR,
-        t[0],
-        t[2],
-        nm->mkNode(kind::MINUS, nm->mkNode(kind::STRING_LENGTH, t[0]), t[2]));
-    Node io2 = nm->mkSkolem("io2", nm->stringType(), "created for indexof");
-    Node io4 = nm->mkSkolem("io4", nm->stringType(), "created for indexof");
+    Node st = nm->mkNode(STRING_SUBSTR,
```
4tXJ7f: Maybe we should do an optimization similar to the `STRING_SUBSTR` reduction above and not do introduce the `STRING_SUBSTR` here if `n == d_zero`.

ajreynol: I think this would be overoptimizing. This would only prevent us from skipping the rewrite substr(x,0,len(x)-0)---> x. Notice that the optimization for the reduction is different: we avoid introducing a skolem if either side of the remainder is empty, e.g. for substr( x, 0, y ) we avoid the prefix remainder and for substr( x, z, len(x) ) we avoid the suffix remainder.

4tXJ7f: You are right, the rewriter should take care of that.

PR #2494, opened 2018-09-18
Decision strategy: incorporate separation logic.

ajreynol:


src/theory/sep/theory_sep.cpp
```diff
@@ -454,12 +455,16 @@ void TheorySep::check(Effort e) {
             if( !use_polarity ){
               // introduce guard, assert positive version
               Trace("sep-lemma-debug") << "Negated spatial constraint asserted to sep theory: " << fact << std::endl;
-              Node lit = Rewriter::rewrite( NodeManager::currentNM()->mkSkolem( "G", NodeManager::currentNM()->booleanType() ) );
-              lit = getValuation().ensureLiteral( lit );
+              Node g = nm->mkSkolem("G", nm->booleanType());
+              d_neg_guard_strategy[g].reset(new DecisionStrategySingleton(
+                  "sep_neg_guard", g, getSatContext(), getValuation()));
+              DecisionStrategySingleton* ds = d_neg_guard_strategy[g].get();
```
HanielB: allocate the pointer and then set the `unique_ptr` map, so that you don't have to retrieve the pointer from the map.

ajreynol: I think that practice is generally frowned upon, it's best to set the unique_ptr directly so that at no point is there "dangling" memory.

PR #2495, opened 2018-09-18
Decision strategy: incorporate arrays.

ajreynol:


src/theory/arrays/theory_arrays.h
```diff
@@ -455,6 +453,27 @@ class TheoryArrays : public Theory {
   /** An equality-engine callback for proof reconstruction */
   ArrayProofReconstruction d_proofReconstruction;
 
+  /** The decision strategy for the theory of arrays */
```
HanielB: say how it operates

ajreynol: Unfortunately I don't know how it does (this is Clark's code).

barrettcw: If the "arrays-eager-index" option is enabled, then whenever a read-over-write lemma is generated, a decision request is also generated for the comparison between the indexes that appears in the lemma.

ajreynol: I'll add this.

PR #2496, opened 2018-09-19
Add rewrites for str.contains + str.replace/substr

4tXJ7f:
This commit adds four rewrites for str.contains combined with
str.replace and str.substr operators. All rewrites were verified with
CVC4 and Z3.

### DIFF
--- a/src/theory/strings/theory_strings_rewriter.cpp
+++ b/src/theory/strings/theory_strings_rewriter.cpp
@@ -1545,6 +1545,35 @@ Node TheoryStringsRewriter::rewriteContains( Node node ) {
           return returnRewrite(node, res, "ctn-rpl-non-ctn");
         }
       }
+
+      // (str.contains x (str.++ w (str.replace x y x) z)) --->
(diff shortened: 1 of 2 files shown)


4tXJ7f:
I ran `--sygus-rr-verify` for 1h on both the string term grammar as well as the predicate grammar without finding an unsound rewrite.

src/theory/strings/theory_strings_rewriter.cpp
```diff
@@ -1749,6 +1778,41 @@ Node TheoryStringsRewriter::rewriteContains( Node node ) {
...
+    if (node[0][2] == nm->mkNode(kind::STRING_LENGTH, node[1]))
+    {
+      Node ret = nm->mkNode(kind::EQUAL, node[0], node[1]);
+      return returnRewrite(node, ret, "ctn-substr");
+    }
+  }
+
+  if (node[1].getKind() == kind::STRING_STRREPL)
+  {
+    // (str.contains x (str.replace y x y)) --->
```
ajreynol: This one is fun :)

PR #2499, opened 2018-09-19
Allow partial models for multiple sygus enumerators

ajreynol:
This improves the fairness scheme for sygus strategies that use multiple enumerators by allowing "partial models" in sygus modules, that is, processing of models where *some* enumerators have model values that have been excluded by symmetry breaking.


src/theory/quantifiers/sygus/synth_conjecture.cpp
```diff
@@ -340,6 +341,19 @@ void SynthConjecture::doCheck(std::vector<Node>& lems)
   std::vector<Node> enum_values;
   getModelValues(terms, enum_values);
 
+  // if the master requires a full model and the model is partial, we fail
+  if (!d_master->allowPartialModel())
+  {
+    for (const Node& v : enum_values)
```
HanielB: instead of checking this here gain why not have `getModelValues` returning `true` iff all terms have non-null model values? Then this block would be only `if (!d_master->allowPartialModel() && !resultGetModelValues)`

PR #2505, opened 2018-09-20
carefully printing trusted assertions in proofs

yoni206:
This PR fixes a bug that occurs when using `--fewer-preprocessing-holes` with `test/regress/regress0/bug217.smt2`.
The generated proof was incorrect, resulting in a type-checking error when using `--check-proofs`.
The problem was that boolean variables were not converted to formulas when `--fewer-preprocessing-holes` was set (although they were converted to formulas when this option was off).
The new function `printTrustedTerm` was added in order to avoid duplicating this conversion.


src/proof/proof_manager.cpp
```diff
@@ -1063,5 +1060,15 @@ void ProofManager::printGlobalLetMap(std::set<Node>& atoms,
 void ProofManager::ensureLiteral(Node node) {
   d_cnfProof->ensureLiteral(node);
 }
-
+void ProofManager::printTrustedTerm(Node term,
+                                    std::ostream& os,
+                                    ProofLetMap& globalLetMap)
+{
+  if (ProofManager::currentPM()->getTheoryProofEngine()->printsAsBool(term))
```
4tXJ7f: Suggestion: Instead of calling `ProofManager::currentPM()` multiple times, maybe just call it in the beginning of the function and store it in a `ProofManager*`? Similar to what we usually do with our `NodeManager`.

src/proof/proof_manager.h
```diff
@@ -239,6 +239,12 @@ class ProofManager {
   // for SMT variable names that have spaces and other things
   static std::string sanitize(TNode var);
 
+  // wrap term with (p_app ... ) if needed and print
```
4tXJ7f: Maybe instead of "if needed" just say that "if the term is printed as a boolean".

PR #2507, opened 2018-09-20
Add skeleton-preprocessing pass (which replaces the non-clausal simplication pass) when triggered

anwu1219:


src/preprocessing/passes/non_clausal_simp_sat.h
```diff
@@ -0,0 +1,46 @@
...
+
+#include "expr/node.h"
+#include "preprocessing/preprocessing_pass.h"
+#include "preprocessing/preprocessing_pass_context.h"
+
+namespace CVC4 {
+namespace preprocessing {
+namespace passes {
+
+class NonClausalSimpSAT : public PreprocessingPass
```
4tXJ7f: Please rename the pass to match the option.

src/preprocessing/passes/non_clausal_simp_sat.cpp
```diff
@@ -0,0 +1,483 @@
...
+ #include "prop/cnf_stream.h"
+ #include "prop/sat_solver_factory.h"
+
+ #include "proof/proof_manager.h"
+
+ #include "context/cdo.h"
+ #include "options/proof_options.h"
+ #include "smt/smt_statistics_registry.h"
+ #include "theory/theory_model.h"
+#include <ctime>
```
4tXJ7f: Please follow the [styleguide for includes](https://github.com/CVC4/CVC4/wiki/Code-Style-Guidelines#names-and-order-of-includes) (also applies to the other changes in the includes). Also, what is `<ctime>` used for?

src/preprocessing/passes/non_clausal_simp_sat.cpp
```diff
@@ -0,0 +1,483 @@
...
+ ** All rights reserved.  See the file COPYING in the top-level source
+ ** directory for licensing information.\endverbatim
+ **
+ ** \brief Non-clausal simplification preprocessing pass.
+ **
+ ** Run the nonclausal solver and try to solve all assigned theory literals.
+ **/
+
+ #include "preprocessing/passes/non_clausal_simp_sat.h"
+ #include <cryptominisat5/cryptominisat.h>
```
4tXJ7f: The implementation needs to be protected by an `#ifdef CVC4_USE_CRYPTOMINISAT` otherwise compilation will fail when CVC4 is not configured with `--with-cryptominisat` because the headers are not available. In the case where we do not compile with CVC4, we should emit an error if the user tries to use the option `--skeleton-preprocessing`.

src/preprocessing/passes/non_clausal_simp_sat.cpp
```diff
@@ -0,0 +1,483 @@
...
+           << "asserting " << (*assertionsToPreprocess)[i] << std::endl;
+       d_cnfStream.convertAndAssert((*assertionsToPreprocess)[i], false, false,
+                                    RULE_GIVEN);
+  }
+
+
+
+  SatValue result = d_satSolver->solve();
+
+  if (result==SAT_VALUE_FALSE){
```
4tXJ7f: Apply clang-format as described [here](https://github.com/CVC4/CVC4/wiki/Developer-Guide#formatting-with-clang-format).

src/preprocessing/passes/non_clausal_simp_sat.cpp
```diff
@@ -0,0 +1,483 @@
...
+    // If in conflict, just return false
+    Trace("non-clausal-simplify-sat")
+        << "conflict in non-clausal propagation" << std::endl;
+    Assert(!options::unsatCores() && !options::fewerPreprocessingHoles());
+    assertionsToPreprocess->clear();
+    Node n = NodeManager::currentNM()->mkConst<bool>(false);
+    assertionsToPreprocess->push_back(n);
+    PROOF(ProofManager::currentPM()->addDependence(n, Node::null()));
+    delete d_satSolver;
```
4tXJ7f: Take a look at [`std::unique_ptr`](https://en.cppreference.com/w/cpp/memory/unique_ptr) to for a better way of managing memory without explicit `delete`s.

src/Makefile.am
```diff
@@ -93,6 +93,8 @@ libcvc4_la_SOURCES = \
 	preprocessing/passes/ite_simp.h \
 	preprocessing/passes/non_clausal_simp.cpp \
 	preprocessing/passes/non_clausal_simp.h \
+        preprocessing/passes/non_clausal_simp_sat.cpp \
```
4tXJ7f: Please fix the alignment.

aniemetz:
Reopen as work-in-progress PR.

src/preprocessing/passes/skeleton_preprocessing.cpp
```diff
@@ -0,0 +1,483 @@
...
+{
+
+  Assert(!options::unsatCores() && !options::fewerPreprocessingHoles());
+
+  d_preprocContext->spendResource(options::preprocessStep());
+
+
+  // Convert the original formula into a boolean formula and solve it
+  // with cryptominisat
+  SatSolver* d_satSolver = SatSolverFactory::createCryptoMinisat(smtStatisticsRegistry(), "non_clausal_simp_solver");
```
mpreiner: Since `d_satSolver` is local here, you don't have to prefix it with `d_`. Variables should be prefixed with `d_` if they are data members of classes (https://github.com/CVC4/CVC4/wiki/Code-Style-Guidelines#naming-1).

Further, use a `std::unique_ptr` so that you don't have to explicitly call delete.

src/preprocessing/passes/skeleton_preprocessing.cpp
```diff
@@ -0,0 +1,483 @@
...
+                                            "toCNF-skeleton-preprocessing");
+
+  Trace("skeleton-preprocessing") << "asserting to sat solver" << std::endl;
+
+  unsigned substs_index = d_preprocContext->getSubstitutionsIndex();
+  for (size_t i = 0, size = assertionsToPreprocess->size(); i < size; ++i)
+  {
+    Assert(Rewriter::rewrite((*assertionsToPreprocess)[i])
+              == (*assertionsToPreprocess)[i]);
+       // Don't reprocess substitutions
```
mpreiner: Document why we don't want to do this.

src/preprocessing/passes/skeleton_preprocessing.cpp
```diff
@@ -0,0 +1,483 @@
...
+
+
+
+  SatValue result = d_satSolver->solve();
+
+  if (result==SAT_VALUE_FALSE){
+    // If in conflict, just return false
+    Trace("skeleton-preprocessing")
+        << "conflict in non-clausal propagation" << std::endl;
+    Assert(!options::unsatCores() && !options::fewerPreprocessingHoles());
```
mpreiner: Remove assertion since this was already asserted at the very beginning of this function and should not have changed.

src/preprocessing/passes/skeleton_preprocessing.cpp
```diff
@@ -0,0 +1,483 @@
...
+    delete d_satSolver;
+    return PreprocessingPassResult::CONFLICT;
+  }
+
+  std::vector<SatLiteral> topLevelUnits = d_satSolver->getTopLevelUnits();
+
+  std::vector<TNode> learned_literals;
+
+
+  for (size_t i = 0; i < topLevelUnits.size(); i++){
```
mpreiner: Cache size `for (size_t i = 0, size = topLevelUnits.size(); i < size; ++i) ...`.

src/preprocessing/passes/skeleton_preprocessing.cpp
```diff
@@ -0,0 +1,483 @@
...
+    SatLiteral lit = topLevelUnits[i];
+    if (d_cnfStream.getNodeCache().find(lit)
```
mpreiner: Store node cache in a separate variable outside of loop.

src/preprocessing/passes/non_clausal_simp.cpp
```diff
@@ -57,68 +80,37 @@ PreprocessingPassResult NonClausalSimp::applyInternal(
...
-    // If in conflict, just return false
-    Trace("non-clausal-simplify")
-        << "conflict in non-clausal propagation" << std::endl;
-    Assert(!options::unsatCores() && !options::fewerPreprocessingHoles());
-    assertionsToPreprocess->clear();
-    Node n = NodeManager::currentNM()->mkConst<bool>(false);
-    assertionsToPreprocess->push_back(n);
-    PROOF(ProofManager::currentPM()->addDependence(n, Node::null()));
-    propagator->setNeedsFinish(true);
+    handleConflictCase(assertionsToPreprocess);
```
4tXJ7f: It looks like you call `handleConflictCase` twice when `propagator->propagate()` returns true, which does not happen in the original code. You may be able to call `handleConflictCase` only once.

src/preprocessing/passes/non_clausal_simp.cpp
```diff
@@ -448,10 +438,154 @@ PreprocessingPassResult NonClausalSimp::applyInternal(
...
+SatValue NonClausalSimp::solveByCryptominisat(
+    CVC4::prop::TseitinCnfStream* d_cnfStream,
+    AssertionPipeline* assertionsToPreprocess,
+    unsigned substs_index)
+{
+  for (size_t i = 0, size = assertionsToPreprocess->size(); i < size; ++i)
+  {
+    Trace("non-clausal-simplify") << "Assertion #" << i << " : "
+                                  << (*assertionsToPreprocess)[i] << std::endl;
+  }
```
4tXJ7f: It looks like this tracing stuff can be moved from the two `solve_*` methods to `applyInternal` because it just prints out the assertions that we are preprocessing.

src/preprocessing/passes/non_clausal_simp.cpp
```diff
@@ -448,10 +438,154 @@ PreprocessingPassResult NonClausalSimp::applyInternal(
...
+    // If in conflict, just return false
+    return std::make_tuple(false, learned_literals);
+  }
+
+  std::vector<SatLiteral> topLevelUnits = d_satSolver->getTopLevelUnits();
+  for (size_t i = 0; i < topLevelUnits.size(); i++)
+  {
+    SatLiteral lit = topLevelUnits[i];
+    if (d_cnfStream.getNodeCache().find(lit)
+        != d_cnfStream.getNodeCache().end())
```
4tXJ7f: What does it mean if the literal is not in the node cache? Why is it safe to just skip those literals? Please add a comment.

src/preprocessing/passes/non_clausal_simp.cpp
```diff
@@ -448,10 +438,154 @@ PreprocessingPassResult NonClausalSimp::applyInternal(
...
+    {
+      // if the literal is in the getNodeCache
+      Node learnedLiteral = Rewriter::rewrite(d_cnfStream.getNode(lit));
```
4tXJ7f: If we are sending rewritten nodes to the CnfStream, I would expect that we only have rewritten subterms in CnfStream and the rewrite is a noop. Am I missing something?

src/preprocessing/passes/non_clausal_simp.cpp
```diff
@@ -448,10 +438,154 @@ PreprocessingPassResult NonClausalSimp::applyInternal(
...
+      if (learnedLiteral.isConst())
```
4tXJ7f: It looks like we are making this distinction already here: https://github.com/CVC4/CVC4/pull/2507/files#diff-8d28c2a099c57ce8aee92960c4ff2db4R144 . Why do it twice?

src/preprocessing/passes/non_clausal_simp.cpp
```diff
@@ -448,10 +438,154 @@ PreprocessingPassResult NonClausalSimp::applyInternal(
...
+                     << std::endl;
+      propagator->assertTrue((*assertionsToPreprocess)[i]);
+    }
+  } // end timing
+
+  Trace("non-clausal-simplify") << "propagating" << std::endl;
+  return propagator->propagate();
+}
+
+std::tuple<bool, std::vector<Node>> NonClausalSimp::preprocessByCircuitPropagator(
```
4tXJ7f: It looks like these `preprocess*` methods are not doing that much (especially if we simplify the rest according to my other comments), so maybe they are not necessary and can just be merged with the `solve*` methods.

src/preprocessing/passes/non_clausal_simp.cpp
```diff
@@ -57,68 +80,37 @@ PreprocessingPassResult NonClausalSimp::applyInternal(
...
-    propagator->finish();
-    propagator->setNeedsFinish(false);
-  }
-  propagator->initialize();
-
-  // Assert all the assertions to the propagator
-  Trace("non-clausal-simplify") << "asserting to propagator" << std::endl;
   unsigned substs_index = d_preprocContext->getSubstitutionsIndex();
-  for (size_t i = 0, size = assertionsToPreprocess->size(); i < size; ++i)
+  std::tuple<bool, std::vector<Node>> resultAndLearnedLiterals;
```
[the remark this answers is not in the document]
anwu1219: I see. std::pair does seem simpler.

src/preprocessing/passes/non_clausal_simp.cpp
anwu1219: These methods are responsible for getting the learned clauses. Do you mean putting the solveBy* method in the preprocessing method?

4tXJ7f: Yeah, exactly. If they are not really doing anything besides calling the `solveBy*` methods, they can be inlined.

src/preprocessing/passes/non_clausal_simp.cpp
anwu1219: I think you are right. Removing this does not make a difference.

src/preprocessing/passes/non_clausal_simp.cpp
4tXJ7f: Right, that's what the if statement says :) However, I think that we need a deeper understanding of what is going on here. The node cache seems to be the literal to node map that CnfStream maintains. Now, the question is: are we expecting every literal that we get to be in that map? If yes, this should become an assertion. Are we expecting some literals to be missing from the map? If yes, we should document why and/or find out if that has a negative impact on our performance (since we are ignoring some learned literals in that case).

src/preprocessing/passes/non_clausal_simp.cpp
anwu1219: I think in the current way, if we detect a false constant node, we can immediately return.

4tXJ7f: It is not clear to me that this is possible. Assuming that the `learnedLiteral`s are already in rewritten form (which seems to be the case according to your other comment), then if a literal was false, we would assert them as such to the CnfStream and the SAT solver would return `SAT_VALUE_FALSE` if it determined that this literal is actually a unit literal and must be true.

src/preprocessing/passes/non_clausal_simp.cpp
anwu1219: I think we should expect every literal that we get to be in that map. getNodeCache() returns the "d_literalToNodeMap" of the cnfStream we created. 
We set the fullLitToNodeMap to true when creating the cnfStream. Therefore, when a new literal is created, whether it's a theory literal or not, it will be added to the d_literalToNodeMap and the SAT-solver.
https://github.com/CVC4/CVC4/blob/8b8e40985b83354e7e4a503d217374eb7fbc45f0/src/prop/cnf_stream.cpp#L197-L200

anwu1219: I believe we should make it an assertion.

anwu1219: Actually, one small twist. 
The SAT_solver always returns "True" and "Not False" (which correspond to satliteral 0 and ~1), in addition to the meaningful top level units. 
So I think I should start from index two when traversing the topLevelUnits

anwu1219: Also should I have the assertion there? It seems a bit expensive. Doing that for each top level units is "n Log (n)"

src/preprocessing/passes/non_clausal_simp.cpp
anwu1219: I just merged the preprocess and solve methods to one because neither of them is that long.

src/preprocessing/passes/non_clausal_simp.cpp
anwu1219: Yeah, that makes sense. It's impossible for the SAT solver to have False or Not True in the learned literal.

src/preprocessing/passes/non_clausal_simp.cpp
mpreiner: I would not start from index two, I would explicitly skip true/false lits. You can use the util function FindOrDie, to either get the element of a map or fail: https://github.com/CVC4/CVC4/blob/987df3df987768e2ce0c36d17469929f8e92fdec/src/base/map_util.h#L88

src/preprocessing/passes/non_clausal_simp.cpp
```diff
@@ -448,10 +444,105 @@ PreprocessingPassResult NonClausalSimp::applyInternal(
...
+    return std::make_pair(false, learned_literals);
+  }
+
+  std::vector<SatLiteral> topLevelUnits = d_satSolver->getTopLevelUnits();
+  // we start from index 2 because the first two top level units much be True
+  // and Not False
+  for (size_t i = 2; i < topLevelUnits.size(); i++)
+  {
+    SatLiteral lit = topLevelUnits[i];
+    Node learnedLiteral = d_cnfStream.getNode(lit);
```
mpreiner: Is it guaranteed that all literals returned by CryptoMiniSat have a corresponding node in CVC4?  CryptoMiniSat supports bounded variable addition as a simplification technique, which adds new variables. CryptoMiniSat may potentially return literals that we don't have mapped to nodes.

src/preprocessing/passes/non_clausal_simp.cpp
4tXJ7f: I agree with @mpreiner, starting from index two is brittle because the order may be different for different SAT solvers.

> Also should I have the assertion that the topLevelUnit must be in the map there? It seems a bit expensive. Doing that for each top level units is "n Log (n)"

Assertions are only checked when compiling a debug version of CVC4 (or with assertions explicitly enabled), so we can and should use them liberally. Assertions document intent and they make CVC4 fail closer to an actual problem instead of later down the road. Also, note that `d_literalToNodeMap` is a hash map, so you'd expect a search to be on the order of O(1) (on average). Additionally, the check can be done without a second search, check out the implementation of `FindOrDie()`.

Could you also add the explanation for why we expect it to be in the cache as a comment in the source code?

src/preprocessing/passes/non_clausal_simp.h
```diff
@@ -48,6 +49,15 @@ class NonClausalSimp : public PreprocessingPass
 
   /** Learned literals */
   std::vector<Node> d_nonClausalLearnedLiterals;
+
+  /** Preprocess the boolean skeleton,
+  which calls solveByCryptominisat or solveByCircuitPropagator */
```
4tXJ7f: This comment is not true anymore.

src/prop/cnf_stream.h
```diff
@@ -211,6 +211,8 @@ class CnfStream {
                                 ProofRule proof_id,
                                 TNode from = TNode::null()) = 0;
 
+  inline SatSolver* getSatSolver() { return d_satSolver; };
```
4tXJ7f: I doesn't look like this is being used. Am I missing something?

src/preprocessing/passes/non_clausal_simp.h
```diff
@@ -48,6 +49,15 @@ class NonClausalSimp : public PreprocessingPass
...
+  std::pair<bool, std::vector<Node>> preprocessByCryptominisat(
+      AssertionPipeline* assertionsToPreprocess, unsigned substs_index);
+
+  std::pair<bool, std::vector<Node>> preprocessByCircuitPropagator(
```
4tXJ7f: Add a comment for this method (also describe the arguments it expects and what it returns).

src/prop/cryptominisat.cpp
```diff
@@ -190,7 +205,15 @@ SatValue CryptoMinisatSolver::modelValue(SatLiteral l){
 
 unsigned CryptoMinisatSolver::getAssertionLevel() const {
   Unreachable("No interface to get assertion level in Cryptominisat");
-  return -1; 
+  return -1;
```
mpreiner: The return is not needed due to `Unreachable(...)`.

src/preprocessing/passes/non_clausal_simp.cpp
```diff
@@ -16,32 +16,55 @@
...
-#include <vector>
 
 using namespace CVC4;
 using namespace CVC4::theory;
 
 namespace CVC4 {
 namespace preprocessing {
+
+namespace {
+void handleConflictCase(AssertionPipeline* assertionsToPreprocess)
```
4tXJ7f: I would actually inline this method again since it is now called from a single place and not doing anything complicated but this is a matter of taste to a certain extent.

src/preprocessing/passes/non_clausal_simp.cpp
```diff
@@ -16,32 +16,55 @@
...
 namespace passes {
 
 /* -------------------------------------------------------------------------- */
 
 NonClausalSimp::Statistics::Statistics()
     : d_numConstantProps(
-          "preprocessing::passes::NonClausalSimp::NumConstantProps", 0)
+          "preprocessing::passes::NonClausalSimp::NumConstantProps", 0),
+      d_addingAssertions(
+          "preprocessing::passes::NonClausalSimp::TimeAddingAssertions")
```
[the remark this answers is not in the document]
anwu1219: Sounds good. I'll change it to CnfTranslateTime

src/prop/cryptominisat.cpp
```diff
@@ -190,7 +205,18 @@ SatValue CryptoMinisatSolver::modelValue(SatLiteral l){
 
 unsigned CryptoMinisatSolver::getAssertionLevel() const {
   Unreachable("No interface to get assertion level in Cryptominisat");
-  return -1; 
+}
+
+std::vector<SatLiteral> CryptoMinisatSolver::getTopLevelUnits(){
+  std::vector<SatLiteral> satLits;
+  for (const auto& lit : d_solver->get_zero_assigned_lits()){
+    if (lit.var() != d_false || lit.var() != d_true)
```
mpreiner: This should be `lit.var() != d_false && lit.var() != d_true`.

src/preprocessing/passes/non_clausal_simp.h
```diff
@@ -48,6 +49,40 @@ class NonClausalSimp : public PreprocessingPass
...
+      - substs_index: the substitution index in the current context
+      Return:
+      whether the boolean skeleton is not satisfiable and the
+      learned unit clauses
+   */
+  std::pair<bool, std::vector<Node>> preprocessByCircuitPropagator(
+      AssertionPipeline* assertionsToPreprocess, unsigned substs_index);
+
+  /** Handle the case that a conflict in the boolean skeleton is detected */
+  inline void handleConflictCase(AssertionPipeline* assertionsToPreprocess)
```
mpreiner: I would move that into the .cpp file. No need to use inline either, let the compiler decide how to optimize.

src/preprocessing/passes/non_clausal_simp.cpp
```diff
@@ -448,10 +431,106 @@ PreprocessingPassResult NonClausalSimp::applyInternal(
...
+      smtStatisticsRegistry(), "non_clausal_simp_solver"));
+  CVC4::prop::NullRegistrar registrar;
+  context::Context context;
+  CVC4::prop::TseitinCnfStream cnfStream(satSolver.get(),
+                                         &registrar,
+                                         &context,
+                                         true,
+                                         "toCNF-skeleton-preprocessing");
+  std::vector<Node> learned_literals;
+  Trace("skeleton-preprocessing") << "asserting to sat solver" << std::endl;
```
4tXJ7f: Right now, the trace tags are inconsistent. I'd just use `non-clausal-simplify` for everything since it is all in the same file.

src/preprocessing/passes/non_clausal_simp.h
```diff
@@ -48,6 +49,29 @@ class NonClausalSimp : public PreprocessingPass
...
+      - substs_index: the substitution index in the current context
+      Return:
+      whether the boolean skeleton is not satisfiable and the
+      learned unit clauses
+   */
+  std::pair<bool, std::vector<Node>> preprocessByCryptoMinisat(
+      AssertionPipeline* assertionsToPreprocess, unsigned substs_index);
+
+  /** Preprocess the boolean skeleton using CryptoMinisat
+      Args:
```
4tXJ7f: Use [doxygen](http://www.stack.nl/~dimitri/doxygen/)-style comments for describing arguments and return values.

src/prop/sat_solver.h
```diff
@@ -98,7 +101,13 @@ class SatSolver {
   virtual bool ok() const = 0;
   
   virtual void setProofLog( BitVectorProof * bvp ) {}
-  
+
+  /** Get the top level unit clauses of a formula **/
```
4tXJ7f: Please describe the contract in greater detail. Specifically, describe what kind of literals we can get back (only the ones added by the user? Different ones?)

src/smt/smt_engine.cpp
```diff
@@ -2185,6 +2185,25 @@ void SmtEngine::setDefaults() {
...
+          "Cannot use MipLibTrick when using cryptominisat instead of circuit"
+          "propagators. Try turn off --skeleton-preprocessing"));
+    }
+    else if (options::skeletonPreprocessing.wasSetByUser()){
+          setOption("skeleton-preprocessing", false);
+    }
+    else if (options::arithMLTrick.wasSetByUser()){
+          setOption("miplib-trick", false);
+    }
+  }
```
4tXJ7f: Also add a check that complains when `options::skeletonPreprocessing()` is set but [`isBuiltWithCryptominisat()`](https://github.com/CVC4/CVC4/blob/master/src/base/configuration.h#L96) returns false.

src/preprocessing/passes/non_clausal_simp.h
```diff
@@ -48,6 +49,29 @@ class NonClausalSimp : public PreprocessingPass
 
   /** Learned literals */
   std::vector<Node> d_nonClausalLearnedLiterals;
+
+
+  /**
+   * Preprocess the boolean skeleton using CryptoMinisat.
```
mpreiner: CryptoMiniSat

src/preprocessing/passes/non_clausal_simp.cpp
```diff
@@ -448,10 +432,105 @@ PreprocessingPassResult NonClausalSimp::applyInternal(
...
+      // Don't reprocess substitutions
+      if (substs_index > 0 && i == substs_index)
+      {
+        continue;
+      }
+      Trace("non-clausal-simplify")
+          << "asserting " << (*assertionsToPreprocess)[i] << std::endl;
+      cnfStream.convertAndAssert(
+          (*assertionsToPreprocess)[i], false, false, RULE_GIVEN);
+    }
```
mpreiner: You are using `(*assertionsToPreprocess)[i]` 4 times in this loop. Maybe store it in a separate variable?

src/preprocessing/passes/non_clausal_simp.cpp
```diff
@@ -448,10 +432,105 @@ PreprocessingPassResult NonClausalSimp::applyInternal(
...
+  } // end timing
+  Trace("non-clausal-simplify") << "solving" << std::endl;
+  if (satSolver->solve() == SAT_VALUE_FALSE)
+  {
+    // If in conflict, just return false
+    return std::make_pair(false, learned_literals);
+  }
+
+  std::vector<SatLiteral> topLevelUnits = satSolver->getTopLevelUnits();
+  for (size_t i = 0; i < topLevelUnits.size(); i++)
```
mpreiner: Use for each loop since you don't need `i` here.

src/preprocessing/passes/non_clausal_simp.cpp
```diff
@@ -448,10 +432,105 @@ PreprocessingPassResult NonClausalSimp::applyInternal(
...
+  {
+    propagator->finish();
+    propagator->setNeedsFinish(false);
+  }
+  propagator->initialize();
+
+  // Assert all the assertions to the propagator
+  Trace("non-clausal-simplify") << "asserting to propagator" << std::endl;
+  {  // for timing
+    TimerStat::CodeTimer timer(d_statistics.d_cnfTranslateTime);
```
mpreiner: If you also use `d_statistics.d_cnfTranslateTime` for the circuit propagator the name of the statistic is not appropriate since the circuit propagator does not do any CNF translation. In both cases you add assertions maybe call it `addAssertionTime` then?

src/prop/cnf_stream.h
mpreiner: Unresolved, since still not removed.

src/preprocessing/passes/non_clausal_simp.cpp
```diff
@@ -57,68 +67,44 @@ PreprocessingPassResult NonClausalSimp::applyInternal(
...
+    std::tie(noconflict, learned_literals) =
+        preprocessByCircuitPropagator(assertionsToPreprocess, substs_index);
   }
-  propagator->initialize();
 
-  // Assert all the assertions to the propagator
-  Trace("non-clausal-simplify") << "asserting to propagator" << std::endl;
-  unsigned substs_index = d_preprocContext->getSubstitutionsIndex();
-  for (size_t i = 0, size = assertionsToPreprocess->size(); i < size; ++i)
+  if (!noconflict)
```
mpreiner: How about renaming `noconflict` to `ok`?

anwu1219: sounds good!

mpreiner:

Issue #1190, opened 2017-10-03
SyGuS invariant synthesis parsing is not robust to variable ordering


ajreynol:
This benchmark does not parse, since declare-primed-var are not in order with the arguments in the pre,post,trans:

(set-option :print-success true)
(set-logic LIA)
(synth-inv inv-f ((x Int) (y Int) (b Bool)))
(declare-primed-var b Bool)
(declare-primed-var x Int)
(declare-primed-var y Int)
(define-fun pre-f ((x Int) (y Int) (b Bool)) Bool
  (and 
    (and (>= x 5) (<= x 9))
    (and (>= y 1) (<= y 3))
  )
)
(define-fun trans-f ((x Int) (y Int) (b Bool) (x! Int) (y! Int) (b! Bool)) Bool
  (and 
    (and (= b! b) (= y! x))
    (ite b (= x! (+ x 10)) (= x! (+ x 12)))
  )
)
(define-fun post-f ((x Int) (y Int) (b Bool)) Bool true)
(inv-constraint inv-f pre-f trans-f post-f)
(check-synth)

ajreynol:
It is questionable whether benchmarks like this should be accepted.  The SyGuS format technically allows it, but the format is a bit nonsensical, since it implicitly considers how variables are shadowed, i.e. (x Int) in pre-f corresponds to the global (declare-primed-var x Int).
There isn't an obvious/clean solution to this that I can see at the moment.

HanielB:
Closed via #2509

PR #2509, opened 2018-09-21
Makes SyGuS parsing more robust in invariant problems

HanielB:
Before CVC4 could only parse invariant problems if the order of `declare-primed-var` was the same as the order of the arguments in the invariant predicate declaration.

Fixes issue #1190


src/parser/smt2/Smt2.g
```diff
@@ -724,31 +724,39 @@ sygusCommand [std::unique_ptr<CVC4::Command>* cmd]
...
-        if( i==2 ){
-          children.insert( children.end(), all.begin(), all.end() );
-        }else{
-          children.insert( children.end(), primed[0].begin(), primed[0].end() );
+        Debug("parser-sygus")
+            << "Make inv-constraint term #" << i << " : " << op << " with type "
+            << op.getType() << "..." << std::endl;
+        std::vector<Expr> children;
+        children.push_back(op);
+        // wether transition relation
```
ajreynol: Comment incomplete?

HanielB: fixed

HanielB:
Yeah, looking at it again indeed they are unnecessary. The format says:

"The number and sorts of the arguments of the pre-condition function and the post-condition function
should be the same as that of the invariant to synthesize (not all arguments must be referred to in the
function definition). Since the function for the transitions needs to refer to both the primed and unprimed
versions of the variables to be synthesized, its list of arguments should be a concatenation of two list of
arguments for the given invariant."

So all the information necessary for building the constraint is available from the previous declarations. Since on the output we have to talk about the variables as they are declared in `synth-inv` anyway, I think we should either ignore `declare-primed-var` or use it only to check that the number of declared variables is the proper one.

What do you think?

ajreynol:
I think we should not rely on declare-primed-var for anything. The synth-inv function should only rely on the argument types on the functions in its arguments.

If we wanted to use the declare-primed-var command as a sanity check, notice its not straightforward: I think we'd be testing (if anything) equality between the multiset of types of arguments of the synth-fun and the declare-primed-var.

HanielB:
> I think we should not rely on declare-primed-var for anything. The validity of the synth-inv command should only rely on the argument types on the functions in its arguments only.

Agreed. 

> If we wanted to use the declare-primed-var command as a sanity check, notice its not straightforward: I think we'd be testing (if anything) equality between the multiset of types of arguments of the synth-fun and the declare-primed-var.

I also think that's what we should be testing, if anything. But since the standard does not require at all that this is checked we should only give a warning that not all variables have been declared.

src/parser/smt2/smt2.h
```diff
@@ -68,17 +68,17 @@ class Smt2 : public Parser {
   std::unordered_map<std::string, Kind> operatorKindMap;
   std::pair<Expr, std::string> d_lastNamedTerm;
   // for sygus
-  std::vector<Expr> d_sygusVars, d_sygusInvVars, d_sygusConstraints,
-      d_sygusFunSymbols;
-  std::map< Expr, bool > d_sygusVarPrimed;
+  std::vector<Expr> d_sygusVars, d_sygusConstraints, d_sygusFunSymbols;
+  /** maps sygus variables to their respective primed variabels */
```
ajreynol: variables.

src/parser/smt2/smt2.cpp
```diff
@@ -1235,14 +1232,21 @@ Expr Smt2::makeSygusBoundVarList(Datatype& dt,
...
+{
+  std::vector<Type> argTypes = t.getArgTypes();
+  for (const Type& ti : argTypes)
   {
-    Expr v = d_sygusInvVars[i];
-    std::map< Expr, bool >::iterator it = d_sygusVarPrimed.find( v );
-    if( it!=d_sygusVarPrimed.end() ){
-      if( it->second==isPrimed ){
-        vars.push_back( v );
+    for (const std::pair<const Expr, Expr>& p : d_sygusVarPrimed)
```
ajreynol: I think concretely this is the part of the code that is unecessary. We are still relying on the d_sygusVarPrimed here, whereas we could just create fresh variables of type ti here instead.  I would update the code to do this.

HanielB:
I have removed the dependency of the command for solving, but we do send out warnings in debug when there is a discrepancy between the invariant type and the declared primed variables. PTAL

HanielB:
Fixed.

src/parser/smt2/smt2.cpp
```diff
@@ -1235,16 +1233,45 @@ Expr Smt2::makeSygusBoundVarList(Datatype& dt,
...
-    std::map< Expr, bool >::iterator it = d_sygusVarPrimed.find( v );
-    if( it!=d_sygusVarPrimed.end() ){
-      if( it->second==isPrimed ){
-        vars.push_back( v );
+    if (type_count.find(ti) == type_count.end())
+    {
+      type_count[ti] = 0;
+    }
+    std::stringstream ss;
+    ss << ti << "_" << type_count[ti]++;
```
ajreynol: This is a pretty nasty name for this variable. I would leave these bound variables have the default names actually.

ajreynol: Also more generally, notice that name clashes do not need to be resolved by the caller of mkBoundVar, mkSkolem, since we automatically append _N for some internal N, so the type_count is not strictly necessary in these kind of cases.

src/parser/parser.h
```diff
@@ -476,6 +476,8 @@ class CVC4_PUBLIC Parser {
 
   /** Create a new CVC4 bound variable expression of the given type. */
   Expr mkBoundVar(const std::string& name, const Type& type);
+  /** as above, but using internal naming scheme */
+  Expr mkBoundVar(const Type& type);
```
ajreynol: Eliminate this function. It is not necessary to define the bound variable in the parser state since it is not being parsed.

src/parser/smt2/smt2.cpp
```diff
@@ -1235,16 +1233,39 @@ Expr Smt2::makeSygusBoundVarList(Datatype& dt,
...
+{
+  std::vector<Type> argTypes = t.getArgTypes();
+  for (const Type& ti : argTypes)
   {
-    Expr v = d_sygusInvVars[i];
-    std::map< Expr, bool >::iterator it = d_sygusVarPrimed.find( v );
-    if( it!=d_sygusVarPrimed.end() ){
-      if( it->second==isPrimed ){
-        vars.push_back( v );
+    vars.push_back(mkBoundVar(ti));
```
ajreynol: This should just be EXPR_MANAGER->mkBoundVar.

PR #2511, opened 2018-09-21
Infrastructure for variable agnostic sygus enumerators

ajreynol:
This is in preparation for a new class of symmetry breaking techniques on sygus enumeration.

It also does minor cleanup to enumerator registration.


src/theory/quantifiers/sygus/term_database_sygus.cpp
```diff
@@ -515,6 +560,33 @@ void TermDbSygus::registerEnumerator(Node e,
     }
   }
   Trace("sygus-db") << "  ...finished" << std::endl;
+
+  d_enum_var_agnostic[e] = isVarAgnostic;
+  if (isVarAgnostic)
+  {
+    // if not done so already, compute type class identifiers for each variable
```
HanielB: where else could that have been computed?

src/theory/quantifiers/sygus/term_database_sygus.cpp
```diff
@@ -881,6 +963,78 @@ bool TermDbSygus::hasSubtermSymbolicCons(TypeNode tn) const
...
+  return itcc->second;
+}
+
+unsigned TermDbSygus::getNumSubclassVars(TypeNode tn, unsigned sc) const
+{
+  std::map<TypeNode, std::map<unsigned, std::vector<Node> > >::const_iterator
+      itv = d_var_subclass_list.find(tn);
+  if (itv == d_var_subclass_list.end())
+  {
+    return 0;
```
HanielB: shouldn't you have the `Assert(false)` cases here as well?

src/theory/quantifiers/sygus/term_database_sygus.cpp
```diff
@@ -881,6 +963,78 @@ bool TermDbSygus::hasSubtermSymbolicCons(TypeNode tn) const
...
+}
+Node TermDbSygus::getVarSubclassIndex(TypeNode tn,
+                                      unsigned sc,
+                                      unsigned i) const
+{
+  std::map<TypeNode, std::map<unsigned, std::vector<Node> > >::const_iterator
+      itv = d_var_subclass_list.find(tn);
+  if (itv == d_var_subclass_list.end())
+  {
+    return Node::null();
```
HanielB: ditto

src/theory/quantifiers/sygus/term_database_sygus.cpp
ajreynol: If this is the second enumerator of this type, we will have already computed this information.

PR #2512, opened 2018-09-22
Unify rewrites related to (str.contains x y) --> (= x y)

4tXJ7f:
This commit unifies rewrites that turn `(str.contains x y)` into a
conjunction of equalities using length-based reasoning. It introduces a
generic helper function for determining which `ys` from an inequality
`y1 + ...  + yn >= x` may be 0 for it to still hold. When we have
`(str.contains x y)` and we can show that `str.len(y) >= str.len(x)`
then we know that `(str.contains x y) --> (= x y)` and we can use the
helper function to determine subterms in `y` that are empty if `y` is in
`x`, which allows us to introduce equalities of the form `(= yi "")`.
This approach unifies multiple existing rewrites. Additional rewrites
will be subsumed once we merge the approximations for arithmetic
entailment.


4tXJ7f:
50+ min of running rewriter verification on both strpred and strterm didn't reveal any issues.

src/theory/strings/theory_strings_rewriter.h
```diff
@@ -534,6 +534,42 @@ class TheoryStringsRewriter {
...
+   * inferZerosInSumGeq( (str.len x), [ (str.len x), (str.len y), 1 ], [] )
+   * --> returns true with ys = [ (str.len x) ] and zeroYs = [ (str.len y), 1 ]
+   *     (can be used to rewrite the inequality to false)
+   *
+   * inferZerosInSumGeq( (str.len x), [ (str.len y) ], [] )
+   * --> returns false because it is not possible to show
+   *     str.len(y) >= str.len(x)
+   */
+  static bool inferZerosInSumGeq(Node x,
+                                 std::vector<Node>* ys,
```
ajreynol: Any reason why these are pointers? I think the common practice is pass by reference std::vector< Node >& (I would prefer this for consistency).

src/theory/strings/theory_strings_rewriter.h
```diff
@@ -534,6 +534,42 @@ class TheoryStringsRewriter {
...
+   *
+   * (and (= x (str.++ y1' ... ym')) (= y1'' "") ... (= yk'' ""))
+   *
+   * where each yi'' are yi that must be empty for (= x y) to hold and yi' are
+   * yi that the function could not infer anything about.  Returns a null node
+   * if the function cannot infer that str.len(y) >= str.len(x). Returns (= x
+   * y) if the function can infer that str.len(y) >= str.len(x) but cannot
+   * infer that any of the yi must be empty.
+   */
+  static Node inferEqsFromContains(Node x, Node y);
```
ajreynol: This will be pushed down to be a rewrite for equality, but its fine as it is here (an analog line of reasoning shows x = y ---> conjunction of equalities).

src/theory/strings/theory_strings_rewriter.cpp
```diff
@@ -2606,6 +2562,15 @@ Node TheoryStringsRewriter::rewritePrefixSuffix(Node n)
   {
     val = NodeManager::currentNM()->mkNode(kind::MINUS, lent, lens);
   }
+
+  // Check if we can turn the prefix/suffix into equalities by showing that the
+  // prefix/suffix is at least as long as the string
+  Node eqs = inferEqsFromContains(n[0], n[1]);
```
ajreynol: We should do something similar here for contains (as I mention in my other comment).  The idea is that contains/prefix whose length entailment holds rewrites to an equality, then subsequently equalities may rewrite to conjunction of equalities in a separate step.

src/theory/strings/theory_strings_rewriter.cpp
```diff
@@ -3809,6 +3774,141 @@ Node TheoryStringsRewriter::getStringOrEmpty(Node n)
...
+  }
+
+  if (y.getKind() != kind::STRING_CONCAT)
+  {
+    if (zeroLens.size() == 1)
+    {
+      // y is not a concatenation and we found that it must be empty, so just
+      // return (= y "")
+      Assert(zeroLens[0][0] == y);
+      return nm->mkNode(kind::EQUAL, y, emp);
```
ajreynol: Remove the kind:: from places (we are in the kind namespace).

4tXJ7f: I somewhat prefer to have enums mentioned explicitly in their namespace to avoid confusion. But yeah, consistency is probably more important.

4tXJ7f:
Addressed the review.

PR #2513, opened 2018-09-23
Refactor strings equality rewriting

ajreynol:
This moves the extended rewrites for string equality to the main strings rewriter as a function rewriteEqualityExt, and makes this function called on every equality that is generated (from non-equalities) by our rewriter.


src/theory/strings/theory_strings_rewriter.cpp
```diff
@@ -1625,6 +1775,12 @@ Node TheoryStringsRewriter::rewriteContains( Node node ) {
     Node ret = NodeManager::currentNM()->mkConst(false);
     return returnRewrite(node, ret, "ctn-len-ineq");
   }
+  else if (checkEntailArith(len_n2, len_n1, false))
```
4tXJ7f: This should be subsumed by #2512.

src/theory/strings/theory_strings_rewriter.cpp
```diff
@@ -3813,5 +3969,15 @@ Node TheoryStringsRewriter::returnRewrite(Node node, Node ret, const char* c)
 {
   Trace("strings-rewrite") << "Rewrite " << node << " to " << ret << " by " << c
                            << "." << std::endl;
+  // standard post-processing
+  if (ret.getKind() == EQUAL && node.getKind() != EQUAL)
```
4tXJ7f: Do the same for conjunction of equalities (e.g. `(and (= x y) (= z w))`).

ajreynol: The slightly annoying thing here is that technically we should do it for any Boolean combination of equalities.

ajreynol:
Regarding your first point about https://github.com/CVC4/CVC4/pull/2513/files#diff-fd25db71e2d85f17645ffb7d54384bf0R256 

This should be safe, see:
https://github.com/CVC4/CVC4/pull/2513/files#diff-fd25db71e2d85f17645ffb7d54384bf0R243

src/theory/strings/theory_strings_rewriter.cpp
ajreynol: I would argue for keeping this rewrite. This rewrite + generalizing inferEqsFromContains to apply bidirectionality to equalities should subsume the current.

src/theory/strings/theory_strings_rewriter.cpp
```diff
@@ -4070,11 +4064,31 @@ Node TheoryStringsRewriter::returnRewrite(Node node, Node ret, const char* c)
...
                            << "." << std::endl;
   // standard post-processing
-  if (ret.getKind() == EQUAL && node.getKind() != EQUAL)
+  // We rewrite (string) equalities immediately here. This allows us to forego
+  // the standard invariant on equality rewrites (that s=t must rewrite to one
+  // of { s=t, t=s, true, false } ).
+  Kind retk = ret.getKind();
+  if( retk==OR || retk==AND )
+  {
+    std::vector< Node > children;
```
4tXJ7f: Could use a NodeBuilder.

PR #2530, opened 2018-09-25
BoolToBV modes (off, ite, all)

makaimann:
There are a couple preprocessing passes/rewrites which are very useful/detrimental for the FAST project. This would add two new flags to control whether these are turned on or not.

--bitwise-eq: Lifts bitvector operators which are asserted to be equal to a one-bit constant to the boolean equivalent. This defaults to true to match the current behavior of CVC4. For FAST we want to turn this off, so we can use the --no-bitwise-eq

--ite-to-bvite: A new preprocessing pass which rewrites ITEs to BITVECTOR_ITEs when it's convenient. This is much less aggressive than the bool-to-bv pass. Convenient meaning the boolean argument can easily be changed to a one-bit bitvector at the top-level without using another ite. This is checked by easyToLower in the pass. For FAST we want this on.


makaimann:
I have one question about CVC4's preprocessing/rewrites.

In the ite-to-bvite pass, at one point I look for patterns like (= X <one-bit const>). Based on some experimentation, it seems like these are always in a normal form where the constant is on the right. Can I count on this or should I also check for (= <one-bit const> X).

Thanks!

4tXJ7f:
No, AFAIK you can't count on that unfortunately. We order equalities based on which node is "larger", e.g. https://github.com/CVC4/CVC4/blob/14b9dbaa0c9e8dce52d1a28595dc1cc80756abed/src/theory/bv/theory_bv_rewrite_rules_core.h#L271-L281.

The `<` operator uses the node id to determine the order: https://github.com/CVC4/CVC4/blob/f4ce78488ae41b4effc140edfc35cbba79d2dcd4/src/expr/node.h#L361-L366

So the order depends on when the node was created.

makaimann:
Thanks @4tXJ7f, I'll fix that. And that makes sense to use a base class for bool-to-bv and ite-to-bvite. I'll work on that.

src/preprocessing/passes/ite_to_bvite.cpp
```diff
@@ -0,0 +1,148 @@
...
+{
+  NodeManager* nm = NodeManager::currentNM();
+  Kind kind = boolTerm.getKind();
+  Kind new_kind = kind;
+  switch (kind)
+  {
+    case kind::EQUAL:
+      // guaranteed because only executed if easyToLower(boolTerm) is true
+      Assert(boolTerm[0].getType().isBitVector());
+      // don't need to handle other case because of rewrites
```
mpreiner: Maybe still `Assert`?

makaimann: Sounds good, thanks. The code has changed based on our offline discussion anyway.

src/preprocessing/passes/ite_to_bvite.cpp
```diff
@@ -0,0 +1,148 @@
...
+PreprocessingPassResult IteToBvite::applyInternal(
+    AssertionPipeline* assertionsToPreprocess)
+{
+  NodeManager::currentResourceManager()->spendResource(
+      options::preprocessStep());
+
+  for (unsigned i = 0; i < assertionsToPreprocess->size(); ++i)
+  {
+    assertionsToPreprocess->replace(
+        i, Rewriter::rewrite(lowerIte(assertionsToPreprocess->operator[](i))));
```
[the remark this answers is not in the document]
makaimann: Done. Thanks

src/preprocessing/passes/ite_to_bvite.cpp
```diff
@@ -0,0 +1,148 @@
...
+IteToBvite::IteToBvite(PreprocessingPassContext* preprocContext)
+    : PreprocessingPass(preprocContext, "ite-to-bvite"), d_statistics(){};
+
+PreprocessingPassResult IteToBvite::applyInternal(
+    AssertionPipeline* assertionsToPreprocess)
+{
+  NodeManager::currentResourceManager()->spendResource(
+      options::preprocessStep());
+
+  for (unsigned i = 0; i < assertionsToPreprocess->size(); ++i)
```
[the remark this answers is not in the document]
makaimann: Sorry, stupid question. What are you referring to? The node manager spendResource thing?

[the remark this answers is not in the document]
makaimann: Nevermind, I get it

src/preprocessing/passes/ite_to_bvite.cpp
```diff
@@ -0,0 +1,148 @@
...
+  }
+
+  NodeBuilder<> builder(new_kind);
+  if (term.getMetaKind() == kind::metakind::PARAMETERIZED)
+  {
+    builder << term.getOperator();
+  }
+
+  builder << lowerIte(firstChild);
+  for (unsigned i = 1; i < term.getNumChildren(); ++i)
```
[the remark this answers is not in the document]
makaimann: Code changed -- using a range iterator now.

src/preprocessing/passes/bool_to_bv.cpp
```diff
@@ -30,183 +29,232 @@ namespace passes {
...
 {
-  Node result;
+  bool optionITE = options::boolToBitvector() == BOOL_TO_BV_ITE;
   NodeManager* nm = NodeManager::currentNM();
-  if (hasLowerCache(current))
-  {
-    result = getLowerCache(current);
-  }
-  else
+  std::vector<TNode> terms;
```
mpreiner: Maybe call this `visit` since it's a to-visit stack.

src/preprocessing/passes/bool_to_bv.cpp
```diff
@@ -30,183 +29,232 @@ namespace passes {
...
+    if ((visited.count(n) == 0) && (n.getNumChildren() > 0))
     {
-      if (current.getKind() == kind::CONST_BOOLEAN)
-      {
-        result = (current == bv::utils::mkTrue()) ? d_one : d_zero;
-      }
-      else
+      visited.insert(n);
+      terms.push_back(n);
+      for (const Node& nn : n)
```
mpreiner: You can use `terms.insert(terms.end(), n.begin(), n.end());` instead of the for loop.

src/preprocessing/passes/bool_to_bv.cpp
```diff
@@ -30,183 +29,232 @@ namespace passes {
...
-          if (current.getType().isBitVector() || current.getType().isBoolean())
+        // check for ite-conditions
+        if (k == kind::ITE)
+        {
+          ite_cond.insert(n[0]);
+        }
+        else if (ite_cond.count(n))
+        {
+          // being part of an ite condition is inherited from the parent
+          for (const Node& nn : n)
```
mpreiner: Ditto.

src/preprocessing/passes/bool_to_bv.cpp
```diff
@@ -30,183 +29,232 @@ namespace passes {
...
 }
 
-bool BoolToBV::hasLowerCache(TNode term) const
+bool BoolToBV::needToRebuild(TNode n) const
 {
-  return d_lowerCache.find(term) != d_lowerCache.end();
+  // check if any children were rebuilt
+  for (const Node& nn : n)
+  {
+    if (d_lowerCache.count(nn))
```
mpreiner: If you only want to check if `nn` is in `d_lowerCache` use `ContainsKey` from the `base/map_util.h`

src/preprocessing/passes/bool_to_bv.cpp
```diff
@@ -30,183 +29,232 @@ namespace passes {
...
-    if (current.getNumChildren() == 0)
+    TNode n = terms.back();
+    terms.pop_back();
+    Kind k = n.getKind();
+    Debug("bool-to-bv") << "BoolToBV::lowerAssertion Post-traversal with " << n
+                        << " and visited = " << visited.count(n) << std::endl;
+
+    // Mark as visited
+    /* Optimization: if it's a leaf, don't need to wait to do the work */
+    if ((visited.count(n) == 0) && (n.getNumChildren() > 0))
```
mpreiner: Ditto (`count`).

src/preprocessing/passes/bool_to_bv.cpp
```diff
@@ -30,183 +29,232 @@ namespace passes {
...
+  std::unordered_set<TNode, TNodeHashFunction> ite_cond;
+
+  while (!terms.empty())
   {
-    if (current.getNumChildren() == 0)
+    TNode n = terms.back();
+    terms.pop_back();
+    Kind k = n.getKind();
+    Debug("bool-to-bv") << "BoolToBV::lowerAssertion Post-traversal with " << n
+                        << " and visited = " << visited.count(n) << std::endl;
```
mpreiner: Ditto (`count`).

src/preprocessing/passes/bool_to_bv.cpp
```diff
@@ -30,183 +29,232 @@ namespace passes {
...
-        case kind::XOR: new_kind = kind::BITVECTOR_XOR; break;
-        case kind::IMPLIES: new_kind = kind::BITVECTOR_OR; break;
-        case kind::ITE:
-          if (current.getType().isBitVector() || current.getType().isBoolean())
+        // check for ite-conditions
+        if (k == kind::ITE)
+        {
+          ite_cond.insert(n[0]);
+        }
+        else if (ite_cond.count(n))
```
mpreiner: Ditto (`count`).

src/preprocessing/passes/bool_to_bv.cpp
```diff
@@ -30,183 +29,232 @@ namespace passes {
...
-        {
-          converted = lowerNode(current[i]);
-          builder << converted;
         }
       }
-      result = builder;
     }
-    if (result.getType().isBoolean())
+    /* Optimization for ite mode */
+    else if (optionITE && (ite_cond.count(n) == 0) && !needToRebuild(n))
```
mpreiner: Ditto (`count`).

src/preprocessing/passes/bool_to_bv.cpp
```diff
@@ -30,183 +29,232 @@ namespace passes {
...
-                             std::vector<Node>& new_assertions)
+void BoolToBV::lowerNode(const TNode& n)
 {
-  for (unsigned i = 0; i < assertions.size(); ++i)
+  NodeManager* nm = NodeManager::currentNM();
+  bool all_bv = true;
+  // check if it was able to convert all children to bitvectors
+  for (const Node& nn : n)
+  {
+    all_bv = all_bv && fromCache(nn).getType().isBitVector();
```
mpreiner: You could `break` as soon as `all_bv` is false.

src/preprocessing/passes/bool_to_bv.cpp
```diff
@@ -30,183 +29,232 @@ namespace passes {
...
+    for (const Node& nn : n)
+    {
+      builder << fromCache(nn);
+    }
+  }
+
+  Debug("bool-to-bv") << "BoolToBV::lowerAssertion " << n << "=>\n"
+                      << builder << std::endl;
+
+  d_lowerCache[n] = builder;
```
mpreiner: For better documentation purpose: `build.contructNode()`.

makaimann:
All right, I think I'm about ready to merge this. I've been comparing --bool-to-bv=all to master's --bool-to-bv option on QF\_BV. I finally have a version which solves more benchmarks than master, although it depends on a strange encoding.

Comparison is attached:
[comparison.txt](https://github.com/CVC4/CVC4/files/2485826/comparison.txt)

master-bool-to-bv: running master with --bool-to-bv
latest-all: the "regular" implementation, which uses bvor and bvnot to encode IMPLIES
latest_alt_implies_encoding: using bvnot and bvand to encode IMPLIES

TL;DR summary: There is no bit-vector equivalent of IMPLIES, so it needs to be encoded using bit-vector operators. I originally used the standard (-p | q) encoding, but that caused a different order in a particular type of node and it solved 4 less benchmarks overall (+1 -5). Using a different encoding, which I consider less natural, actually does better on the benchmarks (+14 overall) so I pushed that change. This encoding is -(p & -q). I don't feel strongly about this so if people prefer the "normal" encoding, I can use that instead.

Some more information below:

Note, some of the benchmarks which timed-out with the (-p | q) encoding weren't originally close to the timeout (so there's a significant difference).

There is no bitvector version of IMPLIES, so it's encoded: (p => q) ---> (-p | q). However, the way the final node is produced happens in a different order between recursive and post-order traversal. This can be seen below:

Original (recursive) traversal:
https://github.com/CVC4/CVC4/blob/ca6f1b0350b18ce3e134701f68f7e02813c3fb5f/src/preprocessing/passes/bool_to_bv.cpp#L148-L154

For a node (=> p q), when it's transformed to (bvor (bvnot lowered(p)) lowered(q)) the order stays the same because all of the transformations on p happen before q.

New (post-order) traversal:
https://github.com/CVC4/CVC4/blob/949b627415824fe4df7134c0707ed3874645130d/src/preprocessing/passes/bool_to_bv.cpp#L236-L240

Unlike the recursive traversal, in the node (=> p q), q has already been lowered to a bitvector at this point and doesn't need to be modified anymore. But to do the encoding, we need to negate p, which changes the node id. Now, instead of (bvor (bvnot lowered(p)) lowered(q)) we get (bvor lowered(q) (bvnot lowered(p))).

It doesn't seem like this should have that much effect because it's now bit-vector operators. However, when this occurs at the top-level, a different pass: BitwiseEq turns this back into boolean operators and now the justification heuristic traverses in a different (unlucky) order.

There is a solution to this problem, which is to encode the implies differently so that both nodes are updated, something like (p => q) --> -(p & -q). Intuitively this doesn't seem better to me, and is not the standard translation. However, empirically it does much better, solving 14 benchmarks *more* than master with --bool-to-bv. Because of this, I pushed that encoding but if anyone is opposed to it I can revert it.

Hopefully that was all clear. Please let me know if I can clarify anything! 

If you'd like, I could also run a comparison with --no-bitwise-eq to turn off that other pass which lifts back to booleans. But personally I don't think that's necessary, because I'm pretty confident that this refactor/adding modes is correct at this point.

src/options/bool_to_bv_mode.h
```diff
@@ -0,0 +1,57 @@
...
+
+/** Enumeration of bool-to-bv modes */
+enum BoolToBVMode {
+  /**
+   * No bool-to-bv pass
+   */
+  BOOL_TO_BV_OFF,
+
+  /**
+   * Only lower bools in condition of ites
```
mpreiner: Nit: ITEs.

src/options/bool_to_bv_mode.h
```diff
@@ -0,0 +1,57 @@
...
+   * Tries to give more info to bit-vector solver
+   * by using bitvector-ites when possible
```
mpreiner: Ditto

src/options/bv_options.toml
```diff
@@ -105,11 +105,22 @@ header = "options/bv_options.h"
...
+
+[[option]]
+  name       = "bitwiseEq"
+  category   = "regular"
+  long       = "bitwise-eq"
   type       = "bool"
-  default    = "false"
-  help       = "convert booleans to bit-vectors of size 1 when possible"
+  default    = "true"
+  help       = "lift equivalence with one-bit bitvectors to be boolean operations"
```
mpreiner: Nit: bit-vectors

src/options/options_handler.cpp
```diff
@@ -1243,6 +1243,48 @@ theory::bv::BvSlicerMode OptionsHandler::stringToBvSlicerMode(
...
 
+const std::string OptionsHandler::s_boolToBVModeHelp =
+    "\
+BoolToBV pass modes supported by the --bool-to-bv option:\n\
+\n\
+off (default)\n\
++ Don't push any booleans to width one bit-vectors\n\
+\n\
+ite\n\
++ Try to turn ITEs into BITVECTOR_ITE when possible. Might fail if not all sub-formulas can be turned to bit-vectors\n\
```
mpreiner: This does not seem to be 80 characters long. clang-format?

src/options/options_handler.cpp
```diff
@@ -1243,6 +1243,48 @@ theory::bv::BvSlicerMode OptionsHandler::stringToBvSlicerMode(
...
+\n\
+all\n\
++ Force all booleans to be bit-vectors of width one except at the top level. Most aggressive mode\n\
```
mpreiner: Ditto

src/preprocessing/passes/bool_to_bv.cpp
```diff
@@ -30,183 +30,253 @@ namespace passes {
...
           "preprocessing::passes::BoolToBV::NumTermsForcedLowered", 0)
 {
-  smtStatisticsRegistry()->registerStat(&d_numTermsLowered);
-  smtStatisticsRegistry()->registerStat(&d_numAtomsLowered);
-  smtStatisticsRegistry()->registerStat(&d_numTermsForcedLowered);
+  smtStatisticsRegistry()->registerStat(&d_numIteToBvite);
+  if (options::boolToBitvector() == BOOL_TO_BV_ALL)
+  {
+    // these statistics wouldn't be correct in the ITE mode,
+    // because might discard rebuilt nodes if it fails to
```
mpreiner: because *it* might?

src/preprocessing/passes/bool_to_bv.h
```diff
@@ -39,24 +39,33 @@ class BoolToBV : public PreprocessingPass
...
-  Node getLowerCache(TNode term) const;
-  bool hasLowerCache(TNode term) const;
-  Node lowerNode(TNode current, bool topLevel = false);
-  NodeNodeMap d_lowerCache;
-  Node d_one;
-  Node d_zero;
+  /* Takes an assertion and tries to create more bit-vector structure */
+  Node lowerAssertion(const TNode& a);
+
+  /* Tries to lower one node to a width-one bitvector */
```
mpreiner: Nit: bit-vector

PR #2532, opened 2018-09-25
Add rewrite for solving stoi

ajreynol:
Also minor changes to rewriteEqualityExt infrastructure.

This also connects rewriteEqualityExt as a context-dependent inference scheme.


src/theory/strings/theory_strings_rewriter.cpp
```diff
@@ -481,12 +492,50 @@ Node TheoryStringsRewriter::rewriteEqualityExt(Node node)
...
+
+
+Node TheoryStringsRewriter::rewriteArithEqualityExt(Node node)
+{
+  Assert( node.getKind()==EQUAL && node[0].getType().isInteger() );
+  
+  NodeManager* nm = NodeManager::currentNM();
+  
+  // cases where we can solve the equality
+  for( unsigned i=0; i<2; i++ )
```
4tXJ7f: Apply clang-format.

src/theory/strings/theory_strings.h
```diff
@@ -602,6 +608,21 @@ class TheoryStrings : public Theory {
    * effort, the call to this method does nothing.
    */
   void registerTerm(Node n, int effort);
+  //-------------------------------------send inferences
+  /** send internal inferences
+   * 
+   * This is called when we have inferred exp => conc, where exp is a set
+   * of equalities and disequalities that hold in the current equality engine.
+   * This method adds equalities and disequalities ~( s = t ) via
+   * sendInference such that both s (resp. t) is either a constant or a term
```
4tXJ7f: "such that both s (resp. t) is" reads a bit strange, maybe just write "such that both s and t are"

src/theory/strings/theory_strings.h
```diff
@@ -602,6 +608,21 @@ class TheoryStrings : public Theory {
...
+   * that already occurs in the equality engine and ~( s = t ) is a consequence
+   * of conc. This function can be seen as a "conservative" version of
+   * sendInference below in that it does not introduce any new non-constant
+   * terms to the state. 
+   * 
+   * The argument c is used for debugging.
```
4tXJ7f: Maybe say something a bit more concrete, e.g. "c is a string identifying the reason for the interference. This string is used for debugging purposes.".

src/theory/strings/theory_strings.h
```diff
@@ -580,7 +580,13 @@ class TheoryStrings : public Theory {
   void doPendingFacts();
   void doPendingLemmas();
   bool hasProcessed();
+  /** 
+   * Adds equality a = b to the vector exp if a and b are distinct. Checks 
+   * that a = b holds in this context, throws an assertion failure via Assert
+   * if not.
```
4tXJ7f: I would just say "a = b must hold in this context" because it is only actually checked in debug mode and this description makes it seem like it is always checked and the exception could be caught by the callee.

src/theory/strings/theory_strings.cpp
```diff
@@ -3809,6 +3842,56 @@ void TheoryStrings::registerTerm( Node n, int effort ) {
...
+  if( conc.getKind()==AND )
+  {
+    for( const Node& cc : conc )
+    {
+      sendInternalInference( exp, cc, c );
+    }
+    return;
+  }
+  Node lit = conc.getKind()==NOT ? conc[0] : conc;
+  bool pol = conc.getKind()!=NOT;
```
4tXJ7f: Suggestion: It would be slightly more elegant to first determine `pol` and then use that in the condition for `lit`.

src/theory/strings/theory_strings.cpp
```diff
@@ -3809,6 +3842,56 @@ void TheoryStrings::registerTerm( Node n, int effort ) {
...
+    }
+    // does it already hold?
+    if( pol ? areEqual( lit[0], lit[1] ) : areDisequal( lit[0], lit[1] ) )
+    {
+      return;
+    }
+  }
+  else if( lit.isConst() )
+  {
+    if( lit.getConst<bool>() )
```
4tXJ7f: Suggestion: Merge with `else if` condition above.

src/theory/strings/theory_strings.cpp
```diff
@@ -1659,12 +1659,28 @@ void TheoryStrings::checkExtfInference( Node n, Node nr, ExtfInfoTmp& in, int ef
...
-  Node exp = n.eqNode(in.d_const);
-  exp = Rewriter::rewrite(exp);
-
+  Trace("strings-extf-infer") << "checkExtfInference: " << n << " : " << nr << " == " << in.d_const << std::endl;
+  
   // add original to explanation
-  in.d_exp.push_back(exp);
+  if( n.getType().isBoolean() )
+  {
+    // if Boolean, its easy
```
4tXJ7f: Nit: its -> it's

src/theory/strings/theory_strings.cpp
```diff
@@ -1793,6 +1809,23 @@ void TheoryStrings::checkExtfInference( Node n, Node nr, ExtfInfoTmp& in, int ef
         getExtTheory()->markReduced(n);
       }
     }
+    return;
+  }
+  
+  // If its not a predicate, see if we can solve the equality n = c, where c
```
4tXJ7f: Nit: its -> it's

src/theory/strings/theory_strings.cpp
ajreynol: This actually would change the control flow. If lit is const and is false, then we should fall through.

4tXJ7f: I'm stupid.

PR #2533, opened 2018-09-25
cmake: Add check for GCC 4.5.1 and warn user.

mpreiner:


CMakeLists.txt
```diff
@@ -216,6 +216,24 @@ endif()
...
 
+#-----------------------------------------------------------------------------#
+# Check GCC version.
+#
+# GCC version 4.5.1 builds MiniSat incorrectly with -O2, which results in
+# incorrect answers.
+
+if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
+  execute_process(
+    COMMAND gcc -dumpversion
```
aniemetz: We should probably not use `gcc` here (but the cmake variable that points to the executable).

mpreiner: Done

PR #2535, opened 2018-09-26
cmake: Exclude examples for coverage target.

mpreiner:
We should not rely on examples to increase the code coverage.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -252,7 +252,7 @@ if(ENABLE_COVERAGE)
   add_definitions(-DCVC4_COVERAGE)
   setup_target_for_coverage_lcov(
     NAME coverage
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -252,7 +252,7 @@ if(ENABLE_COVERAGE)
   add_definitions(-DCVC4_COVERAGE)
   setup_target_for_coverage_lcov(
     NAME coverage
-    EXECUTABLE ctest -j${CTEST_NTHREADS} $(ARGS)
+    EXECUTABLE ctest -j${CTEST_NTHREADS} -LE "example" $(ARGS)
```
aniemetz: Should we call this with `--output-on-failure`?

PR #2538, opened 2018-09-26
Fix Taylor overapproximation for large exponentials

ajreynol:
Currently master gives unsound over-approximations via Taylor for exp( x ) for x >= ~3 since it does not compute and use the minimum degree for which the Taylor approximation of exp( x ) is sound. This fixes the issue.


test/regress/regress1/nl/exp-approx.smt2
```diff
@@ -0,0 +1,15 @@
...
+(set-logic QF_NRAT)
+(set-info :status sat)
+
+(assert (and (<= 3.0 (exp 1.1)) (<= (exp 1.1) 3.1)))
+(assert (and (<= 8.1 (exp 2.1)) (<= (exp 2.1) 8.2)))
+(assert (and (<= 22.1 (exp 3.1)) (<= (exp 3.1) 22.2)))
+(assert (and (<= 60.3 (exp 4.1)) (<= (exp 4.1) 60.4)))
+(assert (and (<= 164.0 (exp 5.1)) (<= (exp 5.1) 164.1)))
+
+; this takes ~10 seconds in production
```
HanielB: ideally we should have the commit hash in which this takes ~10. But we will only have the hash when this is merged...

src/theory/arith/nonlinear_extension.h
```diff
@@ -649,15 +649,32 @@ class NonlinearExtension {
...
-   * approximations of the Taylor series of <k>( 0 ) for degree 2*d where
-   * k is SINE or EXPONENTIAL.
+   * This adds P_l+[x], P_l-[x], P_u+[x], P_u-[x] to pbounds, where x is
+   * d_taylor_real_fv. These are polynomial approximations of the Taylor series
+   * of <k>( 0 ) for degree 2*d where k is SINE or EXPONENTIAL.
    * These correspond to P_l and P_u from Figure 3 of Cimatti et al., CADE 2017,
    * for positive/negative (+/-) values of the argument of <k>( 0 ).
+   *
+   * Notice that for certain bounds (e.g. upper bounds for exponential), the
+   * Taylor approximation for a fixed degree are only sound up to a given
```
HanielB: are only -> is only?

PR #2542, opened 2018-09-27
cmake: Add CxxTest finder module to allow custom paths.

mpreiner:


test/unit/CMakeLists.txt
```diff
@@ -1,3 +1,6 @@
+set(CxxTest_HOME ${CXXTEST_DIR})
+find_package(CxxTest REQUIRED)
```
aniemetz: Can you maybe document why we do not use the default finder?

PR #2553, opened 2018-10-01
Fix coverity warnings in datatypes

ajreynol:
This caches some information regarding tester applications and changes int -> unsigned in a few places.


src/theory/datatypes/theory_datatypes.h
```diff
@@ -42,7 +42,7 @@ namespace datatypes {
 class TheoryDatatypes : public Theory {
  private:
   typedef context::CDList<Node> NodeList;
-  typedef context::CDHashMap<Node, int, NodeHashFunction> NodeIntMap;
+  typedef context::CDHashMap<Node, unsigned, NodeHashFunction> NodeUIntMap;
```
4tXJ7f: If we are already changing the type, let's change it to a type of a fixed bit-width (`uint64_t`/`uint32_t`).

ajreynol: This maps Nodes to an index in a vector. I'll change to size_t.

src/theory/datatypes/theory_datatypes.cpp
```diff
@@ -2121,18 +2132,20 @@ void TheoryDatatypes::printModelDebug( const char* c ){
...
               }
             }
           }
           Trace( c ) << std::endl;
           Trace( c ) << "   Selectors : " << ( ei->d_selectors ? "yes, " : "no " );
-          NodeIntMap::iterator sel_i = d_selector_apps.find( eqc );
+          NodeUIntMap::iterator sel_i = d_selector_apps.find(eqc);
           if( sel_i != d_selector_apps.end() ){
-            for( int j=0; j<(*sel_i).second; j++ ){
+            for (unsigned j = 0; j < (*sel_i).second; j++)
```
4tXJ7f: `(*sel_i).second` now has type `size_t`, please update (also applies to the other cases).

PR #2554, opened 2018-10-01
init scalar class members (coverity issues 1473720 and 1473721)

HanielB:


src/theory/quantifiers/sygus/enum_stream_substitution.cpp
```diff
@@ -31,6 +31,8 @@ namespace quantifiers {
 EnumStreamPermutation::EnumStreamPermutation(quantifiers::TermDbSygus* tds)
     : d_tds(tds)
 {
+  d_first = true;
+  d_curr_ind = 0;
```
4tXJ7f: Why not use the initialization list, i.e. put it behind `d_tds(tds)`? :)

PR #2557, opened 2018-10-01
`make regress` runs **all** regression tests

alex-ozdemir:
Just to be consistent with the documentation in the [Install README](https://github.com/CVC4/CVC4/blob/master/INSTALL.md#custom-targets).


alex-ozdemir:
A note from offline discussion: Clark suggested that rather than changing the behavior of `make regress` we should change the documentation, for historical reasons. Would y'all agree with this, @mpreiner / @aniemetz?

aniemetz:
Yeah, historical reasons is why this was limited to levels 0 and 1. Can you maybe rather amend the documentation (INSTALL.md)?

alex-ozdemir:
Done.

PR #2586, opened 2018-10-04
Update default options for sygus

ajreynol:
This makes --decision=internal default for sygus solving, as well as disabling repair constants.

It also simplifies the logic for enabling cegqi so that this does not depend on decision=internal. Formerly, decision=internal + cegqi were incompatible unless global prenexing was enabled, but recent tests shown combining these is not problematic.  It is likely that the recent decision strategy cleanup for cegqi (regarding outermost-first quantifiers first) eliminated this dependency.

It also fixes a caching issue for PBE symmetry breaking that was uncovered in regress2 by this change.

### DIFF
--- a/examples/api/bitvectors-new.cpp
+++ b/examples/api/bitvectors-new.cpp
@@ -50,10 +50,8 @@ int main()
   // Creating a bit-vector type of width 32
   Sort bitvector32 = slv.mkBitVectorSort(32);
 
-  std::cout << "bitvector32 " << bitvector32 << std::endl;
   // Variables
   Term x = slv.mkVar("x", bitvector32);
-  std::cout << "bitvector32 " << bitvector32 << std::endl;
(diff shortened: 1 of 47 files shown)


HanielB:
Can you post the numbers to justify making this default?

HanielB:
And what about the `--sygus-add-const-grammar` option? Shouldn't we have that as on by default now that `--sygus-repair-const` is off?

ajreynol:
Concerning `--sygus-add-const-grammar`, I'm getting that it is +10-13 on lustre-sygus and +2-0 on sygusComp 2018 (with no other options).  I don't think it should be enabled since its not a huge gain and is by nature a bit hacky.

ajreynol:
Justifying the new defaults:

For lustre-sygus, we are +70-4 with no command line options (30 min)
For sygusComp 2018, we are +42-13 with no command line options (30 min), where the big wins are in PBE strings.

PR #1370, opened 2017-11-14
Add random number generator.

aniemetz:
This adds a deterministic (seeded) random number generator (RNG). It implements the xorshift* generator (see [S. Vigna, An experimental exploration of Marsaglia's xorshift generators, scrambled. ACM Trans. Math. Softw. 42(4): 30:1-30:23, 2016](http://vigna.di.unimi.it/ftp/papers/xorshift.pdf)).

PR #2587, opened 2018-10-04
Option for shuffling condition pool in CegisUnif

HanielB:


src/theory/quantifiers/sygus/sygus_unif_rl.cpp
```diff
@@ -563,6 +563,17 @@ Node SygusUnifRl::DecisionTreeInfo::buildSolAllCond(Node cons,
...
+  // shuffle conditions before bulding DT
+  //
+  // this does not impact whether it's possible to build a solution, but it does
+  // impact the potential size of the resulting solution (can make it smaller,
+  // bigger, or have no impact) and which conditions will be present in the DT,
+  // which influences the "quality" of the solution for cases not covered in the
+  // current data points
+  if (options::sygusUnifShuffleCond())
+  {
+    std::shuffle(d_conds.begin(), d_conds.end(), d_rng);
```
ajreynol: Use std::random_shuffle.  I haven't seen this d_rng thing before. 

Here's an example of shuffling:
https://github.com/CVC4/CVC4/blob/master/src/theory/quantifiers/cegqi/ceg_bv_instantiator.cpp#L309

src/theory/quantifiers/sygus/sygus_unif_rl.h
```diff
@@ -18,6 +18,8 @@
 #define __CVC4__THEORY__QUANTIFIERS__SYGUS_UNIF_RL_H
 
 #include <map>
+#include <random>
```
ajreynol: Remove this.

src/theory/quantifiers/sygus/sygus_unif_rl.h
```diff
@@ -382,6 +387,8 @@ class SygusUnifRl : public SygusUnif
      * enumerated condiotion values
      */
     PointSeparator d_pt_sep;
+    /** random number generator */
+    std::default_random_engine d_rng;
```
ajreynol: Remove this.

src/theory/quantifiers/sygus/sygus_unif_rl.cpp
HanielB: I saw online that `std::shuffle` was preferable to `std::random_shuffle` from C++11 on. And I'm parameterizing it with a random generator so that we can have control over the sequence. This is necessary for reproducing results.

HanielB:
I still have to evaluate, I'll post numbers as soon as I have them.

src/theory/quantifiers/sygus/sygus_unif_rl.cpp
ajreynol: I'm just following what we did for the CAV 2018 BV algorithm. Perhaps @aniemetz can weigh in?

aniemetz: The approach @HanielB proposes is actually the way to go. We want any randomness in CVC4 to be deterministic, hence the existence of d_rng. The example @ajreynol posted should be changed accordingly. It seems that I overlooked this back then when I introduced the random number generator in #1370.

aniemetz: I just realized that @HanielB is not using CVC4's RNG. He should.

src/theory/quantifiers/sygus/sygus_unif_rl.h
aniemetz: See my comment above, this should indeed be removed since CVC4's RNG should be used.

src/theory/quantifiers/sygus/sygus_unif_rl.cpp
aniemetz: I'll submit a PR for the change in ceq_bv_instantiator.cpp.

ajreynol:
Were you able to test this option yet?  If it works, this looks good to me.

HanielB:
It has an impact but not a significant one. 

-  for `--sygus-unif-cond-indpendent` it leads to -29 +27 on lustre and +4 -1 on syguscomp2018_inv
-  for `--sygus-unif-boolean-heuristic-dt` it leads to -4 +2 on lustre and +4 -3 on syguscomp2018_inv

More work is required on applying shuffling multiple times for each candidate solution for us to have more significant impacts with this option.

PR #2593, opened 2018-10-07
Allow multiple synthesis conjectures.

ajreynol:
This allows the synthesis engine to handle multiple synthesis conjectures in parallel. This is in preparation for synthesizing rewrite rules from input formulas.


src/theory/quantifiers/sygus/synth_engine.h
```diff
@@ -31,32 +31,32 @@ class SynthEngine : public QuantifiersModule
...
  private:
-  /** the quantified formula stating the synthesis conjecture */
+  /** the conjecture formula(s) we are waiting to assign */
+  std::vector<Node> d_waiting_conj;
+  /** The synthesis conjectures that this class is managing. */
+  std::vector<std::unique_ptr<SynthConjecture> > d_conjs;
+  /**
+   * The first conjecture in the above vector. We track this conjecture
+   * so that a synthesis conjecture can be preregistered during a call to
+   * preregisterAssertion.
```
HanielB: Can you clarify why this is necessary?

src/theory/quantifiers/sygus/synth_engine.cpp
```diff
@@ -34,78 +34,78 @@ namespace quantifiers {
...
-      // assign conjecture always uses the output channel, we return and
-      // re-check here.
-      return;
-    }
+    assignConjecture(q);
+  }
+  if (assigned)
+  {
+    // assign conjecture always uses the output channel, we return and
+    // re-check here.
```
HanielB: can you clarify why always using the output channel entails we need to return and re-check?

HanielB:
I see now, thanks.

PR #2594, opened 2018-10-08
 Disable extended rewriter when applicable with var agnostic enumeration

ajreynol:
This makes `--sygus-var-agnostic` and `--no-sygus-sym-break-dynamic` (which disables rewriter-based symmetry breaking) work together.


src/theory/quantifiers/sygus/enum_stream_substitution.cpp
```diff
@@ -187,8 +187,11 @@ Node EnumStreamPermutation::getNext()
     bultin_perm_value = d_tds->sygusToBuiltin(perm_value, perm_value.getType());
     Trace("synth-stream-concrete-debug")
         << " ......perm builtin is " << bultin_perm_value;
-    bultin_perm_value =
-        d_tds->getExtRewriter()->extendedRewrite(bultin_perm_value);
+    if (options::options::sygusSymBreakDynamic())
```
HanielB: please take into account the trace messages as well.

PR #2595, opened 2018-10-08
Random: support URNG interface

aniemetz:
Use std::shuffle (with Random as the unified random generator) instead
of std::random_shuffle for deterministic, reproducable random shuffling.


src/util/random.h
```diff
@@ -26,29 +26,40 @@ namespace CVC4 {
 class Random
 {
  public:
-  Random(uint64_t seed) { setSeed(seed); }
+  typedef uint64_t result_type;
```
HanielB: Replace by `using ... = ...` ?

PR #2597, opened 2018-10-08
 Fix string register extended terms

ajreynol:
A regress2 benchmark was failing, due to a recent change in our strings rewriter.

The issue is that our string rewriter is now powerful enough to deduce that certain extended terms like `(str.substr (str.++ x "zb") 1 1)` must be non-empty. As a consequence, our emptiness-split `(str.substr (str.++ x "zb") 1 1) = "" OR len( (str.substr (str.++ x "zb") 1 1) ) > 0` is instead a propagation `len( (str.substr (str.++ x "zb") 1 1) ) > 0`. This means that `(str.substr (str.++ x "zb") 1 1)` may not appear in an assertion sent to strings.

The fix is to ensure that extended function terms in any assertions *or shared terms* are registered. 

This also simplifies the code so that another (now spurious) call to ExtTheory::registerTermRec is removed.


src/theory/strings/theory_strings.cpp
```diff
@@ -1213,7 +1212,9 @@ void TheoryStrings::assertPendingFact(Node atom, bool polarity, Node exp) {
     //getExtTheory()->registerTerm( atom );
   }
   Trace("strings-pending-debug") << "  Now collect terms" << std::endl;
-  //collect extended function terms in the atom
+  // Collect extended function terms in the atom. Notice that we must register
+  // all extended functions occurring in assertions and shared terms. We
+  // may a similar call to registerTermRec in addSharedTerm.
```
4tXJ7f: Missing word.

PR #2598, opened 2018-10-08
Address slow sygus regressions

ajreynol:


test/regress/CMakeLists.txt
```diff
@@ -1752,6 +1751,7 @@ set(regress_3_tests
   regress3/issue2429.smt2
   regress3/pp-regfile.smt
   regress3/qwh.35.405.shuffled-as.sat03-1651.smt
+  regress3/sixfuncs.sy
```
4tXJ7f: Sort alphabetically.

test/regress/Makefile.tests
```diff
@@ -1740,7 +1739,8 @@ REG3_TESTS = \
 	regress3/incorrect2.smt \
 	regress3/issue2429.smt2 \
 	regress3/pp-regfile.smt \
-	regress3/qwh.35.405.shuffled-as.sat03-1651.smt
+	regress3/qwh.35.405.shuffled-as.sat03-1651.smt \
+	regress3/sixfuncs.sy
```
4tXJ7f: Ditto (well here it is not super important since autotools is deprecated).

PR #2599, opened 2018-10-08
Bit vector proof superclass

alex-ozdemir:
## Summary

This is a refactoring diff. The primary change is the introduction of a sub-class of `BitVectorProof` called `ResolutionBitVectorProof`, which contains the parts of the proof specific to resolution. This involed splitting the class, and changing old users of `BitVectorProof` to use `ResolutionBitVectorProof`.

## Things that will be done in future PRs

Some parts of `ResolutionBitVectorProof` have thus been moved into the new `BitVectorProof` class. The division of labor isn't perfect yet, and will be refined in future PRs.

Some of the machinery external to the bv-proof types that refers to `ResolutionBitVectorProof` might be okay just using a `BitVectorProof`. Future PRs will refine this.

## Testing

I ran `make check`. All tests passed. Given that this is a refactor, that was enough to convince me.


src/proof/resolution_bitvector_proof.h
```diff
@@ -0,0 +1,102 @@
...
+ **
+ ** \brief Bitvector proof
+ **
+ ** Bitvector proof
+ **/
+
+#include "cvc4_private.h"
+
+#ifndef __CVC4__RESOLUTION__BITVECTOR__PROOF_H
+#define __CVC4__RESOLUTION__BITVECTOR__PROOF_H
```
4tXJ7f: Please use the format described in the [style guide](https://github.com/CVC4/CVC4/wiki/Code-Style-Guidelines#the-define-guard).

src/proof/resolution_bitvector_proof.h
```diff
@@ -0,0 +1,102 @@
...
+class TheoryBV;
+template <class T> class TBitblaster;
+} /* namespace CVC4::theory::bv */
+} /* namespace CVC4::theory */
+
+class CnfProof;
+} /* namespace CVC4 */
+
+namespace CVC4 {
+
```
4tXJ7f: [We should probably put this in the namespace `proof`](https://github.com/CVC4/CVC4/wiki/Code-Style-Guidelines#namespaces). Some existing files do not follow this style but since we are refactoring, we can introduce the proper style for the refactored parts.

src/proof/resolution_bitvector_proof.h
```diff
@@ -0,0 +1,102 @@
...
+protected:
+  BVSatProof* d_resolutionProof;
+
+  CnfProof* d_cnfProof;
+
+  bool d_isAssumptionConflict;
+
+  theory::TheoryId getTheoryId() override;
+  context::Context d_fakeContext;
```
4tXJ7f: Also include `context/context.h`.

src/proof/resolution_bitvector_proof.h
```diff
@@ -0,0 +1,102 @@
...
+  BVSatProof* getSatProof();
+  CnfProof* getCnfProof() { return d_cnfProof; }
+  void finalizeConflicts(std::vector<Expr>& conflicts);
+
+  void startBVConflict(CVC4::BVMinisat::Solver::TCRef cr);
+  void startBVConflict(CVC4::BVMinisat::Solver::TLit lit);
+  void endBVConflict(const BVMinisat::Solver::TLitVec& confl);
+
+  void markAssumptionConflict() { d_isAssumptionConflict = true; }
+  bool isAssumptionConflict() { return d_isAssumptionConflict; }
```
4tXJ7f: Make this method `const`, i.e. ` bool isAssumptionConflict() const`

src/proof/resolution_bitvector_proof.h
```diff
@@ -0,0 +1,102 @@
...
+
+class CnfProof;
+} /* namespace CVC4 */
+
+namespace CVC4 {
+
+template <class Solver> class TSatProof;
+typedef TSatProof< CVC4::BVMinisat::Solver> BVSatProof;
+
+class ResolutionBitVectorProof : public BitVectorProof {
```
4tXJ7f: Please add a comment for every class, method, and member variable in the header file.

src/proof/bitvector_proof.cpp
```diff
@@ -2,112 +2,45 @@
...
-  // If this term gets used in the final proof, we will want to register it. However,
-  // we don't know this at this point; and when the theory proof engine sees it, if it belongs
-  // to another theory, it won't register it with this proof. So, we need to tell the
-  // engine to inform us.
-
-  if (theory::Theory::theoryOf(term) != theory::THEORY_BV) {
-    Debug("pf::bv") << "\tMarking term " << term << " for future BV registration" << std::endl;
-    d_proofEngine->markTermForFutureRegistration(term, theory::THEORY_BV);
-  }
+std::string BitVectorProof::getBBTermName(Expr expr) {
```
4tXJ7f: [Apply clang format](https://github.com/CVC4/CVC4/wiki/Developer-Guide#formatting-with-clang-format).

src/proof/bitvector_proof.cpp
```diff
@@ -2,112 +2,45 @@
 /*! \file bitvector_proof.cpp
  ** \verbatim
  ** Top contributors (to current version):
- **   Liana Hadarean, Guy Katz, Paul Meng
+ **   Liana Hadarean, Guy Katz, Paul Meng, Alex Ozdemir
```
4tXJ7f: There is no need to update this manually :) You can use the [update-copyright.pl](https://github.com/CVC4/CVC4/blob/0da5ede58934b3a30944e32835698e84626f4c0a/contrib/update-copyright.pl) script (although we also run the script on all files before making a release, so running it on a commit is optional in my opinion).

src/proof/resolution_bitvector_proof.h
alex-ozdemir: What about `getCnfProof` above? Should it be marked `const`?

In general, may methods which return mutable references to a structure be marked `const`?

src/proof/bitvector_proof.cpp
```diff
@@ -394,7 +210,7 @@ void LFSCBitVectorProof::printBitOf(Expr term, std::ostream& os, const ProofLetM
   os << " " << bit << ")";
 }
 
-void LFSCBitVectorProof::printConstant(Expr term, std::ostream& os) {
+void BitVectorProof::printConstant(Expr term, std::ostream& os) {
```
yoni206: [ResolutionBitVectorProof](https://github.com/CVC4/CVC4/blob/609882974d85bb510d9b3256736d0b16a8e63634/src/proof/resolution_bitvector_proof.h#L52) inherits from the current class (`BitVectorProof`), and does not override this function. This suggests that any potential descendent of `ResolutionBitVectorProof` might reuse this function. However, it explicitly prints out `a_bv`, which is unique for LFSC. Should we move it? Or maybe only handle here lfsc proofs (and dismiss the extra layer e.g. in [here](https://github.com/CVC4/CVC4/blob/609882974d85bb510d9b3256736d0b16a8e63634/src/proof/resolution_bitvector_proof.h#L84)) The same holds for other functions here, e.g. `printOperatorNary` that uses `toLFSCKindTerm`.

4tXJ7f: I agree, I would be in favor of removing all LFSC related code from `BitVectorProof` and `ResolutionBitVectorProof`. If additional modifications need to be done first before we can do that, we can also move them in a later PR, though.

src/proof/resolution_bitvector_proof.cpp
```diff
@@ -0,0 +1,426 @@
...
+                               TheoryProofEngine* proofEngine)
+    : BitVectorProof(bv, proofEngine),
+      d_resolutionProof(NULL),
+      d_cnfProof(NULL),
+      d_isAssumptionConflict(false)
+      {}
+
+void ResolutionBitVectorProof::initSatProof(CVC4::BVMinisat::Solver* solver) {
+  Assert (d_resolutionProof == NULL);
+  d_resolutionProof = new BVSatProof(solver, &d_fakeContext, "bb", true);
```
4tXJ7f: It would be good to clarify the owners of `d_resolutionProof` and `d_cnfProof`. Right now, it looks like they are owned by `ResolutionBitVectorProof`. If they are, make them unique pointers, s.t. they are cleaned up when `ResolutionBitVectorProof` is destroyed (it looks like we are leaking them right now?).

src/proof/resolution_bitvector_proof.h
4tXJ7f: I think that ideally, we wouldn't return non-const pointers to objects that we own at all and maybe we could just return `const CnfProof*` in this case (may needs some fixing at the call sites)? However, if we do return mutable references, I think that it is somewhat cleaner to not declare the function to be `const` (not a strong preference though).

src/proof/resolution_bitvector_proof.h
alex-ozdemir: It's easy for me to write comments for members that I've come to understand in the course of implementing DRAT support (most of these are in `BitVectorProof`), and I'll go through and do so.

However,  most of the members in `ResolutionBitVectorProof` are not connected to DRAT, and I haven't looked into what they do. I wouldn't be able to write good comments for them right now.

If needed, I'm happy to look into this, but it would be a new task.

src/proof/resolution_bitvector_proof.h
alex-ozdemir: Agreed on both counts. Alas, the call-sites make very non-const use of the CNF and resolution proof pointers right now...

src/proof/bitvector_proof.cpp
alex-ozdemir: I agree that it would be good to separate LFSC-printing code from proof-structure-storing code.

However for reasons that will be made clear when a subsequent PR introduces `DRATBitVectorProof`, pulling all the LFSC-printing code to the bottom of the class hierarchy isn't the nicest solution, because most LFSC-printing code isn't resolution or DRAT specific.

My plan is to build a parallel hierarchy of LFSC-printers, which take a `BitVectorProof` as their input.

src/proof/bitvector_proof.h
```diff
@@ -39,49 +39,120 @@ typedef std::unordered_map<Expr, unsigned, ExprHashFunction> ExprToId;
...
+ *    2. A declaration of a cnf formula equisatisfiable to the bitblasted
+ *       formula
+ *
+ *       (and a proof that each clause is implied by some bitblasted formula)
+ *
+ *    3. A proof of UNSAT from the clauses.
+ *
+ * This class is responsible for 1 & 2. The proof of UNSAT is delegated to a
+ * subclass.
+ */
```
yoni206: The added documentation in here and above the method signatures is great!

src/proof/resolution_bitvector_proof.h
4tXJ7f: Yeah, I think that it is totally fine to just add comments to the ones you know and address the other ones later.

src/proof/resolution_bitvector_proof.h
4tXJ7f: Yeah, I think that it is fine as-is then. We can address this in future PRs.

alex-ozdemir:
@4tXJ7f, I believe that I've addressed all of your initial comments. Pinging you because I believe you wanted to do a deeper review.

If I haven't addressed some of them, let me know.

src/proof/proof_manager.h
```diff
@@ -69,15 +69,17 @@ class TheoryProof;
 class UFProof;
 class ArithProof;
 class ArrayProof;
-class BitVectorProof;
+
+namespace proof {
```
yoni206: Maybe it is better to call this namespace `bvproof` or something more indicative?

src/proof/bitvector_proof.h
```diff
@@ -9,161 +9,193 @@
...
   void printPredicate(Expr term, std::ostream& os, const ProofLetMap& map);
   void printOperatorParametric(Expr term, std::ostream& os, const ProofLetMap& map);
   void printBitOf(Expr term, std::ostream& os, const ProofLetMap& map);
 
-  ExprToString d_exprToVariableName;
-  ExprToString d_assignedAliases;
-  std::map<std::string, std::string> d_aliasToBindDeclaration;
-  std::string assignAlias(Expr expr);
-  bool hasAlias(Expr expr);
+  void printOwnedSort(Type type, std::ostream& os) override;
```
yoni206: wasn't this public before?

src/proof/proof_manager.h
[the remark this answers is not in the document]
alex-ozdemir: Seems like a good idea. We would probably want to do something like `proof::bv`, and modify the directory structure to reflect this.

I'd like to hold of on this change for now though -- I have 2s PR on top of this one that assume this class will end up in the `CVC4::proof` namespace. Refactoring code across multiple PRs can be done, but it's a bit of a pain, and I'd rather do the refactor in one go, after all the PRs are merged.

Does the delay seem worth it to you?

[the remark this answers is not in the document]
yoni206: np

src/proof/bitvector_proof.h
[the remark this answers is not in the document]
alex-ozdemir: Whoops! Yeah, the overridden method is public... Good catch!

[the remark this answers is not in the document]
yoni206: I think it wasn't the only one (see, e.g., printTermBitblasting?). Of course, if they are not needed externally, it is good to have them private. But if you are not sure, I guess it is better to keep their visibility as it was before.

src/prop/bvminisat/bvminisat.h
```diff
@@ -117,7 +117,7 @@ class BVMinisatSatSolver : public BVSatSolverInterface,
 
   void popAssumption() override;
 
-  void setProofLog(BitVectorProof* bvp) override;
+  void setProofLog(proof::ResolutionBitVectorProof* bvp) override;
```
4tXJ7f: Add/update include.

src/proof/resolution_bitvector_proof.h
```diff
@@ -0,0 +1,129 @@
...
+   * or initializing a BNF proof
+   */
+  void initSatProof(CVC4::BVMinisat::Solver* solver);
+
+  BVSatProof* getSatProof();
+
+  /**
+   * Kind of a mess.
+   * In eager mode this must be invoked before printing a proof of the empty
+   * clause. In lazy mode the behavior is ???
```
4tXJ7f: Please add an explicit `TODO` to clean this up (to make it easily grep-able).

src/Makefile.am
```diff
@@ -167,6 +167,8 @@ libcvc4_la_SOURCES = \
 	proof/proof_output_channel.h \
 	proof/proof_utils.cpp \
 	proof/proof_utils.h \
+	proof/resolution_bitvector_proof.cpp \
+	proof/resolution_bitvector_proof.h \
```
4tXJ7f: This file does not exist anymore after the switch to CMake. To resolve the merge conflict, it is fine to just remove those changes (i.e. nothing else needs to be done instead).

src/proof/bitvector_proof.cpp
```diff
@@ -9,105 +9,44 @@
...
+  Debug("pf::bv") << "BitVectorProof::getBBTermName( " << expr << " ) = bt"
+                  << expr.getId() << std::endl;
+  std::ostringstream os;
+  os << "bt" << expr.getId();
+  return os.str();
 }
 
 void BitVectorProof::registerAtomBB(Expr atom, Expr atom_bb) {
-  Debug("pf::bv") << "BitVectorProof::registerAtomBB( " << atom << ", " << atom_bb << " )" << std::endl;
+  Debug("pf::bv") << "ResolutionBitVectorProof::registerAtomBB( " << atom
```
4tXJ7f: This method hasn't been moved, so the debug output does not need to be changed to say `ResolutionBitVectorProof`.

src/proof/bitvector_proof.cpp
```diff
@@ -119,7 +58,8 @@ void BitVectorProof::registerAtomBB(Expr atom, Expr atom_bb) {
 }
 
 void BitVectorProof::registerTerm(Expr term) {
-  Debug("pf::bv") << "BitVectorProof::registerTerm( " << term << " )" << std::endl;
+  Debug("pf::bv") << "ResolutionBitVectorProof::registerTerm( " << term << " )"
```
4tXJ7f: Ditto.

src/proof/bitvector_proof.cpp
```diff
@@ -147,149 +87,52 @@ void BitVectorProof::registerTerm(Expr term) {
...
-            }
-            if (k == confl.getNumChildren()) {
-              // We couldn't find a match for possibleMatch[j], so not a match
-              matching = false;
-              break;
-            }
-          }
+void BitVectorProof::registerTermBB(Expr term)
+{
+  Debug("pf::bv") << "ResolutionBitVectorProof::registerTermBB( " << term
```
4tXJ7f: Ditto.

src/proof/bitvector_proof.cpp
```diff
@@ -9,105 +9,44 @@
...
-  // If this term gets used in the final proof, we will want to register it. However,
-  // we don't know this at this point; and when the theory proof engine sees it, if it belongs
-  // to another theory, it won't register it with this proof. So, we need to tell the
-  // engine to inform us.
-
-  if (theory::Theory::theoryOf(term) != theory::THEORY_BV) {
-    Debug("pf::bv") << "\tMarking term " << term << " for future BV registration" << std::endl;
-    d_proofEngine->markTermForFutureRegistration(term, theory::THEORY_BV);
-  }
+std::string BitVectorProof::getBBTermName(Expr expr)
```
4tXJ7f: It seems like this method and at least `registerTermBB` were moved, which makes the diff unnecessarily large/complicated. It would be great if this change could be reverted.

src/proof/bitvector_proof.h
```diff
@@ -9,161 +9,194 @@
  ** All rights reserved.  See the file COPYING in the top-level source
  ** directory for licensing information.\endverbatim
  **
- ** \brief Bitvector proof
+ ** \brief Bitvector proof base class
  **
- ** Bitvector proof
+ ** Contains code (e.g. proof printing code) which is common to all bitvector
```
4tXJ7f: I thought that we wanted to move all printing code into a parallel class hierarchy. Am I remembering this wrong or is this going to be done later?

src/proof/bitvector_proof.h
```diff
@@ -9,161 +9,194 @@
...
 typedef std::unordered_map<Expr, Expr, ExprHashFunction> ExprToExpr;
 typedef std::unordered_map<Expr, std::string, ExprHashFunction> ExprToString;
 
-class BitVectorProof : public TheoryProof {
-protected:
+/**
+ * A bitvector proof is best understood as having
+ *
+ *    1. A declaration of a "bitblasted formulas" -- boolean formulas
+ *       that each translations of a BV-literal (a comparison between BVs).
```
4tXJ7f: This sentence does not quite parse.

src/proof/bitvector_proof.h
```diff
@@ -9,161 +9,194 @@
...
+ protected:
+  BitVectorProof(theory::bv::TheoryBV* bv, TheoryProofEngine* proofEngine);
+  virtual ~BitVectorProof(){};
+
+  // Set of BV variables in the input. (e.g. "a" in [ a = 000 ] ^ [ a == 001 ])
   ExprSet d_declarations;
 
-  ExprSet d_usedBB; // terms and formulas that are actually relevant to the proof
+  ExprSet
+      d_usedBB;  // terms and formulas that are actually relevant to the proof
```
4tXJ7f: It looks like the formatting got screwed up here (the declaration is split onto two lines). Maybe move the comment above the line.

src/proof/resolution_bitvector_proof.h
```diff
@@ -0,0 +1,129 @@
...
+ ** Bitvector proof
+ **/
+
+#include "cvc4_private.h"
+
+#ifndef __CVC4__PROOF__RESOLUTIONBITVECTORPROOF_H
+#define __CVC4__PROOF__RESOLUTIONBITVECTORPROOF_H
+
+#include <iostream>
+#include <sstream>
```
4tXJ7f: Instead of including those two headers, we should be able to just include `iosfwd` in the header and then the required headers in the source file. `iostream` is large and slow to compile, so we need to be a bit careful with including it in the header.

src/proof/resolution_bitvector_proof.h
```diff
@@ -0,0 +1,129 @@
...
+#include "context/context.h"
+#include "expr/expr.h"
+#include "proof/bitvector_proof.h"
+#include "proof/theory_proof.h"
+#include "prop/bvminisat/core/Solver.h"
+
+namespace CVC4 {
+
+namespace prop {
+class CnfStream;
```
4tXJ7f: Can we use an include instead of this forward declaration?

src/proof/resolution_bitvector_proof.h
4tXJ7f: The name still doesn't correspond to the style guide (needs underscores between the words).

src/theory/bv/bitblast/eager_bitblaster.cpp
```diff
@@ -19,7 +19,7 @@
 #include "theory/bv/bitblast/eager_bitblaster.h"
 
 #include "options/bv_options.h"
-#include "proof/bitvector_proof.h"
+#include "proof/resolution_bitvector_proof.h"
```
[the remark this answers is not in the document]
alex-ozdemir: Agreed!

src/theory/bv/bitblast/eager_bitblaster.h
```diff
@@ -23,6 +23,7 @@
 
 #include "theory/bv/bitblast/bitblaster.h"
 
+#include "proof/bitvector_proof.h"
```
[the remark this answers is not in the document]
alex-ozdemir: Agreed!

src/theory/bv/bitblast/lazy_bitblaster.cpp
```diff
@@ -19,17 +19,17 @@
 #include "theory/bv/bitblast/lazy_bitblaster.h"
 
 #include "options/bv_options.h"
+#include "proof/proof_manager.h"
+#include "proof/resolution_bitvector_proof.h"
```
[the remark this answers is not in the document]
alex-ozdemir: Agreed!

src/theory/bv/bv_eager_solver.h
```diff
@@ -60,7 +60,7 @@ class EagerBitblastSolver {
   bool d_useAig;
 
   TheoryBV* d_bv;
-  BitVectorProof* d_bvp;
+  proof::ResolutionBitVectorProof* d_bvp;
```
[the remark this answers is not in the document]
alex-ozdemir: Agreed!

src/theory/bv/bv_subtheory.h
```diff
@@ -24,6 +24,11 @@
 #include "theory/theory.h"
 
 namespace CVC4 {
+
+namespace proof {
+class ResolutionBitVectorProof;
```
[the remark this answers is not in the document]
alex-ozdemir: Unfortunately not. `theory_bv.h` depends on `bv_subtheory.h` which depends on `resolution_bitvector_proof.h` which depends on `bitvector_proof.h` which depends on `theory_bv.h`. Gotta break the cycle somewhere.

src/theory/bv/bv_subtheory_bitblast.cpp
```diff
@@ -18,8 +18,8 @@
 #include "decision/decision_attributes.h"
 #include "options/bv_options.h"
 #include "options/decision_options.h"
-#include "proof/bitvector_proof.h"
 #include "proof/proof_manager.h"
+#include "proof/resolution_bitvector_proof.h"
```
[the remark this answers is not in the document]
alex-ozdemir: Yeah, we'll need to forward-declare it there.

src/proof/resolution_bitvector_proof.h
alex-ozdemir: I believe this is outdated. Also, I don't think that the style guide says that there should be spaces between the words. Am I mis-reading it?

alex-ozdemir:
I didn't mean to "review" my own work -- just to comment on Andres's comments.

src/proof/bitvector_proof.cpp
4tXJ7f:
```suggestion
  Debug("pf::bv") << "BitVectorProof::registerAtomBB( " << atom
```

src/proof/bitvector_proof.cpp
4tXJ7f:
```suggestion
  Debug("pf::bv") << "BitVectorProof::registerTerm( " << term << " )"
```

src/proof/bitvector_proof.cpp
4tXJ7f:
```suggestion
  Debug("pf::bv") << "BitVectorProof::registerTermBB( " << term
```

alex-ozdemir:
Thanks for revisiting this. I apologize for missing those comments. I've incorporated your suggested changes. Mind taking another look?

PR #2602, opened 2018-10-09
Fix compiler warnings

4tXJ7f:
This commit fixes two compiler warnings related to parentheses and two
warnings related to derived classes hiding methods from the base class
with the same name.


src/theory/quantifiers/sygus/cegis_unif.h
```diff
@@ -47,6 +47,10 @@ namespace quantifiers {
 class CegisUnifEnumDecisionStrategy : public DecisionStrategyFmf
 {
  public:
+  // Do not hide the zero-argument version of initialize() inherited from the
```
ajreynol: Style question here: should this go beneath the constructor? Also, the comment maybe should be /** */?

src/theory/strings/theory_strings.h
```diff
@@ -728,6 +728,10 @@ class TheoryStrings : public Theory {
   class StringSumLengthDecisionStrategy : public DecisionStrategyFmf
   {
    public:
+    // Do not hide the zero-argument version of initialize() inherited from the
```
ajreynol: Ditto

src/theory/quantifiers/sygus/cegis_unif.h
4tXJ7f: Are you talking about moving the comment or the `using` line as well? Maybe moving it next to the other version of `initialize()` would be a good idea. The `/** */` comments are actually Doxygen comments. Maybe I can use `/* */` though.

ajreynol:
I’m talking about moving the using line as well.
And yes I think I’d prefer having a /* */ comment.

4tXJ7f:
Alright, I addressed the comments, PTAL.

PR #2605, opened 2018-10-09
Fix default setting of CegisUnif options

HanielB:


src/smt/smt_engine.cpp
```diff
@@ -1131,6 +1131,16 @@ void SmtEngine::setDefaults() {
         options::cbqi.set(true);
       }
     }
+    // setting unif requirements
+    if (options::sygusUnifBooleanHeuristicDt()
```
ajreynol: Can you merge master and then move this (and the above options setting in this block) to https://github.com/CVC4/CVC4/blob/master/src/smt/smt_engine.cpp#L1809 ?

src/smt/smt_engine.cpp
```diff
@@ -1812,6 +1800,28 @@ void SmtEngine::setDefaults() {
...
+      if (!options::cbqi.wasSetByUser())
+      {
+        options::cbqi.set(true);
+      }
+    }
+    // setting unif requirements
+    if (options::sygusUnifBooleanHeuristicDt()
+        && !options::sygusUnifCondIndependent())
+    {
+      setOption("sygus-unif-cond-independent", SExpr("true"));
```
ajreynol: this should just be sygusUnifCondIndependent.set(true).

src/smt/smt_engine.cpp
```diff
@@ -1812,6 +1800,28 @@ void SmtEngine::setDefaults() {
...
+    }
+    if (options::sygusUnifCondIndependent() && !options::sygusUnif())
+    {
+      setOption("sygus-unif", SExpr("true"));
```
ajreynol: Ditto

PR #2606, opened 2018-10-09
 Support for basic actively-generated enumerators

ajreynol:
This adds a new sygus enumeration strategy "basic actively-generated enumerators" that (naively) uses a type enumerator instead of the sygus solver for generating candidates.

This PR also cleans/fixes a number of related issues, in particular regarding multiple synthesis conjectures:
- The "active guard" of actively-generated enumerators is managed centrally by SynthConjecture instead of by the sygus modules,
- The "excluded by symmetry breaking" attribute is flipped so that we catch cases where the enumerator has yet to be registered by the datatypes sygus solver,
- Interleave doCheck calls for multiple synthesis conjectures. This ensures that we are fair for basic actively-generated enumerators of infinite type when multiple conjectures are present.

### DIFF
--- a/src/theory/quantifiers/sygus/synth_conjecture.cpp
+++ b/src/theory/quantifiers/sygus/synth_conjecture.cpp
@@ -653,7 +653,7 @@ bool SynthConjecture::getEnumeratedValues(std::vector<Node>& n,
   std::vector<Node> ncheck = n;
   n.clear();
   bool ret = true;
-  for (unsigned i = 0; i < ncheck.size(); i++)
+  for (unsigned i = 0, size = ncheck.size(); i < size; i++)
   {
     Node e = ncheck[i];
     // if it is not active, we return null
@@ -663,6 +663,8 @@ bool SynthConjecture::getEnumeratedValues(std::vector<Node>& n,
       Node gstatus = d_qe->getValuation().getSatValue(g);
       if (gstatus.isNull() || !gstatus.getConst<bool>())
       {
+        Trace("cegqi-engine-debug")
+            << "Enumerator " << e << " is inactive." << std::endl;
         continue;
       }
     }
@@ -725,15 +727,20 @@ class EnumValGeneratorBasic : public EnumValGenerator
 
 Node SynthConjecture::getEnumeratedValue(Node e)
 {
-  if (!e.getAttribute(SygusSymBreakOkAttribute()))
+  bool isEnum = d_tds->isEnumerator(e);
+
+  if (isEnum && !e.getAttribute(SygusSymBreakOkAttribute()))
   {
     // if the current model value of e was not registered by the datatypes
     // sygus solver, or was excluded by symmetry breaking, then it does not
     // have a proper model value that we should consider, thus we return null.
+    Trace("cegqi-engine-debug")
+        << "Enumerator " << e << " does not have proper model value."
+        << std::endl;
     return Node::null();
   }
 
-  if (!d_tds->isEnumerator(e) || d_tds->isPassiveEnumerator(e))
+  if (!isEnum || d_tds->isPassiveEnumerator(e))
   {
     return getModelValue(e);
   }
--- a/src/theory/quantifiers/sygus/synth_conjecture.cpp
+++ b/src/theory/quantifiers/sygus/synth_conjecture.cpp
@@ -253,7 +253,8 @@ bool SynthConjecture::needsCheck()
     }
     else
     {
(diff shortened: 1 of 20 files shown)


src/theory/quantifiers/sygus/term_database_sygus.h
```diff
@@ -69,10 +69,8 @@ class TermDbSygus {
    * (see d_enum_to_active_guard),
    * useSymbolicCons : whether we want model values for e to include symbolic
    * constructors like the "any constant" variable.
-   * isVarAgnostic : if this flag is true, the enumerator will only generate
-   * values whose variables are in canonical order (for example, only x1-x2
-   * and not x2-x1 will be generated, assuming x1 and x2 are in the same
-   * "subclass", see getSubclassForVar).
+   * isActiveGen : if this flag is true, the enumerator will be
+   * actively-generated based on the mode specified by --sygus-active-gen.
```
HanielB: maybe keep the example? It's good to have some examples of what active gen enums do.

src/theory/quantifiers/sygus/term_database_sygus.cpp
```diff
@@ -371,7 +371,9 @@ void TermDbSygus::registerSygusType( TypeNode tn ) {
             Trace("sygus-db") << ", kind = " << sk;
             d_kinds[tn][sk] = i;
             d_arg_kind[tn][i] = sk;
-          }else if( sop.isConst() ){
+          }
+          else if (sop.isConst() && dt[i].getNumArgs() == 0)
```
HanielB: I'm a bit puzzled by this. When `sop.isConst()` is true shouldn't we consider it a constant even if the constructor has arguments?

src/theory/quantifiers/sygus/synth_conjecture.h
```diff
@@ -83,9 +83,19 @@ class SynthConjecture
...
   void doSingleInvCheck(std::vector<Node>& lems);
   /** do syntax-guided enumerative check
+   *
    * This is step 2(a) of Figure 3 of Reynolds et al CAV 2015.
+   *
+   * The method returns true if this conjecture is finished trying solutions
+   * for the given call to SynthEngine::check.
+   *
+   * Notice that we make multiple calls to doCheck on one call to
+   * SynthEngine::check. For example, if we are using an  actively-generated
```
HanielB: double space after "an"

src/theory/quantifiers/sygus/synth_conjecture.h
```diff
@@ -83,9 +83,19 @@ class SynthConjecture
...
+   * enumerator, one enumerated (abstract) term may correspond to multiple
+   * concrete terms t1, ..., tn to check, where we make up to n calls to doCheck
+   * when each of t1, ..., tn fail to satisfy the current refinement lemmas.
```
HanielB: fail -> fails

src/theory/quantifiers/sygus/synth_conjecture.cpp
```diff
@@ -662,38 +650,116 @@ void SynthConjecture::preregisterConjecture(Node q)
...
+ * the type enumerator to generate a stream of (all) sygus terms of a given
+ * type.
+ */
+class EnumValGeneratorBasic : public EnumValGenerator
+{
+ public:
+  EnumValGeneratorBasic(TermDbSygus* tds, TypeNode tn) : d_tds(tds), d_te(tn) {}
+  ~EnumValGeneratorBasic() {}
+  /** initialize (do nothing) */
+  void initialize(Node e) override {}
```
HanielB: isn't this spurious?

src/theory/quantifiers/sygus/synth_conjecture.cpp
```diff
@@ -662,38 +650,116 @@ void SynthConjecture::preregisterConjecture(Node q)
...
+  /** initialize (do nothing) */
+  void addValue(Node v) override {}
```
HanielB: ditto

src/theory/quantifiers/sygus/term_database_sygus.cpp
ajreynol: One example where this was wrong is BV extract operators, whose operators are constant, not builtin, but applications of these are not constants.

src/theory/quantifiers/sygus/synth_conjecture.cpp
ajreynol: The base class's initialize is pure virtual, so this is necessary.

PR #2607, opened 2018-10-10
Skip invalid objects in ContextObj::destroy()

4tXJ7f:
Background
==========

Context-dependent objects store their older version in the pointer
`d_pContextObjRestore`. Additionally, each object is part of a
doubly-linked list of objects in the same scope (i.e. the same level).
When a scope is popped, all objects in the linked list look up their
previous version and insert themselves into the linked list at the level
of the previous version. If an object does not have an older version,
the `d_pContextObjRestore` pointer is `null` and restoring the older
version does not do anything.

When we destroy an object (which should be called from the destructor of
a context-dependent object), remove the object from all scopes by
repeatedly removing it from the current scope, restoring the older
version, removing it from the older scope and so on.

A side note: The doubly-linked lists used are somewhat unusual: Instead
of holding a pointer to the previous object, the context-dependent
objects store a pointer to the next pointer of the previous object.

Problem
=======

The mechanisms described above may be problematic under rare
circumstances. Specifically, this came up because Clang 6.0's ASAN
detected a stack-user-after-scope issue in the
`testTopScopeContextObj()` unit test in context_black.h. Note that Clang
5.0's ASAN was not able to detect this issue. In this unit test the
following happened:

1. We push our context.
2. We create two context-dependent objects: `x` at level 1 (by setting
`allocatedInCMM` to true when calling the `ContextObj` constructor) and
`y` at level 0. At this point, scope 0 has `y` in its linked list and
scope 1 has `x` in its linked list.
3. We call `makeCurrent()` on both objects. This does nothing for `x`
because it already is at level 1. For `y` it creates a backup of the
current information and inserts `y` into the scope 1. Scope 0 now has a
backup of `y` in its linked list and scope 1 has both `x` and `y` in its
linked list. Note that `x`'s `prev()` now points to `y`'s `next()`.
4. We do an additional push, another call to `makeCurrent` on both
objects, and a first `pop()`. After those operations, we have the same
state as after 3.
5. We pop our context again, which restores the original `y` in scope 0.
We now only have scope 0 with `y` in it. As described above, `x` is
untouched because there is no older version to restore.
6. We are at the end of the function and because `y` was created after
`x`, it is destroyed first (standard destruction order of stack objects
in C++). When destroying `x`, `x`'s `prev()` still points to `y`'s
`next()`, so we are doing an invalid write and ASAN complains.

Solution
========

The solution that this commit implements is fairly simple. We can detect
that `x`'s `prev()`/`next()` pointers are stale by checking whether the
objects level is greater than the context's level. If it is, we know
that this object is invalid and that we do not have any old versions to
remove. Note that this situation should be very rare because objects
with `allocatedInCMM` should usually be allocated in context memory,
where they get cleaned up automatically when the context is popped below
the level that they are associated with.

@timothy-king I am adding you to this PR since you may be interested in it.

### DIFF
--- a/src/context/context.cpp
+++ b/src/context/context.cpp
@@ -209,6 +209,7 @@ ContextObj* ContextObj::restoreAndContinue()
 
     Debug("context") << "NULL restore object! " << this << std::endl;
     pContextObjNext = d_pContextObjNext;
+    d_pScope = nullptr;
 
     // Nothing else to do
   } else {
@@ -237,11 +238,6 @@ ContextObj* ContextObj::restoreAndContinue()
 
 void ContextObj::destroy()
 {
-  /* Context can be big and complicated, so we only want to process this output
-   * if we're really going to use it. (Same goes below.) */
-  Debug("context") << "before destroy " << this << " (level " << getLevel()
-                   << "):" << std::endl << *getContext() << std::endl;
-
   // Under rare circumstances, we could be trying to destroy an object that is
   // invalid. In that case, the prev()/next() pointers are stale and must not
   // be used. Since invalid objects have no old versions to restore and remove
@@ -251,13 +247,18 @@ void ContextObj::destroy()
   // when calling the ContextObj constructor) and kept on the stack, we pop
   // down to level 0 (the object keeps level 1 because it does not exist below
   // that level), and at the end of the function the object is destroyed.
-  if (getLevel() > getContext()->getLevel())
+  if (d_pScope == nullptr)
   {
     Assert(d_pContextObjRestore == NULL);
     Debug("context") << "skipping destroy because object already invalid"
                      << std::endl;
     return;
   }
+
+  /* Context can be big and complicated, so we only want to process this output
+   * if we're really going to use it. (Same goes below.) */
+  Debug("context") << "before destroy " << this << " (level " << getLevel()
+                   << "):" << std::endl << *getContext() << std::endl;
 
   for(;;) {
     // If valgrind reports invalid writes on the next few lines,
--- a/src/context/context.cpp
+++ b/src/context/context.cpp
@@ -242,6 +242,23 @@ void ContextObj::destroy()
   Debug("context") << "before destroy " << this << " (level " << getLevel()
                    << "):" << std::endl << *getContext() << std::endl;
 
+  // Under rare circumstances, we could be trying to destroy an object that is
+  // invalid. In that case, the prev()/next() pointers are stale and must not
+  // be used. Since invalid objects have no old versions to restore and remove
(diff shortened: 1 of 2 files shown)


timothy-king:
Adding Clark as a reviewer. He also knows a lot about the context code.

1. Outside of the unit test mentioned, can this happen?
2. What is the measured performance penalty?
3. What has been tested with ASAN? I would suggest a significant fraction of SMT-LIB.
4. What are the lifetimes of the variables used by getLevel() and getContext() this uses?
5. Will the proposed level based check always be correct? Will is always be safe to call?

4tXJ7f:
> Adding Clark as a reviewer. He also knows a lot about the context code.

Good idea, yes.

> 1. Outside of the unit test mentioned, can this happen?

I don't think that we necessarily want it to happen, no. We could turn the negation of the condition into an assertion and add a unit test that checks that this assertion is triggered in such a case. The reason why I didn't turn it into an assertion was because there is a unit test and I thought that maybe we do want to support this kind of pattern. We definitely don't want the status quo, though, because we could potentially silently corrupt data.

> 2. What is the measured performance penalty?

I did not measure the performance penalty but I would expect it to be negligible. Once we agree on a solution, I am happy to run on some benchmarks to confirm that.

> 3. What has been tested with ASAN? I would suggest a significant fraction of SMT-LIB.

The usual regression levels, 0 through 2, and the unit tests. We didn't run with ASAN on SMT-LIB for a while... it would probably be a good idea to run it with and without this patch.

> 4. What are the lifetimes of the variables used by getLevel() and getContext() this uses?
> 5. Will the proposed level based check always be correct? Will is always be safe to call?

This actually brings up a good point. `getLevel()` and `getContext()` depend on `d_pScope`, which is actually already destroyed at that time (I'll double check that though), so we may have to use another solution. Another issue with the current solution may be that if we pop the context until we invalidate the object, then push, and then the object goes out of scope, the check would fail to detect that the object is invalid. One possible solution may be to have an explicit `d_invalid` flag in ContextObj that gets set once we pop below an object's scope. Thoughts?

4tXJ7f:
Alright, I checked with the debug context memory manager and confirmed that `getLevel()` and `getContext()` are really problematic because `d_pScope` does not exist anymore.

4tXJ7f:
I have updated the fix to set `d_pScope` to `nullptr` when an object is invalidated and check for that in `destroy()`. This should fix both issues mentioned above (invalid accesses to `d_pScope` and check that would fail to detect an invalid object). I also seems quite logical because an invalid object is not associated with a scope anymore. Note: I didn't update the comments yet to reflect the solution.

PR #2608, opened 2018-10-10
Synthesize rewrite rules from inputs

ajreynol:
This implements a new feature `--sygus-rr-synth-input` which synthesizes rewrite rules that target the input formula. It does so using a sygus grammar construction that encodes "abstractions of subterms of the input".

This is a major refactor of the original pass `--synth-rr-prep` which searched for equivalences between the (verbatim) subterms of the input.

It is recommended that `--sygus-active-gen=basic` is used in combination with this option currently, for performance reasons.


src/preprocessing/passes/synth_rew_rules.cpp
```diff
@@ -18,143 +18,436 @@
...
+  // Get all usable terms from the input. A term is usable if it does not
+  // contain a quantified subterm
+  std::vector<Node> terms;
+  // all variables (free constants) appearing in the input
+  std::vector<Node> vars;
+
+  // We will generate a fixed number of variables per type. These are the
+  // variables that appear as free variables in the rewrites we generate.
+  unsigned nvars = options::sygusRewSynthInputNVars();
+  // must have at least one variable per type (e.g. true, false for Bool)
```
HanielB: say this also in the header

src/options/quantifiers_options.toml
```diff
@@ -1316,6 +1316,30 @@ header = "options/quantifiers_options.h"
...
+  type       = "bool"
+  default    = "false"
+  help       = "synthesize rewrite rules based on the input formula"
+
+[[option]]
+  name       = "sygusRewSynthInputNVars"
+  category   = "regular"
+  long       = "sygus-rr-synth-input-nvars=N"
+  type       = "int"
+  default    = "3"
```
HanielB: why is this the default?

src/preprocessing/passes/synth_rew_rules.cpp
```diff
@@ -18,143 +18,436 @@
...
-          if (childrenValid)
+          // register type information
+          TypeNode tn = cur.getType();
+          if (tvars.find(tn) == tvars.end())
           {
-            Trace("synth-rr-pass-debug")
-                << "...children are valid, check rewrites..." << std::endl;
-            if (r == 0)
+            // Only make one Boolean variable unless option is set. This ensures
+            // we do not compute purely Boolean rewrites by default.
```
HanielB: what is the motivation for this?

src/preprocessing/passes/synth_rew_rules.cpp
```diff
@@ -18,143 +18,436 @@
...
+        // the operator of each constructor is a no-op
+        Node tbv = nm->mkBoundVar(ctt);
+        Expr lambdaOp =
+            nm->mkNode(LAMBDA, nm->mkNode(BOUND_VAR_LIST, tbv), tbv).toExpr();
+        std::vector<Type> argListc;
+        // the following construction admits any number of repeated factors,
+        // so for instance, t1+t2+t3, we generate the grammar:
+        // T_{t1+t2+t3} -> 
+        //   +( T_{t1+t2+t3}, T_{t1+t2+t3} ) | T_{t1} | T_{t2} | T_{t3}
+        // where we write T_t to denote "the type that abstracts term t.
```
HanielB: put closing quote

src/theory/quantifiers/sygus/sygus_grammar_norm.cpp
```diff
@@ -158,12 +158,20 @@ void SygusGrammarNorm::TypeObject::addConsInfo(SygusGrammarNorm* sygus_norm,
   Node sygus_op = Node::fromExpr(cons.getSygusOp());
   Trace("sygus-grammar-normalize-debug")
       << ".....operator is " << sygus_op << std::endl;
-  Node exp_sop_n = Node::fromExpr(
-      smt::currentSmtEngine()->expandDefinitions(sygus_op.toExpr()));
+  Node exp_sop_n = sygus_op;
+  // Only expand definitions if the operator is not constant. This check is
+  // required to ensure we don't try to expand, e.g. bitvector extract
+  // operators, whose type is undefined.
+  if (!sygus_op.isConst())
```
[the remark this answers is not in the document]
ajreynol: No, things that don't have a type cannot be expanded (who expands a term is determined by its type).  It just happens that indexed operators do not have a type rule, although perhaps they should. Regardless, this avoids these cases.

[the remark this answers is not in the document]
HanielB: What confuses me is why are you checking the property of "not having a type" by checking that its sygus operator is not constant?

src/theory/quantifiers/sygus/sygus_grammar_cons.cpp
```diff
@@ -126,21 +126,18 @@ Node CegGrammarConstructor::process(Node q,
...
     std::stringstream ss;
     ss << sf;
-    if (preGrammarType.isDatatype()
-        && static_cast<DatatypeType>(preGrammarType.toType())
-               .getDatatype()
-               .isSygus())
+    Node sfvl;
+    if (preGrammarType.isDatatype() && preGrammarType.getDatatype().isSygus())
     {
+      sfvl = Node::fromExpr(preGrammarType.getDatatype().getSygusVarList());
```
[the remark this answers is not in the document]
ajreynol: This makes the code more robust. Previously, we assumed that the user had to set a user attribute SygusSynthFunVarListAttribute for *any* sygus function. However, this is redundant if it has SygusSynthGrammarAttribute, and moreover we are assuming that SygusSynthFunVarListAttribute was set to the same variable list as SygusSynthGrammarAttribute. This takes the variable list in the datatype specified by SygusSynthGrammarAttribute as the authority.

src/smt/smt_engine.cpp
```diff
@@ -1850,6 +1852,18 @@ void SmtEngine::setDefaults() {
       options::sygusRewSynth.set(true);
       options::sygusRewVerify.set(true);
     }
+    if (options::sygusRewSynthInput())
+    {
+      // if we are using synthesis rewrite rules from input, we use
+      // sygusRewSynth after preprocessing.
```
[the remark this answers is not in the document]
ajreynol: This should be documented with the class definition in my opinion and not here.

src/options/quantifiers_options.toml
ajreynol: this is a tradeoff between efficiency and generality, 2 is not general enough, 4 is too inefficient.

src/preprocessing/passes/synth_rew_rules.cpp
ajreynol: We don't want any input to generate Boolean rewrites, e.g. (or A A) -> A, (and A A) -> A, etc.

src/smt/smt_engine.cpp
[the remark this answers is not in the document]
HanielB: I think it does not hurt to help the person reading this to have an idea of why this is done without needing to understand things globally (as was my case when I first read the comment)

src/options/quantifiers_options.toml
HanielB: This should be said somewhere, probably in the header of the class using the option

src/preprocessing/passes/synth_rew_rules.cpp
HanielB: Again, this should be said somewhere, probably in the header of the class. It's a simple thing that is clear once said but a bit mysterious if you are not used to it.

src/theory/quantifiers/sygus/sygus_grammar_norm.cpp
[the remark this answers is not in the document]
ajreynol: I'm not. It just happens that if you are not a constant, you definitely have a type. Regardless, we should not be expanding constants anyways (if they are constant, they should not expand).

src/smt/smt_engine.cpp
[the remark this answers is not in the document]
ajreynol: Ive added a reference.

src/theory/quantifiers/sygus/sygus_grammar_norm.cpp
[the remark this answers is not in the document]
HanielB: That was my original question :) Can you clarify a bit the comment to say that?

src/preprocessing/passes/synth_rew_rules.cpp
ajreynol: This has been addressed, see line 48 synth_rew_rules.h

PR #2611, opened 2018-10-10
configure.sh: Fix option parsing to match --help

4tXJ7f:


configure.sh
```diff
@@ -29,7 +29,8 @@ The following flags enable optional features (disable with --no-<option name>).
   --optimized              optimize the build
   --debug-symbols          include debug symbols
   --valgrind               Valgrind instrumentation
-  --debug-context-mm       use the debug context memory manager
+  --debug-context-memory-manager
```
mpreiner: This option name is very long, should we rename the option to `--debug-context-mm` and also rename all the other occurrences of it?

aniemetz: This option is named debug-context-mm on purpose since the old name (debug-context-memory-manager) is way to long. I don't want this to be renamed.

aniemetz: Oh, I see, I forgot to rename the other occurrences. I'd suggest to maybe leave the name as is and rename the other occurrences?

4tXJ7f: Sure, I am fine with that :) I just noticed that the help didn't match the actual name when I tried to use it.

4tXJ7f:
I've finally addressed the comment(s) :)

PR #2615, opened 2018-10-10
Improve reasoning about empty strings in rewriter

4tXJ7f:
This commit adds several rewrites related to reasoning about empty
strings: it strengthens the rewrites for `(= "" _)` and it adds a
rewrite for `(str.replace "" y z)` that simplifies `y` if `(str.contains
"" y)` rewrites to a conjunction of empty-string equalities.

### DIFF
--- a/src/theory/strings/theory_strings_rewriter.cpp
+++ b/src/theory/strings/theory_strings_rewriter.cpp
@@ -2545,8 +2545,11 @@ Node TheoryStringsRewriter::rewriteReplace( Node node ) {
       {
         std::vector<Node> emptyNodesList(emptyNodes.begin(), emptyNodes.end());
         Node nn1 = mkConcat(STRING_CONCAT, emptyNodesList);
-        Node res = nm->mkNode(kind::STRING_STRREPL, node[0], nn1, nn2);
-        return returnRewrite(node, res, "rpl-emp-cnts-substs");
+        if (nn1 != node[1] || nn2 != node[2])
+        {
+          Node res = nm->mkNode(kind::STRING_STRREPL, node[0], nn1, nn2);
+          return returnRewrite(node, res, "rpl-emp-cnts-substs");
+        }
       }
 
       if (nn2 != node[2])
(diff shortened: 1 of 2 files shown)


src/theory/strings/theory_strings_rewriter.cpp
```diff
@@ -512,9 +519,17 @@ Node TheoryStringsRewriter::rewriteStrEqualityExt(Node node)
       {
         Node zero = nm->mkConst(Rational(0));
 
-        // (= "" (str.substr x n m)) ---> (<= (str.len x) n) if n >= 0 and m > 0
         if (checkEntailArith(ne[1], false) && checkEntailArith(ne[2], true))
         {
+          // (= "" (str.substr x 0 m)) ---> "" if m > 0
```
ajreynol: I think you mean (= x "") ?

4tXJ7f: Yup, fixed :)

src/theory/strings/theory_strings_rewriter.cpp
```diff
@@ -2514,6 +2534,21 @@ Node TheoryStringsRewriter::rewriteReplace( Node node ) {
...
+      // be empty:
+      //
+      // (str.replace "" y z) ---> (str.replace "" (str.++ y1 ... yn)  z)
+      // if (str.contains "" y) ---> (and (= y1 "") ... (= yn ""))
+      if (node[0] == empty && allEqs)
+      {
+        std::vector<Node> emptyNodesList(emptyNodes.begin(), emptyNodes.end());
+        Node nn1 = mkConcat(STRING_CONCAT, emptyNodesList);
+        Node res = nm->mkNode(kind::STRING_STRREPL, node[0], nn1, nn2);
+        return returnRewrite(node, res, "rpl-emp-cnts-substs");
```
ajreynol: Do we ensure that res != node here?  What if y is a CONCAT?

4tXJ7f: Good point, done.

src/theory/strings/theory_strings_rewriter.cpp
```diff
@@ -2504,6 +2520,10 @@ Node TheoryStringsRewriter::rewriteReplace( Node node ) {
             emptyNodes.insert(c[0]);
           }
         }
+        else
```
ajreynol: Should we also set allEqs to false if we are EQUAL but neither side is empty?  Perhaps this should be allEmptyEqs?

4tXJ7f: Yeah, good point. Done.

PR #2618, opened 2018-10-11
 Fix string ext inference for rewrites that introduce negation

### LINKED ISSUE #2616: Strings input does not terminate when reaching timeout
When running the following input, it does not terminate after 3 seconds when using the options `--strings-fmf --strings-exp --tlimit 3000`:

```
(set-logic SLIA)
(declare-const x String)
(declare-const y String)
(assert (not (= (str.replace "A" (str.replace x "A" y) x) (str.replace "A" x (str.replace x y x)))))
(check-sat)
```

ajreynol:
Fixes #2616.

### DIFF
--- a/src/theory/strings/theory_strings.cpp
+++ b/src/theory/strings/theory_strings.cpp
@@ -1778,18 +1778,20 @@ void TheoryStrings::checkExtfInference( Node n, Node nr, ExtfInfoTmp& in, int ef
           Node conc =
               nm->mkNode(STRING_STRCTN, pol ? nr[1] : onr, pol ? onr : nr[1]);
           conc = Rewriter::rewrite(conc);
+          conc = conc.negate();
           bool do_infer = false;
+          bool pol = conc.getKind()!=NOT;
+          Node lit = pol ? conc : conc[0];
           if (conc.getKind() == EQUAL)
           {
-            do_infer = !areDisequal(conc[0], conc[1]);
+            do_infer = pol ? !areEqual(conc[0], conc[1]) : !areDisequal(conc[0], conc[1]);
           }
           else
           {
-            do_infer = !areEqual(conc, d_false);
+            do_infer = !areEqual(conc, pol ? d_true : d_false);
           }
           if (do_infer)
           {
-            conc = conc.negate();
             std::vector<Node> exp_c;
             exp_c.insert(exp_c.end(), in.d_exp.begin(), in.d_exp.end());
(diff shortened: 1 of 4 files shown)


src/theory/strings/theory_strings.cpp
```diff
@@ -1778,18 +1778,21 @@ void TheoryStrings::checkExtfInference( Node n, Node nr, ExtfInfoTmp& in, int ef
...
           conc = Rewriter::rewrite(conc);
+          conc = conc.negate();
           bool do_infer = false;
+          bool pol = conc.getKind() != NOT;
+          Node lit = pol ? conc : conc[0];
           if (conc.getKind() == EQUAL)
           {
-            do_infer = !areDisequal(conc[0], conc[1]);
+            do_infer = pol ? !areEqual(conc[0], conc[1])
+                           : !areDisequal(conc[0], conc[1]);
```
4tXJ7f: Shouldn't this be `lit[0]` and `lit[1]` instead of `conc[0]` and `conc[1]`?

src/theory/strings/theory_strings.cpp
```diff
@@ -1778,18 +1778,21 @@ void TheoryStrings::checkExtfInference( Node n, Node nr, ExtfInfoTmp& in, int ef
           Node conc =
               nm->mkNode(STRING_STRCTN, pol ? nr[1] : onr, pol ? onr : nr[1]);
           conc = Rewriter::rewrite(conc);
+          conc = conc.negate();
           bool do_infer = false;
+          bool pol = conc.getKind() != NOT;
+          Node lit = pol ? conc : conc[0];
           if (conc.getKind() == EQUAL)
```
4tXJ7f: Shouldn't this be `lit.getKind()`?

src/theory/strings/theory_strings.cpp
```diff
@@ -1778,18 +1778,21 @@ void TheoryStrings::checkExtfInference( Node n, Node nr, ExtfInfoTmp& in, int ef
...
           {
-            do_infer = !areDisequal(conc[0], conc[1]);
+            do_infer = pol ? !areEqual(conc[0], conc[1])
+                           : !areDisequal(conc[0], conc[1]);
           }
           else
           {
-            do_infer = !areEqual(conc, d_false);
+            do_infer = !areEqual(conc, pol ? d_true : d_false);
```
4tXJ7f: Shouldn't that be `pol ? d_false : d_true`? `conc` should be not equal to `d_false` if it is positive.

PR #2623, opened 2018-10-12
 Add rewrites for str.replace in str.contains 

4tXJ7f:
This commit adds two rewrites for `(str.contains (str.replace x y x) z) ---> (str.contains x z)`, either when `z = y` or `(str.len z) <= 1`. Additionally, the commit adds `(str.contains (str.replace x y z) w) ---> true` if `(str.contains x w) --> true` and `(str.contains z w) ---> true`.

### DIFF
--- a/test/unit/theory/theory_strings_rewriter_white.h
+++ b/test/unit/theory/theory_strings_rewriter_white.h
@@ -80,6 +80,20 @@ class TheoryStringsRewriterWhite : public CxxTest::TestSuite
     TS_ASSERT_DIFFERS(res_t1, res_t2);
   }
 
+  void testCheckEntailArith()
+  {
+    TypeNode intType = d_nm->integerType();
+    TypeNode strType = d_nm->stringType();
+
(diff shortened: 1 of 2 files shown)


test/unit/theory/theory_strings_rewriter_white.h
```diff
@@ -80,6 +80,21 @@ class TheoryStringsRewriterWhite : public CxxTest::TestSuite
...
+  void testCheckEntailArith()
+  {
+    TypeNode intType = d_nm->integerType();
+    TypeNode strType = d_nm->stringType();
+
+    Node z = d_nm->mkVar("z", strType);
+    Node n = d_nm->mkVar("n", intType);
+    Node one = d_nm->mkConst(Rational(1));
+
+    // (str.len (str.substr z n 1)) >= 1 ---> true
```
ajreynol: I think this is `<=` not `>=`.

4tXJ7f: Yup, you are right, fixed it.

PR #2624, opened 2018-10-12
Reset input language for ExprMiner subsolver

4tXJ7f:
Currently, the subsolver created by the `ExprMiner` copies all the
options from the main solver, including the input language. When
synthesizing rewrites and checking them with the subsolver, this is
problematic because the default decision engine used by SyGuS is
`--decision=internal` but we want `--decision=justification` for solving
string queries for example. This commit changes the `ExprMiner` to set
the input language of the subsolver to `smt2`, s.t.
`SmtEngine::setDefaults()` chooses the correct decision engine.

### DIFF
--- a/src/theory/quantifiers/expr_miner.cpp
+++ b/src/theory/quantifiers/expr_miner.cpp
@@ -91,6 +91,7 @@ void ExprMiner::initializeChecker(std::unique_ptr<SmtEngine>& checker,
       checker->setTimeLimit(options::sygusExprMinerCheckTimeout(), true);
       checker->setLogic(smt::currentSmtEngine()->getLogicInfo());
       checker->setOption("sygus-rr-synth-input", false);
+      checker->setOption("input-language", "smt2");
       Expr equery = squery.toExpr().exportTo(&em, varMap);
       checker->assertFormula(equery);
     }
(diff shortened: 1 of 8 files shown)


src/theory/quantifiers/expr_miner.cpp
```diff
@@ -91,6 +91,7 @@ void ExprMiner::initializeChecker(std::unique_ptr<SmtEngine>& checker,
       checker->setTimeLimit(options::sygusExprMinerCheckTimeout(), true);
       checker->setLogic(smt::currentSmtEngine()->getLogicInfo());
       checker->setOption("sygus-rr-synth-input", false);
+      checker->setOption("input-language", "smt2");
```
ajreynol: Should we drop the condition on line 80? At the moment the behavior is not inuitive: if you set a timeout or the other option is set, we setup the proper defaults, otherwise we dont.

4tXJ7f:
Alright, after our discussion offline, I've added the option `--sygus-expr-miner-check-use-export` (true by default), which decides whether we use a separate `ExprManager` for our subsolver. We usually want to do this because it allows us to set separate options for the subsolver and avoids any potential issues from the subsolver setting options, e.g. during `SmtEngine::setDefaults()`, that are then also applied to the main solver. Some sorts unfortunately do not allow exporting right now (primarily datatypes) but it should be fine for most sorts that we are interested in. I've also added regressions for FP rewrite synthesis to make sure that we are not breaking their export.

ajreynol:
The new policy looks good to me.

PR #2625, opened 2018-10-12
Improvements to rewrite rules from inputs

ajreynol:
This includes:
- Not including Boolean variables in rewrites if the input does not have any.
- An option to filter "non-linear" rewrites --sygus-rr-synth-filter-nl. When this option is enabled, we only print rewrites where at least one side of the rewrite has a term with no repeated variables. @4tXJ7f , you may be interested in this.
- Better ordering so that constants (e.g. true) are preferred to others.

### DIFF
--- a/src/expr/node_algorithm.cpp
+++ b/src/expr/node_algorithm.cpp
@@ -63,6 +63,55 @@ bool hasSubterm(TNode n, TNode t, bool strict)
   return false;
 }
 
+bool hasSubtermMulti(TNode n, TNode t)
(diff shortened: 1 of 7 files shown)


HanielB:
> This, although not guaranteed (say if the decision heuristic was highly random), should make constants come first:
> https://github.com/CVC4/CVC4/pull/2625/files#diff-ab72d2ada0745ee9aa540ef1753cd34bR129

Cool. Please add a comment in the code.

PR #2633, opened 2018-10-15
Lazy model construction in TheoryEngine

ajreynol:
This commit makes model construction in TheoryEngine lazy. The motivation is to optimize our performance on incremental benchmarks (e.g. Kind2) where the user only queries for model values after a strict subset of the check-sat commands.

This PR:
- Splits TheoryEngine::getModel to two functions. (1) getModel, which is used to get a pointer to the model object of theory engine, and (2) getBuiltModel, which should be called by callees that want TheoryEngine to ensure that the model is built. The latter is called by SmtEngine only at the moment.
- Combines PropEngine::resetTrail, TheoryEngine::postsolve into doPendingPops. This is required to ensure that we maintain the proper state information after check-sat calls so that model building is available as long as possible.
- Removes a call to doPendingPops during addToModelCommandAndDump.

With this PR, we are roughly 3x faster on many of the Kind2 incremental benchmarks where CVC4 is reported to be much slower than z3/yices.

I'm marking this as complex since it'd be good to have multiple reviews.

@barrettcw , you may be interested in this PR.


ajreynol:
One open issue I have some concerns about: I think that API users can call SmtEngine::getModel() and store a Model* object on their end.  If the user then issues a second check-sat, then using the same model object without calling SmtEngine::getModel() again, this will not trigger the model being generated.  This wasn't an issue before this PR since we always eagerly built the model after each check-sat that gave "sat".

mpreiner:
Is there a way to invalidate the model so that the user is forced to call SmtEngine::getModel() again?

ajreynol:
As discussed with @mpreiner offline, there isn't an obvious clean solution.  One could make the Model object know whether or not its dirty/clean, but then it would likely need a pointer to the object that is capable of building it (TheoryEngineModelBuilder), which is a bit ugly.

ajreynol:
I've implemented the "fallback to eager model building mode" solution as suggested by Clark. I opened an issue #2648 to revisit this.

I now believe that there are no shortcomings with this PR, should be ready to review.

ajreynol:
From Daniel:

```
Hi Andy,

I have the results of the test.

Using this branch version, Kind 2 can prove 1 valid problem more than before while improving times.

[New results] Safe: 482 in 1069s; Unsafe: 379 in 482s.
[Old results] Safe: 481 in 1151s; Unsafe: 379 in 481s.

If we compare times over the same set of proved problems (i.e. without the new one), we have:
[New results] Safe: 481 in 1007s; Unsafe: 379 in 482s.

Best,
Daniel
```

So it seems the results are good on the Kind2 end. I'm not sure how to run the incremental unsat core tests, I assume this is setup on the Stanford cluster?

ajreynol:
This is work towards #2305.

PR #2634, opened 2018-10-15
literal-block-distance based learnt clause removal and learning rate based branching heuristics

anwu1219:
1. Add options: --decision==lrb (learning rate based branching heuristics) and justification-stoponly-lrb
2. Add option: --lbd (literal block distance aggressive conflict clause removal), which is compatible with every branching heuristics (by default, on for lrb and lrb-stoponly, off for the others).


src/prop/minisat/core/Solver.h
```diff
@@ -252,8 +252,14 @@ class Solver {
     // Mode of operation:
     //
     int       verbosity;
+    // for LRB
+    double step_size;
+    double step_size_dec;
+    double min_step_size;
```
4tXJ7f: It looks like these three variables are not actually initialized. Am I missing something?

src/options/options_handler.cpp
```diff
@@ -1410,16 +1410,27 @@ justification-stoponly\n\
...
+    options::decisionStopOnly.set(2);
+    return decision::DECISION_STRATEGY_JUSTIFICATION;
+  }
+  else if (optarg == "lrb")
+  {
+    return decision::DECISION_STRATEGY_LRB;
+  }
+  else if (optarg == "help")
+  {
     puts(s_decisionModeHelp.c_str());
```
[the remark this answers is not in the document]
anwu1219: How should I update the help string?

4tXJ7f:
Also, please add the licensing information of the code from external sources in the source file itself and in the COPYING file.

src/prop/minisat/core/Solver.h
4tXJ7f: Note: if my comment is correct and the variables are not initialized, we probably need to reevaluate the strategies affected by it.

anwu1219: Yeah, I think you are correct that they are not initialized. This is probably why LRB did not win much. I think I need to re-evaluate two strategies: lrb and lrb+stoponly. The other ones should be fine.

mpreiner:

PR #2636, opened 2018-10-15
Improve reduction for str.to.int

ajreynol:
Implements analogous improvements as #2629 for str.to.int.


src/theory/strings/theory_strings_preprocess.cpp
```diff
@@ -275,104 +276,90 @@ Node StringsPreprocess::simplify( Node t, std::vector< Node > &new_nodes ) {
...
-      str.eqNode(NodeManager::currentNM()->mkConst(::CVC4::String(""))),
-      pret.eqNode(negone));
+    Node sEmpty = s.eqNode(d_empty_str);
+    Node k = nm->mkSkolem("k", nm->integerType());
+    Node kc1 = nm->mkNode(GEQ, k, d_zero);
+    Node kc2 = nm->mkNode(LT, k, lens);
+    Node codeSk = nm->mkNode(
+        MINUS,
+        nm->mkNode(STRING_CODE, nm->mkNode(STRING_SUBSTR, s, k, d_one)),
+        nm->mkConst(Rational(48)));
```
4tXJ7f: Use `c0` instead.

PR #2641, opened 2018-10-16
Improve strings reductions including skolem caching for contains

ajreynol:
This PR:
- Merges the skolems for positive contains and replace/indexof.
- Rewrites terms before looking up skolems in the cache.
- Marks (when appropriate) reductions as SAT-context-independent, which leads to fewer extraneous inferences.

It is a net +250 on the CMU CAV 2017 set. As an estimate, +220 of these or so can be attributed to the first optimization, +30 to the third.

### DIFF
--- a/src/theory/strings/theory_strings.cpp
+++ b/src/theory/strings/theory_strings.cpp
@@ -453,9 +453,9 @@ int TheoryStrings::getReduction( int effort, Node n, Node& nr ) {
           Node s = n[1];
           //positive contains reduces to a equality
           Node sk1 = d_sk_cache.mkSkolemCached(
-              x, s, SkolemCache::SK_ID_CTN_PRE, "sc1");
+              x, s, SkolemCache::SK_FIRST_CTN_PRE, "sc1");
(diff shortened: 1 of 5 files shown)


src/theory/strings/theory_strings.cpp
```diff
@@ -482,14 +492,15 @@ int TheoryStrings::getReduction( int effort, Node n, Node& nr ) {
           sendInference( d_empty_vec, nnlem, "Reduction", true );
           //we've reduced this n
           Trace("strings-extf-debug") << "  resolve extf : " << n << " based on reduction." << std::endl;
-          return 1;
+          isCd = false;
+          return true;
         }
       }else{
-        return 1;
+        return true;
```
4tXJ7f: Shouldn't we also set `isCd` explicitly here?

ajreynol: Yes, although this case should never fire anymore. I'll reorganize/remove this.

ajreynol:
I addressed the review. Unfortunately I had to change indentation in doReductions.

PR #2652, opened 2018-10-18
Non-contributing find replace rewrite

ajreynol:
Solves 2/8 remaining candidate rewrites for term size 2 strings.


ajreynol:
Ran --sygus-rr-verify for ~15 minutes with no issues.

src/theory/strings/theory_strings_rewriter.cpp
```diff
@@ -2845,6 +2845,54 @@ Node TheoryStringsRewriter::rewriteReplace( Node node ) {
...
+      else
+      {
+        break;
+      }
+    }
+    if (!lastLhs.isNull())
+    {
+      std::vector<Node> remc;
+      remc.insert(
+          remc.end(), children0.begin() + lastCheckIndex, children0.end());
```
4tXJ7f: Nit: the previous two statements can be merged to `std::vector<Node> remc(children0.begin() + lastCheckIndex, children0.end());`.

PR #2653, opened 2018-10-18
Non-implied mode for model cores

ajreynol:
This introduces a new "non-implied" mode for model cores.

@barrettcw, you may be interested in this PR.


test/regress/regress0/model-core.smt2
```diff
@@ -1,4 +1,4 @@
-; COMMAND-LINE: --produce-model-cores
+; COMMAND-LINE: --produce-models --model-cores=simple
```
yoni206: Perhaps we should test this both for simple and for non-implied modes?

```suggestion
; COMMAND-LINE: --produce-models --model-cores=simple
;COMMAND-LINE: --produce-models --model-core=non-implied
```

yoni206: What do you think about adding Alessandro's example as another test case for both modes (maybe in regress1)?

yoni206: Is there a test case where the model core is predictable and robust, in which we can use ;EXPECT in order to verify that the model core is what we expected it to be?

src/theory/subs_minimize.cpp
```diff
@@ -124,8 +262,6 @@ bool SubstitutionMinimize::find(Node n,
   Trace("subs-min") << "--- Compute relevant variables..." << std::endl;
   std::unordered_set<Node, NodeHashFunction> rlvFv;
   // only variables that occur in assertions are relevant
```
yoni206: Does this comment still belong here?

ajreynol: Yes, relevant here means "possibly required or implied".

test/regress/regress0/model-core.smt2
4tXJ7f: Is it possible to implement something similar to `--check-models` for model cores (maybe out-of-scope for this PR)?

PR #2654, opened 2018-10-18
Add helper to detect length one string terms

4tXJ7f:
This commit introduces a helper function to detect string terms that
have exactly/at most length one. This is useful for a lot of rewrites
because strings of at most length one are guaranteed to not overlap
multiple components in a concatenation, which allows for more aggressive
rewriting.

This commit has been tested with --sygus-rr-synth-check for >1h on the
string term grammar.


ajreynol:
Looks good, thanks for doing this.

4tXJ7f:
For the record: I was referring to`--sygus-rr-synth-verify`, not `--sygus-rr-synth-check` :)

PR #2656, opened 2018-10-18
Fail for SWIG 3.0.8

makaimann:
Ran into this bug when compiling with python3 bindings: https://github.com/swig/swig/issues/588

Instantiating any object crashes python. Since swig3.0.8 is currently the apt-get install for Ubuntu 16.04, I thought it'd be good to have a check for that. If python3 is preferred and the swig version is 3.0.8, it errors out and asks users to downgrade or upgrade SWIG.


src/bindings/CMakeLists.txt
```diff
@@ -3,6 +3,13 @@ if(NOT ENABLE_SHARED)
 endif()
 
 find_package(SWIG 3.0.0 REQUIRED)
+
+if(USE_PYTHON3 AND (${SWIG_VERSION} MATCHES 3.0.8))
```
mpreiner: Use `(SWIG_VERSION VERSION_EQUAL 3.0.8)`.

makaimann: Done.

PR #2657, opened 2018-10-19
Only build CryptoMiniSat library, no binary

4tXJ7f:
This commit changes the contrib/get-cryptominisat script to only build
the CryptoMiniSat library instead of both the library and the binary.
The advantage of this is that we can compile a static version of the
CryptoMiniSat library without having a static version of glibc or
libstdc++ (this is fine as long as we do a shared library build of
CVC4). This is an issue on Fedora (tested on version 25) where the
contrib/get-cryptominisat script was failing when building the
CryptoMiniSat binary due to the static version of these libraries not
being available. Since we just want to build the library, the commit
also changes the script to not install CryptoMiniSat anymore and updates
the CMake find script to accomodate the new folder structure. Side note:
the folder structure generated after this commit is a bit more uniform
with, e.g. the CaDiCaL script: The source files are directly in the
cryptominisat5 folder, not in a subfolder.


mpreiner:
I would actually prefer to use the `make install` approach since this is more robust rather than finding the built library somewhere in the build directory. If the location of the library in the build directory changes we won't find it. CaDiCaL doesn't have a `make install` rule and therefore needs to be handled differently.

PR #2660, opened 2018-10-19
Sygus streaming non-implied predicates

### LINKED ISSUE #1212: SyGuS stream solution filtering
Daniel Larraz requested a sygus streaming option that does not produce implied invariants, so we would not produce this output to an invariant synthesis problem with --sygus-stream:

(declare-fun I ((x Int)) Bool (> x 1))
(declare-fun I ((x Int)) Bool (> x 0)) 

since the second solution is implied by the first.

ajreynol:
Work towards #1212.

The new option --sygus-sol-filter-implied implies that sygus solutions in streaming mode are not collectively entailed by previous solutions. One target application for this feature is auxilary invariant generation in Kind2.


src/theory/quantifiers/solution_filter.cpp
```diff
@@ -0,0 +1,77 @@
...
+{
+  ExprMiner::initialize(vars, ss);
+}
+
+bool SolutionFilter::addTerm(Node n, std::ostream& out)
+{
+  if( !n.getType().isBoolean() )
+  {
+    // currently, should not register non-Boolean terms here
+    Assert(false);
```
HanielB: why assert false? Having no effect isn't enough?

src/theory/quantifiers/solution_filter.cpp
```diff
@@ -0,0 +1,77 @@
...
+  }
+  if (n.isConst())
+  {
+    if (n.getConst<bool>())
+    {
+      return false;
+    }
+    else
+    {
+      // could abort after printing here
```
HanielB: agreed. Just do it?

src/theory/quantifiers/solution_filter.cpp
```diff
@@ -0,0 +1,77 @@
...
+    }
+  }
+  NodeManager* nm = NodeManager::currentNM();
+  Node imp = d_conj.isNull() ? n.negate() : nm->mkNode(AND, d_conj, n.negate());
+  Trace("sygus-sol-implied") << "  implies: check " << imp << "..." << std::endl;
+  // make the satisfiability query
```
HanielB: could it be worthwhile to rewrite the implication before testing?

PR #2663, opened 2018-10-19
Travis: run examples and avoid building them twice

4tXJ7f:
`make check` builds the examples but does not run them. This commit
changes our Travis script to run the examples after building them and
removes `makeExamples()` to avoid building them twice.


.travis.yml
```diff
@@ -76,12 +76,8 @@ script:
    #       error "DISTCHECK (WITH NEWTHEORY TESTS) FAILED";
    #   }
    makeCheck() {
-     cd build
      make -j2 check ARGS='-LE regress[1-4]' CVC4_REGRESSION_ARGS='--no-early-exit' || error "BUILD/UNIT/SYSTEM/REGRESSION TEST FAILED"
-   }
-   makeExamples() {
-     cd build
-     make -j2 examples || error "COULD NOT BUILD EXAMPLES${normal}";
+     ctest -j2 -R examples/ || error "RUNNING EXAMPLES FAILED"
```
mpreiner: `-L example` should also work here.

aniemetz: Yes, and it's also the cleaner way to do this here.

4tXJ7f: Yup, that makes sense :)

PR #2665, opened 2018-10-20
Add substr, contains and equality rewrites

4tXJ7f:
This commit adds three new rewrites:

- `(str.contains (str.replace x y z) w) ---> (str.contains (str.replace
x y "") w)` if `(str.contains z w) ---> false` and `(str.len w) = 1`
- `(= (str.replace x y z) z) ---> (or (= x y) (= x z))` if `(str.len y)
= (str.len z)`
- `(str.substr s x y) --> ""` if `0 < y |= x >= str.len(s)`

### DIFF
--- a/src/theory/strings/theory_strings_rewriter.cpp
+++ b/src/theory/strings/theory_strings_rewriter.cpp
@@ -575,6 +575,21 @@ Node TheoryStringsRewriter::rewriteStrEqualityExt(Node node)
           return returnRewrite(node, ret, "str-eq-repl-not-ctn");
         }
       }
+
+      // (= (str.replace x y z) z) --> (or (= x y) (= x z))
+      // if (str.len y) = (str.len z)
+      if (repl[2] == x)
+      {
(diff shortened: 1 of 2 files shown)


src/theory/strings/theory_strings_rewriter.cpp
```diff
@@ -1733,6 +1748,15 @@ Node TheoryStringsRewriter::rewriteSubstr(Node node)
         return returnRewrite(node, ret, "ss-start-entails-zero-len");
       }
 
+      // (str.substr s x y) --> "" if 0 < y |= x >= str.len(s)
+      Node non_zero_len =
+          Rewriter::rewrite(nm->mkNode(kind::LT, zero, node[2]));
+      if (checkEntailArithWithAssumption(non_zero_len, node[1], tot_len, false))
```
ajreynol: I think this function expects the assumption to be rewritten. Maybe we should just build this into `checkEntailArithWithAssumption`?

PR #2668, opened 2018-10-22
Do not use lazy trie for sygus-rr-verify

ajreynol:


src/theory/datatypes/datatypes_sygus.cpp
```diff
@@ -1096,45 +1096,46 @@ Node SygusSymBreakNew::registerSearchValue(Node a,
...
+          bve = its->second.evaluate(bv, i);
+          bvre = its->second.evaluate(bvr, i);
+          if (bve != bvre)
+          {
+            ptDisequal = true;
+            pt_index = i;
+            break;
+          }
+        }
         // bv and bvr should be equivalent under examples
```
4tXJ7f:
```suggestion
         // bve and bvre should be equivalent under examples
```

PR #2669, opened 2018-10-22
Modification to HandleAndOrEasy()

anwu1219:
1. Create an option "--handle-and-or-easy-check2", when turned on, we check whether the children of a node can already justify the parent (2 children at a time) before calling findSplitterRec() on those children. The default of the option is false.
2. Make JustificationHeuristic a template class. JustificationHeuistic<true> calls the handleAndOrEasy that does the checking; JustificationHeuistic<false> calls the one that does not do the checking.


src/decision/justification_heuristic.cpp
```diff
@@ -552,36 +580,81 @@ JustificationHeuristic::findSplitterRec(TNode node, SatValue desiredVal)
...
-    TNode curNode = getChildByWeight(node, i, desiredVal);
-    if ( tryGetSatValue(curNode) != desiredValInverted ) {
-      SearchResult ret = findSplitterRec(curNode, desiredVal);
-      if(ret != DONT_KNOW) {
-        return ret;
+  if (check2)
+  {
+    for (int i = 0; i < numChildren; i += 2)
+    {
+      std::vector<Node> nodes;
```
4tXJ7f: Instead of creating a vector for every iteration of the outer loop, declare it outside that loop and clear it in here for better performance. I also believe that it should be ok to use `TNode`s here.

src/decision/justification_heuristic.h
```diff
@@ -38,7 +38,15 @@
...
 
-class JustificationHeuristic : public ITEDecisionStrategy {
+template <bool ref_count>
+class JustificationHeuristic;
+
+template <bool check2>
+class JustificationHeuristic : public ITEDecisionStrategy
+{
+  friend class JustificationHeuristic<true>;
+  friend class JustificationHeuristic<false>;
```
4tXJ7f: Are those `friend` declarations necessary?

src/decision/justification_heuristic.h
```diff
@@ -38,7 +38,15 @@
 namespace CVC4 {
 namespace decision {
 
-class JustificationHeuristic : public ITEDecisionStrategy {
+template <bool ref_count>
+class JustificationHeuristic;
```
4tXJ7f: I don't think that this forward declaration is necessary.

src/decision/justification_heuristic.h
```diff
@@ -38,7 +38,15 @@
...
+
+template <bool check2>
```
4tXJ7f: `check2` is not very descriptive and sounds a bit like there should be a `check1`, maybe something like `checkInPairs`? Maybe someone has an even better idea.

src/decision/justification_heuristic.h
```diff
@@ -176,7 +184,10 @@ class JustificationHeuristic : public ITEDecisionStrategy {
                         TNode node2, SatValue desiredVal2);
   SearchResult handleITE(TNode node, SatValue desiredVal);
   SearchResult handleEmbeddedITEs(TNode node);
-};/* class JustificationHeuristic */
+}; /* class JustificationHeuristic */
+
+typedef JustificationHeuristic<true> JustificationHeuristicNoCheck;
+typedef JustificationHeuristic<false> JustificationJeuristicCheck2;
```
4tXJ7f: Are those typedefs necessary?

4tXJ7f:
Also, it looks like it does not currently compile on Clang (see Travis build).

ajreynol:
Just a style preference here, I don't like the templates introduced in this PR. What is the harm in having a data member d_doCheck2 (or something with a better name) that gives the preference for whether the check is performed?  Or even just referring to the option directly in the code? My concern is such templates could get out of hand, say, with more optional flags to the class.

4tXJ7f:
The motivation for that was performance. I think that you are bringing up a good point. We should just try to use a data member and check how much worse the performance is (I suspect that the difference is negligible).

src/decision/justification_heuristic.h
anwu1219: You're right. I don't think they are.

anwu1219:
@4tXJ7f @ajreynol After a more thorough test, it seems that the runtime increase is negligible when using a data member that stores the flag "handle-and-or-easy-check-in-pairs". Therefore, I think there is good reason to go with this implementation.

anwu1219:
Here is the runtime comparison of using the template implementation and the data member implementation. Both solve the exact same amount of instances for each benchmark tested.

  Benchmark                         | Template | No Template
 QF_AUFLIA_subset (1303) | 2326.0     | 2339.4
 QF_AUFNIA_subset (17)     | 118.0         | 117.9
         QF_AX_subset (551)   | 304.3        | 302.7
     QF_BVFP_subset (5000)| 2898.1      | 2897.3
         QF_BV_subset (5000)| 235511.8   | 235622.9

mpreiner:

PR #2671, opened 2018-10-24
CMake: Set RPATH on installed binary

4tXJ7f:
Currently, when installing CVC4 with a custom installation directory on
macOS, the resulting binary cannot be executed because the linker cannot
find the required libraries (e.g. our parser). This commit changes our
build system to use the `CMAKE_INSTALL_RPATH` variable to add the
installation directory to the RPATH list in the exectuable.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -31,6 +31,17 @@ set(CMAKE_CXX_STANDARD 11)
 # plugins.
 set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 
+# Embed the installation prefix as an RPATH in the executable such that the
+# linker can find our libraries (such as libcvc4parser) when executing the cvc4
+# binary. This is for example useful when installing CVC4 with a custom prefix
+# on macOS (e.g. when using homebrew in a non-standard directory). If we do not
+# set this option, then the linker will not be able to find the required
+# libraries when trying to run CVC4.
+#
+# More information on RPATH in CMake:
+# https://gitlab.kitware.com/cmake/community/wikis/doc/cmake/RPATH-handling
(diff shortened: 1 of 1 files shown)


4tXJ7f:
Note: This (or some other solution) is required for our macOS nightlies.

CMakeLists.txt
```diff
@@ -31,6 +31,17 @@ set(CMAKE_CXX_STANDARD 11)
...
+# linker can find our libraries (such as libcvc4parser) when executing the cvc4
+# binary. This is for example useful when installing CVC4 with a custom prefix
+# on macOS (e.g. when using homebrew in a non-standard directory). If we do not
+# set this option, then the linker will not be able to find the required
+# libraries when trying to run CVC4.
+#
+# More information on RPATH in CMake:
+# https://gitlab.kitware.com/cmake/community/wikis/doc/cmake/RPATH-handling
+set(CMAKE_INSTALL_RPATH "${CMAKE_INSTALL_PREFIX}/lib")
+
```
aniemetz: Do we want to set this in general or only for macOS?

4tXJ7f: Good question. When I submitted the PR, I was under the impression that it at least shouldn't hurt on other plattforms but your comment prompted me to actually test it on Linux. It looks like the problem that I described above also applies to Linux when using current `master`:

```
$ ./configure.sh --prefix=./install
...
$ cd build && make install
...
$ cd .. && install/bin/cvc4
install/bin/cvc4: error while loading shared libraries: libcvc4parser.so.5: cannot open shared object file: No such file or directory
```

This commit fixes the issue under Linux as well.

PR #2673, opened 2018-10-24
Sketch of refactoring the Commands to use new API

4tXJ7f:
This commit is a sketch of how we can transform `Command`s to use the
new API and transform the parser bit-by-bit. The overall idea is that
`Command::invoke()` now takes an `Solver*` as an argument instead of an
`SmtEngine*`. The default implementation `Command::invoke()` just calls
the old implementation. As an example, the PR transforms the
`GetValueCommand`.

### DIFF
--- a/src/printer/ast/ast_printer.cpp
+++ b/src/printer/ast/ast_printer.cpp
@@ -20,6 +20,7 @@
 #include <typeinfo>
 #include <vector>
 
+#include "api/cvc4cpp.h"  // for ExprSetDepth etc..
 #include "expr/expr.h" // for ExprSetDepth etc..
 #include "expr/node_manager_attributes.h" // for VarNameAttr
 #include "options/language.h" // for LANG_AST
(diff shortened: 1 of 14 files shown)


src/printer/ast/ast_printer.cpp
```diff
@@ -20,6 +20,7 @@
 #include <typeinfo>
 #include <vector>
 
+#include "api/cvc4cpp.h"  // for ExprSetDepth etc..
```
mpreiner: Is the header really needed for `ExprSetDepth` etc.?

src/printer/cvc/cvc_printer.cpp
```diff
@@ -24,11 +24,12 @@
 #include <typeinfo>
 #include <vector>
 
-#include "expr/expr.h" // for ExprSetDepth etc..
-#include "expr/node_manager_attributes.h" // for VarNameAttr
-#include "options/language.h" // for LANG_AST
-#include "printer/dagification_visitor.h"
+#include "api/cvc4cpp.h"                   // for ExprSetDepth etc..
```
mpreiner: Ditto

PR #2674, opened 2018-10-24
API/Smt2 parser: refactor termAtomic

4tXJ7f:
This commit refactors `termAtomic` in our SMT 2.x/SyGuS parser to use
the new API. Instead of returning an `Expr`, `termAtomic` now returns an
`api::Term` and all the constants parsed by `termAtomic` are created
with the new API.  This required some changes to the API: The commit
implements additional methods for accessing datatypes, creating
floating-point constants, creating bit-vector constants, and changes the
API to make sure that it returns a constant of sort real when
`Solver::mkReal()`/`Solver::mkConst(CONST_RATIONAL, ...)` is used.


ajreynol:
I think we need to think more about ensureRealSort from this PR.  It's questionable whether building in the divide-by-one trick into the API is a good idea. Notice that #1005 is indirectly related to this.

4tXJ7f:
Yes, I think that this is something that we should discuss in a larger group. My reasoning was that if you use `mkReal` as opposed to `mkInteger` then you really expect to get something of sort real. Maybe we could make it optional?

4tXJ7f:
We decided to remove `mkInteger` and make it very clear that `mkReal` can return either and integer or a real. `ensureRealSort` should become a user-facing utility function. `mkTuple` should be added to the API (maybe in another PR).

4tXJ7f:
Note: this PR currently includes #2680, which is why I am temporarily marking it as "do not merge".

src/api/cvc4cpp.h
```diff
@@ -1818,129 +1857,87 @@ class CVC4_PUBLIC Solver
...
-   * Create an Real constant.
-   * @param s the string represetntation of the constant
-   * @param base the base of the string representation
-   * @return the Real constant
+   * Create a real constant.
+   * @param s the string representation of the constant
+   * @return a constant of sort Real or Integer (if the string is an integer)
    */
-  Term mkReal(const char* s, uint32_t base = 10) const;
+  Term mkReal(const char* s) const;
```
aniemetz: Why did you remove the `base` argument?

src/api/cvc4cpp.h
```diff
@@ -1818,129 +1857,87 @@ class CVC4_PUBLIC Solver
...
-   * Create an Real constant.
-   * @param s the string represetntation of the constant
-   * @param base the base of the string representation
-   * @return the Real constant
+   * Create a real constant.
+   * @param s the string representation of the constant
+   * @return a constant of sort Real or Integer (if the string is an integer)
    */
-  Term mkReal(const std::string& s, uint32_t base = 10) const;
+  Term mkReal(const std::string& s) const;
```

src/api/cvc4cpp.h
```diff
@@ -1818,129 +1857,87 @@ class CVC4_PUBLIC Solver
...
+   * @return a constant of sort Integer
    */
   Term mkReal(int32_t val) const;
 
   /**
-   * Create an Real constant.
+   * Create a real constant.
    * @param val the value of the constant
-   * @return the Real constant
+   * @return a constant of sort Integer
```
aniemetz: Is the sort indeed Integer?

src/api/cvc4cpp.h
```diff
@@ -1818,129 +1857,87 @@ class CVC4_PUBLIC Solver
...
    */
   Term mkReal(int64_t val) const;
 
   /**
-   * Create an Real constant.
+   * Create a real constant.
    * @param val the value of the constant
-   * @return the Real constant
+   * @return a constant of sort Integer
```

src/api/cvc4cpp.h
```diff
@@ -1818,129 +1857,87 @@ class CVC4_PUBLIC Solver
...
    */
   Term mkReal(uint32_t val) const;
 
   /**
-   * Create an Real constant.
+   * Create a real constant.
    * @param val the value of the constant
-   * @return the Real constant
+   * @return a constant of sort Integer
```

src/api/cvc4cpp.h
4tXJ7f: The reason is that we don't support constructing real numbers from something like `101.010` (binary base). The `Rational(const std::string& s, unsigned base)` constructor expects a string of the form `1101/101` (binary base). For the parser, I needed to support parsing constants like `9.123`, so I changed `mkReal()` to use `Rational::fromDecimal()` and removed the `base` parameter. We could split this up into two constructors `mkReal()` and `mkRational()` though but I wasn't sure if we really need that.

src/api/cvc4cpp.h
```diff
@@ -2044,6 +2045,64 @@ class CVC4_PUBLIC Solver
...
   Term mkBitVector(std::string& s, uint32_t base = 2) const;
 
+  /**
+   * Create a bit-vector constant of a given bit-width from a given string.
+   * @param s the string represetntation of the constant
+   * @param base the base of the string representation
+   * @param sz the bit-width of the constant
+   * @return the bit-vector constant
+   */
+  Term mkBitVector(const char* s, uint32_t base, uint32_t sz) const;
```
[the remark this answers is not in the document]
4tXJ7f: I guess the idea was that you can just use the length of the string to determine the size but I agree, I'd just keep the one with an explicit `size` parameter. I'll address the consistency issues, not sure why I didn't think of that.

src/parser/smt2/Smt2.g
```diff
@@ -2207,148 +2216,165 @@ termNonVariable[CVC4::Expr& expr, CVC4::Expr& expr2]
...
-    { expr = MK_CONST( AntlrInput::tokenToInteger($INTEGER_LITERAL) ); }
-
+    {
+      std::string intStr = AntlrInput::tokenText($INTEGER_LITERAL);
+      atomTerm = SOLVER->mkReal(intStr);
+    }
   | DECIMAL_LITERAL
-    { // FIXME: This doesn't work because an SMT rational is not a
+    {
+      // FIXME: This doesn't work because an SMT rational is not a
```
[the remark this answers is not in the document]
4tXJ7f: Actually this `FIXME` does not apply anymore (neither in the code before nor after this PR), I'll remove it. This actually goes back to my explanation above about parsing decimal literals like `9.123`.

src/api/cvc4cpp.h
4tXJ7f: For the record: we have decided to support only the decimal base but both styles `123/456` and `123.456`.

4tXJ7f:
All the comments should (finally) be addressed. Note that I found something strange: type checking the universe set with a boolean type throws an error, so I removed type checking from that method. We should fix that issue but I'd do it in another PR (because it will probably involve some discussions about _how_ to fix it). If you agree to postpone fixing this to a later PR, I'll open an issue for it.

src/api/cvc4cpp.cpp
```diff
@@ -2124,6 +2141,61 @@ Term Solver::mkBitVector(const std::string& s, uint32_t base) const
   return mkBVFromStrHelper(s, base);
 }
 
+Term Solver::mkBitVector(uint32_t size, const char* s, uint32_t base) const
+{
+  std::string str(s);
+  return mkBitVector(size, str, base);
```
aniemetz: Do not use nested API calls, will lead to problems with API tracing.
Introduce a new variant of `mkBVFromStrHelper()` and use this variant here and below.

src/api/cvc4cpp.cpp
```diff
@@ -2124,6 +2141,61 @@ Term Solver::mkBitVector(const std::string& s, uint32_t base) const
...
+}
+
+Term Solver::mkBitVector(uint32_t size, std::string& s, uint32_t base) const
+{
+  Integer val(s, base);
+  CVC4_API_CHECK(val.modByPow2(size) == val)
+      << "Overflow in bitvector construction (specified bitvector size " << size
+      << " too small to hold value " << s << ")";
+  return d_exprMgr->mkConst(BitVector(size, val));
```
aniemetz: Use`mkConstHelper()` (wrapper that ensures type checking, throws CVC4APIException).

src/api/cvc4cpp.cpp
```diff
@@ -2124,6 +2141,61 @@ Term Solver::mkBitVector(const std::string& s, uint32_t base) const
...
+}
+
+Term Solver::mkPosInf(uint32_t exp, uint32_t sig) const
+{
+  CVC4_API_CHECK(Configuration::isBuiltWithSymFPU())
+      << "Expected CVC4 to be compiled with SymFPU support";
+  return d_exprMgr->mkConst(
+      FloatingPoint::makeInf(FloatingPointSize(exp, sig), false));
```
aniemetz: Ditto

src/api/cvc4cpp.cpp
```diff
@@ -2124,6 +2141,61 @@ Term Solver::mkBitVector(const std::string& s, uint32_t base) const
...
+}
+
+Term Solver::mkNegInf(uint32_t exp, uint32_t sig) const
+{
+  CVC4_API_CHECK(Configuration::isBuiltWithSymFPU())
+      << "Expected CVC4 to be compiled with SymFPU support";
+  return d_exprMgr->mkConst(
+      FloatingPoint::makeInf(FloatingPointSize(exp, sig), true));
```
aniemetz: Ditto

src/api/cvc4cpp.cpp
```diff
@@ -2124,6 +2141,61 @@ Term Solver::mkBitVector(const std::string& s, uint32_t base) const
...
+}
+
+Term Solver::mkNaN(uint32_t exp, uint32_t sig) const
+{
+  CVC4_API_CHECK(Configuration::isBuiltWithSymFPU())
+      << "Expected CVC4 to be compiled with SymFPU support";
+  return d_exprMgr->mkConst(
+      FloatingPoint::makeNaN(FloatingPointSize(exp, sig)));
```
aniemetz: Ditto

src/api/cvc4cpp.cpp
```diff
@@ -2124,6 +2141,61 @@ Term Solver::mkBitVector(const std::string& s, uint32_t base) const
...
+}
+
+Term Solver::mkPosZero(uint32_t exp, uint32_t sig) const
+{
+  CVC4_API_CHECK(Configuration::isBuiltWithSymFPU())
+      << "Expected CVC4 to be compiled with SymFPU support";
+  return d_exprMgr->mkConst(
```
aniemetz: Ditto

src/api/cvc4cpp.cpp
```diff
@@ -2124,6 +2141,61 @@ Term Solver::mkBitVector(const std::string& s, uint32_t base) const
...
+      FloatingPoint::makeZero(FloatingPointSize(exp, sig), false));
+}
+
+Term Solver::mkNegZero(uint32_t exp, uint32_t sig) const
+{
+  CVC4_API_CHECK(Configuration::isBuiltWithSymFPU())
+      << "Expected CVC4 to be compiled with SymFPU support";
+  return d_exprMgr->mkConst(
```
aniemetz: Ditto

src/api/cvc4cpp.cpp
```diff
@@ -3356,6 +3445,31 @@ void Solver::setOption(const std::string& option,
   d_smtEngine->setOption(option, value);
 }
 
+Term Solver::ensureTermSort(const Term& t, const Sort& s) const
```
aniemetz: Did you double check if this can be used in other places, too? I'd imagine there are several places.

test/unit/api/solver_black.h
```diff
@@ -26,6 +27,7 @@ class SolverBlack : public CxxTest::TestSuite
   void setUp() override;
   void tearDown() override;
 
+  void testGetNullSort();
```
aniemetz: Sort alphabetically.

test/unit/api/solver_black.h
```diff
@@ -47,7 +49,9 @@ class SolverBlack : public CxxTest::TestSuite
   void testMkTupleSort();
   void testMkUninterpretedSort();
 
+  void testMkTuple();
```
aniemetz: Ditto

test/unit/api/solver_black.h
```diff
@@ -47,7 +49,9 @@ class SolverBlack : public CxxTest::TestSuite
...
   void testMkBitVector();
+  void testMkFloatingPoint();
```
aniemetz: Ditto

test/unit/api/solver_black.h
```diff
@@ -234,6 +243,22 @@ void SolverBlack::testMkTupleSort()
                    CVC4ApiException&);
 }
 
+void SolverBlack::testMkTuple()
```
aniemetz: Ditto

test/unit/api/solver_black.h
```diff
@@ -234,6 +243,22 @@ void SolverBlack::testMkTupleSort()
...
+{
+  TS_ASSERT_THROWS(d_solver.mkTuple({}, {d_solver.mkBitVector("101", 2)}),
```
aniemetz: Can you maybe group the THROWS and THROWS_NOTHING?

src/api/cvc4cpp.cpp
```diff
@@ -2077,7 +2095,6 @@ Term Solver::mkUniverseSet(Sort sort) const
     CVC4_API_ARG_CHECK_EXPECTED(!sort.isNull(), sort) << "non-null sort";
     Term res =
         d_exprMgr->mkNullaryOperator(*sort.d_type, CVC4::kind::UNIVERSE_SET);
-    (void)res.d_expr->getType(true); /* kick off type checking */
```
aniemetz: I'd rather have this not removed if this is undesired behavior. Disable and assign an issue. We can fix this in a separate PR.

src/api/cvc4cpp.cpp
4tXJ7f: Good point. Done.

src/api/cvc4cpp.cpp
4tXJ7f: Done (the method now calls the new `mkBVFromStrHelper()` which uses `mkConstHelper()`.

src/api/cvc4cpp.cpp
4tXJ7f: Done.

src/api/cvc4cpp.cpp
4tXJ7f: Done.

src/api/cvc4cpp.cpp
4tXJ7f: Done.

src/api/cvc4cpp.cpp
4tXJ7f: Done.

src/api/cvc4cpp.cpp
4tXJ7f: Done.

src/api/cvc4cpp.cpp
4tXJ7f: As far as I can tell, the only two places where this is useful for now is tuple construction and parsing decimal numbers. It may be useful for datatypes, too, but I couldn't see a place where we do this conversion. We should definitely keep this in mind for the future, though.

PR #2684, opened 2018-11-02
unreachable code removal

yoni206:
The PR removes [this](https://github.com/CVC4/CVC4/blob/050747cbceef232b11f1226081bc3dbc74c8ff77/src/preprocessing/passes/bv_to_bool.cpp#L109) `if` statement, which always resolves to `false`:
The `if` statement is executed after the negation of its condition is [asserted](https://github.com/CVC4/CVC4/blob/050747cbceef232b11f1226081bc3dbc74c8ff77/src/preprocessing/passes/bv_to_bool.cpp#L144).

Note, however, that this statement might be reachable in production, as we use `Assert` and not `AlwaysAssert`.

### DIFF
--- a/src/preprocessing/passes/bv_to_bool.cpp
+++ b/src/preprocessing/passes/bv_to_bool.cpp
@@ -106,9 +106,6 @@ bool BVToBool::isConvertibleBvAtom(TNode node)
 
 bool BVToBool::isConvertibleBvTerm(TNode node)
 {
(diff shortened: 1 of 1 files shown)


4tXJ7f:
As discussed offline, I think that it is fine to remove the if statement because an assertion should never be triggered (for performance reasons we just don't check in production). I would however add the same assertion at the top of the function that you removed the if statement from. This would show what restrictions you have on the inputs.

Issue #1667, opened 2018-03-14
Implement constant arrays fully


barrettcw:
We currently support the creation of constant arrays (via kind ARRAY_STORE_ALL), but under some conditions when there is a chain of equalities connecting two array constants, the solver doesn't know what to do and aborts.

We should 1) fail more gracefully in this case (return unknown) and 2) implement constant arrays properly.

dddejan:
The simple smt2 examples I am trying with constant arrays seem to be working.

A characterization of when the solver will fail, or a simple example would be very helpful.

barrettcw:
Here's how I would characterize the current support.  You can create constant arrays and use them however you want, but there is an incompleteness problem.  The problem comes when you use equalities between array expressions.  If there is some chain of equalities from one constant array to another constant array, then all bets are off and you may run into the write chains error message.

barrettcw:
Note that it should not be possible to get wrong answers, only assertion failures.

hajduakos:
For our simple examples with @dddejan it seems to work even if we create two constant arrays and check for equality between elements and/or the whole array

barrettcw:
It might only be a problem if you store some things into the array and then check for equality.  If it is working, consider yourselves lucky.  :)

And do let us know when you run into the inevitable case that doesn't work...

PR #2685, opened 2018-11-03
Adding default SyGuS grammar construction for arrays

HanielB:


src/theory/quantifiers/sygus/sygus_grammar_cons.cpp
```diff
@@ -371,6 +371,38 @@ void CegGrammarConstructor::mkSygusConstantsForType(TypeNode type,
   {
     ops.push_back(nm->mkConst(String("")));
   }
+  else if (type.isArray())
+  {
+    TypeNode constituentType = TypeNode::fromType(
+        static_cast<ArrayType>(type.toType()).getConstituentType());
+    Node firstElem;
```
ajreynol: I'm not sure we should add a constant for arrays. The issue is that this constant is not handled by the array decision procedure, so we'd fail the verification step of CEGIS.

Regardless, you can use mkGroundTerm or the TypeEnumerator for this type so that you don't have to do special cases here.

But I think we should just do nothing (until the array theory handles array constants).

ajreynol:
Also, can you add (maybe 2) simple examples that require select and store in the solution respectively?

src/theory/quantifiers/sygus/sygus_grammar_cons.cpp
HanielB: Indeed, I just emulated what the type enumerator for arrays does. I'll add an issue for adding a constant array here once we solve #1667.

ajreynol:
Can you check what happens when you have an Array return type but no Array arguments for a function-to-synthesize?  I am guessing something goes wrong (no constructors).  We should perhaps catch this case.

PR #2689, opened 2018-11-05
Change default sygus enumeration mode to auto

ajreynol:
This updates the default sygus enumeration to an "auto" mode, which uses the optimized enumerator for:
- PBE enumerators,
- Single (non-Boolean) function conjectures whose grammars do not contain ITE,
- When using --sygus-stream.

It also disables --sygus-pbe-multi-fair by default, since disabling this option works better with the new default.

This decreases the runtime of regress2 on production on my machine to 331 seconds, down from 365 seconds.

Will post results on sygusComp 2018 soon.


HanielB:
You decided to reuse the active-gen modes for the auto strategy. Isn't it best to use a separate option (whether to use auto-strategy picking or not) which would be on by default?

src/theory/quantifiers/sygus/term_database_sygus.cpp
```diff
@@ -554,6 +559,65 @@ void TermDbSygus::registerEnumerator(Node e,
...
+      // generate a "product" of two actively-generated enumerators. That is,
+      // given a conjecture with two functions-to-synthesize with enumerators
+      // e_f and e_g, and if these enumerators generated:
+      // e_f -> t1, ..., tn
+      // e_g -> s1, ..., sm
+      // The sygus module in charge of this conjecture would expect
+      // constructCandidates calls of the form
+      //   (e_f,e_g) -> (ti, sj)
+      // for each i,j. We instead use passive enumeration in this case.
+      // If the enumerator is constrained, it cannot be actively generated.
```
HanielB: add a blank comment line before this

ajreynol:
I prefer having --sygus-active-gen=auto, since some combinations of flag + active gen modes don't make sense (say, --sygus-active-gen=none --sygus-active-gen-auto). It also allows us to use multiple kinds of active generated enumerators (say if we wanted to use variable-agnostic in combination with sygus-unif, but enum otherwise).

ajreynol:
sygusComp 2018 results: auto solves 1438, optimized enumerator (enum) solves 1434, datatypes solver (old default) solves 1401. Unique solves (e.g. auto is +9-5 vs enum):
```
auto       X       9      38
enum       5       X      43
  dt       1      10       X
```

PR #2690, opened 2018-11-06
Incorporate static PBE symmetry breaking lemmas into SygusEnumerator

ajreynol:
This PR makes the SygusEnumerator capable of interpreting symmetry breaking lemmas provided by an external source such as SygusPbe as an explicit filter on the value it should generate.

This required a bit of refactoring to the sygus extension of datatypes so that it does not process symmetry breaking lemmas for basic enumerators. It also refactors this class a bit to correct some initialization issues where the active guard for enumerators could be uninitialized after the first call to SygusSymBreakNew::check.

This is +2-0 on PBE strings and +0-0 on PBE BV.

### DIFF
--- a/src/theory/quantifiers/sygus/sygus_enumerator.cpp
+++ b/src/theory/quantifiers/sygus/sygus_enumerator.cpp
@@ -44,7 +44,8 @@ void SygusEnumerator::initialize(Node e)
   NodeManager* nm = NodeManager::currentNM();
   std::vector<Node> sbl;
   d_tds->getSymBreakLemmas(e, sbl);
-  TNode ag = d_tds->getActiveGuardForEnumerator(e);
+  Node ag = d_tds->getActiveGuardForEnumerator(e);
+  TNode agt = ag;
   Node truen = nm->mkConst(true);
   TNode truent = truent;
   Assert(d_tcache.find(d_etype) != d_tcache.end());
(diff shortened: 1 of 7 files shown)


src/theory/quantifiers/sygus/sygus_enumerator.cpp
```diff
@@ -31,10 +31,71 @@ SygusEnumerator::SygusEnumerator(TermDbSygus* tds, SynthConjecture* p)
...
+  Assert(d_tcache.find(d_etype) != d_tcache.end());
+  const Datatype& dt = d_etype.getDatatype();
+  for (const Node& lem : sbl)
+  {
+    if (!d_tds->isSymBreakLemmaTemplate(lem))
+    {
+      // substitute its active guard by true and rewrite
+      Node slem = lem.substitute(agt, truent);
+      slem = Rewriter::rewrite(slem);
+      // break into conjuncts
```
HanielB: what if the static sym break lemma is a disjunction of lemmas?

ajreynol: Then we cant use it :)

HanielB: OK. Add a comment and maybe open an issue.

ajreynol: I dont think this is issue worthy (it's not immediately clear that we should use them).  But I'll comment that we dont use them currently.

PR #2692, opened 2018-11-07
Fix collectEmptyEqs in string rewriter

4tXJ7f:
`TheoryStringsRewriter::collectEmptyEqs()` was not checking whether the
input node is a conjunction if it is not an equality. This commitadds
that check.


ajreynol:
Can you also add that triggering benchmark I sent you as a regression?

4tXJ7f:
Done.

PR #2701, opened 2018-11-12
Use https for antlr3.org downloads

tomsmeding:
If CVC4 is built on a machine where wget is not avaiable (while curl is), using http:// URL's in place of https:// URL's for www.antlr3.org (used in contrib/get-antlr-3.4) downloads just the redirect HTML instead of the actual file.
This is an attempt to fix this; another option would be to add automatic redirect following to the curl invocation in `webget`. This is difficult to do correctly, however, since curl will then also output the redirection page HTML together with the downloaded file, which makes the download corrupted.


tomsmeding:
I'm sorry, no, I ran into the issue with the version distributed with the macOS homebrew tap. I believe I checked whether adding -L fixed the issue with curl, but apparently I didn't do that well enough since it does seem to work after all.

So this PR was premature, I fear; feel free to close it if you want to, since as far as I can see the version on master works fine.

tomsmeding:
No problem whatsoever, and thanks for merging!

PR #2703, opened 2018-11-12
Move ss-combine rewrite to extended rewriter

4tXJ7f:
We found that the `ss-combine` rewrite hurts solving performance, so
this commit is moving it to the extended rewriter.


src/theory/strings/theory_strings_rewriter.cpp
```diff
@@ -1822,13 +1832,13 @@ Node TheoryStringsRewriter::rewriteSubstr(Node node)
...
-        Node ret =
-            nm->mkNode(kind::STRING_SUBSTR, node[0][0], new_start, new_len);
+        Node new_start = nm->mkNode(PLUS, start_inner, start_outer);
+        Node ret = nm->mkNode(STRING_SUBSTR, node[0][0], new_start, new_len);
         return returnRewrite(node, ret, "ss-combine");
       }
     }
   }
+
   Trace("strings-rewrite-nf") << "No rewrites for : " << node << std::endl;
```
ajreynol: Maybe remove this trace? Could be confused with the other one on line 1793.

PR #2706, opened 2018-11-13
Obvious rewrites to floating-point < and <=.

martin-cs:
Found via @ajreynol's rewrite tools.


src/theory/fp/theory_fp_rewriter.cpp
```diff
@@ -283,6 +283,33 @@ namespace rewrite {
...
 
+  RewriteResponse leqId(TNode node, bool isPreRewrite)
+  {
+    Assert(node.getKind() == kind::FLOATINGPOINT_LEQ);
+
+    if (node[0] == node[1])
+    {
+      NodeManager *nm = NodeManager::currentNM();
+      return RewriteResponse(
+          REWRITE_DONE,
```
4tXJ7f: If this rewrite is used in a postrewrite, we should probably return `REWRITE_AGAIN_FULL` because it could be that `(fp.isNaN x)` can be rewritten to something simpler, e.g. if `x` is a BV-to-FP conversion where we know that the bits of the exponent.

martin-cs: Fair point.  In postrewrite ```x``` is in canonical form but ```(fp.isNaN x)``` isn't.

martin-cs: Fixed.

martin-cs:
@4tXJ7f No worries; I've got buckets more of these to do, so not urgent.

PR #2707, opened 2018-11-13
 Expand definitions prior to model core computation

ajreynol:
Currently, the model cores computed are highly suboptimal when definitions are present since expand definitions is not called on the input formula. This PR fixes this issue.

### DIFF
--- a/src/smt/smt_engine.cpp
+++ b/src/smt/smt_engine.cpp
@@ -4308,10 +4308,10 @@ Model* SmtEngine::getModel() {
     std::vector<Expr> easserts = getAssertions();
     // must expand definitions
     std::vector<Expr> eassertsProc;
+    std::unordered_map<Node, Node, NodeHashFunction> cache;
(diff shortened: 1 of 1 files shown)


src/smt/smt_engine.cpp
```diff
@@ -4306,7 +4306,16 @@ Model* SmtEngine::getModel() {
     // If we enabled model cores, we compute a model core for m based on our
     // assertions using the model core builder utility
     std::vector<Expr> easserts = getAssertions();
-    ModelCoreBuilder::setModelCore(easserts, m, options::modelCoresMode());
+    // must expand definitions
+    std::vector<Expr> eassertsProc;
+    for (unsigned i = 0, nasserts = easserts.size(); i < nasserts; i++)
+    {
+      Node ea = Node::fromExpr(easserts[i]);
+      std::unordered_map<Node, Node, NodeHashFunction> cache;
```
yoni206: Could the cache be declared outside the loop? Or should it be different for every assertion? otherwise LGTM.

PR #2708, opened 2018-11-13
Fix E-matching for case where candidate generator is not properly initialized

ajreynol:
This came up recently when using --relational-triggers


yoni206:
What was the result of using `--relational-triggers` without initializing the generator, before your fix? 
Was it an exception/assertion error? 
If so, could you please add a regression test that makes sure this is not the case?

PR #2711, opened 2018-11-16
 Clausify context-dependent simplifications in ext theory

ajreynol:
This converts theory lemmas generated by context-dependent simplification to clauses. 
This ensures that parts of the proof infrastructure don't get confused.

### DIFF
--- a/src/theory/ext_theory.cpp
+++ b/src/theory/ext_theory.cpp
@@ -249,15 +249,15 @@ bool ExtTheory::doInferencesInternal(int effort,
             // This ensures the proof infrastructure can process this as a
             // normal theory lemma.
             Node lem = terms[i].eqNode(sr);
-            if( !exp[i].empty() )
+            if (!exp[i].empty())
             {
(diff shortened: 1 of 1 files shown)


src/theory/ext_theory.cpp
```diff
@@ -245,15 +246,24 @@ bool ExtTheory::doInferencesInternal(int effort,
...
+            Node lem = eq;
+            if (!exp[i].empty())
+            {
+              std::vector<Node> eei;
+              for (const Node& e : exp[i])
+              {
+                eei.push_back(e.negate());
+              }
+              eei.push_back(eq);
+              lem = nm->mkNode(kind::OR, eei);
```
4tXJ7f: Suggestion: could use a NodeBuilder here to be slightly more efficient.

ajreynol: This is not performance critical code. I prefer using vectors for cases like this since it makes the code more flexible.

PR #2712, opened 2018-11-16
Change lemma proof step storage & iterators

alex-ozdemir:
Proof steps were in a std::list, which is a linked list, but really, we
only needed a stack, so I changed it to a vector, because LL's are
usually slower.

Also added an iterator for the proof steps, and << implementations


alex-ozdemir:
Ah, I'll investigate the failures.

PR #2719, opened 2018-11-21
Information gain heuristic for PBE

ajreynol:
This implements the standard information gain heuristic for decision tree learning in SygusUnifIo.  It generalizes a look-ahead heuristic, which is removed by this PR.
This leads to more compact solutions, the performance impact looks to be negligible (will verify this with more testing).


HanielB:
I think you should guard the usage of the information gain for solution minimization with an option (which should be on by default).

src/theory/quantifiers/sygus/sygus_unif_io.cpp
```diff
@@ -1454,6 +1377,84 @@ Node SygusUnifIo::constructSol(
...
+  {
+    return SygusUnif::constructBestConditional(ce, solved);
+  }
+  UnifContextIo& x = d_context;
+  // use information gain heuristic
+  Trace("sygus-sui-dt-igain") << "Best information gain in context ";
+  print_val("sygus-sui-dt-igain", x.d_vals);
+  Trace("sygus-sui-dt-igain") << std::endl;
+  std::vector<unsigned> activeIndices;
+  std::map<unsigned, std::map<Node, std::map<Node, unsigned>>> eval;
```
HanielB: comment what this is used for

src/theory/quantifiers/sygus/sygus_unif_io.cpp
```diff
@@ -1454,6 +1377,84 @@ Node SygusUnifIo::constructSol(
...
+  std::map<unsigned, std::map<Node, unsigned>> evalCount;
```
HanielB: ditto

src/theory/quantifiers/sygus/sygus_unif_io.cpp
```diff
@@ -1454,6 +1377,84 @@ Node SygusUnifIo::constructSol(
...
+  unsigned nsolved = solved.size();
+  EnumCache& ecache = d_ecache[ce];
+  std::vector<unsigned> eindex;
+  for (unsigned j = 0; j < nsolved; j++)
```
HanielB: describe loop's goal

src/theory/quantifiers/sygus/sygus_unif_io.cpp
```diff
@@ -1454,6 +1377,84 @@ Node SygusUnifIo::constructSol(
...
+  {
+    eindex.push_back(ecache.d_enum_val_to_index[solved[j]]);
+  }
+  unsigned activePoints = 0;
+  for (unsigned i = 0, npoints = x.d_vals.size(); i < npoints; i++)
```
HanielB: ditto

src/theory/quantifiers/sygus/sygus_unif_io.cpp
```diff
@@ -1454,6 +1377,84 @@ Node SygusUnifIo::constructSol(
...
+      {
+        Node resn = ecache.d_enum_vals_res[eindex[j]][i];
+        Assert(resn.isConst());
+        eval[j][resn][eo]++;
+        evalCount[j][resn]++;
+      }
+    }
+  }
+  AlwaysAssert(activePoints > 0);
+  double minEntropy = 2.0;
```
HanielB: explain

src/theory/quantifiers/sygus/sygus_unif_io.h
```diff
@@ -460,6 +438,13 @@ class SygusUnifIo : public SygusUnif
                     NodeRole nrole,
                     int ind,
                     std::vector<Node>& lemmas) override;
+  /** construct best conditional
+   *
+   * This returns the condition in solved that maximizes information gain with
+   * respect to the current active points in d_context.
```
HanielB: explain in detail how this is computed (i.e. refer to Alur et al. TACAS 2017)

src/theory/quantifiers/sygus/sygus_unif.h
```diff
@@ -156,18 +156,29 @@ class SygusUnif
...
-   * x, currently  return the first. */
-  virtual Node constructBestStringSolvedTerm(const std::vector<Node>& solved);
-  /** Heuristically choose the best solved conditional term  from solved in
-   * context x, currently random */
-  virtual Node constructBestSolvedConditional(const std::vector<Node>& solved);
-  /** Heuristically choose the best conditional term  from conds in context x,
-   * currently random */
-  virtual Node constructBestConditional(const std::vector<Node>& conds);
+  /**
+   * Heuristically choose the best solved term for enumerator e,
```
HanielB: explain what it is a "solved term"

src/theory/quantifiers/sygus/sygus_unif.h
```diff
@@ -156,18 +156,29 @@ class SygusUnif
...
+   */
+  virtual Node constructBestSolvedTerm(Node e, const std::vector<Node>& solved);
+  /**
+   * Heuristically choose the best solved string term for enumerator e,
+   * return the first by default.
+   */
+  virtual Node constructBestStringSolvedTerm(Node e,
+                                             const std::vector<Node>& solved);
+  /**
+   * Heuristically choose the best solved conditional term for condition
```
HanielB: explain what it is a "solved conditional"

src/theory/quantifiers/sygus/sygus_unif_io.h
```diff
@@ -460,6 +438,14 @@ class SygusUnifIo : public SygusUnif
...
                     int ind,
                     std::vector<Node>& lemmas) override;
+  /** construct best conditional
+   *
+   * This returns the condition in solved that maximizes information gain with
+   * respect to the current active points in d_context. For example, see
+   * Alur TACAS 2017 for an example of information gain.
+   */
+  Node constructBestConditional(Node ce,
+                                const std::vector<Node>& solved) override;
```
HanielB: change `solved` to `conds`

src/theory/quantifiers/sygus/sygus_unif_io.h
```diff
@@ -460,6 +438,14 @@ class SygusUnifIo : public SygusUnif
                     NodeRole nrole,
                     int ind,
                     std::vector<Node>& lemmas) override;
+  /** construct best conditional
+   *
+   * This returns the condition in solved that maximizes information gain with
+   * respect to the current active points in d_context. For example, see
+   * Alur TACAS 2017 for an example of information gain.
```
HanielB: Alur et al.

src/theory/quantifiers/sygus/sygus_unif_io.cpp
```diff
@@ -1454,6 +1377,93 @@ Node SygusUnifIo::constructSol(
...
+      unsigned ecount = evalCount[j][ej.first];
+      if (ecount > 0)
+      {
+        double probBranch = double(ecount) / double(activePoints);
+        Trace("sygus-sui-dt-igain") << ej.first << " -> ( ";
+        for (std::pair<const Node, unsigned>& eej : ej.second)
+        {
+          if (eej.second > 0)
+          {
+            double probVal = double(eej.second) / double(ecount);
```
HanielB: So if I understand correctly you are looking only at the actual output of each example and not at all at the terms to be assigned?

It would be best to have an explicit description of how (based on what parameters) you are computing the entropy, which is the key thing here.

ajreynol: I've added an explanation

src/theory/quantifiers/sygus/sygus_unif_io.cpp
```diff
@@ -1427,8 +1427,15 @@ Node SygusUnifIo::constructBestConditional(Node ce,
...
-  for (unsigned j = 0; j < nsolved; j++)
+  for (unsigned j = 0; j < nconds; j++)
   {
+    // To compute the entropy for a condition C, for pair of terms (s, t), let
+    //   prob(t) be the probability C evaluates to t on an active point,
+    //   prob(s|t) be the probability that an active point on which C
+    //     evaluates to t has output s.
+    // Then, the entropy of C is:
+    //   sum{t}. prob(t)*( sum{s}. -prob(s|t)*log2(prob(s|t)) )
+    // where notice this is always between 0 and 1.
```
HanielB: If I understand correctly you are optimizing for conditions that lead to sets of points with the same output. This is fine if you have only T/F as output, but for the general case I think you really need to take into account the terms you are assigning to the points. This is what is done in Alur et al. TACAS'17.

HanielB: I meant that for T/F this is already optimal. But yeah, that can be addressed in a future PR. Please open an issue to keep track of it.

Issue #2720, opened 2018-11-25
Echo doesn't flush output


agacek:
Starting sometime after CVC 1.4, the `echo` command no longer causes output to be flushed. Here is Python code demonstrating the issue:

```
import subprocess

p = subprocess.Popen(['cvc4', '--lang', 'smt'], stdout=subprocess.PIPE, stdin=subprocess.PIPE)
p.stdin.write(b'(echo "foo")\n')
p.stdin.flush()
print(p.stdout.readline())
```

With `cvc4-1.4-win32-opt.exe` the output is

```
b'"foo"\r\n'
```

With `cvc4-1.6-win64-opt.exe`, there is no output and the test program hangs.

agacek:
Git bisect says the issue was introduced in 0da095c3f5be79a85c085b4b46d7a0a0513ecdd6. Based on that commit, perhaps the issue is that CVC4 is buffering the input?

4tXJ7f:
The issue should be fixed on master now. Tomorrow morning, a fixed Windows build should be available at http://cvc4.cs.stanford.edu/downloads/builds/win64-opt/unstable/ (assuming that our nightly tests pass). If you find time, it would be great if you could confirm that the issue has been fixed.

agacek:
Yes, I confirm that this resolves the original issue which led to this test case. Thanks!

PR #2721, opened 2018-11-27
Reduce lookahead when parsing string literals

4tXJ7f:
Fixes issue #2720. Between versions 2.0 and 2.5, the SMT-LIB standard changed
the way that string literals esacpe double quotes (") within strings from \" to
"". As a consequence, we have two different ways of lexing string literals
depending on the language used. The code produced by ANTLR first uses a
Deterministic Finite Automata (DFA) to predict the next token. Because the DFA
is static, it considers both versions of the string literal. If we are parsing
a string without escape sequences, e.g. "foo", it ends up reading three
characters past the closing double quote (I am not sure whether this is by
design or a bug). This is problematic in interactive mode because the third
character may be on the next line, which may not be available yet. Consider the
example from the issue mentioned above:

```
import subprocess

p = subprocess.Popen(['cvc4', '--lang', 'smt'], stdout=subprocess.PIPE, stdin=subprocess.PIPE)
p.stdin.write(b'(echo "foo")\n')
p.stdin.flush()
print(p.stdout.readline())
```

In this example, "foo" is a valid string for both types of string literals and
ANTLR reads ", \n, and then tries to access the character on the next line,
which leads to a deadlock because CVC4 is waiting for the next line on stdin
and the Python program is waiting for the output of CVC4. Note: this is why
writing `(echo "foo")\n\n` or `(echo "foo" )\n` both do not lead to a deadlock.

The solution this commit implements is to reorganize the way the string
literals are parsed: we now have a single token `STRING_LITERAL` for both types
of string literals, which is actually a bit simpler because it allows us to
share common code for processing the literal. The code that ANTLR generates
for that pattern does not exhibit the problem described above. Additionally, the
commit removes `SYGUS_QUOTED_LITERAL`, which was unused and was also
had overlapping matches with the string literals.

As to why it worked with older versions: we allocated memory for the input in
increments of 1024 bytes and in our `myConsume` method, we did not actually
check whether we were reading a character from the input or just from the
buffer that had not been filled in yet---so I am assuming that we got lucky.

### DIFF
--- a/src/parser/smt2/Smt2.g
+++ b/src/parser/smt2/Smt2.g
@@ -2611,64 +2611,49 @@ termList[std::vector<CVC4::Expr>& formulas, CVC4::Expr& expr]
  * Matches a string, and strips off the quotes.
  */
 str[std::string& s, bool fsmtlib]
-  : STRING_LITERAL_2_0
-    { s = AntlrInput::tokenText($STRING_LITERAL_2_0);
+  : STRING_LITERAL
+    {
+      s = AntlrInput::tokenText($STRING_LITERAL);
       /* strip off the quotes */
       s = s.substr(1, s.size() - 2);
-      for(size_t i=0; i<s.size(); i++) {
-        if((unsigned)s[i] > 127 && !isprint(s[i])) {
-          PARSER_STATE->parseError("Extended/unprintable characters are not "
-                                   "part of SMT-LIB, and they must be encoded "
-                                   "as escape sequences");
+      for (size_t i = 0; i < s.size(); i++)
+      {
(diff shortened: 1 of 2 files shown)


4tXJ7f:
A couple of additional notes:

- The issue also did not appear when a string literal was not matched by both versions of the string literal rule (e.g. `"abc ""foo"" cba"`, which is only valid in newer versions of SMT-LIB).
- I was able to reproduce the issue on Linux (the original issue mentioned the CVC4 Windows binaries).
- I've confirmed that the behavior is the same when reading from a file by hacking the ANTLR library to provide debug output in `antlr38BitConsume()`.

PR #2722, opened 2018-11-27
[LRA proof] More complete LRA example proofs.

alex-ozdemir:
Short Version:

Refined the LRA signature and used the refined version to write two new
test proofs which are close to interface compatible with the LRA proofs
that CVC4 will produce.

Love Version:

LRA proofs have the following interface:
   * Given predicates between real terms
   * Prove bottom

However, even though the type of the interface does not express this,
the predicates are **linear bounds**, not arbitrary real bounds. Thus
LRA proofs have the following structure:

   1. Prove that the input predicates are equivalent to a set of linear
      bounds.
   2. Use the linear bounds to prove bottom using farkas coefficients.

Notice that the distinction between linear bounds (associated in the
signature with the string "poly") and real predicates (which relate
"term Real"s to one another) matters quite a bit. We have certain inds
of axioms for one, and other axioms for the other.

The signature used to muddy this distinction using a constructor called
"term_poly" which converted between them. I decided it was better to buy
into the distinction fully.

Now all of the axioms for step (2) use the linear bounds and axioms for
step (1) use both kinds of bounds, which makes sense because step (1) is
basically a conversion.

Also had to add an axiom or two, because some were missing.


proofs/signatures/th_lra_test.plf
```diff
@@ -1,4 +1,5 @@
 ; Depends On: th_lra.plf
+;; Polynomial proof
```
yoni206: Can you provide a short explanation before each test saying what does it prove? In this example, it would be something like:
```suggestion
;;This is a proof that the following set of constraints in UNSAT in LRA:
; 2 + (-x) + (-1/2)y >=0
; -8 + x + y >=0
; 0 + x - y >=0
```

proofs/signatures/th_lra_test.plf
```diff
@@ -11,9 +12,9 @@
   (@ p1 (polyc 2/1 m1)
   (@ p2 (polyc (~ 8/1) m2)
   (@ p3 (polyc 0/1 m3)
-  (% pf_nonneg_1 (th_holds (>=0_Real (poly_term p1)))
-  (% pf_nonneg_2 (th_holds (>=0_Real (poly_term p2)))
-  (% pf_nonneg_3 (th_holds (>=0_Real (poly_term p3)))
+  (% pf_nonneg_1 (th_holds (>=0_poly p1))
+  (% pf_nonneg_2 (th_holds (>=0_poly p2))
+  (% pf_nonneg_3 (th_holds (>=0_poly p3))
      (lra_contra_>=
```
yoni206: I think that currently in cvc4, it is usually crucial to know what the proof proves, and not only that it is a valid proof. If that is the case with LRA proofs as well, I would suggest including a `( : (holds ...)` in every test proof.

proofs/signatures/th_lra.plf
```diff
@@ -102,59 +120,65 @@
 
 (declare poly_term (! p poly (term Real)))
```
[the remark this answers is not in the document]
alex-ozdemir: Good catch.

Agreed. I don't use any of the axioms like `poly_norm_=` anyways, so I'll just remove them.

proofs/signatures/th_lra_test.plf
alex-ozdemir: That's a good idea. I'll do so!

proofs/signatures/th_lra_test.plf
alex-ozdemir: You're right.
I got out of the habit when I was writing tests for `program`s, but I should get back into the habit.
I'll add the annotations.

alex-ozdemir:
Okay, I believe that I have addressed all of the comments. Let me know if that is not the case!

PR #2726, opened 2018-11-28
Combine sygus stream with PBE

ajreynol:
Previously, using "--sygus-stream --sygus-pbe" was incompatible (PBE cached a solution once it found one and then didnt build new ones). 

The PR makes it so that when these two modes are used in combination, the PBE solver streams solutions that smaller and smaller. This allows a "progressive generalization" for PBE problems, which is useful in applications, in particular has been very useful for floating point invertibility conditions.

It also includes some minor improvements to the information gain heuristic, and various fixes to ensure that minimal size terms can be selected at the leafs in decision tree learning.


src/theory/quantifiers/sygus/sygus_unif_io.cpp
```diff
@@ -238,11 +238,14 @@ Node SubsumeTrie::addTermInternal(Node t,
...
-          d_term = Node::null();
-        }
+        // If we are only interested in feasibility, we could set d_term to null
+        // here. However, d_term still could be useful, since it may be
+        // smaller than t and suffice as a solution under some condition.
+        // As a simple example, consider predicate synthesis and a case where we
+        // enumerate a C that is correct for all I/O points whose output is
+        // true. Then, C subsumes true. However, true may be preferred, e.g.
+        // to generate a solution ite( C, true, D ) instead of ite( C, C, D ),
+        // since true is conditionally correct under C, and is smaller than C.
```
HanielB: Did you test whether this leads to performance loss?

ajreynol: No performance impact: https://www.starexec.org/starexec/secure/details/job.jsp?id=31478

HanielB: I don't have access but I believe you. :)

PR #2728, opened 2018-11-29
Optimizations for PBE strings

ajreynol:
This includes:
- Caching solutions that satisfy *some* but not all I/O pairs, which is useful for non-deterministic strategies that are combined with ITE,
- Not registering terms to concatenation PBE pool that are not contained in all examples.


ajreynol:
I've addressed the review in my latest commit.

PR #2729, opened 2018-11-29
 Infrastructure for sygus side conditions

ajreynol:
This adds a quantifiers attribute for "sygus side condition", which is a quantifier-free condition which a solution to a sygus conjecture must satisfy.

This is part 3 / 5 for sygus abduction.


src/theory/quantifiers/quantifiers_attributes.h
```diff
@@ -73,6 +73,17 @@ struct SygusPrintProxyAttributeId
 typedef expr::Attribute<SygusPrintProxyAttributeId, Node>
     SygusPrintProxyAttribute;
 
+/** Attribute for specify a "side condition" for a sygus function-to-synthesize
```
HanielB: for specifying?

PR #2732, opened 2018-12-03
Arith Constraint Proof Loggin

alex-ozdemir:
Added some debugging statements to reveal how constraints in the arithmetic theory are justified.

Also a small documentation elaboration.


src/theory/arith/constraint.cpp
```diff
@@ -1196,6 +1201,15 @@ void Constraint::impliedByIntHole(ConstraintCP a, bool nowInConflict){
 }
 
 void Constraint::impliedByIntHole(const ConstraintCPVec& b, bool nowInConflict){
+  Debug("constraints::pf") << "impliedByIntHole(" << this;
+  if (Debug.isOn("constraints::pf")) {
+    for (const ConstraintCP p : b)
```
aniemetz: Nit: `const ConstraintCP&`

src/theory/arith/constraint.cpp
```diff
@@ -1224,6 +1238,15 @@ void Constraint::impliedByIntHole(const ConstraintCPVec& b, bool nowInConflict){
  *   coeff.back() corresponds to the current constraint. 
  */
 void Constraint::impliedByFarkas(const ConstraintCPVec& a, RationalVectorCP coeffs, bool nowInConflict){
+  Debug("constraints::pf") << "impliedByFarkas(" << this;
+  if (Debug.isOn("constraints::pf")) {
+    for (const ConstraintCP p : a)
```
aniemetz: Ditto

src/theory/arith/constraint.cpp
alex-ozdemir: Question: Do you always have your iteratee be a reference, even if the type is small enough for copying to be a non-issue?

aniemetz: Not for base types like int, bool, ... ConstraintCP is a pointer, right? Didn't realize that this is the case when I reviewed, so in this case not using a reference is fine. Doesn't hurt, though.

PR #2733, opened 2018-12-03
Enable BV proofs when using an eager bitblaster

alex-ozdemir:
Specifically:
   * Removed assertions that blocked them.
   * Made sure that only bitvectors were stored in the BV const let-map
   * Prevented true/false from being bit-blasted by the eager bitblaster

Also:
   * uncommented "no-check-proofs" from relevant tests

Tested with `make check` which (after this PR) includes 3 tests of eager bitblasting proofs.

### DIFF
--- a/src/options/options_handler.cpp
+++ b/src/options/options_handler.cpp
@@ -1632,7 +1632,14 @@ void OptionsHandler::setProduceAssertions(std::string option, bool value)
 
 void OptionsHandler::proofEnabledBuild(std::string option, bool value)
 {
-#ifndef CVC4_PROOF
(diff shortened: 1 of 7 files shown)


mpreiner:
This only works with bvminisat right?

alex-ozdemir:
Yes, this only works with bvminisat.

Do you think I should leave the assertions, but change them?

mpreiner:
We should add a check to ensure that we have the right solver. An assertion is not enough.

alex-ozdemir:
Okay, per our offline discussion I understand what you meant by "checking",  and have modified the options handler accordingly.

PR #2737, opened 2018-12-05
[LRAT] A C++ data structure for LRAT.

alex-ozdemir:
Added a data structure for storing (abstract) LRAT proofs.

Includes
  * Text output
  * Text input

The constructor will take a drat binary proof and convert it to LRAT
using drat-trim. However, this is unimplemented in this PR.

Subsequent PRs will add:
   * LFSC representation of LRAT
   * Bitvector Proofs based on LRAT
   * Enabled tests for those proofs


alex-ozdemir:
Hmm. Travis has one test failing, and that test doesn't fail on my laptop or workstation (which uses GCC).

I'm going to rerun Travis.

alex-ozdemir:
Looks like it consistently fails... Some error coming out of the allocator.. Maybe we're clobbering some of it's memory?

This is  exceedingly odd, since I've only added dead code. I'll have to look into it more closely.

4tXJ7f:
Hahaha, interesting issue. I can take a look.

src/proof/lrat/lrat_proof.h
```diff
@@ -0,0 +1,207 @@
...
+    textOut(o, l) << " ";
+  }
+  return o << "0";
+}
+
+// Prints the trace as a space-separated list of (+) ints with a space at the
+// end.
+inline std::ostream& operator<<(std::ostream& o, const LRATUPTrace& trace)
+{
+  for (const auto i : trace)
```
4tXJ7f: Maybe make this `const auto&` to be consistent with `const auto& idx : del.clauses`:
```suggestion
  for (const auto& i : trace)
```

src/proof/lrat/lrat_proof.cpp
```diff
@@ -0,0 +1,199 @@
...
+  // TODO(aozdemir) Add invocation of DRAT trim, once I get CMake to bundle it
+  // into CVC4 correctly.
+  Unimplemented();
+
+  std::ifstream lratStream(lratFilename);
+  LRATProof lratProof(lratStream);
+  remove(formulaFilename);
+  remove(dratFilename);
+  remove(lratFilename);
+  *this = std::move(lratProof);
```
4tXJ7f: Instead of doing this (I am actually not quite sure whether the C++ standard allows that or not), I'd suggest using a static class method `static std::unique_ptr<LratProof> LratProof::fromDratProof(...);`.

src/proof/lrat/lrat_proof.cpp
```diff
@@ -0,0 +1,199 @@
...
+}
+
+std::istream& operator>>(std::istream& in, prop::SatLiteral& l)
+{
+  long int i;
```
4tXJ7f: This should be a `uint64_t` or a `size_t`. [Note that `long int` is actually 32-bit on Windows (even for 64-bit programs).](https://en.wikipedia.org/wiki/64-bit_computing#64-bit_data_models)

src/proof/lrat/lrat_proof.cpp
```diff
@@ -0,0 +1,199 @@
...
+      Assert(!firstS.fail(), "Couldn't parse first literal from addition line");
+      for (; lit != 0; textualProof >> lit)
+      {
+        Assert(textualProof.good());
+        SatLiteral l(lit.getSatVariable() - 1, lit.isNegated());
+        instr.additionData.clause.push_back(l);
+      }
+
+      // Now we read the AT UP trace. It ends at the first non-(+) #
+      long int i;
```
4tXJ7f: Make this a `uint64_t` or `size_t` (see above comment).

src/proof/lrat/lrat_proof.cpp
```diff
@@ -0,0 +1,199 @@
...
+      textualProof >> i;
+      for (; i > 0; textualProof >> i)
+      {
+        Assert(textualProof.good());
+        instr.additionData.atTrace.push_back(i);
+      }
+
+      // For each RAT hint... (each RAT hint starts with a (-).
```
4tXJ7f:
```suggestion
      // For each RAT hint... (each RAT hint starts with a (-)).
```

src/proof/lrat/lrat_proof.cpp
```diff
@@ -0,0 +1,199 @@
...
+
+  std::ifstream lratStream(lratFilename);
+  LRATProof lratProof(lratStream);
+  remove(formulaFilename);
+  remove(dratFilename);
+  remove(lratFilename);
+  *this = std::move(lratProof);
+}
+
+std::istream& operator>>(std::istream& in, prop::SatLiteral& l)
```
4tXJ7f: For consistency:

```suggestion
std::istream& operator>>(std::istream& in, SatLiteral& l)
```

src/proof/lrat/lrat_proof.cpp
```diff
@@ -0,0 +1,199 @@
...
+{
+  long int i;
+  in >> i;
+  l = prop::SatLiteral(labs(i), i < 0);
```
4tXJ7f: For consistency:
```suggestion
  l = prop::SatLiteral(labs(i), i < 0);
```

src/proof/lrat/lrat_proof.cpp
```diff
@@ -0,0 +1,199 @@
...
+      std::istringstream firstS(first);
+      SatLiteral lit;
+      firstS >> lit;
+      Trace("pf::lrat") << "First lit: " << lit << std::endl;
+      Assert(!firstS.fail(), "Couldn't parse first literal from addition line");
+      for (; lit != 0; textualProof >> lit)
+      {
+        Assert(textualProof.good());
+        SatLiteral l(lit.getSatVariable() - 1, lit.isNegated());
+        instr.additionData.clause.push_back(l);
```
4tXJ7f: [Optional] Suggestion: you can combine the above two lines to `instr.additionData.clause.emplace_back(lit.getSatVariable() - 1, lit.isNegated());`, avoiding a copy of the `SatLiteral`.

src/proof/lrat/lrat_proof.cpp
```diff
@@ -0,0 +1,199 @@
...
+        instr.additionData.atTrace.push_back(i);
+      }
+
+      // For each RAT hint... (each RAT hint starts with a (-).
+      for (; i<0; textualProof>> i)
+      {
+        Assert(textualProof.good());
+        // Create an entry in the RAT hint list
+        instr.additionData.resolvants.push_back(
+            make_pair(-i, std::vector<ClauseIdx>()));
```
4tXJ7f: [Optional] As with the suggestion above, this could be written as `instr.additionData.resolvants.emplace_back(-i, std::vector<ClauseIdx>())`.

src/proof/lrat/lrat_proof.h
```diff
@@ -0,0 +1,207 @@
...
+namespace proof {
+namespace lrat {
+
+// Refers to clause position within an LRAT proof
+using ClauseIdx = size_t;
+
+enum LRATInstructionKind
+{
+  lratDeletion,
+  lratAddition,
```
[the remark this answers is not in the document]
alex-ozdemir: Sounds good!

src/proof/lrat/lrat_proof.h
```diff
@@ -0,0 +1,207 @@
...
+  }
+
+ private:
+  // The instructions in the proof. Each is a deletion or addition.
+  std::vector<LRATInstruction> d_instructions;
+};
+
+// Prints the literal as a (+) or (-) int
+// Not operator<< b/c that represents negation as ~
+inline std::ostream& textOut(std::ostream& o, const prop::SatLiteral& l)
```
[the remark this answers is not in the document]
alex-ozdemir: Sounds good!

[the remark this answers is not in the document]
alex-ozdemir: I'll move all declarations but `operator<<` for `LratProof` to the implementation file.

src/proof/lrat/lrat_proof.h
```diff
@@ -0,0 +1,207 @@
...
+#include "prop/sat_solver.h"
+
+namespace CVC4 {
+namespace proof {
+namespace lrat {
+
+// Refers to clause position within an LRAT proof
+using ClauseIdx = size_t;
+
+enum LRATInstructionKind
```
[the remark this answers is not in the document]
alex-ozdemir: Sounds good.

src/proof/lrat/lrat_proof.cpp
alex-ozdemir: Sounds good!

src/proof/lrat/lrat_proof.h
```diff
@@ -0,0 +1,207 @@
...
+ **/
+
+#include "cvc4_private.h"
+
+#ifndef __CVC4__PROOF__LRAT__LRAT_PROOF_H
+#define __CVC4__PROOF__LRAT__LRAT_PROOF_H 0
+
+#include <vector>
+#include <unordered_map>
+#include <string>
```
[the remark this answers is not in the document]
alex-ozdemir: Sounds good!

src/proof/lrat/lrat_proof.h
```diff
@@ -0,0 +1,207 @@
...
+{
+  lratDeletion,
+  lratAddition,
+};
+
+struct LRATDeletionData
+{
+  // This idx doesn't really matter, but it's in the format anyway, so we parse
+  // it.
+  ClauseIdx idxOfClause;
```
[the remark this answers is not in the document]
alex-ozdemir: Sounds good!

src/proof/lrat/lrat_proof.h
```diff
@@ -0,0 +1,207 @@
...
+  LRATUPTrace atTrace;
+
+  // Clauses that can resolve with `clause` on its first variable,
+  // together with a UP trace after that resolution.
+  // Used for RAT checks.
+  std::vector<std::pair<ClauseIdx, LRATUPTrace>> resolvants;
+};
+
+// This is conceptually an Either<Addition,Deletion>
+struct LRATInstruction
```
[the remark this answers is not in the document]
alex-ozdemir: My reason for choosing a disjoint union over an inheritance hierarchy was that I thought it would produce more understandable code.  In particular, when implementing a particular **verb** for an inheritance hierarchy, the code get's split up, while the implementation  of a particular **verb** for a disjoint union is all in one places.

That being said, while writing this code I didn't (yet) know how to implement disjoin  unions correctly in C++. Now I do, so I'll clean it up. Take a look at the result, and if you'd like, I'm happy to continue the conversation.

src/proof/lrat/lrat_proof.cpp
alex-ozdemir: Sounds good. I'll make it an `int64_t`.

src/proof/lrat/lrat_proof.cpp
alex-ozdemir: Sounds good. I'll make it an `int64_t`.

src/proof/lrat/lrat_proof.cpp
alex-ozdemir: Sounds good.

src/proof/lrat/lrat_proof.cpp
alex-ozdemir: Sounds good.

src/proof/lrat/lrat_proof.cpp
```diff
@@ -0,0 +1,199 @@
...
+  char dratFilename[] = "/tmp/cvc4-drat-XXXXXX";
+  char lratFilename[] = "/tmp/cvc4-lrat-XXXXXX";
+  int r;
+  r = mkstemp(formulaFilename);
+  Assert(r > 0);
+  r = mkstemp(dratFilename);
+  Assert(r > 0);
+  r = mkstemp(lratFilename);
+  Assert(r > 0);
+  std::ofstream formStream(formulaFilename);
```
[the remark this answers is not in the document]
alex-ozdemir: I'd  like to avoid the temporary files, but  I'm not sure how to.

In particular, we're going to link against drat-trim, so even its interface supports receiving one of its two in puts on stdin, its stdin is our stdin, so I'm not sure if we can use that part of the interface.

I could spend some time figuring out memfd, but I suspect that would be hard to do cross-platform.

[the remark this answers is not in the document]
alex-ozdemir: In the meantime, I'll  close the fd produced by `mkstemp`.

src/proof/lrat/lrat_proof.cpp
alex-ozdemir: I like the idea of a static constructor-like function (incidentally, that's how **all** Rust constructors work). How about just returning an `LratProof`, and letting return-value-optimization do its work?

4tXJ7f: Yes, that's actually even better.

alex-ozdemir:
Thanks for the feedback! I believe that I have either replied to or directly addressed each of the comments. If I missed any, let me know.

src/proof/lrat/lrat_proof.cpp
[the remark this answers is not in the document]
4tXJ7f: I didn't realize that we were going to link against drat trim. I was somehow assuming that we would execute it as a separate process. I think that the current solution is ok for now and we can try to make it file-free later

src/proof/lrat/lrat_proof.cpp
```diff
@@ -0,0 +1,345 @@
...
+      firstS >> lit;
+      Trace("pf::lrat") << "First lit: " << lit << std::endl;
+      Assert(!firstS.fail(), "Couldn't parse first literal from addition line");
+
+      SatClause clause;
+      for (; lit != 0; textualProof >> lit)
+      {
+        Assert(textualProof.good());
+        clause.emplace_back(
+            SatLiteral(lit.getSatVariable() - 1, lit.isNegated()));
```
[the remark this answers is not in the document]
alex-ozdemir: Sounds good. Will do!

src/proof/lrat/lrat_proof.cpp
```diff
@@ -0,0 +1,345 @@
...
+                                           std::move(atTrace),
+                                           std::move(resolvants))));
+    }
+    Debug("pf::lrat") << "Instr: " << d_instructions.back() << std::endl;
+  }
+}
+
+// Prints the literal as a (+) or (-) int
+// Not operator<< b/c that represents negation as ~
+inline std::ostream& textOut(std::ostream& o, const SatLiteral& l)
```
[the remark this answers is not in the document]
alex-ozdemir: Sounds good! I'll create an unamed namespace.

src/proof/lrat/lrat_proof.cpp
```diff
@@ -0,0 +1,345 @@
...
+    const std::vector<ClauseId>& clauseOrder,
+    const std::string& dratBinary)
+{
+  std::ostringstream cmd;
+  char formulaFilename[] = "/tmp/cvc4-dimacs-XXXXXX";
+  char dratFilename[] = "/tmp/cvc4-drat-XXXXXX";
+  char lratFilename[] = "/tmp/cvc4-lrat-XXXXXX";
+  int r;
+  r = mkstemp(formulaFilename);
+  Assert(r > 0);
```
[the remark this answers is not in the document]
alex-ozdemir: Good idea. Will do!

alex-ozdemir:
Mk, that serves me right for pushing without compiling. I misplaced a function, and had a few other issues with the anonymous namespace.

src/proof/lrat/lrat_proof.cpp
```diff
@@ -239,7 +311,7 @@ LratProof::LratProof(std::istream& textualProof)
       {
         Assert(textualProof.good());
         // Create an entry in the RAT hint list
-        resolvants.emplace_back(make_pair(-i, std::vector<ClauseIdx>()));
+        resolvants.emplace_back(i, std::vector<ClauseIdx>());
```
4tXJ7f: It looks like you changed from `-i` to `i`. Was that on purpose?

alex-ozdemir: Good catch, thanks!

src/proof/lrat/lrat_proof.cpp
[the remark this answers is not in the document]
alex-ozdemir: I've now removed the `inline` annotations.

src/proof/lrat/lrat_proof.cpp
```diff
@@ -0,0 +1,345 @@
...
+  remove(dratFilename);
+  remove(lratFilename);
+  return lrat;
+}
+
+std::istream& operator>>(std::istream& in, SatLiteral& l)
+{
+  int64_t i;
+  in >> i;
+  l = SatLiteral(labs(i), i < 0);
```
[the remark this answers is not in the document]
alex-ozdemir: That is nifty. Yeah, seems like the right thing to do.

src/proof/lrat/lrat_proof.h
```diff
@@ -0,0 +1,164 @@
...
+  LratInstructionKind d_kind;
+  union LratInstructionData
+  {
+    LratAdditionData d_addition;
+    LratDeletionData d_deletion;
+    ~LratInstructionData(){/* Empty destructor */};
+    LratInstructionData(){/* Empty constructor */};
+  } d_data;
+
+ public:
```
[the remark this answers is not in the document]
alex-ozdemir: You're right. I'll remove  it.

alex-ozdemir:
I believe that I've responded to or addressed all of your comments. If that isn't the case, let me know!

PR #2738, opened 2018-12-05
 Take into account minimality and types for cached PBE solutions

ajreynol:
Fixes a bug where we could have a type mismatch on cached PBE solutions, and an issue with minimality (cached solutions didnt take minimality into account).

### DIFF
--- a/src/theory/quantifiers/sygus/sygus_unif_io.cpp
+++ b/src/theory/quantifiers/sygus/sygus_unif_io.cpp
@@ -1013,6 +1013,7 @@ Node SygusUnifIo::constructSol(
   bool retValMod = x.isReturnValueModified();
 
   Node ret_dt;
+  Node cached_ret_dt;
   if (nrole == role_equal)
   {
     if (!retValMod)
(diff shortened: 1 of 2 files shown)


src/theory/quantifiers/sygus/sygus_unif_io.cpp
```diff
@@ -1135,12 +1136,25 @@ Node SygusUnifIo::constructSol(
...
+        else
+        {
+          ret_dt = *intersection.begin();
+        }
         if (Trace.isOn("sygus-sui-dt"))
         {
           indent("sygus-sui-dt", ind);
           Trace("sygus-sui-dt") << "ConstructPBE: found in cache: ";
-          TermDbSygus::toStreamSygus("sygus-sui-dt", ret_dt);
+          TermDbSygus::toStreamSygus("sygus-sui-dt", *intersection.begin());
```
HanielB: This whole debug block should move within the `else`, no? And I think you probably want to print something in the spirit of this block in the case that you are looking for minimal solutions.

PR #2743, opened 2018-12-11
[LRAT] signature robust against duplicate literals

alex-ozdemir:
The LRAT signature previously had complex, surprising, and occasionally
incorrect behavior when given clauses with duplicate literals.

Now it does not. Now clauses have true set semantics, and clauses with
duplicate literals are treated identically to those without.


proofs/signatures/lrat_test.plf
```diff
@@ -689,15 +741,15 @@
...
       (@ pf_cmap
-         (CMapc_proof 1 _ _ pf_c1
-         (CMapc_proof 2 _ _ pf_c2
-         (CMapc_proof 3 _ _ pf_c3
-         (CMapc_proof 4 _ _ pf_c4
-         (CMapc_proof 5 _ _ pf_c5
-         (CMapc_proof 6 _ _ pf_c6
-         (CMapc_proof 7 _ _ pf_c7
-         (CMapc_proof 8 _ _ pf_c8
-         (CMapc_proof 9 _ _ pf_c9
```
yoni206: Since duplicate clauses seems like a delicate issue, maybe it would be a good idea to add a test in which clause1 and clause9 are identical modulo ordering (e.g. clause9 is ~v3,v2,v1)?

alex-ozdemir: Good idea! I'll add another test to that effect.

PR #2744, opened 2018-12-11
Sygus abduction feature

ajreynol:
Adds `--sygus-abduct` preprocessing option.  

Apart from the preprocessing pass, the main change in this PR is the ability to mark SmtEngine as an "internal subsolver", that is, an instance of the SmtEngine class that was constructed internally by one of our modules.  This is required to support --sygus-abduct within going into an infinite loop where the subsolver also applies the abduction preprocessing pass. 

I'm open to suggestions about this design.  Also notice that if it was easy to pass new option objects to SmtEngine, this would not be necessary.


4tXJ7f:
@ajreynol Could you also update the NEWS file?

src/preprocessing/passes/sygus_abduct.cpp
```diff
@@ -0,0 +1,174 @@
...
+  Trace("sygus-abduct-debug") << "...finish" << std::endl;
+
+  res = theory::Rewriter::rewrite(res);
+
+  Trace("sygus-abduct") << "Generate: " << res << std::endl;
+
+  Node trueNode = nm->mkConst(true);
+
+  assertionsToPreprocess->replace(0, res);
+  for (unsigned i = 1, size = assertionsToPreprocess->size(); i < size; ++i)
```
4tXJ7f:
```suggestion
  for (size_t i = 1, size = assertionsToPreprocess->size(); i < size; ++i)
```

src/preprocessing/passes/sygus_abduct.cpp
```diff
@@ -0,0 +1,174 @@
+/*********************                                                        */
+/*! \file sygus_abduct.cpp
+ ** \verbatim
+ ** Top contributors (to current version):
+ **   Andrew Reynolds
+ ** This file is part of the CVC4 project.
+ ** Copyright (c) 2009-2018 by the authors listed in the file AUTHORS
```
4tXJ7f:
```suggestion
 ** Copyright (c) 2009-2019 by the authors listed in the file AUTHORS
```

src/preprocessing/passes/sygus_abduct.h
```diff
@@ -0,0 +1,72 @@
+/*********************                                                        */
+/*! \file sygus_abduct.h
+ ** \verbatim
+ ** Top contributors (to current version):
+ **   Andrew Reynolds
+ ** This file is part of the CVC4 project.
+ ** Copyright (c) 2009-2018 by the authors listed in the file AUTHORS
```
4tXJ7f:
```suggestion
 ** Copyright (c) 2009-2019 by the authors listed in the file AUTHORS
```

PR #2747, opened 2018-12-12
 [LRA Proof] Storage for LRA proofs 

alex-ozdemir:
During LRA solving the `ConstraintDatabase` contains the reasoning
behind different constraints. Combinations of constraints are
periodically used to justify lemmas (conflict clauses, propegations, ...
?). `ConstraintDatabase` is SAT context-dependent.

ArithProofRecorder will be used to store concise representations of the
proof for each lemma raised by the (LR)A theory. The (LR)A theory will
write to it, and the ArithProof class will read from it to produce LFSC
proofs.

Right now, it's pretty simplistic -- it allows for only Farkas proofs.

In future PRs I'll:
   1. add logic that stores proofs therein
   2. add logic that retrieves and prints proofs
   3. enable LRA proof production, checking, and testing

### DIFF
--- a/src/proof/arith_proof_recorder.cpp
+++ b/src/proof/arith_proof_recorder.cpp
@@ -29,13 +29,21 @@ void ArithProofRecorder::saveFarkasCoefficients(
 {
   Assert(conflict.getKind() == kind::AND);
   Assert(conflict.getNumChildren() == farkasCoefficients->size());
+    for (size_t i = 0; i < conflict.getNumChildren(); ++i)
+    {
+      const Node& child = conflict[i];
+      Assert(child.getType().isBoolean() && child[0].getType().isReal());
+    }
   Debug("pf::arith") << "Saved Farkas Coefficients:" << std::endl;
-  for (size_t i = 0; i < conflict.getNumChildren(); ++i)
+  if (Debug.isOn("pf::arith"))
   {
-    const Node& child = conflict[i];
-    const Rational& r = (*farkasCoefficients)[i];
-    Debug("pf::arith") << "  " << std::setw(8) << r;
-    Debug("pf::arith") << "  " << child << std::endl;
+    for (size_t i = 0; i < conflict.getNumChildren(); ++i)
+    {
+      const Node& child = conflict[i];
+      const Rational& r = (*farkasCoefficients)[i];
+      Debug("pf::arith") << "  " << std::setw(8) << r;
+      Debug("pf::arith") << "  " << child << std::endl;
+    }
   }
 
   std::set<Node> lits;
--- a/src/proof/arith_proof_recorder.cpp
+++ b/src/proof/arith_proof_recorder.cpp
@@ -29,11 +29,11 @@ void ArithProofRecorder::saveFarkasCoefficients(
 {
   Assert(conflict.getKind() == kind::AND);
   Assert(conflict.getNumChildren() == farkasCoefficients->size());
-    for (size_t i = 0; i < conflict.getNumChildren(); ++i)
-    {
-      const Node& child = conflict[i];
-      Assert(child.getType().isBoolean() && child[0].getType().isReal());
-    }
+  for (size_t i = 0; i < conflict.getNumChildren(); ++i)
+  {
+    const Node& child = conflict[i];
+    Assert(child.getType().isBoolean() && child[0].getType().isReal());
+  }
   Debug("pf::arith") << "Saved Farkas Coefficients:" << std::endl;
   if (Debug.isOn("pf::arith"))
   {
--- a/src/proof/arith_proof_recorder.cpp
+++ b/src/proof/arith_proof_recorder.cpp
@@ -9,7 +9,7 @@
  ** All rights reserved.  See the file COPYING in the top-level source
  ** directory for licensing information.\endverbatim
  **
- ** \brief A class for saving the skeletons of a arithmetic proofs for later.
+ ** \brief A class for saving the skeletons of arithmetic proofs for later.
  **/
 
 #include "proof/arith_proof_recorder.h"
--- a/src/proof/arith_proof_recorder.cpp
+++ b/src/proof/arith_proof_recorder.cpp
@@ -1,0 +1,71 @@
+/*********************                                                        */
+/*! \file arith_proof_recorder.cpp
+ ** \verbatim
+ ** Top contributors (to current version):
+ **   Alex Ozdemir
+ ** This file is part of the CVC4 project.
+ ** Copyright (c) 2009-2018 by the authors listed in the file AUTHORS
+ ** in the top-level source directory) and their institutional affiliations.
+ ** All rights reserved.  See the file COPYING in the top-level source
+ ** directory for licensing information.\endverbatim
+ **
+ ** \brief A class for saving the skeletons of a arithmetic proofs for later.
+ **/
+
+#include "proof/arith_proof_recorder.h"
+
+#include <algorithm>
+#include <vector>
+
+namespace CVC4 {
+namespace proof {
+
+ArithProofRecorder::ArithProofRecorder() : d_lemmasToFarkasCoefficients()
+{
+  // Nothing else
+}
+void ArithProofRecorder::saveFarkasCoefficients(
+    Node conflict, theory::arith::RationalVectorCP farkasCoefficients)
+{
+  Assert(conflict.getKind() == kind::AND);
+  Assert(conflict.getNumChildren() == farkasCoefficients->size());
+  Debug("pf::arith") << "Saved Farkas Coefficients:" << std::endl;
+  for (size_t i = 0; i < conflict.getNumChildren(); ++i)
+  {
+    const Node& child = conflict[i];
+    const Rational& r = (*farkasCoefficients)[i];
+    Debug("pf::arith") << "  " << std::setw(8) << r;
+    Debug("pf::arith") << "  " << child << std::endl;
+  }
+
+  std::set<Node> lits;
+  std::copy(
+      conflict.begin(), conflict.end(), std::inserter(lits, lits.begin()));
+
+  d_lemmasToFarkasCoefficients[lits] =
+      std::make_pair(std::move(conflict), *farkasCoefficients);
+}
+
+bool ArithProofRecorder::hasFarkasCoefficients(
+    const std::set<Node>& conflict) const
+{
+  return d_lemmasToFarkasCoefficients.find(conflict)
+         != d_lemmasToFarkasCoefficients.end();
+}
+
+std::pair<Node, theory::arith::RationalVectorCP>
+ArithProofRecorder::getFarkasCoefficients(const std::set<Node>& conflict) const
+{
+  if (hasFarkasCoefficients(conflict))
+  {
+    return std::make_pair(d_lemmasToFarkasCoefficients.at(conflict).first,
+                          &d_lemmasToFarkasCoefficients.at(conflict).second);
+  }
+  else
+  {
+    return std::make_pair(Node(), theory::arith::RationalVectorCPSentinel);
+  }
+}
+
+}  // namespace proof
+}  // namespace CVC4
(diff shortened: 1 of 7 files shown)


alex-ozdemir:
I've added @yoni206 because he wanted to see this, and @mpreiner because Yoni though there should be a C++ guru on the review list. I'll wait for both of your approvals.

src/proof/arith_proof_recorder.cpp
```diff
@@ -0,0 +1,71 @@
...
+ ** \verbatim
+ ** Top contributors (to current version):
+ **   Alex Ozdemir
+ ** This file is part of the CVC4 project.
+ ** Copyright (c) 2009-2018 by the authors listed in the file AUTHORS
+ ** in the top-level source directory) and their institutional affiliations.
+ ** All rights reserved.  See the file COPYING in the top-level source
+ ** directory for licensing information.\endverbatim
+ **
+ ** \brief A class for saving the skeletons of a arithmetic proofs for later.
```
yoni206:
```suggestion
 ** \brief A class for saving the skeletons of arithmetic proofs for later.
```

src/proof/arith_proof_recorder.h
```diff
@@ -0,0 +1,101 @@
...
+ **      and a parallel ordering of farkas coefficients.
+ **
+ ** Farkas proofs have the following cocre structure:
+ **    For a list of affine bounds: c[i] dot x >= b[i]
+ **    and a list of non-negative coefficients: f[i],
+ **    compute
+ **
+ **             sum_i{ (c[i] dot x) * f[i] }     and   sum_i{b[i]*f[i]}
+ **
+ **    and then verify that the left is actually < the right, a (=><=)
```
yoni206: 1. Is `x` known at this stage?
2. What is `(=><=)`?
3. Where is this check done in the code?

src/proof/arith_proof_recorder.cpp
```diff
@@ -0,0 +1,71 @@
...
+
+ArithProofRecorder::ArithProofRecorder() : d_lemmasToFarkasCoefficients()
+{
+  // Nothing else
+}
+void ArithProofRecorder::saveFarkasCoefficients(
+    Node conflict, theory::arith::RationalVectorCP farkasCoefficients)
+{
+  Assert(conflict.getKind() == kind::AND);
+  Assert(conflict.getNumChildren() == farkasCoefficients->size());
```
yoni206: maybe it would be useful to add an assertion that all children are literals?

src/proof/arith_proof_recorder.cpp
```diff
@@ -0,0 +1,71 @@
...
+  Debug("pf::arith") << "Saved Farkas Coefficients:" << std::endl;
+  for (size_t i = 0; i < conflict.getNumChildren(); ++i)
+  {
+    const Node& child = conflict[i];
```
yoni206: Can you move the declaration of `child` and `r` before the loop and just do the assignments in the loop?

src/proof/arith_proof_recorder.cpp
```diff
@@ -0,0 +1,71 @@
...
+{
+  // Nothing else
+}
+void ArithProofRecorder::saveFarkasCoefficients(
+    Node conflict, theory::arith::RationalVectorCP farkasCoefficients)
+{
+  Assert(conflict.getKind() == kind::AND);
+  Assert(conflict.getNumChildren() == farkasCoefficients->size());
+  Debug("pf::arith") << "Saved Farkas Coefficients:" << std::endl;
+  for (size_t i = 0; i < conflict.getNumChildren(); ++i)
```
yoni206: Is there a way to prevent this loop from running unless in debug mode?

src/proof/arith_proof_recorder.h
alex-ozdemir: 1. I'll add a definition -- it's a vector of variables.
2. It represents contradiction. I'll replace it with the word.
3. This code doesn't check proofs, it just stores them. I'll add that to the documentation.

src/proof/arith_proof_recorder.cpp
alex-ozdemir: I'm not sure exactly how to do that. I'll add an assertion that each child is
   1. a boolean with
   2. a real child
but let me know if there's a better way.

src/proof/arith_proof_recorder.cpp
alex-ozdemir: I guess it's a stylistic question.

Since `child` and `r` only have meaning inside the loop, declaring them inside the loop aligns their lexical scopes with their regions of semantic validity.

What would be the advantage of declaring them outside the loop?

src/proof/arith_proof_recorder.cpp
alex-ozdemir: Yes, I'll do that.

alex-ozdemir:
Thanks for the feedback!

I believe that I've responded to or addressed each comment. Let me know if that's not the case.

src/proof/arith_proof_recorder.cpp
yoni206: Won't performance on debug suffer from re-creating these two variables over and over again? Or are we certain this will be solved by the compiler?

yoni206:
Thank you for the fixes! Just left one thing comment unresolved, regarding the right place to put the declarations in. What do you think?

src/proof/arith_proof_recorder.cpp
alex-ozdemir: Yeah, the compiler should lift declarations if doing so is a performance win.

yoni206: Doesn't [this](https://github.com/CVC4/CVC4/wiki/Code-Style-Guidelines#local-variables) guideline apply here?

alex-ozdemir: Hmm, well these variables aren't objects.

alex-ozdemir: Yeah, actually, I can't lift them out of the loop anyway, because you can't re-assign references.

src/proof/arith_proof_recorder.h
```diff
@@ -0,0 +1,106 @@
...
+   *
+   * @return whether or not there is actually a proof for them.
+   */
+  bool hasFarkasCoefficients(const std::set<Node>& conflict) const;
+
+  /**
+   * @brief Get the Farkas Coefficients object
+   *
+   * @param conflict a bunch of literals which if conjoined, form a conflict
+   * @return theory::arith::RationalVectorCP -- the farkas coefficients
```
mpreiner: Nit: Farkas

src/proof/arith_proof_recorder.h
```diff
@@ -0,0 +1,106 @@
...
+   *         Node -- a conjunction of the problem literals in coefficient order
+   *
+   *         theory::arith::RationalVectorCPSentinel if there is no entry for
+   * these lits
+   */
+  std::pair<Node, theory::arith::RationalVectorCP> getFarkasCoefficients(
+      const std::set<Node>& conflict) const;
+
+ protected:
+  // For each lemma, save the farkas coefficients of that lemma
```
mpreiner: Nit: Farkas

src/proof/arith_proof_recorder.cpp
```diff
@@ -0,0 +1,79 @@
...
+ ** \verbatim
+ ** Top contributors (to current version):
+ **   Alex Ozdemir
+ ** This file is part of the CVC4 project.
+ ** Copyright (c) 2009-2018 by the authors listed in the file AUTHORS
+ ** in the top-level source directory) and their institutional affiliations.
+ ** All rights reserved.  See the file COPYING in the top-level source
+ ** directory for licensing information.\endverbatim
+ **
+ ** \brief A class for saving the skeletons of arithmetic proofs for later.
```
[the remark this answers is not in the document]
alex-ozdemir: Ah yes, they're saved for use during proof-production. I'll modify the documentation.

src/proof/arith_proof_recorder.cpp
```diff
@@ -0,0 +1,79 @@
...
+    const std::set<Node>& conflict) const
+{
+  return d_lemmasToFarkasCoefficients.find(conflict)
+         != d_lemmasToFarkasCoefficients.end();
+}
+
+std::pair<Node, theory::arith::RationalVectorCP>
+ArithProofRecorder::getFarkasCoefficients(const std::set<Node>& conflict) const
+{
+  if (hasFarkasCoefficients(conflict))
```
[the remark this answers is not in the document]
alex-ozdemir: Sounds good.

src/proof/arith_proof_recorder.h
```diff
@@ -0,0 +1,106 @@
...
+ ** \verbatim
+ ** Top contributors (to current version):
+ **   Alex Ozdemir
+ ** This file is part of the CVC4 project.
+ ** Copyright (c) 2009-2018 by the authors listed in the file AUTHORS
+ ** in the top-level source directory) and their institutional affiliations.
+ ** All rights reserved.  See the file COPYING in the top-level source
+ ** directory for licensing information.\endverbatim
+ **
+ ** \brief A class for saving the skeletons of a arithmetic proofs for later.
```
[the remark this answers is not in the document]
alex-ozdemir: Corrected, as above

src/proof/arith_proof_recorder.h
```diff
@@ -0,0 +1,106 @@
...
+namespace CVC4 {
+namespace proof {
+
+class ArithProofRecorder
+{
+ public:
+  ArithProofRecorder();
+
+  /**
+   * @brief For a set of incompatible literal, save the Farkas coefficients
```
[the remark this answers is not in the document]
alex-ozdemir: Good catch!

src/proof/arith_proof_recorder.h
```diff
@@ -0,0 +1,106 @@
...
+   * demonstrating their incompatibilty
```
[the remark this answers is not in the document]
alex-ozdemir: Good catch!

src/proof/arith_proof_recorder.h
```diff
@@ -0,0 +1,106 @@
...
+   *
+   * @param conflict a bunch of conjoined literals which form a conflict
```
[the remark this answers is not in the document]
alex-ozdemir: I like the first!

src/proof/arith_proof_recorder.h
```diff
@@ -0,0 +1,106 @@
...
+   * @param farkasCoefficients a list of rational coefficients to multiple each
+   *        literal by
```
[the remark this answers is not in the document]
alex-ozdemir: Yeah, that syntax is a bit funky. I'll change it.

src/proof/arith_proof_recorder.h
```diff
@@ -0,0 +1,106 @@
...
+   *
+   * The orders of the two vectors must agree!
+   */
+  void saveFarkasCoefficients(
+      Node conflict, theory::arith::RationalVectorCP farkasCoefficients);
+
+  /**
+   * @brief Determine whether some literals have a farkas proof of their
```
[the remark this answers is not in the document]
alex-ozdemir: :+1:

src/proof/arith_proof_recorder.h
```diff
@@ -0,0 +1,106 @@
...
+   * @param conflict the literals which if conjoined, putatively imply bottom
+   *
+   * @return whether or not there is actually a proof for them.
+   */
+  bool hasFarkasCoefficients(const std::set<Node>& conflict) const;
+
+  /**
+   * @brief Get the Farkas Coefficients object
+   *
+   * @param conflict a bunch of literals which if conjoined, form a conflict
```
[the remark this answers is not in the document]
alex-ozdemir: I'll use the above  language here too.

src/proof/arith_proof_recorder.h
alex-ozdemir: :+1:

src/proof/arith_proof_recorder.h
alex-ozdemir: :+1:

alex-ozdemir:
Thanks for the feedback, @mpreiner!. I believe that I've addressed each of your comments. Take a look and let me know if I missed any, or if you think that further work should be done for any.

PR #2755, opened 2018-12-14
New C++ API: Add tests for term object.

aniemetz:
This further renames Expr::iffExpr() to Expr::eqExpr() to reflect its actual behavior. Accordingly, Term::iffTerm() is renamed to Term::eqTerm().

### DIFF
--- a/test/unit/api/term_black.h
+++ b/test/unit/api/term_black.h
@@ -49,17 +49,11 @@ void TermBlack::testEq()
   Term y = d_solver.mkVar("y", uSort);
   Term z;
 
-  TS_ASSERT_THROWS_NOTHING(x == x);
   TS_ASSERT(x == x);
-  TS_ASSERT_THROWS_NOTHING(x == y);
-  TS_ASSERT(!(x == y));
-  TS_ASSERT_THROWS_NOTHING(x != x);
   TS_ASSERT(!(x != x));
-  TS_ASSERT_THROWS_NOTHING(x != y);
+  TS_ASSERT(!(x == y));
   TS_ASSERT(x != y);
-  TS_ASSERT_THROWS_NOTHING(x == z);
   TS_ASSERT(!(x == z));
-  TS_ASSERT_THROWS_NOTHING(x != z);
   TS_ASSERT(x != z);
 }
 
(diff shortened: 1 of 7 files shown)


test/unit/api/term_black.h
```diff
@@ -24,14 +24,520 @@ class TermBlack : public CxxTest::TestSuite
...
+
+void TermBlack::testEq()
+{
+  Sort uSort = d_solver.mkUninterpretedSort("u");
+  Term x = d_solver.mkVar("x", uSort);
+  Term y = d_solver.mkVar("y", uSort);
+  Term z;
+
+  TS_ASSERT_THROWS_NOTHING(x == x);
+  TS_ASSERT(x == x);
```
4tXJ7f: Suggestion: use `TS_ASSERT_EQUALS(x, x)` instead (it has some nicer output if something goes wrong).

test/unit/api/term_black.h
```diff
@@ -24,14 +24,520 @@ class TermBlack : public CxxTest::TestSuite
...
 };
+
+void TermBlack::testEq()
+{
+  Sort uSort = d_solver.mkUninterpretedSort("u");
+  Term x = d_solver.mkVar("x", uSort);
+  Term y = d_solver.mkVar("y", uSort);
+  Term z;
+
+  TS_ASSERT_THROWS_NOTHING(x == x);
```
4tXJ7f: I think if the equality throws something, then the `TS_ASSERT` below also doesn't go through, so maybe we don't need two separate checks.

test/unit/api/term_black.h
```diff
@@ -24,14 +24,520 @@ class TermBlack : public CxxTest::TestSuite
...
+{
+  Sort bvSort = d_solver.mkBitVectorSort(8);
+  Sort intSort = d_solver.getIntegerSort();
+  Sort boolSort = d_solver.getBooleanSort();
+  Sort funSort1 = d_solver.mkFunctionSort(bvSort, intSort);
+  Sort funSort2 = d_solver.mkFunctionSort(intSort, boolSort);
+
+  Term b = d_solver.mkTrue();
+  Term x = d_solver.mkVar("x", d_solver.mkBitVectorSort(8));
+  TS_ASSERT_THROWS(x.andTerm(b), CVC4ApiException&);
```
4tXJ7f: `b.andTerm(b)` ist actually missing in those (I don't think it is a big deal since the tests are already pretty comprehensive but just mentioning it for completeness).

test/unit/api/term_black.h
aniemetz: The line below was rather to check if it produces the correct result. I can remove if you think this is not necessary, but it certainly doesn't hurt.

test/unit/api/term_black.h
aniemetz: We decided offline to not use `TS_ASSERT_EQUALS` and `TS_ASSERT_DIFFERS` in this case since they can't be used for `!(x != x)` and the like. So for consistency, we use `TS_ASSERT` here.

PR #2762, opened 2018-12-15
Fix issues with REWRITE_DONE in floating point rewriter

ajreynol:
This was triggered by sygus + FP with fp.min/fp.max in the grammar.


src/theory/fp/theory_fp_rewriter.cpp
```diff
@@ -175,8 +175,9 @@ namespace rewrite {
     if (node[0] == node[1]) {
       return RewriteResponse(REWRITE_DONE, NodeManager::currentNM()->mkConst(true));
     } else if (!isPreRewrite && (node[0] > node[1])) {
-	Node normal = NodeManager::currentNM()->mkNode(kind::EQUAL,node[1],node[0]);
-	return RewriteResponse(REWRITE_DONE, normal);
+      Node normal =
+          NodeManager::currentNM()->mkNode(kind::EQUAL, node[1], node[0]);
+      return RewriteResponse(REWRITE_AGAIN_FULL, normal);
```
4tXJ7f: This change should not be needed (there is no other rewrite that could apply and `node[0]`/`node[1]` are already rewritten at that point.

src/theory/fp/theory_fp_rewriter.cpp
```diff
@@ -191,7 +192,7 @@ namespace rewrite {
 	   (k == kind::FLOATINGPOINT_MIN_TOTAL) || (k == kind::FLOATINGPOINT_MAX_TOTAL));
 #endif
     if (node[0] == node[1]) {
-      return RewriteResponse(REWRITE_DONE, node[0]);
+      return RewriteResponse(REWRITE_AGAIN_FULL, node[0]);
```
4tXJ7f: `REWRITE_AGAIN` should be sufficient here (to make sure that we rewrite as much as possible on the prerewriting pass).

src/theory/fp/theory_fp_rewriter.cpp
```diff
@@ -204,7 +205,7 @@ namespace rewrite {
 
     if (node[0] > node[1]) {
       Node normal = NodeManager::currentNM()->mkNode(kind::FLOATINGPOINT_EQ,node[1],node[0]);
-      return RewriteResponse(REWRITE_DONE, normal);
+      return RewriteResponse(REWRITE_AGAIN_FULL, normal);
```
4tXJ7f: This change should not be needed (there is no other rewrite that could apply and `node[0]`/`node[1]` are already rewritten (or just about to be rewritten) at that point.

src/theory/fp/theory_fp_rewriter.cpp
```diff
@@ -217,7 +218,7 @@ namespace rewrite {
 
     if (node[1] > node[2]) {
       Node normal = NodeManager::currentNM()->mkNode(k,node[0],node[2],node[1]);
-      return RewriteResponse(REWRITE_DONE, normal);
+      return RewriteResponse(REWRITE_AGAIN_FULL, normal);
```
4tXJ7f: Ditto.

src/theory/fp/theory_fp_rewriter.cpp
```diff
@@ -229,7 +230,7 @@ namespace rewrite {
 
     if (node[1] > node[2]) {
       Node normal = NodeManager::currentNM()->mkNode(kind::FLOATINGPOINT_FMA,node[0],node[2],node[1],node[3]);
-      return RewriteResponse(REWRITE_DONE, normal);
+      return RewriteResponse(REWRITE_AGAIN_FULL, normal);
```
4tXJ7f: Ditto.

src/theory/fp/theory_fp_rewriter.cpp
```diff
@@ -280,7 +281,11 @@ namespace rewrite {
 			   nm->mkNode(kind::FLOATINGPOINT_REM, working[0][0], working[1]));
     }
 
-    return RewriteResponse(REWRITE_DONE, working);
+    if (working != node)
+    {
+      return RewriteResponse(REWRITE_AGAIN_FULL, working);
```
4tXJ7f: Technically, not all rewrites need `REWRITE_AGAIN_FULL`, only the last one.

src/theory/fp/theory_fp_rewriter.cpp
ajreynol: Yes, you're right.

src/theory/fp/theory_fp_rewriter.cpp
ajreynol: Right.

PR #2767, opened 2018-12-28
[DRAT] DRAT data structure 

alex-ozdemir:
Since we're going to (in some situations) print and check DRAT proofs, we need a data structure for storing them.

This pull request creates such a data structure along with:
* a way of parsing DRAT from its binary format (since cryptominisat gives us binary DRAT proofs)
* a way of printing DRAT proofs textually (presumably this will useful for debugging).

There are no tests yet, since this code is (presently) all dead.

The next step will be to write code for printing DRAT in the format dictated by our LFSC drat signature.

I've added Yoni for DRAT review, and Mathias for C++ review.


src/proof/drat/drat_proof.h
```diff
@@ -0,0 +1,120 @@
...
+  DratProof(DratProof&&) = default;
+
+  ~DratProof() = default;
+
+  /**
+   * Parses a DRAT proof from the binary format.
+   * The format is described at:
+   *    https://www.cs.utexas.edu/~marijn/drat-trim/#contact
+   *
+   * @param binaryProof a string containing the bytes of the "binary" proof
```
yoni206: could you clarify what does "binary" mean?

src/proof/drat/drat_proof.cpp
```diff
@@ -0,0 +1,235 @@
...
+ **/
+
+#include "proof/drat/drat_proof.h"
+#include <bitset>
+
+namespace CVC4 {
+namespace proof {
+namespace drat {
+
+// Forward-declare helper functions for parsing the binary DRAT format.
```
yoni206: Could you please [justify](https://github.com/CVC4/CVC4/wiki/Code-Style-Guidelines#forward-declarations) this forward declaration.

src/proof/drat/drat_proof.cpp
```diff
@@ -0,0 +1,235 @@
...
+
+// DratInstruction implementation
+
+DratInstruction::DratInstruction(DratInstructionKind kind, SatClause clause)
+    : d_kind(kind), d_clause(clause)
+{
+  // All intialized
+}
+
+void DratInstruction::outputAsText(std::ostream& os) const
```
yoni206: `os` is not used in the function.

src/proof/drat/drat_proof.cpp
```diff
@@ -0,0 +1,235 @@
...
+{
+  switch (d_kind)
+  {
+    case DratInstructionKind::addition:
+    {
+      for (const SatLiteral& l : d_clause)
+      {
+        if (l.isNegated())
```
yoni206: Maybe this `if` statement should be a function (it is repeated in the next case as well)?

src/proof/drat/drat_proof.cpp
```diff
@@ -0,0 +1,235 @@
...
+        {
+          Debug("pf::drat") << '-' << l.getSatVariable() + 1 << ' ';
```
yoni206: Why `+1`?

src/proof/drat/drat_proof.h
alex-ozdemir: Yeah, this documentation could be improved.

I'll add a note explaining that binary DRAT proofs should be understood as sequences of bytes, not sequences of characters, and why it's okay to use a `std::string` to represent them, even though most people think of strings as being used to store character sequences.

src/proof/drat/drat_proof.cpp
alex-ozdemir: Good catch! I don't have a good justification. I'll lift the implementations to this point in the source.

src/proof/drat/drat_proof.cpp
alex-ozdemir: Oh man! Thanks for catching this -- that's embarrassing.

I've changed the `Debug` streams within to `os`.

I've also checked all the other `Debug` uses in this file, to verify that I didn't make the mistake elsewhere.

src/proof/drat/drat_proof.cpp
alex-ozdemir: Sounds good. I'll extract the "write DIMACS literal" idea as a function.

src/proof/drat/drat_proof.cpp
alex-ozdemir: Ah yeah. The reason for that is that our internal SAT variables are numbered starting at 0 (because this makes representing them as a uint, boolean pair easy), but externally (e.g. when communicating with a SAT solver) they're numbered starting at 1 (because this is how the DIMACS format works).

I'll add a comment to this effect.

alex-ozdemir:
Thanks for your review, Yoni. I think that I've addressed all your comments. If I missed any, or if you think further work should be done, then let me know!

src/proof/drat/drat_proof.h
```diff
@@ -77,11 +77,29 @@ class DratProof
...
+   * They just mean that proofs in this format should be understood as
+   * sequences of bytes, not sequences of ASCII/Unicode/your favorite character
+   * set characters.
+   *
+   * @param binaryProof a string containing the bytes of the binary proof.
+   *        Even though the proof isn't text, its safe to store it in a string
+   *        because C++ strings don't make any gaurantees about the encoding of
+   *        their contents. This makes them (effectively) just byte sequences.
+   *
+   *        A brief aside: In languages where strings are semantically
```
yoni206: I would keep the side note out of the source files (but maybe leave it in the discussion of this PR).

alex-ozdemir: Sounds good. I'll put the discussion in a comment.

alex-ozdemir:
Question: Why use a `std::string` to represent a byte sequence?

Answer:

>A brief aside: In languages where strings are semantically
character sequences with gaurantees about the encoding (e.g. Rust),
we wouldn't be able to use a string here. Instead we'd have to use
a byte buffer of some sort. However, in C++ a lot of useful
iostream machinery is built around strings, making `std::string` a
more convenient byte buffer than `std::vector<uint_8t>`. Between
this convenience advantage and the lack of encoding constraints on
`std::string`, `std::string` is the best choice for a byte buffer
in C++.

alex-ozdemir:
Sounds good, Yoni. I'll wait for Mathias.

src/proof/drat/drat_proof.h
```diff
@@ -0,0 +1,128 @@
...
+  InvalidDratProofException(const std::string& msg) : Exception(msg) {}
+
+  InvalidDratProofException(const char* msg) : Exception(msg) {}
+}; /* class InvalidDratProofException */
+
+enum DratInstructionKind
+{
+  addition,
+  deletion
+};
```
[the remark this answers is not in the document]
alex-ozdemir: Whoops! Good catch!
I'll fix this.

src/proof/drat/drat_proof.cpp
```diff
@@ -0,0 +1,243 @@
...
+ *
+ * Since internally +0 and -0 are valid literals, we add one to each
+ * literal's number (SAT variable) when outputtting it.
+ *
+ * @param os the stream to write to
+ * @param l the literal to write
+ */
+void outputLiteralAsDimacs(std::ostream& os, SatLiteral l)
+{
+  if (l.isNegated())
```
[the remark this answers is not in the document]
alex-ozdemir: Sure, sounds good.

src/proof/drat/drat_proof.h
```diff
@@ -0,0 +1,128 @@
...
+   * The format is described at:
+   *    https://www.cs.utexas.edu/~marijn/drat-trim/#contact
+   *
+   * What do the standard authors mean by the format being "binary"?
+   * They just mean that proofs in this format should be understood as
+   * sequences of bytes, not sequences of ASCII/Unicode/your favorite character
+   * set characters.
+   *
+   * @param binaryProof a string containing the bytes of the binary proof.
+   *        Even though the proof isn't text, its safe to store it in a string
```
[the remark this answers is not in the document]
alex-ozdemir: Good catch! I'll fix this.

src/proof/drat/drat_proof.cpp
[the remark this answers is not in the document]
alex-ozdemir: Yeah, that is a lot nicer. Good spot.

src/proof/drat/drat_proof.cpp
```diff
@@ -0,0 +1,243 @@
...
+{
+  // lit is encoded as uint represented by a variable-length byte sequence
+  uint64_t literal_represented_as_uint = 0;
+  for (int shift = 0; start != proof_end; ++start, shift += 7)
+  {
+    unsigned char byte = *start;
+    // The MSB of the byte is an indicator of whether the sequence continues
+    bool continued = (byte >> 7) & 1;
+    unsigned char numeric_part = byte & 0x7f;
+    literal_represented_as_uint |= numeric_part << shift;
```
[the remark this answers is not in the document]
alex-ozdemir: Good catch!

Agreed, I'll raise an error if the literal is too big.

I think you're correct regarding the conversion to `unsigned char`, per 
> Otherwise, both operands are converted to the unsigned counterpart of the signed operand's type. 

And yes, that could result in undefined behavior. I'll make `numeric_part` a `uint_64t` and make `shift` a `uint_8t`.

src/proof/drat/drat_proof.cpp
```diff
@@ -35,12 +35,20 @@ SatLiteral parse_binary_literal(std::string::const_iterator& start,
 {
   // lit is encoded as uint represented by a variable-length byte sequence
   uint64_t literal_represented_as_uint = 0;
-  for (int shift = 0; start != proof_end; ++start, shift += 7)
+  for (uint8_t shift = 0; start != proof_end; ++start, shift += 7)
   {
+    // Check whether shift is so large that we're going to lose some
+    // information
+    if (shift + 7 > 64)
```
mpreiner: This must be actually `if (shift + 7 >= 64)` as it is undefined behavior if it is equal or greater than the bit width (or if it is negative).

mpreiner:
I know you mentioned that there are no tests yet, but actually I would prefer that you add some unit tests to test most of the functionality (since **it is** dead code ;-)).

src/proof/drat/drat_proof.cpp
alex-ozdemir: Good catch!. I'll fix it.

alex-ozdemir:
Okay, I added a few unit tests for the DRAT proof class.

test/unit/proof/CMakeLists.txt
```diff
@@ -0,0 +1,4 @@
+#-----------------------------------------------------------------------------#
+# Add unit tests
+
+cvc4_add_unit_test_white(drat_proof_black proof)
```
mpreiner: `cvc4_add_unit_test_black`

PR #2768, opened 2019-01-03
RNG: Avoid modulo bias in random numbers

4tXJ7f:
Computing `rand() % range` is problematic because it is biased if the
range does not evenly divide the largest possible random value [0].
This commit uses `std::uniform_int_distribution` to generate a random
value within a given range, avoiding the problem. Similarly, generating
floating-point values in an unbiased matter is tricky, and the commit
uses `std::uniform_real_distribution` to avoid those issues. Note that
`std::uniform_real_distribution` generates values in the range [a, b).
The commit uses the suggested solution in [1] to generate values in [a,
b].

[0] https://en.wikipedia.org/wiki/Fisher%E2%80%93Yates_shuffle#Modulo_bias
[1] https://en.cppreference.com/w/cpp/numeric/random/uniform_real_distribution#Notes

*Note:* I am adding @ajreynol as a reviewer because the output of one of the quantifier regressions changed.

### DIFF
--- a/src/util/random.cpp
+++ b/src/util/random.cpp
@@ -19,6 +19,8 @@
 #include "util/random.h"
 
 #include <cfloat>
+#include <random>
+
 #include "base/cvc4_assert.h"
 
 namespace CVC4 {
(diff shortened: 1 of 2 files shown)


4tXJ7f:
We have discussed this on Slack and came to the conclusion that we should implement our own version of `std::uniform_int_distribution` to ensure that we get the same random numbers regardless of the STL implementation.

aniemetz:
@4tXJ7f any updates on this? Do we still plan to do this?

PR #2772, opened 2019-01-04
New C++ API: Add missing catch blocks for std::invalid_argument.

aniemetz:


aniemetz:
What about the TypeCheckingException catches? Same?

PR #2773, opened 2019-01-04
New C++ API: Add missing getType() calls to kick off type checking.

aniemetz:


aniemetz:
> Maybe we could also add some simple unit tests for type checking (I can do it if you'd like)?

:tada: :tada: :tada:

4tXJ7f:
I just realized that we actually already have unit tests for type checking for those functions (which makes sense because type checking was already being done in debug mode before).

PR #2776, opened 2019-01-06
LFSC drat output

alex-ozdemir:
Adds a method to the `DratProof` class which prints it in the format specified by our LFSC DRAT signature.

### DIFF
--- a/src/proof/drat/drat_proof.cpp
+++ b/src/proof/drat/drat_proof.cpp
@@ -15,7 +15,10 @@
  **/
 
 #include "proof/drat/drat_proof.h"
+#include <algorithm>
 #include <bitset>
+#include <iostream>
+#include "proof/proof_manager.h"
 
 namespace CVC4 {
 namespace proof {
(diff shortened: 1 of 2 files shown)


src/proof/drat/drat_proof.cpp
```diff
@@ -15,7 +15,10 @@
  **/
 
 #include "proof/drat/drat_proof.h"
```
mpreiner: Add newline

src/proof/drat/drat_proof.cpp
```diff
@@ -15,7 +15,10 @@
...
+#include <algorithm>
 #include <bitset>
+#include <iostream>
```
mpreiner: Ditto

src/proof/drat/drat_proof.cpp
alex-ozdemir: :+1:

src/proof/drat/drat_proof.cpp
alex-ozdemir: :+1:

alex-ozdemir:
Adding a unit test was a good idea, and I have done so.

Let me know if you think anything else should be done, Mathias.

alex-ozdemir:
Yeah, the code is clang-formatted, although clang-format seems to be being a bit weird in some places (e.g. [here](https://github.com/CVC4/CVC4/pull/2776/files#diff-9d7950c07b864c5bcdfb95fd9029a3dcR270))

alex-ozdemir:
Hmm. Github isn't picking up the success of the Travis build. I'm rerunning the build from the Travis interface.

PR #2779, opened 2019-01-08
Clause proof printing

alex-ozdemir:
Adds two features to the LFSC proof printer:
1. A function for printing LFSC clauses
2. A function for printing proofs of cnf formulas (given that the constituent clauses have already been printed)

Also a unit test.

### DIFF
--- a/src/proof/lfsc_proof_printer.h
+++ b/src/proof/lfsc_proof_printer.h
@@ -79,8 +79,8 @@ class LFSCProofPrinter
    * Assuming that each clause has alreay been individually proven,
    * defines a proof of the input to the SAT solver.
    *
-   * Returns the name of the LFSC value holding the proof, i.e. a value of type
-   * (clauses_hold ...)
+   * Prints an LFSC value corresponding to the proof, i.e. a value of type
+   * (cnf_holds ...)
    *
    * @param clauses The clauses to print a proof of
    * @param out The stream to print to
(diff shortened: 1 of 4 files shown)


alex-ozdemir:
I thought this was small enough/had a simple enough interaction with the codebase, that you could be the only reviewer, but if you'd like to add someone else, feel free!

src/proof/lfsc_proof_printer.h
```diff
@@ -74,7 +74,35 @@ class LFSCProofPrinter
                                          std::ostream& out,
                                          std::ostream& paren);
 
+  /**
+   * The SAT solver is given a list of clauses.
+   * Assuming that each clause has alreay been individually proven,
+   * defines a proof of the input to the SAT solver.
+   *
+   * Returns the name of the LFSC value holding the proof, i.e. a value of type
```
yoni206: returns or prints?

src/proof/lfsc_proof_printer.h
```diff
@@ -74,7 +74,35 @@ class LFSCProofPrinter
...
+   * (clauses_hold ...)
```
yoni206: where is this type defined?

src/proof/lfsc_proof_printer.h
```diff
@@ -74,7 +74,35 @@ class LFSCProofPrinter
...
+   *
+   * @param clauses The clauses to print a proof of
```
yoni206: Any assumption on the order of the clauses in this vector?

src/proof/lfsc_proof_printer.h
alex-ozdemir: Good catch! Yes, I mean prints.

src/proof/lfsc_proof_printer.h
alex-ozdemir: This is a _very_ good catch! You're right that no such type is defined.

I don't actually want to be printing a proof that some `cnf` holds, I want to be printing a proof that some `CMap` holds.

I think that at some point I was using the `clauses_hold` type, but I've since removed it.

I would have discovered this once I actually tried to put the pieces together, but all the better to discover it now!

Thanks!

src/proof/lfsc_proof_printer.h
alex-ozdemir: Nope!

The function will work with any order. And in fact, the order won't influence the correctness of the proof, since this is for DRAT, where the working formula is a multiset of clauses (which has no concept of order!)

alex-ozdemir:
Thanks for catching that mistake, Yoni! I've responded to or addressed each of your comments.

I also added a new function, inspired by Yoni's comment, for printing proofs of `CMap`s, for use with LRAT, eventually.

src/proof/lfsc_proof_printer.h
yoni206: Right, mixed it with lrat.

PR #2787, opened 2019-01-12
LFSC LRAT Output

alex-ozdemir:
In this PR I implement LFSC output for LRAT proofs.

I also added a unit test for this output.

The natural reviewers seemed like Yoni (since he knows LFSC) and Andres (since he's reviewed the rest of the `LratProof` class). Since I just dropped a painfully huge diff on Andres, I've assigned Yoni, but Yoni, if you want to call in someone else to be another pair of eyes, feel free.

### DIFF
--- a/src/proof/lrat/lrat_proof.cpp
+++ b/src/proof/lrat/lrat_proof.cpp
@@ -75,6 +75,54 @@ std::ostream& operator<<(std::ostream& o, const LratUPTrace& trace)
   return o;
 }
 
+/**
+ * Print a list of clause indices to go to while doing UP.
+ *
+ * i.e. a value of type Trace
+ */
+void printTrace(std::ostream& o, const LratUPTrace& trace)
+{
+  for (ClauseIdx idx : trace)
+  {
+    o << "(Tracec " << idx << " ";
+  }
+  o << "Tracen";
+  std::fill_n(std::ostream_iterator<char>(o), trace.size(), ')');
+}
+
+/**
+ * Print the RAT hints for a clause addition.
+ *
+ * i.e. prints an LFSC value of type RATHints
+ */
+void printHints(std::ostream& o,
+                const std::vector<std::pair<ClauseIdx, LratUPTrace>>& hints)
+{
+  for (auto& hint : hints)
+  {
+    o << "\n    (RATHintsc " << hint.first << " ";
+    printTrace(o, hint.second);
+    o << " ";
+  }
+  o << "RATHintsn";
+  std::fill_n(std::ostream_iterator<char>(o), hints.size(), ')');
+}
+
+/**
+ * Print an index list
+ *
+ * i.e. prints an LFSC value of type CIList
+ */
+void printIndices(std::ostream& o, const std::vector<ClauseIdx>& indices)
+{
+  for (ClauseIdx idx : indices)
+  {
+    o << "(CIListc " << idx << " ";
+  }
+  o << "CIListn";
+  std::fill_n(std::ostream_iterator<char>(o), indices.size(), ')');
+}
+
 }  // namespace
 
 // Prints the LRAT addition line in textual format
@@ -248,6 +296,17 @@ LratProof::LratProof(std::istream& textualProof)
   }
 }
 
+void LratProof::outputAsLfsc(std::ostream& o) const
+{
+  std::ostringstream closeParen;
+  for (const auto& i : d_instructions)
+  {
+    i->outputAsLfsc(o, closeParen);
+  }
+  o << "LRATProofn";
+  o << closeParen.str();
+}
+
 void LratAddition::outputAsText(std::ostream& o) const
 {
   o << d_idxOfClause << " ";
@@ -261,6 +320,18 @@ void LratAddition::outputAsText(std::ostream& o) const
   o << "0\n";
 }
 
+void LratAddition::outputAsLfsc(std::ostream& o, std::ostream& closeParen) const
+{
+  o << "\n    (LRATProofa " << d_idxOfClause << " ";
+  closeParen << ")";
+  LFSCProofPrinter::printSatClause(d_clause, o, "bb");
+  o << " ";
+  printTrace(o, d_atTrace);
+  o << " ";
+  printHints(o, d_resolvants);
+  o << " ";
+}
+
 void LratDeletion::outputAsText(std::ostream& o) const
 {
   o << d_idxOfClause << " d ";
@@ -269,6 +340,14 @@ void LratDeletion::outputAsText(std::ostream& o) const
     o << idx << " ";
   }
   o << "0\n";
+}
+
+void LratDeletion::outputAsLfsc(std::ostream& o, std::ostream& closeParen) const
+{
+  o << "\n    (LRATProofd ";
+  closeParen << ")";
+  printIndices(o, d_clauses);
+  o << " ";
 }
 
 std::ostream& operator<<(std::ostream& o, const LratProof& p)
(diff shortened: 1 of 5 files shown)


test/unit/proof/lrat_proof_black.h
```diff
@@ -0,0 +1,112 @@
...
+  std::ostringstream lfsc;
+  proof.outputAsLfsc(lfsc);
+  std::ostringstream lfscWithoutWhitespace;
+  for (char c : lfsc.str())
+  {
+    if (!std::isspace(c))
+    {
+      lfscWithoutWhitespace << c;
+    }
+  }
```
4tXJ7f: Suggestion: Create a helper function for this.

test/unit/proof/lrat_proof_black.h
```diff
@@ -0,0 +1,112 @@
...
+ ** \verbatim
+ ** Top contributors (to current version):
+ **   Alex Ozdemir
+ ** This file is part of the CVC4 project.
+ ** Copyright (c) 2009-2018 by the authors listed in the file AUTHORS
+ ** in the top-level source directory) and their institutional affiliations.
+ ** All rights reserved.  See the file COPYING in the top-level source
+ ** directory for licensing information.\endverbatim
+ **
+ ** \brief Black box testing of the DRAT proof class
```
4tXJ7f:
```suggestion
 ** \brief Black box testing of the LRAT proof class
```

test/unit/proof/lrat_proof_black.h
```diff
@@ -0,0 +1,112 @@
...
+ **
+ ** In particular, tests DRAT binary parsing.
```
4tXJ7f:
```suggestion
 ** In particular, tests LRAT LFSC output.
```

test/unit/proof/lrat_proof_black.h
```diff
@@ -0,0 +1,112 @@
+/*********************                                                        */
+/*! \file lrat_proof_black.h
+ ** \verbatim
+ ** Top contributors (to current version):
+ **   Alex Ozdemir
+ ** This file is part of the CVC4 project.
+ ** Copyright (c) 2009-2018 by the authors listed in the file AUTHORS
```
4tXJ7f:
```suggestion
 ** Copyright (c) 2009-2019 by the authors listed in the file AUTHORS
```

src/proof/lrat/lrat_proof.cpp
```diff
@@ -75,6 +75,54 @@ std::ostream& operator<<(std::ostream& o, const LratUPTrace& trace)
   return o;
 }
 
+/**
+ * Print a list of clause indices to go to while doing UP.
+ *
+ * i.e. a value of type Trace
```
4tXJ7f: Add comments for the `@param`s.

test/unit/proof/lrat_proof_black.h
alex-ozdemir: Good idea! This is also an opportunity to use some STL.

alex-ozdemir:
Thanks for the feedback, Andres! I believe that I've addressed all your comments. If that's not the case, or if you think further work should be done to address any of them, let me know.

test/unit/proof/lrat_proof_black.h
4tXJ7f: Nice :)

src/proof/lrat/lrat_proof.cpp
```diff
@@ -75,6 +75,63 @@ std::ostream& operator<<(std::ostream& o, const LratUPTrace& trace)
...
+ * Print an index list
+ *
+ * i.e. prints an LFSC value of type CIList
+ *
+ * @param o where to print to
+ * @param indices the list of indices to print
+ */
+void printIndices(std::ostream& o, const std::vector<ClauseIdx>& indices)
+{
+  for (ClauseIdx idx : indices)
```
yoni206: Perhaps add an assertion that the list is sorted (as assumed by `lrat.plf`)?

src/proof/lrat/lrat_proof.h
```diff
@@ -52,6 +52,15 @@ class LratInstruction
    * @param out the stream to write to
    */
   virtual void outputAsText(std::ostream& out) const = 0;
```
yoni206: What tool can be used to process `out` after the operation of `outputAsText`, if any? Does it correspond to the format of the examples in the LRAT paper?

test/unit/proof/lrat_proof_black.h
```diff
@@ -0,0 +1,113 @@
...
+std::string filterWhitespace(const std::string& s)
+{
+  std::string out;
+  std::copy_if(s.cbegin(), s.cend(), std::inserter(out, out.end()), [](char c) {
+    return !std::isspace(c);
+  });
+  return out;
+}
+
+void LfscProofBlack::testOutputThreeAsLfsc()
```
yoni206: What does `Three` stand for in `testOutputThreeAsLfsc`?

test/unit/proof/lrat_proof_black.h
```diff
@@ -0,0 +1,113 @@
...
+{
+  std::vector<std::unique_ptr<LratInstruction>> instructions;
+
+  // 6   d 1 2
```
yoni206: Is this proof taken from some paper or something? If so, it might be useful to include lines 1--5 (which I assume are the original clauses?).

test/unit/proof/lrat_proof_black.h
```diff
@@ -0,0 +1,113 @@
...
+      "  RATHintsn "
+      "(LRATProofa 8 "
+      "  (clc (pos bb.v2) cln)"
+      "  Tracen "
+      "  (RATHintsc 1 (Tracec 3 Tracen)"
+      "    (RATHintsc 5 (Tracec 2 Tracen)"
+      "    RATHintsn)) "
+      "LRATProofn)))";
+
+  TS_ASSERT_EQUALS(filterWhitespace(lfsc.str()), filterWhitespace(expectedLfsc));
```
yoni206: What do you think about applying `filterWhitespace` on both strings, in case in the future someone adds a space somewhere?

src/proof/lrat/lrat_proof.h
alex-ozdemir: Good question. The answer is the `ClausalBitVectorProof` (in review) and the `ProofManager`.

Since this method just writes an LFSC LRAT proof **value** (not a proof!), the `ClausalBitVectorProof` will write the axiom which uses a side condition to check the proof value and then prove bottom.

The `ProofManager` also needs to write a bunch of closing parentheses after that.

Unfortunately the answer is yes -- the proof manager (or some other component) write a bunch of closing parentheses to the stream.

test/unit/proof/lrat_proof_black.h
alex-ozdemir: Ahh -- three lines. I'll remove that detail. It made sense in a different context, but not here.

test/unit/proof/lrat_proof_black.h
alex-ozdemir: Ah -- I actually just make this proof up.

I didn't think it was important to have a semantically valid proof, since this test is really just about formatting.

test/unit/proof/lrat_proof_black.h
alex-ozdemir: I think that's a great idea!

alex-ozdemir:
Thanks for the review Yoni!

I think that I've addressed or responded to each of your comments. Let me know if I missed any, or if you think any could be better addressed.

PR #2761, opened 2018-12-15
 Strings: More aggressive skolem normalization

4tXJ7f:
This PR makes the skolem normalization in the string solver quite a bit
more aggressive by reducing more skolems to prefix and suffix skolems.
Experiments show that this PR significantly improves performance on CMU
benchmarks.

Marking as "do not merge" because it includes #2759 and because one of the regressions times out (regress2/strings/non_termination_regular_expression6.smt2). We need to decide what to do about that.

PR #2794, opened 2019-01-15
Strings: Add option to change loop process mode

4tXJ7f:
This commit adds an option `--strings-process-loop-mode` that allows
finer-grained control over CVC4 processes looping word equation. In
particular, performing normal loop breaking sometimes leads to worse
performance. The "simple" mode disables that inference.

### DIFF
--- a/src/theory/strings/theory_strings.cpp
+++ b/src/theory/strings/theory_strings.cpp
@@ -3083,15 +3083,29 @@ void TheoryStrings::processSimpleNEq( std::vector< std::vector< Node > > &normal
             Trace("strings-solve-debug") << "Non-simple Case 2 : must compare strings" << std::endl;
             int loop_in_i = -1;
             int loop_in_j = -1;
+            ProcessLoopResult plr = ProcessLoopResult::SKIPPED;
             if( detectLoop( normal_forms, i, j, index, loop_in_i, loop_in_j, rproc ) ){
               if( !isRev ){  //FIXME
               getExplanationVectorForPrefixEq( normal_forms, normal_form_src, normal_forms_exp, normal_forms_exp_depend, i, j, -1, -1, isRev, info.d_ant );
               //set info
-              if( processLoop( normal_forms, normal_form_src, i, j, loop_in_i!=-1 ? i : j, loop_in_i!=-1 ? j : i, loop_in_i!=-1 ? loop_in_i : loop_in_j, index, info ) ){
+              plr = processLoop(normal_forms,
+                                normal_form_src,
+                                i,
+                                j,
+                                loop_in_i != -1 ? i : j,
+                                loop_in_i != -1 ? j : i,
+                                loop_in_i != -1 ? loop_in_i : loop_in_j,
+                                index,
+                                info);
+              if (plr == ProcessLoopResult::INFERENCE)
+              {
                 info_valid = true;
               }
               }
-            }else{
+            }
+
+            if (plr == ProcessLoopResult::SKIPPED)
+            {
               //AJR: length entailment here?
               if( normal_forms[i][index].getKind() == kind::CONST_STRING || normal_forms[j][index].getKind() == kind::CONST_STRING ){
                 unsigned const_k = normal_forms[i][index].getKind() == kind::CONST_STRING ? i : j;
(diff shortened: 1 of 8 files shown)


src/theory/strings/theory_strings.cpp
```diff
@@ -3311,18 +3325,26 @@ bool TheoryStrings::detectLoop( std::vector< std::vector< Node > > &normal_forms
...
+{
+  if (options::stringProcessLoopMode() == ProcessLoopMode::ABORT)
   {
-    d_out->setIncomplete();
-    return false;
+    throw LogicException("Looping word equation encountered.");
+  }
+  else if (options::stringProcessLoopMode() == ProcessLoopMode::NONE)
+  {
+    return ProcessLoopResult::SKIPPED;
```
ajreynol: I think we need to track the setIncomplete() from the other side. This ensures we answer "unknown" instead of "sat" for cases where we encounter but choose not to process a looping word equation.

src/theory/strings/theory_strings.cpp
```diff
@@ -3464,6 +3486,15 @@ bool TheoryStrings::processLoop( std::vector< std::vector< Node > > &normal_form
   }
   else
   {
+    if (options::stringProcessLoopMode() == ProcessLoopMode::SIMPLE_ABORT)
+    {
+      throw LogicException("Normal loop breaking not allowed");
```
ajreynol: I think this should say something like `Normal looping word equation encountered.`

4tXJ7f:
Here are the results (all with a 300s timeout, also including #2761):

```
            DIRECTORY |                      qf_slia_process_loop_full |                qf_slia_process_loop_full_no_fmf |                    qf_slia_process_loop_simple |              qf_slia_process_loop_simple_no_fmf |
            BENCHMARK |  STAT  SLVD   TOT  TO MO ERR  CPU[s] SPACE[MB] |  STAT  SLVD   TOT  TO MO ERR   CPU[s] SPACE[MB] |  STAT  SLVD   TOT  TO MO ERR  CPU[s] SPACE[MB] |  STAT  SLVD   TOT  TO MO ERR   CPU[s] SPACE[MB] |
    20180523-Reynolds |    ok 25336 25421  85  0   0 66848.1  463180.3 |    ok 25223 25421 198  0   0 122033.4  492512.1 |    ok 25339 25421  82  0   0 66825.8  462752.2 |    ok 25223 25421 198  0   0 122145.6  492937.7 |
appscan-pisa-smtlib25 |    ok    20    20   0  0   0   188.3     317.9 |    ok    20    20   0  0   0     55.9     301.6 |    ok    20    20   0  0   0   188.6     318.0 |    ok    20    20   0  0   0     53.4     301.5 |
    str-pred-small-rw |    ok   818   835  17  0   0  5103.4   12137.9 |    ok   829   835   6  0   0   1801.1    1930.5 |    ok   818   835  17  0   0  5103.2   12451.2 |    ok   827   835   8  0   0   2400.9    2579.4 |
    str-term-small-rw |    ok  1037  1045   8  0   0  2465.8    7131.8 |    ok  1035  1045  10  0   0   3009.8    3279.7 |    ok  1037  1045   8  0   0  2461.1    7149.8 |    ok  1035  1045  10  0   0   3009.6    3289.9 |
             __totals |    ok 27211 27321 110  0   0 74605.6  482767.9 |    ok 27107 27321 214  0   0 126900.2  498023.9 |    ok 27214 27321 107  0   0 74578.7  482671.2 |    ok 27105 27321 216  0   0 127609.5  499108.5 |
````

And for the "POPL"  benchmark set:
```
                          DIRECTORY |  TOT |  SLVD |  SAT |  UNSAT |  UNKWN |   TO |  MO |  ERR |   CPU[s] |  SPACE[MB] |
         popl2017_process_loop_full |  600 |   397 |   49 |    348 |      0 |  203 |   0 |    0 |  66482.1 |   131818.3 |
  popl2017_process_loop_full_no_fmf |  600 |   409 |   54 |    355 |      0 |  191 |   0 |    0 |  58953.8 |    56516.8 |
       popl2017_process_loop_simple |  600 |   500 |  151 |    349 |      0 |  100 |   0 |    0 |  35624.5 |    95278.3 |
popl2017_process_loop_simple_no_fmf |  600 |   464 |  108 |    356 |      0 |  136 |   0 |    0 |  42417.4 |    49988.7 |
```

As you can see, `simple` is crushing it on the "POPL" benchmark set (both with and without fmf). On the other  benchmark sets, there is essentially no difference when running with fmf (the benchmarks that were not solved by both configs were all very close to timeout). Without fmf, we are losing two rewriting benchmarks:

```
                                   DIRECTORY | qf_slia_process_loop_full_no_fmf | qf_slia_process_loop_simple_no_fmf |
                                   BENCHMARK |  STAT  RES EXIT CPU[s] SPACE[MB] |    STAT  RES EXIT CPU[s] SPACE[MB] |
str-pred-small-rw-str-pred-small-rw_725.smt2 |    ok   20    0   0.09      15.5 |      to    -    2 300.09     395.8 |
str-pred-small-rw-str-pred-small-rw_732.smt2 |    ok   20    0   0.09      15.3 |      to    -    2 300.02     395.7 |
```

I would be quite comfortable to make `simple` the default for fmf and full the default otherwise. What do you think?

ajreynol:
Yes, those defaults sound correct to me.  Note that it makes sense to have `simple` for `fmf` since fmf does not strictly require normal loop breaking to make progress, since its always working with a fixed bound.

PR #2795, opened 2019-01-15
fix mismatch between "delete" and "new []"

ptroja:
It is an undefined behaviour to call `delete` on a value returned by a call
to operator `new []` (i.e. the array form of allocator). Detected with
cppcheck static analyser.


src/theory/arith/approx_simplex.cpp
```diff
@@ -1487,7 +1487,7 @@ static void glpkCallback(glp_tree *tree, void *info){
         RowsDeleted* rd = new RowsDeleted(exec, nrows, num);
 
         node.addCut(rd);
-        delete num;
+        delete [] num;
```
4tXJ7f:
```suggestion
        delete[] num;
```

4tXJ7f:
Btw. how many other issues does cppcheck detect? We are running Coverity nightly but if it turns out that cppcheck finds a lot of other issues, it might be worth it to look into running it nightly as well.

ptroja:
With cppcheck 1.76.1-1 and its default rules, this was the only issue that looked legitimately. With `--enable=all` (which enables all extra checks) I got ~650 messages, which is far too much for me to investigate, but those with `--enable=performance` also look reasonable, e.g. ```Function parameter 'XXX' should be passed by reference```.

4tXJ7f:
@ptroja Sorry for the delay. We finally settled on a process for accepting external contributions that we outlined [here](https://github.com/CVC4/CVC4/blob/master/CONTRIBUTING.md). Could you please squash your two commits and sign-off the resulting commit?

Thank you for you patience and contribution.

aniemetz:
@ptroja We would like to go forward with merging in this PR. Can you please squash your two commits and sign-off the resulting commit as instructed above? If we don't hear back from you until Friday Aug. 23 we'll go ahead and apply your fixes ourselves.

ptroja:
@aniemetz Sorry for the delay, I believe this commit should be ready to go now.

PR #2800, opened 2019-01-16
Extract DIMACS Printing

alex-ozdemir:
Creating LRAT proofs requires writing SAT problems in the DIMACS format.
Before this code was in the LRAT class.

However, since creating ER proofs will also require writing DIMACS, I
decided to extract it this code into helper functions.

At the same time I realized that my prior representation of used clauses
was unnecessarily poor. I had chosen it to interface with
`CnfProof::collectAtomsForClauses`, but I have since discovered that the
aforementioned method is super simple, so choosing a poor representation to re-use it is not the right call.

### DIFF
--- a/src/proof/dimacs_printer.cpp
+++ b/src/proof/dimacs_printer.cpp
@@ -1,0 +1,77 @@
+/*********************                                                        */
+/*! \file dimacs_printer.cpp
+ ** \verbatim
+ ** Top contributors (to current version):
+ **   Alex Ozdemir
+ ** This file is part of the CVC4 project.
+ ** Copyright (c) 2009-2018 by the authors listed in the file AUTHORS
+ ** in the top-level source directory) and their institutional affiliations.
+ ** All rights reserved.  See the file COPYING in the top-level source
+ ** directory for licensing information.\endverbatim
+ **
+ ** \brief DIMACS SAT Problem Format
+ **
+ ** Defines serialization for SAT problems as DIMACS
+ **/
+
+#include "proof/dimacs_printer.h"
+
+#include <iostream>
+
+namespace CVC4 {
+namespace proof {
+
+// Prints the literal as a (+) or (-) int
+// Not operator<< b/c that represents negation as ~
+std::ostream& textOut(std::ostream& o, const prop::SatLiteral& l)
+{
+  if (l.isNegated())
+  {
+    o << "-";
+  }
+  return o << l.getSatVariable() + 1;
+}
+
+// Prints the clause as a space-separated list of ints
+// Not operator<< b/c that represents negation as ~
+std::ostream& textOut(std::ostream& o, const prop::SatClause& c)
+{
+  for (const auto l : c)
+  {
+    textOut(o, l) << " ";
+  }
+  return o << "0";
+}
+
+void printDimacs(
+    std::ostream& o,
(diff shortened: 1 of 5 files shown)


src/proof/dimacs_printer.h
```diff
@@ -0,0 +1,46 @@
...
+#include <memory>
+
+#include "proof/clause_id.h"
+#include "prop/sat_solver_types.h"
+
+namespace CVC4 {
+namespace proof {
+
+// Prints the literal as a (+) or (-) int
+// Not operator<< b/c that represents negation as ~
```
aniemetz: Use a style consistent with comments in other header files:
```
/* ... */

/**
  ...
  ...
*/
```

src/proof/dimacs_printer.h
```diff
@@ -0,0 +1,46 @@
...
+std::ostream& textOut(std::ostream& o, const prop::SatLiteral& l);
+
+// Prints the clause as a space-separated list of ints
+// Not operator<< b/c that represents negation as ~
```
aniemetz: Ditto

src/proof/dimacs_printer.h
```diff
@@ -0,0 +1,46 @@
...
+std::ostream& textOut(std::ostream& o, const prop::SatClause& c);
+
+void printDimacs(
```
aniemetz: Document.

src/proof/dimacs_printer.cpp
```diff
@@ -0,0 +1,77 @@
...
+    o << "-";
+  }
+  return o << l.getSatVariable() + 1;
+}
+
+// Prints the clause as a space-separated list of ints
+// Not operator<< b/c that represents negation as ~
+std::ostream& textOut(std::ostream& o, const prop::SatClause& c)
+{
+  for (const auto l : c)
```
aniemetz: `for (const auto& ...)`

src/proof/dimacs_printer.cpp
```diff
@@ -0,0 +1,77 @@
...
+}
+
+void printDimacs(
+    std::ostream& o,
+    const std::vector<std::pair<ClauseId, prop::SatClause>>& usedClauses)
+{
+  size_t maxVar = 0;
+  for (auto& c : usedClauses)
+  {
+    for (auto l : c.second)
```
aniemetz: `for (const auto& ...)`

src/proof/dimacs_printer.cpp
alex-ozdemir: Sounds good!

src/proof/dimacs_printer.cpp
alex-ozdemir: Sounds good!

src/proof/dimacs_printer.h
alex-ozdemir: Sounds good!

src/proof/dimacs_printer.h
alex-ozdemir: :+1:

src/proof/dimacs_printer.h
alex-ozdemir: Whoops! Good catch. Will do!

alex-ozdemir:
Thanks for the review, Aina! I believe that I've addressed each of your comments. Let me know if I missed any, or if further work should be done.

PR #2801, opened 2019-01-16
Bugfix: LFSC clause equality

alex-ozdemir:
My implementation of clause equality had an undocumented assumption that
the clauses didn't have any duplicate literals. Now that assumption is
gone, and the tests suite has been expanded.

You can run the (updated) tests by running

    lfscc sat.plf smt.plf lrat.plf drat.plf drat_test.plf

in the `signatures` directory.

### DIFF
--- a/proofs/signatures/drat.plf
+++ b/proofs/signatures/drat.plf
@@ -29,10 +29,20 @@
 ; We could sort them, but that would require defining an order on our variables,
 ; and incurring the cost of sorting.
 ; Instead, we do the following:
-;  1. Sweep the first clause, marking variables with flags 1 (pos) and 2 (neg)
-;  2. Sweep the second clause, erasing marks.
+;  1. Sweep the first clause, marking variables with flags 1,3 (pos) and 2,4 (neg)
+;  2. Sweep the second clause, erasing marks 1,2.
 ;  3. Unsweep the first clause, returning FALSE on marks.
-;     Also unmarking
+;     Also unmarking 3,4, and 1,2 if needed
+;
+; So the mark 1 means (seen as + in c1, but not yet in c2)
+;    the mark 3 means (seen as + in c1)
+;    the mark 2 means (seen as - in c1, but not yet in c2)
+;    the mark 4 means (seen as - in c1)
+;
+; This implementation is robust to literal order, repeated literals, and
+; literals of opposite polarity in the same clause. It is true equality on sets
+; literals.
+;
 ; TODO(aozdemir) This implementation could be further optimized b/c once c1 is
 ; drained, we need not continue to pattern match on it.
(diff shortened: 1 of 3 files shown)


proofs/signatures/drat.plf
```diff
@@ -29,10 +29,20 @@
 ; We could sort them, but that would require defining an order on our variables,
 ; and incurring the cost of sorting.
 ; Instead, we do the following:
-;  1. Sweep the first clause, marking variables with flags 1 (pos) and 2 (neg)
-;  2. Sweep the second clause, erasing marks.
+;  1. Sweep the first clause, marking variables with flags 1,3 (pos) and 2,4 (neg)
+;  2. Sweep the second clause, erasing marks 1,2.
 ;  3. Unsweep the first clause, returning FALSE on marks.
```
yoni206: Returning FALSE on marks 3,4? Or also on 1,2?

proofs/signatures/drat_test.plf
```diff
@@ -1,4 +1,95 @@
...
+
+(check
+  (% a var
+  (% b var
+     (@ c1 (clc (pos a) (clc (pos a) cln))
+     (@ c2 (clc (pos a) (clc (neg b) cln))
+        (: TestSuccess
+           (test_clause_eq c1 c2 ff)))))))
+
+(check
```
yoni206: Isn't it the same as the previous one?

proofs/signatures/drat.plf
```diff
@@ -29,10 +29,20 @@
...
+;    the mark 2 means (seen as - in c1, but not yet in c2)
+;    the mark 4 means (seen as - in c1)
+;
+; This implementation is robust to literal order, repeated literals, and
+; literals of opposite polarity in the same clause. It is true equality on sets
+; literals.
+;
 ; TODO(aozdemir) This implementation could be further optimized b/c once c1 is
 ; drained, we need not continue to pattern match on it.
 (program clause_eq ((c1 clause) (c2 clause)) bool
```
yoni206: Does this program assumes marks 1,2,3,4 are unset upon entry, and unsets them upon exit? If so, maybe worth documenting.

proofs/signatures/drat.plf
alex-ozdemir: Good question!
Yeah -- FALSE is returned on marks 1,2 but not 3,4. I'll fix the documentations.

proofs/signatures/drat.plf
alex-ozdemir: The program does make those assumptions, and they should be documented.

Thanks!

proofs/signatures/drat_test.plf
alex-ozdemir: Whoops! I think I meant to do an ordering test. I'll fix it!

PR #2805, opened 2019-01-16
Add option to print BV constants in binary

4tXJ7f:
This commit adds the option `--bv-print-consts-in-binary` to print
bit-vector constants in binary, e.g. `#b0001`, instead of decimal, e.g.
`(_ bv1 4)`). The option is on by default to match the behavior of Z3
and Boolector.


HanielB:
I'm not sure we want to change this option to default now so we do not surprise our users. I think we should wait until next release.

4tXJ7f:
Your wish is my command. Done.

PR #2813, opened 2019-01-20
Language bindings: Enable catching of exceptions

### LINKED ISSUE #2810: Python wrapper does not encapsulate exceptions
The CVC4 python wrapper does not encapsulate C++ exceptions into python exceptions causing a crash of the entire python application when an exception is thrown by CVC4.

For example, consider the snipped below:
```python
import CVC4
cvc4 = CVC4.SmtEngine(CVC4.ExprManager())
cvc4.setOption("non-existing", CVC4.SExpr("true"))
```

this produces the following error that kills python:
```
terminate called after throwing an instance of 'CVC4::UnrecognizedOptionException'
  what():  Error in option parsing: Unrecognized informational or option key or setting: non-existing
Aborted
```

The expected behavior should be a python exception that is catch-able by python.

4tXJ7f:
Fixes #2810. SWIG relies on `throw` specifiers to determine which
exceptions a method can throw. The wrappers generated by SWIG catch
those C++ exceptions and turn them into exceptions for the target
language. However, we have removed `throw` specifiers because they have
been deprecated in C++11, so SWIG did not know about any of our
exceptions. This commit fixes the issue using the `%catches` directive,
declaring that all methods may throw a `CVC4::Exception` or a general
exception. Note: This means that users of the language bindings will
just receive a general `CVC4::Exception` instead of more specific
exceptions like `TypeException`s. Given that we are planning to have a
single exception type for the new CVC4 API, this seemed like a natural
choice.
Additionally, the commit (significantly) simplifies the mapping of C++
to Java exceptions and fixes an issue with Python exceptions not
inheriting from `BaseException`. Finally, the commit adds API examples
for Java and Python, which demonstrate catching exceptions, and adds
  Python examples as tests in our build system.


examples/api/python/CMakeLists.txt
```diff
@@ -0,0 +1,15 @@
+set(EXAMPLES_API_PYTHON
+  exceptions
+)
+
+foreach(example ${EXAMPLES_API_PYTHON})
+  set(example_test example/api/python/${example})
+  add_test(
+    NAME ${example_test}
+    COMMAND
+      "${PYTHON_EXECUTABLE}" "${CMAKE_SOURCE_DIR}/examples/api/python/${example}.py"
```
aniemetz: I think this needs to depend on the CVC4 Python module.

aniemetz:
@4tXJ7f what's the status of this?

PR #2816, opened 2019-01-22
CVC: Fix printing of tuple and record selectors

4tXJ7f:
Previously, tuple and record selectors were printed as postifx
operations, e.g. `a(r)` instead of `r.a`. This commit fixes the issue.
The commit also adds a regression test. Note that the test uses
`--no-dt-share-sel` because shared selectors make
`getSelectorIndexInternal()` return -1, which should be fixed in a
future commit.

### DIFF
--- a/src/printer/cvc/cvc_printer.cpp
+++ b/src/printer/cvc/cvc_printer.cpp
@@ -404,20 +404,26 @@ void CvcPrinter::toStream(
       }
       break;
     case kind::APPLY_SELECTOR:
-    case kind::APPLY_SELECTOR_TOTAL: {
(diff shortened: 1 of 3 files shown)


ajreynol:
This makes output wrong for non-tuples, for example:
```
% EXPECT: invalid
DATATYPE
  tree = node(left : tree, right : tree) | leaf
END;
x : tree;
GET_VALUE left(x);
```
causes this to be printed as `x.leaf` instead of `leaf(x)`.  I will open a separate PR for fixing this.

4tXJ7f:
Alright, thanks. I've reviewed to other PR. Closing this one.

PR #2817, opened 2019-01-22
Strings: Strengthen multiset reasoning

4tXJ7f:
This commit introduces three helper methods for performing multiset
reasoning: an entailment check whether a term is always a strict subset
of another term in the multiset domain (`checkEntailMultisetSubset()`),
a check whether a string term is always homogeneous
(`checkEntailHomogeneousString()`), and an overapproximation for the
multiset domain (`getMultisetApproximation()`). It also adds unit tests
related to multiset reasoning.


src/theory/strings/theory_strings_rewriter.h
```diff
@@ -540,6 +540,50 @@ class TheoryStringsRewriter {
                                      std::vector<Node>& approx,
                                      bool isOverApprox = false);
 
+  /**
+   * Checks whether it is always true that `a` is a strict subset of `b` in the
+   * multiset domain.
+   *
+   * Example:
+   *
+   * a = (str.++ "A" x), b = (str.++ "A" x "B") ---> true
```
ajreynol: Perhaps use a more informative example, say `a = "A" ++ x, b = "B" ++ x ++ "AA"`

test/unit/theory/theory_strings_rewriter_white.h
```diff
@@ -866,6 +866,14 @@ class TheoryStringsRewriterWhite : public CxxTest::TestSuite
                        d_nm->mkNode(kind::STRING_STRREPL, x, a, empty),
                        a);
     sameNormalForm(lhs, rhs);
+
+    {
+      // (str.contains (str.++ x "A") (str.++ x "B")) ---> false
```
ajreynol: FWIW, this one should be doable with strip constant endpoints.

test/unit/theory/theory_strings_rewriter_white.h
```diff
@@ -945,9 +953,7 @@ class TheoryStringsRewriterWhite : public CxxTest::TestSuite
     Node empty_x = d_nm->mkNode(kind::EQUAL, x, empty);
     sameNormalForm(p_xx, empty_x);
 
-    // (str.suffix x (str.++ x x "A")) --> false
-    //
-    // (= x "")
+    // (str.suffix x (str.++ x x "A")) ---> false
```
ajreynol: The arguments in this comment of this are (and were previously) flipped.

test/unit/theory/theory_strings_rewriter_white.h
```diff
@@ -1075,16 +1082,96 @@ class TheoryStringsRewriterWhite : public CxxTest::TestSuite
...
+
+    {
+      // (= (str.++ x "A") "") ---> false
+      Node eq = d_nm->mkNode(
+          kind::EQUAL, d_nm->mkNode(kind::STRING_CONCAT, x, a), empty);
+      sameNormalForm(eq, f);
+    }
+
+    {
+      // (= (str.++ x "A") "") ---> false
```
ajreynol: This is a duplicate of above?

src/theory/strings/theory_strings_rewriter.cpp
```diff
@@ -385,91 +385,65 @@ Node TheoryStringsRewriter::rewriteStrEqualityExt(Node node)
...
-        }
-      }
-      if (isHomogeneous)
+      Assert(cn.isConst());
+      Assert(cn.getConst<String>().size() == 1);
+      unsigned hchar = cn.getConst<String>().front();
+
+      std::vector<Node> trimmed[2];
+      size_t numHChars[2] = {0, 0};
+      for (size_t j = 0; j < 2; j++)
```
ajreynol: Can you comment a bit on what this loop is doing?  Took me a while to figure this out in the new version.

src/theory/strings/theory_strings_rewriter.cpp
```diff
@@ -385,91 +385,65 @@ Node TheoryStringsRewriter::rewriteStrEqualityExt(Node node)
...
-          Assert(lhss.size() >= rmChar);
-          // we trimmed
-          lhs = nm->mkConst(lhss.substr(0, lhss.size() - rmChar));
-        }
-        Node ss = mkConcat(STRING_CONCAT, trimmed);
-        if (lhs != node[i] || ss != node[1 - i])
+      }
+
+      size_t trimmedConst = std::min(numHChars[0], numHChars[1]);
+      for (size_t j = 0; j < 2; j++)
```
ajreynol: Also add a comment here about adding the difference to the size that has less.

src/theory/strings/theory_strings_rewriter.cpp
4tXJ7f: Done

src/theory/strings/theory_strings_rewriter.cpp
4tXJ7f: Done

src/theory/strings/theory_strings_rewriter.h
4tXJ7f: I've added your example and a negative example as well.

test/unit/theory/theory_strings_rewriter_white.h
4tXJ7f: Great point, I changed it to `(str.contains (str.++ x "A") (str.++ "B" x)) ---> false`

test/unit/theory/theory_strings_rewriter_white.h
4tXJ7f: Fixed

test/unit/theory/theory_strings_rewriter_white.h
4tXJ7f: Hmm, right. Not sure what happened. Removed.

PR #2818, opened 2019-01-22
 Fix tuple and record CVC printing

ajreynol:
Alternative fix to #2816.

This also fixes the issue with the index of shared selectors being -1.


test/regress/regress0/printer/tuples_and_records.cvc
```diff
@@ -0,0 +1,19 @@
+% COMMAND-LINE: --no-dt-share-sel
```
4tXJ7f:
```suggestion
```

test/regress/regress0/printer/tuples_and_records.cvc
```diff
@@ -0,0 +1,19 @@
...
+#];
+r : R;
+
+y: [REAL, INT, REAL];
+
+ASSERT r = (# a := "active", b := "who knows?" #);
+ASSERT y = ( 4/5, 9, 11/9 );
+QUERY r.a = "what?";
+GET_VALUE r.a;
+GET_VALUE y.1;
```
4tXJ7f: Can you add your tree example here?

src/printer/cvc/cvc_printer.cpp
```diff
@@ -407,12 +407,30 @@ void CvcPrinter::toStream(
...
+            else
+            {
+              sindex = dt[0].getSelectorIndexInternal(opn.toExpr());
+            }
+            Assert(sindex >= 0);
+            out << '.' << sindex;
+          }
+          else
+          {
+            out << '.';
```
4tXJ7f: Nit: This can be moved before the if statement.

PR #2835, opened 2019-02-13
Introduce template classes for simple type rules

4tXJ7f:
This commit introduces two template classes `SimpleTypeRule` and
`SimpleTypeRuleVar` to help define simple type rules without writing
lots of redundant code. The main advantages of this approach are (IMHO):

- Less code
- More consistent error reporting
- Easier to extend type checking with other functionality (e.g. getting
the type of a symbol)

_Note_: This is a bit of a proposal, so let me know what you think of the approach.


src/theory/arith/kinds
```diff
@@ -94,40 +94,40 @@ typerule POW ::CVC4::theory::arith::ArithOperatorTypeRule
 
 typerule CONST_RATIONAL ::CVC4::theory::arith::ArithConstantTypeRule
 
-typerule LT ::CVC4::theory::arith::ArithPredicateTypeRule
-typerule LEQ ::CVC4::theory::arith::ArithPredicateTypeRule
-typerule GT ::CVC4::theory::arith::ArithPredicateTypeRule
-typerule GEQ ::CVC4::theory::arith::ArithPredicateTypeRule
+typerule LT "SimpleTypeRule<RBool, AReal, AReal>"
```
mpreiner: It's a bit counter-intuitive to use `<RBool, AReal, AReal>` instead of `<AReal, AReal, RBool>`, but I guess there is no other way with the optional arguments.

4tXJ7f: Yeah, I am not sure if there is a better way, unfortunately. Any ideas?

4tXJ7f:
@ajreynol I seem to remember that you were not a huge fan of this change. What were your concerns again?

ajreynol:
I think I was concerned that it makes the code harder to make tweaks to individual type rules.

Overall I think I'm in favor of merging this though.

4tXJ7f:
Yeah, so the idea was that it shouldn't be harder to make special type
rules (the process for that actually stays exactly the same) but just to
make it really easy to implement common type rules.

ajreynol:
I'm fine with merging.

PR #2837, opened 2019-02-13
New C++ API: Remove redundant declareFun function.

aniemetz:


src/api/cvc4cpp.h
```diff
@@ -2270,15 +2270,6 @@ class CVC4_PUBLIC Solver
   Sort declareDatatype(const std::string& symbol,
                        const std::vector<DatatypeConstructorDecl>& ctors) const;
 
-  /**
-   * Declare 0-arity function symbol.
-   * SMT-LIB: ( declare-fun <symbol> ( ) <sort> )
```
4tXJ7f: Maybe change the comment for `declareSort` to say that it also corresponds to `( declare-fun <symbol> ( ) <sort> )`?

aniemetz: Good point, will do.

aniemetz: Oh wait, it already does.

PR #2839, opened 2019-02-15
 Move tuple/record update elimination from ppRewrite to expandDefinition

### LINKED ISSUE #2838: incomplete model generated combining arrays and records
The following formula is satisfiable, but when I ask CVC4 to print the model, it prints a complicated expression for x instead of giving an integer value.

Ints_0 : ARRAY INT OF INT;
C : TYPE = [# i : INT #];
CType : TYPE = ARRAY INT OF C;
C_0 : CType;
x : INT;
C_1 : CType = C_0 WITH [0].i := 2;

ASSERT C_0[0].i = 0;
ASSERT C_0[1].i = 1;
ASSERT Ints_0[2] = Ints_0[0];
ASSERT x = Ints_0[C_1[0].i];
ASSERT x /= Ints_0[C_1[1].i];
CHECKSAT

ajreynol:
Fixes #2838.

### DIFF
--- a/src/theory/datatypes/theory_datatypes.cpp
+++ b/src/theory/datatypes/theory_datatypes.cpp
@@ -578,8 +578,7 @@ Node TheoryDatatypes::expandDefinition(LogicRequest &logicRequest, Node n) {
       }else{
         selector_use = selector;
       }
-      Node sel = nm->mkNode(
-          kind::APPLY_SELECTOR_TOTAL, selector_use, n[0]);
+      Node sel = nm->mkNode(kind::APPLY_SELECTOR_TOTAL, selector_use, n[0]);
       if (options::dtRewriteErrorSel())
       {
         return sel;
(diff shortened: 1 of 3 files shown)


test/regress/regress0/datatypes/issue2838.cvc
```diff
@@ -0,0 +1,14 @@
+% EXPECT: sat
```
4tXJ7f: Can we maybe get the value for `x` and check whether it matches what we expect it to be?

src/theory/datatypes/theory_datatypes.cpp
```diff
@@ -596,56 +597,70 @@ Node TheoryDatatypes::expandDefinition(LogicRequest &logicRequest, Node n) {
     }
   }
     break;
-  default:
-    return n;
+    case TUPLE_UPDATE:
```
4tXJ7f: The formatting here is off: `case` shouldn't be at the same level of indentation as `break`. Maybe reformat the whole switch statement?

ajreynol:
I've addressed the review in my latest commits.

test/regress/regress0/datatypes/issue2838.cvc
4tXJ7f: I've discussed this offline with @ajreynol and he pointed out that current master fails the regression because `--check-models` detects a non-constant value. Thus, getting the value does not add anything here.

PR #2847, opened 2019-03-02
Enable CryptoMiniSat-backed BV proofs

alex-ozdemir:
* Connect the plumbing so that BV proofs are enabled when using
CryptoMiniSat
* Also fixed a bug in CNF-proof generation
   * Specifically, CNF proofs broke when proving tautological clauses.
     Now they don't.


src/proof/cnf_proof.h
```diff
@@ -164,6 +164,12 @@ class CnfProof {
                            std::ostream& paren,
                            ProofLetMap &letMap) = 0;
 
+  // Actually returns std::optional<std::pair<unsigned, unsigned>>
+  // First field is true if the option is filled.
+  // Detects whether a clause has x v ~x for some x
+  // If so, returns the positive occurence's idx first, then the negative's
+  static std::tuple<bool, unsigned, unsigned> detectTrivialTautology(
```
4tXJ7f: We actually have a [`Maybe` type](https://github.com/CVC4/CVC4/blob/master/src/util/maybe.h) that you should be able to use for those purposes.

src/proof/bitvector_proof.cpp
```diff
@@ -221,6 +221,14 @@ void BitVectorProof::printOwnedTerm(Expr term,
   }
 }
 
+void BitVectorProof::printEmptyClauseProof(std::ostream& os,
```
4tXJ7f: Remove the `= 0` from the header file.

src/proof/cnf_proof.h
alex-ozdemir: This is fantastic! Will do!

src/proof/bitvector_proof.cpp
alex-ozdemir: Whoops! Will do

alex-ozdemir:
Thanks for the review, @4tXJ7f. I think I've addressed all of your comments. Let me know if I missed anything.

src/proof/cnf_proof.cpp
```diff
@@ -431,6 +465,31 @@ void LFSCCnfProof::printCnfProofForClause(ClauseId id,
...
+                    << std::endl;
+
+    std::string proofOfPos =
+        ProofManager::getLitName((*clause)[negIndexInClause], d_name);
+    std::string proofOfNeg =
+        ProofManager::getLitName((*clause)[posIndexInClause], d_name);
+    os << "(contra _ " << proofOfPos << " " << proofOfNeg << ")";
+  }
+  else
+  {
```
4tXJ7f: The indentation of the code within those parentheses is off.

src/proof/cnf_proof.cpp
```diff
@@ -349,6 +349,40 @@ void CnfProof::collectAssertionsForClauses(const IdToSatClause& clauses,
   }
 }
 
+// Detects whether a clause has x v ~x for some x
+// If so, returns the positive occurence's idx first, then the negative's
+Maybe<std::pair<unsigned, unsigned>> CnfProof::detectTrivialTautology(
```
4tXJ7f: Use `size_t` for indices. `unsigned` is likely to be only 32-bit, even on 64-bit platforms.

src/proof/theory_proof.cpp
```diff
@@ -85,8 +85,29 @@ void TheoryProofEngine::registerTheory(theory::Theory* th) {
         if (options::bitblastMode() == theory::bv::BITBLAST_MODE_EAGER
             && options::bvSatSolver() == theory::bv::SAT_SOLVER_CRYPTOMINISAT)
         {
-          proof::BitVectorProof* bvp =
-              new proof::LfscClausalBitVectorProof(thBv, this);
+          proof::BitVectorProof* bvp = nullptr;
+          switch (options::bvProofFormat())
+          {
+            case theory::bv::BvProofFormat::BITVECTOR_PROOF_DRAT:
+              {
```
4tXJ7f: Clang-Format should actually not add an extra indentation here. Can you reformat?

src/proof/cnf_proof.cpp
```diff
@@ -431,6 +465,31 @@ void LFSCCnfProof::printCnfProofForClause(ClauseId id,
...
+  // recording the location of variables in the clause assumes the clause is
+  // not tautological
+  Maybe<std::pair<unsigned, unsigned>> isTrivialTaut =
+      detectTrivialTautology(*clause);
+  if (isTrivialTaut.just())
+  {
+    unsigned posIndexInClause = isTrivialTaut.value().first;
+    unsigned negIndexInClause = isTrivialTaut.value().second;
+    Trace("cnf-pf") << "; Indices " << posIndexInClause << " (+) and "
+                    << negIndexInClause << " (-) make this clause a taut"
```
4tXJ7f:
```suggestion
                    << negIndexInClause << " (-) make this clause a tautology"
```

src/proof/cnf_proof.cpp
```diff
@@ -349,6 +349,40 @@ void CnfProof::collectAssertionsForClauses(const IdToSatClause& clauses,
   }
 }
 
+// Detects whether a clause has x v ~x for some x
+// If so, returns the positive occurence's idx first, then the negative's
+Maybe<std::pair<unsigned, unsigned>> CnfProof::detectTrivialTautology(
+    const prop::SatClause& clause)
+{
+  // a map from a SatVariable to its previous occurence's polarity and location
+  std::map<prop::SatVariable, std::pair<bool, unsigned>> varsToPolsAndIndices;
```
4tXJ7f: It probably doesn't change much in this case but I tend to prefer using an `unordered_map` if the keys don't need to be sorted (`unordered_map` is usually implemented as a hash map and `map` as some sort of tree data structure).

src/proof/clausal_bitvector_proof.h
```diff
@@ -87,6 +85,49 @@ class LfscClausalBitVectorProof : public ClausalBitVectorProof
...
+                         TheoryProofEngine* proofEngine)
+      : LfscClausalBitVectorProof(bv, proofEngine)
+  {
+  }
+
+  void printEmptyClauseProof(std::ostream& os, std::ostream& paren) override;
+};
+
+/**
+ * An Extended Resolution proof for a bitvector problem
```
4tXJ7f:
```suggestion
 * An Extended Resolution proof for a bit-vector problem
```

src/proof/clausal_bitvector_proof.h
```diff
@@ -87,6 +85,49 @@ class LfscClausalBitVectorProof : public ClausalBitVectorProof
...
+                         TheoryProofEngine* proofEngine)
+      : LfscClausalBitVectorProof(bv, proofEngine)
+  {
+  }
+
+  void printEmptyClauseProof(std::ostream& os, std::ostream& paren) override;
+};
+
+/**
+ * An LRAT proof for a bitvector problem
```
4tXJ7f:
```suggestion
 * An LRAT proof for a bit-vector problem
```

src/proof/clausal_bitvector_proof.h
```diff
@@ -87,6 +85,49 @@ class LfscClausalBitVectorProof : public ClausalBitVectorProof
   void printBBDeclarationAndCnf(std::ostream& os,
                                 std::ostream& paren,
                                 ProofLetMap& letMap) override;
+};
+
+/**
+ * A DRAT proof for a bitvector problem
```
4tXJ7f:
```suggestion
 * A DRAT proof for a bit-vector problem
```

src/options/options_handler.cpp
```diff
@@ -1769,7 +1769,8 @@ void OptionsHandler::proofEnabledBuild(std::string option, bool value)
 {
 #ifdef CVC4_PROOF
   if (value && options::bitblastMode() == theory::bv::BITBLAST_MODE_EAGER
-      && options::bvSatSolver() != theory::bv::SAT_SOLVER_MINISAT)
+      && options::bvSatSolver() != theory::bv::SAT_SOLVER_MINISAT
+      && options::bvSatSolver() != theory::bv::SAT_SOLVER_CRYPTOMINISAT)
   {
     throw OptionException(
         "Eager BV proofs only supported when minisat is used");
```
4tXJ7f:
```suggestion
        "Eager BV proofs only supported when MiniSat or CryptoMiniSat is used");
```

src/proof/clausal_bitvector_proof.cpp
```diff
@@ -84,7 +87,16 @@ void ClausalBitVectorProof::calculateAtomsInBitblastingProof()
     Debug("bv::clausal") << "Printing DRAT proof ... " << std::endl;
     dratProof.outputAsText(Debug("bv::clausal"));
   }
-  Unimplemented();
+
+  // Empty any old record of which atoms for used
```
4tXJ7f:
```suggestion
  // Empty any old record of which atoms were used
```

src/proof/clausal_bitvector_proof.cpp
```diff
@@ -101,13 +113,91 @@ void LfscClausalBitVectorProof::printBBDeclarationAndCnf(std::ostream& os,
...
+  os << "(lrat_proof_of_bottom _ proofOfCMap lratProof "
+     << "\n)";
+}
+
+void LfscErBitVectorProof::printEmptyClauseProof(std::ostream& os,
+                                                 std::ostream& paren)
+{
+  Assert(options::bitblastMode() == theory::bv::BITBLAST_MODE_EAGER,
+         "the BV theory should only be proving bottom directly in the eager "
+         "bitblasting mode");
```
4tXJ7f: Instead of repeating these checks, we could just call the implementation in the base class. I am a bit on the fence on whether that would be cleaner or not but it is worth considering.

4tXJ7f:
For the regressions, I think that you can actually just have a single test with the following header:

```
; REQUIRES: cryptominisat
; COMMAND-LINE: --bv-proof-format= lrat --check-proofs
; COMMAND-LINE: --bv-proof-format= drat --check-proofs
; COMMAND-LINE: --bv-proof-format= er --check-proofs
```

This should run the test three times with the different formats.

src/proof/cnf_proof.cpp
alex-ozdemir: Yeah. I'll format this file.

src/proof/cnf_proof.cpp
alex-ozdemir: :+1:

src/proof/theory_proof.cpp
alex-ozdemir: :+1:

src/proof/cnf_proof.cpp
alex-ozdemir: :+1:

src/proof/clausal_bitvector_proof.cpp
alex-ozdemir: I thought about that too. I thought it would be weird to call a function named `print` in another `print` function, so I didn't.

alex-ozdemir:
Okay, @4tXJ7f . I believe that I've addressed your comments. Let me know if this is not the case!

src/proof/cnf_proof.cpp
4tXJ7f: Hmm, formatting the whole file is a bit excessive and makes it hard to see your changes. Any chance that you could just format your changes? [These](https://github.com/CVC4/CVC4/wiki/Developer-Guide#formatting-with-clang-format) explanations may be helpful. The easiest way is probably to do a `git reset HEAD~1` to remove your last commit, then add the non-formatting changes to a separate commit with `git add -p`, commit those changes, discard the remaining changes, then something like `git clang-format master`.

src/proof/er/er_proof.cpp
```diff
@@ -126,12 +127,12 @@ ErProof ErProof::fromBinaryDratProof(const ClauseUseRecord& usedClauses,
   // Parse the resulting TRACECHECK proof into an ER proof.
   std::ifstream tracecheckStream(tracecheckFilename);
   ErProof proof(usedClauses, TraceCheckProof::fromText(tracecheckStream));
-  unlink(tracecheckFilename);
-
-  formStream.close();
-  dratStream.close();
   tracecheckStream.close();
 
+  unlink(formulaFilename);
```
4tXJ7f: Nit: `remove()` is standard C while `unlink()` is POSIX, so it is a bit better to use `remove()` here.

alex-ozdemir: Sounds good! Will do.

src/proof/cnf_proof.cpp
alex-ozdemir: You're right -- thanks for keeping me honest...

I'll do it right and force-push over this.

alex-ozdemir:
Okay, @4tXJ7f , I think I've addressed your latest round of comments. Sorry for the whole-file format. I got a little sloppy there.

Paging @mpreiner. The regression tests that I added are currently skipped because our CI builds are not configured with `drat2er`. Could we add `drat2er` to their configuration?

PR #2849, opened 2019-03-06
Use TMPDIR environment variable for temp files

4tXJ7f:
Previosuly, we were just writing temporary files to `/tmp/` but this
commit allows the user to use the `TMPDIR` environment variable to
determine which directory the temporary file should be written to. The
commit adds a helper function for this and also includes some minor
cleanup of existing code.


4tXJ7f:
The new version of this PR depends on some other commits of Alex' branch, so I'll update once those are in.

4tXJ7f:
I have merged current master and I've changed the new calls to `mkstemp()`. Removing the "do not merge" tag.

PR #2850, opened 2019-03-06
Add statistics for proof gen./checking time, size

4tXJ7f:
This commit adds a statistic that records the total size of all proofs
generated by an instance of `SmtEngine`. The commit also moves
`SmtEngine::checkProof()` into `smt_engine.cpp` because it needs to know
the complete type of `d_stats` (and the separate file for that method
didn't seem that useful). Additionally, it changes
`smt::SmtEngine::checkProofTime` to `smt::SmtEngine::lfscCheckProofTime`
that only measures the time spent in LFSC and adds a statistic
`proof::ProofManager::proofProductionTime` that measures the proof
production time separately (also works with `get-proof`/`--dump-proof`).


src/proof/proof_manager.cpp
```diff
@@ -30,6 +30,7 @@
 #include "proof/theory_proof.h"
 #include "smt/smt_engine.h"
 #include "smt/smt_engine_scope.h"
+#include "smt/smt_statistics_registry.h"
```
mpreiner: This shouldn't be required here since it's already included in `proof/proof_manager.h`

4tXJ7f: Hmm, I don't see there. Am I missing something? We only seem to include `util/statistics_registry.h` there.

mpreiner: You are right! Must have misread...

PR #2852, opened 2019-03-12
 Add getFreeVariables method to node algorithm

ajreynol:
This commit is needed for some of the higher-order work for the CADE 2019 HO-SMT submission.


src/expr/node_algorithm.h
```diff
@@ -59,6 +59,18 @@ bool hasBoundVar(TNode n);
  */
 bool hasFreeVar(TNode n);
 
+/**
+ * Get the free variables in n, that is, the subterms of n of kind
+ * BOUND_VARIABLE that are not bound in n, adds these to fvs.
+ * @param n The node under investigation
+ * @param computeFv If this false is false, then we only return true/false and
```
4tXJ7f:
```suggestion
 * @param computeFv If this flag is false, then we only return true/false and
```

src/expr/node_algorithm.h
```diff
@@ -59,6 +59,18 @@ bool hasBoundVar(TNode n);
...
 
+/**
+ * Get the free variables in n, that is, the subterms of n of kind
+ * BOUND_VARIABLE that are not bound in n, adds these to fvs.
+ * @param n The node under investigation
+ * @param computeFv If this flag is false, then we only return true/false and
+ * do not add to fvs.
+ * @return true iff this node contains a free variable.
+ */
+bool getFreeVariables(TNode n,
```
HanielB: also describe parameter `fvs`

PR #2853, opened 2019-03-12
Properly handle lambdas in relevant domain

ajreynol:
Fix from HO-SMT CADE 2019 branch.


src/theory/quantifiers/relevant_domain.cpp
```diff
@@ -169,7 +169,9 @@ void RelevantDomain::computeRelevantDomain( Node q, Node n, bool hasPol, bool po
         computeRelevantDomainOpCh( rf, n[i] );
       }
     }
-    if( n[i].getKind()!=FORALL ){
+    // do not recurse under nested quantifiers/lambdas
+    if (n[i].getKind() != FORALL && n[i].getKind() != LAMBDA)
```
HanielB: Why don't we add a general utility function `isBinder` that is called in different places of the code whenever you should have such a test?

ajreynol: Oh, strange, we aren't have it: Node::isClosure.  I'll change to use this.

HanielB: Funny this is not used anywhere. BTW shouldn't `isClosure` also include choice and lets?

ajreynol: There is no such thing as a let. I've already updated it to include choice.

PR #2854, opened 2019-03-12
Check for rt library in configuration -- support for glibc<2.17

makaimann:
This is a minor fix for systems with glibc version < 2.17. In that case, we need to link with `-lrt` according to the clock_gettime man page.

Using `check_library_exists` in this way still seems to work for more recent versions of glibc, but I don't know cmake very well so I don't know if it's the *right* way to solve this. Let me know if you'd like any changes!


cmake/ConfigureCVC4.cmake
```diff
@@ -53,7 +54,7 @@ if(CVC4_WINDOWS_BUILD)
     add_c_cxx_flag(-pthread)
   endif()
 else()
-  check_symbol_exists(clock_gettime "time.h" HAVE_CLOCK_GETTIME)
+  check_library_exists(rt clock_gettime "time.h" HAVE_CLOCK_GETTIME)
```
4tXJ7f: Hm, the documentation doesn't say that `check_library_exists` actually adds the library to the linking flags, so maybe we have to do something like:

```
check_symbol_exists(clock_gettime "time.h" HAVE_CLOCK_GETTIME)
if(NOT HAVE_CLOCK_GET_TIME)
  check_library_exists(rt clock_gettime "time.h" HAVE_CLOCK_GETTIME)
  find_library(RT_LIBRARIES NAMES rt)
endif()
```

And then `target_link_libraries(cvc4 ${RT_LIBRARIES})` in the appropriate places.

@mpreiner What do you think?

makaimann: Good point -- it does actually build with only check_library_exists but I'm not sure why.

Just out of curiosity, is this the documentation you checked? https://cmake.org/cmake/help/latest/module/CheckLibraryExists.html

4tXJ7f: Yep, that's the one. Which OS was this on? With `make VERBOSE=1`, are you seeing the `-lrt` flag?

makaimann: CentOS. And huh, that’s very strange... I grepped for it in the verbose output and don’t see `-lrt`...but it builds and if I undo the change it complains about clock_gettime symbols..

4tXJ7f: It could be that some other library includes rt. I'll take a look.

makaimann:
Update -- I made the changes that @4tXJ7f suggested. It still works fine for glibc >= 2.17, but unfortunately it doesn't work on the CentOS environment. I used some `message` statements in cmake to track what's going on and it looks like it enters the if statement fine (because `HAVE_CLOCK_GETTIME` isn't defined) but then `check_library_exists` doesn't seem to execute here: https://github.com/makaimann/CVC4/blob/11a807724920053fa88035d5d6a0fa3fe7ba1827/cmake/ConfigureCVC4.cmake#L59

I can tell because it's supposed to print something like `looking for clock_gettime in rt` and it does not. Additionally, `HAVE_CLOCK_GETTIME` is still not defined after that line. So, it then fails at build time. Strangely, it works if I leave everything the same but comment out this line: https://github.com/makaimann/CVC4/blob/11a807724920053fa88035d5d6a0fa3fe7ba1827/cmake/ConfigureCVC4.cmake#L57

Anyway, I'm very confused -- do you all have any ideas? I'm trying to create a Docker image to reproduce this but am having trouble finding a gcc version that uses an old enough glibc.

makaimann:
Here's a Dockerfile for reproducing this: [docker.tar.gz](https://github.com/CVC4/CVC4/files/2968523/docker.tar.gz)

## Instructions
```
tar -xzvf docker.tar.gz
cd docker && ./get-centos6-docker.sh && cd ../
sudo docker build -t centos6 ./docker
sudo docker run --rm -it centos6
```

Unfortunately the build fails for another reason. But you can still see the rt issue by running `./configure.sh` from the CVC4 repository (which the docker images sets up for you). If you then remove the `check_symbol_exists` line that I pointed out above, it's able to find `clock_gettime`. Let me know if I can provide any more information! I'm continuing to look into it myself.

mpreiner:
`man clock_gettime`  says the following:

> Link with -lrt (only for glibc versions before 2.17).

The problem is that for glibc versions before 2.17 you *additionally* have to link against `rt`. That means you have to check if symbol `clock_gettime` exists in time.h *and* link against the library. 

I guess for this case we have to do a `check_c_source_compiles` check like (untested):

```
set(CHECK_RT_SRC
 "                                                                                 
  #include <time.h>                                                               
  int main(void)
  {
    timespec s;
    (void) clmock_gettime(CLOCK_MONOTONIC, &s);
    return 0;
  }
  "
)
check_c_source_compiles("${CHECK_RT_SRC}" HAVE_CLOCK_GETTIME)

if(NOT HAVE_CLOCK_GETTIME)
  set(CMAKE_REQUIRED_LINK_OPTIONS rt)
  check_c_source_compiles("${CHECK_RT_SRC}" HAVE_CLOCK_GETTIME)
  unset(CMAKE_REQUIRED_LINK_OPTIONS)
  if(HAVE_CLOCK_GETTIME)
    set(REQUIRE_RT_LIBRARY TRUE)
  endif()
endif()
```

In `src/CMakeLists.txt` you would need to add something like:
```
  if(REQUIRE_RT_LIBRARY)
    target_link_libraries(cvc4 rt)
  endif()
```

makaimann:
Thanks @mpreiner for looking into that. I made the suggested change and with some slight tweaking it appears to work. PSA on that -- I mentioned above that it wasn't running the check again and I found documentation on that at the bottom of this page: https://cmake.org/cmake/help/latest/module/CheckCSourceCompiles.html. Turns out it won't re-evaluate if you use the same result variable, unless you explicitly remove it from the CACHE first. Earlier I was using a different `check_*` command before, but my guess is it behaves the same.

The only thing I'm currently worried about is this line:
https://github.com/CVC4/CVC4/blob/5ec1aa3658120e15227e4493b46d566ee10e3b05/cmake/ConfigureCVC4.cmake#L74
I was following the pattern that is used in the rest of the file of setting and unsetting, and I also checked that `CMAKE_REQUIRED_LIBRARIES` was not set before I set it a few lines above. But, is there any chance we'd rely on it being persistent at some point in the future?

mpreiner:
CMAKE_REQUIRED_LIBRARIES is only used for add additional libraries for the check. This should not be used as a system-wide library dependency variable anyways and we don't rely on it.

src/CMakeLists.txt
```diff
@@ -770,6 +770,13 @@ endif()
 target_link_libraries(cvc4 ${GMP_LIBRARIES})
 target_include_directories(cvc4 PUBLIC ${GMP_INCLUDE_DIR})
 
+# Add rt library
+# Note: for glibc >= 2.17 this is not necessary because rt is already included
```
mpreiner:
```suggestion
# Note: For glibc < 2.17 we have to additionally link against rt (man clock_gettime).
```

src/CMakeLists.txt
```diff
@@ -770,6 +770,13 @@ endif()
...
+#       in that case RT_LIBRARIES will be empty
```
mpreiner:
```suggestion
```

src/CMakeLists.txt
```diff
@@ -770,6 +770,13 @@ endif()
 target_link_libraries(cvc4 ${GMP_LIBRARIES})
 target_include_directories(cvc4 PUBLIC ${GMP_INCLUDE_DIR})
 
+# Add rt library
```
mpreiner:
```suggestion
```

mpreiner:
@makaimann Can you maybe check again if the `check_library_exists` as suggested by @4tXJ7f works with clearing the cache?

makaimann:
Good point, I tried it and that does work. The latest commit reverts to that implementation.

PR #2859, opened 2019-03-14
Add more NEWS

4tXJ7f:


4tXJ7f:
Note: I intentionally didn't include QF_LRA proofs, DRAT/LRAT proofs, and news related to higher-order support because those features are not fully merged in yet.

NEWS
```diff
@@ -4,17 +4,31 @@ Changes since 1.6
...
+    (enabled by default).
+* SyGuS:
+  * Support for two new term enumerator strategies: variable agnostic
+    (`--sygus-active-gen=var-agnostic`) and fast (`--sygus-active-gen=enum`).
+    By default, CVC4 tries to choose the best term enumerator strategy
+    automatically based on the input (`--sygus-active-gen=auto`).
 
 Improvements:
+* Strings:
+  * Significantly better performance on common string benchmarks.
```
ajreynol: Maybe elaborate on common: string benchmarks with extended string functions like substring, contains, replace.

4tXJ7f: As discussed offline (just reiterating here for the record), our performance also improved on benchmarks without extended string functions.

ajreynol: Regardless this should be made explicit here though instead of just saying "common".  Maybe something like:

`string benchmarks over the core theory, and those with extended string functions like substring, contains, replace.`

NEWS
```diff
@@ -4,17 +4,36 @@ Changes since 1.6
 =================
 
 New Features:
-* Strings: Support for `str.replaceall` operator.
+* Proofs:
+  * Support for bit-vector proofs with eager bitblasting (older versions only
+    supported proofs with lazy bitblasting).
+* Strings:
+  * Support for `str.replaceall` operator.
+  * New option `--re-elim` to reduce regular expressions to extended string
```
ajreynol: I would indicate this is enabled by default.

PR #2865, opened 2019-03-14
Add higher-order elimination preprocessing pass

ajreynol:


HanielB:
What is the contract of `eliminateHo` when `options::HoElim` is false? The documentation of the preprocessing pass makes it look like this module is used only to reduce HOL to FOL, but when we are not doing `hoElim` it removes the lambdas and maybe adds store axioms, is that it?

ajreynol:
I've added a comment that describes how options affect the pass.

PR #2867, opened 2019-03-14
Adding capture avoiding substitution

HanielB:
Introduces utility functions to perform capture avoiding substitution in nodes. Whenever a binder is found during the recursive application of a substitution, a renaming is prepended to the substitution when "going into" the binder, so that bound variables are renamed. Getting out of the binder, the renaming is removed.

Since we are modifying the substitution during the recursion, this new function uses vectors and Nodes.

### DIFF
--- a/src/expr/node.h
+++ b/src/expr/node.h
@@ -239,9 +239,17 @@ class NodeTemplate {
 
 public:
 
+  /**
+   * Cache-aware, recursive version of substituteCaptureAvoiding() used by the
+   * public member function with a similar signature.
+   */
   Node substituteCaptureAvoiding(Node node, Node replacement,
                   std::unordered_map<Node, Node, NodeHashFunction>& cache) const;
 
+  /**
+   * Cache-aware, recursive version of substituteCaptureAvoiding() used by the
+   * public member function with a similar signature.
+   */
   Node substituteCaptureAvoiding(
       std::vector<Node> source,
       std::vector<Node> dest,
@@ -467,7 +475,7 @@ public:
     assertTNodeNotExpired();
     return getMetaKind() == kind::metakind::VARIABLE;
   }
-  
+
   /**
    * Returns true if this node represents a nullary operator
    */
@@ -475,7 +483,7 @@ public:
     assertTNodeNotExpired();
     return getMetaKind() == kind::metakind::NULLARY_OPERATOR;
   }
-  
+
   inline bool isClosure() const {
     assertTNodeNotExpired();
     return getKind() == kind::LAMBDA ||
@@ -532,8 +540,16 @@ public:
    */
   TypeNode getType(bool check = false) const;
 
+  /**
+   * Substitution of Nodes in a capture avoiding way.
+   */
   Node substituteCaptureAvoiding(Node node, Node replacement) const;
 
+  /**
+   * Simultaneous substitution of Nodes in a capture avoiding way.  Elements in
+   * source will be replaced by their corresponding element in dest.  Both
+   * vectors should have the same size.
+   */
   Node substituteCaptureAvoiding(std::vector<Node> source,
                                  std::vector<Node> dest) const;
 
@@ -1388,8 +1404,8 @@ Node NodeTemplate<ref_count>::substituteCaptureAvoiding(
     std::unordered_map<Node, Node, NodeHashFunction>& cache) const
 {
   // in cache?
-  typename std::unordered_map<Node, Node, NodeHashFunction>::const_iterator
-      i = cache.find(*this);
+  typename std::unordered_map<Node, Node, NodeHashFunction>::const_iterator i =
+      cache.find(*this);
   if (i != cache.end())
   {
     return (*i).second;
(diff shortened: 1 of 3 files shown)


src/expr/node.h
```diff
@@ -239,6 +239,22 @@ class NodeTemplate {
 
 public:
 
+  /**
+   * Cache-aware, recursive version of substituteCaptureAvoiding() used by the
+   * public member function with a similar signature.
+   */
+  Node substituteCaptureAvoiding(Node node, Node replacement,
```
4tXJ7f: I would prefer having those functions in `node_algorithm` in the interest of compilation times since they are not very commonly used.

src/expr/node.h
```diff
@@ -239,6 +239,22 @@ class NodeTemplate {
 
 public:
 
+  /**
+   * Cache-aware, recursive version of substituteCaptureAvoiding() used by the
+   * public member function with a similar signature.
```
4tXJ7f: This one is public, too, no? :)

src/expr/node.h
```diff
@@ -1331,6 +1360,117 @@ TypeNode NodeTemplate<ref_count>::getType(bool check) const
...
 }
 
+template <bool ref_count>
+inline Node NodeTemplate<ref_count>::substituteCaptureAvoiding(
+    Node node, Node replacement) const
+{
+  if (node == *this)
+  {
+    return replacement;
+  }
```
4tXJ7f: I guess one other quick check that we may want to do is `node == replacement ---> return *this`.

HanielB:
The original substitution is and that does not seem to be an issue...

src/expr/node.h
```diff
@@ -1331,6 +1360,117 @@ TypeNode NodeTemplate<ref_count>::getType(bool check) const
   return NodeManager::currentNM()->getType(*this, check);
 }
 
+template <bool ref_count>
+inline Node NodeTemplate<ref_count>::substituteCaptureAvoiding(
```
4tXJ7f: I wouldn't use `inline` here. If it is here for performance reasons, the compiler should automatically figure out whether to inline the code or not and we probably don't need it for linking reasons here either.

src/expr/node.h
```diff
@@ -1331,6 +1360,117 @@ TypeNode NodeTemplate<ref_count>::getType(bool check) const
...
+  std::vector<Node> source;
+  std::vector<Node> dest;
+  source.push_back(node);
+  dest.push_back(replacement);
+  return substituteCaptureAvoiding(source, dest, cache);
+}
+
+template <bool ref_count>
+inline Node NodeTemplate<ref_count>::substituteCaptureAvoiding(
+    std::vector<Node> source, std::vector<Node> dest) const
```
4tXJ7f: Make vectors `const std::vector<Node>&` to avoid copying the vectors.

src/expr/node.h
```diff
@@ -1331,6 +1360,117 @@ TypeNode NodeTemplate<ref_count>::getType(bool check) const
...
+  }
+  NodeBuilder<> nb(getKind());
+  if (getMetaKind() == kind::metakind::PARAMETERIZED)
+  {
+    // push the operator
+    nb << getOperator().substituteCaptureAvoiding(source, dest, cache);
+  }
+  for (const_iterator i = begin(), iend = end(); i != iend; ++i)
+  {
+    nb << (*i).substituteCaptureAvoiding(source, dest, cache);
```
4tXJ7f: We should make this function non-recursive. Right now, we are copying the `source` and the `dest` vectors every time we go to a child node and the nesting of nodes can get very deep.

4tXJ7f:
> The original substitution is and that does not seem to be an issue...

This is actually something that bothered me for a while. Substitution really should be done in a non-recursive fashion because you can get very deep nesting and because it is a widely used method, so it should be as efficient as possible. Your current implementation also copies the source and dest vectors every time you descend into a child (as mentioned in one of my comments) while the "normal" `substitute()` method doesn't have that issue because it uses iterators instead of vectors.

4tXJ7f:
After a discussion offline, we have decided to keep `substituteCaptureAvoiding()` separate from `substitute()` because most of our code only deals with shadowing-free terms and `isClosure()` would fire often, so there would be unnecessary overhead.

src/expr/node.h
HanielB: OK, will do

HanielB:
I've addressed the reviews, PTAL

src/expr/node_algorithm.cpp
```diff
@@ -273,5 +273,121 @@ void getSymbols(TNode n,
...
+
+        NodeManager* nm = NodeManager::currentNM();
+        for (const Node& v : curr[0])
+        {
+          vars.push_back(v);
+          renames.push_back(nm->mkBoundVar(v.getType()));
+        }
+        // have new vars -> renames subs in the beginning of current sub
+        src.insert(src.begin(), vars.begin(), vars.end());
+        dest.insert(dest.begin(), renames.begin(), renames.end());
```
4tXJ7f: This is not critical for this PR but I think that it would be a bit more efficient to insert the new variables at the end of `src`/`dest` (and without creating `vars` and `renames`) and then use `std::find(src.rbegin(), src.rend(), curr)` on line 312.

HanielB: Good suggestion! I changed accordingly

4tXJ7f: Cool, this looks good :)

PR #2871, opened 2019-03-18
Trace tags for dumping the decision tree in org-mode format

makaimann:
This would add tracing options to print the decision tree in [org-mode](https://orgmode.org/) format which can be viewed with emacs or [vim-orgmode](https://github.com/jceb/vim-orgmode). 

In the raw format, the number of asterisks denote the decision level of a decision, and within a propagation node, the number of spaces denote the level. Most viewers render the asterisks as indented bullets.

There are some options for controlling verbosity:
`dtview` - prints the decisions (basic option)
`dtview::command` - also prints smt-lib commands as they're issued
`dtview::conflict` - also prints conflicts
`dtview::prop` - also prints propagations 

Example usage:
`cvc4 -t dtview -t dtview::command -t dtview::conflict -t dtview::prop <example smt2 file> &> example-trace.org`

The resulting file when opened with an org-mode viewer has collapsible nodes, where "..." marks a node with children. Here's an example screenshot: example-trace


src/prop/minisat/core/Solver.cc
```diff
@@ -607,6 +607,20 @@ Lit Solver::pickBranchLit()
...
+                               - (options::incrementalSolving() ? 1 : 0),
+                           '*')
+            << " " << proxy->getNode(MinisatSatSolver::toSatLiteral(nextLit))
+            << " :THEORY-DECISION:" << std::endl;
+        Trace("dtview::prop")
+            << std::string(context->getLevel() + 1
+                               - (options::incrementalSolving() ? 1 : 0),
+                           '*')
+            << " /Propagations [Last Decision Repeated]/" << std::endl;
+
```
mpreiner: How about wrapping these trace statements in a function to reduce the code duplication?

src/prop/minisat/core/Solver.cc
```diff
@@ -630,10 +644,23 @@ Lit Solver::pickBranchLit()
...
+                                 - (options::incrementalSolving() ? 1 : 0),
+                             '*')
+                      << " "
+                      << proxy->getNode(MinisatSatSolver::toSatLiteral(nextLit))
+                      << " :DE-DECISION:" << std::endl;
+      Trace("dtview::prop")
+          << std::string(context->getLevel() + 1
+                             - (options::incrementalSolving() ? 1 : 0),
+                         '*')
+          << " /Propagations [Last Decision Repeated]/" << std::endl;
```
mpreiner: Call trace helper function instead.

src/prop/minisat/core/Solver.cc
```diff
@@ -667,12 +694,31 @@ Lit Solver::pickBranchLit()
...
+          << std::string(
+                 context->getLevel() - (options::incrementalSolving() ? 1 : 0),
+                 '*')
+          << " " << proxy->getNode(MinisatSatSolver::toSatLiteral(decisionLit))
+          << " :DE-DECISION:" << std::endl;
+      Trace("dtview::prop")
+          << std::string(context->getLevel() + 1
+                             - (options::incrementalSolving() ? 1 : 0),
+                         '*')
+          << " /Propagations [Last Decision Repeated]/" << std::endl;
```
mpreiner: Ditto.

src/prop/minisat/core/Solver.cc
makaimann: Thanks for the feedback! Does something like this look okay? https://github.com/CVC4/CVC4/blob/4f4cc1ac23b9cef207fd21d924b4e74508b75e16/src/prop/minisat/core/Solver.h#L460

4tXJ7f:
I think that adding a helper function in these cases is probably ok but a word of warning when adding helper functions for debug output: the arguments to those functions are always evaluated no matter whether tracing is active or not. This can be [expensive](https://github.com/CVC4/CVC4/pull/2937#issuecomment-482358212). Note that this is not the case with `Trace` because of [macro magic](https://github.com/CVC4/CVC4/blob/master/src/base/output.h#L428).

makaimann:
To be honest, I'm not sure how I feel about having any overhead for this tracing feature. Although code duplication is also bad... Even if it's inlined they're still evaluated? Would it be stupid to make another macro? I try to avoid macros when possible. I'll defer to better judgement on what to do here, but I am uncomfortable adding overhead for a feature that I  don't anticipate will be used very often, and probably only be developers.

4tXJ7f:
One possible solution is to instead have a helper function that returns a string and then use it as follows: `Trace("foo") << myHelper()`. It would be a bit more expensive to actually produce the trace output but you don't have to pay for evaluating the arguments when you are not tracing. Thoughts?

makaimann:
I like that a lot better, good idea! Just implemented that [here](https://github.com/CVC4/CVC4/pull/2871/commits/522236d34240310804dd97fe1d72b224c6282d09).

src/prop/minisat/core/Solver.h
```diff
@@ -455,6 +455,26 @@ class Solver {
...
+    }
+
+    static inline std::string dtviewPropagationHeaderHelper(size_t level)
+    {
+      std::ostringstream msg;
+      msg << std::string(level + 1 - (options::incrementalSolving() ? 1 : 0),
+                         '*')
+          << " /Propagations/";
+      return msg.str();
+    }
```
mpreiner: This doesn't need to be a member of  the Solver class. You can move the helpers to `Solver.cc` into an anonymous namespace.

mpreiner:
> I think that adding a helper function in these cases is probably ok but a word of warning when adding helper functions for debug output: the arguments to those functions are always evaluated no matter whether tracing is active or not. This can be [expensive](https://github.com/CVC4/CVC4/pull/2937#issuecomment-482358212). Note that this is not the case with `Trace` because of [macro magic](https://github.com/CVC4/CVC4/blob/master/src/base/output.h#L428).

What about wrapping the function calls in an `Trace.isOn("dtview")`?

src/prop/minisat/core/Solver.h
makaimann: Done. Thanks for the feedback.

makaimann:
> What about wrapping the function calls in an `Trace.isOn("dtview")`?

I didn't even realize you could do that -- cool! I put the functions in anonymous namespace, switched them back to void and guarded with `Trace.isOn(...)`. It's a little strange, because we check the `Trace` twice, but I think it's worth it to avoid the overhead when we're not tracing. After all, this is placed in a pretty critical part of the code.

PR #2873, opened 2019-03-19
Fix fairness issue with fast sygus enumerator

ajreynol:
Previously, a sygus enumerator could exhibit non-terminating behavior if there was a sygus type with infinite cardinality and had a subfield type with infinite cardinality but admitted only finitely many semantically unique terms based on PBE symmetry breaking.

This commit makes two changes to the optimized enumerator for ensuring we are fair for these cases:
- Master enumerators now "break" (i.e. they return the null term while indicating increment success) once at each size boundary,
- Slave enumerators now abort their increment method based on the current size of master enumerators.

This corrects a non-termination bug from the cvc-users mailing list.

### DIFF
--- a/src/theory/quantifiers/sygus/sygus_enumerator.cpp
+++ b/src/theory/quantifiers/sygus/sygus_enumerator.cpp
@@ -840,6 +840,7 @@ bool SygusEnumerator::TermEnumMaster::incrementInternal()
           Trace("sygus-enum-debug2") << "master(" << d_tn
                                      << "): failed addTerm\n";
           // we will return null (d_currTermSet is true at this point)
+          Assert(d_currTermSet);
           d_currTerm = Node::null();
         }
       }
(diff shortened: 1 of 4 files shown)


src/theory/quantifiers/sygus/sygus_enumerator.cpp
```diff
@@ -819,6 +839,7 @@ bool SygusEnumerator::TermEnumMaster::incrementInternal()
           // the term was not unique based on rewriting
           Trace("sygus-enum-debug2") << "master(" << d_tn
                                      << "): failed addTerm\n";
+          // we will return null (d_currTermSet is true at this point)
```
4tXJ7f: Maybe we should assert this?

ajreynol: sure

PR #2876, opened 2019-03-20
Use empty vector instead of false in query with null Expr assumption

makaimann:
This solution is less confusing than using a `false` assumption.


makaimann:
@aniemetz that makes sense -- I added `std::` there and to `checkSatisfiability`. I didn't add it throughout the whole file though, just to keep the commit small.

PR #2878, opened 2019-03-20
Fix RewriteITEBv to ensure rewrite to fixpoint

4tXJ7f:
`TheoryBVRewriter::RewriteITEBv()` is currently always returning the
status `REWRITE_DONE`. This can result in a situation where a rewritten
node can be rewritten again (which breaks the contract of our rewriter).
The unit test in this commit illustrates the issue. The commit fixes the
issue by returning `REWRITE_AGAIN` or `REWRITE_AGAIN_FULL` if a node
changed. `REWRITE_AGAIN_FULL` is needed if the resulting node may have a
child that is not a subterm of the original expression.


mpreiner:
@4tXJ7f Can you run the old and the new version on the cluster on QF_BV just to make sure that this was not done for a reason?

4tXJ7f:
`BITVECTOR_ITE` currently only seems to be used with the `--bool-to-bv` preprocessing pass. Maybe we could test @makaimann's benchmarks that the preprocessing pass was designed for?

aniemetz:
Just for the record, `BITVECTOR_ITE` nodes can be created through the (new) C++ (and Python) API.

makaimann:
I can run it with this change using bool-to-bv=ite/all. That's how this bug came up and it only occurred on 13 (or so) benchmarks from QF_ABV. Interestingly, they only failed on --bool-to-bv=all and not --bool-to-bv=ite, not sure why. 

I'll get back to you with results when they're done running. We just care about errors right?

4tXJ7f:
Errors and performance, ideally. I think that this commit could potentially improve performance because we are rewriting more aggressively but most likely it won't have any effect. The issue only occurred with some very specific patterns (see unit test), so I am not too surprised that it only happened with `--bool-to-bv=all`.

makaimann:
All right, here are the comparisons on both QF_ABV and QF_BV. The options used are `--simplification=none --bool-to-bv=<ite|all>`. I used simplification=none to help find errors because simplification can sometimes remove terms that would otherwise be rewritten by bool-to-bv. `master-*` runs are from this branch before this commit and `rewrite-*` include this commit. There is still one error, which I'm looking into currently. Otherwise, performance seems to be slightly worse but I think it's just noise -- screenshots attached.
QF_BV_comparison_fixRewriteITEBv
QF_ABV_comparison_fixRewriteITEBv

4tXJ7f:
The error was fixed in #2903. Performance looks good to me. The slightly longer time it takes with the fix could be because we don't error out (passing `--penalty <timeout>` to the compare-smt-runs script would take care of that).

@mpreiner Do you agree that this is ready to be merged?

PR #2879, opened 2019-03-21
Add Python bindings using Cython -- see below for more details

makaimann:
### A brief summary of the changes
* In src/api/python, added Cython files for generating python bindings
* Created python objects to wrap the new c++ API classes/methods.
* genkinds.py automatically reads cvc4cppkind.h and generates cvc4kinds.pxd and cvc4kinds.pxi
* cvc4.pxd and cvc4.pxi define the rest of the bindings.
* The *.pxd files are like headers, and the *.pxi files are included in pycvc4.pyx, so that there's only one generated cython .so file.
* There were some minor changes needed in the src/api/cvc4cpp.* files, mainly just nullary constructors, which Cython requires.
* Finally, in examples/api/python there are 9 examples files demonstrating the Python API based on the examples/api/*-new.cpp examples
* @mpreiner then added the necessary cmake infrastructure, which can be invoked with the `--python-bindings` option of `configure.sh`


CMakeLists.txt
```diff
@@ -519,7 +526,8 @@ print_config("Valgrind             :" ENABLE_VALGRIND)
 message("")
 print_config("Shared libs          :" ENABLE_SHARED)
 print_config("Static binary        :" ENABLE_STATIC_BINARY)
-print_config("Java bindings        :" BUILD_BINDINGS_JAVA)
+print_config("Java bindings        :" BUILD_SWIG_BINDINGS_JAVA)
+print_config("Python bindings      :" BUILD_SWIG_BINDINGS_PYTHON)
```
4tXJ7f: This seems to print "Python bindings" twice. Maybe we can have two separate titles?

makaimann: Good feedback! What do you think of the fix in 395c653074dad07b42e2e8785f6bc2e205179f86? Basically just calls them "SWIG bindings" if they're generated by SWIG.

4tXJ7f: That works for me :+1:

src/api/python/genkinds.py
```diff
@@ -0,0 +1,243 @@
+#!/usr/bin/env python
```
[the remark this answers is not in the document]
makaimann: Done with: 4458a33dea158e5dc17b93cec472e5e24427ae9f

src/api/python/cvc4.pxi
```diff
@@ -0,0 +1,1104 @@
...
+                v.push_back((<Term?> a).cterm)
+            term.cterm = self.csolver.mkTerm(k.k,
+                                             (<OpTerm?> args[0]).copterm, v)
+        else:
+            for a in args:
+                v.push_back((<Term?> a).cterm)
+            term.cterm = self.csolver.mkTerm(k.k, v)
+        return term
+
+    def mkOpTerm(self, kind k, *args):
```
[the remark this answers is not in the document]
makaimann: Good point, fixed in 2fa466d

src/api/python/cvc4.pxi
```diff
@@ -0,0 +1,1104 @@
...
+    def getValue(self, Term t):
+        cdef Term term = Term()
+        term.cterm = self.csolver.getValue(t.cterm)
+        return term
+
+    def pop(self, nscopes=None):
+        if nscopes is None:
+            self.csolver.pop()
+        else:
+            self.csolver.pop(nscopes)
```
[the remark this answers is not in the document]
makaimann: Makes sense! Fixed here 0aae4b4.

makaimann:
Hey @mpreiner, I've been pushing a few commits to keep us up to date with master. On my desktop, I get the following cmake warning (which actually stops configuration):
```
CMake Warning (dev) at cmake/UseCython.cmake:148 (if):
  Policy CMP0057 is not set: Support new IN_LIST if() operator.  Run "cmake
  --help-policy CMP0057" for policy details.  Use the cmake_policy command to
  set the policy and suppress this warning.

  IN_LIST will be interpreted as an operator when the policy is set to NEW.
  Since the policy is not set the OLD behavior will be used.
Call Stack (most recent call first):
  src/api/python/CMakeLists.txt:25 (add_cython_target)
This warning is for project developers.  Use -Wno-dev to suppress it.

CMake Error at cmake/UseCython.cmake:148 (if):
  if given arguments:

    "C" "IN_LIST" "languages"

  Unknown arguments specified
Call Stack (most recent call first):
  src/api/python/CMakeLists.txt:25 (add_cython_target)


-- Configuring incomplete, errors occurred!
See also "/home/makaim/repos/forks/CVC4/build/CMakeFiles/CMakeOutput.log".
See also "/home/makaim/repos/forks/CVC4/build/CMakeFiles/CMakeError.log".
```

I'm able to fix this locally by adding `cmake_policy(SET CMP0057 NEW)` in `src/api/python/CMakeLists.txt`. However, it looks like this "new" behaviour was not introduced until CMake 3.3 (https://cmake.org/cmake/help/v3.3/policy/CMP0057.html), but we only require 3.2. Do you have any suggestions? I could just require 3.3 and set the policy. Or maybe we could modify `UseCython.cmake` to not use `IN_LIST`?

mpreiner:
@makaimann You can check if the policy exists and if this is the case you can enable the policy to NEW behavior.

```
if(POLICY CMP0057)
  cmake_policy(SET CMP0057 NEW)
endif()
```

makaimann:
Great, thanks! Just made that change. Do we test it with cmake 3.2 anywhere? I can try it in a Docker build sometime. I'm not sure if it will still run properly using the OLD behavior of IN_LIST.

Other than that, I think everything is updated. So we're ready for more reviews when people have a chance. @mpreiner suggested @barrettcw, @aniemetz, @4tXJ7f, and maybe @ajreynol.

src/api/cvc4cpp.h
```diff
@@ -82,6 +82,9 @@ class CVC4_PUBLIC Result
    */
   Result(const CVC4::Result& r);
 
+  // Nullary constructor for Cython
```
aniemetz:
```suggestion
  /** Nullary constructor (required for Cython). */
```

src/api/cvc4cpp.h
```diff
@@ -1235,6 +1238,9 @@ class CVC4_PUBLIC DatatypeConstructor
     friend class DatatypeConstructor;  // to access constructor
 
    public:
+    // Nullary constructor for Cython
```
aniemetz:
```suggestion
    /** Nullary constructor (required for Cython). */
```

src/api/cvc4cpp.cpp
```diff
@@ -712,6 +712,9 @@ class CVC4ApiExceptionStream
 
 Result::Result(const CVC4::Result& r) : d_result(new CVC4::Result(r)) {}
 
+// Nullary constructor for Cython
```
```suggestion

```

src/api/cvc4cpp.cpp
```diff
@@ -1801,6 +1807,9 @@ Datatype::Datatype(const CVC4::Datatype& dtype)
 {
 }
 
+// Nullary constructor for Cython
```
```suggestion

```

src/api/cvc4cpp.cpp
```diff
@@ -1866,6 +1877,9 @@ Datatype::const_iterator::const_iterator(const CVC4::Datatype& dtype,
   d_idx = begin ? 0 : cons->size();
 }
 
+// Nullary constructor for Cython
```
```suggestion

```

src/api/cvc4cpp.h
```diff
@@ -1338,6 +1344,9 @@ class CVC4_PUBLIC Datatype
    */
   Datatype(const CVC4::Datatype& dtype);
 
+  // Nullary constructor for Cython
```
aniemetz:
```suggestion
  /** Nullary constructor (required for Cython). */
```

src/api/cvc4cpp.h
```diff
@@ -1387,6 +1396,9 @@ class CVC4_PUBLIC Datatype
     friend class Datatype;  // to access constructor
 
    public:
+    // Nullary constructor for Cython
```
aniemetz:
```suggestion
    /** Nullary constructor (required for Cython). */
```

src/api/python/genkinds.py
```diff
@@ -0,0 +1,254 @@
...
+"""
+This script reads CVC4/src/api/cvc4cppkind.h and generates
+.pxd and .pxi files which declare all the CVC4 kinds and
+implement a Python wrapper for kinds, respectively. The
+default names are kinds.pxd / kinds.pxi, but the name is
+configurable from the command line with --kinds-file-prefix.
+
+The script is aware of the '#if 0' pattern and will ignore
+kinds declared between '#if 0' and '#endif'.
```
aniemetz: Can it also handle nested `#if 0` (or rather, `#if`* ) blocks? (It doesn't look like it, correct me if I'm wrong.)

makaimann: It cannot, but if that's important I could try to add that.

makaimann: So it actually does handle nested `#if 0` blocks, but no other macros at this point.

aniemetz: As discussed offline with @makaimann , that's absolutely sufficient for us here.

aniemetz: Can you maybe add a comment that it supports nested `#if 0` blocks?

makaimann: Definitely! Done with **0fa83b5910feb94a0587b4070ab6d44263d87d0c**

makaimann:
Of course, thanks for all the great feedback! Looks like there's one conflict, so I'll just resolve that quickly.

makaimann:
@mpreiner I had to make a temporary change to the cmake infrastructure because of this recent change: https://github.com/CVC4/CVC4/blob/8e4dbbe1ef9a82622a784f0a905d01ce4fdc1e7d/src/api/cvc4cpp.h#L25

I just added the necessary include directory for now: 9cb2ac7547392e8e3c343483c11df9de3a07d710

What do you think?

PR #2885, opened 2019-03-22
Strip non-matching beginning from indexof operator

4tXJ7f:
This commit adds a rewrite that strips endpoints from `str.indexof`
operators if they don't overlap with the string that is being searched
for using `stripConstantEndpoints()`. In addition to that, it makes
`stripConstantEndpoints()` slightly more aggressive by allowing it to
drop substring components that have zero overlap with the string that we
are looking at. Finally, the commit fixes the default argument for
`fullRewriter` of `checkEntailContains()` to be true instead of false,
which should allow for more rewriting opportunities.

Benchmarks are running.

### DIFF
--- a/src/theory/strings/theory_strings_rewriter.cpp
+++ b/src/theory/strings/theory_strings_rewriter.cpp
@@ -2323,7 +2323,7 @@ Node TheoryStringsRewriter::rewriteIndexof( Node node ) {
         if (stripConstantEndpoints(children0, children1, nb, ne, 1))
         {
           // str.indexof(str.++("AB", x, "C"), "C", 0) --->
-          // str.indexof(str.++(x, "C"), "C", 0)
+          // 2 + str.indexof(str.++(x, "C"), "C", 0)
           Node ret =
               nm->mkNode(kind::PLUS,
                          nm->mkNode(kind::STRING_LENGTH,
--- a/src/theory/strings/theory_strings_rewriter.cpp
+++ b/src/theory/strings/theory_strings_rewriter.cpp
@@ -2318,6 +2318,22 @@ Node TheoryStringsRewriter::rewriteIndexof( Node node ) {
           Node ret = nm->mkNode(kind::STRING_STRIDOF, nn, node[1], node[2]);
           return returnRewrite(node, ret, "idof-def-ctn");
         }
+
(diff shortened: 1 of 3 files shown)


4tXJ7f:
Here are the results:
```
             DIRECTORY |                                      qf_slia_master |                            qf_slia_index_of_rewrite |
             BENCHMARK |  STAT  SLVD   TOT  TO MO ERR  CPU[s] SPACE[MB] UNIQ |  STAT  SLVD   TOT  TO MO ERR  CPU[s] SPACE[MB] UNIQ |
     20180523-Reynolds |    ok 25298 25421 123  0   0 80233.4  480723.7    3 |    ok 25296 25421 125  0   0 80228.2  474559.4    1 |
 appscan-pisa-smtlib25 |    ok    20    20   0  0   0   174.5     312.6    0 |    ok    20    20   0  0   0   182.8     310.5    0 |
     str-pred-small-rw |    ok   818   835  17  0   0  5103.3   12953.8    0 |    ok   818   835  17  0   0  5103.1   12805.0    0 |
     str-term-small-rw |    ok  1039  1045   6  0   0  1852.0    1675.1    0 |    ok  1039  1045   6  0   0  1848.0    1640.1    0 |
              __totals |    ok 27175 27321 146  0   0 87363.2  495665.2    3 |    ok 27173 27321 148  0   0 87362.1  489315.0    1 |
```

All differences are close to timeouts.

src/theory/strings/theory_strings_rewriter.cpp
```diff
@@ -2318,6 +2318,22 @@ Node TheoryStringsRewriter::rewriteIndexof( Node node ) {
           Node ret = nm->mkNode(kind::STRING_STRIDOF, nn, node[1], node[2]);
           return returnRewrite(node, ret, "idof-def-ctn");
         }
+
+        // Strip components from the beginning that are guaranteed not to match
+        if (stripConstantEndpoints(children0, children1, nb, ne, 1))
+        {
+          // str.indexof(str.++("AB", x, "C"), "C", 0) --->
+          // str.indexof(str.++(x, "C"), "C", 0)
```
ajreynol: 2+str.indexof

4tXJ7f: Good catch, fixed.

Issue #1881, opened 2018-05-05
Refactor strings theory


ajreynol:
This includes making a few things enums.

ajreynol:
Made this into a project, closing.

PR #2891, opened 2019-03-22
 Split regular expression solver

ajreynol:
This splits the regular expression specific portions of TheoryStrings to its own file and improves some documentation. It makes some functions in TheoryStrings public as a result and documents them.

Notice that I am intentionally not adding further documentation to the private block of RegExpSolver since I did not write this code (Tianyi Liang did) and I am not 100% sure what it does.  This code will hopefully be phased out with further improvements.

### DIFF
--- a/src/theory/strings/regexp_solver.h
+++ b/src/theory/strings/regexp_solver.h
@@ -69,7 +69,7 @@ private:
   Node getMembership( Node n, bool isPos, unsigned i );
   unsigned getNumMemberships( Node n, bool isPos );
   CVC4::String getHeadConst( Node x );
-  bool deriveRegExp( Node x, Node r, Node ant );
+  bool deriveRegExp( Node x, Node r, Node atom, std::vector< Node >& ant );
   Node getNormalSymRegExp(Node r, std::vector<Node> &nf_exp);
   //--------------------------------end for checkMemberships
   // regular expression memberships
--- a/src/theory/strings/regexp_solver.h
+++ b/src/theory/strings/regexp_solver.h
@@ -26,10 +26,10 @@
 namespace CVC4 {
 namespace theory {
 namespace strings {
-  
+
 class TheoryStrings;
 
-class RegExpSolver 
+class RegExpSolver
 {
   typedef context::CDList<Node> NodeList;
   typedef context::CDHashMap<Node, bool, NodeHashFunction> NodeBoolMap;
@@ -39,24 +39,25 @@ class RegExpSolver
 
  public:
   RegExpSolver(TheoryStrings& p, context::Context* c, context::UserContext* u);
-  ~RegExpSolver(){}
-  
+  ~RegExpSolver() {}
+
   /** add membership
-   * 
+   *
    * This informs this class that assertion is asserted in the current context.
    * We expect that assertion is a (possibly negated) regular expression
    * membership.
    */
   void addMembership(Node assertion);
   /** check
-   * 
+   *
    * Tells this solver to check whether the regular expressions asserted to it
    * are consistent. If they are not, then this class will call the
    * sendInference method of its parent TheoryString object, indicating that
    * it requires a conflict or lemma to be processed.
    */
   void check();
-private:
+
+ private:
   // Constants
   Node d_emptyString;
   Node d_emptyRegexp;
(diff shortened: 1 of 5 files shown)


ajreynol:
This is work towards #1881

src/theory/strings/theory_strings.cpp
```diff
@@ -258,6 +248,39 @@ Node TheoryStrings::getLength( Node t, std::vector< Node >& exp ) {
...
+        for (unsigned i = 0; i < x.getNumChildren(); i++)
+        {
+          Node nc = getNormalString(x[i], nf_exp);
+          vec_nodes.push_back(nc);
+        }
+        return mkConcat(vec_nodes);
+      }
+    }
+  }
+  return x;
```
4tXJ7f: Why are we returning `x` here and not the representative of `x`? (Just curious, the code is the same as in the original)

src/theory/strings/theory_strings.cpp
```diff
@@ -3497,7 +3544,6 @@ TheoryStrings::ProcessLoopResult TheoryStrings::processLoop(
                      nm->mkNode(kind::REGEXP_STAR,
                                 nm->mkNode(kind::STRING_TO_REGEXP, restr))));
       cc = cc == d_true ? conc2 : nm->mkNode(kind::AND, cc, conc2);
-      d_regexp_ant[conc2] = ant;
```
4tXJ7f: Why is it ok to remove this?

src/theory/strings/theory_strings.cpp
```diff
@@ -3547,11 +3593,6 @@ TheoryStrings::ProcessLoopResult TheoryStrings::processLoop(
     conc = nm->mkNode(kind::AND, vec_conc);
   }  // normal case
 
-  // set its antecedant to ant, to say when it is relevant
-  if (!str_in_re.isNull())
-  {
-    d_regexp_ant[str_in_re] = ant;
-  }
```
4tXJ7f: Ditto.

src/theory/strings/regexp_solver.h
```diff
@@ -0,0 +1,102 @@
...
+ **/
+
+#include "cvc4_private.h"
+
+#ifndef __CVC4__THEORY__STRINGS__REGEXP_SOLVER_H
+#define __CVC4__THEORY__STRINGS__REGEXP_SOLVER_H
+
+#include "context/cdhashset.h"
+#include "context/cdlist.h"
+#include "theory/strings/regexp_operation.h"
```
4tXJ7f: Missing `context/cdhashmap.h`, `context/context.h`, `expr/node.h`, `util/regexp.h`, `<map>`

src/theory/strings/theory_strings.cpp
ajreynol: I think the two are equivalent. If x has a representative that is distinct from x, then it exists in the equality engine, and hence should have a normal form. I'll add an assertion for this.

src/theory/strings/theory_strings.cpp
ajreynol: This is a very old optimization. It appears that this map is storing an alternative way of explaining a regular expression. For example, `x ++ "ab" = "ba" ++ x` explains `x in ("ba")* ++ "b"`.

When untangling the regular expression solver, I found this map would need to be accessed by both the core strings solver and the regular expression solver. 

Since this optimization is not very important (and its correctness is questionable to me), I decided now is a good time to remove it.

src/theory/strings/regexp_solver.h
4tXJ7f: `<map>` goes before the other includes :)

src/theory/strings/theory_strings.cpp
```diff
@@ -264,18 +264,18 @@ Node TheoryStrings::getNormalString(Node x, std::vector<Node>& nf_exp)
           << "Term: " << x << " has a normal form " << ret << std::endl;
       return ret;
     }
-    else
+    // if x does not have a normal form, then it should not occur in the
+    // equality engine and hence should be its own representative.
+    Assert( xr==x );
```
4tXJ7f: Apply clang-format.

src/theory/strings/theory_strings.cpp
4tXJ7f: That makes sense, thanks for the explanation.

PR #2901, opened 2019-03-26
Fix issues in cvc parser

ajreynol:
This reduces our number of warnings from 1322 to 860 and fixes major issues in Cvc.g for parsing relations. It fixes relations regressions which did not conform to the documented syntax (http://cvc4.cs.stanford.edu/wiki/Sets), because these lead to syntax errors in the corrected parser.

This is work towards #2525.


ajreynol:
The syntax has not changed. The parser is now more strict such that it only accepts the documented syntax.

4tXJ7f:
> The syntax has not changed. The parser is now more strict such that it only accepts the documented syntax.

Ok, thanks. I don't think we need to document it then.

PR #2907, opened 2019-03-29
Apply empty splits more aggressively in sets+cardinality

### LINKED ISSUE #2904: Wrong model on CVC4 1.6
Running on windows 10 64bit

cvc4-1.6-win64-opt.exe -m --smt cvc4_bug_example.txt

The following specification is unsatisfiable, but the solver says otherwise.
The model suggested as a satisfying model does not actually satisfy the model.
I got:
sat
(model
(define-fun n () Int 3)
(define-fun f () Int 1)
(define-fun m () Int 0)
(define-fun b () (Set Int) (as emptyset (Set Int)))
(define-fun c () (Set Int) (singleton 0))
(define-fun UNIVERALSET () (Set Int) (union (union (singleton 0) (singleton (- 1))) (singleton (- 2))))
)

But in that model, |(U/b) & (U/c)|=2, and as n=3 the assertion on line 25 does not hold.

[cvc4_bug_example.txt](https://github.com/CVC4/CVC4/files/3019191/cvc4_bug_example.txt)

ajreynol:
Fixes #2904.

The current implementation of sets+cardinality uses a saturation criterion that is optimized towards avoiding certain unnecessary splits. This criterion was too aggressive: in terms of the [IJCAR 2016](http://homepage.divms.uiowa.edu/~ajreynol/ijcar16b.pdf) algorithm, it did not apply "GUESS EMPTY SET" enough (fig 8 of that paper), and hence could saturate for unsatisfiable problems.

This corrects the issue by split with empty on all "leaves" of cardinality graphs. 

A diagram of the state of the solver that reported a false model is attached (found in a minimized version of the original). Solid lines denote edges of the cardinality graph, dotted lines denote equalities. Here, `u\c` was assigned { -1 } in the model. This leads to a conflict, since subsequently -1 must occur in u\c and u\b, violating u\b^u\c=empty. Instead, we require the solver to decide whether `u\c` (which is a leaf node modulo equality) is empty. This ensures that the conflict is discovered after witness the non-emptiness of `u\c` via SET DISEQUALITY (fig 3 [IJCAR 2016](http://homepage.divms.uiowa.edu/~ajreynol/ijcar16b.pdf)).

A85C2848-2ED7-4A0E-AC17-E9550859A056


src/theory/sets/theory_sets_private.cpp
```diff
@@ -1494,6 +1494,11 @@ void TheorySetsPrivate::checkNormalForm( Node eqc, std::vector< Node >& intro_se
         Assert( false );
       }
     }else{
+      // must ensure disequal from empty
+      if (!eqc.isConst() && !ee_areDisequal(eqc, emp_set))
+      {
+        split(eqc.eqNode(emp_set));
```
yoni206: Could it be that such a split is missing also in other branches of this code?

ajreynol:
The place I added it is exactly where we find a "leaf" node in the graph. I'm planning to document/refactor the entire implementation more soon (after this PR).

PR #2910, opened 2019-03-30
FP: Fix wrong model due to partial assignment

4tXJ7f:
For a simple query `(not (= (fp.isSubnormal x) false))`, we were getting
a wrong model. The issue was that `(sign x)` was not assigned a value
and did not appear in the shared terms. In
`TheoryFp::collectModelInfo()`, however, we generate an expression that
connects the components of `x` to `x`, which contains `(sign x)`. As a
result, the normalization while building a model did not result in a
constant. This commit fixes the issue by marking `(sign x)` (and
`(significand x)`) as assignable. Assignable terms can take any value
while building a model if there is no existing value.


ajreynol:
This looks good, although my main question is whether we need to make `(exponent x)` assignable as well.

src/theory/theory_model_builder.cpp
```diff
@@ -47,6 +47,14 @@ bool TheoryEngineModelBuilder::isAssignable(TNode n)
       return !n.getType().isFunction();
     }
   }
+  else if (n.getKind() == kind::FLOATINGPOINT_COMPONENT_SIGN
+           || n.getKind() == kind::FLOATINGPOINT_COMPONENT_SIGNIFICAND)
```
martin-cs: I don't think significand should be treated the same.  The auxiliary constraints that keep the components in sync should always mention it so the BV solver should always return a term for it.  Also it is not correct to simply pick any value.

martin-cs:
( There is still some discussion ongoing so don't merge this immediately! )

PR #2915, opened 2019-04-02
[BV] An option for SAT proof optimization

alex-ozdemir:
The option doesn't **do** anything yet, but exists.

Another PR will follow with the optimization machinery -- I split the changes to make reviewing easier.


src/options/bv_bitblast_mode.h
```diff
@@ -88,13 +88,35 @@ enum BvProofFormat
   BITVECTOR_PROOF_LRAT,
 };
 
+/**
+ * When the BV solver does eager bitblasting backed by CryptoMiniSat, proofs
```
mpreiner:
```suggestion
 * When the BV solver does eager bit-blasting backed by DRAT-producing SAT solvers, proofs
```

We will also support CaDiCaL in the future.

src/options/options_handler.cpp
```diff
@@ -1245,6 +1245,45 @@ theory::bv::BvProofFormat OptionsHandler::stringToBvProofFormat(
   }
 }
 
+const std::string OptionsHandler::s_bvOptimizeSatProofHelp =
+    "\
+Optimization levels currently supported by the --bv-optimize-sat-proof option:\n\
+\n\
+  none    : Do not optimize that SAT proof\n\
```
mpreiner:
```suggestion
  none    : Do not optimize the SAT proof\n\
```

alex-ozdemir:
I believe that I've applied all of your changes, @mpreiner. Let me know if anything else should be done!

PR #2916, opened 2019-04-02
SatClauseSetHashFunction

alex-ozdemir:
Called "SatClauseSetHashFunction" because it's a hash function for a SAT clause encoded as a set.

Added to the same file as SatLiteralHashFunction.

Some SAT proof optimization machinery needs to (temporarily) build a map which is keyed by clause, which is why this is needed.

I wish there was a way to lift a user-defined hash for `T` into a hash for `unordered_set<T>` using a standardized lift, but I don't know how to do so.


alex-ozdemir:
Whoops. Good catch!

Issue #2855, opened 2019-03-12
Fix QF_LRA proofs


4tXJ7f:
The following regressions fail with QF_LRA proofs:

- [ ] https://github.com/CVC4/CVC4/blob/master/test/regress/regress0/lemmas/sc_init_frame_gap.induction.smt
- [ ] https://github.com/CVC4/CVC4/blob/master/test/regress/regress1/lemmas/pursuit-safety-8.smt
- [ ] https://github.com/CVC4/CVC4/blob/master/test/regress/regress1/lemmas/simple_startup_9nodes.abstract.base.smt
- [ ] https://github.com/CVC4/CVC4/blob/master/test/regress/regress2/arith/pursuit-safety-11.smt
- [ ] https://github.com/CVC4/CVC4/blob/master/test/regress/regress2/arith/pursuit-safety-12.smt
- [x]  https://github.com/CVC4/CVC4/blob/master/test/regress/regress0/uflra/constants0.smt
- [ ]  https://github.com/CVC4/CVC4/blob/master/test/regress/regress0/uflra/pb_real_10_0200_10_22.smt
- [ ]  https://github.com/CVC4/CVC4/blob/master/test/regress/regress0/uflra/pb_real_10_0200_10_26.smt

Our support for QF_LRA proofs is limited but we should output trust statements instead of failing the proof check.

4tXJ7f:
Found some more tests that fail for QF_UFLRA.

alex-ozdemir:
Following up on this, I've found a few bugs

## ITE translation bug

This is actually a bug in the both `TheoryProof` and `ArithProof`. `TheoryProof` will convert some `Bool`s to `formula`s in ITEs. It does this with the support of the `printsAsBool` function, which should be subclass defined. `ArithProof` neglected to define it, and `TheoryProof` was doing something slightly wrong anyway.

Tests

* [https://github.com/CVC4/CVC4/blob/master/test/regress/regress1/lemmas/simple_startup_9nodes.abstract.base.smtv1.smt2](https://github.com/CVC4/CVC4/blob/master/test/regress/regress1/lemmas/simple_startup_9nodes.abstract.base.smtv1.smt2)

## Integer Reasoning 

Sometimes we do integer reasoning, even if the problem in LRA. The current proof system produces invalid proofs in this case.

Tests

[...]

PR #2928, opened 2019-04-03
Enable proof checking for QF_LRA benchmarks

4tXJ7f:
Due to issues in the current proof code, this commit also disables proof
checking for five QF_LRA benchmarks (see issue #2855). This commit also
fixes issues related to comparing `InfoLogic`s with/without higher-order
or cardinality constraints.


src/smt/smt_engine.cpp
```diff
@@ -4431,7 +4431,7 @@ void SmtEngine::checkProof()
   if (!(
           // Pure logics
           logicString == "QF_UF" || logicString == "QF_AX"
-          || logicString == "QF_BV" ||
+          || logicString == "QF_BV" || logicString == "QF_LRA" ||
           // Non-pure logics
```
yoni206: Any reason to not include `QF_UFLRA` and other combinations?

4tXJ7f: Good point, I've enabled it now and fixed some issues with comparing `LogicInfo`s in the process.

4tXJ7f:
@ajreynol I've added you to the PR if you are interested to look at the changes in `LogicInfo` related to comparing logics with/without cardinalities/higher-order constraints.

4tXJ7f:
I've updated #2855 to include some additional failing benchmarks.

PR #2929, opened 2019-04-03
Eliminate negative constant coefficients in div/mod

### LINKED ISSUE #1399: Suspiciously hard LIA benchmark
Hello, for the following benchmark CVC4 seems to run indefinitely and consumes increasingly more memory.

```
(set-logic LIA)

(define-fun findIdx ((y1 Int)(y2 Int)(k1 Int)) Int (div (* (- 7) (mod y1 (- 5))) 2))
(declare-fun x1 () Int)
(declare-fun x2 () Int)
(declare-fun k () Int)

(assert (not (and (=> (< x1 x2) (=> (< k x1) (= (findIdx x1 x2 k) 0)))
  (=> (< x1 x2) (=> (> k x2) (= (findIdx x1 x2 k) 2)))
  (=> (< x1 x2) (=> (and (> k x1) (< k x2)) (= (findIdx x1 x2 k) 1))))))
(check-sat)
(get-value (x1 x2 k))
```

ajreynol:
Fixes #1399.


4tXJ7f:
This seems like reasonable rewrites to do. Is there a concern that this breaks assumptions about how arithmetic terms are normalized? If not, I'd be in favor of merging this.

Issue #2183, opened 2018-07-20
Potentially wrong result when down-casting floating-point numbers using RNA rounding


4tXJ7f:
Z3 and CVC4 report a different result on the following benchmark:
```
(set-logic QF_FP)
(declare-fun r () (_ FloatingPoint 5 9))
(assert (= (fp #b1 #b00000 #xff) ((_ to_fp 5 9) RNA (fp #b1 #b00000 #b1111111110))))
(check-sat)
```

Z3 reports `sat` while CVC4 reports `unsat`. My MPFR wrapper agrees with Z3, so chances that this is a bug in CVC4/symfpu are non-zero.

4tXJ7f:
Interestingly, it looks like Z3 4.7.1 agrees with CVC4 (the `sat` answer was the [online](https://rise4fun.com/z3) version). It also makes more sense given the input but I am leaving this issue open to confirm that CVC4's answer of `(fp #b1 #b00001 #x00000000)` for the conversion is indeed correct.

martin-cs:
It should be UNSAT because the value does indeed round to `(fp #b1 #b00001 #x00000000)`.  @florianschanda did report some bugs in Z3, especially around RNA so there change is likely a fix of that issue.  However I added it as a regression test : https://github.com/CVC4/CVC4/pull/3008

4tXJ7f:
For the record, I've tested it again with a recent Z3 version (commit bd46c52f959e01ac7c65bbf634994c6e9649f927) and it answered `unsat`.

4tXJ7f:
Closing the issue since we've determined that CVC4's answer was correct.

Issue #2181, opened 2018-07-19
Failing precondition when dividing 5-bit significand/3-bit exponent floats


4tXJ7f:
The following input causes a precondition to fail:
```
(set-logic QF_FP)
(declare-fun r () (_ FloatingPoint 3 5))
(assert (= r (fp.div RNE (fp #b0 #b000 #b0000) (fp #b0 #b000 #b0000))))
(check-sat)
```

Error:
```
(error "Assertion failure.
static void CVC4::symfpuLiteral::traits::precondition(const prop&)
../../../../../src/util/floatingpoint.cpp:444:

  p
")
```

Note: the error seems to be connected to the bit-width. The same division works fine for single precision floats and the precondition also fails with arguments other than `(fp #b0 #b000 #b0000)`.

PR #2931, opened 2019-04-03
Ignoring FP benchmarks with "unsafe" sizes unless option

HanielB:
Conservative solution until issues #2181 and #2183 are solved


ajreynol:
One recommendation for the option name is `--fp-exp`. This follows the convention of `--strings-exp` for "experimental".

src/theory/fp/theory_fp.cpp
```diff
@@ -876,7 +879,23 @@ bool TheoryFp::isRegistered(TNode node) {
   return !(d_registeredTerms.find(node) == d_registeredTerms.end());
 }
 
-void TheoryFp::preRegisterTerm(TNode node) {
+void TheoryFp::preRegisterTerm(TNode node)
+{
+#ifdef CVC4_USE_SYMFPU
```
4tXJ7f: Suggestion: I always prefer to use `Configuration::isBuiltWithSymFPU()` if possible to avoid as many preprocessing directives as possible.

4tXJ7f:
> This follows the convention of --strings-exp for "experimental".

TIL

HanielB:
> > This follows the convention of --strings-exp for "experimental".
> 
> TIL

TIL what "TIL" is

PR #2937, opened 2019-04-04
Cache explanations in the equality engine

ajreynol:
Leads to solving a few hard QF_UF and TPTP benchmarks while adding a small overhead. @HanielB has numbers on this.


ajreynol:
I've updated the branch to guard debug messages that were absolutely killing our performance on hard TPTP benchmarks where proofs are enabled, since the entire proof was being traversed (for the debug_print) even when the debug message was not enabled.

src/theory/uf/equality_engine.cpp
```diff
@@ -1024,15 +1036,51 @@ void EqualityEngine::explainPredicate(TNode p, bool polarity,
...
+      {
+        // We may have cached null in its place, create the trivial proof now.
+        Assert(d_nodes[t1Id] == d_nodes[t2Id]);
+        Assert(eqp->d_id == MERGED_THROUGH_REFLEXIVITY);
+        eqp->d_node = d_nodes[t1Id];
+      }
+    }
+    return;
+  }
+  cache[cacheKey] = eqp;
```
barrettcw: Setting the cache here before filling in the proof seems a little dangerous.  Can you double-check that there's no path where you end up caching an incomplete or invalid proof?  Alternatively - and to make sure we don't make mistakes in future revisions - maybe we should wait until the end of the function to set the cache.

ajreynol:
Can you post the benchmark we lose in QF_ABV with the longer timeout? And perhaps run it with a longer timeout to see when we solve it?

HanielB:
I updated my previous post with the diff in QF_ABV and checking with a longer timeout. Since I was testing on faster hardware, though, both times (master and branch) went down. So in the original setting the banch was probably close to solving that benchmark.

src/theory/uf/equality_engine.cpp
ajreynol: I've checked this is the case. At a higher level, the contract of the function says that `eqp` must always be filled in with a proper proof if it is non-null, so whatever case where eqp was incomplete or invalid, it would be breaking the contract of the overall function.

ajreynol:
Based on the discussion in the meeting, it appears this one is ready to proceed? Any further review comments?

PR #2939, opened 2019-04-05
Fail fast strategy for propagating instances

ajreynol:
This revisits propagating instances in quant_conflict_find.h/cpp.

In particular, it implements a "fail fast" strategy that recognizes eagerly when an instantiation is guaranteed not be propagating. It also adds explanations to TermDb::evaluateTerm.

This along with the equality engine caching leads to +35-4 on TPTP:
```
                           CVC4-031419_d    7957   8753.08      73   5.47917      8030     22011
                       CVC4-040319        7988   8759.16      73   5.38905      8061     22011
```


src/theory/quantifiers/term_database.cpp
```diff
@@ -509,88 +509,194 @@ bool TermDb::inRelevantDomain( TNode f, unsigned i, TNode r ) {
...
+      // if we short circuited
+      if (computeExp)
+      {
+        exp.clear();
+        exp.insert(exp.end(), tempExp.begin(), tempExp.end());
+      }
+    }
+    else
+    {
+      TNode f = getMatchOperator(n);
```
HanielB: say what `f` corresponds to (it'd clarify the test below)

src/theory/quantifiers/term_database.cpp
```diff
@@ -509,88 +509,194 @@ bool TermDb::inRelevantDomain( TNode f, unsigned i, TNode r ) {
...
+            }
+          }
+          ret = qy->getRepresentative(nn);
+          Trace("term-db-eval") << "return rep" << std::endl;
+          ret_set = true;
+          reqHasTerm = false;
+          Assert(!ret.isNull());
+          if (computeExp)
+          {
+            if (n != ret)
```
HanielB: join with above

src/theory/quantifiers/term_database.cpp
```diff
@@ -509,88 +509,194 @@ bool TermDb::inRelevantDomain( TNode f, unsigned i, TNode r ) {
...
-            if( ret.getKind()==kind::EQUAL || ret.getKind()==kind::GEQ ){
-              for( unsigned j=0; j<2; j++ ){
-                std::pair<bool, Node> et = d_quantEngine->getTheoryEngine()->entailmentCheck(THEORY_OF_TYPE_BASED, j==0 ? ret : ret.negate() );
-                if( et.first ){
-                  ret = j==0 ? d_true : d_false;
-                  break;
+        }
+        if (useEntailmentTests)
+        {
+          if (ret.getKind() == EQUAL || ret.getKind() == GEQ)
```
HanielB: join with the above

src/theory/quantifiers/term_database.cpp
```diff
@@ -509,88 +509,194 @@ bool TermDb::inRelevantDomain( TNode f, unsigned i, TNode r ) {
...
   }
-  Trace("term-db-eval") << "evaluated term : " << n << ", got : " << ret << std::endl;
+  // must have the term
+  if (reqHasTerm && !ret.isNull())
+  {
+    Kind k = ret.getKind();
+    if (k != OR && k != AND && k != EQUAL && k != ITE && k != NOT
+        && k != FORALL)
+    {
+      if (!qy->hasTerm(ret))
```
HanielB: join with the above

src/theory/quantifiers/quant_conflict_find.cpp
```diff
@@ -615,27 +620,6 @@ bool QuantInfo::isTConstraintSpurious( QuantConflictFind * p, std::vector< Node
...
-  }else if( n.getKind()==NOT || n.getKind()==AND || n.getKind()==OR || n.getKind()==EQUAL || n.getKind()==ITE || 
-            ( n.getKind()==EQUAL && n[0].getType().isBoolean() ) ){
-    for( unsigned i=0; i<n.getNumChildren(); i++ ){
-      if( !isPropagatingInstance( p, n[i] ) ){
-        return false;
-      }
-    }
-    return true;
-  }else{
-    if( p->getEqualityEngine()->hasTerm( n ) || isGroundSubterm( n ) ){
```
HanielB: removing the test of `isGroundSubterm( n )` (and the auxiliary information to set and test this) is what leads to the performance losses, isn't it? It does not seem to simplify things too much. I think we should keep this for now and only remove it when we add an alternative (and cleaner) way of dealing with these cases.

src/theory/quantifiers/term_database.cpp
ajreynol: I prefer these separate.

src/theory/quantifiers/term_database.cpp
ajreynol: I prefer these separate.

src/theory/quantifiers/quant_conflict_find.cpp
ajreynol: There are a handful of sledgehammer examples that have axioms forall x : U. x = f where f is a ground term. It was by chance that this hack led to solving these benchmarks. In general, this is a bad idea.

The proper way to handle this is to fix our trigger selection to make x a "variable trigger" in this case.

src/theory/quantifiers/term_database.cpp
ajreynol: I prefer these separate.

src/theory/quantifiers/quant_conflict_find.cpp
HanielB: Why is it a bad idea in general?

ajreynol: It's the same reason why we don't register terms that occur in quantifier bodies for E-matching.

HanielB: You tested and confirmed this is a bad idea? I had an option to add ground terms from quantified formulas to the term index used for CCFV (for both conflict and trigger based instantiations), which was very helpful sometimes, but I don't remember how it was in general.

ajreynol: Yes, it is a bad idea. There is no need in retesting this. 
Let's focus on making the code do the right thing.

PR #2942, opened 2019-04-08
fix copyright year in configuration file

HanielB:


mpreiner:
We should try to automate this with the build system for later releases.

PR #2957, opened 2019-04-17
Add rewrite for splitting equalities

4tXJ7f:
This commit adds an extended rewrite that splits a given string equality
into multiple equalities if it can find prefixes of equal lengths.

For example: `(= (str.++ x "AB" z) (str.++ "A" x y)) ---> (and (=
(str.++ x "A") (str.++ "A" x)) (= (str.++ "B" z) y))`


4tXJ7f:
The results on the APLAS benchmarks are pretty promising:
```
         DIRECTORY |  Tot |  Slvd |  Sat |  Unsat |  Unk |   To |  Mo |  Err |   Cpu[s] |   Mem[MB] |
 flat_ext_rew_prep |  600 |   359 |  142 |    217 |    0 |  241 |   0 |    0 |  74840.6 |  232369.9 |
flat_split_eq_rew2 |  600 |   440 |  142 |    298 |    0 |  160 |   0 |    0 |  48110.8 |  188712.7 |
```
`flat_ext_rew_prep` is master with the extended rewriting preprocessing pass. `flat_split_eq_rew2` is this PR with the extended rewriting preprocessing pass.

Performance otherwise is not affected much:
```
             DIRECTORY |                           qf_slia_ext_rew_prep |                          qf_slia_split_eq_rew2 |
             Benchmark |  Stat  Slvd   Tot  To Mo Err   Cpu[s]  Mem[MB] |  Stat  Slvd   Tot  To Mo Err   Cpu[s]  Mem[MB] |
     20180523-Reynolds |    ok 25167 25421 254  0   0 132823.1 503931.2 |    ok 25163 25421 258  0   0 133104.6 505515.0 |
 appscan-pisa-smtlib25 |    ok    20    20   0  0   0    185.0    316.6 |    ok    20    20   0  0   0    191.6    315.3 |
     str-pred-small-rw |    ok   822   835  13  0   0   3903.3   3102.0 |    ok   822   835  13  0   0   3902.7   3072.3 |
     str-term-small-rw |    ok  1039  1045   6  0   0   1853.2   1644.9 |    ok  1039  1045   6  0   0   1855.4   1640.2 |
              __totals |    ok 27048 27321 273  0   0 138764.6 508994.7 |    ok 27044 27321 277  0   0 139054.3 510542.8 |
```

All four lost instances are close to timeout.

test/unit/theory/theory_strings_rewriter_white.h
```diff
@@ -1300,6 +1303,49 @@ class TheoryStringsRewriterWhite : public CxxTest::TestSuite
...
+                             d_nm->mkNode(kind::STRING_CONCAT, b, x, y),
+                             d_nm->mkNode(kind::STRING_CONCAT, x, aaa, z));
+      sameNormalForm(eq, f);
+    }
+
+    {
+      Node xrepl = d_nm->mkNode(kind::STRING_STRREPL, x, a, b);
+
+      // Same normal form for:
+      //
```
ajreynol: Could you make this example even trickier: replace one of the str.replace( x, "A", "B" ) with x ?

src/theory/strings/theory_strings_rewriter.cpp
```diff
@@ -592,6 +592,106 @@ Node TheoryStringsRewriter::rewriteStrEqualityExt(Node node)
...
+                             mkConcat(STRING_CONCAT, pfxv0)
+                                 .eqNode(mkConcat(STRING_CONCAT, sfxv1)));
+              return returnRewrite(node, ret, "split-eq-strip-l");
+            }
+
+            // If the prefix of the left-hand side is (strictly) longer than
+            // the prefix of the right-hand side, then we don't need to check
+            // that right-hand prefix for future left-hand prefixes anymore
+            // (since they are increasing in length)
+            startRhs = j + 1;
```
ajreynol: This is pretty cool, so the overall run time is linear wrt the number of string components in LHS+RHS?

4tXJ7f: In the worst case, it can still be quadratic if we can't show any of the length entailments. If we can, however, then the number of entailment checks should be linear w.r.t. the number of string components in LHS+RHS.

PR #2964, opened 2019-04-18
 Less aggressive caching in equality engine when proofs are enabled

ajreynol:
This fixes the nightlies.

The recent caching in getExplanation #2937 led to a failure for UF proof checking. The issue is that a proof of t1 = t2 could be cached as a proof for t2 = t1. The proof code in src/proof/uf_proof.cpp is unfortunately not robust to this.

The solution is to make the caching order-dependent when proofs are enabled. We should still maintain most of the benefits of caching when proofs are enabled (we still expect cache hits when the order of equalities is correct). 

The performance of this method when proofs are disabled is unchanged with this PR.

If src/proof/uf_proof.cpp was made more robust to ordering in proofs, then we could make the caching more aggressive in this call, but I consider this to be low priority.

### DIFF
--- a/src/theory/uf/equality_engine.cpp
+++ b/src/theory/uf/equality_engine.cpp
@@ -1069,7 +1069,9 @@ void EqualityEngine::getExplanation(
   else
   {
     // If proofs are enabled, note that proofs are sensitive to the order of t1
-    // and t2, so we don't sort the ids in this case.
+    // and t2, so we don't sort the ids in this case. Depending on how issue
+    // #2965 is resolved, we may be able to revisit this, if it is the case
+    // that proof/uf_proof.h,cpp is robust to equality ordering.
     cacheKey = std::pair<EqualityNodeId, EqualityNodeId>(t1Id, t2Id);
     it = cache.find(cacheKey);
     if (it != cache.end())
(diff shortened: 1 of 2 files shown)


HanielB:
Can you open an issue about the proof code being robust to the order of the equalities?

src/theory/uf/equality_engine.cpp
```diff
@@ -1052,23 +1052,35 @@ void EqualityEngine::getExplanation(
...
+    if (it != cache.end())
+    {
+      return;
+    }
+  }
+  else
+  {
+    // If proofs are enabled, note that proofs are sensitive to the order of t1
+    // and t2, so we don't sort the ids in this case.
+    cacheKey = std::pair<EqualityNodeId, EqualityNodeId>(t1Id, t2Id);
```
barrettcw: Can we add a TODO here with the issue number for making proofs robust to order?

ajreynol:
Thinking about this more, its not clear 

As I recall, we have debated in the past whether the equality engine's proof producing code should actually take more responsibility in ensuring that its proofs are more coherent.

ajreynol:
Thinking about it more, I don't think we want to make things even more difficult on the proof printing code (we have mentioned in the past that the equality engine proof production could possibly take more responsibilty for making its proofs coherent). Thus, the code in the equality engine for proof production should generate properly oriented equalities.  I've opened #2965 for revisiting the proof printing code for UF.

barrettcw:
Let's discuss this at next CVC4 meeting.  If you add this discussion point to the issue, I'm fine with approving this PR.

ajreynol:
I've also added a comment in the equality engine code.

src/theory/uf/equality_engine.cpp
4tXJ7f: @ajreynol It looks like you updated the comment after @barrettcw's comment and it looks good to me but could you add an explicit `TODO` just to make it grep-able?

PR #2974, opened 2019-04-24
Do not use __ prefix for header guards.

mpreiner:
Fixes 2887.


elfring:
I find these include guards still questionable.
Please [reconsider the usage of double underscores in affected identifiers](https://stackoverflow.com/questions/228783/what-are-the-rules-about-using-an-underscore-in-a-c-identifier#answer-228797 "What are the rules about using an underscore in a C++ identifier?") (not only the leading ones).

Issue #1428, opened 2017-12-05
Revisit whether expandDefinitions is necessary


ajreynol:
It seems unnecessary that we have an expand definitions pass in SMT engine.  If we do substitutions + beta-reductions in the rewriter, this should suffice.

We also maybe don't need an APPLY kind.

ajreynol:
Fixed in #2976.

PR #2976, opened 2019-04-25
Eliminate APPLY kind

ajreynol:
Previously, an APPLY kind was used whose purpose was to mark defined functions during parsing.  This design choice has led to a number of complications involving which kind (APPLY or APPLY_UF) should be used/checked in various places.  

This PR eliminates the APPLY kind and replaces all of its previous uses with APPLY_UF. This means we now have a uniform representation (APPLY_UF) for applications of (defined or undefined) uninterpreted functions.  The fixes required:
- The parser no longer automatically rewrites defined functions from f to (APPLY f).
- SmtEngine::expandDefinitions was changed so that defined constants are themselves expanded. This was not necessary before since it appears it was assumed that all definition expansions were limited to APPLY/APPLY_UF,
- SmtEngine::expandDefinitions no longer has to special case on options where function definitions could be later added. Instead, heads of applications of APPLY_UF are always checked whether they are defined functions.
- A fix for sygus constructors whose operator was a defined constant.

One consequence is that users can now ask for models for defined functions (see regression).  Previously, a defined function f would be erroneously turned into the expression (APPLY f) by the parser, which would then lead to a type error if it was not fully applied.

This PR also eliminates an unused FUNCTION kind that appears to be unused.

This fixes the question I raised in #1428, we can close this issue now.


ajreynol:
I would prefer APPLY_UF since it is more descriptive. All uninterpreted functions have an implied interpretation, regardless of whether they have a concrete definition.

PR #2977, opened 2019-04-25
New C++ API: Clean up API: mkVar vs mkConst vs mkBoundVar.

aniemetz:
This cleans up naming of API functions to create first-order constants and variables.
`mkVar`             -> `mkConst`
`mkBoundVar` -> `mkVar`
`declareConst` is redundant (= `mkConst`) and thus, in an effort to avoid redundancy, removed.

Note that we want to avoid redundancy in order to reduce code duplication and maintenance
overhead (we do not allow nested API calls, since this is problematic when tracing API calls).

@makaimann

### DIFF
--- a/src/api/cvc4cpp.cpp
+++ b/src/api/cvc4cpp.cpp
@@ -1970,7 +1970,7 @@ Term Solver::mkPi() const
 }
 
 template <typename T>
-Term Solver::mkConstHelper(T t) const
+Term Solver::mkValHelper(T t) const
 {
   try
   {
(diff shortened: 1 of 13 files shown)


src/api/cvc4cpp.cpp
```diff
@@ -2436,7 +2439,7 @@ void Solver::checkMkTerm(Kind kind, uint32_t nchildren) const
       mk == kind::metakind::PARAMETERIZED || mk == kind::metakind::OPERATOR,
       kind)
       << "Only operator-style terms are created with mkTerm(), "
-         "to create variables and constants see mkVar(), mkBoundVar(), "
+         "to create variables and constants see mkConst(), mkVar(), "
```
4tXJ7f: This message says `mkConst()` twice.

PR #2979, opened 2019-04-27
Optimization for evaluation with unfolding

### LINKED ISSUE #2972: Enable fast evaluator for problems without PBE solver
The example https://github.com/zhanghongce/inv-syn-examples/blob/master/aes-sygus/wrapper.smt2
is much faster when using `--sygus-pbe` despite the PBE solver not being the required solver (due to a lack of ITE in the grammar).  Regardless, we should be using fast evaluation in the standard CEGIS loop when it is possible. This mainly applies to PBE conjectures without ITEs in the grammar.

This was brought up on #2911.

ajreynol:
This PR makes it so that the evaluator is called on subterms of CEGIS lemmas that are evaluatable.
This fixes #2972.

### DIFF
--- a/src/theory/quantifiers/sygus/term_database_sygus.cpp
+++ b/src/theory/quantifiers/sygus/term_database_sygus.cpp
@@ -1633,6 +1633,31 @@ Node TermDbSygus::evaluateWithUnfolding(
     while (ret.getKind() == DT_SYGUS_EVAL
            && ret[0].getKind() == APPLY_CONSTRUCTOR)
     {
+      if( ret==n && ret[0].isConst() )
+      {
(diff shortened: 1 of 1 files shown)


src/theory/quantifiers/sygus/term_database_sygus.cpp
```diff
@@ -1633,6 +1633,31 @@ Node TermDbSygus::evaluateWithUnfolding(
     while (ret.getKind() == DT_SYGUS_EVAL
            && ret[0].getKind() == APPLY_CONSTRUCTOR)
     {
+      if( ret==n && ret[0].isConst() )
```
4tXJ7f: Please apply Clang-Format before merging.

PR #2980, opened 2019-04-27
Add support for re.all

4tXJ7f:


src/parser/smt2/smt2.cpp
```diff
@@ -293,6 +293,12 @@ void Smt2::addTheory(Theory theory) {
     defineType("String", getExprManager()->stringType());
     defineType("RegLan", getExprManager()->regExpType());
     defineType("Int", getExprManager()->integerType());
+
+    defineVar("re.all",
```
ajreynol: Should this be a part of addStringOperators?

4tXJ7f: Yeah, I suppose that would be better.

PR #2983, opened 2019-04-30
Fix concat-find regexp elimination

### LINKED ISSUE #2982: Incorrect model with --re-elim-agg
On the following benchmark, we answer `sat` with an incorrect model when using `--re-elim-agg`:
```
(set-logic QF_S)

(declare-fun var_0 () String)
(declare-fun var_1 () String)
(declare-fun var_2 () String)
(declare-fun var_3 () String)
(declare-fun var_4 () String)
(declare-fun var_5 () String)
(declare-fun var_6 () String)
(declare-fun var_7 () String)
(declare-fun var_8 () String)
(declare-fun var_9 () String)
(declare-fun var_10 () String)
(declare-fun var_11 () String)
(declare-fun var_12 () String)

(assert (str.in.re (str.++ var_7 "z" var_7 ) (re.* (str.to.re "z"))))
(assert (str.in.re var_7 (re.* (re.range "a" "u"))))
(assert (not (str.in.re (str.++ "a" var_7 "z" "a" var_7 ) (re.++ (re.* (re.union (str.to.re "z") (re.++ (str.to.re "a") (re.++ (re.* (str.to.re "a")) (str.to.re "z"))))) (re.++ (str.to.re "a") (re.* (str.to.re "a")))))))
(assert (and (<= (str.len var_7) 0 ) (<= 0  (str.len var_7))))
(check-sat)
```

We answer `unsat` without `--re-elim-agg`. The regular expression eliminations look ok to me. Note that some assertions fail but they seem bogus to me and I've fixed them [here](https://github.com/4tXJ7f/CVC4/commit/2bde55cd2a435a58118f2a6e6392e88753f9c653).

4tXJ7f:
Fixes #2982. The `concat-find` regexp elimination has an off-by-one
error in the upper bound of the quantifier. For example, for `(str.in.re
x (re.++ R1 "A" R2))`, the upper bound for the quantifier variable in
the eliminated term was `(< k (- (str.len x) 1))`. However, it should be
`(<= k (- (str.len x) 1))` because we have to account for the case where
`"A"` is at the end of the string, i.e. at position `(- (str.len x) 1)`.
This commit fixes the issue and moves two assertions that were too
aggressive.


ajreynol:
Can you update your comment above to `(<= k (- (str.len x) 1))`?

4tXJ7f:
Good catch, not sure what happened there. Done.

PR #2986, opened 2019-05-01
 Use total versions of div/mod in re-elim-agg

ajreynol:
We were using partial versions of div/mod in --re-elim-agg when total versions would suffice since it is division by constant. This PR changes them to total.


src/theory/strings/regexp_elim.cpp
```diff
@@ -547,19 +547,23 @@ Node RegExpElimination::eliminateStar(Node atom)
         Node lens = nm->mkNode(STRING_LENGTH, s);
         lens = Rewriter::rewrite(lens);
         Assert(lens.isConst());
+        Assert(lens.getConst<Rational>().sgn()>0);
```
4tXJ7f: Please apply clang-format before merging.

PR #2988, opened 2019-05-01
Simple optimizations to core strings theory and fix for extf handling

ajreynol:
This PR:
- Eliminates two kinds of redundant lemmas that were being generated by the core string solver. The first involves length splits for proxy variables that map to constants or concatenation terms. The value of these variables is already implied by an equality, so splitting on whether it is zero is redundant. The second involves a simple check for the LEN-NORM inference rule that ensures the lemma we generate has not already been generated.  These two lead to significant reduction in the number of lemmas generated.
- Make the string solver handle extended functions even if they are eagerly eliminated (--no-strings-lazy-pp). One reason for this is that some RE inferences introduce extended functions during the search.  This means that --no-strings-lazy-pp could be incomplete when the problem also involves regular expressions.  This PR eliminates this possibility.

### DIFF
--- a/src/theory/strings/theory_strings.cpp
+++ b/src/theory/strings/theory_strings.cpp
@@ -146,16 +146,16 @@ TheoryStrings::TheoryStrings(context::Context* c,
   d_equalityEngine.addFunctionKind(kind::STRING_CONCAT);
   d_equalityEngine.addFunctionKind(kind::STRING_IN_REGEXP);
   d_equalityEngine.addFunctionKind(kind::STRING_CODE);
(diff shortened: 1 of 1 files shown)


ajreynol:
+36-2 on CMU benchmarks:
```
                CVC4-050119-so_strings_f    3828   636.759   20934   22458.6     24762     25421
                  CVC4050119-2_strings_f    3818   723.671   20911   23971.7     24729     25421
```

ajreynol:
Quite consistently 12% faster:

ajreynol:
cmu-050219

PR #2992, opened 2019-05-08
Fixing Issues in LRA Proofs

yoni206:
This PR fixes the following issues:

1. Some applications of Farkas lemma were not valid ones, and so the corresponding proofs failed to check. The current solution is to check whether the Farkas coefficients that are stored in the proof are valid - if they are, we print the proof step. Otherwise, we print a hole in the proof. This solves #2855 .
2. Some LRA proofs included integer terms, that caused a failure while checking. Now we only allow real terms in LRA proofs.
3. ite terms were printed in a wrong way in proofs, since they should be printed differently if the second and third arguments are booleans. This is solved by checking their type before printing.

This PR depends on another PR that @4tXJ7f plans to submit, which is the reason for the "do not merge" label.


src/proof/arith_proof.cpp
```diff
@@ -24,15 +24,84 @@
...
+    else
+    {
+      result = false;
+      break;
+    }
+  }
+  return result;
+}
+
+// Verify that that the farkas coefficients indeed create a contradiction
```
4tXJ7f:
```suggestion
// Verify that the Farkas coefficients indeed create a contradiction
```

src/proof/arith_proof.cpp
```diff
@@ -24,15 +24,84 @@
...
 
 namespace CVC4 {
+using namespace CVC4::theory;
 
 inline static Node eqNode(TNode n1, TNode n2) {
   return NodeManager::currentNM()->mkNode(kind::EQUAL, n1, n2);
 }
 
+// Proofs are supported only for contraints of the form __ <= __ .
+inline static bool allChildrenGeq(const Node& conflict)
```
4tXJ7f:
```suggestion
static bool allChildrenGeq(const Node& conflict)
```

src/proof/arith_proof.cpp
```diff
@@ -24,15 +24,84 @@
...
+{
+  bool result = true;
+  size_t nChildren = conflict.getNumChildren();
+  for (size_t i = 0; i < nChildren; i++)
```
4tXJ7f: Why not `for (const Node& n : conflict)`?

src/proof/arith_proof.cpp
```diff
@@ -24,15 +24,84 @@
...
+  {
+    const Node& n = conflict[i];
+    const Node& nonneg = n.getKind() == kind::NOT ? n[0] : n;
+    if (nonneg.getKind() == kind::GEQ)
+    {
+      continue;
```
4tXJ7f: This branch is not needed, just do:

```
if (nonneg.getKind() != kind::GEQ)
{
  result = false;
  break;
}
```

src/proof/theory_proof.cpp
```diff
@@ -879,17 +879,27 @@ void LFSCTheoryProofEngine::printCoreTerm(Expr term, std::ostream& os, const Pro
...
+    if (isBooleanTerm)
+    {
+      os << "(ifte ";
+    }
+    else
+    {
+      os << "(ite _ ";
+    }
+    Assert(term[0].getType().isBoolean());
+    if (printsAsBool(term[0])) os << "(p_app ";
```
4tXJ7f: Use braces around branch.

src/proof/theory_proof.cpp
```diff
@@ -879,17 +879,27 @@ void LFSCTheoryProofEngine::printCoreTerm(Expr term, std::ostream& os, const Pro
...
     printBoundTerm(term[0], os, map);
-    if (booleanCase && printsAsBool(term[0])) os << ")";
-
-    os << " ";
-    printBoundTerm(term[1], os, map);
+    if (printsAsBool(term[0])) os << ")";
```
4tXJ7f: Use braces around branch.

src/proof/arith_proof.cpp
```diff
@@ -24,15 +24,84 @@
...
+    {
+      result = false;
+      break;
+    }
+  }
+  return result;
+}
+
+// Verify that that the farkas coefficients indeed create a contradiction
+inline static bool hasContradiction(
```
4tXJ7f:
```suggestion
static bool hasContradiction(
```

(may have to reformat)

src/proof/arith_proof.cpp
```diff
@@ -1004,14 +1065,15 @@ void LFSCArithProof::printTheoryLemmaProof(std::vector<Expr>& lemma,
                    return NodeManager::currentNM()->fromExpr(e).negate();
                  });
 
+  // Get farkas coefficients & literal order
```
4tXJ7f:
```suggestion
  // Get Farkas coefficients & literal order
```

src/proof/arith_proof.cpp
```diff
@@ -1068,90 +1130,91 @@ void LFSCArithProof::printTheoryLemmaProof(std::vector<Expr>& lemma,
...
+       *       / \
+       *      *   +               2
+       *         / \
+       *        *  ...            i
+       *             \
+       *              +           n-1
+       *             / \
+       *            *   (0 >= 0)
+       *
+       * Where each * is a linearized antecedant being scaled by a farkas
```
4tXJ7f:
```suggestion
       * Where each * is a linearized antecedant being scaled by a Farkas
```

src/proof/arith_proof.cpp
```diff
@@ -24,15 +24,84 @@
...
+    const Node& lem = conflict[i];
+    Rational c = (*farkasCoefficients)[i].abs();
+    // if lem has the form ~(x <= y), this is equivalent to x > y.
+    // Hence the Farkas coefficient is multiplied by -1, to keep the inequality
+    // in the right direction. Also, the resulting inequality will be strict.
+    if (lem.getKind() == kind::NOT)
+    {
+      hasStrictIneq = true;
+      currentLeft = lem[0][0];
+      Assert(lem[0].getKind() == kind::GEQ);
```
4tXJ7f: It would be more natural for me to move this assertion up by a line.

src/proof/arith_proof.cpp
```diff
@@ -1068,90 +1130,91 @@ void LFSCArithProof::printTheoryLemmaProof(std::vector<Expr>& lemma,
         }
         default: Unreachable();
       }
-    }
+      }
```
4tXJ7f: Something went wrong with the formatting here. This parentheses was moved even though it shouldn't have been. I think all the changes below are actually just due to this formatting change.

src/proof/arith_proof.cpp
yoni206: fixed in next commit.

src/proof/arith_proof.cpp
yoni206: fixed in next commit.

src/proof/theory_proof.cpp
yoni206: fixed in next commit.

src/proof/arith_proof.cpp
yoni206: fixed in next commit.

src/proof/arith_proof.cpp
yoni206: fixed in next commit.

src/proof/arith_proof.cpp
yoni206: Will touch this line and then clang format changes.

yoni206:
Closing this PR because there are several PRs that will subsume it.
I talked with @alex-ozdemir, and he has a better solution for all three issues that this PR solves.
The first PR he has in this direction is #3484 , which solves item 3 of this PR. But the other items are in the pipeline as well.

PR #2994, opened 2019-05-10
Disable relational triggers by default

ajreynol:
Fixes the nightlies (caused by a timeout on a strings benchmark).

Relational triggers has some unforeseen consequences when used in combination with bounded integer quantifiers, e.g. for strings: `forall x. 0 <= x < str.len(s) => P( x )`.  I believe they should be disabled by default.


ajreynol:
No, the user should still be able to use relational triggers with strings, just not by default.

PR #2998, opened 2019-05-14
Fix model of Boolean vars with eager bit-blaster

4tXJ7f:
When bit-blasting eagerly, we were not assigning values to the Boolean
variables in the `TheoryModel`. With eager bit-blasting, the BV SAT
solver gets all (converted) terms, including the Boolean ones, so
`EagerBitblaster::collectModelInfo()` is responsible for assigning
values to Boolean variables. However, it has only been assigning values
to bit-vector variables, which lead to wrong models. This commit fixes
the issue by asking the `CnfStream` for the Boolean variables, querying
the SAT solver for their value, and assigning them in the `TheoryModel`.

### DIFF
--- a/src/theory/bv/bitblast/eager_bitblaster.cpp
+++ b/src/theory/bv/bitblast/eager_bitblaster.cpp
@@ -271,25 +271,12 @@ bool EagerBitblaster::collectModelInfo(TheoryModel* m, bool fullModel)
   d_cnfStream->getBooleanVariables(vars);
   for (TNode var : vars)
   {
-    Node const_value;
-
-    if (d_cnfStream->hasLiteral(var))
-    {
-      prop::SatLiteral bit = d_cnfStream->getLiteral(var);
-      prop::SatValue bit_value = d_satSolver->value(bit);
-      Assert(bit_value != prop::SAT_VALUE_UNKNOWN);
-      const_value = nm->mkConst(bit_value == prop::SAT_VALUE_TRUE);
-    }
-    else
-    {
-      if (!fullModel)
-      {
-        continue;
-      }
-      const_value = nm->mkConst(false);
-    }
-
-    if (!m->assertEquality(var, const_value, true))
+    Assert(d_cnfStream->hasLiteral(var));
+    prop::SatLiteral bit = d_cnfStream->getLiteral(var);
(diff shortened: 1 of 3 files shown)


src/theory/bv/bitblast/eager_bitblaster.cpp
```diff
@@ -262,6 +265,35 @@ bool EagerBitblaster::collectModelInfo(TheoryModel* m, bool fullModel)
       }
     }
   }
+
+  // Collect the values for the Boolean variables
+  std::vector<TNode> vars;
+  d_cnfStream->getBooleanVariables(vars);
```
aniemetz: Careful, this returns ALL Boolean variables, right?

src/theory/bv/bitblast/eager_bitblaster.cpp
```diff
@@ -262,6 +265,35 @@ bool EagerBitblaster::collectModelInfo(TheoryModel* m, bool fullModel)
...
+    Node const_value;
+
+    if (d_cnfStream->hasLiteral(var))
+    {
+      prop::SatLiteral bit = d_cnfStream->getLiteral(var);
+      prop::SatValue bit_value = d_satSolver->value(bit);
+      Assert(bit_value != prop::SAT_VALUE_UNKNOWN);
+      const_value = nm->mkConst(bit_value == prop::SAT_VALUE_TRUE);
+    }
+    else
```
aniemetz: This case should never happen, right?

4tXJ7f: Yeah, as discussed offline, this probably shouldn't be the case. I removed it and replaced it with an assertion.

src/theory/bv/bitblast/eager_bitblaster.cpp
4tXJ7f: No, we've discussed this offline and we have discovered that it only returns Boolean variables that appear in the input and none of the Boolean variables that are generated during bit-blasting. The variables generated during bit-blasting are `BOOLEAN_TERM_VARIABLE`s and are not counted as Boolean variables by `CnfStream`.

aniemetz:
Did you do a run with `--check-models` on QF_BV? If not I can run your branch with the model validator, two birds with one stone.

PR #3021, opened 2019-05-24
Enhancement: make the bool-to-bv pass more robust and targeted

makaimann:
## Summary
This pull request is an improvement to the `bool-to-bv` preprocessing pass. The existing pass is both too weak and too strong, depending on the circumstance. Throughout this description, "lower" refers to lowering a boolean to a bit-vector.

Note: We did discuss a bug that was in this branch, but to my knowledge there is no bug in the `master` implementation. I.e. this is not a bug fix, only an enhancement.

I am still running tests for both correctness and to verify that this actually improves performance when compared to the `master` implementation of `bool-to-bv`. Please do not merge this until I complete those runs.

### Example of being too weak
If there are arrays in the formula, the pass will give up. This is because before rebuilding a node, the existing pass checks if all the cached children are bit-vectors and if not it gives up. It does not recognize that there are scenarios in which a node should be rebuilt, even if it's not being lowered itself.
e.g. `(= A (store B idx (ite cond val1 val2)))` should lower the `ite` to a `BITVECTOR_ITE` but this does not happen with the current pass

### Example of being too strong
When using `--bool-to-bv=ite`, the pass should only try to lower `ITEs`. However, due to node-hashing, there may be some unintended consequences.

For example, `(and (= x y) (= z (ite (= x y) val1 val2)))` would lower both occurrences of `(= x y)` even in the `bool-to-bv=ite` mode. We would prefer, to only lower the `ite`.

## Solution

This pull request separates the notion of rebuilding a node and lowering it. Even if a node cannot be lowered, it might need to be rebuilt so that changes to it's children are incorporated. This is especially important when there are other sorts, such as arrays, present in the formula. Equality between arrays can't be lowered, but there could be descendants that were safely turned into bit-vectors and the parent node needs to be rebuilt to keep those changes.

Therefore, there is now a distinction between lowering a node and rebuilding it. We have to be careful in both cases, to avoid creating badly typed nodes.

* it is safe to lower a node if all of its children have been turned into bit-vectors
* it is safe to rebuild a node if none of its children will change types (causing it to not be well typed)

Furthermore, to address the second issue, we treat ITEs differently from other nodes. If the mode is `bool-to-bv=ite`, we traverse the formula until we hit an `ITE` and only then do we start trying to lower the node. We also have a separate cache for ITEs that is only added to. There's a separate cache for every other type of node, which is cleared to avoid unnecessarily lowering a node that's not an `ITE` (due to node hashing). Note: in the `ite` mode, it does not convert boolean variables because that requires introducing another `ITE`.

@aniemetz 
@mpreiner


makaimann:
All right, the tests finished running and the current version doesn't have any errors or discrepancies. I ran the following benchmark sets with both debug and production with assertions for `--bool-to-bv=ite` and `--bool-to-bv=all`:

- `QF_ABV`
- `QF_AUFBV`
- `QF_BV`
- `QF_ABV` incremental
- `QF_AUFBV` incremental
- `QF_BV` incremental

I haven't run performance comparisons with the `bool-to-bv` pass implemented in master yet.

src/preprocessing/passes/bool_to_bv.h
```diff
@@ -46,23 +46,72 @@ class BoolToBV : public PreprocessingPass
...
+   *  Returns the lowered node
+   */
+  Node lowerNode(const TNode& node, bool force = false);
+
+  /** Tries to lower one node to a width-one bit-vector
+   *  Caches the result if successful
+   *
+   *  force = true causes booleans to be converted to bit-vectors using an ITE
+   *     this is only used by mode ALL currently, but could conceivably be used
+   *     in new modes.
```
alex-ozdemir: (optional) Perhaps call this parameter `allowIteIntroduction` then (and apply the same name change throughout)?

src/preprocessing/passes/bool_to_bv.h
```diff
@@ -46,23 +46,72 @@ class BoolToBV : public PreprocessingPass
...
+   */
+  void lowerNodeHelper(const TNode& n, bool force = false);
```
alex-ozdemir: (optional) Since you seem to be implementing a traversal, and this is your visit function, perhaps just call it `visit`, and in the header comment say that it lowers a node or rebuilds it or skips it.

src/preprocessing/passes/bool_to_bv.h
```diff
@@ -46,23 +46,72 @@ class BoolToBV : public PreprocessingPass
...
+      forces the top-level assertion. There's no point forcing the assertion
+      to be a bit-vector when it will just be converted back into a boolean.
+  */
+  Node lowerAssertion(const TNode& node, bool force = false);
+
+  /** Traverses subterms to turn booleans into bit-vectors using lowerNodeHelper
+   *  Passes the force argument to lowerNodeHelper
+   *  Returns the lowered node
+   */
+  Node lowerNode(const TNode& node, bool force = false);
```
alex-ozdemir: (optional) Since this does a traversal, I might call it `recursivelyLowerNode`.

src/preprocessing/passes/bool_to_bv.h
```diff
@@ -46,23 +46,72 @@ class BoolToBV : public PreprocessingPass
...
+   *
+   *  force = true causes booleans to be converted to bit-vectors using an ITE
+   *     this is only used by mode ALL currently, but could conceivably be used
+   *     in new modes.
+   */
+  void lowerNodeHelper(const TNode& n, bool force = false);
+
+  /* Traverses formula looking for ITEs to lower to BITVECTOR_ITE using
+   * lowerNode*/
+  Node lowerIte(const TNode& node);
```
alex-ozdemir: (optional) Since this does a traversal, call it `recrusivelyLowerItes`?

src/preprocessing/passes/bool_to_bv.cpp
```diff
@@ -221,7 +215,157 @@ void BoolToBV::lowerNode(const TNode& n)
...
+    Debug("bool-to-bv") << "BoolToBV::lowerNodeHelper skipping: " << n
+                        << std::endl;
+  }
+}
+
+Node BoolToBV::lowerIte(const TNode& node)
+{
+  std::vector<TNode> visit;
+  visit.push_back(node);
+  std::unordered_set<TNode, TNodeHashFunction> visited;
```
alex-ozdemir: (optional) Perhaps `childrenEnqueued`?

src/preprocessing/passes/bool_to_bv.cpp
```diff
@@ -39,165 +39,159 @@ PreprocessingPassResult BoolToBV::applyInternal(
       options::preprocessStep());
 
   unsigned size = assertionsToPreprocess->size();
-  for (unsigned i = 0; i < size; ++i)
+
+  if (options::boolToBitvector() == BOOL_TO_BV_ALL)
```
alex-ozdemir: Perhaps use a switch/case?

src/preprocessing/passes/bool_to_bv.cpp
```diff
@@ -39,165 +39,159 @@ PreprocessingPassResult BoolToBV::applyInternal(
...
+}
+
+Node BoolToBV::lowerNode(const TNode& node, bool force)
 {
-  bool optionITE = options::boolToBitvector() == BOOL_TO_BV_ITE;
-  NodeManager* nm = NodeManager::currentNM();
   std::vector<TNode> visit;
-  visit.push_back(a);
+  visit.push_back(node);
   std::unordered_set<TNode, TNodeHashFunction> visited;
```
alex-ozdemir: (optional) Perhaps call this `childrenEnqueued`?

src/preprocessing/passes/bool_to_bv.cpp
```diff
@@ -39,165 +39,159 @@ PreprocessingPassResult BoolToBV::applyInternal(
...
 
-    int numChildren = n.getNumChildren();
-    Kind k = n.getKind();
-    Debug("bool-to-bv") << "BoolToBV::lowerAssertion Post-traversal with " << n
-                        << " and visited = " << ContainsKey(visited, n)
+    Debug("bool-to-bv") << "BoolToBV::lowerNode: Post-order traversal with "
+                        << n << " and visited = " << ContainsKey(visited, n)
                         << std::endl;
 
+    int numChildren = n.getNumChildren();
```
alex-ozdemir: I'd move this to the (single) use-site.

src/preprocessing/passes/bool_to_bv.cpp
```diff
@@ -39,165 +39,159 @@ PreprocessingPassResult BoolToBV::applyInternal(
...
+
     // Mark as visited
-    /* Optimization: if it's a leaf, don't need to wait to do the work */
-    if (!ContainsKey(visited, n) && (numChildren > 0))
+    if (!ContainsKey(visited, n))
```
alex-ozdemir: Since this, `if` has an `else`, why not switch the bodies and get rid of the `!`, eliminating the implicit double-negation?

src/preprocessing/passes/bool_to_bv.cpp
```diff
@@ -221,7 +215,157 @@ void BoolToBV::lowerNode(const TNode& n)
...
+  bool safe_to_lower =
+      (new_kind != k);  // don't need to lower at all if kind hasn't changed
+
+  // it's safe to rebuild if rebuilding doesn't change any of the types of the
+  // children
+  bool safe_to_rebuild = true;
+
+  for (const Node& nn : n)
+  {
+    if (safe_to_lower)
```
alex-ozdemir: (optional) You could do `safe_to_lower = safe_to_lower && ...;`

My preference is generally to minimize explicit control flow, which is why I suggest this.

src/preprocessing/passes/bool_to_bv.cpp
```diff
@@ -221,7 +215,157 @@ void BoolToBV::lowerNode(const TNode& n)
...
+    {
+      safe_to_lower = fromCache(nn).getType().isBitVector();
+    }
+    if (safe_to_rebuild)
```
alex-ozdemir: (optional) Same as above

src/preprocessing/passes/bool_to_bv.cpp
```diff
@@ -221,7 +215,157 @@ void BoolToBV::lowerNode(const TNode& n)
...
+    {
+      safe_to_rebuild = (fromCache(nn).getType() == nn.getType());
+    }
+    // if it's already not safe to do either, stop checking
+    if (!safe_to_lower && !safe_to_rebuild)
```
alex-ozdemir: Good catch. I was surprised to discover that [GCC doesn't find this kind optimization on its own](https://godbolt.org/z/XUba0d).

src/preprocessing/passes/bool_to_bv.cpp
```diff
@@ -221,7 +215,157 @@ void BoolToBV::lowerNode(const TNode& n)
...
+
+  if (new_kind != k)
+  {
+    if (safe_to_lower)
+    {
+      rebuildNode(n, new_kind);
+      return;
+    }
+
+    else if (force && fromCache(n).getType().isBoolean())
```
alex-ozdemir: You can eliminate this `else if` my moving it's inner if into two else-if's below (adding the `new_kind != k` condition and merging the if above with its outer if.

This has the nice effect of producing an (almost) flat conditional tree with branches that correspond to the four visitation actions:
* lower to BV
* lower to BV using an ITE
* rebuild to incorporate child changes
* do nothing.

which you could then write appropriate comments for.

src/preprocessing/passes/bool_to_bv.cpp
```diff
@@ -221,7 +215,157 @@ void BoolToBV::lowerNode(const TNode& n)
...
+{
+  std::vector<TNode> visit;
+  visit.push_back(node);
+  std::unordered_set<TNode, TNodeHashFunction> visited;
+
+  while (!visit.empty())
+  {
+    TNode n = visit.back();
+    visit.pop_back();
+    Kind k = n.getKind();
```
alex-ozdemir: I'd inline this to its (single) use-site.

src/preprocessing/passes/bool_to_bv.cpp
```diff
@@ -39,165 +39,159 @@ PreprocessingPassResult BoolToBV::applyInternal(
       options::preprocessStep());
 
   unsigned size = assertionsToPreprocess->size();
-  for (unsigned i = 0; i < size; ++i)
+
+  if (options::boolToBitvector() == BOOL_TO_BV_ALL)
+  {
+    for (unsigned i = 0; i < size; ++i)
```
aniemetz: I'd like us to get into the habit of using a) the correct type size without implicitly down casting and b) fixed-width types wherever applicable. In this particular case, `size` should be `size_t` (see https://github.com/CVC4/CVC4/blob/master/src/preprocessing/assertion_pipeline.h#L39).

src/preprocessing/passes/bool_to_bv.cpp
```diff
@@ -39,165 +39,159 @@ PreprocessingPassResult BoolToBV::applyInternal(
...
+    {
+      Node newAssertion = lowerAssertion((*assertionsToPreprocess)[i], true);
+      assertionsToPreprocess->replace(i, Rewriter::rewrite(newAssertion));
+    }
+  }
+  else if (options::boolToBitvector() == BOOL_TO_BV_ITE)
+  {
+    for (unsigned i = 0; i < size; ++i)
```
aniemetz: Ditto

src/preprocessing/passes/bool_to_bv.cpp
aniemetz: It's only two cases, so I'd say no to switch/case. But I'd suggest to store option values that are accessed multiple times in a variable so that you have to access the option only once (not performance critical here but a good habit in general to reduce overhead).

src/preprocessing/passes/bool_to_bv.cpp
```diff
@@ -39,165 +39,159 @@ PreprocessingPassResult BoolToBV::applyInternal(
...
+    {
+      assertionsToPreprocess->replace(
+          i, Rewriter::rewrite(lowerIte((*assertionsToPreprocess)[i])));
+    }
+  }
+  else
   {
-    assertionsToPreprocess->replace(
-        i, Rewriter::rewrite(lowerAssertion((*assertionsToPreprocess)[i])));
+    // Unhandled option
```
aniemetz: Since we only have two cases here, I'd suggest to assert that the second case is the case you expect. Avoid using `Unreachable()` when possible.

src/preprocessing/passes/bool_to_bv.cpp
```diff
@@ -39,165 +39,159 @@ PreprocessingPassResult BoolToBV::applyInternal(
...
+
+  // now try lowering the assertion, but don't force it (even in mode all)
+  lowerNode(assertion, false);
+  Node newAssertion = fromCache(assertion);
+  TypeNode newAssertionType = newAssertion.getType();
+  if (newAssertionType.isBitVector())
+  {
+    Assert(newAssertionType.getBitVectorSize() == 1);
+    NodeManager* nm = NodeManager::currentNM();
+    newAssertion = nm->mkNode(kind::EQUAL, newAssertion, bv::utils::mkOne(1));
```
aniemetz: You could merge these two lines since you use `nm` only once.

src/preprocessing/passes/bool_to_bv.cpp
aniemetz: `numChildren` should be `size_t` (see https://github.com/CVC4/CVC4/blob/master/src/expr/node.h#L571).

src/preprocessing/passes/bool_to_bv.cpp
```diff
@@ -39,165 +39,159 @@ PreprocessingPassResult BoolToBV::applyInternal(
...
-    if (!ContainsKey(visited, n) && (numChildren > 0))
+    if (!ContainsKey(visited, n))
     {
       visited.insert(n);
       visit.push_back(n);
 
       // insert children in reverse order so that they're processed in order
-      //     important for rewriting which sorts by node id
+      //    important for rewriting which sorts by node id
       for (int i = numChildren - 1; i >= 0; --i)
```
aniemetz: Then change this accordingly (see above).

src/preprocessing/passes/bool_to_bv.cpp
aniemetz: `visited` is fine, that's the pattern we commonly use.

src/preprocessing/passes/bool_to_bv.h
aniemetz: No, don't call it `recursively` if it's implemented iteratively -- it might be a bit misleading. And we want to get rid of recursion in general wherever possible.

src/preprocessing/passes/bool_to_bv.cpp
```diff
@@ -39,165 +39,159 @@ PreprocessingPassResult BoolToBV::applyInternal(
...
-  else
-  {
-    Assert(a == fromCache(a));
-    return a;
-  }
+  return fromCache(node);
 }
 
-void BoolToBV::lowerNode(const TNode& n)
+void BoolToBV::lowerNodeHelper(const TNode& n, bool force)
```
aniemetz: This is short and only used once -- why not inline it?

aniemetz:
> I think that we could clean this preprocessing pass up quite a bit by using the node visitor, but there's no need to block this PR on that change.

@alex-ozdemir the problem with NodeVisitor is that it is quite clunky, in particular for light weight use cases. That's the main reason we've not been using it much -- you have to implement a class for every use case. I suggest to not switch to NodeVisitor here for the same reason.

makaimann:
Thanks for all the great feedback @alex-ozdemir and @aniemetz! This is on my todo list but not near the top yet. I should get to it before too long. Unless it needs to be expedited for some reason, in which case I can prioritize it more.

src/preprocessing/passes/bool_to_bv.h
makaimann: Good idea, done with eeae87d6fb90d88145462825066888f943b6738b

src/preprocessing/passes/bool_to_bv.h
makaimann: Done with e74a30cdff8c3dd950be461310a847aaa6a2ab9c

src/preprocessing/passes/bool_to_bv.h
makaimann: Left it as-is.

src/preprocessing/passes/bool_to_bv.h
makaimann: Left it as-is to avoid the "recursively" name.

src/preprocessing/passes/bool_to_bv.cpp
makaimann: Left it the same.

src/preprocessing/passes/bool_to_bv.cpp
makaimann: Good point, done with 516680e786d4ac6bf38d1771cdbd1f250028203a.

src/preprocessing/passes/bool_to_bv.cpp
makaimann: Done with 516680e786d4ac6bf38d1771cdbd1f250028203a.

src/preprocessing/passes/bool_to_bv.cpp
makaimann: Good idea, done with 03d3d6df59cfe118cf430d8483607274e60c4f11

src/preprocessing/passes/bool_to_bv.cpp
makaimann: Good observation, done with 1c069405200345bc284cea108e51aaef63c35190

src/preprocessing/passes/bool_to_bv.cpp
makaimann: Sounds good -- done with 68cd1a8524ba835196351c8eb9a84cd3f15b9df2

src/preprocessing/passes/bool_to_bv.cpp
makaimann: Done with 68cd1a8524ba835196351c8eb9a84cd3f15b9df2

src/preprocessing/passes/bool_to_bv.cpp
makaimann: Note: renamed this to `visit` based on other comments. You're right that it's only used once, but it's actually pretty long (that first return is just one of the exit conditions).

src/preprocessing/passes/bool_to_bv.cpp
makaimann: Done with 2b4d2bafd6aa2381b092701c0af9953a1602c428.

src/preprocessing/passes/bool_to_bv.cpp
makaimann: Done with 7367bdb1ad9a3e59afedfff86ad8df441c6c6794

src/preprocessing/passes/bool_to_bv.cpp
makaimann: Done with a6e29354a9f971683a33ec622bf4ec8716c10936

src/preprocessing/passes/bool_to_bv.cpp
makaimann: Done with a6e29354a9f971683a33ec622bf4ec8716c10936

makaimann:
> But I'd suggest to store option values that are accessed multiple times in a variable so that you have to access the option only once (not performance critical here but a good habit in general to reduce overhead).

In d3dbfa82863aa08af3e9247ba640701b1d93e258 I stored the option in a member variable. This required moving the `bv_option` include to the header. Is this okay? Also, for some reason I can't find the guidelines for the member variable naming scheme. Is it in the Developer guide? `d_` is for "data", right? What should I call a cached option?

makaimann:
After looking at other code, I think it should be called `d_boolToBVMode`, so I just made that change with 59071b1b474cbeb5d1b2cf92c8021cc85654cd0b.

alex-ozdemir:
Sorry for the delay, Makai. I didn't realize you had updated this (although, now that I think about it, you said so on Monday!). Looks good to me!

alex-ozdemir:
Adding "do not merge" so that this doesn't accidentally get merged before Aina takes another look.

PR #3048, opened 2019-06-10
Optimization for negative concatenation membership.

ajreynol:
Documents and adds an optimization for our reduction of negative memberships for concat RE.

### DIFF
--- a/src/theory/strings/regexp_operation.cpp
+++ b/src/theory/strings/regexp_operation.cpp
@@ -827,12 +827,12 @@ void RegExpOpr::simplifyNRegExp( Node s, Node r, std::vector< Node > &new_nodes 
         // then the conclusion of the reduction is quantifier-free:
         //    ~( substr(s,0,n) in R1 ) OR ~( substr(s,n,len(s)-n) in R2)
         Node reLength = TheoryStringsRewriter::getFixedLengthForRegexp(r[0]);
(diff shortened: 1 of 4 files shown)


src/theory/strings/regexp_operation.cpp
```diff
@@ -808,18 +811,42 @@ void RegExpOpr::simplifyNRegExp( Node s, Node r, std::vector< Node > &new_nodes
...
+        //      0 <= x <= len(s) =>
+        //        ~( substr(s,0,x) in R1 ) OR ~( substr(s,x,len(s)-x) in R2)
+        Node lens = nm->mkNode(STRING_LENGTH, s);
+        Node b1;
+        Node b1v;
+        // As an optimization to the above reduction, if we can determine that
+        // all strings in the language of R1 have the same length, say n,
+        // then the conclusion of the reduction is quantifier-free:
+        //    ~( substr(s,0,n) in R1 ) OR ~( substr(s,n,len(s)-n) in R2)
+        Node reLength = TheoryStringsRewriter::getFixedLengthForRegexp(r[0]);
```
4tXJ7f: We can do the same if the last element in the `re.++` has a fixed length, no?

ajreynol: Yes thats true. I can try adding this.

PR #3049, opened 2019-06-10
Do not require sygus constructors to be flattened

ajreynol:
This is in preparation for the sygus v2 parser.

This commit allows for "unflattened" operators in sygus constructors. As an example, the grammar:
A -> A+x | 0
where x is an argument of the function to synthesize would previously be converted to:
A -> A+B | 0
B -> x
Now, we leave the grammar unchanged.  This change potentially makes the sygus solver more efficient for grammars having complex terms as sygus operators. 

The change requires modifications for how the sygus -> builtin conversion works, and minor changes to the heuristic for evaluation unfolding. Another subtle change is that nullary datatype constructors `op` are not stored as sygus operators as `(APPLY_CONSTRUCTOR op)` (a term) instead of `op` (an operator).

### DIFF
--- a/src/theory/datatypes/datatypes_rewriter.cpp
+++ b/src/theory/datatypes/datatypes_rewriter.cpp
@@ -16,6 +16,8 @@
 
 #include "theory/datatypes/datatypes_rewriter.h"
 
+#include "expr/node_algorithm.h"
+
 using namespace CVC4;
 using namespace CVC4::kind;
 
@@ -115,8 +117,7 @@ RewriteResponse DatatypesRewriter::postRewrite(TNode in)
     if (ev.getKind() == APPLY_CONSTRUCTOR)
     {
       Trace("dt-sygus-util") << "Rewrite " << in << " by unfolding...\n";
-      const Datatype& dt =
-          static_cast<DatatypeType>(ev.getType().toType()).getDatatype();
+      const Datatype& dt = ev.getType().getDatatype();
       unsigned i = indexOf(ev.getOperator());
       Node op = Node::fromExpr(dt[i].getSygusOp());
       // if it is the "any constant" constructor, return its argument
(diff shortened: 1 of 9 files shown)


src/theory/datatypes/datatypes_rewriter.h
```diff
@@ -51,6 +51,28 @@ struct SygusSymBreakOkAttributeId
 typedef expr::Attribute<SygusSymBreakOkAttributeId, bool>
     SygusSymBreakOkAttribute;
 
+/** sygus var free
+ *
+ * This attribute is used to mark whether sygus operators have free occurrences
+ * of variables from the formal argument list of the function-to-synthesize.
```
HanielB: why "formal argument list" rather than "argument list"?

ajreynol: "Formal argument list" means the expected arguments in the declaration of the function.

HanielB: What I don't understand is the need for "formal".

ajreynol: This is standard terminology?

src/theory/datatypes/datatypes_rewriter.cpp
```diff
@@ -186,6 +181,67 @@ RewriteResponse DatatypesRewriter::postRewrite(TNode in)
...
+    return args[vn];
+  }
+  // n is an application of operator op.
+  // We must compute the free variables in op to determine if there are
+  // any substitutions we need to make to n.
+  TNode val;
+  if (!op.hasAttribute(SygusVarFreeAttribute()))
+  {
+    std::unordered_set<Node, NodeHashFunction> fvs;
+    if (expr::getFreeVariables(op, fvs))
```
HanielB: Did you ever try to implement a global cache of the free variables of each node/expression? It should be a common enough operation to justify it.

ajreynol: This would be too expensive, since we'd effectively take up O(n^2) space in the worst case.  Notice that the use of attributes in this function avoids the need for recomputing free variables since the attribute is applied once and for all to sygus operators.

PR #3064, opened 2019-06-19
Enable Clang-Format for Java

4tXJ7f:
This commit adds Java to our Clang-Format configuration, which is useful
to format our Java examples.


4tXJ7f:
Note: I am not quite sure how we should handle the curly braces in Java. Our current examples don't put the opening braces on a separate line, so I followed that convention. We may want to discuss this before merging this PR.

mpreiner:
I don't have an issue with keeping the curly braces as is in the Java code.

PR #3066, opened 2019-06-20
Support sygus version 2 format

ajreynol:
This adds support for sygus version 2.0 https://sygus.org/assets/pdf/SyGuS-IF_2.0.pdf

The language is not yet enabled by default.  We will support both version 1.0 and 2.0 for the time being, although eventually support for 1.0 will be removed.

Note: Smt2.g needs to be clang-formatted.


src/parser/smt2/smt2.h
```diff
@@ -332,6 +334,30 @@ class Smt2 : public Parser
...
+   * Adds a constructor to sygus datatype dt whose sygus operator is term.
+   *
+   * ntsToUnres contains a mapping from non-terminal symbols to the unresolved
+   * types they correspond to. This map indicates how the argument term should
+   * be interpreted (instances of symbols from the domain of ntsToUnres
+   * correspond to constructor arguments).
+   *
+   * The sygus operator that is actually added to dt corresponds to replacing
+   * each occurrence of non-terminal symbols from the domain of ntsToUnres
+   * with bound variables via purifySygusGTerm, and bounding these variables
```
HanielB: bounding --> binding

src/parser/smt2/smt2.cpp
```diff
@@ -497,13 +497,18 @@ bool Smt2::logicIsSet() {
...
+    // allow unary minus in sygus version 1
     return getExprManager()->mkConst(Rational(name));
-  }else if(isAbstractValue(name)) {
+  }
+  else if (isAbstractValue(name))
+  {
     return mkAbstractValue(name);
-  }else{
+  }
+  else
```
HanielB: no-else-after-return

src/parser/smt2/smt2.cpp
```diff
@@ -1431,6 +1448,125 @@ Expr Smt2::makeSygusBoundVarList(Datatype& dt,
...
+          pureVar = false;
+          break;
+        }
+      }
+    }
+    Trace("parser-sygus2") << "Pure var is " << pureVar
+                           << ", hasOp=" << op.hasOperator() << std::endl;
+    if (pureVar && op.hasOperator())
+    {
+      // optimization: just use the operator if it an application to only vars
```
HanielB: I'd write "use just" to avoid confusion. Also, fix "if it an"

PR #3067, opened 2019-06-20
Stratify unfolding of regular expressions based on polarity

ajreynol:
This ensures that if `x in R1` and `NOT x in R2` are simultaneously asserted, then we unfold the former only.  The motivation is that positive memberships are easier to check and are more likely to lead to conflicts. Moreover, it is often the case that a negative membership for x is often trivially satisfied by unfolding a positive membership for x. 

This helps the performance of some Amazon benchmarks.


src/theory/strings/regexp_solver.cpp
```diff
@@ -172,23 +172,35 @@ void RegExpSolver::check()
   if (!addedLemma)
   {
     NodeManager* nm = NodeManager::currentNM();
-    for (unsigned i = 0; i < d_regexp_memberships.size(); i++)
+    // representatives of strings that are the LHS of positive memberships that
+    // we unfolded
+    std::unordered_set<Node, NodeHashFunction> repUnfold;
+    // check positive (e=0), then negative (e=1) memberships
+    for (unsigned e = 0; e < 2; e++)
```
4tXJ7f: Suggestion: we could do `for (bool e : {true, false})` here instead.

ajreynol: I prefer not using more fancy C++ language features.

PR #3072, opened 2019-06-25
Make mkOpTerm const

makaimann:
It looks like mkOpTerm can be const but is not declared as such. This pull request would rectify that.


makaimann:
This came up because the generic api makes as many member functions as possible `const` and without this change that’s not possible for the CVC4 implementation.

PR #3073, opened 2019-06-26
New C++ API: Add templated getIndices method for OpTerm

makaimann:
This pull request implements a templated `getIndices` method for `OpTerm` which allows retrieving the indices used to build the `OpTerm`. For indices that support arbitrary precision integers, such as `Divisible`, the index is returned as a string only.


makaimann:
Let me know what you think! Templates seemed like the easiest solution to me, but I'm sure there are other ways of implementing this.

I also wrote a unit test, but I'm not sure where the right place to put it is. Should I just put them in `test/unit/api`? My tests are different than those because it requires a solver to build the `OpTerm` in the first place. I've attached the unit test here.

[unit-index.zip](https://github.com/CVC4/CVC4/files/3331698/unit-index.zip)

4tXJ7f:
This seems like a pretty reasonable solution to me style-wise. The only other option that I can think for dealing with the different types would be to have what is currently a template argument as part of the name: `getStringIndex()`, `getIntIndex()`, etc. One other question is whether we want to directly return all indices or pass the index of the index to retrieve :) I would be happy to keep the current design though.

Regarding the unit tests: It seems like adding them to [opterm_black.h](https://github.com/CVC4/CVC4/blob/master/test/unit/api/opterm_black.h) would be a good option or is there an issue with that?

makaimann:
Thanks for the feedback! I’m pretty open style-wise, so I don’t have any strong opinions between templates and including it as part of the function name. As for the unit tests, that definitely works. I didn’t notice that opterm_black.h already has a solver in it.

src/api/cvc4cpp.cpp
```diff
@@ -1280,6 +1280,143 @@ Sort OpTerm::getSort() const
...
 bool OpTerm::isNull() const { return d_expr->isNull(); }
 
+template <>
+std::string OpTerm::getIndices() const
+{
+  CVC4_API_CHECK_NOT_NULL;
+  std::string i;
+  Kind k = intToExtKind(d_expr->getKind());
+
+  if (k == DIVISIBLE_OP)
```
aniemetz: Why is `DIVISIBLE_OP` here? It's index arg is uint32_t not string: 
https://github.com/CVC4/CVC4/blob/master/src/api/cvc4cpp.cpp#L2735
https://github.com/CVC4/CVC4/blob/master/src/api/cvc4cpp.cpp#L2749

src/api/cvc4cpp.cpp
```diff
@@ -1280,6 +1280,143 @@ Sort OpTerm::getSort() const
...
+    CVC4::Integer _int = d_expr->getConst<Divisible>().k;
+    i = _int.toString();
+  }
+  else if (k == RECORD_UPDATE_OP)
+  {
+    i = d_expr->getConst<RecordUpdate>().getField();
+  }
+  else
+  {
+    CVC4ApiExceptionStream().ostream() << "Can't get string index from"
```
```suggestion
    CVC4_API_CHECK(false) << "Can't get string index from"
```
Your solution is in general fine, but I'd prefer to use a consistent pattern for all checks/throwing API exceptions. It also 'hides' the 'internal' (stream) handling here, which makes it a bit simpler (as a user of the macro you don't have to think about how and why the OStreamVoider is used and when it is fine to not use it).

src/api/cvc4cpp.cpp
```diff
@@ -1280,6 +1280,143 @@ Sort OpTerm::getSort() const
...
+Kind OpTerm::getIndices() const
+{
+  CVC4_API_CHECK_NOT_NULL;
+  Kind kind = intToExtKind(d_expr->getKind());
+  CVC4_API_KIND_CHECK_EXPECTED(kind == CHAIN_OP, kind) << "CHAIN_OP";
+  return intToExtKind(d_expr->getConst<Chain>().getOperator());
+}
+
+template <>
+uint32_t OpTerm::getIndices() const
```
aniemetz: See above, `DIVISIBLE_OP` belongs here.

src/api/cvc4cpp.cpp
```diff
@@ -1280,6 +1280,143 @@ Sort OpTerm::getSort() const
...
+  }
+  else if (k == FLOATINGPOINT_TO_FP_GENERIC_OP)
+  {
+    CVC4::FloatingPointToFPGeneric ext =
+        d_expr->getConst<FloatingPointToFPGeneric>();
+    indices = std::make_pair(ext.t.exponent(), ext.t.significand());
+  }
+  else
+  {
+    CVC4ApiExceptionStream().ostream()
```
```suggestion
    CVC4_API_CHECK(false)
```

test/unit/api/opterm_black.h
```diff
@@ -27,6 +27,11 @@ class OpTermBlack : public CxxTest::TestSuite
   void testGetKind();
   void testGetSort();
   void testIsNull();
+  void testStringIndices();
```
aniemetz:
```suggestion
  void testGetIndicesString();
```

test/unit/api/opterm_black.h
```diff
@@ -27,6 +27,11 @@ class OpTermBlack : public CxxTest::TestSuite
...
+  void testKindIndices();
```
aniemetz:
```suggestion
  void testGetIndicesKind();
```

test/unit/api/opterm_black.h
```diff
@@ -27,6 +27,11 @@ class OpTermBlack : public CxxTest::TestSuite
...
+  void testUintIndices();
```
aniemetz:
```suggestion
  void testGetIndicesUint();
```

test/unit/api/opterm_black.h
```diff
@@ -27,6 +27,11 @@ class OpTermBlack : public CxxTest::TestSuite
...
+  void testPairUintIndices();
```
aniemetz:
```suggestion
  void testGetIndicesPairUint();
```

test/unit/api/opterm_black.h
```diff
@@ -55,3 +60,117 @@ void OpTermBlack::testIsNull()
   x = d_solver.mkOpTerm(BITVECTOR_EXTRACT_OP, 31, 1);
   TS_ASSERT(!x.isNull());
 }
+
+void OpTermBlack::testStringIndices()
```
aniemetz:
```suggestion
void OpTermBlack::testGetIndicesString()
```

test/unit/api/opterm_black.h
```diff
@@ -55,3 +60,117 @@ void OpTermBlack::testIsNull()
...
+{
+  OpTerm divisible_ot = d_solver.mkOpTerm(DIVISIBLE_OP, 4);
```
aniemetz: Reminder: does not belong here

test/unit/api/opterm_black.h
```diff
@@ -55,3 +60,117 @@ void OpTermBlack::testIsNull()
...
+  std::string divisible_idx = divisible_ot.getIndices<std::string>();
+  TS_ASSERT(divisible_idx == "4");
+
+  OpTerm record_update_ot = d_solver.mkOpTerm(RECORD_UPDATE_OP, "test");
+  std::string record_update_idx = record_update_ot.getIndices<std::string>();
+  TS_ASSERT(record_update_idx == "test");
+}
+
+void OpTermBlack::testKindIndices()
```
aniemetz:
```suggestion
void OpTermBlack::testGetIndicesKind()
```

test/unit/api/opterm_black.h
```diff
@@ -55,3 +60,117 @@ void OpTermBlack::testIsNull()
...
+{
+  OpTerm chain_ot = d_solver.mkOpTerm(CHAIN_OP, AND);
+  Kind chain_idx = chain_ot.getIndices<Kind>();
+  TS_ASSERT(chain_idx == AND);
+}
+
+void OpTermBlack::testUintIndices()
```
aniemetz:
```suggestion
void OpTermBlack::testGetIndicesUint()
```

test/unit/api/opterm_black.h
```diff
@@ -55,3 +60,117 @@ void OpTermBlack::testIsNull()
...
+  OpTerm floatingpoint_to_sbv_total_ot = d_solver.mkOpTerm(FLOATINGPOINT_TO_SBV_TOTAL_OP, 14);
+  uint32_t floatingpoint_to_sbv_total_idx = floatingpoint_to_sbv_total_ot.getIndices<uint32_t>();
+  TS_ASSERT(floatingpoint_to_sbv_total_idx == 14);
+
+  OpTerm tuple_update_ot = d_solver.mkOpTerm(TUPLE_UPDATE_OP, 5);
+  uint32_t tuple_update_idx = tuple_update_ot.getIndices<uint32_t>();
+  TS_ASSERT(tuple_update_idx == 5);
+}
+
+void OpTermBlack::testPairUintIndices()
```
aniemetz:
```suggestion
void OpTermBlack::testGetIndicesPairUint()
```

test/unit/api/opterm_black.h
```diff
@@ -55,3 +60,117 @@ void OpTermBlack::testIsNull()
...
+  OpTerm floatingpoint_to_fp_unsigned_bitvector_ot = d_solver.mkOpTerm(FLOATINGPOINT_TO_FP_UNSIGNED_BITVECTOR_OP, 4, 25);
+  std::pair<uint32_t, uint32_t> floatingpoint_to_fp_unsigned_bitvector_indices = floatingpoint_to_fp_unsigned_bitvector_ot.getIndices<std::pair<uint32_t, uint32_t>>();
+  TS_ASSERT((floatingpoint_to_fp_unsigned_bitvector_indices == std::pair<uint32_t, uint32_t>{4, 25}));
+
+  OpTerm floatingpoint_to_fp_generic_ot = d_solver.mkOpTerm(FLOATINGPOINT_TO_FP_GENERIC_OP, 4, 25);
+  std::pair<uint32_t, uint32_t> floatingpoint_to_fp_generic_indices = floatingpoint_to_fp_generic_ot.getIndices<std::pair<uint32_t, uint32_t>>();
+  TS_ASSERT((floatingpoint_to_fp_generic_indices == std::pair<uint32_t, uint32_t>{4, 25}));
+}
+
+void OpTermBlack::testGetIndicesExceptions()
```
aniemetz: I'd prefer having these checks in the corresponding functions above rather than in a separate function (that's the pattern used in all the API unit tests so far).

src/api/cvc4cpp.cpp
makaimann: That’s actually incorrect, it’s an arbitrary precision integer: https://github.com/CVC4/CVC4/blob/57524fd9f204f8e85e5e37af1444a6f76d809aee/src/util/divisible.h#L34
I return it as a string to support arbitrary precision values.
`mkOpTerm` also supports building it with a string for the same reason: https://github.com/CVC4/CVC4/blob/57524fd9f204f8e85e5e37af1444a6f76d809aee/src/api/cvc4cpp.h#L1829

src/api/cvc4cpp.cpp
makaimann: Gotcha, thanks for the feedback!

src/api/cvc4cpp.cpp
makaimann: Moving conversation to above. `DIVISIBLE_OP` has an arbitrary precision `Integer` index which can't always be represented by a `uint64_t`.

test/unit/api/opterm_black.h
makaimann: Fixed with 02da5ec1ab49dc09625bd26e005489bd4fba0ed3

src/api/cvc4cpp.cpp
aniemetz: Hm, your link to cvc4cpp.h actually confirms my above, currently we only have support for uin32_t. We should probably support both (it is quite inconvenient if you have to convert strings that fit into uint back and forth). String as argument should then also be supported in the API though, which it currently isn't. @mpreiner @4tXJ7f what do you think?

test/unit/api/opterm_black.h
```diff
@@ -55,3 +59,106 @@ void OpTermBlack::testIsNull()
...
+
+  OpTerm tuple_update_ot = d_solver.mkOpTerm(TUPLE_UPDATE_OP, 5);
+  uint32_t tuple_update_idx = tuple_update_ot.getIndices<uint32_t>();
+  TS_ASSERT(tuple_update_idx == 5);
+  TS_ASSERT_THROWS(tuple_update_ot.getIndices<std::string>(), CVC4ApiException&);
+}
+
+void OpTermBlack::testGetIndicesPairUint()
+{
+    OpTerm bitvector_extract_ot = d_solver.mkOpTerm(BITVECTOR_EXTRACT_OP, 4, 0);
```
aniemetz: Indent?

src/api/cvc4cpp.cpp
makaimann: Oh duh, I was reading the function declaration above the comment instead of below it, oops.

It is inconvenient to convert it, but I still don't really like the extra infrastructure needed for bounds checking... Either way, string as argument should be supported in the API right? Otherwise functionality is limited.

test/unit/api/opterm_black.h
makaimann: Thanks

makaimann:
I did, but that didn't fix the indent -- fixed it manually.

src/api/cvc4cpp.cpp
aniemetz: We can do string and uint23_t (or change this to uint64_t actually) to create, and string to return (as you have now). Sounds reasonable?

aniemetz:
The lines in opterm_black.h are definitely longer than 80 chars - that should not be the case after clang-format. How did you format? (command)

src/api/cvc4cpp.cpp
makaimann: Yeah, sounds good to me!

makaimann:
Ahh I'm doing `git clang-format HEAD~<N>`. So I probably forgot to `clang-format` in earlier commits, I'll make sure I cover the whole pull request.

src/api/cvc4cpp.cpp
mpreiner: Sounds good.

makaimann: Added support for creating DIVISIBLE_OP with a string in f87e6bb4c5d7e4de31099973244b3b5fc4874316

test/unit/api/opterm_black.h
```diff
@@ -55,3 +60,157 @@ void OpTermBlack::testIsNull()
   x = d_solver.mkOpTerm(BITVECTOR_EXTRACT_OP, 31, 1);
   TS_ASSERT(!x.isNull());
 }
+
+void OpTermBlack::testMakeDivisibleOp()
```
aniemetz: This should actually go into solver_black, just add it in this function: https://github.com/CVC4/CVC4/blob/master/test/unit/api/solver_black.h#L445

makaimann: Oops, fixed with 4d91687f33ba4186d7230e307126221b1e62d86c. Thanks for the pointer.

src/api/cvc4cpp.cpp
```diff
@@ -2724,10 +2862,26 @@ OpTerm Solver::mkOpTerm(Kind kind, Kind k) const
...
+  {
+    case RECORD_UPDATE_OP:
+      res = *mkValHelper<CVC4::RecordUpdate>(CVC4::RecordUpdate(arg))
+                 .d_expr.get();
+      break;
+    case DIVISIBLE_OP:
+      res = *mkValHelper<CVC4::Divisible>(CVC4::Divisible(CVC4::Integer(arg)))
+                 .d_expr.get();
+      break;
+    default:
```
aniemetz: This case can never occur, kind is guarded in the check above. You can safely convert this `switch` to an `if`. For API functions we prefer having argument checks at the top of the function, hence the preference to convert this to `if` and keep the guard above.

makaimann: No worries, good catch! Fixed with f0cd7ef0844e57cd483907946c0f909b9b49af1f. I turned it into an if/else without checking the second case because there's only two possibilities. I also took at the assertion that `res` is not null because that should never happen.

One last question: a possible issue is that the user passes in a string that doesn't represent a valid `Integer`. Should I catch that exception and do something special or just let the API try/catch macro handle it?

aniemetz: In general we would want to always catch such cases on the API level (and then guard internally with assertions that no invalid argument is passed internally -- think of API guards as a 'fence').

With Integer it's a bit of a special case, unfortunately, since we support two different back end implementations (GMP, CLN), and they do not necessarily agree on what is (in)valid input. So, yes, we generally want to explicitly guard against it in the API, but since we use a 3rd-party backend we have no control over and which throws an exception anyways, we delegate it to the back end and let the macro handle it.

But good thing you brought this up, because there is one thing we actually need to do here due to the above mentioned differences. Add a guard like this one here:
https://github.com/CVC4/CVC4/blob/master/src/api/cvc4cpp.cpp#L2067

And let me also add what I wrote above as documentation of the file. Will submit PR.

makaimann: Cool, added the guard in b3068b68424b2bd0be4d94dd852be7e8da25a17a.

PR #3074, opened 2019-06-28
Optimize DRAT optimization: clause matching

alex-ozdemir:
tldr: I made a bad data-structure/algorithm choice when implementing
part of DRAT/CNF-optimization, which consumed **a lot** of time on some
bechmarks. This commit fixes that choice.

Long Version:

Set-Keyed Maps Considered Harmful
=================================

Algorithmic Problem
-------------------

The DRAT optimization process spits out a unsatifiable core of the CNF.
The clauses in the core are all from the original formula, but their
literals may have been reordered and deduplicated.  We must match the
old clauses with new ones, so we know which old clauses are in the core.

Old (BAD) Solution
------------------

Before I didn't really think about this process very much. I built a
solution in which clauses were canonically represented by hash sets of
literals, and made a hash map from canonical clauses to clause indices
into the original CNF.

Problem With Old Solution
-------------------------

In hindsight this was a bad idea. First, it required a new hash set to
be heap-allocated for every clause in the CNF. Second, the map lookups
required set-hashes (reasonable -- linear time, once) and hash-set
equality (not reasonable -- quadratic time, multiple times) on every
lookup.

New Solution
------------

The ideal solution is probably to have the map from clauses to clause
ids be something like a trie. STL doesn't have a trie, but representing
clauses as sorted, deduped vectors of literal in a tree based on
lexicographical comparison is pretty closed to this. On randomly chosen
examples it seems to be a huge improvement over the old
map-keyed-by-sets solution, and I'm in the process of running a full set
of bechmarks.

Also, we store pointers to the clauses already stored elsewhere in the
proof, instead of allocating new memory for them.

Measuring the Improvment
-----------------

Using the statistics also included in this PR, I measured how the amount
of time spent doing core-input clause matching changed. `set-keyed hashmap` 
is the old system. `vector-keyed treemap` is the new. Notice that for some
instances, the old system spent a lot of the proof production time matching
clauses (how embarrassing :frowning_face:). Now it spends significantly less
time. 

### Clause Matching Time as a Fraction of DRAT Optimization Time

clause-match-v-opt-time

### Clause Matching Time as a Fraction of Proof Production Time
clause-match-v-production-time

### Time Spent on Different Tasks
task-time-dist

### Solved Instances
```
DIRECTORY |    Tot |   Slvd |  Sat |  Unsat |  Unk |    To |   Mo |  Err |     Cpu[s] |      Mem[MB] |
run-er    |  21127 |  13822 |    2 |  13820 |    0 |  6791 |   99 |  415 |  4493334.7 |  128195543.1 |
run-er    |  21127 |  13857 |    2 |  13855 |    0 |  6623 |  102 |  545 |  4354398.9 |  128616361.1 |
```

We check 35 more proofs! We also error more. I haven't verified why, but I
assume it is because we attempt to check more proofs and have LFSC fail us :cry:. 

Future Work
-----------

It may also be reasonable to do a hash map of sorted, deduped, vector
clauses, or actually do a trie. I haven't tried this, yet (there's a TODO in the code).

### DIFF
--- a/src/proof/clausal_bitvector_proof.h
+++ b/src/proof/clausal_bitvector_proof.h
@@ -101,7 +101,8 @@ class ClausalBitVectorProof : public BitVectorProof
     TimerStat d_totalTime;
     // Time that drat-trim actually spent optimizing the DRAT proof/formula
     TimerStat d_toolTime;
-    // Time that drat-trim actually spent optimizing the DRAT proof/formula
+    // Time that was spent matching clauses in drat-trim's output to clauses in
+    // its input
     TimerStat d_clauseMatchingTime;
     // Bytes in binary DRAT proof before optimization
     IntStat d_initialDratSize;
--- a/src/proof/clausal_bitvector_proof.h
+++ b/src/proof/clausal_bitvector_proof.h
@@ -87,6 +87,10 @@ class ClausalBitVectorProof : public BitVectorProof
   // of clause actually needed to check that proof (a smaller UNSAT core)
   void optimizeDratProof();
 
+  // Given reference to a SAT clause encoded as a vector of literals, puts the
+  // literals into a canonical order
+  static void canonicalizeClause(prop::SatClause& clause);
+
   struct DratOptimizationStatistics
   {
     DratOptimizationStatistics();
--- a/src/proof/clausal_bitvector_proof.h
+++ b/src/proof/clausal_bitvector_proof.h
@@ -33,6 +33,7 @@
 #include "prop/cnf_stream.h"
 #include "prop/sat_solver_types.h"
 #include "theory/bv/theory_bv.h"
+#include "util/statistics_registry.h"
 
 namespace CVC4 {
 
@@ -67,10 +68,48 @@ class ClausalBitVectorProof : public BitVectorProof
   std::ostringstream d_binaryDratProof{};
   std::vector<ClauseId> d_coreClauseIndices{};
 
+  struct DratTranslationStatistics
+  {
+    DratTranslationStatistics();
+    ~DratTranslationStatistics();
+
+    // Total time spent doing translation (optimized binary DRAT -> in memory
+    // target format) (including IO, postprocessing, etc.)
+    TimerStat d_totalTime;
+    // Time that the external tool actually spent
+    TimerStat d_toolTime;
+  };
+
+  DratTranslationStatistics d_dratTranslationStatistics;
+
  private:
   // Optimizes the DRAT proof stored in `d_binaryDratProof` and returns a list
   // of clause actually needed to check that proof (a smaller UNSAT core)
   void optimizeDratProof();
+
+  struct DratOptimizationStatistics
+  {
+    DratOptimizationStatistics();
+    ~DratOptimizationStatistics();
+
+    // Total time spent using drat-trim to optimize the DRAT proof/formula
+    // (including IO, etc.)
+    TimerStat d_totalTime;
+    // Time that drat-trim actually spent optimizing the DRAT proof/formula
+    TimerStat d_toolTime;
+    // Time that drat-trim actually spent optimizing the DRAT proof/formula
+    TimerStat d_clauseMatchingTime;
+    // Bytes in binary DRAT proof before optimization
+    IntStat d_initialDratSize;
+    // Bytes in binary DRAT proof after optimization
+    IntStat d_optimizedDratSize;
+    // Bytes in textual DIMACS bitblasted formula before optimization
+    IntStat d_initialFormulaSize;
+    // Bytes in textual DIMACS bitblasted formula after optimization
+    IntStat d_optimizedFormulaSize;
+  };
+
+  DratOptimizationStatistics d_dratOptimizationStatistics;
 };
 
 /**
--- a/src/proof/clausal_bitvector_proof.h
+++ b/src/proof/clausal_bitvector_proof.h
@@ -74,7 +74,7 @@ class ClausalBitVectorProof : public BitVectorProof
     ~DratTranslationStatistics();
 
     // Total time spent doing translation (optimized binary DRAT -> in memory
-    // target format) (including IO, postprocessing, etc.)
+    // target format including IO, postprocessing, etc.)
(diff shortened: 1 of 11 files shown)


src/proof/proof_manager.cpp
```diff
@@ -561,8 +561,8 @@ void LFSCProof::toStream(std::ostream& out, const ProofLetMap& map) const
 void LFSCProof::toStream(std::ostream& out) const
 {
   TimerStat::CodeTimer proofProductionTimer(
-      *ProofManager::currentPM()->getProofProductionTime());
-
+      ProofManager::currentPM()->getStats().d_proofProductionTime);
+  ProofManager::currentPM()->getStats().d_skeletonProofTraceTime.start();
```
4tXJ7f: Instead of manually starting/stopping timers, you can also use a `CodeTimer` within a block (just a set of curly braces). The advantage of that is that the timer automatically gets stopped, no matter how the block is exited (e.g. due to an exception).

src/proof/clausal_bitvector_proof.h
```diff
@@ -67,10 +68,52 @@ class ClausalBitVectorProof : public BitVectorProof
...
+  {
+    DratOptimizationStatistics();
+    ~DratOptimizationStatistics();
+
+    // Total time spent using drat-trim to optimize the DRAT proof/formula
+    // (including IO, etc.)
+    TimerStat d_totalTime;
+    // Time that drat-trim actually spent optimizing the DRAT proof/formula
+    TimerStat d_toolTime;
+    // Time that drat-trim actually spent optimizing the DRAT proof/formula
```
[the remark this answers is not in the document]
alex-ozdemir: Good catch!

src/proof/er/er_proof.cpp
```diff
@@ -80,7 +80,7 @@ TraceCheckProof TraceCheckProof::fromText(std::istream& in)
   return pf;
 }
 
-ErProof ErProof::fromBinaryDratProof(
+std::pair<ErProof, TimerStat> ErProof::fromBinaryDratProof(
```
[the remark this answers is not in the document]
alex-ozdemir: Good idea!

src/proof/lrat/lrat_proof.cpp
```diff
@@ -124,7 +124,7 @@ void printIndices(std::ostream& o, const std::vector<ClauseIdx>& indices)
 
 // Prints the LRAT addition line in textual format
 
-LratProof LratProof::fromDratProof(
+std::pair<LratProof, TimerStat> LratProof::fromDratProof(
```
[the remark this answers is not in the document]
alex-ozdemir: Good idea.
I was somewhat unhappy with returning the timer. Glad to have a better solution.

src/proof/proof_manager.cpp
alex-ozdemir: That's neat -- I didn't know about the exception interaction.

Sounds good. One of these days I should split this method up too.

alex-ozdemir:
> This is awesome! Great findings and a good lesson for all of us. I just a have a couple of minor comments (see below).
> 
> Some other thoughts:
> 
> > new hash set to be heap-allocated for every clause in the CNF
> 
> Just for the record: The vectors are still doing heap allocations under the hood but allocating a vector should be cheaper overall than allocating a hash set. Using pointers instead of allocating them again is of course much cheaper (but you could do the same with the hash sets).

Agreed. The new solution uses pointers to vector-clauses already stored in the proof.

> 
> > hash-set equality (not reasonable -- quadratic time, multiple times)
> 
> To be fair, that's only the worst case (if all elements in the "other" set are in a single bucket). Average complexity should be `O(n)`. What's probably worse though is that such a comparison results in a lot more random memory accesses than comparing two vectors lexicographically.

Good point -- I hadn't considered the memory access pattern and its significance.

> > It may also be reasonable to do a hash map of sorted, deduped, vector clauses, or actually do a trie.
> 
> I'd be very interested in seeing a comparison with a hash map. A comparison with a "real" trie would be interesting as well but I am not sure if it is worth all the work to implement it.

Agreed. Something to look into when timer permits.

> Again, really great work finding this issue and coming up with the fix.

src/proof/proof_manager.cpp
```diff
@@ -562,198 +562,249 @@ void LFSCProof::toStream(std::ostream& out) const
...
 
-  NodeSet::iterator it3;
-  Debug("pf::pm") << std::endl << "Used assertions: " << std::endl;
-  for (it3 = used_assertions.begin(); it3 != used_assertions.end(); ++it3)
-    Debug("pf::pm") << "\t assertion = " << *it3 << std::endl;
+    // Debug("pf::pm") << std::endl << "Used lemmas: " << std::endl;
+    // for (it2 = used_lemmas.begin(); it2 != used_lemmas.end(); ++it2) {
+    //   Debug("pf::pm") << "\t lemma = " << *(it2->second) << std::endl;
+    // }
+    // Debug("pf::pm") << std::endl;
```
4tXJ7f: Hmm, maybe remove the commented out code above?

src/proof/proof_manager.cpp
```diff
@@ -562,198 +562,249 @@ void LFSCProof::toStream(std::ostream& out) const
...
-    utils::collectAtoms(*it, atoms);
-    // utils::collectAtoms(*it, newAtoms);
-  }
+    // collects the atoms in the assertions
+    for (NodeSet::const_iterator it = used_assertions.begin();
+         it != used_assertions.end();
+         ++it)
+    {
+      utils::collectAtoms(*it, atoms);
+      // utils::collectAtoms(*it, newAtoms);
```
4tXJ7f: Remove commented out code?

src/proof/proof_manager.cpp
alex-ozdemir: Sounds good. If I'm already messing up the git-blame, I might as well clean things.

src/proof/proof_manager.cpp
alex-ozdemir: Sounds good. If I'm already messing up the git-blame, I might as well clean things.

PR #3076, opened 2019-06-28
Refactor strings to use an inference manager object

ajreynol:
This organizes the theory of strings so that the processing of inferences is handled by a central, customized inference manager object, which manages whether inferences are facts/lemmas/conflicts. It can be seem as a wrapper around the output channel of TheoryStrings and the equality engine of TheoryStrings.

This required cleaning up a few other minor things (mkAnd, utilities related to proxy variables) to enable this split.

The behavior of the code is mostly unchanged.

This is work towards #1881.


src/theory/strings/output_channel.cpp
```diff
@@ -0,0 +1,448 @@
...
+  if (eq.isConst())
+  {
+    return false;
+  }
+  NodeManager* nm = NodeManager::currentNM();
+  Node lemma_or = nm->mkNode(OR, eq, nm->mkNode(NOT, eq));
+  Trace("strings-lemma") << "Strings::Lemma " << c << " SPLIT : " << lemma_or
+                         << std::endl;
+  d_pending_lem.push_back(lemma_or);
+  sendPhaseRequirement(eq, preq);
```
4tXJ7f: The splits statistic is missing. Can we maybe add it to `OutputChannelStrings`?

src/theory/strings/output_channel.cpp
```diff
@@ -0,0 +1,448 @@
...
+}
+
+Node OutputChannelStrings::getRepresentative(Node t)
+{
+  if (d_ee.hasTerm(t))
+  {
+    return d_ee.getRepresentative(t);
+  }
+  return t;
+}
```
4tXJ7f: IMHO, we should only have an implementation here or in `TheoryStrings` and not in both.

src/theory/strings/output_channel.cpp
```diff
@@ -0,0 +1,448 @@
...
+
+bool OutputChannelStrings::hasTerm(Node a) { return d_ee.hasTerm(a); }
```
4tXJ7f: Ditto.

src/theory/strings/output_channel.cpp
```diff
@@ -0,0 +1,448 @@
...
+
+bool OutputChannelStrings::areEqual(Node a, Node b)
```
4tXJ7f: Ditto.

src/theory/strings/output_channel.cpp
```diff
@@ -0,0 +1,448 @@
...
+    return true;
+  }
+  else if (hasTerm(a) && hasTerm(b))
+  {
+    return d_ee.areEqual(a, b);
+  }
+  return false;
+}
+
+bool OutputChannelStrings::areDisequal(Node a, Node b)
```
4tXJ7f: Ditto.

src/theory/strings/output_channel.cpp
```diff
@@ -0,0 +1,448 @@
...
+    }
+    i++;
+  }
+  d_pending.clear();
+  d_pending_exp.clear();
+}
+
+void OutputChannelStrings::doPendingLemmas()
+{
+  if (!hasConflict())
```
4tXJ7f: In the original code, we don't do that if we don't have pending lemmas. Is this change intentional? It seems like it could make a difference if we have pending phase requirements but no pending lemmas.

src/theory/strings/output_channel.h
```diff
@@ -0,0 +1,308 @@
...
+#include "cvc4_private.h"
+
+#ifndef CVC4__THEORY__STRINGS__OUTPUT_CHANNEL_H
+#define CVC4__THEORY__STRINGS__OUTPUT_CHANNEL_H
+
+#include "context/cdhashset.h"
+#include "theory/uf/equality_engine.h"
+
+#include <climits>
+#include <deque>
```
4tXJ7f: Those includes should come before the project includes.

src/theory/strings/output_channel.h
```diff
@@ -0,0 +1,308 @@
...
+ ** \brief Customized output channel for the theory of strings
+ **/
+
+#include "cvc4_private.h"
+
+#ifndef CVC4__THEORY__STRINGS__OUTPUT_CHANNEL_H
+#define CVC4__THEORY__STRINGS__OUTPUT_CHANNEL_H
+
+#include "context/cdhashset.h"
+#include "theory/uf/equality_engine.h"
```
4tXJ7f: Add missing includes for `Context`, `Node`, `std::vector`, `OutputChannel`, `std::map`.

src/theory/strings/output_channel.h
```diff
@@ -0,0 +1,308 @@
...
+  /** same as above, but where exp_n is empty */
+  void sendInference(std::vector<Node>& exp,
+                     Node eq,
+                     const char* c,
+                     bool asLemma = false);
+  /** Send split
+   *
+   * This requests that ( a = b V a != b ) is sent on the output channel as a
+   * lemma. We additionally request that a phase requirement the equality a=b
+   * to polarity preq.
```
4tXJ7f: The last sentence does not seem to parse for me.

src/theory/strings/output_channel.h
```diff
@@ -0,0 +1,308 @@
...
+   * debugging.
+   */
+  void sendLemma(Node ant, Node conc, const char* c);
+  /**
+   * Indicates that conc should be added to the equality engine of this class
+   * with explanation eq_exp. It must be the case that eq_exp is a (conjunction
+   * of) literals that each are explainable, i.e. they already hold in the
+   * equality engine of this class.
+   */
+  void sendInfer(Node eq_exp, Node eq, const char* c);
```
4tXJ7f: Why are the two methods above `protected` instead of `private`?

src/theory/strings/output_channel.h
```diff
@@ -0,0 +1,308 @@
...
+   * This is a reference to the output channel of the theory of strings.
+   */
+  OutputChannel& d_out;
+  /** Common constants */
+  Node d_true;
+  Node d_false;
+  /** The list of pending literals to assert to the equality engine */
+  std::vector<Node> d_pending;
+  /** A map from the literals in the above vector to their explanation */
+  std::map<Node, Node> d_pending_exp;
```
4tXJ7f: Since this is a new class, can we rename the variables to be camelCase to follow the style guide?

src/theory/strings/output_channel.h
```diff
@@ -0,0 +1,308 @@
...
+   * For example, say that v1 was introduced as a proxy variable for "ABC", and
+   * v2 was introduced as a proxy variable for "AA".
+   *
+   * Given the input n := v1 = "ABC" ^ v2 = x ^ x = "AA", this method sets:
+   * vars = { x },
+   * subs = { v2 },
+   * unproc = {}.
+   * In particular, since says that the information content of n essentially
+   * says that x = v2, while the first and third conjunctions can be dropped
+   * since they effectively give the definition of the proxy variables.
```
4tXJ7f: The sentence above does not seem to parse for me.

src/theory/strings/output_channel.cpp
```diff
@@ -0,0 +1,448 @@
...
+      Trace("strings-pending") << "Require phase : " << prp.first
+                               << ", polarity = " << prp.second << std::endl;
+      d_out.requirePhase(prp.first, prp.second);
+    }
+  }
+  d_pending_lem.clear();
+  d_pending_req_phase.clear();
+}
+
+bool OutputChannelStrings::hasConflict() const { return d_parent.d_conflict; }
```
4tXJ7f: Wouldn't it be more natural to move `d_parent.d_conflict` to `OutputChannelStrings`?

src/theory/strings/theory_strings_utils.cpp
```diff
@@ -0,0 +1,50 @@
...
+ **/
+
+#include "theory/strings/theory_strings_utils.h"
+
+using namespace CVC4::kind;
+
+namespace CVC4 {
+namespace theory {
+namespace strings {
+namespace utils {
```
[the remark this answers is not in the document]
ajreynol: I was following https://github.com/CVC4/CVC4/blob/master/src/theory/bv/theory_bv_utils.h, although I am not strongly for/against this namespace.

src/theory/strings/theory_strings_utils.h
```diff
@@ -0,0 +1,43 @@
...
+namespace CVC4 {
+namespace theory {
+namespace strings {
+namespace utils {
+
+/**
+ * Make the conjunction of nodes in a. Removes duplicate conjuncts, returns
+ * true if a is empty, and a single literal if a has size 1.
+ */
+Node mkAnd(std::vector<Node>& a);
```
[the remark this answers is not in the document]
ajreynol: I think this is a bigger discussion, that we've debated a lot in the past and won't be easy to quickly resolve. I'd like to revisit this but its out of scope for the PR.

This PR is regardless progress towards this since it is making strings/BV more uniform as they both will now have a utils file.

src/theory/strings/output_channel.h
```diff
@@ -0,0 +1,308 @@
+/*********************                                                        */
+/*! \file output_channel.h
```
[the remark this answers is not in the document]
ajreynol: I've changed the name to inference_manager and the class name to InferenceManager.

src/theory/strings/output_channel.h
```diff
@@ -0,0 +1,308 @@
...
+ * In particular, inferences are given to this class via calls to functions:
+ *
+ * sendInternalInference, sendInference, sendSplit
+ *
+ * This class decides how these calls will be processed, where the main design
+ * decisions it considers are whether the conclusions corresponding to these
+ * calls will be processed as:
+ *
+ * (1) Internally in the strings solver, via calls to the equality engine. We
+ * refer to these literals as "facts",
```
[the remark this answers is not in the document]
ajreynol: What I call "facts" does refer to literals. 

A "conclusion" indeed may be an arbitrary formula, which can be processed to multiple literal "facts".

src/theory/strings/output_channel.cpp
ajreynol: This stat was never really used and is not very important, hence I'd like to remove it for simplicity.

src/theory/strings/output_channel.cpp
ajreynol: Yes, the previous code was a bit nonsensical.  Lemmas and phase requirements essentially should be processed in parallel (phase requirements typically say what polarity to decide for literals that appear in lemmas).  The new code makes it so that we don't do anything if we are in conflict.

src/theory/strings/output_channel.cpp
ajreynol: There would be a circular include.

ajreynol:
I've addressed all points of the review except those that I've commented on. 

I have left the strings utilities in the `strings::utils::` namespace.  My main reason why I like it is that it is a way of marking utility functions in the code. That is, seeing `utils::mkAnd` immediately tells me that its from `strings/theory_strings_utils.h`, whereas if it were a global function in the same namespace it would be a bit more confusing.  I'm also basing this on my experience in the bit-vector theory with theory_bv_utils, where I generally like having the `utils::` prefix there too.

This is up for discussion.

src/theory/strings/theory_strings_utils.cpp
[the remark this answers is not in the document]
4tXJ7f: I would prefer that there is a `utils` folder insides `src/theory/{bv,strings}` in that case but I agree that that discussion is out-of-scope for this PR. We should discuss this though.

src/theory/strings/theory_strings_utils.h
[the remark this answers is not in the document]
4tXJ7f: I tend to agree with this and yes, we should discuss this again. To have a utils file is at least a step forward :)

src/theory/strings/output_channel.h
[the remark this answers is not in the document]
4tXJ7f: That sounds good to me.

src/theory/strings/output_channel.cpp
4tXJ7f: In that case, we should probably get rid of the variable completely.

src/theory/strings/output_channel.cpp
4tXJ7f: Hm, I am not quite following. `theory_strings.h` already includes `inference_manager.h` https://github.com/CVC4/CVC4/pull/3076/files#diff-452ff9fd1723599f909e96f23d6f366dR27, no?

src/theory/strings/output_channel.cpp
4tXJ7f: That's fair, thanks for the explanation.

src/theory/strings/output_channel.h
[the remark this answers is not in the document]
4tXJ7f: Right. In the context, it is still a bit confusing IMHO because the text says: 

> whether the _conclusions_ will be processed [...] We refer to these literals as "facts"

It could be understood that the conclusion has to be a simple literal in case (1), which is AFAIK not true. Maybe we could say:

> We refer to literals in these conclusions as "facts"

src/theory/strings/output_channel.cpp
ajreynol: Yes, so inference_manager cannot include theory_strings.h. Hence, this function needs to be in the cpp of inference_manager.

4tXJ7f: I think there was a bit of a misunderstanding here :) I was suggesting to move the `d_conflict` variable to `InferenceManager` ;) Not too important, we can potentially address this in a later PR.

src/theory/strings/inference_manager.h
```diff
@@ -0,0 +1,292 @@
...
+   * definition of all proxy variables introduced by the theory of strings.
+   *
+   * For example, say that v1 was introduced as a proxy variable for "ABC", and
+   * v2 was introduced as a proxy variable for "AA".
+   *
+   * Given the input n := v1 = "ABC" ^ v2 = x ^ x = "AA", this method sets:
+   * vars = { x },
+   * subs = { v2 },
+   * unproc = {}.
+   * In particular, this says that the information content of n essentially
```
4tXJ7f:
```suggestion
   * In particular, this says that the information content of n is essentially
```

src/theory/strings/output_channel.cpp
ajreynol: Ah.  Yes I considered this.

I think in the end I'm going to split another class off of TheoryStrings, which relates to the "state" of the solver, which includes things like read queries to equality engine and the normal form information.  Otherwise I think the inference manager will become too big of a class.  The intention of this class is to mostly act as a means of performing the "output" of strategies and not means of tracking the state of the solver.

I am following a very similar approach in refactoring sets, see InferenceManager, SetsState (https://github.com/CVC4/CVC4/compare/master...ajreynol:setsCardSplit).

aniemetz:
> I've addressed all points of the review except those that I've commented on.
> 
> I have left the strings utilities in the `strings::utils::` namespace. My main reason why I like it is that it is a way of marking utility functions in the code. That is, seeing `utils::mkAnd` immediately tells me that its from `strings/theory_strings_utils.h`, whereas if it were a global function in the same namespace it would be a bit more confusing. I'm also basing this on my experience in the bit-vector theory with theory_bv_utils, where I generally like having the `utils::` prefix there too.
> 
> This is up for discussion.

I agree. I prefer having util functions in the utils:: namespace. I also prefer to always use `utils::` explicitly in the code (no `using utils::...`).

PR #3081, opened 2019-07-02
Add methods to `CVC4::api::Term` for retrieving an operator

makaimann:
This pull request implements several `has` queries and getters for obtaining the operator of a term. There are three cases:

* `BUILTIN` operator
  * Check with `hasBuiltinOperator()`
  * Get it with `getKind()`
* OpTerm operator, e.g. indexed operator
  * Check with `hasOpTermOperator()`
  * Get it with `getOpTerm()`
* Uninterpreted function operator (operator is a Term)
  * Check with `hasUFOperator()`
  * Get it with `getUF()`

Each getter is guarded with a `CVC4_API_CHECK`, and there's also a `hasOperator()` that returns true iff the term has an operator (of any type).

### DIFF
--- a/src/api/cvc4cpp.h
+++ b/src/api/cvc4cpp.h
@@ -591,6 +591,43 @@ class CVC4_PUBLIC Term
   bool isNull() const;
 
   /**
+   * @return true iff this Term has an operator
+   */
+  bool hasOperator() const;
+
+  /**
+   * @return true iff this Term has a builtin operator
+   *
+   * in this case, getKind() will return the operator
+   */
+  bool hasBuilltinOperator() const;
+
+  /**
+   * @return true iff this Term has an indexed operator
+   */
+  bool hasOpTermOperator() const;
+
+  /**
+   * @return true iff this Term is the result of
+   *         applying an uninterpreted function
+   */
+  bool hasUFOperator() const;
+
+  /**
+   * @return the OpTerm used to create this time
+   *
+   * valid to call when hasOpTermOperator() returns true
+   */
+  OpTerm getOpTerm() const;
+
+  /**
+   * @return the OpTerm used to create this time
+   *
+   * valid to call when hasUFOperator() returns true
+   */
+  Term getUF() const;
+
+  /**
    * Boolean negation.
    * @return the Boolean negation of this term
    */
--- a/src/api/cvc4cpp.h
+++ b/src/api/cvc4cpp.h
@@ -718,7 +718,8 @@ class CVC4_PUBLIC Term
   bool hasOpTermOperator() const;
 
   /**
-   * Returns true if this Term was created with an uninterpreted function application.
+   * Returns true if this Term was created with an
+   * uninterpreted function application.
    * If true, getUFOperator() returns the operator UF.
    * @return true iff this Term is the result of
    *         applying an uninterpreted function
--- a/src/api/cvc4cpp.h
+++ b/src/api/cvc4cpp.h
@@ -524,6 +524,114 @@ struct CVC4_PUBLIC SortHashFunction
 };
 
 /* -------------------------------------------------------------------------- */
+/* OpTerm                                                                     */
+/* -------------------------------------------------------------------------- */
+
+/**
+ * A CVC4 operator term.
+ * An operator term is a term that represents certain operators, instantiated
+ * with its required parameters, e.g., a term of kind BITVECTOR_EXTRACT.
+ */
+class CVC4_PUBLIC OpTerm
+{
+  friend class Solver;
+  friend struct OpTermHashFunction;
+
+ public:
+  /**
+   * Constructor.
+   */
+  OpTerm();
+
+  // !!! This constructor is only temporarily public until the parser is fully
+  // migrated to the new API. !!!
+  /**
+   * Constructor.
+   * @param e the internal expression that is to be wrapped by this term
+   * @return the Term
+   */
+  OpTerm(const CVC4::Expr& e);
+
+  /**
+   * Destructor.
+   */
+  ~OpTerm();
+
+  /**
+   * Syntactic equality operator.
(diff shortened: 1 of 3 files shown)


makaimann:
I don't feel very strongly about the style of this, please let me know if you think it should be implemented differently (e.g. templates), or have different names.

makaimann:
Sorry about all the force-pushes. I originally had built this on top of PR #3073, but then decided to make this independent and ran into a few issues rebasing. It should be working now, but I'll keep an eye on Travis. It's been taking a long time to queue up Travis jobs on my fork so I haven't been getting feedback.

src/api/cvc4cpp.h
```diff
@@ -523,6 +523,114 @@ struct CVC4_PUBLIC SortHashFunction
   size_t operator()(const Sort& s) const;
 };
 
+/* -------------------------------------------------------------------------- */
```
aniemetz: Is there a good reason why you moved this block?

src/api/cvc4cpp.h
```diff
@@ -590,6 +698,43 @@ class CVC4_PUBLIC Term
...
+  /**
+   * @return true iff this Term has an indexed operator
+   */
+  bool hasOpTermOperator() const;
+
+  /**
+   * @return true iff this Term is the result of
+   *         applying an uninterpreted function
+   */
+  bool hasUFOperator() const;
```
aniemetz: The name of this function does not really imply what it actually does. Can we maybe find a better name?

src/api/cvc4cpp.h
```diff
@@ -590,6 +698,43 @@ class CVC4_PUBLIC Term
    */
   bool isNull() const;
 
+  /**
+   * @return true iff this Term has an operator
+   */
+  bool hasOperator() const;
+
+  /**
+   * @return true iff this Term has a builtin operator
```
aniemetz: Can you change the format of the documentation of this and the function below? First description, then parameters (if any), then return value. And start description sentence with upper case.

src/api/cvc4cpp.h
```diff
@@ -590,6 +698,43 @@ class CVC4_PUBLIC Term
...
+  /**
+   * @return true iff this Term is the result of
+   *         applying an uninterpreted function
+   */
+  bool hasUFOperator() const;
+
+  /**
+   * @return the OpTerm used to create this time
+   *
+   * valid to call when hasOpTermOperator() returns true
```
aniemetz: Ditto (documentation)

src/api/cvc4cpp.h
```diff
@@ -590,6 +698,43 @@ class CVC4_PUBLIC Term
...
+   */
+  OpTerm getOpTerm() const;
+
+  /**
+   * @return the OpTerm used to create this time
```

src/api/cvc4cpp.h
```diff
@@ -590,6 +698,43 @@ class CVC4_PUBLIC Term
...
+   *
+   * valid to call when hasUFOperator() returns true
+   */
+  Term getUF() const;
```
aniemetz: The name of this function is not really fitting, as above. Maybe try to find a better name?
Also, can you remind me why we even need this?

src/api/cvc4cpp.h
makaimann: Nevermind, I see what you mean. Fixed with 6d9f9f2e531d824e57f6b0ac6e0f06d86abef3cb.

src/api/cvc4cpp.h
makaimann: Fixed with 6d9f9f2e531d824e57f6b0ac6e0f06d86abef3cb.

src/api/cvc4cpp.h
makaimann: Fixed with 6d9f9f2e531d824e57f6b0ac6e0f06d86abef3cb.

src/api/cvc4cpp.h
aniemetz: Maybe a solution to the naming problem here and above is to switch the order and use these names:

hasOperatorBuiltin
hasOperatorOpTerm
hasOperatorUF

getOperatorBuiltin
getOperatorOpTerm
getOperatorUF

What do you think?

makaimann: Makes sense to me, good idea! The only issue I have with it is that `getOperatorBuiltin` and `getKind` would do the same thing. Do you think it's more important to have a predictable API (e.g. has `getOperatorBuiltin` to match the other versions), or no redundancy in the API (don't include it because `getKind` already does that). 

I guess we could also just remove `hasOperatorBuiltin` and then you'd have to check that both `hasOperatorOpTerm` and `hasOperatorUF` return false to know that it's a builtin operator. What do you think?

src/api/cvc4cpp.h
aniemetz: OK, that's fine.

src/api/cvc4cpp.h
aniemetz: > Makes sense to me, good idea! The only issue I have with it is that `getOperatorBuiltin` and `getKind` would do the same thing. Do you think it's more important to have a predictable API (e.g. has `getOperatorBuiltin` to match the other versions), or no redundancy in the API (don't include it because `getKind` already does that).

Hm, yeah, actually we don't want redundancy if possible. But at the same time we want predictability.  I'm a bit torn, tbh, but leaning towards less redundancy. @mpreiner what do you think?

> I guess we could also just remove `hasOperatorBuiltin` and then you'd have to check that both `hasOperatorOpTerm` and `hasOperatorUF` return false to know that it's a builtin operator. What do you think?

No, I'd keep it.

makaimann:
Temporarily closing for refactoring - will have OpTerm wrap Kind, and UF so that getting an operator always returns an OpTerm.

PR #3085, opened 2019-07-09
Split, refactor and document the theory of sets

ajreynol:
This splits 3 new utilities from theory of sets:
- A "solver state", which manages state information during full effort checks
- An "inference manager", which is the solver's standard interface for adding new facts, lemmas and conflicts.  This is inspired by TheoryStrings.
- The "cardinality extension", which is the porition of the solver that is responsible for the Bansal et al IJCAR 2016 procedure.

The code for sets is now fairly well documented.

The code is heavily refactored, but no major behavior changes were made in this commit.


src/theory/sets/solver_state.h
```diff
@@ -0,0 +1,290 @@
...
+  std::map<Kind, std::vector<Node> > d_op_list;
+  /** the skolem cache */
+  SkolemCache d_skCache;
+  /** is set disequality entailed internal
+   *
+   * This returns true if disequality between sets a and b is entailed in the
+   * current context. We use an incomplete test based on equality and membership
+   * information.
+   *
+   * eqE is the representative of the equivalence class of the empty set
```
mudathirmahgoub: `eqE` is the representative of .... Do you mean parameter `re` ?

src/theory/sets/theory_sets_private.cpp
```diff
@@ -501,45 +288,32 @@ void TheorySetsPrivate::fullEffortCheck(){
...
     d_full_check_incomplete = false;
-    d_set_eqc.clear();
-    d_set_eqc_list.clear();
-    d_eqc_emptyset.clear();
-    d_eqc_univset.clear();
-    d_eqc_singleton.clear();
-    d_congruent.clear();
-    d_nvar_sets.clear();
-    d_var_set.clear();
     d_most_common_type.clear();
```
mudathirmahgoub: Wouldn't it be nice to have a reset function for TheorySetsPrivate similar to SolverState and others ?

src/theory/sets/theory_sets_private.cpp
```diff
@@ -1032,739 +773,33 @@ void TheorySetsPrivate::checkDisequalities( std::vector< Node >& lemmas ) {
   }
 }
 
-void TheorySetsPrivate::checkCardBuildGraph( std::vector< Node >& lemmas ) {
```
mudathirmahgoub: this function is renamed into CardinalityExtension::checkRegister

src/theory/sets/solver_state.h
ajreynol: Yes, I'll fix.

src/theory/sets/theory_sets_private.cpp
ajreynol: Will do.

src/theory/sets/theory_sets_private.cpp
ajreynol: Yes, that's the better name for it after the refactor.

src/theory/sets/cardinality_extension.h
```diff
@@ -0,0 +1,343 @@
...
+                          std::vector<Node>& curr,
+                          std::vector<Node>& exp);
+  /** check normal forms
+   *
+   * This method attempts to assign "normal forms" to all set equivalence
+   * classes whose types have cardinality constraints. Normal forms are
+   * defined recursively.
+   *
+   * A "normal form" of an equivalence class [r] (where [r] denotes the
+   * equivalence class whose representative is r) is a set of representatives
```
yoni206: Isn't it an ordered set? the argument to the function is a vector.

src/theory/sets/cardinality_extension.h
```diff
@@ -0,0 +1,343 @@
...
+   * U = { r1, ..., rn }. If there exists at least one set in [r] that has a
+   * "flat form", then all sets in the equivalence class have flat form U.
+   * If no set in E has a flat form, then U = { r } if r does not contain
```
yoni206: How is `r` chosen in this case?
Who is E?

src/theory/sets/cardinality_extension.h
```diff
@@ -0,0 +1,343 @@
...
+   * form, then the resulting intro_sets is guaranteed to be non-empty.
+   *
+   * As an example, say we have a context with equivalence classes:
+   *   {A, D}, {C, A^B}, {E, C^D}, {C\D}, {D\C}, {A\B}, {empty, B\A},
+   * An ordered list d_oSetEqc for this context:
+   *   A, C, E, C\D, D\C, A\B, empty, ...
+   * The normal form of {empty, B\A} is {}, since it contains the empty set.
+   * The normal forms for each of the singleton equivalence classes are
+   * themselves.
+   * The flat form of each of E and C^D does not exist, hence the normal form
```
yoni206: Why E and not C^D?

src/theory/sets/cardinality_extension.h
```diff
@@ -0,0 +1,343 @@
...
+   * of {E, C^D} is {E}.
+   * The flat form of C is {E, C\D}, noting that C^D and C\D are terms whose
+   * parents are C, hence {E, C\D} is the normal form for class {C, A^B}.
```
yoni206:
```suggestion
   * parent is C, hence {E, C\D} is the normal form for class {C, A^B}.
```

src/theory/sets/cardinality_extension.h
```diff
@@ -0,0 +1,343 @@
...
+  void checkMinCard();
+  /** check cardinality cycles
+   *
+   * The purpose of this inference schema is construct two data structures:
+   *
+   * (1) d_card_parent, which maps set terms (A op B) for op in { \, ^ } to
+   * equivalence class representatives of their "parents", where:
+   *   parent( A ^ B ) = A, B
+   *   parent( A \ B ) = A
+   *   parent( B \ A ) = B
```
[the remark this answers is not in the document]
ajreynol: Why remove this?

src/theory/sets/cardinality_extension.h
ajreynol: E -> U, will fix. Representatives are chosen by the equality engine.

src/theory/sets/cardinality_extension.h
ajreynol: What do you mean?  Neither E nor C^D have a flat form since there are no sets that have them as parents.

src/theory/sets/cardinality_extension.h
ajreynol: No, the order of the normal form doesn't matter.

`intro_sets` doesn't have to do with normal forms, see documentation at bottom of the function.

src/theory/sets/cardinality_extension.h
```diff
@@ -0,0 +1,343 @@
...
+  void check();
+  /** Is model value basic?
+   *
+   * This returns true if equivalence class eqc is a "leaf" in the cardinality
+   * graph. In this case, its model value is simply
+   */
+  bool isModelValueBasic(Node eqc);
+  /** get model elements
+   *
+   * This method updates els so that it is the set of elements that occur in eqc
```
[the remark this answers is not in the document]
ajreynol: I've updated the documentation.

src/theory/sets/cardinality_extension.h
```diff
@@ -0,0 +1,343 @@
...
+   *
+   * This method applies various inference schemas in succession until an
+   * inference is made. These inference schemas are given in the private
+   * methods of this class (e.g. checkMinCard, checkCardBuildGraph, etc.).
+   */
+  void check();
+  /** Is model value basic?
+   *
+   * This returns true if equivalence class eqc is a "leaf" in the cardinality
+   * graph. In this case, its model value is simply
```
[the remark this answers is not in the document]
ajreynol: This sentence is spurious, will remove.

src/theory/sets/cardinality_extension.h
```diff
@@ -0,0 +1,343 @@
...
+   * This method is inspired by the checkCycles inference schema in the theory
+   * of strings.
+   */
+  void checkCardCycles();
+  /**
+   * Helper function for above. Called when wish to process equivalence class
+   * eqc.
+   *
+   * Argument curr contains the equivalence classes we are currently processing,
+   * which are descendants of eqc in the cardinality graph.
```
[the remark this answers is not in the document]
ajreynol: Yes.

src/theory/sets/cardinality_extension.h
yoni206: Why is the normal form of {E,C^D} is {E} and not {C^D}?

ajreynol: I am assuming the first term in each set is the representative, will update the documentation to say this.

src/theory/sets/inference_manager.h
```diff
@@ -0,0 +1,159 @@
...
+   * Called at the beginning of a full effort check. Resets the information
+   * related to this class regarding whether facts and lemmas have been
+   * processed.
+   */
+  void reset();
+  /**
+   * Add facts corresponding to ( exp => fact ) via calls to the assertFact
+   * method of TheorySetsPrivate.
+   *
+   * The argument lemmas is updated to contain poritions of fact that were
```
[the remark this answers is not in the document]
ajreynol: This was refactored, will fix.

src/theory/sets/cardinality_extension.h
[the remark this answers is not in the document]
ajreynol: I guess that's one way to look at this; we tend to think in terms of those 3 base nodes but I guess that is spurious in this context.

src/theory/sets/theory_sets_private.cpp
```diff
@@ -404,142 +276,49 @@ bool TheorySetsPrivate::assertFact( Node fact, Node exp ){
...
 void TheorySetsPrivate::fullEffortCheck(){
   Trace("sets") << "----- Full effort check ------" << std::endl;
   do{
     Trace("sets") << "...iterate full effort check..." << std::endl;
-    Assert( d_equalityEngine.consistent() );
-    d_sentLemma = false;
-    d_addedFact = false;
-    d_full_check_incomplete = false;
-    d_set_eqc.clear();
-    d_set_eqc_list.clear();
```
mudathirmahgoub: All other variables are reset in the new function fullEffortReset except `d_set_eqc_list` which I can not find in the new version. Is it removed or renamed ?

ajreynol: This was an unused variable, I've removed it.

src/theory/sets/cardinality_extension.cpp
```diff
@@ -0,0 +1,911 @@
...
+    registerCardinalityTerm(n[0]);
+  }
+  if (tnc.isInterpretedFinite())
+  {
+    std::stringstream ss;
+    ss << "ERROR: cannot use cardinality on sets with finite element "
+          "type (term is "
+       << n << ")." << std::endl;
+    throw LogicException(ss.str());
+    // TODO (#1123): extend approach for this case
```
mudathirmahgoub: Interesting. I didn't know the cardinality in the example below is not supported. 
```
(set-logic ALL_SUPPORTED)
(set-info :status unsat)
(declare-fun S () (Set Bool))
(declare-fun x () Int)
(assert (= (card S) 3))
(check-sat)
```
```
(error "ERROR: cannot use cardinality on sets with finite element type (term is (card S)).")
```

src/theory/sets/theory_sets_private.cpp
```diff
@@ -404,142 +276,49 @@ bool TheorySetsPrivate::assertFact( Node fact, Node exp ){
...
-    d_members_index.clear();
-    d_singleton_index.clear();
-    d_bop_index.clear();
-    d_op_list.clear();
-    d_card_enabled = false;
-    d_t_card_enabled.clear();
-    d_rels_enabled = false;
-    d_eqc_to_card_term.clear();
-
-    std::vector< Node > lemmas;
```
mudathirmahgoub: Just to confirm. This local variable `lemmas` is replaced with `InferenceManager::d_pendingLemmas`. Right?

src/theory/sets/theory_sets_private.cpp
```diff
@@ -672,123 +387,132 @@ void TheorySetsPrivate::fullEffortCheck(){
       }
     }
 
-    flushLemmas( lemmas );
```
mudathirmahgoub: Why isn't this line replaced with `d_im.flushPendingLemmas`?

src/theory/sets/cardinality_extension.cpp
ajreynol: Yep, see https://github.com/CVC4/CVC4/issues/1123.  We should implement support for this.

src/theory/sets/theory_sets_private.cpp
ajreynol: Yes, that's right. It's easier this way since we don't have to pass that local variable around everywhere.

src/theory/sets/theory_sets_private.cpp
ajreynol: We now insist that lemmas are flushed locally by the modules that generate them.  See e.g. https://github.com/CVC4/CVC4/pull/3085/files#diff-afb263928d078d3259ca8f32785fa402R177

src/theory/sets/theory_sets_private.cpp
```diff
@@ -672,123 +387,132 @@ void TheorySetsPrivate::fullEffortCheck(){
...
-                      debugPrintSet(intro_sets[0], "sets-nf");
-                      Trace("sets-nf") << std::endl;
-                      Node k = getProxy(intro_sets[0]);
-                      d_sentLemma = true;
-                    }
-                  }
-                }
-              }
+              // call the check method of the cardinality solver
+              d_cardSolver->check();
```
mudathirmahgoub: What is the rationale of removing all occurrences of `flushLemmas( lemmas );` in the code block above from the function `d_cardSolver::check`?

src/theory/sets/theory_sets_private.cpp
```diff
@@ -798,23 +522,27 @@ void TheorySetsPrivate::checkDownwardsClosure( std::vector< Node >& lemmas ) {
...
                 std::vector< Node > exp;
-                if( ee_areEqual( mem, pmem ) ){
+                if (d_state.areEqual(mem, pmem))
+                {
                   exp.push_back( pmem );
-                }else{
+                }
+                else
+                {
                   nmem = NodeManager::currentNM()->mkNode( kind::OR, pmem.negate(), nmem );
```
mudathirmahgoub: This function can benefit from the common used  shorthand `NodeManager* nm = NodeManager::currentNM();`

ajreynol: I agree, but that part of the code didn't change.  Trying to keep the changes in this PR from encompassing the entire solver :)

src/theory/sets/theory_sets_private.cpp
```diff
@@ -824,160 +552,176 @@ void TheorySetsPrivate::checkDownwardsClosure( std::vector< Node >& lemmas ) {
...
             TypeNode tn = NodeManager::currentNM()->mkSetType( e.getType() );
             Node u;
             std::map< TypeNode, Node >::iterator itu = univ_set.find( tn );
             if( itu==univ_set.end() ){
-              std::map< TypeNode, Node >::iterator itu = d_eqc_univset.find( tn );
+              Node ueqc = d_state.getUnivSetEqClass(tn);
               // if the universe does not yet exist, or is not in this equivalence class
-              if( itu==d_eqc_univset.end() || itu->second!=it->first ){
-                u = getUnivSet( tn );
+              if (itu->second != ueqc)
```
mudathirmahgoub: When the execution reaches this line, the condition  `itu == d_eqc_univset.end() ` evaluates to `true`. Is it safe to dereference the iterator `it` in `itu->second != ueqc`? Is it possible to replace this condition with `ueqc.isNull()`?

src/theory/sets/theory_sets_private.h
```diff
@@ -32,51 +34,73 @@ namespace theory {
...
   void eqNotifyPreMerge(TNode t1, TNode t2);
   void eqNotifyPostMerge(TNode t1, TNode t2);
   void eqNotifyDisequal(TNode t1, TNode t2, TNode reason);
-private:
-  bool ee_areEqual( Node a, Node b );
-  bool ee_areDisequal( Node a, Node b );
-  bool ee_areCareDisequal( Node a, Node b );
+  /** Assert fact holds in the current context with explanation exp.
+   *
+   * exp should be explanable by the equality engine of this class, and fact
```
mudathirmahgoub: typo `explainable`

src/theory/sets/theory_sets_private.cpp
```diff
@@ -1774,13 +814,6 @@ void TheorySetsPrivate::check(Theory::Effort level) {
   Trace("sets-check") << "Sets finish Check effort " << level << std::endl;
 }/* TheorySetsPrivate::check() */
 
-bool TheorySetsPrivate::needsCheckLastEffort() {
```
mudathirmahgoub: I didn't find any equivalent code for this. Why is this function no longer needed?

src/theory/sets/theory_sets_private.cpp
```diff
@@ -1774,13 +814,6 @@ void TheorySetsPrivate::check(Theory::Effort level) {
...
-  if( !d_var_elim.empty() ){
-    return true;
-  }
-  return false;
-}
-
```
mudathirmahgoub: typo in the function below `TheorySetsPrivate`
```
void TheorySetsPrivate::addSharedTerm(TNode n) {
  Debug("sets") << "[sets] ThoerySetsPrivate::addSharedTerm( " << n << ")" << std::endl;
  d_equalityEngine.addTriggerTerm(n, THEORY_SETS);
}
```

src/theory/sets/theory_sets_private.cpp
```diff
@@ -1978,50 +1020,41 @@ bool TheorySetsPrivate::collectModelInfo(TheoryModel* m)
...
-        std::map< Node, std::map< Node, Node > >::iterator itm = d_pol_mems[0].find( eqc );
-        if( itm!=d_pol_mems[0].end() ){
-          for( std::map< Node, Node >::iterator itmm = itm->second.begin(); itmm != itm->second.end(); ++itmm ){
-            Node t = NodeManager::currentNM()->mkNode( kind::SINGLETON, itmm->first );
+        const std::map<Node, Node>& emems = d_state.getMembers(eqc);
+        if (!emems.empty())
+        {
+          for (const std::pair<const Node, Node>& itmm : emems)
+          {
+            Node t = nm->mkNode(kind::SINGLETON, itmm.first);
```
mudathirmahgoub: Suggestion: Why don't we construct sets using `insert` instead of `union` which is more verbose? compare this
```
(define-fun S () (Set Int) (insert 1 2 3 (singleton 4)))
```
versus
```
(define-fun S () (Set Int) (union (union (union (singleton 1) (singleton 2)) (singleton 3)) (singleton 4)))
```

src/theory/sets/theory_sets_private.cpp
ajreynol: This function was written for an experimental piece of code that was never integrated (check at LAST_CALL effort in sets was a no-op https://github.com/CVC4/CVC4/pull/3085/files/1a1c83e4667af47eecfe62a779c054e7f6264358#diff-aaa774b6c62e6fbcec9f9c49f7d60a3eL1725); this PR cleans this function up.

src/theory/sets/theory_sets_private.cpp
ajreynol: `insert` is syntax sugar. If we insisted that the rewritten form involved `insert`, then our decision procedures would have to handle `union` and `insert`, thus making the code for handling set union twice as complicated.

src/theory/sets/theory_sets_private.cpp
ajreynol: Yes, good catch.

ajreynol: I actually meant to change this to `ueqc != it->first`.

src/theory/sets/theory_sets_private.cpp
```diff
@@ -2133,58 +1117,40 @@ bool TheorySetsPrivate::propagate(TNode literal) {
...
 {
-  d_conflictNode = explain(a.eqNode(b));
-  d_external.d_out->conflict(d_conflictNode);
-  Debug("sets") << "[sets] conflict: " << a << " iff " << b
-                << ", explaination " << d_conflictNode << std::endl;
-  Trace("sets-lemma") << "Equality Conflict : " << d_conflictNode << std::endl;
-  d_conflict = true;
+  Node conf = explain(a.eqNode(b));
+  d_state.setConflict(conf);
+  Debug("sets") << "[sets] conflict: " << a << " iff " << b << ", explaination "
```
mudathirmahgoub: typo `explanation`

src/theory/sets/theory_sets_private.cpp
```diff
@@ -2285,25 +1251,10 @@ Theory::PPAssertStatus TheorySetsPrivate::ppAssert(TNode in, SubstitutionMap& ou
...
-  if( status==Theory::PP_ASSERT_STATUS_SOLVED ){
-    Trace("sets-var-elim") << "Sets : ppAssert variable eliminated : " << in << ", var = " << var << std::endl;
-    /*
-    if( var.getType().isSet() ){
-      //we must ensure that subs is included in the universe set
-      d_var_elim[var] = true;
-    } 
-    */
-    if( options::setsExt() ){
-      Assert( !var.getType().isSet() );
```
mudathirmahgoub: Since this assertion is removed, the local variable `var` is no longer needed, and it can be removed (as CLion says).

src/theory/sets/theory_sets_rels.cpp
```diff
@@ -32,24 +32,22 @@ typedef std::map< Node, std::unordered_set< Node, NodeHashFunction > >::iterator
...
-    : d_eqEngine(eq),
-      d_sets_theory(set),
-      d_trueNode(NodeManager::currentNM()->mkConst<bool>(true)),
-      d_falseNode(NodeManager::currentNM()->mkConst<bool>(false)),
-      d_shared_terms(u),
-      d_satContext(c)
+TheorySetsRels::TheorySetsRels(SolverState& s,
+                               InferenceManager& im,
+                               eq::EqualityEngine& e,
+                               context::Context* c,
```
mudathirmahgoub: CLion detected that this argument `c` is never used,  and it can be removed from the constructor. Previously it was used to initialize the deleted field `d_satContext` which was never used.

src/theory/sets/theory_sets_rels.cpp
```diff
@@ -32,24 +32,22 @@ typedef std::map< Node, std::unordered_set< Node, NodeHashFunction > >::iterator
...
 {
-  d_eqEngine->addFunctionKind(PRODUCT);
-  d_eqEngine->addFunctionKind(JOIN);
-  d_eqEngine->addFunctionKind(TRANSPOSE);
-  d_eqEngine->addFunctionKind(TCLOSURE);
-  d_eqEngine->addFunctionKind(JOIN_IMAGE);
-  d_eqEngine->addFunctionKind(IDEN);
-  d_eqEngine->addFunctionKind(APPLY_CONSTRUCTOR);
+  d_trueNode = NodeManager::currentNM()->mkConst(true);
+  d_falseNode = NodeManager::currentNM()->mkConst(false);
```
mudathirmahgoub: `d_trueNode` and `d_falseNode` are redundant. They are already privately defined in `TheorySetsPrivate` as `d_true` and `d_false`. How about making them public constants in `TheorySetsPrivate` or better move them to  `Node` class which I think is is the ideal place for them?

src/theory/sets/theory_sets_rels.cpp
```diff
@@ -1058,49 +1056,72 @@ void TheorySetsRels::check(Theory::Effort level)
...
+      Node lem = NodeManager::currentNM()->mkNode(OR, exp.negate(), conc);
+      d_im.assertInference(lem, d_trueNode, c, 1);
+      return;
+    }
+    // try to assert it as a fact
+    d_im.assertInference(conc, exp, c);
+  }
+
   bool TheorySetsRels::isRelationKind( Kind k ) {
     return k == kind::TRANSPOSE || k == kind::PRODUCT || k == kind::JOIN || k == kind::TCLOSURE;
```
mudathirmahgoub: Why are Kinds `JOIN_IMAGE, IDEN, APPLY_CONSTRUCTOR` not included in this function?

src/theory/sets/theory_sets_rels.cpp
```diff
@@ -32,24 +32,22 @@ typedef std::map< Node, std::unordered_set< Node, NodeHashFunction > >::iterator
...
-      d_sets_theory(set),
-      d_trueNode(NodeManager::currentNM()->mkConst<bool>(true)),
-      d_falseNode(NodeManager::currentNM()->mkConst<bool>(false)),
-      d_shared_terms(u),
-      d_satContext(c)
+TheorySetsRels::TheorySetsRels(SolverState& s,
+                               InferenceManager& im,
+                               eq::EqualityEngine& e,
+                               context::Context* c,
+                               context::UserContext* u)
```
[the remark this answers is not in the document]
ajreynol: I hadn't really thought about it.  I think in my head I was modelling reference vs pointer after what is done for `Theory`: https://github.com/CVC4/CVC4/blob/master/src/theory/theory.h#L197, where state / output channels are references and contexts are pointers.

One minor advantage of references is that `.` is one character shorter than `->` :).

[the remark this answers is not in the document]
4tXJ7f: Yeah, I guess it is fine for now but we should somehow find a consistent style in the future IMHO :)

PR #3091, opened 2019-07-15
 Add string rewrite to distribute character stars over concatenation

### LINKED ISSUE #3090: Performance issue on solving assertion with regexps
Hello,

I'm used **cvc4 v1.7** (besides, behavior is reproduced on the older versions of **cvc4**) with the options: ` -L smt2.5 --output-lang smt2.5 --produce-models --strings-exp --tlimit-per 1000000 --fmf-fun-rlv` and try to solve the simple assertion which contains restrictions as regular expression as are shown on the below:

```
(set-logic ALL_SUPPORTED)
(declare-const id String)
(assert (and (str.in.re id (re.+ (re.range "0" "9"))) (str.contains id "value")))
(check-sat)
```
But the solving is stopped by timeout (1000 seconds) and the result is `unknown` when I expected that the solving was ended at much less time and with a correct answer.

Additionally, **cvc4 v1.5** also had ended solving by timeout, but it took only 250mb of RAM, when **cvc4 v1.7** took 1800mb.

If I miss something, please let me know about it.

ajreynol:
Fixes #3090.

### DIFF
--- a/src/theory/strings/theory_strings_rewriter.cpp
+++ b/src/theory/strings/theory_strings_rewriter.cpp
@@ -1231,9 +1231,31 @@ Node TheoryStringsRewriter::rewriteMembership(TNode node) {
         }
       }
     }
+    else if( x.getKind() == STRING_CONCAT )
+    {
+      // (str.in.re (str.++ x1 x1) (str.* R)) -->
+      //   (str.in.re x1 (re.* R)) AND (str.in.re x2 (re.* R))
+      //     if the length of all strings in R is one
+      Node flr = getFixedLengthForRegexp(r[0]);
+      if( !flr.isNull() )
+      {
+        Node one = nm->mkConst( Rational( 1 ) );
+        if( flr==one )
+        {
+          std::vector< Node > conj;
+          for( const Node& xc : x )
+          {
+            conj.push_back(nm->mkNode(STRING_IN_REGEXP,xc,r));
+          }
+          Node retNode = nm->mkNode(AND, conj);
+          return returnRewrite(node, retNode, "re-in-dist-char-star");          
+        }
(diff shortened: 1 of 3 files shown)


src/theory/strings/theory_strings_rewriter.cpp
```diff
@@ -1231,9 +1231,31 @@ Node TheoryStringsRewriter::rewriteMembership(TNode node) {
...
+      // (str.in.re (str.++ x1 ... xn) (str.* R)) -->
+      //   (str.in.re x1 (re.* R)) AND ... AND (str.in.re xn (re.* R))
+      //     if the length of all strings in R is one.
+      Node flr = getFixedLengthForRegexp(r[0]);
+      if (!flr.isNull())
+      {
+        Node one = nm->mkConst(Rational(1));
+        if (flr == one)
+        {
+          std::vector<Node> conj;
```
4tXJ7f: Could use a `NodeBuilder` instead of an `std::vector`.

src/theory/strings/theory_strings_rewriter.cpp
```diff
@@ -1231,9 +1231,31 @@ Node TheoryStringsRewriter::rewriteMembership(TNode node) {
         }
       }
     }
+    else if (x.getKind() == STRING_CONCAT)
+    {
+      // (str.in.re (str.++ x1 ... xn) (str.* R)) -->
```
GeorgeAlexandria: @ajreynol, `str.* R` => `re.* R` or  did I miss something?

ajreynol: You're right, that was a typo. Will fix.

PR #3092, opened 2019-07-15
Add support for str.tolower and str.toupper

ajreynol:
Basic support for str.tolower and str.toupper. Rewriter is constant evaluator, reduction is the expected one using str.code.

### DIFF
--- a/src/parser/smt2/smt2.cpp
+++ b/src/parser/smt2/smt2.cpp
@@ -169,8 +169,11 @@ void Smt2::addStringOperators() {
   addOperator(kind::STRING_STRIDOF, "str.indexof" );
   addOperator(kind::STRING_STRREPL, "str.replace" );
   addOperator(kind::STRING_STRREPLALL, "str.replaceall");
-  addOperator(kind::STRING_TOLOWER, "str.tolower");
-  addOperator(kind::STRING_TOUPPER, "str.toupper");
+  if (!strictModeEnabled())
+  {
+    addOperator(kind::STRING_TOLOWER, "str.tolower");
+    addOperator(kind::STRING_TOUPPER, "str.toupper");
+  }
   addOperator(kind::STRING_PREFIX, "str.prefixof" );
   addOperator(kind::STRING_SUFFIX, "str.suffixof" );
   // at the moment, we only use this syntax for smt2.6.1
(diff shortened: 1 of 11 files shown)


src/theory/strings/kinds
```diff
@@ -28,6 +28,8 @@ operator STRING_SUFFIX 2 "string suffixof"
 operator STRING_ITOS 1 "integer to string"
 operator STRING_STOI 1 "string to integer (total function)"
 operator STRING_CODE 1 "string to code, returns the code of the first character of the string if it has length one, -1 otherwise"
+operator STRING_TOLOWER 1 "string to lowercase conversion"
+operator STRING_TOUPPER 1 "string to lowercase conversion"
```
4tXJ7f:
```suggestion
operator STRING_TOUPPER 1 "string to uppercase conversion"
```

src/parser/smt2/smt2.cpp
```diff
@@ -169,6 +169,8 @@ void Smt2::addStringOperators() {
   addOperator(kind::STRING_STRIDOF, "str.indexof" );
   addOperator(kind::STRING_STRREPL, "str.replace" );
   addOperator(kind::STRING_STRREPLALL, "str.replaceall");
+  addOperator(kind::STRING_TOLOWER, "str.tolower");
+  addOperator(kind::STRING_TOUPPER, "str.toupper");
```
4tXJ7f: Guard with `if (!strictModeEnabled())`

src/theory/strings/theory_strings_rewriter.cpp
```diff
@@ -2974,6 +2978,41 @@ Node TheoryStringsRewriter::rewriteReplaceInternal(Node node)
   return Node::null();
 }
 
+Node TheoryStringsRewriter::rewriteStrConvert(Node node)
+{
+  if (node[0].isConst())
```
4tXJ7f: Add assertion on kinds.

src/theory/strings/theory_strings_rewriter.cpp
```diff
@@ -2974,6 +2978,41 @@ Node TheoryStringsRewriter::rewriteReplaceInternal(Node node)
...
+        {
+          newChar = newChar + 32;
+        }
+      }
+      newChar = CVC4::String::convertCodeToUnsignedInt(newChar);
+      nvec[i] = newChar;
+    }
+    Node retNode = NodeManager::currentNM()->mkConst(String(nvec));
+    return returnRewrite(node, retNode, "str-conv-const");
+  }
```
4tXJ7f: Note: another basic rewrite that we should do is to distribute `str.tolower`/`str.toupper` over `str.++`. We can do that in a future PR though.

PR #3094, opened 2019-07-17
Get operators in node

yoni206:
This PR adds a function to `node_algorithm.{h,cpp}` that returns the operators that occur in a given node.
The operators are mapped according to their type.
This is needed for a future PR for sygus-based interpolation.
In that future PR, the set of operators of the input formulas needs to be computed in order to generate the right grammar for the corresponding sygus problem.

The current PR also includes a new unit test file for `node_algorithm`, including a test for the newly added function, as well as two test for an existing function.


src/expr/node_algorithm.h
```diff
@@ -91,6 +92,19 @@ void getSymbols(TNode n, std::unordered_set<Node, NodeHashFunction>& syms);
...
                 std::unordered_set<TNode, TNodeHashFunction>& visited);
+
+/**
+ * For term n, this function collects the operators that occur in n.
+ * @param n The node under investigation
+ * @param syms The map (from each type to operators of that type) which the operators of
+ * n are added to
+ */
+void getOperatorsMap(TNode n, std::map<TypeNode, std::unordered_set<Node, NodeHashFunction>>& ops);
+/** Same as above, with a visited cache */
```
4tXJ7f: It is better to write out the comment IMHO (in case we generate the documentation one day).

src/expr/node_algorithm.cpp
```diff
@@ -304,6 +304,48 @@ void getSymbols(TNode n,
...
+    visit.pop_back();
+    //if cur is in the cache, do nothing
+    if (visited.find(cur) == visited.end())
+    {
+      //fetch the correct type and add it to the resulting map if needed
+      TypeNode tn = cur.getType();
+      if (ops.find(tn) == ops.end())
+      {
+        ops[tn] = std::unordered_set<Node, NodeHashFunction>();
+      }
```
4tXJ7f: This whole if statement above is not needed. When you do `ops[tn].insert(NodeManager::currentNM()->operatorOf(cur.getKind()))` below, the `std::unordered_set` will automatically be created when `ops[tn]` is executed.

src/expr/node_algorithm.h
4tXJ7f: The other option would be to explicitly refer to the name of the other function.

PR #3095, opened 2019-07-18
Basic rewrites for tolower/toupper

ajreynol:


src/theory/strings/theory_strings_rewriter.cpp
```diff
@@ -3008,9 +3009,27 @@ Node TheoryStringsRewriter::rewriteStrConvert(Node node)
...
     return returnRewrite(node, retNode, "str-conv-const");
   }
+  else if (node[0].getKind() == STRING_CONCAT)
+  {
+    std::vector<Node> cc;
+    for (const Node& nc : node[0])
+    {
+      cc.push_back(nm->mkNode(nk, nc));
+    }
+    Node retNode = nm->mkNode(STRING_CONCAT, cc);
```
4tXJ7f: Use `NodeBuilder` :)

src/theory/strings/theory_strings_rewriter.cpp
```diff
@@ -3008,9 +3009,27 @@ Node TheoryStringsRewriter::rewriteStrConvert(Node node)
...
+    return returnRewrite(node, retNode, "str-conv-minscope-concat");
+  }
+  else if (node[0].getKind() == STRING_TOLOWER
+           || node[0].getKind() == STRING_TOUPPER)
+  {
+    // tolower( tolower( x ) ) = tolower( x )
+    // tolower( toupper( x ) ) = tolower( x )
+    Node retNode = nm->mkNode(nk, node[0][0]);
+    return returnRewrite(node, retNode, "str-conv-idem");
```
4tXJ7f: Another basic rewrite would be: `tolower/toupper(int.to.str(n)) ---> int.to.str(n)` (maybe future PR).

src/theory/strings/theory_strings_rewriter.cpp
```diff
@@ -3008,9 +3009,27 @@ Node TheoryStringsRewriter::rewriteStrConvert(Node node)
...
+  else if (node[0].getKind() == STRING_CONCAT)
+  {
+    std::vector<Node> cc;
+    for (const Node& nc : node[0])
+    {
+      cc.push_back(nm->mkNode(nk, nc));
+    }
+    Node retNode = nm->mkNode(STRING_CONCAT, cc);
+    return returnRewrite(node, retNode, "str-conv-minscope-concat");
+  }
```
4tXJ7f: Another normalization that we should do is to decide between `str.tolower(str.substr(x, n, m))` and `str.substr(str.tolower(x), n, m)`. Pushing `str.tolower` out is maybe better because you get `m` as a bound for the quantifier of `str.tolower`?

src/theory/strings/theory_strings_rewriter.cpp
ajreynol: Ah, yeah, thats a good point, I'll add.

src/theory/strings/theory_strings_rewriter.cpp
ajreynol: This is tricky, pushing to inner also has advantages: say I have multiple `tolower(substr(x,i1,n1)) ... tolower(substr(x,im,nm))`, then pushing inwards makes it so that I only have a single application of `tolower`.

PR #3100, opened 2019-07-18
SyGuS grammar refactor

yoni206:
This PR refactors the code that generates SyGuS grammars.
These refactors are needed for a future PR that introduces SyGuS-based interpolation.
It makes the following changes:

1. In `mkSygusDefaultType` (and as a consequence, also in `mkSygusDefaultGrammar`), the arguments `extra_cons` and `exclude_cons` used to be maps from `TypeNode`s to vectors of `Node`s. 
The first includes operators that need to be included in the grammar, even if they are not in the default grammar.
The second includes operators that should be excluded from the grammar, even if they are in the default grammar.
For each type, the maps included a vector of nodes that represent operators. When generating such maps, however, it is more convenient to create a set rather than a vector, in order to avoid duplicates. Therefore, I have changed these two maps, and now they are from `TypeNode` to an unordered set of Nodes.
2. In the same functions, I have added another parameter: `include_cons`. This includes operators that should be included in the grammar, only if they are in the default grammar (unlike `extra_cons` that adds operators regardless of their inclusion in the default grammar).

Both changes required a few changes in places that call these functions, or that use these parameters.


src/theory/quantifiers/sygus/sygus_grammar_cons.h
```diff
@@ -63,30 +63,42 @@ class CegGrammarConstructor
...
  bool isSyntaxRestricted() { return d_is_syntax_restricted; }
  /** make the default sygus datatype type corresponding to builtin type range
- *   bvl is the set of free variables to include in the grammar
- *   fun is for naming
- *   extra_cons is a set of extra constant symbols to include in the grammar,
- *   exclude_cons is used to exclude operators from the grammar,
- *   term_irrelevant is a set of terms that should not be included in the
- *      grammar.
- */
+  *   bvl is the set of free variables to include in the grammar
```
ajreynol: Can you bullet point this description somehow so that he arguments can each be seen clearly?

src/theory/quantifiers/sygus/sygus_grammar_cons.h
```diff
@@ -63,30 +63,42 @@ class CegGrammarConstructor
...
+  *   fun is for naming
+  *   extra_cons is a set of extra constant symbols to include in the grammar,
+  *   regardless of their inclusion in the default grammar.
+  *   exclude_cons is used to exclude operators from the grammar,
+  *   term_irrelevant is a set of terms that should not be included in the
+  *      grammar.
+  *   include_cons is a set of operators such that if this set is not empty,
```
ajreynol: Can you clarify this point? Do you mean that if include_cons is non-empty, then it defines the complete set of operators for that grammar?

PR #3104, opened 2019-07-19
Eliminate BITVECTOR_ITE in bv_to_bool preprocessing pass

yoni206:
Currently the `bv_to_bool` preprocessing pass does not eliminate `BITVECTOR_ITE` nodes.
However, in a future PR that will introduce the new `bv_to_int` preprocessing pass, it is convenient for me to assume that such nodes do not appear.

This PR translates `BITVECTOR_ITE` nodes into "regular" `ITE` nodes.

While working on a unit test for this feature, I noticed that the unit test for the `bv_gauss` preprocessing pass does not run with `make units`. This PR fixes that as well.


test/unit/preprocessing/pass_bv_to_bool_white.h
```diff
@@ -0,0 +1,86 @@
+/*********************                                                        */
+/*! \file pass_bv_to_bool_white.h
+ ** \verbatim
+ ** Top contributors (to current version):
+ **   Aina Niemetz, Andres Noetzli
```
mpreiner: Shouldn't that be Yoni Zohar?

yoni206: correct, fixed.

src/preprocessing/passes/bv_to_bool.cpp
```diff
@@ -171,9 +169,15 @@ Node BVToBool::convertBvTerm(TNode node)
...
-  if (kind == kind::ITE)
+  if (kind == kind::ITE || kind == kind::BITVECTOR_ITE)
   {
-    Node cond = liftNode(node[0]);
+    Node cond = node[0];
+    // Translate bit-vector condition to an equality to (_ bv1 1)
+    if (kind == kind::BITVECTOR_ITE)
+    {
+      cond = NodeManager::currentNM()->mkNode(kind::EQUAL, cond, d_one);
+    }
```
mpreiner: What happens if we have e.g.,

```
(BITVECTOR_ITE (BITVECTOR_COMP ...) x y)
```

`BITVECTOR_COMP` gets converted to `EQUAL`. Can it happen that you will compare an `EQUAL` against  `d_one`?

yoni206:
Eliminating `BITVECTOR_ITE` is possible, but a bit more complex than the other operators, as the type of `BITVECTOR_ITE` terms may be different than `(_ BitVec 1)`. Since the only known application of this elimination is my future preprocessing pass, I will just implement it there.

PR #3105, opened 2019-07-19
Refactoring of bit-vector elimination rules

yoni206:
This PR makes the following minor refactors to `src/theory/bv/theory_bv_rewrite_rules_operator_elimination.h`:
1. Including `options/bv_options.h`: this is needed because this header file is being used. 
2. Marking all functions as `inline`: details in a discussion inside the PR.


src/theory/bv/theory_bv_rewrite_rules_operator_elimination.h
```diff
@@ -163,12 +172,45 @@ Node RewriteRule<CompEliminate>::apply(TNode node)
...
+
+template <>
+inline Node RewriteRule<AshrEliminate>::apply(TNode node)
+{
+  Debug("bv-rewrite") << "RewriteRule<AshrEliminate>(" << node << ")"
+                      << std::endl;
+  NodeManager* nm = NodeManager::currentNM();
+  Node s = node[0];
+  Node t = node[1];
+  /*  From smtlib:
```
mpreiner:
```suggestion
  /*  From SMT-LIB:
```

src/theory/bv/theory_bv_rewrite_rules_operator_elimination.h
```diff
@@ -163,12 +172,45 @@ Node RewriteRule<CompEliminate>::apply(TNode node)
...
+   *  (bvashr s t) abbreviates
+        (ite (= ((_ extract |m-1| |m-1|) s) #b0)
+             (bvlshr s t)
+             (bvnot (bvlshr (bvnot s) t)))
+   *
```
mpreiner: Prefix all the comment lines with `*`

src/theory/bv/theory_bv_rewrite_rules_operator_elimination.h
```diff
@@ -163,12 +172,45 @@ Node RewriteRule<CompEliminate>::apply(TNode node)
...
+  return (node.getKind() == kind::BITVECTOR_ASHR);
+}
+
+template <>
+inline Node RewriteRule<AshrEliminate>::apply(TNode node)
+{
+  Debug("bv-rewrite") << "RewriteRule<AshrEliminate>(" << node << ")"
+                      << std::endl;
+  NodeManager* nm = NodeManager::currentNM();
+  Node s = node[0];
```
mpreiner:
```suggestion
  TNode s = node[0];
```

src/theory/bv/theory_bv_rewrite_rules_operator_elimination.h
```diff
@@ -163,12 +172,45 @@ Node RewriteRule<CompEliminate>::apply(TNode node)
...
+  Node t = node[1];
```
mpreiner:
```suggestion
  TNode t = node[1];
```

src/theory/bv/theory_bv_rewrite_rules_operator_elimination.h
```diff
@@ -19,20 +19,22 @@
 
 #pragma once
 
+#include "options/bv_options.h"
```
mpreiner: Why do you need this include here?

test/unit/theory/theory_bv_rewriter_white.h
```diff
@@ -100,6 +102,29 @@ class TheoryBvRewriterWhite : public CxxTest::TestSuite
...
+    TS_ASSERT(RewriteRule<AshrEliminate>::applies(fifteen_ashr_one));
+    TS_ASSERT(RewriteRule<AshrEliminate>::applies(three_ashr_one));
+    TS_ASSERT(!RewriteRule<AshrEliminate>::applies(three_lshr_one));
+    Node fifteen_ashr_one_rw =
+        RewriteRule<AshrEliminate>::apply(fifteen_ashr_one);
+    Node three_ashr_one_rw = RewriteRule<AshrEliminate>::apply(three_ashr_one);
+    TS_ASSERT_EQUALS(Rewriter::rewrite(fifteen_ashr_one),
+                     Rewriter::rewrite(fifteen_ashr_one_rw));
+    TS_ASSERT_EQUALS(Rewriter::rewrite(three_ashr_one),
+                     Rewriter::rewrite(three_ashr_one_rw));
```
mpreiner: Not sure how meaningful this test is. Would it make more sense to do a case split on the MSB and check the following two cases:
```
(rewrite (bvashr (concat #b1 s) t)) == (bvnot (bvlshr (bvnot (concat #b1 s)) t))
(rewrite (bvashr (concat #b0 s) t)) == (bvlshr (concat #b0 s) t)
```

src/theory/bv/theory_bv_rewrite_rules_operator_elimination.h
yoni206: Now that I am including `theory_bv_rewrite_rules_operator_elimination.h` in the test file `test/unit/theory/theory_bv_rewriter_white.h`, i needed to added this include so `make units` works. Should I put the inclusion elsewhere?

test/unit/theory/theory_bv_rewriter_white.h
yoni206: Wouldn't the test then be too similar to the rewrite itself? What if we decide at some point to eliminate `bvashr` in some other manner? Then the test will break. Do you think it makes sense to do as you say, and then change the test if we ever change the elimination code?

src/theory/bv/theory_bv_rewrite_rules_operator_elimination.h
```diff
@@ -19,20 +19,22 @@
...
 #include "theory/bv/theory_bv_utils.h"
 
 namespace CVC4 {
 namespace theory {
 namespace bv {
 
-template<>
-bool RewriteRule<UgtEliminate>::applies(TNode node) {
+template <>
+inline bool RewriteRule<UgtEliminate>::applies(TNode node)
```
martin-cs: I appreciate that C++ has become a maze of twisty little passages but it concerns me when the solution to a problem is to add `inline`.  Maybe we should dig a bit deeper into what is going on here.

src/theory/bv/theory_bv_rewrite_rules_operator_elimination.h
```diff
@@ -163,12 +172,43 @@ Node RewriteRule<CompEliminate>::apply(TNode node)
...
+  Debug("bv-rewrite") << "RewriteRule<AshrEliminate>(" << node << ")"
+                      << std::endl;
+  NodeManager* nm = NodeManager::currentNM();
+  TNode s = node[0];
+  TNode t = node[1];
+  /*  From SMT-LIB:
+   *  (bvashr s t) abbreviates
+   *    (ite (= ((_ extract |m-1| |m-1|) s) #b0)
+   *         (bvlshr s t)
+   *         (bvnot (bvlshr (bvnot s) t)))  */
```
martin-cs: A problem with using this directly is that it introduces two shifts and so doubles the size of the bit-blasting.  That is a significant performance overhead.  If we must do this rewrite, can we at least:
```
(let mask
      ((_ repeat m) ((_ extract |m-1| |m-1|) s))
      (bvxor mask (bvlshr (bvxor mask s) t)))
```

src/theory/bv/theory_bv_rewrite_rules.h
```diff
@@ -54,6 +54,7 @@ enum RewriteRuleId
   SgtEliminate,
   RedorEliminate,
   RedandEliminate,
+  AshrEliminate,
```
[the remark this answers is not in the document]
yoni206: Hi Martin, thanks for the comments.
This will not be used by default.
I am writing a pre-processing pass that translates bit-vector formulas to integer formulas.
In that pass, it is convenient to assume that some operators are already eliminated, including `bvashr`.
Therefore, I would like to eliminate this operator before I reach that preprocessing pass.
Now, this elimination procedure is very similar to other elimination procedures that occur in `src/theory/bv/theory_bv_rewrite_rules_operator_elimination.h`, and this is the reason I added it to that file.

Following an offline discussion with @aniemetz , I am now wondering whether this rewrite rule should instead be implemented inside my preprocessing pass, leaving the bit-vector rewriter intact. @mpreiner, @aniemetz, @martin-cs  --  what do you think?

src/theory/bv/theory_bv_rewrite_rules_operator_elimination.h
yoni206: Since the only purpose of this rewrite rules is to make the bv-to-int preprocessing pass easier, this optimization actually makes that translation harder. First, `bvxor` is harder to handle in integers than the alternatives. Second, it is also convenient to assume that `repeat` does not occur when translating to integers :)

src/theory/bv/theory_bv_rewrite_rules_operator_elimination.h
yoni206: This is an important point, beyond the inclusion or exclusion of the new rewriting rule in the bit-vector rewriter discussed above. The inline were added not because of the new rule, but because of a unit test that imports `src/theory/bv/theory_bv_rewrite_rules_operator_elimination.h`.
The same problem occurs in my preprocessing pass, since in it I also import the same file and am having the same issue.

My understanding is that this problems occurs just now and not before, because this is the first time `src/theory/bv/theory_bv_rewrite_rules_operator_elimination.h` is being included in more than one file. As the post I linked to earlier (and am linking [again](http://learningcppisfun.blogspot.com/2009/01/inline-specializations-and-multiple.html)), this creates two definitions of the same function, that generates errors of the form:
```
In function ... multiple definition of ... first defined here 
```

The post suggests two possible solutions: either add an `inline` or change the architecture of the rewrite rules, making all `applies` and `apply` functions not template-specializations but overloaded functions, using separate `.h` and `.cpp` files.

The main reason for me choosing the first option is that this seems to be what was chosen in all other cases in the bit-vector rewriter.
The rewriting rules are separated into the following files:
```
src/theory/bv/theory_bv_rewrite_rules_constant_evaluation.h 
src/theory/bv/theory_bv_rewrite_rules_simplification.h          
src/theory/bv/theory_bv_rewrite_rules_operator_elimination.h
src/theory/bv/theory_bv_rewrite_rules_normalization.h       
```

In all of them except for `src/theory/bv/theory_bv_rewrite_rules_operator_elimination.h`, all functions are inlined. In addition, all these functions are implemented as specializations of [this](https://github.com/CVC4/CVC4/blob/2a96d7de381aa565a6eacf724848c0e7839c7cf6/src/theory/bv/theory_bv_rewrite_rules.h#L352) class. Notice that in it, the functions are declared with `inline` (see, e.g., [here](https://github.com/CVC4/CVC4/blob/2a96d7de381aa565a6eacf724848c0e7839c7cf6/src/theory/bv/theory_bv_rewrite_rules.h#L411)), suggesting that the specialization should be inlined too.

Perhaps there is also a third option, but I am not aware of one.

src/theory/bv/theory_bv_rewrite_rules_operator_elimination.h
mpreiner: If options are need in `theory_bv_rewrite_rules_operator_elimination.h` then it should be included there.

test/unit/theory/theory_bv_rewriter_white.h
```diff
@@ -100,6 +102,29 @@ class TheoryBvRewriterWhite : public CxxTest::TestSuite
     TS_ASSERT_EQUALS(nr, Rewriter::rewrite(nr));
   }
 
+  void testRewriteAshrEliminate()
+  {
+    TypeNode bvType = d_nm->mkBitVectorType(4);
```
mpreiner: `bvtype` is unused.

test/unit/theory/theory_bv_rewriter_white.h
```diff
@@ -100,6 +102,29 @@ class TheoryBvRewriterWhite : public CxxTest::TestSuite
...
+    Node zero = d_nm->mkConst(BitVector(4, 0u));
```
mpreiner: `zero` is also unused.

test/unit/theory/theory_bv_rewriter_white.h
mpreiner: What would be actually better is to make a sat call and check the equivalence of the original and the rewritten node. But I'm not sure if this is overkill here. You can leave it for now.

src/theory/bv/theory_bv_rewrite_rules.h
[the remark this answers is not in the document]
mpreiner: If it's not on by default and currently only used in the new preprocessing pass, it should be ok. But what I actually don't like (and this is a problem with more elimination rules) is that with the ITE we introduce new Boolean terms. We should probably think of using `BITVECTOR_ITE` in this case so that everything is sent to the BV theory. @yoni206, you don't have to change that now, but this is something we will evaluate in the future for the new bit-vector solver.

src/theory/bv/theory_bv_rewrite_rules_operator_elimination.h
yoni206: Then this particular issue is resolved?

mpreiner: Ah sure... This is the header file, I was confused. Then it's resolved.

src/theory/bv/theory_bv_rewrite_rules_operator_elimination.h
mpreiner: I agree with @martin-cs, if it's better for your preprocessing pass to eliminate `bvashr` in this way, it's probably better to move the elimination into your preprocessing pass. In general, if we have to eliminate `bvashr` I would prefer @martin-cs's encoding.

yoni206: Sounds good. I will eliminate `bvashr` internally in my pass. For now, I have changed this PR to a small refactoring PR, that adds the inclusion of the options header file and adds `inline` to the functions.

src/theory/bv/theory_bv_rewrite_rules.h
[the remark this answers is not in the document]
martin-cs: I would be happy for it to live with the other bit-vector re-write but to be called from your pass.  As previously discussed I would be concerned if it was on by default.

src/theory/bv/theory_bv_rewrite_rules_operator_elimination.h
martin-cs: Having opened this can of worms I am now going to do the cowardly thing and try to shunt this on to someone else's desk.  I know @mpreiner and @aniemetz have Plans for the BV solver.  Perhaps they have already attacked the refactoring of this part?

mpreiner: No we haven't yet. But for me it's ok to apply these changes since it's consistent with the other rewrite rule header files.

src/theory/bv/theory_bv_rewrite_rules.h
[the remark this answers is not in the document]
yoni206: Thinking of it a bit more, @martin-cs, I realized that I can even optimize this further for integer: The condition in the ITE is whether or not the msb is 0. This is done with extracts. But when translating to integers, it is much better to check this by comparison to 2^{k-1}, k being the bitwidth. So looking at this integer-oriented version of the rewrite rule, I don't think it should be in the bv rewriter, because it will never be used by the bv rewriter.

yoni206:
@martin-cs, if you are ok with leaving this rewriting rule in the future preprocessing pass, could you please approve or mark the change-request as resolved so that i can merge? If not, we can continue the discussion.

martin-cs:
Sorry for the lag @yoni206 .

PR #3107, opened 2019-07-20
Split infer info data structure in strings

ajreynol:
Also refactors the mkExplain method and adds some const to some arguments.


src/theory/strings/inference_manager.h
```diff
@@ -125,16 +126,22 @@ class InferenceManager
...
                      bool asLemma = false);
   /** same as above, but where exp_n is empty */
-  void sendInference(std::vector<Node>& exp,
+  void sendInference(const std::vector<Node>& exp,
                      Node eq,
                      const char* c,
                      bool asLemma = false);
+  /** Send inference
+   *
+   * Calls the appropriate call to send inference based on the infer info
```
4tXJ7f: "Calls the appropriate call" reads a bit weird. Maybe just refer to the documentation of `sendInference`?

src/theory/strings/theory_strings.cpp
```diff
@@ -4110,51 +4088,57 @@ Node TheoryStrings::mkLength( Node t ) {
...
+    if (std::find(ap.begin(), ap.begin() + i, api) != ap.begin() + i)
+    {
+      // already processed
+      continue;
+    }
+    if (api.getKind() == AND)
+    {
+      for (const Node& apic : api)
+      {
+        ap.push_back(apic);
```
4tXJ7f: I am not a huge fan of modifying `ap` while iterating over it. The for loop above makes it seem like the size is fixed. Why not do two loops? Instead of copying `a` to `ap` in the beginning, iterate over `ap` and flatten the `AND`s into `ap`. This shouldn't be less efficient (in fact you save copying some useless elements into `ap`).

src/theory/strings/theory_strings.cpp
```diff
@@ -4110,51 +4088,57 @@ Node TheoryStrings::mkLength( Node t ) {
...
+    }
+    Debug("strings-explain") << "Add to explanation " << api << std::endl;
+    if (api.getKind() == NOT && api[0].getKind() == EQUAL)
+    {
+      Assert(hasTerm(api[0][0]));
+      Assert(hasTerm(api[0][1]));
+      // ensure that we are ready to explain the disequality
+      AlwaysAssert(d_equalityEngine.areDisequal(api[0][0], api[0][1], true));
     }
+    Assert(api.getKind() != EQUAL || d_equalityEngine.areEqual(api[0], api[1]));
```
4tXJ7f: One detail that is different here is that we are not skipping cases where `api[0] == api[1]`. Is this change intentional?

ajreynol: Yes, I realized that the core `explain` method is robust to this case, so this was being checked twice.

src/theory/strings/theory_strings.cpp
ajreynol: The flattening of AND may need to be done recursively, which makes this a bit of a pain.  I'll make a utility to do this.

src/theory/strings/theory_strings_utils.cpp
```diff
@@ -44,6 +44,21 @@ Node mkAnd(std::vector<Node>& a)
   return NodeManager::currentNM()->mkNode(AND, au);
 }
 
+void getConjuncts(Node n, std::vector<Node>& conj)
```
4tXJ7f: Sorry for being annoying but how about making this function take a `Kind` to flatten and making it non-recursive?

src/theory/strings/theory_strings.cpp
4tXJ7f: Yeah, but it is worth it IMHO :) The updated code is much easier to read.

src/theory/strings/theory_strings_utils.cpp
ajreynol: Done.  Notice that this is pretty performance critical code, so I've made a special case for when the top symbol is not the kind, in which we don't do a traversal.

src/theory/strings/theory_strings.cpp
```diff
@@ -4110,50 +4088,44 @@ Node TheoryStrings::mkLength( Node t ) {
...
-        for( unsigned j=ps; j<antec_exp.size(); j++ ) {
-          Debug("strings-explain") << "  " << antec_exp[j] << std::endl;
-        }
-        Debug("strings-explain") << std::endl;
-      }
+  // copy to processing vector
+  std::vector<Node> aconj;
+  for (const Node& ac : a)
+  {
+    utils::flattenOp(ac, aconj);
```
```suggestion
    utils::flattenOp(AND, ac, aconj);
```

src/theory/strings/theory_strings_utils.cpp
```diff
@@ -44,6 +44,45 @@ Node mkAnd(std::vector<Node>& a)
   return NodeManager::currentNM()->mkNode(AND, au);
 }
 
+void getConjuncts(Kind k, Node n, std::vector<Node>& conj)
```
4tXJ7f:
```suggestion
void flattenOp(Kind k, Node n, std::vector<Node>& conj)
```

PR #3110, opened 2019-07-22
Eager conflict detection in strings based on constant prefix/suffix

ajreynol:
This implements an eager conflict detection scheme in TheoryStrings based on finding when two terms in the same equivalence class have a conflicting prefix (or suffix).

This leads to significant improvements on Amazon benchmarks having many simple RE memberships.


src/theory/strings/theory_strings.cpp
```diff
@@ -1147,8 +1251,40 @@ void TheoryStrings::eqNotifyNewClass(TNode t){
...
+}
+
+void TheoryStrings::addPrefixToEqcInfo(Node t, Node concat, Node eqc)
+{
+  EqcInfo* ei = nullptr;
+  // check each side
+  for (unsigned r = 0; r < 2; r++)
+  {
+    unsigned index = r == 0 ? 0 : concat.getNumChildren() - 1;
+    Node c = utils::getConstantComponent(concat[index]);
```
4tXJ7f: Aren't the two lines above essentially what `getConstantPrefix()` does?

src/theory/strings/theory_strings.cpp
```diff
@@ -1147,8 +1251,40 @@ void TheoryStrings::eqNotifyNewClass(TNode t){
...
+  }
+  else if (k == STRING_CONCAT)
+  {
+    addPrefixToEqcInfo(t, t, t);
+  }
+}
+
+void TheoryStrings::addPrefixToEqcInfo(Node t, Node concat, Node eqc)
+{
+  EqcInfo* ei = nullptr;
```
4tXJ7f: Assert that `concat` is of kind `STRING_CONCAT`/`REGEXP_CONCAT`.

src/theory/strings/theory_strings.cpp
```diff
@@ -1147,8 +1251,40 @@ void TheoryStrings::eqNotifyNewClass(TNode t){
...
+    if (!c.isNull())
+    {
+      if (ei == nullptr)
+      {
+        ei = getOrMakeEqcInfo(eqc);
+      }
+      Trace("strings-eager-pconf-debug")
+          << "New term: " << concat << " for " << t << " with prefix " << c
+          << " (" << (r == 1) << ")" << std::endl;
+      setPendingConflict(ei->addPrefixConst(t, c, r == 1));
```
4tXJ7f: Hm, if we use `setPendingConflict()` in this way, maybe we should rename it to `setPendingConflictIf()`

src/theory/strings/theory_strings.cpp
ajreynol: Yes; this was done in a more verbose way for optimization.
In my estimation, the performance of this code is highly critical (it is run at standard effort), so this is e.g. ensuring we don't have to check the kind twice.

src/theory/strings/theory_strings.cpp
```diff
@@ -1319,6 +1488,14 @@ void TheoryStrings::assertPendingFact(Node atom, bool polarity, Node exp) {
   Trace("strings-pending-debug") << "  Finished collect terms" << std::endl;
 }
 
+void TheoryStrings::setPendingConflict(Node conf)
+{
+  if (!conf.isNull() && d_pendingConflict.get().isNull())
```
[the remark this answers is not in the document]
ajreynol: I don't think there is a reasonable heuristic that would make a difference here. Certainly if one is a subset of the other, we could do this, but this is too expensive to check (and actually will never occur in the current strategy).

src/theory/strings/theory_strings.cpp
ajreynol: Renaming to `setPendingConflictWhen`

src/theory/strings/theory_strings.cpp
4tXJ7f: I don't think that it matters too much but fair enough :)

src/theory/strings/theory_strings.cpp
```diff
@@ -1097,10 +1098,113 @@ TheoryStrings::EqcInfo::EqcInfo(context::Context* c)
...
+  {
+    Trace("strings-eager-pconf-debug") << "Check conflict " << prev << ", " << t
+                                       << " post=" << isSuf << std::endl;
+    Node prevC = utils::getConstantPrefix(prev, isSuf);
+    Assert(!prevC.isNull());
+    if (c.isNull())
+    {
+      c = utils::getConstantPrefix(t, isSuf);
+      Assert(!c.isNull());
+    }
```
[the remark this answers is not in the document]
4tXJ7f: Ah, I was actually thinking to assert it after the closing brace :) Maybe move it before merging?

src/theory/strings/theory_strings.cpp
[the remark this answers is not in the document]
4tXJ7f: Yeah, that's fair. I guess the best check would be to take the conflict with the lowest level.

src/theory/strings/theory_strings.cpp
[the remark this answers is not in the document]
4tXJ7f: The "do not merge" is just for this.

PR #3111, opened 2019-07-22
Properly parse qualified identifiers 

### LINKED ISSUE #2832: datatypes: type checking issue
I've a bit of an old cvc4, compiled Apr 15 2018. If the master fixes this issue, please do let me know!

For this benchmark:

```
(set-logic ALL)
(declare-datatypes ((SBVEither 2)) ((par (T1 T2)
                                    ((left_SBVEither  (get_left_SBVEither  T1))
                                     (right_SBVEither (get_right_SBVEither T2))))))

(declare-fun s0 () (_ BitVec 8))
(define-fun s1 () (SBVEither (_ BitVec 8) (_ BitVec 8)) (left_SBVEither s0))
(check-sat)
(get-model)
```

I get:

```
$ cvc4 c.smt2
(error "Declared type of defined constant does not match its definition
The constant   : s1
Declared type  : (SBVEither (_ BitVec 8) (_ BitVec 8)) 275
The definition : (left_SBVEither (as s0 T1))
Definition type: (SBVEither (_ BitVec 8) T2) 278")
```

which is just not true. (Also some bizarre numbers being printed in the message, but I guess that's just an artifact of the system.)

For what it's worth, z3 accepts this benchmark as is.

ajreynol:
This refactors the smt2 parser so that it properly parses qualified identifiers (the generic form of function application heads in smt2). This required refactoring of the termNonVariable grammar in Smt2.g.

Specifically:
- This PR merges the cases of termNonVariable of (1) a type-ascripted term with head `as`, (2) function variables applied to argument list (3) indexed functions applied to argument list, and (4) array constants, which can be seen as `(as const (Array T1 T2))` applied to an argument. 
These are now all cases of a qualified identifier (`qualIdentifier`) applied to argument list, which follows terms in section 3.6 of http://smtlib.cs.uiowa.edu/papers/smt-lib-reference-v2.6-r2017-07-18.pdf.
- Merges indexed function names and symbols into a `identifier`, following the definition of identifier in section 3.3 of http://smtlib.cs.uiowa.edu/papers/smt-lib-reference-v2.6-r2017-07-18.pdf.
- A few checks had to be rearranged to preserve the old behavior, including moving the hack for unary minus in sygus version 1.0.

This fixes #2832.


4tXJ7f:
I didn't have time to look at this in detail yet. One high-level comment that I have: We have decided to move logic out of the parser files as much as possible (to make it easier to move away from ANTLR3) but this PR adds quite a bit of logic in the parser file itself. Is there any chance that we could move this code to the parser class?

ajreynol:
I agree we should move code to the parser files.  This PR actually doesn't add much code, it just rearranges it. Perhaps we should do this on a followup PR?

mpreiner:
Yes, I think it's a good idea to move the logic to the parser files in a follow-up PR.

ajreynol:
Ready to merge?

mpreiner:
I think we should still wait to get a second review for that. @4tXJ7f do you have any time soon to do a quick review?

4tXJ7f:
I'll try to do it tomorrow evening (Thursday).

src/parser/smt2/Smt2.g
```diff
@@ -2395,6 +2432,211 @@ termNonVariable[CVC4::Expr& expr, CVC4::Expr& expr2]
...
+ * (3) An expression expr.
+ * (4) A type t.
+ *
+ * A qualified identifier is the generic case of function heads.
+ * With respect to the standard definition (Section 3.6 of SMT-LIB version 2.6)
+ * of qualified identifiers, we additionally parse:
+ * - "Array constant specifications" of the form (as const (Array T1 T2)),
+ * which notice are used as function heads e.g. ((as const (Array Int Int)) 0)
+ * specifies the constant array over integers consisting of constant 0. This
+ * is handled as it were a special case of an operator here.
```
aniemetz:
```suggestion
 * is handled as if it were a special case of an operator here.
```

src/parser/smt2/Smt2.g
```diff
@@ -2395,6 +2432,211 @@ termNonVariable[CVC4::Expr& expr, CVC4::Expr& expr2]
...
+ *
+ * Examples:
+ *
+ * (Identifiers)
+ *
+ * - For declared functions f, we return (2).
+ * - Indexed functions like testers (_ is C) and bitvector extract
```
aniemetz:
```suggestion
 * - For indexed functions like testers (_ is C) and bitvector extract
```

PR #3112, opened 2019-07-23
Model blocker feature

ajreynol:
This adds the commands `block-model` and `block-model-values` to the Smt2 interface and the public methods of SmtEngine. 

*** (block-model-values [term list])

Valid in SAT mode. Command (block-model (t1 ... tn)) is syntax sugar for:

(assert (or (not (= t1 (M t1))) ... (not (= tn (M tn)))))

where (M t1) ... (M tn) are the current model values for t1 ... tn from the last
check-sat query. Like assert, moves solver to Assert mode.

*** (block-model)

Valid in SAT mode, moves solver to Assert mode. Instructs the solver to block
the last model based on solver-specific options. For instance, CVC4 can
choose to block the model based on model values, based on literal assignments,
etc., which would be based on whatever options we define.  This is determined by
the option `--block-models`.


src/smt/model_blocker.h
```diff
@@ -0,0 +1,65 @@
...
+   *    x > 0 ^ ( y < 0 V z < 0 V w < 0 )
+   * and m is { x -> 1, y -> 2, z -> -1, w -> -1 }, then this method may
+   * return ~(z < 0) or ~(w < 0) when set to mode "literals".
+   *
+   * Notice that we do not require that L1...Ln entail unit top-level assertions
+   * since these literals are trivially entailed to be true in all models of
+   * our input. In other words, we do not return ~(x < 0) V ~(w < 0) since the
+   * left disjunct is always false.
+   */
+  static Expr getModelBlocker(
```
yoni206: Can you add a description of the parameters? In particular that `exprToBlock` is optional because it is only used when `mode` is `BLOCK_MODELS_VALUES `?

src/smt/model_blocker.cpp
```diff
@@ -0,0 +1,292 @@
...
+    } while (!visit.empty());
+    Assert(visited.find(formula) != visited.end());
+    Assert(!visited.find(formula)->second.isNull());
+    blocker = visited[formula].negate();
+  }
+  else
+  {
+    Assert(mode == BLOCK_MODELS_VALUES);
+    std::vector<Node> blockers;
+    // if specific terms were not specified in get-value, block all variables of
```
yoni206:
```suggestion
    // if specific terms were not specified, block all variables of
```

src/smt/model_blocker.cpp
```diff
@@ -0,0 +1,292 @@
...
+      {
+        if (s.getType().getKind() != kind::FUNCTION_TYPE)
+        {
+          Node v = m->getValue(s);
+          Node a = nm->mkNode(DISTINCT, s, v);
+          blockers.push_back(a);
+        }
+      }
+    }
+    // otherwise, block all terms that were specified in get-value
```
yoni206:
```suggestion
    // otherwise, block all terms that were specified
```

src/smt/model_blocker.cpp
```diff
@@ -0,0 +1,292 @@
...
+  }
+  else
+  {
+    Assert(mode == BLOCK_MODELS_VALUES);
+    std::vector<Node> blockers;
+    // if specific terms were not specified in get-value, block all variables of
+    // the model
+    if (nodesToBlock.empty())
+    {
+      Trace("model-blocker") << "no get-value recognized" << std::endl;
```
yoni206:
```suggestion
      Trace("model-blocker") << "no specific terms to block recognized" << std::endl;
```

src/smt/model_blocker.cpp
```diff
@@ -0,0 +1,292 @@
...
+  for (unsigned i = 0, size = exprToBlock.size(); i < size; i++)
+  {
+    nodesToBlock.push_back(Node::fromExpr(exprToBlock[i]));
+  }
+  Trace("model-blocker") << "Compute model blocker, assertions:" << std::endl;
+  Node blocker;
+  if (mode == BLOCK_MODELS_LITERALS)
+  {
+    Assert(nodesToBlock.empty());
+    // optimization: filter to only top-level disjunctions
```
yoni206: What about implications? Why aren't they needed?

src/options/options_handler.cpp
```diff
@@ -1754,6 +1754,49 @@ ModelCoresMode OptionsHandler::stringToModelCoresMode(std::string option,
   }
 }
 
+const std::string OptionsHandler::s_blockModelsHelp =
+    "\
+Blocking models modes are currently supported by the --simplification option:\n\
```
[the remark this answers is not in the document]
ajreynol: :)

src/smt/model_blocker.h
```diff
@@ -0,0 +1,65 @@
...
+class ModelBlocker
+{
+ public:
+  /** get model blocker
+   *
+   * This returns a disjunction of literals ~L1 V ... V ~Ln with the following
+   * properties:
+   * (1) L1 ... Ln hold in the current model (given by argument m),
+   * (2) if mode is set to "literals", L1 ... Ln are literals that occur in
+   * assertions and propositionally entail all non-unit top-level assertions.
```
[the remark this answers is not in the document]
ajreynol: This is an optimization: the unit assertions are not interesting to block since they always must hold.  Hence we don't care about them when model blocking

src/smt/model_blocker.h
ajreynol: I fixed the documentation (I had renamed nodesToBlock -> exprToBlock).  I've added a bit more documentation.

src/smt/model_blocker.cpp
ajreynol: Implications are disjunctions, I'll clarify the comment.

src/smt/smt_engine.h
```diff
@@ -374,6 +374,16 @@ class CVC4_PUBLIC SmtEngine {
...
+  /** ensure that the model is available and return a pointer to it
+   *
+   * This ensures that the model is currently available, which means that
+   * CVC4 is producing models, and is in "SAT mode", otherwise an exception
+   * is thrown.
+   *
+   * The flag c is used for giving an error message to indicate the context
+   * this method was called.
+   */
+  theory::TheoryModel* ensureAvailableModel(const char* c) const;
```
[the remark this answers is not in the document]
ajreynol: Changed.

src/smt/model_blocker.h
```diff
@@ -0,0 +1,69 @@
...
+   * (3) if mode is set to "values", L1 ... Ln are literals of the form x=c,
+   * where c is the value of x in the current model.
+   * (4) if exprToBlock is not empty, L1 ... Ln are literals of the form t=c,
+   * where t is an element of exprToBlock and c is its value in the current
+   * model.
+   *
+   * We expect exprToBlock to be non-empty if and only if mode is
```
yoni206: if and only if? 
Looks like we only rely on "only if" but not on "if" (in the cpp file):

```
 Assert(mode == BLOCK_MODELS_VALUES);
    std::vector<Node> blockers;
    // if specific terms were not specified, block all variables of
    // the model
    if (nodesToBlock.empty())
    {
      Trace("model-blocker") << "no specific terms to block recognized" << std::endl;
      std::unordered_set<Node, NodeHashFunction> symbols;
      for (Node n : tlAsserts)
      {
        expr::getSymbols(n, symbols);
      }
      for (Node s : symbols)
      {
        if (s.getType().getKind() != kind::FUNCTION_TYPE)
        {
          Node v = m->getValue(s);
          Node a = nm->mkNode(DISTINCT, s, v);
          blockers.push_back(a);
        }
      }
    }
```

ajreynol: The "only if" is here: https://github.com/CVC4/CVC4/pull/3112/files#diff-391e93405e2e7168c5717b5b3ade56cdR48

yoni206: right, but where is the "if"? :)

ajreynol: Actually it appears the code may need to change here.  The question is what does (block-model-values ()) mean?

yoni206: Currently, if I understand correctly, it is syntactic sugar for `(model-values (x1 ... xn))` with `x1...xn` being the variables in the formula. This was the original implementation of the `values` option.
However, at least for Alessandro's team, this option was much less useful than the potential option of actually specifying the terms they wanted to block. So maybe we shouldn't allow it. What do you think?

ajreynol: Yes, let's disallow it. If it were allowed, it would be syntax sugar for `(assert false)` actually.

src/smt/smt_engine.cpp
```diff
@@ -4188,6 +4218,16 @@ Expr SmtEngine::getValue(const Expr& ex) const
   return resultNode.toExpr();
 }
 
+vector<Expr> SmtEngine::getValues(const vector<Expr>& exprs)
+{
+  vector<Expr> result;
+  for (unsigned i = 0, size = exprs.size(); i < size; i++)
```
mpreiner: for each?

src/smt/smt_engine.cpp
```diff
@@ -4394,6 +4458,21 @@ std::pair<Expr, Expr> SmtEngine::getSepHeapAndNilExpr(void)
       "expressions from theory model.");
 }
 
+std::vector<Expr> SmtEngine::getExpandedAssertions()
+{
+  std::vector<Expr> easserts = getAssertions();
+  // must expand definitions
+  std::vector<Expr> eassertsProc;
+  std::unordered_map<Node, Node, NodeHashFunction> cache;
+  for (unsigned i = 0, nasserts = easserts.size(); i < nasserts; i++)
```
mpreiner: for each?

src/smt/model_blocker.cpp
```diff
@@ -0,0 +1,294 @@
...
+
+Expr ModelBlocker::getModelBlocker(const std::vector<Expr>& assertions,
+                                   theory::TheoryModel* m,
+                                   BlockModelsMode mode,
+                                   const std::vector<Expr>& exprToBlock)
+{
+  NodeManager* nm = NodeManager::currentNM();
+  // convert to nodes
+  std::vector<Node> tlAsserts;
+  for (unsigned i = 0, size = assertions.size(); i < size; i++)
```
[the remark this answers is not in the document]
ajreynol: I wasn't sure it was common practice to do for-each loops for vectors of Expr.  My performance testing was on vectors of Nodes only.  (Actually, do our unit tests ensure that this works?).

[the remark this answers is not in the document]
mpreiner: We should just get rid of the Expr layer... But in this case if you use `for (const Expr& ...)` there shouldn't be any issue at all.

ajreynol:
@mpreiner , you're right, for-each Expr is doable. I've addressed the comments now.

src/smt/model_blocker.cpp
```diff
@@ -0,0 +1,288 @@
...
+    } while (!visit.empty());
+    Assert(visited.find(formula) != visited.end());
+    Assert(!visited.find(formula)->second.isNull());
+    blocker = visited[formula].negate();
+  }
+  else
+  {
+    Assert(mode == BLOCK_MODELS_VALUES);
+    std::vector<Node> blockers;
+    // if specific terms were not specified, block all variables of
```
yoni206: Didn't you say that now we do not allow this case?

src/smt/model_blocker.cpp
```diff
@@ -0,0 +1,288 @@
...
+    // the model
+    if (nodesToBlock.empty())
```
yoni206: Shouldn't this become an assertion?

ajreynol: See above.

src/smt/model_blocker.cpp
ajreynol: No, we do not allow an empty set of values from the smt2 command: https://github.com/CVC4/CVC4/pull/3112/commits/8e6add556b0dae71218e79af9ecc486153677a00#diff-c78de56e9d21a5aa6d0deae68fa5c202R4421

We still allow it in the utility, as documented.

yoni206: But in the h file we have:
`We expect exprToBlock to be non-empty only if mode is BLOCK_MODELS_VALUES.`

ajreynol: There is no `if`. :)

PR #3116, opened 2019-07-23
Minor refactoring of regexp operation

ajreynol:


src/theory/strings/regexp_operation.cpp
```diff
@@ -1432,70 +1432,60 @@ Node RegExpOpr::intersectInternal( Node r1, Node r2, std::map< PairNodes, Node >
...
+      std::vector<Node> vec_nodes;
+      for (const Node& rc : r)
+      {
+        Node tmpNode = removeIntersection(rc);
+        vec_nodes.push_back(tmpNode);
       }
-      default: {
-        Unreachable();
+      retNode =
+          Rewriter::rewrite(NodeManager::currentNM()->mkNode(rk, vec_nodes));
```
4tXJ7f: Since we are already refactoring: maybe switch to our `nm`-for-`NodeManager` pattern and use a `NodeBuilder` instead of `vec_nodes`?

src/theory/strings/regexp_operation.cpp
```diff
@@ -1432,70 +1432,60 @@ Node RegExpOpr::intersectInternal( Node r1, Node r2, std::map< PairNodes, Node >
...
+      break;
+    }
+
+    case REGEXP_INTER:
+    {
+      retNode = removeIntersection(r[0]);
+      for (unsigned i = 1, nchild = r.getNumChildren(); i < nchild; i++)
```
```suggestion
      for (size_t i = 1, nchild = r.getNumChildren(); i < nchild; i++)
```

PR #3117, opened 2019-07-24
Use python realpath instead of relying on shell realpath

makaimann:
Mac OS does not have `realpath`, thus the `./contrib/get-antlr-3.4` script fails. This uses an alternative implementation to avoid this issue.

First reported here: https://github.com/makaimann/smt-switch/issues/3


makaimann:
I haven't had a chance to test this on Mac yet, I'll try it later today and update the label if it works.

makaimann:
Tested on macOS Mojave and it works.

contrib/get-antlr-3.4
```diff
@@ -3,6 +3,21 @@
 source "$(dirname "$0")/get-script-header.sh"
 ANTLR_HOME_DIR=antlr-3.4
 
+# re-implementation of realpath, because Mac OS doesn't have realpath
+# https://stackoverflow.com/questions/3572030/bash-script-absolute-path-with-os-x/18443300#18443300
+portable_realpath() {
+    OURPWD=$PWD
```
4tXJ7f: The comments on the Stack Overflow post recommend using `local`, lowercase variables, which are good suggestions. Could we apply them? Also, maybe we could move this script to a separate file in the contrib directory in case some other script needs a portable copy of `realpath`.

makaimann: Thanks for the helpful feedback! Just made the requested changes. Let me know if the style seems okay.

contrib/portable_realpath_fun.sh
```diff
@@ -0,0 +1,16 @@
...
+    cd "$(dirname "$1")"
+    local link=$(readlink "$(basename "$1")")
+    while [ "$link" ]; do
+        cd "$(dirname "$link")"
+        link=$(readlink "$(basename "$1")")
+    done
+    local ourrealpath="$PWD/$(basename "$1")"
+    cd "$ourpwd"
+    echo "$ourrealpath"
+}
```
mpreiner: Not sure if this script is actually doing what it should. For example, for path `/foo/bar/cvc4/contrib/../..` above script produces `/foo/bar/cvc4/..`, whereas realpath produces `/foo/bar`. So why not just use `$(pwd)/../..` as prefix and don't use realpath at all? If we want to resolve symbolic links, I would actually vote for the Python approach. Sure we have to call the Python interpreter, but it is less error prone than bash hacks.

makaimann: Thanks for checking that -- should have unit-tested it. Based on comments it looks like we should just use an absolute path to avoid warnings so I went the python route.

contrib/get-antlr-3.4
```diff
@@ -46,7 +46,7 @@ cd libantlr3c-3.4
 
 # Use an absolute path for the installation directory to avoid spurious libtool
 # warnings about the ANTLR library having moved
-PREFIX=$(realpath `pwd`/../..)
+PREFIX=$(python -c "import os,sys; print os.path.realpath(sys.argv[1])" `pwd`/../..)
```
4tXJ7f:
```suggestion
PREFIX=$(python -c "import os,sys; print os.path.realpath(sys.argv[1])" $(pwd)/../..)
```
Instead of mixing the two types of command substitution :)

mpreiner:
```suggestion
PREFIX=$(python -c "import os; print(os.path.realpath('$(pwd)/../..'))")
```

If we want to get even more fancy ;-)

makaimann: Good point! Updated.

makaimann: I missed the second suggested change, but I just applied it now.

PR #3119, opened 2019-07-25
Input user grammar in sygus abduct

ajreynol:
This is work towards incorporating user grammars for sygus abduct.

This makes a static public utility function for constructing sygus abduction conjectures that takes as input an optional sygus datatype argument, indicating the grammar.

### DIFF
--- a/src/preprocessing/passes/sygus_abduct.cpp
+++ b/src/preprocessing/passes/sygus_abduct.cpp
@@ -136,7 +136,7 @@ Node SygusAbduct::mkAbductionConjecture(const std::vector<Node>& asserts,
 
     // datatype types we need to process
     std::vector<TypeNode> dtToProcess;
-// datatype types we have processed
+    // datatype types we have processed
     std::map<TypeNode, TypeNode> dtProcessed;
     dtToProcess.push_back(abdGType);
     std::stringstream ssutn0;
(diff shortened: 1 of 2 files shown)


src/preprocessing/passes/sygus_abduct.h
```diff
@@ -56,6 +56,18 @@ class SygusAbduct : public PreprocessingPass
  public:
   SygusAbduct(PreprocessingPassContext* preprocContext);
 
+  /**
+   * Returns the sygus conjecture corresponding to the abduction problem for
+   * input problem given by asserts, and axioms given by axioms. Note that
+   * axioms is expected to be a subset of asserts.
```
yoni206: Could you clarify what `axioms` and `asserts` are? Maybe in terms of traditional A and B such that A does not imply B? In particular, why should `axioms` be a subset of `asserts`?

src/preprocessing/passes/sygus_abduct.cpp
```diff
@@ -52,15 +52,43 @@ PreprocessingPassResult SygusAbduct::applyInternal(
...
+    for (size_t i = 0, astart = assertionsToPreprocess->getAssumptionsStart();
+         i < astart;
+         i++)
     {
+      // if we are not an assumption, add it to the set of axioms
       axioms.push_back(asserts[i]);
     }
   }
+
+  // the abduction type we are using (null for now)
```
yoni206:
```suggestion
  // the abduction grammar type we are using (null for now, until a future commit)
```

src/preprocessing/passes/sygus_abduct.cpp
```diff
@@ -93,6 +123,152 @@ PreprocessingPassResult SygusAbduct::applyInternal(
...
+  {
+    Assert(abdGType.isDatatype() && abdGType.getDatatype().isSygus());
+    // must convert all constructors to version with bound variables in "vars"
+    std::vector<Datatype> datatypes;
+    std::set<Type> unres;
+
+    Trace("sygus-abduct-debug") << "Process abduction type:" << std::endl;
+    Trace("sygus-abduct-debug") << abdGType.getDatatype() << std::endl;
+
+    // datatype types we have processed
```
yoni206:
```suggestion
    // datatype types we need to process
```

src/preprocessing/passes/sygus_abduct.cpp
```diff
@@ -93,6 +123,152 @@ PreprocessingPassResult SygusAbduct::applyInternal(
...
+    std::vector<TypeNode> dtToProcess;
+    std::map<TypeNode, TypeNode> dtProcessed;
```
yoni206:
```suggestion
// datatype types we have processed
    std::map<TypeNode, TypeNode> dtProcessed;
```

src/preprocessing/passes/sygus_abduct.cpp
```diff
@@ -93,6 +124,170 @@ PreprocessingPassResult SygusAbduct::applyInternal(
...
+    // variables of the input problem, and varlist is the formal argument list
+    // of the abduct-to-synthesize. For example, given user-provided sygus
+    // grammar:
+    //   G -> a | +( b, G )
+    // we synthesize a abduct A with two arguments x_a and x_b corresponding to
+    // a and b, and reconstruct the grammar:
+    //   G' -> x_a | +( x_b, G' )
+    // In this way, x_a and x_b are treated as bound variables and handled as
+    // arguments of the abduct-to-synthesize instead of as free variables with
+    // no relation to A. We additionally require that x_a, when printed, prints
```
yoni206:
```suggestion
    // no relation to a. We additionally require that x_a, when printed, prints
```

ajreynol: No, I mean A (for the abduct-to-synthesize) here.  :)

PR #3120, opened 2019-07-25
 Handle RE intersections modulo equality

ajreynol:
This PR ensures RE solver considers equality over strings when detecting intersection conflicts. Previously, `R1` and `R2` would not be intersected when `x in R1 ^ x = y ^ y in R2` was asserted.

The code also changes so that intersection "inferences" are made, e.g.:
`x in R1 ^ y in R2 ^ x = y => x  in intersect( R1, R2 )`
Notice this is a reducing inference, so `x in R1` and `y in R2` can be marked reduced.
Previously, the RE solver was caching the intermediate value of `intersect( R1, R2 )` internally, in a cache `d_inter_cache`.  Using an inference makes this much cleaner, since we do not rely on the internal state of the RE solver and proceed as usual on the next call to `RegexpSolver::check(...)`.

This PR simplifies the code for managing RE memberships, as it was maintaining redundant SAT-context dependent information that was redundant (and reconstructed in TheoryStrings::checkMemberships).


src/theory/strings/theory_strings.cpp
```diff
@@ -1056,24 +1056,28 @@ void TheoryStrings::checkMemberships()
 {
   // add the memberships
   std::vector<Node> mems = getExtTheory()->getActive(kind::STRING_IN_REGEXP);
+  // maps representatives regular expression memberships for that class
```
```suggestion
  // maps representatives to regular expression memberships in that class
```
(the original sentence doesn't quite parse for me)

src/theory/strings/regexp_solver.h
```diff
@@ -49,23 +49,33 @@ class RegExpSolver
...
    *
-   * Tells this solver to check whether the regular expressions asserted to it
+   * Tells this solver to check whether the regular expressions in mems
    * are consistent. If they are not, then this class will call the
    * sendInference method of its parent TheoryString object, indicating that
    * it requires a conflict or lemma to be processed.
+   *
+   * The argument mems maps representative string terms r to memberships of the
+   * form (t in R) or ~(t in R), where t = r currently holds in the equality
+   * engine of theory of strings.
```
4tXJ7f:
```suggestion
   * engine of the theory of strings.
```

src/theory/strings/regexp_solver.h
```diff
@@ -49,23 +49,33 @@ class RegExpSolver
...
+  /**
+   * Check memberships for equivalence class.
+   * The vector mems is a vector of memberships of the form:
+   *   (~) (x1 in R1 ) ... (~) (xn in Rn)
+   * where x1 = ... = xn in the current context.
+   *
+   * This method may add lemmas or conflicts via the inference manager.
+   *
+   * This method returns false if it discovered a conflict for this set of
+   * assertions, and false otherwise. It discovers a conflict e.g. if mems
```
4tXJ7f:
```suggestion
   * assertions, and true otherwise. It discovers a conflict e.g. if mems
```

src/theory/strings/regexp_operation.h
```diff
@@ -86,14 +87,23 @@ class RegExpOpr {
...
   int derivativeS( Node r, CVC4::String c, Node &retNode );
   Node derivativeSingle( Node r, CVC4::String c );
+  /**
+   * Returns the regular expression intersection of r1 and r2. If r1 or r2 is
+   * not constant, then this method returns null and sets spflag to true.
+   */
   Node intersect(Node r1, Node r2, bool &spflag);
-
-  std::string mkString( Node r );
+  /** Get the pretty printed version of r */
```
4tXJ7f:
```suggestion
  /** Get the pretty printed version of the regular expression r */
```

src/theory/strings/regexp_solver.cpp
```diff
@@ -174,14 +79,28 @@ void RegExpSolver::check()
...
+    {
+      for (const Node& m : mr.second)
+      {
+        bool polarity = m.getKind() != NOT;
+        if (polarity || !options::stringIgnNegMembership())
+        {
+          allMems.push_back(m);
+        }
+      }
+    }
```
4tXJ7f: Just a thought: I am not quite sure if it is better to compute this once and then use it twice or to just compute it twice below (saving a bunch of copies and allocations).

src/theory/strings/regexp_solver.cpp
```diff
@@ -327,6 +246,90 @@ void RegExpSolver::check()
...
+      {
+        vec_nodes.push_back(mi[0].eqNode(m[0]));
+      }
+      Node conc;
+      d_im.sendInference(vec_nodes, conc, "INTERSECT CONFLICT", true);
+      // conflict, return
+      return false;
+    }
+    // rewrite to ensure the equality checks below are precise
+    Node mres = Rewriter::rewrite(nm->mkNode( STRING_IN_REGEXP, mi[0],resR));
```
4tXJ7f: Apply ClangFormat before merging.

src/theory/strings/regexp_solver.cpp
ajreynol: I think I may refactor this on a future PR; I don't think this is a big deal performance wise.

PR #3121, opened 2019-07-26
 New C++ API: Introduce macros for try-catch blocks in Solver.

aniemetz:
We use a repeated pattern to guard the API functions of the solver objects. This PR introduces macros for this pattern.


src/smt/smt_engine.h
```diff
@@ -472,6 +472,11 @@ class CVC4_PUBLIC SmtEngine {
    */
   bool isFullyInited() { return d_fullyInited; }
 
+  /**
+   * Return the number of user context levels.
```
4tXJ7f:
```suggestion
   * Return the user context level.
```

src/api/cvc4cpp.cpp
```diff
@@ -3332,12 +3331,17 @@ std::vector<Term> Solver::getValue(const std::vector<Term>& terms) const
  */
 void Solver::pop(uint32_t nscopes) const
 {
-  // CHECK: incremental enabled?
-  // CHECK: nscopes <= d_smtEngine->d_userLevels.size()
+  CVC4_API_CHECK(d_smtEngine->getOption("incremental").toString() == "true")
+      << "Cannot push when not solving incrementally (use --incremental)";
```
4tXJ7f:
```suggestion
      << "Cannot pop when not solving incrementally (use --incremental)";
```

src/api/cvc4cpp.cpp
```diff
@@ -27,6 +27,7 @@
 #include "expr/type.h"
 #include "options/main_options.h"
 #include "options/options.h"
+#include "options/smt_options.h"
```
4tXJ7f: I may be mistaken but could it be that this include is not actually needed because you are using `SmtEngine::getOption()` to check whether incremental mode is enabled or not?

aniemetz: indeed

aniemetz:
> Just some minor comments below. One question that I would like to raise is whether we should put the `CVC4_API_SOLVER_TRY_CATCH_BEGIN` always in the beginning of each function because the checks themselves may potentially throw exceptions. Thoughts?

All checks that are currently outside shouldn't throw, but it probably doesn't hurt and is less error prone. Will do.

aniemetz:
@4tXJ7f all addressed PTAL

PR #3122, opened 2019-07-26
Support get-abduct smt2 command

ajreynol:
This adds support for the smt2 command get-abduct.

This PR removes the old preprocessing pass `--sygus-abduct`. This option was used to rewrite the entire input to a sygus conjecture corresponding to the abduction problem.  

Now, we use a separate copy of the SMT engine to solve abduction problems whenever `--produce-abducts` is enabled and a `(get-abduct ...)` command is issued while leaving the assertion stack unchanged.


test/regress/regress1/sygus-abduct-test.smt2
```diff
@@ -1,4 +1,4 @@
-; COMMAND-LINE: --sygus-abduct --sygus-abort-size=2 
+; COMMAND-LINE: --produce-abducts --sygus-stream --sygus-abort-size=2
```
yoni206: I am wondering, from a usability standpoint, whether it makes sense to "mix" abduction options with SyGuS options. 
Isn't it a bit weird that the user asks for an abduct with `(get-abduct ...)` but then to get a stream of abducts, a SyGuS options needs to be enabled? 
SyGuS here is the technique, and the user asking for an abduct shouldn't really care about it.

Maybe there is room for an additional command `(get-abducts ...)` (that will automatically turn on the required sygus flag, if that's a possibility)? 
Or `(get-abducts ... N)` where  N is `sygus-abort-size`?

test/regress/regress1/sygus-abduct-test-user.smt2
```diff
@@ -0,0 +1,28 @@
+; COMMAND-LINE: --produce-abducts
```
yoni206: Is it possible, with this new architecture, to include also `--check-sygus-sol`?

ajreynol: Yes, this should work as expected.

test/regress/regress1/sygus-abduct-test.smt2
ajreynol: We haven't standardized the "streaming" version of get-abduct.  In fact, it probably does not make sense from an API point of view to tell CVC4 to do an infinite loop streaming solutions.

What you said would require a further discussion to the smt2 interface, but I don't think it's necessary to make this public facing.

PR #3124, opened 2019-07-28
Minor improvement for rewriter for str.replace

ajreynol:
We were unecessarily rewriting, e.g. `(str.replace "A" "A" "B") ---> (str.++ "" "B" "")` instead of `"B"`.


ajreynol:
I think these branches are already being taken hundreds if not thousands of times :)  But I've added a regression for good measure.

PR #3125, opened 2019-07-28
Fix match trie for polymorphic operators

ajreynol:
This fixes a bug in the MatchTrie where type constraints were not considered.  This means that e.g. `(= x y)` with `x:Int y:Int` could be matched to `(= t s)` with `t:String s:String`.

This utility is currently used for filtering rewrite rules with `--sygus-rr` where such spurious matches likely did not affect much. (However, in the context of the proof infrastructure https://github.com/ajreynol/CVC4/tree/proofDb, this bug needs to be fixed).

### DIFF
--- a/src/theory/quantifiers/candidate_rewrite_filter.cpp
+++ b/src/theory/quantifiers/candidate_rewrite_filter.cpp
@@ -135,6 +135,10 @@ bool MatchTrie::getMatches(Node n, NotifyMatch* ntm)
               recurse = false;
             }
           }
+          else if (!var.getType().isSubtypeOf(cn.getType()))
+          {
+            recurse = false;
+          }
           else
           {
             vars.push_back(var);


src/theory/quantifiers/candidate_rewrite_filter.cpp
```diff
@@ -135,6 +135,10 @@ bool MatchTrie::getMatches(Node n, NotifyMatch* ntm)
               recurse = false;
             }
           }
+          else if (!var.getType().isSubtypeOf(cn.getType()))
```
HanielB: I imagine you would not want to merge this `if` condition with the one above it? As is it can be a little annoying to see when `recurse` is actually gonna be false. Also, you could avoid the lookup in the map if you tested the type first, no?

ajreynol: Computing types is actually (much) more expensive than a map lookup.  The code is organized such that this is minimized.

HanielB: I was afraid that was the case. :)

PR #3131, opened 2019-07-30
Parsing THF and adding several regressions

HanielB:
Modifies the TPTP grammar to allow parsing THF, its higher-order extension. Also adds several regressions that were instrumental for exercising features of the parser and of higher-order solving in CVC4. When possible the actual exercised feature was marked in the regression name.


src/options/options.h
```diff
@@ -241,7 +241,7 @@ class CVC4_PUBLIC Options {
   unsigned getParseStep() const;
   unsigned getThreadStackSize() const;
   unsigned getThreads() const;
-
+  bool getHoFlattenTotal() const;
```
ajreynol: I would prefer if this was not a parsing option. Can you instead implement this as a rewrite internally after parsing?

HanielB: I'll purge it from this PR and add in another one then. This option is actually not really impactful, although sometimes it helps

src/parser/tptp/Tptp.g
```diff
@@ -637,8 +759,320 @@ fofBinaryNonAssoc[CVC4::parser::tptp::NonAssoc& na]
...
+      //    (define-fun Q ((x Int)) (-> Int Int) (lambda y (P x)))
+      // is equivalent to the command:
+      //    (define-fun Q ((x Int) (z Int)) Int (@ (lambda y (P x)) z))
+      // where @ is (higher-order) application. In this example, z is added to
+      // flattenVars.
+
+      // flatten body type
+      Type range = expr.getType();
+      std::vector<Expr> flattenVars;
+      if (range.isFunction())
```
ajreynol: There is a utility that does this block of code (almost exactly I think): https://github.com/CVC4/CVC4/blob/master/src/parser/parser.h#L592.
Would it be easy just to call this function?

HanielB: That was actually my inspiration. IIRC though, that is used only for types, while here it's about creating the term. While the loop part is the same between the two pieces of code, what they actually produce is different.

HanielB: I changed the code to use the existing function. There will be the spurious computation of some type information, but that is negligible.

src/parser/tptp/Tptp.g
```diff
@@ -637,8 +759,303 @@ fofBinaryNonAssoc[CVC4::parser::tptp::NonAssoc& na]
...
+          for (unsigned i = 1; i < args.size(); ++i)
+          {
+            expr = MK_EXPR(kind::HO_APPLY, expr, args[i]);
+          }
+        }
+      }
+    )?
+  ;
+
+// atom
```
ajreynol: What are these comments?

src/parser/tptp/Tptp.g
```diff
@@ -107,6 +107,7 @@ using namespace CVC4::parser;
 #include "parser/antlr_input.h"
 #include "parser/parser.h"
 #include "parser/tptp/tptp.h"
+#include "options/options.h"
```
ajreynol: Do we still need this?

ajreynol:
Looks good to me, although there is a failing regression now?

ajreynol:
Failing regression is `regress0/ho/nested_lambdas-AGT034^2.smt2`, looks to be a timeout.

HanielB:
Issue was with proof checking. Fixed now (by disabling it for that regression).

HanielB:
I updated the regressions to respect these criteria (I was checking the times only on production). I also moved a bunch of the old ho regressions (in fact, all but one) from regress0 to regress1 because they failed the criteria.

PR #3132, opened 2019-07-30
 Track solver execution mode

ajreynol:
This refactors the SMT engine so that solver execution mode (Figure 4.1 of http://smtlib.cs.uiowa.edu/papers/smt-lib-reference-v2.6-r2017-07-18.pdf) is explicitly tracked.

This is required for further work on `get-abduct` / `get-next-abduct`.

### DIFF
--- a/src/smt/smt_engine.cpp
+++ b/src/smt/smt_engine.cpp
@@ -5054,9 +5054,7 @@ Expr SmtEngine::doQuantifierElimination(const Expr& e, bool doFull, bool strict)
   }
 }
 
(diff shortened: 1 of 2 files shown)


src/smt/smt_engine.cpp
```diff
@@ -5223,7 +5241,7 @@ void SmtEngine::push()
   // The problem isn't really "extended" yet, but this disallows
   // get-model after a push, simplifying our lives somewhat and
   // staying symmtric with pop.
```
mpreiner:
```suggestion
  // staying symmetric with pop.
```

PR #3133, opened 2019-07-30
Add some missing cases in evaluator

ajreynol:


src/theory/evaluator.cpp
```diff
@@ -259,6 +259,12 @@ EvalResult Evaluator::evalInternal(TNode n,
           break;
         }
 
+        case kind::UMINUS:
+        {
+          const Rational& x = results[currNode[0]].d_rat;
+          results[currNode] = EvalResult(-x);
+        }
+        break;
```
4tXJ7f: Maybe move this `break` inside of the block above to be consistent with the other code.

PR #3134, opened 2019-07-30
 Regular expression intersection modes

ajreynol:
This PR introduces a mode for when to compute regular expression intersections. Since regular expression intersection is used as a way of increasing refutational completeness, we can vary our effort as a heuristic while maintaining (model) soundness.

This PR changes the default behavior so that regular expressions with `re.allchar` are *not* taken into consideration when computing intersections.


ajreynol:
We should merge https://github.com/CVC4/CVC4/pull/3110 before this.

src/theory/strings/regexp_operation.h
```diff
@@ -58,7 +73,8 @@ class RegExpOpr {
   std::map<PairNodeStr, Node> d_dv_cache;
   std::map<PairNodeStr, std::pair<Node, int> > d_deriv_cache;
   std::map<Node, std::pair<Node, int> > d_compl_cache;
-  std::map<Node, bool> d_cstre_cache;
+  /** cache mapping regular expressions to whether they contain constants */
+  std::unordered_map<Node, RegExpConstType, NodeHashFunction> d_constCache;
```
4tXJ7f: This is very nit-picky but I think a better name would be `d_regExpTypeCache`

src/theory/strings/regexp_solver.cpp
```diff
@@ -264,9 +269,13 @@ bool RegExpSolver::checkEqcIntersect(const std::vector<Node>& mems)
...
-    if (!d_regexp_opr.checkConstRegExp(m))
+    RegExpConstType rct = d_regexp_opr.getRegExpConstType(m);
+    if (rct == RE_C_VARIABLE
+        || (options::stringRegExpInterMode() == RE_INTER_CONSTANT
+            && rct != RE_C_CONRETE_CONSTANT))
     {
-      // cannot do intersection on RE with variables
+      // cannot do intersection on RE with variables, or with re.allchar based
+      // on option.
       continue;
```
4tXJ7f: Hmm, how about allowing the one of the regexps to have `re.allchar`s? Our intersections shouldn't be too bad in that case.

src/options/strings_options.toml
```diff
@@ -216,3 +216,13 @@ header = "options/strings_options.h"
...
+
+[[option]]
+  name       = "stringRegExpInterMode"
+  category   = "expert"
+  long       = "re-inter-mode=MODE"
+  type       = "CVC4::theory::strings::RegExpInterMode"
+  default    = "CVC4::theory::strings::RE_INTER_CONSTANT"
+  handler    = "stringToRegExpInterMode"
+  includes   = ["options/strings_modes.h"]
+  help       = "determines which regular expressions to compute intersections"
```
4tXJ7f:
```suggestion
  help       = "determines which regular expressions intersections to compute"
```

src/options/strings_modes.h
```diff
@@ -9,24 +9,22 @@
...
- **
- ** Modes for processing looping word equations in the string solver.
+ ** \brief Modes for the string solver.
  **/
 
 #include "cvc4_private.h"
 
-#ifndef CVC4__THEORY__STRINGS__PROCESS_LOOP_MODE_H
-#define CVC4__THEORY__STRINGS__PROCESS_LOOP_MODE_H
+#ifndef CVC4__BASE__STRINGS_MODES_H
```
4tXJ7f: The header guard should follow the namespaces, so `CVC4__THEORY__STRINGS__STRINGS_MODES_H`

src/theory/strings/regexp_operation.cpp
```diff
@@ -47,24 +47,61 @@ RegExpOpr::RegExpOpr()
...
+      }
+    }
+    else if (it->second == RE_C_UNKNOWN)
+    {
+      RegExpConstType ret = RE_C_CONRETE_CONSTANT;
+      for (const Node& cn : cur)
+      {
+        it = d_constCache.find(cn);
+        Assert(it != d_constCache.end());
+        if (it->second > ret)
```
4tXJ7f: Maybe add a comment to the `enum` that the order matters or assert that the order between the `enum` values is as you expect them to be.

src/theory/strings/regexp_operation.h
ajreynol: This would actually be `d_regExpConstTypeCache` though?  I prefer the shorter name; the other words there are not very unique to this cache.

src/theory/strings/regexp_solver.cpp
ajreynol: I can add this as a mode.

src/theory/strings/regexp_operation.cpp
ajreynol: Done.

src/theory/strings/regexp_operation.h
4tXJ7f: Hm, I just thought that `d_constCache` is very generic whereas `d_regExpTypeCache` is at least saying that we cache the type of regexps that we've seen. Not very critical though.

src/theory/strings/regexp_solver.cpp
```diff
@@ -264,15 +270,29 @@ bool RegExpSolver::checkEqcIntersect(const std::vector<Node>& mems)
...
     }
+    if (options::stringRegExpInterMode() == RE_INTER_ONE_CONSTANT)
+    {
+      if (!mi.isNull() && rcti == RE_C_CONSTANT && rct == RE_C_CONSTANT)
```
4tXJ7f:
```suggestion
      if (!mi.isNull() && rcti >= RE_C_CONSTANT && rct >= RE_C_CONSTANT)
```

ajreynol: Actually this doesn't change the behavior (m and mi cannot be set if their constant type is strictly greater than `RE_C_CONSTANT`), but I suppose this makes the code a bit more robust.

PR #3139, opened 2019-07-31
Move some generic utilities out of quantifiers

ajreynol:
This includes the term canonizer, and the match trie (discrimination tree) implementation used in the SAT 19 paper (which is split out of candidate_rewrite_filter.h/cpp).


src/expr/term_canonize.cpp
```diff
@@ -107,13 +106,22 @@ Node TermCanonize::getCanonicalFreeVar(TypeNode tn, unsigned i)
...
 }
 
+size_t TermCanonize::getIndexForFreeVariable(Node v) const
+{
+  std::map<Node, size_t>::const_iterator it = d_fvIndex.find(v);
+  if (it == d_fvIndex.end())
+  {
+    return 0;
+  }
+  return it->second;
```
HanielB: don't you wanna inline this?

ajreynol: I think the compiler takes care of these low level issues.

PR #3140, opened 2019-07-31
Collapse @ chains in SMT2 printer

HanielB:
This is useful for generating, from THF benchmarks, benchmarks that can be parsed by CVC4 in the HOL extension of SMT2


ajreynol:
I am wondering if this should be a feature of the printer.  It seems that this should be somewhere else (a preprocessing pass / a script?)  I am a bit wary when the printer is deceiving the user what the node itself is.

HanielB:
Maybe it'd be better if it were an option for the printer? The structure of the term does not affect the solver, only the parser. The use case of this for now is just to generate problems that CVC4 can parser.

ajreynol:
If CVC4 can't parse the output, shouldn't we extend the parser instead of the printer?

HanielB:
I think we should do both. :) It's a useful feature to normalize the output, but indeed it should not be the default case. So should we settle on opening an issue to extend the hosmt parser and to add an option to print normalized chains?

src/printer/smt2/smt2_printer.cpp
```diff
@@ -509,7 +509,30 @@ void Smt2Printer::toStream(std::ostream& out,
   // uf theory
   case kind::APPLY_UF: typeChildren = true; break;
   // higher-order
-  case kind::HO_APPLY: break;
+  case kind::HO_APPLY:
+    // collapse "@" chains, i.e.
+    //
+    // @(@(a, b), c) --> (@ a b c)
```
ajreynol: This isn't the actual smt2 syntax right? Can you comment on what is actually being printed?

HanielB: the syntax for the above example would be 

`((a b) c)` becomes `(a b c)`

ajreynol: I mean, comment in the code :)

HanielB: ah, right hahaha will do

PR #3148, opened 2019-08-02
Support default sygus grammar for strings

ajreynol:
Currently we include only concatenation and length.

This PR also updates when to apply single invocation techniques: if the CegInstantiator utility says that it cannot handle the FO quantifier, then we do not apply single invocation techniques.  In particular, this catches the case of strings + no grammars, where CVC4 would by default try to use CEGQI to solve this.


ajreynol:
@4tXJ7f this may be of interest to you.

ajreynol:
Synthesis of string functions is highly undecidable. I believe that coming up with a grammar that is "complete" is in fact impossible. I am using length and concatenation as a heuristic for now since other operators may be too expensive to check.

PR #3151, opened 2019-08-02
Call separate SMT engine for single invocation sygus

ajreynol:
This refactors the algorithm in sec 3 of http://homepage.divms.uiowa.edu/~ajreynol/cav15a.pdf to call a separate instance of the SMT engine.

Some indentation has changed in a few blocks of code.

The PR eliminates two options for minimizing the size of solutions from single invocation based on unsat core analysis, which was very lightly used/tested.

It also documents a few methods that were used.

This PR is in preparation for "incremental" sygus.


HanielB:
When you say the unsat cores options were lightly used you mean they were generally ineffective? It does sound like something we should be trying to do well, so maybe open an issue to revisit this?

ajreynol:
They weren't very effective. We don't need an issue to revisit.

src/theory/quantifiers/sygus/synth_conjecture.h
```diff
@@ -165,6 +162,8 @@ class SynthConjecture
   TermDbSygus* d_tds;
   /** The feasible guard. */
   Node d_feasible_guard;
+  /** do we have a solution in this user context? */
```
HanielB: please say more why it's necessary to keep this tracking at the level of the user context (which is not done for anything else in the module)

src/smt/smt_engine.h
```diff
@@ -962,15 +962,34 @@ class CVC4_PUBLIC SmtEngine {
...
+   *
+   * If q was a quantified formula that was instantiated on the last call to
+   * check-sat (i.e. q is returned as part of the vector in the method
+   * getInstantiatedQuantifiedFormulas above), then the list of instantiations
+   * of that formula that were generated are added to insts.
+   *
+   * In particular, if q is of the form forall x. P(x), then insts is a list
+   * of formulas of the form P(t1), ..., P(tn).
+   */
+  void getInstantiations(Expr q, std::vector<Expr>& insts);
```
HanielB: why are you doing this with expressions rather than with nodes? It introduces a level of indirection in the usage by the the single invocation module that is unnecessary. If expressions are needed for other applications of this method just have two versions?

src/smt/smt_engine.cpp
```diff
@@ -1803,8 +1795,12 @@ void SmtEngine::setDefaults() {
     Notice() << "SmtEngine: turning off cbqi to support instMaxLevel" << endl;
     options::cbqi.set(false);
   }
-  //track instantiations?
-  if( options::cbqiNestedQE() || ( options::proof() && !options::trackInstLemmas.wasSetByUser() ) ){
+  // Do we need to track instantiations?
+  // Needed for sygus due to single invocation techniques.
+  if (options::cbqiNestedQE()
+      || (options::proof() && !options::trackInstLemmas.wasSetByUser())
+      || is_sygus)
```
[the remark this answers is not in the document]
ajreynol: The only time we do instantiations with sygus is when we do single invocation.  :)
So there is no loss here; also notice that tracking instantiations isn't a huge overhead regardless (tends to add 10% overhead or so).

src/smt/smt_engine.h
ajreynol: This is an API-level call, so it must be expressions.

src/theory/quantifiers/sygus/synth_conjecture.h
ajreynol: I've documented, note this flag is in preparation of sygus + incremental (it will be clearer on future PRs).

src/smt/smt_engine.h
HanielB: So why not have an internal-level call? This function converts nodes to expressions, which the sygus module converts back to nodes afterwards.

ajreynol: There isn't any need to do an internal level call, since this is not critical to performance. Moreover, this code has the advantage of exercising the API (much like the parser does), which is good for code coverage.

PR #3152, opened 2019-08-02
Implement check abduct feature

### LINKED ISSUE #3126: --check-abduct feature
It would be useful to have a `--check-abduct` flag, that gets the abduction solution (see #3122) and verifies that it is indeed an abduct for the input.
Using the existing `--check-synth-sol` is possible, but will only check that the synthesized solution satisfies the SyGuS constraint -- this does not verify that the translation to a SyGuS problem was correct.

ajreynol:
Fixes #3126 .


src/smt/smt_engine.cpp
```diff
@@ -4918,6 +4918,70 @@ void SmtEngine::checkSynthSolution()
...
+void SmtEngine::checkAbduct(Expr a)
+{
+  Assert(a.getType().isBoolean());
+  Trace("check-abduct") << "SmtEngine::checkAbduct: get expanded assertions"
+                        << std::endl;
+
+  std::vector<Expr> asserts = getExpandedAssertions();
+  asserts.push_back(a);
+
+  // two checks: first, consistent with assertions, second, implies negated goal
```
yoni206: 1. Where is the goal?
2. In another PR, didn't you change the polarity of the goal?

ajreynol: For 1, this is stored at d_abdConj, a data member of SMT engine introduced in this PR. I am using conjecture/goal interchangably, I'll try to make this more consistent.
For 2, I changed the polarity on the smt2 interface, but didn't change how its stored internally.

PR #3154, opened 2019-08-02
Fix printing issue related to nested quotes

ajreynol:
This ensures we do choose names for certain identifiers we introduce (including bound variables and uninterpreted constants) that contain `|`.

### DIFF
--- a/src/expr/uninterpreted_constant.cpp
+++ b/src/expr/uninterpreted_constant.cpp
@@ -16,10 +16,10 @@
 
 #include "expr/uninterpreted_constant.h"
 
+#include <algorithm>
 #include <iostream>
 #include <sstream>
 #include <string>
-#include <algorithm>
 
 #include "base/cvc4_assert.h"
 
--- a/src/expr/uninterpreted_constant.cpp
+++ b/src/expr/uninterpreted_constant.cpp
@@ -34,7 +34,13 @@ UninterpretedConstant::UninterpretedConstant(Type type, Integer index)
 }
 
 std::ostream& operator<<(std::ostream& out, const UninterpretedConstant& uc) {
-  return out << "uc_" << uc.getType() << '_' << uc.getIndex();
+  std::stringstream ss;
+  ss << uc.getType();
+  std::string st(ss.str());
+  // must remove delimiting quotes from the name of the type
+  // this prevents us from printing symbols like |@uc_|T|_n|
+  std::replace( st.begin(), st.end(), '|', '@');
+  return out << "uc_" << st.c_str() << "_" << uc.getIndex();
 }
 
 }/* CVC4 namespace */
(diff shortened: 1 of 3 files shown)


src/expr/uninterpreted_constant.cpp
```diff
@@ -34,7 +35,18 @@ UninterpretedConstant::UninterpretedConstant(Type type, Integer index)
...
+  std::string st(ss.str());
+  // must remove delimiting quotes from the name of the type
+  // this prevents us from printing symbols like |@uc_|T|_n|
+  std::string q("|");
+  size_t pos;
+  while ((pos = st.find(q)) != std::string::npos)
+  {
+    st.replace(pos, 1, "");
+  }
+  return out << "uc_" << st.c_str() << "_" << uc.getIndex();
```
4tXJ7f: Sorry for being late to the party but wouldn't it be better to add this code to `maybeQuoteSymbol()` and apply that to `st`? According to SMT-LIB a sort is an `<identifier>`, which is an (indexed) `<symbol>`, so it would make sense to unify those implementations. Doing this is also a bit dangerous since we may up printing the same string for different sorts if we are unlucky.

Also, it shouldn't be necessary to call `st.c_str()` here (just `st` should be fine) and the loop can be written as [`st.erase(std::remove(st.begin(), st.end(), '|'), st.end())`](https://en.wikipedia.org/wiki/Erase%E2%80%93remove_idiom).

ajreynol: It is indeed true that we could map uninterpreted constants to the same identifier if there was an input language that accepted `|` within identifiers, although I'm not sure if any language we have is like this.  (FYI, we've had more serious issues with renaming constants before, see https://github.com/CVC4/CVC4/issues/1031 and the fix here: https://github.com/CVC4/CVC4/commit/e799ab0164722e8a4f192ee13223d0eeec6ec004#diff-331deedda9ea9cc1bf7a5b62f50a4ccd).

IMHO the right solution here is to catch the problem at the time the symbol is named (which is what this PR does). This ensures that no symbol is given an illegal name, regardless of whether it is from the input or introduced internally.  Otherwise the function `maybeQuoteSymbol` would have to be responsible for doing further renaming, which would be counter-intuitive.  However I do agree that we should probably have a more general solution than the code in this PR.

4tXJ7f: Oh, I didn't realize that this is being applied at creation time. Since it is only really a problem with internal symbols, wouldn't it be best to just assert that there are no `|` in the name? Less overhead and no problems with naming things.

ajreynol: It's not being applied at creation time.  :)

Sorry what I said may have been a bit confusing: this is the code that determines what the name of uninterpreted constants are, which is implemented within a custom print method.  My point was that we should always use legal names for terms that print as symbols (regardless of how they are sent to the smt2 printer).

Notice the reason why this code was wrong previously is because it was invoking the smt2 printer on the type, and `maybeQuoteSymbol` was introducing the `|`, so that uninterpreted constants could be named `uc_|T|` if the type `T` needed to be quoted.  This clearly is a name that we want to avoid.

PR #3155, opened 2019-08-04
Fix regression script for incremental SMT-LIB v2 benchmarks.

mpreiner:
The regression script did not extract the expected status from incremental
SMT-LIB v2 benchmarks correctly if status was given via (set-info :status ...). The
script used re.search for finding the status, which only searches for
the first occurrence instead of finding all (set-info :status ...).

This commit fixes the issue by using re.findall instead.


4tXJ7f:
LGTM :)

PR #3158, opened 2019-08-05
 Fix HO model construction for functions having Boolean arguments

### LINKED ISSUE #3136: HOL extension failing constant invariant in model generation
Running on debug with `--uf-ho --ho-elim` on  
[bug_f_defIsConst.zip](https://github.com/CVC4/CVC4/files/3452543/bug_f_defIsConst.zip)
 we get:

```(error "Assertion failure.
void CVC4::theory::TheoryModel::assignFunctionDefinition(CVC4::Node, CVC4::Node)
/home/hbarbosa/cvc/src/theory/theory_model.cpp:574:

  f_def.isConst()
")

### PR DESCRIPTION
**Pull request by ajreynol**
Fixes #3136. 

This fixes the function -> array conversion so that we process Boolean variables as conditions.

Previously the regression failed due to an assigned function model value `lambda x : Bool. x` which was not being interpreted as canonical (although it is).

### CODE REVIEW

File: src/theory/builtin/theory_builtin_rewriter.cpp [hunk 1]
```diff
@@ -214,26 +213,57 @@ Node TheoryBuiltinRewriter::getArrayRepresentationForLambdaRec(TNode n,
   std::vector< Node > conds;
   std::vector< Node > vals;
   Node curr = n[1];
-  while( curr.getKind()==kind::ITE || curr.getKind()==kind::EQUAL || curr.getKind()==kind::NOT ){
-    Trace("builtin-rewrite-debug2") << "  process condition : " << curr[0] << std::endl;
+  Kind ck = curr.getKind();
+  NodeManager* nm = NodeManager::currentNM();
```
HanielB: Nit: you could move this to beginning of the method so that you use `nm` also in the if block before this.

src/theory/builtin/theory_builtin_rewriter.cpp
```diff
@@ -214,26 +213,57 @@ Node TheoryBuiltinRewriter::getArrayRepresentationForLambdaRec(TNode n,
...
+      // lambda x. (ite (= x v) true false)
+      index_eq = curr;
+      curr_val = nm->mkConst(true);
+      next = nm->mkConst(false);
+    }
+    if (index_eq.getKind() != kind::EQUAL)
+    {
+      bool pol = index_eq.getKind() != kind::NOT;
+      Node indexEqAtom = pol ? index_eq : index_eq[0];
+      if (indexEqAtom.getKind() == kind::BOUND_VARIABLE)
```
HanielB: You made the control flow of this loop much more complicated... wouldn't it be simpler if you moved this test to beginning? You will might need to recompute a couple things but I think it would be best to not make the whole thing more complicated because of this special case.

ajreynol: I don't know how you mean to move this check to the beginning of the loop.  The first step of the loop is to determine `index_eq`, then we we have to check whether its an equality or a variable?
I'll add documentation on what is going on to make this clearer.

HanielB: You could turn it into an equality in the beginning, I mean, so that the rest of the loop can go back to how it was before.

ajreynol: There are still two cases: when the Boolean variable is a condition of an ITE, and when it is in the base case position.  So I would have to duplicate the control flow of the block between [1] and [2] in my latest documentation.

HanielB: I see now. Too bad. :(

PR #3163, opened 2019-08-06
Scripts for CASC-27

HanielB:


contrib/run-script-casc27-tfn-nr
```diff
@@ -0,0 +1,36 @@
+#!/bin/bash
```
ajreynol: You can exclude this one.

PR #3164, opened 2019-08-07
Add subdirectories to contrib for competition scripts

ajreynol:
The `contrib/` folder is overcrowded with competition scripts, this PR organizes this.


ajreynol:
I've updated the directory structure based on the suggestion.

mpreiner:
How about dropping the `Comp` for smt and sygus since it is in the `competitions` directory?

ajreynol:
I think I prefer the Comp there since it is accurate to the competition name (e.g. what if there was another kind of sygus comp?)

mpreiner:
Ok, how about `sygus-comp` and `smt-comp` then? :)

ajreynol:
Haha sure. All this coding lately has gotten me into using camel case for everything :)

PR #3165, opened 2019-08-07
Introduce smt2 parsing utility ParseOp and refactor

ajreynol:
This refactors the Smt2.g file to move logic into the Smt2Parser.  It introduces a ParseOp utility and refactors the code so that a lot of the logic can be seen as operations involving this utility.


src/parser/smt2/parse_op.h
```diff
@@ -0,0 +1,75 @@
...
+
+#include "expr/expr.h"
+#include "expr/kind.h"
+
+namespace CVC4 {
+
+/** A parsed operator
+ *
+ * The purpose of this class is to store information regarding a parsed term
+ * in the smt2 language that we are might not be ready to associate with an
```
mpreiner:
```suggestion
 * in the smt2 language that might not be ready to associate with an
```

src/parser/smt2/parse_op.h
```diff
@@ -0,0 +1,75 @@
...
+ * selecting from.
+ * - For array constant specifications prior to type ascription e.g. when we
+ * have parsed "const", we store (1), setting the kind to STORE_ALL.
+ * - For array constant specifications (as const (Array T1 T2)), we store (1)
+ * and (4), where kind is set to STORE_ALL and type is set to (Array T1 T2).
+ * When parsing this as an operator e.g. ((as const (Array T1 T2)) t), we
+ * interpret this operator by converting the next parsed constant of type T2 to
+ * an Array of type (Array T1 T2) over that constant.
+ */
+class ParseOp
```
mpreiner: Nit: How about changing this to `struct` if it's just storing data?

src/parser/smt2/smt2.cpp
```diff
@@ -1566,5 +1566,355 @@ InputLanguage Smt2::getLanguage() const
...
+    }
+    if (p.d_expr.isNull())
+    {
+      std::stringstream ss;
+      ss << "Could not resolve expression with name " << p.d_name
+         << " and type " << type << std::endl;
+      parseError(ss.str());
+    }
+  }
+  ExprManager* em = getExprManager();
```
mpreiner: Just as a note: In smt2.cpp `getExprManager()` is called quite often. It would probably make sense to store a `d_exprmgr` member in `Smt2`.

ajreynol: I tried doing this but its a bit tricky, since the expression manager may not be ready or may be changed from the time `Smt2` is created and when we parse.  More generally caching this pointer makes things a bit rigid since it makes the assumption that a single expression manager is used throughout the course of the parser object.

mpreiner: Why would we have different expression managers while parsing?

ajreynol: Say if we had an architecture where we used a single parser object to parse multiple files in multiple threads.  I'm not 100% sure of the intended architecture, but caching the expression manager in this class would introduce an assumption that we weren't planning on doing that kind of thing.

mpreiner: I'm not even sure if the parser object itself can handle multi-threading. I would not want to go there :-) The whole thread-safety/singleton business needs to be revisited anyways.

src/parser/smt2/smt2.cpp
```diff
@@ -1566,6 +1566,356 @@ InputLanguage Smt2::getLanguage() const
...
+    {
+      /* Special treatment for associative operators with lots of children
+       */
+      return em->mkAssociative(kind, args);
+    }
+    else if (!strictModeEnabled() && (kind == kind::AND || kind == kind::OR)
+             && args.size() == 1)
+    {
+      /* Unary AND/OR can be replaced with the argument.
+       * It just so happens expr should already be the only argument. */
```
4tXJ7f: This comment is out of date (there is no more `expr`).

src/parser/smt2/smt2.cpp
4tXJ7f: I tend to agree that we probably don't have to worry about multi-threading in that sense. We may want to support parallel parsing but in that case, we'd have likely one parser instance per thread because that seems like the most straightforward architecture. However, in the current version `getExprManager()` is only called three times, which I find reasonable.

PR #3167, opened 2019-08-08
Add string rewrite involving allchar stars

ajreynol:
Regular expressions that accept the empty string can be dropped from concatentations if they are between `(re.* re.allchar)`'s.


ajreynol:
Yes, you're right, that would be better, I'll update

ajreynol:
I've updated the code to do the more general check for when to drop components. I've also refactored the loop a bit (moving one of the rewrites to a separate loop, since the code was getting a bit complicated).

src/theory/strings/theory_strings_rewriter.cpp
```diff
@@ -814,10 +835,29 @@ Node TheoryStringsRewriter::rewriteConcatRegExp(TNode node)
...
-    // handles all cases where consecutive re constants are combined, and cases
-    // where arguments are swapped, as described in the loop above.
+    // handles all cases where consecutive re constants are combined or dropped
+    // as described in the loop above.
     return returnRewrite(node, retNode, "re.concat");
   }
+
+  // flipping adjacent star arguments
+  changed = false;
+  for (unsigned i = 0, size = cvec.size(); (i + 1) < size; i++)
```
```suggestion
  for (size_t i = 0, size = cvec.size() - 1; i < size; i++)
```
I find the change to `cvec.size() - 1` easier to read (just a suggestion though). Also, `size_t` is more appropriate here.

src/theory/strings/theory_strings_rewriter.cpp
```diff
@@ -814,10 +835,29 @@ Node TheoryStringsRewriter::rewriteConcatRegExp(TNode node)
...
+  {
+    if (cvec[i].getKind() == REGEXP_STAR && cvec[i][0] == cvec[i + 1])
+    {
+      // by convention, flip the order (a*)++a ---> a++(a*)
+      Node tmp = cvec[i + 1];
+      cvec[i + 1] = cvec[i];
+      cvec[i] = tmp;
```
4tXJ7f: AFAIK, this could be written as `std::swap(cvec[i], cvec[i + 1])`.

src/theory/strings/theory_strings_rewriter.cpp
```diff
@@ -763,49 +763,70 @@ Node TheoryStringsRewriter::rewriteConcatRegExp(TNode node)
...
       {
         curr = Node::null();
       }
     }
     if (!curr.isNull())
+    {
+      lastAllStar = false;
+      if (curr.getKind() == REGEXP_STAR)
+      {
+        // we can group stars (a*)++(a*) ---> a*
```
```suggestion
        // we can group stars (a)* ++ (a)* ---> (a)*
```
To be consistent with the comment below.

PR #3170, opened 2019-08-09
 New C++ API: Add checks and tests for Solver::simplify.

aniemetz:
Suggestion: Review commit-wise (easier because code move in first commit).

### DIFF
--- a/test/unit/api/solver_black.h
+++ b/test/unit/api/solver_black.h
@@ -52,7 +52,7 @@ class SolverBlack : public CxxTest::TestSuite
   void testMkAbstractValue();
   void testMkBitVector();
   void testMkBoolean();
-  void testMkBoundVar();
+  void testMkConst();
   void testMkEmptySet();
   void testMkFalse();
   void testMkFloatingPoint();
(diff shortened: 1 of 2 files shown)


test/unit/api/solver_black.h
```diff
@@ -949,6 +950,80 @@ void SolverBlack::testSetInfo()
...
+  Term v2 = d_solver->mkConst(d_solver->getIntegerSort(), "v2");
+  TS_ASSERT_THROWS_NOTHING(d_solver->simplify(v2));
+  Term f1 = d_solver->mkConst(funSort1, "f1");
+  TS_ASSERT_THROWS_NOTHING(d_solver->simplify(f1));
+  Term f2 = d_solver->mkConst(funSort2, "f2");
+  TS_ASSERT_THROWS_NOTHING(d_solver->simplify(f2));
+  d_solver->defineFunsRec({f1, f2}, {{b1, b2}, {b3}}, {v1, v2});
+  TS_ASSERT_THROWS_NOTHING(d_solver->simplify(f1));
+  TS_ASSERT_THROWS_NOTHING(d_solver->simplify(f2));
+}
```
mpreiner: How about adding a test where `simplify` actually throws an exception?

aniemetz: Can you think of a case that would not already trigger an exception before calling simplify?
I couldn't..... or wait. If the argument is a null Term, that should throw. Any other cases?

aniemetz: Added argument null check and test.

PR #3171, opened 2019-08-09
Fix handling of bvxnor operator with arity > 2.

mpreiner:
Previously, (bvxnor a b c) was translated to (bvxnor a (bvxnor b c))
instead of creating a node (bvxnor a b c). Since (bvxnor a b) == (bvnot
(bvxor a b)) the above translation was not correct for arity > 2.

For example, (bvxnor #b0 #b0 #b0) = #b1, however, with above (wrong)
translation it yields (bvxnor #b0 (bvxnor #b0 #b0)) = #b0

### DIFF
--- a/src/parser/smt2/Smt2.g
+++ b/src/parser/smt2/Smt2.g
@@ -2120,8 +2120,7 @@ termNonVariable[CVC4::Expr& expr, CVC4::Expr& expr2]
         if (args.size() > 2)
         {
           if (kind == CVC4::kind::INTS_DIVISION || kind == CVC4::kind::XOR
-              || kind == CVC4::kind::MINUS || kind == CVC4::kind::DIVISION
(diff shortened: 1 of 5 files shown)


4tXJ7f:
Hmm, the standard says:

> (bvxnor s_1 s_2 ... s_n) abbreviates (bvxnor (bvxnor s_1 s_2 ...) s_n).

So `(bvxnor #b0 #b0 #b0) = (bvxnor (bvxnor #b0 #b0) #b0) = #b0`.

mpreiner:
The standard doesn't make sense in this case. First it states that `bvxnor` is equivalent to `(bvnot (bvxor ...))` and with arity > 2 the behavior changes?

mpreiner:
Closing this since `bvxnor` will probably not be left-associative anymore in the near future.

PR #3172, opened 2019-08-09
Clean smt2 parsing of named attributes

### LINKED ISSUE #2919: Move helper functions out of SMT2 parser
Helper functions should be moved out of the parser into the Smt2 class:
- https://github.com/CVC4/CVC4/blob/master/src/parser/smt2/Smt2.g#L153
- https://github.com/CVC4/CVC4/blob/master/src/parser/smt2/Smt2.g#L183

ajreynol:
Fixes #2919.

### DIFF
--- a/src/parser/smt2/smt2.cpp
+++ b/src/parser/smt2/smt2.cpp
@@ -1566,16 +1566,17 @@ InputLanguage Smt2::getLanguage() const
   return em->getOptions().getInputLanguage();
 }
 
-
 Expr Smt2::setNamedAttribute(Expr& expr, const SExpr& sexpr)
 {
-  if(!sexpr.isKeyword()) {
+  if (!sexpr.isKeyword())
+  {
     parseError("improperly formed :named annotation");
   }
   std::string name = sexpr.getValue();
   checkUserSymbol(name);
   // ensure expr is a closed subterm
-  if(expr.hasFreeVariable()) {
+  if (expr.hasFreeVariable())
+  {
     std::stringstream ss;
     ss << ":named annotations can only name terms that are closed";
     parseError(ss.str());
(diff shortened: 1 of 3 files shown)


src/parser/smt2/smt2.cpp
```diff
@@ -1566,5 +1566,29 @@ InputLanguage Smt2::getLanguage() const
...
+    parseError("improperly formed :named annotation");
+  }
+  std::string name = sexpr.getValue();
+  checkUserSymbol(name);
+  // ensure expr is a closed subterm
+  if (expr.hasFreeVariable())
+  {
+    std::stringstream ss;
+    ss << ":named annotations can only name terms that are closed";
+    parseError(ss.str());
```
mpreiner: I guess there is no way to report the free variables here as it was done in the old code?

src/parser/smt2/smt2.h
```diff
@@ -425,6 +425,17 @@ class Smt2 : public Parser
...
     }
   }
+  /** Set named attribute
+   *
+   * This is called when expression expr is annotated with a name, i.e.
+   * (! expr :named sexpr). It sets up the necessary information to process
+   * this naming, including marking that expr is the last named term.
+   *
+   * We construct an expression symbol whose name is the name of s-expression
+   * which is used later for tracking assertions in unsat cores. This
```
mpreiner:
```suggestion
   * sexpr, which is used later for tracking assertions in unsat cores. This
```

The name of s-expression you are referring to is the argument `sexpr` right?

ajreynol: No, it is a symbol we construct whose name is the same as the `sexpr` (at least that's how its currently implemented).

src/parser/smt2/smt2.cpp
ajreynol: Not unless we extend the ExprManager interface; this is super low priority so I don't think it was worth the trouble.

PR #3173, opened 2019-08-09
Add option to only dump unsolved queries for --sygus-query-gen

ajreynol:
Adds an option to configure which queries to dump to file when using the query generation feature.

This is useful for dumping only the "hard" queries which a subsolver could not solve via `--sygus-expr-miner-check-timeout=N`.

### DIFF
--- a/src/theory/quantifiers/query_generator.cpp
+++ b/src/theory/quantifiers/query_generator.cpp
@@ -196,7 +196,7 @@ void QueryGenerator::dumpQuery(Node qy, unsigned spIndex)
   // Print the query and the query + its model (commented) to queryN.smt2
   std::vector<Node> pt;
   d_sampler->getSamplePoint(spIndex, pt);
-  unsigned nvars = d_vars.size();
+  size_t nvars = d_vars.size();
   AlwaysAssert(pt.size() == d_vars.size());
   std::stringstream fname;
   fname << "query" << d_queryCount << ".smt2";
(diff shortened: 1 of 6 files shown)


src/theory/quantifiers/query_generator.cpp
```diff
@@ -207,11 +179,51 @@ void QueryGenerator::checkQuery(Node qy, unsigned spIndex)
...
 
   d_queryCount++;
 }
 
+void QueryGenerator::dumpQuery(Node qy, unsigned spIndex)
+{
+  // Print the query and the query + its model (commented) to queryN.smt2
+  std::vector<Node> pt;
+  d_sampler->getSamplePoint(spIndex, pt);
+  unsigned nvars = d_vars.size();
```
4tXJ7f:
```suggestion
  size_t nvars = d_vars.size();
```

src/theory/quantifiers/query_generator.cpp
```diff
@@ -207,11 +179,51 @@ void QueryGenerator::checkQuery(Node qy, unsigned spIndex)
...
+  AlwaysAssert(pt.size() == d_vars.size());
+  std::stringstream fname;
+  fname << "query" << d_queryCount << ".smt2";
+  std::ofstream fs(fname.str(), std::ofstream::out);
+  fs << "(set-logic ALL)" << std::endl;
+  for (unsigned i = 0; i < 2; i++)
+  {
+    for (unsigned j = 0; j < nvars; j++)
```
4tXJ7f:
```suggestion
    for (size_t j = 0; j < nvars; j++)
```

PR #3176, opened 2019-08-09
Properly implement logic info for separation logic

### LINKED ISSUE #2230: CVC4 cannot differentiate between `QF_SEP` and `QF_S`
As part of #2229, it would be good to add a test that validates that a `RecoverableModalException` is raised if (e.g.,) `getHeadExpr` is called without using `QF_SEP`.

However, currently it is only possible use separation logic with `QF_ALL_SUPPORTED`. If you try to use `QF_SEP`, then CVC4 thinks it is `QF_S` (for strings) with the extra characters `EP` on the end:

```
LogicInfo::setLogicString(): junk ("EP") at end of logic string: QF_SEP
```

ajreynol:
Fixes #2230.  This changes logic strings that involve the separation logic theory (which was not properly implemented before) and enables it on a few regressions.

### DIFF
--- a/src/theory/logic_info.cpp
+++ b/src/theory/logic_info.cpp
@@ -283,10 +283,11 @@ std::string LogicInfo::getLogicString() const {
       if(!isQuantified()) {
         ss << "QF_";
       }
-      if(d_theories[THEORY_SEP]) {
+      if (d_theories[THEORY_SEP])
+      {
         ss << "SEP_";
         ++seen;
-      }    
+      }
       if(d_theories[THEORY_ARRAYS]) {
         ss << (d_sharingTheories == 1 ? "AX" : "A");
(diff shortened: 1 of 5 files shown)


src/theory/logic_info.cpp
```diff
@@ -412,6 +413,11 @@ void LogicInfo::setLogicString(std::string logicString)
     } else {
       enableQuantifiers();
     }
+    if (!strncmp(p, "SEP_", 3))
```
4tXJ7f:
```suggestion
    if (!strncmp(p, "SEP_", 4))
```

Otherwise, we'd match `"SEPx"`

ajreynol:
@4tXJ7f  I did update the logic_info_white unit test for the new expected logic strings.

PR #3177, opened 2019-08-09
Simplify how defined functions are tracked during parsing

### LINKED ISSUE #451: Defined functions screwy in DeclarationScope (Bugzilla #231)
**Imported from Bugzilla**
Reporter: *Morgan Deters*
Status: *NEW*
Severity: *enhancement*
Assigned to: *Clark Barrett*
Component: *{Internals} Expression package*
Milestone: *1.6*
Version: *master*
Platform: *All*
OS: *All*

On 2010-10-31 09:42:59 -0700, Morgan Deters wrote:
> Defined functions, as implemented, are a little strange.  The set of functions are kept as expression bodies (rather than names) in a CDSet<>.  So that means if you have a defined function f()\=x\+y and a let binding e\=x\+y, then e is considered a defined function?
> 
> Need to review this, and clean this up.

ajreynol:
This PR simplifies how defined functions are managed during parsing.

For some background, we used to insist that applications of defined functions use a special APPLY kind.  This meant that we needed to keep track of which functions were introduced with define-fun.  This was uncessary: recently, the APPLY kind was eliminated (https://github.com/CVC4/CVC4/pull/2976) where we uniformly use APPLY_UF instead.

This PR further simplifies the parser by not tracking which functions are defined and which aren't.  It eliminates several variants of how symbols are bound during parsing.

This fixes #451, which is no longer relevant after this commit.

### DIFF
--- a/src/parser/smt2/Smt2.g
+++ b/src/parser/smt2/Smt2.g
@@ -401,7 +401,7 @@ command [std::unique_ptr<CVC4::Command>* cmd]
       // permitted)
       // we allow overloading for function definitions
       Expr func = PARSER_STATE->mkVar(name, t,
-                                           ExprManager::VAR_FLAG_DEFINED, true);
+                                      ExprManager::VAR_FLAG_DEFINED, true);
       cmd->reset(new DefineFunctionCommand(name, func, terms, expr));
     }
   | DECLARE_DATATYPE_TOK datatypeDefCommand[false, cmd]
--- a/src/parser/smt2/Smt2.g
+++ b/src/parser/smt2/Smt2.g
@@ -745,7 +745,8 @@ sygusCommand [std::unique_ptr<CVC4::Command>* cmd]
     }
     ( symbol[name,CHECK_NONE,SYM_VARIABLE] {
         if( !terms.empty() ){
-          if( !PARSER_STATE->isDeclared(name) ){
+          if (!PARSER_STATE->isDeclared(name))
+          {
             std::stringstream ss;
             ss << "Function " << name << " in inv-constraint is not defined.";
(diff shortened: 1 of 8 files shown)


src/parser/smt2/Smt2.g
```diff
@@ -988,11 +988,9 @@ sygusGTerm[CVC4::SygusGTerm& sgt, std::string& fun]
           // fail, but we need an operator to continue here..
           Debug("parser-sygus")
               << "Sygus grammar " << fun << " : op (declare="
-              << PARSER_STATE->isDeclared(name) << ", define="
-              << PARSER_STATE->isDefinedFunction(name) << ") : " << name
+              << PARSER_STATE->isDeclared(name) << ") : " << name
               << std::endl;
-          if(!PARSER_STATE->isDeclared(name) &&
-             !PARSER_STATE->isDefinedFunction(name) ){
+          if(!PARSER_STATE->isDeclared(name) ){
```
4tXJ7f:
```suggestion
          if(!PARSER_STATE->isDeclared(name))
          {
```

src/parser/smt2/Smt2.g
```diff
@@ -745,7 +745,7 @@ sygusCommand [std::unique_ptr<CVC4::Command>* cmd]
     }
     ( symbol[name,CHECK_NONE,SYM_VARIABLE] {
         if( !terms.empty() ){
-          if( !PARSER_STATE->isDefinedFunction(name) ){
+          if( !PARSER_STATE->isDeclared(name) ){
```
4tXJ7f:
```suggestion
          if(!PARSER_STATE->isDeclared(name))
          {
```

PR #3179, opened 2019-08-10
cmake: Export CVC4 library interface.

mpreiner:
The `find_package` configuration information is now generated by CVC4's build system.

This allows cmake projects that use CVC4 to search for CVC4 libraries via  `find_package(CVC4)` and link CVC4 and all its dependencies via `target_link_libraries(<target> CVC4::cvc4)` instead of defining a custom finder module.


4tXJ7f:
Hmm, from a cursory look at Travis, it looks like making `target_include_directories` `PRIVATE` does not play well with the parser because the parser is not part of the CVC4 library but needs access to the definition of floating-point numbers. Any idea of how we could fix this? I suppose that long-term, we don't want src/util/floatingpoint.h to be part of the public API anymore, which would fix the problem.

mpreiner:
Well for now we have to mark all includes that are exported to the public API as `PUBLIC`, all others can be `PRIVATE`. As soon as the new API is default we can make all others also `PRIVATE`.

mpreiner:
Seems like the only dependencies that get exported to the public API are GMP, CLN and SymFPU.

CMakeLists.txt
```diff
@@ -481,6 +481,26 @@ if(BUILD_BINDINGS_JAVA OR BUILD_BINDINGS_PYTHON)
...
+#-----------------------------------------------------------------------------#
+# Package configuration
+#
+# Install config and configversion for CVC4 for find_package(CVC4).
+
+include(CMakePackageConfigHelpers)
+write_basic_package_version_file(
+  ${CMAKE_CURRENT_BINARY_DIR}/CVC4ConfigVersion.cmake
+  VERSION ${CVC4_RELEASE_STRING}
+  COMPATIBILITY SameMajorVersion
```
4tXJ7f: Hmm, isn't it more like `SameMinorVersion` for us? E.g. for 1.7 we [changed the API a bit](https://github.com/CVC4/CVC4/blob/eba03c5af0112aea04d83977333ae37e8a13137d/NEWS#L39-L40).

CMakeLists.txt
```diff
@@ -481,6 +481,26 @@ if(BUILD_BINDINGS_JAVA OR BUILD_BINDINGS_PYTHON)
...
+)
+
+# Install the config, configversion and custom find modules
+install(FILES
+  ${CMAKE_CURRENT_LIST_DIR}/cmake/CVC4Config.cmake
```
4tXJ7f: Why aren't we using `configure_package_config_file()` to generate the real `CVC4Config.cmake` file from a template as [recommended](https://cmake.org/cmake/help/latest/module/CMakePackageConfigHelpers.html#command:configure_package_config_file)?

CMakeLists.txt
mpreiner: Yep, changed.

CMakeLists.txt
mpreiner: Changed it to use `configure_package_config_file()`. Tested it with another cmake project and `find_package(CVC4)` works as expected.

mpreiner:
Yeah, that's a good idea!

PR #3180, opened 2019-08-11
Update option to disable symbolic definitions in strings

ajreynol:
This PR extends the stringsInferSym option to also cover the extended function evaluation lemma schema.

This option may be better to disable for Amazon benchmarks (although this PR does not change the default behavior).

This PR also does some minor improvements to trace messages in strings.


src/theory/strings/theory_strings.cpp
```diff
@@ -1778,7 +1791,12 @@ void TheoryStrings::checkExtfEval( int effort ) {
           //    x = "" => str.replace( x, x, x ) == ""
           //    y = "" => str.replace( y, y, y ) == ""
           Trace("strings-extf-debug") << "  get symbolic definition..." << std::endl;
-          Node nrs = getSymbolicDefinition( sn, exps );
+          Node nrs;
+          // only if option do we use symbolic definitions
```
4tXJ7f: Doesn't quite parse :)

src/theory/strings/theory_strings.cpp
```diff
@@ -1002,17 +1003,29 @@ void TheoryStrings::check(Effort e) {
...
       runStrategy(sbegin, send);
       // flush the facts
       addedFact = d_im.hasPendingFact();
       addedLemma = d_im.hasPendingLemma();
       d_im.doPendingFacts();
       d_im.doPendingLemmas();
+      if (Trace.isOn("strings-check"))
+      {
+        Trace("strings-check") << "  ...finish run strategy: ";
+        Trace("strings-check") << (addedFact ? "addedFact " : " ");
```
```suggestion
        Trace("strings-check") << (addedFact ? "addedFact " : "");
```

Mini-nit: I think the space is actually not necessary :)

src/theory/strings/theory_strings.cpp
```diff
@@ -1002,17 +1003,29 @@ void TheoryStrings::check(Effort e) {
...
+        Trace("strings-check") << (addedLemma ? "addedLemma " : " ");
```
```suggestion
        Trace("strings-check") << (addedLemma ? "addedLemma " : "");
```

src/theory/strings/theory_strings.cpp
```diff
@@ -1002,17 +1003,29 @@ void TheoryStrings::check(Effort e) {
...
+        Trace("strings-check") << (d_conflict ? "conflict " : " ");
```
```suggestion
        Trace("strings-check") << (d_conflict ? "conflict " : "");
```

PR #3182, opened 2019-08-13
 Track sygus variable to term relationship via attribute

ajreynol:
This generalizes support for sygus functions-to-synthesize whose formal argument lists encode free symbols. 

We now track an attribute on formal arguments to functions-to-synthesize to the term they encode (of the same name).  This allows us to print:

`(declare-fun A () (> x 0))`

instead of:

`(declare-fun A ((x Int)) (> x 0))`

when `x` is intended to refer to a free variable in the scope of, e.g. a `get-abduct` command.

This eliminates the need to track the substitution on solutions for get-abduct and makes our printing of abducts with `--sygus-stream` consistent with the non-streaming version.

### DIFF
--- a/src/theory/quantifiers/sygus/synth_conjecture.cpp
+++ b/src/theory/quantifiers/sygus/synth_conjecture.cpp
@@ -1017,7 +1017,7 @@ void SynthConjecture::printSynthSolution(std::ostream& out)
   {
     return;
   }
-  NodeManager * nm = NodeManager::currentNM();
+  NodeManager* nm = NodeManager::currentNM();
   for (unsigned i = 0, size = d_embed_quant[0].getNumChildren(); i < size; i++)
   {
     Node sol = sols[i];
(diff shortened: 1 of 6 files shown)


src/theory/quantifiers/sygus/synth_conjecture.cpp
```diff
@@ -1081,13 +1082,33 @@ void SynthConjecture::printSynthSolution(std::ostream& out)
       if (is_unique_term)
       {
         out << "(define-fun " << f << " ";
-        if (dt.getSygusVarList().isNull())
+        // Only include variables that are truly bound variables of the
+        // function-to-synthesize. This means we exclude variables that encode
+        // external terms. This ensures that --sygus-stream prints
+        // solutions with no arguments on the predicate for responses to
+        // the get-abduct command.
+        std::vector<Node> pvs;
```
yoni206: What is the job of pvs?

ajreynol: These are the variables that will be printed in the argument list.

PR #3183, opened 2019-08-13
SmtEngine: Reorganize class according to guidelines and some clean up.

aniemetz:
This reorganizes class SmtEngine according to our coding guidelines.

### DIFF
--- a/src/smt/smt_engine.h
+++ b/src/smt/smt_engine.h
@@ -65,18 +65,26 @@ class Model;
 class LogicRequest;
 class StatisticsRegistry;
 
+/* -------------------------------------------------------------------------- */
+
 namespace context {
   class Context;
   class UserContext;
 }/* CVC4::context namespace */
 
+/* -------------------------------------------------------------------------- */
+
 namespace preprocessing {
 class PreprocessingPassContext;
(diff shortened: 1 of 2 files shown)


src/smt/smt_engine.h
```diff
@@ -113,1089 +123,1108 @@ namespace theory {
...
+   * solve a sygus conjecture based on the current assertions. In particular,
+   * consider the input:
+   *   (assert A)
+   *   (get-abduct B)
+   * In the copy of the SMT engine where these commands are issued, we maintain
+   * A in the assertion stack. To solve the abduction problem, instead of
+   * modifying the assertion stack to remove A and add the sygus conjecture
+   * (exists I. ...), we invoke a fresh copy of the SMT engine and leave the
+   * assertion stack unchaged. This copy of the SMT engine can be further
+   * queried for information regar
```
ajreynol: You'll get a merge conflict with these (they were removed in https://github.com/CVC4/CVC4/pull/3182).

aniemetz: Good catch, in this case it actually didn't produce a conflict and I might have missed it.

PR #3195, opened 2019-08-15
Towards incremental SyGuS in SMT engine

ajreynol:
This modifies the SMT engine to work in incremental mode with check-synth calls.  It is not operational yet since the underlying SyGuS solver in CVC4 does not support multiple calls with synthesis conjectures yet.

The PR also simplifies some of the data maintained for SyGuS at the SMT engine.


src/smt/smt_engine.h
```diff
@@ -1237,6 +1237,18 @@ class CVC4_PUBLIC SmtEngine
 
   /** Container for the lemma input and output channels for this SmtEngine.*/
   LemmaChannels* d_channels;
+
+  /*---------------------------- sygus commands  ---------------------------*/
+
+  /**
+   * Set sygus conjecture is stale. This is called when new sygus constraints
```
HanielB: Please detail the whole "stale" business. It's not clear what it means neither why it is set in these cases.

src/smt/smt_engine.cpp
```diff
@@ -3905,28 +3904,41 @@ void SmtEngine::declareSynthFun(const std::string& id,
                                 bool isInv,
                                 const std::vector<Expr>& vars)
 {
+  SmtScope smts(this);
+  finalOptionsAreSet();
+  doPendingPops();
```
HanielB: Do you need to do this now because of the attributes or because of the stale business maybe popping stuff? I don't remember well the dependencies.

PR #3196, opened 2019-08-16
Use separate CMake project for CVC4 examples.

mpreiner:
This commit restructures the examples directory to use a separate CMake project to test CVC4's exported package configuration (`find_package(CVC4)`) as suggested in #3179. This also fixes the includes for the example files.

Note: In theory it should be enough to just include `<cvc4/cvc4.h>`, but in some examples we have to additionally include other public headers. This is probably due to the main header not including all public cvc4 headers. I'm not sure if this intended behavior or not, @4tXJ7f do you know more about that?


4tXJ7f:
> Note: In theory it should be enough to just include `<cvc4/cvc4.h>`, but in some examples we have to additionally include other public headers. This is probably due to the main header not including all public cvc4 headers. I'm not sure if this intended behavior or not, @4tXJ7f do you know more about that?

Judging by the comment [here](https://github.com/CVC4/CVC4/blob/master/src/include/cvc4.h#L14-L15) this seems to be intentional.

examples/CMakeLists.txt
```diff
@@ -45,56 +35,52 @@ macro(cvc4_add_example name src_files libs output_dir)
...
+
+  add_executable(${name} ${src_files_list})
+  target_link_libraries(${name} CVC4::cvc4 CVC4::cvc4parser)
+
+  # The test target is prefixed with the path,
   # e.g., for '<output_dir>/myexample.cpp'
-  #   we create test target 'example/<output_dir>/myexample'
-  #   and run it with 'ctest -R "example/<output_dir>/myunittest"'.
+  #   we create test target '<output_dir>/myexample'
+  #   and run it with 'ctest -R "<output_dir>/myunittest"'.
```
4tXJ7f:
```suggestion
  #   and run it with 'ctest -R "<output_dir>/myexample"'.
```

PR #3197, opened 2019-08-19
New C++ API: Add checks for Solver::checkValid and Solver::checkValidAssuming.

aniemetz:


src/api/cvc4cpp.cpp
```diff
@@ -3084,33 +3084,58 @@ Term Solver::simplify(const Term& t)
...
   CVC4_API_SOLVER_TRY_CATCH_END;
 }
 
 Result Solver::checkValid(void) const
 {
-  // CHECK:
-  // if d_queryMade -> incremental enabled
+  CVC4_API_SOLVER_TRY_CATCH_BEGIN;
+  CVC4_API_CHECK(!d_smtEngine->isQueryMade()
+                 || d_smtEngine->getOption("incremental").toString() == "true")
```
mpreiner: Would it be better to check `options::incrementalSolving()` here?

aniemetz: Definitely. I'm wondering why all option checks were already of this form. If possible, I'll change.

aniemetz: OK, so this requires to open an `ExprManagerScope` before calling `options::incrementalSolving` due to our internal singleton madness. The current solution is a bit ugly but we decided in an offline team discussion that we will get rid of singletons and clean up all of this in the near future. So for now we'll open the scope as I did in my most recent commit, and this will become obsolete in the future.

PR #3218, opened 2019-08-23
 Infer emptiness instead of splitting when a string equality rewrites to a constant

### LINKED ISSUE #3217: Incorrect sat result on QF_S formula
Hi, 
On the latest reversion of CVC4:
```
(declare-fun a () String)
(declare-fun b () String)
(declare-fun c () String)
(declare-fun d () String)
(assert 
    (or 
        (not (= ( str.suffixof "B" ( str.replace "A" b "B")) (= ( str.substr a 0 (str.len b)) "A"))) 
        (not (= (not (= c "A")) ( str.suffixof "A" ( str.replace "A" c "B"))))))
(assert (= a (str.++ (str.++ b "") d)))
(check-sat)     
(get-model) 
```
CVC4 will report `sat` on this formula, and give a model:
```
(model
(define-fun a () String "")
(define-fun b () String "")
(define-fun c () String "")
(define-fun d () String "")
)
```
If I feed this model to the formula, CVC4 will report `unsat`

OS: Ubuntu 18.04
Reversion: b58bdc9c5672430cf15914c64129136b24050152
Commandline: cvc4 --strings-exp --produce-model file.smt2

ajreynol:
Fixes #3217.

The issue was caused by a term  `(str.substr (str.replace "A" b "B") (+ (- 1) (str.len (str.replace "A" b "B"))) 1)` that is entailed not to be empty. Our inference scheme split on whether this term was empty, leading to the split `(or true false)` after rewriting, which was tautological and hence ignored. Thus we terminated with no lemmas.

The fix is to recognize these cases and infer the direction of the split that the rewriter infers.  This can be properly communicated via purification variables (e.g. a fresh variable for the term above).


ajreynol:
Changes to performance on CMU benchmarks is minimal (I don't think this case triggers for it).  I get no change in status apart from benchmarks close to timeout.

PR #3228, opened 2019-08-27
Remove SMT1 parser.

### LINKED ISSUE #1313: Unify handling of logics
Right now, smt1 handles logics [separately](https://github.com/CVC4/CVC4/blob/master/src/parser/smt1/smt1.cpp) from the other inputs. We should probably unify and simplify that.

### LINKED ISSUE #2948: Remove SMT1 parser
We have decided that it is not worth it to continue supporting SMT1. E.g. we don't want to migrate the parser to the new API.

mpreiner:
This commit removes the SMT1 parser infrastructure and adds the SMT2 translations of the SMT1 regression tests. For now this commit removes regression test `regress3/pp-regfile.smt` since the SMT2 translation has a file size of 887M (vs. 172K for the SMT1 version).

Fixes #2948 and fixes #1313.


mpreiner:
Yes, I hope I can fix it today.

PR #3230, opened 2019-08-28
[proof] Eliminate side-condition from ER signature

alex-ozdemir:
Yoni recently informed me that while I claimed in my SAT talk that our ER signature used no side conditions, it actually used one. This PR tweaks the axioms slightly to remove it.

We can merge this PR in the interest of truthful advertising, or not.


proofs/signatures/er.plf
```diff
@@ -64,13 +64,17 @@
...
 
-; Takes a `list` of literals and a clause, `suffix`, adds the suffix to each
-; literal and returns a list of clauses as a `cnf`.
-(program clause_add_suffix_all ((list clause) (suffix clause)) cnf
-         (match list
-                (cln cnfn)
-                ((clc l rest) (cnfc (clc l suffix)
-                                    (clause_add_suffix_all rest suffix)))))
+; Represents multiple conjoined clauses.
+; There is a list, `heads` of literals, each of which should is suffixed by the
```
yoni206:
```suggestion
; There is a list, `heads` of literals, each of which is suffixed by the
```

proofs/signatures/er.plf
```diff
@@ -64,13 +64,17 @@
...
+; same `tail`.
+(declare common_tail_cnf_t type)
```
yoni206: What does "_t" stand for? I expected to see "_n" as in other places.

proofs/signatures/er.plf
```diff
@@ -64,13 +64,17 @@
...
+(declare common_tail_cnf
+         (! heads clause
+         (! tail  clause common_tail_cnf_t)))
+
+; A member of this type is a proof of a common_tail_cnf.
+(declare common_tail_cnf_holds
+         (! cnf common_tail_cnf_t type))
```
yoni206: is the tail a cnf or a clause?

proofs/signatures/er.plf
```diff
@@ -89,37 +93,38 @@
...
-(declare cnfc_unroll_towards_bottom
-         (! first clause
-            (! rest cnf
-               (! pf (cnf_holds (cnfc first rest))
-                  (! pf_continuation (! r1 (holds first) (! r2 (cnf_holds rest) (holds cln)))
-                    (holds cln))))))
-
+;    1. a proof of some `common_tail_cnf`
+; Then
+;    1. the first axiom give you a proof of the first `clause` therein and
```
yoni206:
```suggestion
;    1. the first axiom gives you a proof of the first `clause` therein and
```

src/proof/er/er_proof.cpp
```diff
@@ -220,30 +221,43 @@ void ErProof::outputAsLfsc(std::ostream& os) const
   for (const ErDefinition& def : d_definitions)
   {
     os << "\n    (clausify_rat_elimination_def _ _ _ "
```
yoni206: Where is `clausify_rat_elimination_def` defined?

alex-ozdemir: Oh! It isn't. This is definitely broken. I'll find a benchmark that actually triggers this codepath and add it to the test suite.

proofs/signatures/er.plf
alex-ozdemir: It is suppose to stand for "type". My idea here was to define a new type (`common_tail_cnf_t`) and a constructor for that type (`common_tail_cnf`).
Note that this type is **not** intended to be recursive -- it's just a pair of `clause`s -- the first clause is a list of literals and the second is a clause. The idea is that  `common_tail_cnf head tail` represents a cnf in which each clause begins with a literal from `head`, and ends with the same suffix -- `tail`.

proofs/signatures/er.plf
alex-ozdemir: A clause, see above.

alex-ozdemir:
@yoni206 I've added the aforementioned test!

test/unit/proof/er_proof_black.h
```diff
@@ -414,63 +427,38 @@ void ErProofBlack::testErTraceCheckOutputMedium()
...
+  std::string pf_footer = R"EOF(
+          )
+        ))))))))
+      ))))
     )
     )EOF";
 
-  TS_ASSERT_EQUALS(filterWhitespace(lfsc.str()), filterWhitespace(out));
+  std::stringstream actual_pf;
+  actual_pf << proof::plf_signatures << pf_header << actual_pf_body.str() << pf_footer;
```
yoni206: Something isn't clear to me, that perhaps could be clarified by a comment. It looks like you create an lfsc proof in a string, and then call lfsc to check it. Why are you doing this in a cpp unit test and not as an lfsc test case?

alex-ozdemir: The goal of [`er_test.plf`](https://github.com/CVC4/CVC4/blob/master/proofs/signatures/er_test.plf) is to verify that the ER signature works; i.e. 

The goal of this unit test is to verify that the `ErProof::outputAsLfsc` function works correctly. I see that I accidentally stripped out the call to that function during my PR-preparation edits. Good catch!

PR #3231, opened 2019-08-28
 Fixes related to destructing null

### LINKED ISSUE #2888: Remove unnecessary null pointer checks
[An extra null pointer check is not needed in functions](https://isocpp.org/wiki/faq/freestore-mgmt#delete-handles-null "Do I need to check for null before delete p?") like the following.
- [CutInfo](https://github.com/CVC4/CVC4/blob/14b9dbaa0c9e8dce52d1a28595dc1cc80756abed/src/theory/arith/cut_log.cpp#L113 "Destructor for CutInfo")
- [IllegalArgumentException::formatVariadic](https://github.com/CVC4/CVC4/blob/047e75b485ad16a729083c210ba4064943d2e7c5/src/base/exception.cpp#L59 "IllegalArgumentException::formatVariadic function")
- [TypeSet](https://github.com/CVC4/CVC4/blob/14b9dbaa0c9e8dce52d1a28595dc1cc80756abed/src/theory/type_set.cpp#L22 "Destructor for TypeSet")

ajreynol:
Fixes #2888.

Also ensures we destruct two members of CutLog, where there was a bug.


src/theory/arith/cut_log.cpp
```diff
@@ -106,13 +106,11 @@ CutInfo::CutInfo(CutInfoKlass kl, int eid, int o)
   , d_cutVec()
   , d_mAtCreation(-1)
   , d_rowId(-1)
-  , d_exactPrecision(NULL)
-  , d_explanation(NULL)
+  , d_exactPrecision(nullptr)
```
elfring: Can these two initialisations be omitted after smart pointers are applied for the class members?

ajreynol: They probably can, although I'm not sure. @mpreiner  do you know?

mpreiner: You could, but I think until now we always explicitly initialized them.

elfring: How do you think about to pick any further opportunities up for source code reduction because of smart pointer usage?

PR #3233, opened 2019-08-28
Fix DAGification for printer.

mpreiner:
Due to the unique parent optimization the translation of
regress3/pp-regfile.smt to SMT2 increased the file size by a factor of
5000 (172k vs 880M).

To avoid tree traversals due to this optimization, this commit restricts
the unique parent check to nodes of arity 1.


ajreynol:
Unit test failing?

PR #3236, opened 2019-08-29
Remove portfolio

ajreynol:
This is a somewhat conservation elimination of components related to portfolio in CVC4.

There are other modifications that can likely be made that would involve some restructuring of the code, but this PR is "delete only".

Issue https://github.com/CVC4/CVC4/issues/2844 can be closed after this is merged, as it is no longer relevant.


src/prop/prop_engine.cpp
```diff
@@ -96,11 +96,13 @@ PropEngine::PropEngine(TheoryEngine* te, DecisionEngine *de, Context* satContext
 
   d_registrar = new theory::TheoryRegistrar(d_theoryEngine);
   d_cnfStream = new CVC4::prop::TseitinCnfStream
-    (d_satSolver, d_registrar, userContext,
+    (d_satSolver, d_registrar, userContext,true);
```
mpreiner: clang-format

PR #3239, opened 2019-08-29
Refactoring CEGQI interface

ajreynol:
This simplifies some of the interface for "instantiator" objects for counterexample-guided instantiation.  It also moves the `BvInverter` object from `QuantifiersEngine` to the `InstStrategyCegqi` module for the purposes of eliminating global pointers in quantifiers engine.


ajreynol:
Notice that many of these simplifications are due to the refactor from https://github.com/CVC4/CVC4/commit/c130a81b3578898dcb5cacaf746e4dd923e2c5d6, which made it so that we use a fresh copy of the SMT engine when using CEGQI for sygus.

src/theory/quantifiers/cegqi/ceg_arith_instantiator.cpp
```diff
@@ -267,7 +266,7 @@ bool ArithInstantiator::processAssertion(CegInstantiator* ci,
           << pvmod << " -> " << uval << ", styp = " << uires << std::endl;
     }
     // take into account delta
-    if (ci->useVtsDelta() && (uires == 2 || uires == -2))
+    if (uires == 2 || uires == -2)
```
aniemetz: Can `uires` (and `ires`) take on any value or only a specific set of values? If the latter, it may be a good idea to introduce macros for the values and give them names that hint at what the values mean.

ajreynol: I agree, looking into this a bit, this will be a fairly major refactor, I'll do this on a follow up PR.

aniemetz: SG

Issue #2851, opened 2019-03-07
Add API to retrieve representatives for uninterpreted sorts from model


4tXJ7f:

4tXJ7f:
See #2848 for context.

PR #3243, opened 2019-08-30
 Model API for domain elements

ajreynol:
This adds an API call from theory_model / model to get the domain elements of an uninterpreted sort.  It updates the smt2 printer to use this call.

@barrettcw , this is relevant to the discussion of whether CVC4 can reparse its own models (it is a step towards this).

The new API will have to call into TheoryModel::getDomainElements to add this feature.


ajreynol:
This fixes #2851.

src/smt/model.h
```diff
@@ -96,6 +96,14 @@ class Model {
    * is a predicate over t that indicates a property that t satisfies.
    */
   virtual std::vector<std::pair<Expr, Expr> > getApproximations() const = 0;
+  /** get the domain elements for uninterpreted sort s
+   *
+   * This method is gets the interpretation of an uninterpreted sort s.
```
aniemetz:
```suggestion
   * This method gets the interpretation of an uninterpreted sort s.
```

aniemetz: Also, what is `uninterpreted sort s` here -- the given `Type t`? Maybe make this more clear?

ajreynol: Yes, meant to say `t`.

PR #3254, opened 2019-09-04
Fix case of disjunctive conclusion in strings

### LINKED ISSUE #3247: Assertion failure with ---sygus-inference  and --strings-exp options
Hi,
for the following formula `cvc4 --sygus-inference --strings-exp` gives an assertion failure 

```
(declare-fun a () String) 
(declare-fun b () String) 
(declare-fun c () String)
(declare-fun d () String)
(declare-fun f () String)
(declare-fun e () String)
(assert
  (not
    (=
      (str.contains
        c
        (str.replace d (str.substr b 0 (str.len d)) "A")
      )
      (str.contains c "A")
    )
  )
)
(assert (= a (str.++ c f)))
(assert (= b (str.++ d e)))
(check-sat) 
```

```
void CVC4::theory::strings::TheoryStrings::assertPendingFact(CVC4::Node, bool, CVC4::Node)
/home/windomin/solver_binaries/cache/CVC4/src/theory/strings/theory_strings.cpp:1462:

  atom.getKind() != kind::OR
Infer error: a split.")
```
Revision: f727de33 (compiled with assertions)
OS: Ubuntu 18.04

ajreynol:
Fixes #3247. 

The error in the above issue was caused by the theory of strings inferring a conclusion e.g. `(str.contains s t)` that rewrote to a disjunction `(str.contains s1 t1) OR (str.contains s2 t2)`.  The strings inference manager would send a disjunction to the equality engine as a predicate. 

This fixes this case and also makes our check for redundancy in the inference that triggered this (CTN_Trans) more robust.


src/theory/strings/theory_strings.cpp
```diff
@@ -2077,6 +2077,20 @@ void TheoryStrings::checkExtfInference( Node n, Node nr, ExtfInfoTmp& in, int ef
             do_infer = pol ? !areEqual(lit[0], lit[1])
                            : !areDisequal(lit[0], lit[1]);
           }
+          else if (lit.getKind() == OR)
```
4tXJ7f: There are some rewrites of `str.contains` that result in an conjunction, so we should probably handle it here as well?

src/theory/strings/theory_strings.cpp
```diff
@@ -2077,6 +2077,20 @@ void TheoryStrings::checkExtfInference( Node n, Node nr, ExtfInfoTmp& in, int ef
...
+          {
+            do_infer = pol;
```
4tXJ7f: I don't quite get it. Let's say that we have `(not (or (str.contains x y) (str.contains z w)))` and both `(str.contains x y)` and `(str.contains z w)` are not equal to true in the current context. Don't we want to do `CTN_Trans` in that case?

ajreynol:
On second thought, I think we should avoid applying `CTN_Trans` if the conclusion rewrites, since the termination argument for this inference plus the other inferences that strings does isn't straightforward to me.  I've updated the code to guard against this; the 2 comments above are now outdated.

src/theory/strings/theory_strings.cpp
```diff
@@ -2065,46 +2065,37 @@ void TheoryStrings::checkExtfInference( Node n, Node nr, ExtfInfoTmp& in, int ef
...
-          conc = Rewriter::rewrite(conc);
-          conc = conc.negate();
-          bool do_infer = false;
-          bool pol = conc.getKind() != NOT;
-          Node lit = pol ? conc : conc[0];
-          if (lit.getKind() == EQUAL)
+          Node conc = Rewriter::rewrite(concOrig);
+          // For termination concerns, we only do the inference if the contains
+          // does not rewrite (and thus does not introduce new terms).
+          if( conc==concOrig )
```
4tXJ7f: Apply ClangFormat.

PR #3259, opened 2019-09-06
Disallow let in sygus grammars, check for free variables in sygus constructors

ajreynol:
We previously were not gracefully catching malformed sygus inputs like:

```
(declare-var z Int)
(declare-var w Int)
(synth-fun foo ((x Int) (y Int)) Int 
((Start Int))
 ((Start Int (z w (+ Start Start)))))
(constraint (= (foo z w) (+ z w)))
(check-synth)
```
This is illegal since z and w are not bound by the formal argument list of foo.

This PR has been updated to eliminate let terms from sygus grammars, which violated the above assertion.  Let terms in sygus grammars were previously handled in an extremely hacky way. This PR eliminates this code.  Notice that sygus let terms are no longer of importance since they have been eliminated from SyGuS comp; users should never use them since they are not guaranteed to work properly in the current implementation.


ajreynol:
Some old regressions fail due to the `let` terms allowed in sygus v1 grammars.

PR #3262, opened 2019-09-06
Infrastructure for instantiation rewriter

ajreynol:
This adds a generic callback of an "instantiation rewriter".  This detangles a hardcoded dependency of the core `Instantiate` class on `InstStrategyCegqi` and eliminates another global pointer from `QuantifiersEngine`.


src/theory/quantifiers/cegqi/inst_strategy_cegqi.h
```diff
@@ -181,6 +222,14 @@ class InstStrategyCegqi : public QuantifiersModule
   NodeIntMap d_nested_qe_waitlist_proc;
   std::map< Node, std::vector< Node > > d_nested_qe_waitlist;
 
+  /** Do nested quantifier elimination.
+   *
+   * This rewrites the quantified formulas in inst based on nested quantifier
+   * elimination. In this method, lem is the instantiation of quantified formula
```
mpreiner: Where is `lem`?

src/theory/quantifiers/cegqi/inst_strategy_cegqi.h
```diff
@@ -181,6 +222,14 @@ class InstStrategyCegqi : public QuantifiersModule
...
+   * q for the vector terms. The flag doVts is if we must apply virtual term
```
mpreiner:
```suggestion
   * q for the vector terms. The flag doVts indicates whether we must apply virtual term
```

src/theory/quantifiers/cegqi/inst_strategy_cegqi.h
ajreynol: Typo, should say `inst`, will fix.

PR #3265, opened 2019-09-09
Fix helper function in cegqi for strict bounds

ajreynol:
Might be causing the nightly failures (although I wasn't able to reproduce the failure).


aniemetz:
Was able to reproduce the nightlies failure, this PR does not fix it.

PR #3268, opened 2019-09-10
Avoid computing cardinality when constructing models

### LINKED ISSUE #3266: Benchmark for performance on simple ADT problems
As a user of CVC4, I like to get its output fast.
One of the use cases in which I use CVC4 is with data types.
I have attached a smt2.5 file in which I define a number of data types.
For each data type I introduce a variable
  and for each constructor of the data type, 
       I ask CVC4 whether I can make a data instance with this constructor (it should) and when that is the case I ask for a model.

CVC4 gives the right answers, yet its performance is slow.
[performanceADT.txt](https://github.com/CVC4/CVC4/files/3596135/performanceADT.txt)

I hope this file can help to improve the performance of CVC4 even further.

Note, if wanted I can provide a wide variety of these kind of files: they are generated as part of the unit tests of the new TorXakis version (under development).

ajreynol:
Fixes #3266.

That benchmark solved in ~169 seconds; it is solved in 0.03 seconds if we don't compute exact cardinality for datatypes.

There is still a few (very obscure) places where getCardinality can be called. Will fix in a follow-up PR.

@barrettcw , this may be of interest to you (I'm copying this code over from the dtNonSimpleRec branch).


src/expr/type_node.cpp
```diff
@@ -66,6 +66,15 @@ Cardinality TypeNode::getCardinality() const {
   return kind::getCardinality(*this);
 }
 
+/** Attribute true for types that are finite */
+struct IsFiniteTag
+{
+};
+struct IsFiniteComputedTag
```
aniemetz: Document? Or does the above apply to both?

PR #3270, opened 2019-09-11
Update to standard implementation of contains term

ajreynol:


src/expr/node_algorithm.cpp
```diff
@@ -118,6 +127,60 @@ bool hasSubtermMulti(TNode n, TNode t)
...
+  {
+    return true;
+  }
+
+  std::unordered_set<TNode, TNodeHashFunction> visited;
+  std::vector<TNode> toProcess;
+
+  toProcess.push_back(n);
+
+  for (unsigned i = 0; i < toProcess.size(); ++i)
```
mpreiner: for each?

ajreynol: `toProcess` changes over the course of the execution of the function actually. Does the for each iteration work in this case?

mpreiner: Oh I see, in this case leave it, but maybe add a comment so that it's clear when you look at the for loop header.

ajreynol:
Regarding the code duplication, that is true, although I'm not sure if we should be concerned about performance. My motivation for having 2 copies was performance.

mpreiner:
Hmm, do you think that it is a big performance hit if you pass a vector of size 1 instead of having the TNode version?

PR #3278, opened 2019-09-13
Move virtual term substitution utilities to own file and document

ajreynol:


src/theory/quantifiers/cegqi/vts_term_cache.h
```diff
@@ -0,0 +1,145 @@
...
+
+namespace quantifiers {
+
+/** Virtual term substitution term cache
+ *
+ * This class stores skolems corresponding to virtual terms (e.g. delta and
+ * infinity) and has methods for performing virtual term substitution.
+ *
+ * In detail, there are three kinds of virtual terms:
+ * (1) delta, of Real type, representing a infinitessimally small real value,
```
mpreiner:
```suggestion
 * (1) delta, of Real type, representing a infinitesimally small real value,
```

PR #3280, opened 2019-09-13
Move higher-order matching predicate

### LINKED ISSUE #1115: Revisit when to apply app completion for higher-order
In the theory UF solver on the ho-smt branch and pull requests, currently checkAppCompletion is too eager when introducing equivalences between APPLY_UF and HO_APPLY terms.  This should be revisited.

ajreynol:
This moves the higher-order matching predicate from TermUtil to TermDb (it is more appropriate to be there, which manages other globally used terms).

Will close #1115.

### DIFF
--- a/src/theory/uf/ho_extension.cpp
+++ b/src/theory/uf/ho_extension.cpp
@@ -316,7 +316,8 @@ unsigned HoExtension::checkAppCompletion()
             // add to pending list
             apply_uf[rop].push_back(n);
           }
-          // arguments are also relevant operators  FIXME (github issue #1115)
+          // Arguments are also relevant operators.
+          // It might be possible include fewer terms here, see #1115.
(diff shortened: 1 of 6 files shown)


src/theory/uf/ho_extension.cpp
```diff
@@ -316,7 +316,8 @@ unsigned HoExtension::checkAppCompletion()
             // add to pending list
             apply_uf[rop].push_back(n);
           }
-          // arguments are also relevant operators  FIXME (github issue #1115)
+          // Arguments are also relevant operators.
+          // It might be possible include fewer terms here, see #1115.
```
HanielB: Since you're gonna close the issue shouldn't you remove mention to it?

ajreynol: No, unless there is a FIXME or TODO, we can reference issues in comments. This is helpful for remembering what the issue was.

PR #3281, opened 2019-09-13
Support context-(in)dependent decision strategies.

ajreynol:
This allows three kinds of decision strategies to be registered with the DecisionManager: user-context dependent strategies, solve-local strategies, and context-independent strategies. 

Previously all strategies were treated as local to a `check-sat` call (the second kind above) and hence needed to be readded during `presolve` if we are in incremental mode.

This makes further work on sygus+incremental much easier to manage, as most sygus strategies are user-context-dependent.


src/theory/arrays/theory_arrays.cpp
```diff
@@ -1252,9 +1252,9 @@ bool TheoryArrays::collectModelInfo(TheoryModel* m)
 void TheoryArrays::presolve()
 {
   Trace("arrays")<<"Presolving \n";
-  // add the decision strategy
-  getDecisionManager()->registerStrategy(DecisionManager::STRAT_ARRAYS,
-                                         d_dstrat.get());
+  // add the decision strategy, which is user-context-independent
+  getDecisionManager()->registerStrategy(
+      DecisionManager::STRAT_ARRAYS, d_dstrat.get(), true);
```
HanielB: shouldn't the third argument here be `false`, which indicates context-independent?

src/theory/strings/theory_strings.cpp
```diff
@@ -570,8 +570,9 @@ void TheoryStrings::presolve() {
       inputVars.push_back(*itr);
     }
     d_sslds->initialize(inputVars);
+    // it is context-independent
     getDecisionManager()->registerStrategy(
-        DecisionManager::STRAT_STRINGS_SUM_LENGTHS, d_sslds.get());
+        DecisionManager::STRAT_STRINGS_SUM_LENGTHS, d_sslds.get(), true);
```
HanielB: ditto

src/theory/uf/cardinality_extension.cpp
```diff
@@ -504,8 +504,10 @@ void SortModel::initialize( OutputChannel* out ){
   if (d_c_dec_strat.get() != nullptr && !d_initialized)
   {
     d_initialized = true;
+    // Strategy is context-independent, since SortModel is constructed in a
+    // context-independent way.
     d_thss->getTheory()->getDecisionManager()->registerStrategy(
-        DecisionManager::STRAT_UF_CARD, d_c_dec_strat.get());
+        DecisionManager::STRAT_UF_CARD, d_c_dec_strat.get(), true);
```
HanielB: ditto

src/theory/arrays/theory_arrays.cpp
ajreynol: Yes, you're right, good catch.

Notice that in all three of these cases, the code is robust to the user-context-dependent-ness of the strategy, since it is always re-registered during presolve calls.

ajreynol: Actually it was a bit more complicated than i thought, we need a third option (strategies that are local to a solve scope).  I've made an enum in my latest commit, PTAL.

Issue #3269, opened 2019-09-11
assertion failure in getBuiltModel when using option tlimit 


yannickmoy:
We're using option --tlimit-per when calling CVC4 in SPARK as a way to both bound the effort and make results reproducible. On the VC attached, this leads to an assertion failure with the latest stable release 1.7:

$ cvc4-1.7-x86_64-linux-opt --tlimit-per=5000 pbm.smt2
unknown
(error "Assertion failure.
CVC4::theory::TheoryModel* CVC4::TheoryEngine::getBuiltModel()
/barrett/scratch/preiner/release-1.7/dest/src/cvc4-2019-04-09/src/theory/theory_engine.cpp:903:
  d_inSatMode && options::produceModels()
")

[pbm.zip](https://github.com/CVC4/CVC4/files/3601624/pbm.zip)

Is there a way to avoid this failure? 
That might relate to the inability sometimes to produce relevant models when interrupting CVC4, which is of great importance to us, as we'd rather have no model than an incorrect one.
Thanks

martin-cs:
It looks like a failure during model construction.  I have run into similar issues when trying to construct models after an "unknown" response.

For other CVC4 folks:

```
(set-option :produce-models true)
(set-option :incremental true)
;; ...
(check-sat)
(get-model)
```

The relevant parts of the SMT-LIB 2.6 standard are:

[ General semantics ]
"The solver must respond with an error when given a command not permitted in the
current mode."

[ `(check-sat)` ]
"On reporting sat or unknown the solver should move to sat mode—and then respond to
get-assignment, get-model, and get-value commands provided that the corresponding
enabling option is set to true."

[ `(get-model)` ]
"The next three commands can be issued only when the solver is in sat mode, and provide
information related to the most recent check command."

"The model A is required to satisfy all currently asserted
formulas and current assumptions only if the most recent check command reported sat."

So, as I read it, we should be returning something but it can really be anything.

Also, time limits are not necessarily reproducible.  Have you tried?

`(set-option :reproducible-resource-limit 5000)`

yannickmoy:
@martin-cs thanks for answer. Regarding reproducible limits, you're right in that case we're just using the time limit, as this is the call of CVC4 to generate a counterexample, not the call to attempt proof (we generate different VCs for these two calls).

ajreynol:
I can reproduce the assertion failure.

[...]

PR #3283, opened 2019-09-13
Return RecoverableModalException when model is not available.

ajreynol:
Possible fix for #3269. 

This returns a null pointer in `TheoryEngine` when the model is not available, and throws a `RecoverableModalException` in `SmtEngine` when this is the case.

It also makes the calls to getting the model in `SmtEngine` more standard: now everything goes through the internal call `getAvailableModel`.


src/theory/theory_engine.h
```diff
@@ -752,6 +752,9 @@ class TheoryEngine {
    *
    * If the model is not already built, this will cause this theory engine
    * to build to the model.
```
aniemetz:
```suggestion
   * to build the model.
```

PR #3291, opened 2019-09-16
Adding new scripts for CASC/TPTP

HanielB:


HanielB:
Done. I was indeed a bit puzzled by that "nr" on the other script :)

PR #3296, opened 2019-09-17
Add support for creating constant arrays to the new C++ API

makaimann:
I believe it is currently not possible to create a constant array through the new C++ API. This pull request adds this support.

### DIFF
--- a/src/api/cvc4cpp.cpp
+++ b/src/api/cvc4cpp.cpp
@@ -2558,6 +2558,7 @@ Term Solver::mkBitVector(uint32_t size, std::string& s, uint32_t base) const
 Term Solver::mkConstArray(Sort sort, Term val) const
 {
   CVC4_API_SOLVER_TRY_CATCH_BEGIN;
+  CVC4_API_ARG_CHECK_NOT_NULL(val);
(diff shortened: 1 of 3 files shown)


src/api/cvc4cpp.cpp
```diff
@@ -2555,6 +2555,16 @@ Term Solver::mkBitVector(uint32_t size, std::string& s, uint32_t base) const
   CVC4_API_SOLVER_TRY_CATCH_END;
 }
 
+Term Solver::mkConstArray(Sort sort, Term val) const
+{
+  CVC4_API_SOLVER_TRY_CATCH_BEGIN;
+  CVC4_API_CHECK(sort.isArray())  << "Not an array sort.";
+  CVC4_API_CHECK(sort.getArrayElementSort() == val.getSort())  << "Value does not match element sort.";
```
aniemetz: Isn't here a check missing? `val` may not be null, right? (macro `CVC4_API_ARG_CHECK_NOT_NULL`)

makaimann: Great catch, thanks! Just added that.

PR #3297, opened 2019-09-17
Decouple fmf-bound and finite-model-find

ajreynol:
This decouples `fmf-bound` (instantiation based on bounded integer quantification) and `finite-model-find` (UF sort minmization + instantiation based on finite uninterpreted sorts).  Previously, the latter was implied by the former.

It updates the default options when `strings-exp` is enabled (since now `fmf-bound` but not `finite-model-find` is not enabled by default).

One consequence of this is that by default, when `strings-exp` is enabled, we no longer do uninterpreted sort minimization via UF+cardinality.  This dependency was a bit counter-intuitive.


src/smt/smt_engine.cpp
```diff
@@ -1215,43 +1215,39 @@ void SmtEngine::setDefaults() {
...
       d_logic = d_logic.getUnlockedCopy();
       d_logic.enableQuantifiers();
       d_logic.lock();
       Trace("smt") << "turning on quantifier logic, for strings-exp"
                    << std::endl;
     }
-    if(! options::fmfBound.wasSetByUser()) {
+    if (!options::fmfBound.wasSetByUser())
+    {
+      // We require bounded quantifier handling.
```
4tXJ7f: For consistency, move comment out of `if` statement?

4tXJ7f:
Just two minor comments, LGTM otherwise.

ajreynol:
There is no change needed to the SMT run script.

PR #3301, opened 2019-09-19
Avoid cases of empty sygus grammars

ajreynol:
This fixes a few corner cases in the sygus v2 parser where the user generates an empty grammar, for example:

```
(synth-fun f ((x Int) (y Int)) Int
  ((Start Int) (B Bool))
  (
    (Start Int (x y 0 1 (ite B Start Start)))
    (B Bool ((Variable Bool)))
  )
)
```

It also:
- Fixes a bug where we would throw an exception for grammars containing only (Constant T),
- Catches malformed grammars with no production rules for a non-terminal symbol.


src/parser/smt2/Smt2.g
```diff
@@ -1122,6 +1122,30 @@ sygusGrammar[CVC4::Type & ret,
...
+      Trace("parser-sygus2") << "- " << datatypes[i].getName()
+                             << ", #cons= " << datatypes[i].getNumConstructors()
+                             << ", aci= " << aci << std::endl;
+      // We can be in a case where the only rule specified was (Variable T)
+      // and there are no variables of type T, in which case this is a bogus
+      // grammar. This results in the error below.
+      // We can also be in a case where the only rule specified was
+      // (Constant T), in which case we have not yet added a constructor. We
+      // ensure an arbitrary constant is added in this case.
+      if (datatypes[i].getNumConstructors()==0)
```
HanielB: clang-format

PR #3303, opened 2019-09-20
Add isParameterized function to Expr

ajreynol:
This fixes a bug in the sygus v2 parser where we were creating invalid Expr, in particular for BV extract terms.

AFAIK, there was a limitation where term conversions could not be applied in a generic way to Expr since it was not clear when the operator needed to passed back to `mkExpr`.  At the Node level, the check for this case is `getMetaKind() == kind::metakind::PARAMETERIZED`.  This makes this check available in Expr as a function `isParameterized()`.

@barrettcw this may be of interest to you.


ajreynol:
You mean add `isParameterized` to the new API?

It looks as though the new API is missing a number of related functions from Expr, including `getOperator`.  Perhaps we need a discussion to ensure we carry over the necessary functions from Expr to Term?

aniemetz:
Yeah, it probably misses functionality that was added recently. This should all be exported to the new API.

aniemetz:
For now, it would be great if you added this one already on this PR, and for the rest we could discuss how to proceed in the meeting tomorrow.

ajreynol:
I've added it to the API.

The unit test I would add would involve the functionality I discussed (re-constructing a node based on its children), however, this will only be possible once there is a way to do `getOperator`.

ajreynol:
Opened: https://github.com/CVC4/CVC4/issues/3314.

PR #3304, opened 2019-09-20
Return choice functions for approximate values in get-value

### LINKED ISSUE #3300: Accessing algebraic reals
For this benchmark:

```
(set-logic ALL)
(declare-fun x () Real)
(assert (= (* x x) 2))
(check-sat)
(get-value (x))
```

cvc4 says:

```
sat
((x (/ (- 741455) 524288)))
```

This is very confusing, as obviously square-root of 2 is not rational. The output of `get-model` is much more helpful:

```
sat
(model
(approximation (and (>= x (/ (- 1482917) 1048576)) (>= (* (- 1.0) x) (/ 741455 524288))))
(define-fun x () Real (/ (- 741455) 524288))
)
```

As an up-stream tool, I simply do a `get-value` on my variables; and it's not even clear how to access this "approximation" using the SMTLib interface without getting the whole model. Nor it is clear how to robustly parse that information.

Would it be worth considering an alternative design where `get-value` returns some indication that the value is only approximate?

For reference, this is what z3 prints for this benchmark:

```
sat
((x (root-obj (+ (^ x 2) (- 2)) 1)))
```

which makes it clear to the up-stream solver that there is an approximation going on.

What's the recommended way of accessing this info in CVC4? Would be great if CVC4 and z3 agreed on the presentation to simplify up-stream tools, though I should emphasize that I'm not advocating one way or the other; just something that both tools can standardize on.

ajreynol:
This returns choice functions in response to `get-value` if the value was approximate.  See discussion on #3300.

This partially addresses #3300. What remains is to provide approximate witnesses for choice functions.

### DIFF
--- a/src/theory/theory_model.cpp
+++ b/src/theory/theory_model.cpp
@@ -257,14 +257,14 @@ Node TheoryModel::getModelValue(TNode n, bool hasBoundVars) const
     return ret;
   }
   // it might be approximate
-  std::map< Node, Node >::const_iterator ita = d_approximations.find(n);
-  if (ita!=d_approximations.end())
+  std::map<Node, Node>::const_iterator ita = d_approximations.find(n);
+  if (ita != d_approximations.end())
   {
     // If the value of n is approximate based on predicate P(n), we return
     // choice z. P(z).
     Node v = nm->mkBoundVar(n.getType());
-    Node bvl = nm->mkNode(BOUND_VAR_LIST,v);
-    Node ret = nm->mkNode(CHOICE,bvl,ita->second.substitute(n,v));
+    Node bvl = nm->mkNode(BOUND_VAR_LIST, v);
+    Node ret = nm->mkNode(CHOICE, bvl, ita->second.substitute(n, v));
     d_modelCache[n] = ret;
     return ret;
   }
--- a/src/theory/theory_model.cpp
+++ b/src/theory/theory_model.cpp
@@ -256,6 +256,18 @@ Node TheoryModel::getModelValue(TNode n, bool hasBoundVars) const
     d_modelCache[n] = ret;
     return ret;
   }
+  // it might be approximate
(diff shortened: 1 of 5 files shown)


src/theory/theory_model.cpp
```diff
@@ -256,6 +256,18 @@ Node TheoryModel::getModelValue(TNode n, bool hasBoundVars) const
...
     return ret;
   }
+  // it might be approximate
+  std::map<Node, Node>::const_iterator ita = d_approximations.find(n);
+  if (ita != d_approximations.end())
+  {
+    // If the value of n is approximate based on predicate P(n), we return
+    // choice z. P(z).
+    Node v = nm->mkBoundVar(n.getType());
+    Node bvl = nm->mkNode(BOUND_VAR_LIST, v);
```
ahmed-irfan: in the rest of the file, it is kind::BOUND_VAR_LIST.
So, perhaps for readability, we should have the same here.

src/theory/theory_model.cpp
```diff
@@ -256,6 +256,18 @@ Node TheoryModel::getModelValue(TNode n, bool hasBoundVars) const
...
+    Node ret = nm->mkNode(CHOICE, bvl, ita->second.substitute(n, v));
```
ahmed-irfan: ditto

src/theory/theory_model.cpp
ajreynol: Actually, we are progressively trying to go the opposite direction. :)

Notice that we include namespace `CVC4::kind` at the top of this file.  Thus, the `kind::` is spurious.  There are a number of places in the CVC4 code base where `kind::` is unnecessarily used. As a rule of thumb, all new code I add doesn't use `kind::`. Progressively I'd like to remove all existing unnecessary `kind::` from existing code, but not all at once.

src/theory/theory_model.cpp
ajreynol: See above.

PR #3305, opened 2019-09-20
Fix printing of instantiation patterns

ajreynol:
Previously we were assuming all children of INST_PATTERN_LIST were of kind INST_ATTRIBUTE or INST_PATTERN, which is not the case.

### DIFF
--- a/src/printer/smt2/smt2_printer.cpp
+++ b/src/printer/smt2/smt2_printer.cpp
@@ -862,7 +862,8 @@ void Smt2Printer::toStream(std::ostream& out,
     out << ')';
     return;
   }
-  case kind::INST_PATTERN: break;
+  case kind::INST_PATTERN:
+  case kind::INST_NO_PATTERN: break;
   case kind::INST_PATTERN_LIST:
   {
     for (const Node& nc : n)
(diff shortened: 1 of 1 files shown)


src/printer/smt2/smt2_printer.cpp
```diff
@@ -862,7 +862,8 @@ void Smt2Printer::toStream(std::ostream& out,
     out << ')';
     return;
   }
-  case kind::INST_PATTERN: break;
+  case kind::INST_PATTERN:
+  case kind::INST_NO_PATTERN: break;
```
4tXJ7f: This case shouldn't really occur because of the `nc[0]` below, no?

ajreynol: That's right, although the kind should appear here regardless.

4tXJ7f: Should we maybe make it `Unreachable()` in that case?

ajreynol: No, we still should be able to print it out internally.  It's just that it's main use case (as a child of forall) forgoes printing it.

4tXJ7f: That's fair.

PR #3306, opened 2019-09-20
Make getSynthSolution return a Bool

ajreynol:
This is work towards sygus + incremental.

This adds the feature that we can explicitly know via a Bool return value whether a call to check-synth was successful.


src/smt/smt_engine.cpp
```diff
@@ -4795,12 +4795,16 @@ void SmtEngine::checkSynthSolution()
...
+  if (!d_theoryEngine->getSynthSolutions(sol_map))
   {
     Trace("check-synth-sol") << "No solution to check!\n";
     return;
   }
+  if (sol_map.empty())
+  {
+    Trace("check-synth-sol") << "Got empty solution!\n";
+    return;
+  }
```
HanielB: Shouldn't this be an assertion that if we got here. i.e. if we successfully got solutions for all conjectures, then the map is not empty? Or are we allowing conjectures that have no functions-to-synthesize?

ajreynol: This preserves the previous behavior, although I will think a bit whether it should change.

ajreynol:
Playing around with this, I think it's a good idea to throw an error when we call checkSynthSolution and there is not solution or it is an empty one.  

My latest commit does several fixes to ensure we are able to do this correctly.  This included moving where checkSynthSolution is invoked (from checkSatisfiability to checkSynth, where it belongs).

src/theory/quantifiers/sygus/synth_engine.cpp
```diff
@@ -376,15 +387,21 @@ void SynthEngine::printSynthSolution(std::ostream& out)
...
+  bool ret = true;
   for (unsigned i = 0, size = d_conjs.size(); i < size; i++)
   {
     if (d_conjs[i]->isAssigned())
     {
-      d_conjs[i]->getSynthSolutions(sol_map);
+      if (!d_conjs[i]->getSynthSolutions(sol_map))
+      {
+        // if one conjecture fails, we fail overall
+        ret = false;
```
HanielB: break after this

PR #3312, opened 2019-09-24
Add support for set comprehension

### LINKED ISSUE #3251: feature request: set comprehensions
Hi,

I've looked at http://cvc4.cs.stanford.edu/wiki/Sets and while I could work out how to do most things I am a missing set comprehensions.

So what I want to do is something like this, to build the set of all positive even integers:
```
{ i : Int | i >= 0 @ i * 2 }
```

Perhaps syntax like this: `(comprehension (DECLARATIONS) PREDICATE EXPRESSION)`

For example the above I could write as:
```
   (comprehension ((i Int)) (>= i 0) (* 2 i))
```

Or the set of prime numbers could be:
```
   (define-const n2 (Set Int) (comprehension ((i Int)) (>= i 2) i))
   (define-const primes (Set Int)
     (setminus n2
       (comprehension ((a Int) (b Int)) (and (member a n2) (member b n2)) (* a b))))
```

The use case here is that I am thinking about translating the Z notation to SMT-LIB, and the set comprehension is a very common concept there.

ajreynol:
This adds basic support for set comprehensions via a reduction to quantifiers.

Fixes #3251.


test/regress/regress1/sets/comp-positive.smt2
```diff
@@ -0,0 +1,12 @@
+; COMMAND-LINE: --sets-ext
+; EXPECT: unsat
```
yoni206: Would it be a good idea to also include at least one SAT test?

src/parser/parser.h
```diff
@@ -720,12 +720,28 @@ class CVC4_PUBLIC Parser {
...
     if(!bindingLevel) {
       d_assertionLevel = scopeLevel();
     }
   }
+  /**
+   * Pushes a scope with new bound variables are bound to types based on the
+   * entries in sortedVarNames. This method returns a bound variable list
+   * containing each of these variables.
+   */
+  Expr pushScopeWithDefs(
```
4tXJ7f: What do you think about doing something like `SynthFunFactory` (#3335), e.g. `DefinitionScope` to make sure that the popping of the scope happens automatically? We can address this in a separate PR.

ajreynol: Possibly. There are advantages and disadvantages to this in my opinion. Sometimes it's nice to just see the push/pop scopes in plain sight.

test/regress/regress0/sets/comp-qf-error.smt2
```diff
@@ -0,0 +1,14 @@
+; COMMAND-LINE: --sets-ext
+; EXPECT: (error "Set comprehensions require quantifiers in the background logic.")
+; EXIT: 1
+(set-logic QF_UFLIAFS)
+(set-info :status unsat)
```
yoni206: is it really unsat?

ajreynol: Not sure :) I'll remove.

ajreynol:
The semantics are currently given in the Kinds file.  I'm not sure if there is a better place for that?

I will add a bit more text there to make it clear what the impact of the semantics are wrt infinite sets.

src/theory/sets/kinds
```diff
@@ -53,6 +53,11 @@ nullaryoperator UNIVERSE_SET "(finite) universe set, all set variables must be i
 # A comprehension C with the above form has members given by the following
 # semantics:
 # forall y. ( exists x1...xn. P[x1...xn] ^ t[x1...xn] = y ) <=> (member y C)
```
yoni206: What is the range of `y`? Is it the sort of elements in C? Or is it the universe set?

ajreynol: Sort of elements in C, will add this in a comment.

yoni206: In that case, I don't understand the semantics of the `COMPREHENSION` operator.
If the sort is `Int`, for example, then according to the above formula, `C` can be infinite. But according to the semantics of the theory of sets in CVC4, `C` cannot be infinite.
The explanation below talks about CVC4's ability to construct models, but at for me, to better understand this operator, this explanation is missing a concrete definition of the `COMPREHENSION` operator.

A simple solution would be having `y` above to range over the universe set, but, is it a valid solution?

ajreynol: No, I don't think that would be intuitive for a user.  A set comprehension should be interpreted as enforced as being syntactic sugar for a (quantified) membership constraint. This in fact impacts the interpretation of the universe set, not the other way around.

Thus, the semantics for this operator need to be explained *axiomatically* (as above), not model-theoretically, which I think is what you are wanting. 

Since the only way to explain the semantics here is axiomatically, we necessarily need to talk about CVC4's ability to construct models.  Notice that this issue is no different from combining sets with arbitrary quantified formulas. COMPREHENSION is just syntax sugar.

yoni206: Completely agree, and so the question is not what is the semantics of comprehension (this is syntactic sugar), but what is the semantics of quantified set formulas.
Consider the following variant of one of your regressions:
```
(set-logic ALL)
(declare-fun x () (Set Int))
(assert (= x (comprehension ((z Int)) true (* 2 z))))
(check-sat)
```

Is this formula SAT or UNSAT?

ajreynol: Let's ask a simpler question. Is this SAT or UNSAT:
```
(set-logic ALL)
(declare-fun x () (Set Int))
(assert (forall ((z Int) (member (* 2 z) x)))
(check-sat)
```

yoni206: In my opinion it is UNSAT: x is a finite set of integers, and so it cannot contain all even numbers.
If the formula inside the quantifier was guarded by a "z in the universe", then I would think it is SAT.

Am I right? What do I get?

ajreynol: No, this is SAT.  However, CVC4 cannot construct a model for this input, because our theory solver limits itself to only generating models where all sets are interpreted as finite.

However, the absence of a (constructible) model does not mean that the problem is UNSAT.  Indeed in this example there are no inferences based on the axioms of the theory that would allow us to conclude UNSAT.  Thus, this problem is unsolvable by CVC4.

At an even more basic level (not related to finite sets), it is the same reason why this problem is unsolvable by standard SMT solvers:
```
(set-logic UF)
(declare-sort U 0)
(declare-fun a () U)
(declare-fun f (U) U)
(declare-fun g (U) U)
(assert (forall ((x U)) (not (= (f x) a))))
(assert (forall ((x U) (y U)) (=> (= (f x) (f y)) (= x y))))
(check-sat)
```
Since U and the interpretation of f/g must be infinite. 

You may be interested in this paper: https://link.springer.com/content/pdf/10.1007/s10817-010-9216-8.pdf

yoni206: The new example is clear to me, and thanks for the reference to the paper!

But the previous example (the one about finite sets) still isn't clear to me. 
If you go to the finite sets paper, you see that the theory of finite sets is defined by a structure in which the "Set" sort is interpreted as the set of finite subset of the interpretation of the "Element" sort (page 5).

In this structure, the above formula is not satisfiable (right?).

In the UF example, the definition of the UF theory is just first-order logic, and then yes, the formula should be SAT, but it is unsolvable due to the approach taken in SMT-solvers.

But in the finite sets example, at least according to my understanding of the paper that describes this theory, the problem itself is supposed to be UNSAT.

Or am I still missing something? Perhaps the implemented theory is some kind of an extension of the theory defined in the paper?

ajreynol: You're right, the IJCAR 2016 does mention that Set(T) is finite subset of T.  This paper considers quantifier-free set constraints, where no sets can be satified by *only* infinite models.

However, what is implemented in CVC4 does not make this distinction, as this would be highly unintuitive for users of sets+quantifiers, since if they accidentally defined an infinite set, the entire problem becomes UNSAT.

Thus, in practice, CVC4 implements a theory of sets where (in theory) the models are arbitrary subsets and the models it generates are finite.

yoni206: I see. Well, as a(n occasional) user of this theory in cvc4, including quantifier, I actually assumed the opposite of what's implemented (I just merged the ijcar 16 paper with the ordinary definition of quantifiers).

I think this means that we should document this distinction somewhere. Where would be the right place? I guess that this PR isn't the right place to discuss it :)

But it is the right place to discuss documentation of comprehension. I think that in the kinds file it should be clearly stated the the semantics is of infinite sets, and maybe include a variant of your last message here. What do you think? I would also put this explanation somewhere outside of the comprehension feature.

ajreynol: This should be on the wiki: http://cvc4.cs.stanford.edu/wiki/Sets

yoni206: Agreed.

src/theory/sets/theory_sets.cpp
```diff
@@ -15,7 +15,11 @@
  **/
 
 #include "theory/sets/theory_sets.h"
+#include "options/sets_options.h"
```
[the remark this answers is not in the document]
4tXJ7f: Still need to format again :)

PR #3313, opened 2019-09-24
Fixes for SyGuS + regular expressions

ajreynol:
This fixes numerous issues involving the combination of SyGuS and regular expressions.

Combining SyGuS and regular expressions may involve constructing regular expressions that are neither variables nor builtin regular expression operators.  The code was not robust for this case, either throwing spurious assertion failures or having incorrect behavior.

This PR fixes these issues.


src/theory/strings/regexp_operation.cpp
```diff
@@ -481,8 +505,8 @@ int RegExpOpr::derivativeS( Node r, CVC4::String c, Node &retNode ) {
         break;
       }
       default: {
-        //Trace("strings-error") << "Unsupported term: " << mkString( r ) << " in derivative of RegExp." << std::endl;
-        Unreachable();
+        Assert(!isRegExpKind(r.getKind()));
+        break;
```
4tXJ7f: I think that we also have to add a `ret = 0` here to say that the result is unknown (`ret` is 1 by default).

src/theory/strings/regexp_operation.cpp
```diff
@@ -655,8 +679,8 @@ Node RegExpOpr::derivativeSingle( Node r, CVC4::String c ) {
         break;
       }
       default: {
-        Trace("strings-error") << "Unsupported term: " << mkString( r ) << " in derivative of RegExp." << std::endl;
-        Unreachable();
+        Assert(!isRegExpKind(k));
+        break;
```
4tXJ7f: I think that we should return the null node here and then bail in `intersectInternal` if we get a null node, otherwise the intersection may be empty even though it should not be.

ajreynol: I'm going to change this back to `Unreachable();`. This shouldn't be called on regular expressions with subterms that are not regular expression kinds.

ajreynol:
I've addressed the review on this one, do let me know if there is anything else.

src/theory/strings/regexp_operation.cpp
```diff
@@ -791,8 +785,17 @@ void RegExpOpr::firstChars(Node r, std::set<unsigned> &pcset, SetNodes &pvset)
         firstChars(r[0], cset, vset);
         break;
       }
+      case kind::REGEXP_SIGMA:
       default: {
-        Assert(!isRegExpKind(k));
+        // should not call this function on regular expressions that aren't a
+        // standard regular expression kind.
+        Assert(k == REGEXP_SIGMA);
```
4tXJ7f: Hmm, something is off here: we execute this code for both `case kind::REGEXP_SIGMA` and `default` but we assert here that it is only for `REGEXP_SIGMA`. It seems that we should just make this the `REGEXP_SIGMA` case and make the `default` case `Unreachable()`?

ajreynol: That would change the behavior (in production).

The code here conservatively says any character can be the first character if we have sigma or an unknown operator. We additionally want to throw an assertion in debug, since we aren't expect to execute this for the latter.

4tXJ7f: Hmm, I understand your reasoning but I am really not a huge fan of code that changes behavior in production because it makes it much more difficult to spot potential issues (i.e. everything may be fine for our users but actually, it just works by luck). If we keep this code as-is, we should at least comment on why this is safe to do for an unknown operator.

ajreynol: The code is doing the right thing in production.  This is the common use case of a debug assertion (to proactively catch cases where we are executing something unexpected).

I can add a comment about why the behavior is correct for the default case.

ajreynol: I've added a comment, let me know what you think.

4tXJ7f: It is a common use case that I tend to disagree with :) However, the comment is good enough for me.

src/theory/strings/regexp_operation.cpp
```diff
@@ -767,9 +785,20 @@ void RegExpOpr::firstChars(Node r, std::set<unsigned> &pcset, SetNodes &pvset)
         firstChars(r[0], cset, vset);
         break;
       }
+      case kind::REGEXP_SIGMA:
       default: {
-        Trace("regexp-error") << "Unsupported term: " << r << " in firstChars." << std::endl;
-        Unreachable();
+        // we do not expect to call this function on regular expressions that
+        // aren't a standard regular expression kind. However, it we do, then
```
```suggestion
        // aren't a standard regular expression kind. However, if we do, then
```

src/theory/strings/regexp_operation.cpp
4tXJ7f: I think I brought this up previously but I think that we should have a different macro for these cases, e.g. `Expect()` instead of `Assert()` to signal that we do the right thing even if the condition does not hold.

PR #3315, opened 2019-09-26
Ackermann Task1: Make ackermannization generally applicable rather than just BV

Ying1123:
The ackermannization process is currently already support general theories rather than specifically for BV. In this pull request, an option has been added to turn on ackermannization independently.


src/CMakeLists.txt
```diff
@@ -46,8 +46,8 @@ libcvc4_add_sources(
   preprocessing/passes/bool_to_bv.h
   preprocessing/passes/bv_abstraction.cpp
   preprocessing/passes/bv_abstraction.h
-  preprocessing/passes/bv_ackermann.cpp
-  preprocessing/passes/bv_ackermann.h
+  preprocessing/passes/ackermann.cpp
+  preprocessing/passes/ackermann.h
```
4tXJ7f: Sort alphabetically

src/smt/smt_engine.cpp
```diff
@@ -1214,6 +1214,10 @@ void SmtEngine::setDefaults() {
   {
     d_logic = LogicInfo("QF_BV");
   }
+  else if (d_logic.getLogicString() == "QF_UFBV" && options::ackermann())
+  {
+    d_logic = LogicInfo("QF_BV");
+  }
```
4tXJ7f: If I am understanding correctly, this should apply to all logics that include `UF`. So it should be something like:
```
if (d_logic.isTheoryEnabled(THEORY_UF) && options::ackermann()) {
  d_logic.disableTheory(THEORY_UF);
}
```

src/smt/smt_engine.cpp
```diff
@@ -1214,6 +1214,10 @@ void SmtEngine::setDefaults() {
   {
     d_logic = LogicInfo("QF_BV");
   }
+  else if (d_logic.getLogicString() == "QF_UFBV" && options::ackermann())
+  {
+    d_logic = LogicInfo("QF_BV");
```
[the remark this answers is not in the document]
yoni206: see @4tXJ7f comment above.

test/regress/regress0/arith/ackermann1.smt2
```diff
@@ -0,0 +1,9 @@
+; COMMAND-LINE: --ackermann
```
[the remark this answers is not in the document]
4tXJ7f: Yes, I think that this is a great idea.

src/preprocessing/passes/ackermann.cpp
```diff
@@ -21,7 +21,7 @@
 ￼**   https://cs.nyu.edu/media/publications/hadarean_liana.pdf
  **/
 
-#include "preprocessing/passes/bv_ackermann.h"
+#include "preprocessing/passes/ackermann.h"
 
 #include "options/bv_options.h"
```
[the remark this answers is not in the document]
4tXJ7f: `#include "options/bv_options.h"` can be removed because it looks like it was only used for the `Assert` that this commit removes. However, `#include "theory/bv/theory_bv_utils.h"` is still needed for `bv::utils::mkAnd()`, which we should replace by a global `mkAnd` at some point but we don't have a good solution for that yet.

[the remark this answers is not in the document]
yoni206: Well, `mkAnd()` is pretty simple:
https://github.com/CVC4/CVC4/blob/923abd7000a2ab6e3c0776c59d159bdc3a4d9a52/src/theory/bv/theory_bv_utils.cpp#L324

I guess we could inline its code here and that way get rid of this dependency.

Ying1123:
> line 166:
> @what does ABV means? I currently delete this assert, since we are not supposing using the eager bitblasting, neither QF_BV.

Realized it should be array bit vector. I change the comment to "cannot use ackermannization on QF_ABV formula with stores"

src/smt/smt_engine.cpp
```diff
@@ -3238,7 +3242,12 @@ void SmtEnginePrivate::processAssertions() {
   if (options::bitblastMode() == theory::bv::BITBLAST_MODE_EAGER
       && !options::incrementalSolving())
   {
-    d_passes["bv-ackermann"]->apply(&d_assertions);
+    options::ackermann.set(true);
+  }
+
+  if (options::ackermann())
+  {
+    d_passes["ackermann"]->apply(&d_assertions);
```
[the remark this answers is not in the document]
Ying1123: I tried to make some changes. But I'm not sure about the condition. How does ackermannization related to incremental solving? Does ackermannization supports combined theory? And what is the proof you mentioned?
Needs second time review about this part.

4tXJ7f:
Yes, QF_ABV stands for quantifier-free bit-vectors with arrays. Since we are now applying the ackermannization pass also to non-bit-vector inputs maybe we should change the comment to: `Cannot use ackermannization on formula with stores to arrays`?

src/preprocessing/passes/ackermann.cpp
```diff
@@ -164,7 +170,7 @@ void collectFunctionsAndLemmas(FunctionToArgsMap& fun_to_args,
       {
         AlwaysAssert(
             term.getKind() != kind::STORE,
-            "Cannot use eager bitblasting on QF_ABV formula with stores");
+            "Cannot use ackermannization on QF_ABV formula with stores");
```
yoni206:
```suggestion
            "Cannot use Ackermannization on QF_ABV formula with stores");
```

src/smt/smt_engine.cpp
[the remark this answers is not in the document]
yoni206: 1. Ackermannization is not supported currently with incremental solving.
2. Which combined theory?
3. Actually, I think proofs are ok so you can ignore my comments on proofs (@4tXJ7f please correct me if I am wrong -- this should be no different than any other preprocessing hole in proofs, so we can support proofs but print holes right?)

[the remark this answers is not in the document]
Ying1123: 1. I changed the condition to run ackermann to with no incremental solving. But just want to point out, that there was a logic inconsistency in the code about eager bitblasting. In setDefaults(), the error is only threw when incrementalSolving() and !isPure(THEORY_BV), but in processAssertions(), it is turn off whenever incrementalSolving()=true. I will throw an error if incrementalSolving()=true and ackermann()=true.
2. The eager bitblast only supports pure BV, so how about ackermann? I currently didn't check if it is in a pure theory.

test/regress/regress0/bv/ackermann3.smt2
```diff
@@ -18,6 +18,5 @@
 
 (assert (= v0 v1))
 
-
```
4tXJ7f: Please remove this change.

src/smt/smt_engine.cpp
```diff
@@ -3235,10 +3273,9 @@ void SmtEnginePrivate::processAssertions() {
                          "Try --bv-div-zero-const to interpret division by zero as a constant.");
   }
 
-  if (options::bitblastMode() == theory::bv::BITBLAST_MODE_EAGER
-      && !options::incrementalSolving())
+  if (options::ackermann() && !options::incrementalSolving())
```
4tXJ7f: The check for whether incremental solving and ackermannization are both enabled is already done earlier, no need to repeat it here.

```suggestion
  if (options::ackermann())
```

src/preprocessing/passes/ackermann.cpp
```diff
@@ -35,8 +34,7 @@ namespace passes {
 
 /* -------------------------------------------------------------------------- */
 
-namespace
-{
+namespace {
```
4tXJ7f: This change seems unrelated and does not conform to our style. Please revert.

src/smt/smt_engine.cpp
```diff
@@ -1208,11 +1212,45 @@ void SmtEngine::setDefaults() {
...
+            || d_logic.isTheoryEnabled(THEORY_UF)))
+    {
+      if (options::produceModels.wasSetByUser())
+      {
+        throw OptionException(std::string(
+            "Ackermannization currently does not support model generation."));
+      }
+      Notice() << "SmtEngine: turn off ackermannization to support model"
+               << "generation" << endl;
+      options::ackermann.set(false);
```
4tXJ7f: I think that it is slightly better to disable model generation in that case because it makes the option checking easier (otherwise we are checking whether incremental solving has been enabled below even though we just disabled ackermannization).

src/smt/smt_engine.cpp
```diff
@@ -1208,11 +1212,45 @@ void SmtEngine::setDefaults() {
...
+    }
+
+    if (options::incrementalSolving())
+    {
+      throw OptionException(
+          "Incremental Ackermannization is currently not supported.");
+    }
+
+    if (options::ackermann())
```
4tXJ7f: If we make the change as I suggested above, there is no need to recheck the option.

yoni206:
Other than @4tXJ7f 's comments above, LGTM, no more comments on my side. Added @barrettcw as a reviewer also because I think he requested it.

src/smt/smt_engine.cpp
Ying1123: The option of produceModel is read-only. Seems I cannot set it to false.

src/preprocessing/passes/ackermann.cpp
Ying1123: This is changed by clang-format, and I see that other files in cvc4 are using the style with connected {. Are you sure to revert this?

4tXJ7f: Oh yes, you are right. It is actually matching our style (we have the `{` on separate line for other constructs in the language). However, we usually try to only format the changes and not the whole file (that way, we are slowly changing old files slowly to the new formatting style instead of having a huge commit changing the style in all our files). You can use `git clang-format <commit>` to format the diff between the current state and `<commit>`. For example, if your branch is up-to-date with `master`, you should be able to do `git clang-format master` and it should only format the changes that you made.

Ying1123: OK, thanks for pointing out.

src/smt/smt_engine.cpp
4tXJ7f: Ok, let's do it the other way around in that case. Let's have a separate block (i.e. outside of the current `if (options::ackermann())` block) that does something like:
```
if (options::produceModels()) {
  if (options::ackermann() && (d_logic.isTheoryEnabled(THEORY_ARRAYS) || d_logic.isTheoryEnabled(THEORY_UF)) {
    if (options::ackermann.wasSetByUser()) {
      // throw error
    }
    // disable ackermannizations
  }
}
```

This is cleaner (IMHO :) ) than rechecking below whether ackermannization is enabled.

src/smt/smt_engine.cpp
```diff
@@ -1208,11 +1212,45 @@ void SmtEngine::setDefaults() {
   {
     d_logic = LogicInfo("QF_NIA");
   }
-  else if ((d_logic.getLogicString() == "QF_UFBV"
-            || d_logic.getLogicString() == "QF_ABV")
-           && options::bitblastMode() == theory::bv::BITBLAST_MODE_EAGER)
+
+  // set options about ackermannization
+  if (options::ackermann())
```
4tXJ7f: This condition would actually move inside the check for `options::produceModels()` (see the code [my comment](https://github.com/CVC4/CVC4/pull/3315/files#r331704039)). Let's discuss briefly tomorrow to make sure that we are on the same page :)

yoni206:
Thanks for all the changes! LGTM.

Ying1123:
Thank you for your patient reviews!

PR #3319, opened 2019-09-26
Support UF in default sygus grammars

ajreynol:
This adds support for UF in default constructed sygus grammars.  This makes it possible to do `get-abduct` in logics with UF.

It also adds two related regressions, one for higher-order sygus.


src/theory/quantifiers/sygus/sygus_grammar_cons.cpp
```diff
@@ -520,18 +530,42 @@ void CegGrammarConstructor::mkSygusDefaultGrammar(
     Type unres_t = unres_types[i];
     //add variables
     for (unsigned j = 0, size_j = sygus_vars.size(); j < size_j; ++j)
```
HanielB: you forgot to remove this?

src/theory/quantifiers/sygus/sygus_grammar_cons.cpp
```diff
@@ -520,18 +530,42 @@ void CegGrammarConstructor::mkSygusDefaultGrammar(
...
+          ss << sv;
+          Trace("sygus-grammar-def")
+              << "...add for variable " << ss.str() << std::endl;
+          ops[i].push_back(sv.toExpr());
+          cnames[i].push_back(ss.str());
+          cargs[i].push_back(std::vector<Type>());
+          pcs[i].push_back(nullptr);
+          weights[i].push_back(-1);
+        }
+        else if (svt.isFunction() && svt.getRangeType() == types[i])
```
HanielB: Could you add a comment to say how you chose to handle function applications?

src/theory/quantifiers/sygus/sygus_grammar_cons.cpp
ajreynol: yes good catch :)

PR #3324, opened 2019-09-26
Do not substitute beneath arithmetic terms in the non-linear solver

### LINKED ISSUE #3307: Assertion failure NRA formula
Hi,
for the following NRA formula, CVC4 gives an assertion failure

```
(set-logic NRA)
(declare-fun a () Real)
(declare-fun b () Real)
(assert
 (and
  (> b 1)
  (< a 0)
  (>= (/ 0 (+ (* a b) (/ (- a) 0))) a)
  )
 )
(check-sat)
```

```
(error "Assertion failure.
bool CVC4::theory::arith::NonlinearExtension::simpleCheckModelMsum(const std::map<CVC4::NodeTemplate<true>, CVC4::NodeTemplate<true> >&, bool)
/home/windomin/solver_binaries/cache/CVC4/src/theory/arith/nonlinear_extension.cpp:1741:

  false
")

```
OS:Ubuntu 18.04
Revision: 7988675

ajreynol:
This fixes #3307 and #3472. 

The bug was caused by substituting within arithmetic terms that appear in non-linear monomials. This means that a term like `f(x)*f(x) + x*x` would turn into `f(7)*f(7) + 7*7` if the non-linear extension inferred that `x = 7`.  The problem was then that we have cached information for the term `f(x)` but not `f(7)`.  The solution is to not traverse beneath applications of functions that belong to a theory other than arithmetic for certain substitution operations.

Note that the bad behavior was hindering our ability to say "sat", although it is not related to correctness.


ajreynol:
Btw, I'm not super happy about adding a special substitution function in this PR; I considered generalizing this to a generic utility function.  This is up for discussion.

ahmed-irfan:
cool.... checking it now

src/theory/arith/nonlinear_extension.cpp
```diff
@@ -387,6 +387,85 @@ Node NonlinearExtension::computeModelValue(Node n, unsigned index) {
...
+      std::vector<Node> children;
+      if (cur.getMetaKind() == kind::metakind::PARAMETERIZED)
+      {
+        children.push_back(cur.getOperator());
+      }
+      for (const Node& cn : cur)
+      {
+        it = visited.find(cn);
+        Assert(it != visited.end());
+        Assert(!it->second.isNull());
```
mpreiner: Use `CVC4_DCHECK`

src/theory/arith/nonlinear_extension.cpp
```diff
@@ -387,6 +387,85 @@ Node NonlinearExtension::computeModelValue(Node n, unsigned index) {
...
+      }
+      if (childChanged)
+      {
+        ret = nm->mkNode(cur.getKind(), children);
+      }
+      visited[cur] = ret;
+    }
+  } while (!visit.empty());
+  Assert(visited.find(n) != visited.end());
+  Assert(!visited.find(n)->second.isNull());
```
mpreiner: Ditto

ajreynol:
We decided to generalize the `arithSubstitute` in this PR to a general utility with callbacks.

ajreynol:
After some more thought about this, I think I actually am happy with the status of this PR. The downside of generalizing and moving arithSubstitute to expr/node_algorithm.h is that we may be overdesigning.  I think we should only generalize this when there are at least 2 compelling variants of substitute that fit this pattern, but after some thought I'm not sure there is even another use case in the existing code. 

@ahmed-irfan @4tXJ7f @mpreiner what do you think?

ahmed-irfan:
there is a conflict in `test/regress/CMakeLists.txt`

PR #3327, opened 2019-09-27
Disable proofs for unsupported logics

yoni206:
This PR makes CVC4 complain if the user asked for proofs for an unsupported logic (in this contest, `ALL` is considered unsupported).
Changes in the regression script are introduced as well, in order to only request proofs for regressions in supported logics.


src/smt/smt_engine.cpp
```diff
@@ -2238,6 +2238,10 @@ void SmtEngine::setDefaults() {
 
   if (options::proof())
   {
+    if (d_logic > LogicInfo("QF_AUFBVLRA")) {
+        throw OptionException(
+            "Proofs are only supported for sub-logics of QF_AUFBVLIA."); 
+    }
```
4tXJ7f: We can now remove the check in `getProof()` since it is covered here.

test/regress/run_regression.py
```diff
@@ -79,6 +80,32 @@ def get_cvc4_features(cvc4_binary):
...
 
+def logic_supported_with_proofs(logic):
+    result = False
+    if logic in [
+            #single theories
+            "QF_BV",
+            "QF_UF",
+            "QF_A",
+            "QF_LRA",
+            #two theoreis
```
4tXJ7f:
```suggestion
            #two theories
```

test/regress/run_regression.py
```diff
@@ -79,6 +80,32 @@ def get_cvc4_features(cvc4_binary):
...
+            "QF_UFBV",
+            "QF_UFLRA",
+            "QF_AUF",
+            "QF_ALRA",
+            "QF_ABV",
+            "QF_BVLRA"
+            #Three theories
```
4tXJ7f:
```suggestion
            #three theories
```

test/regress/run_regression.py
```diff
@@ -79,6 +80,32 @@ def get_cvc4_features(cvc4_binary):
     return features
 
 
+def logic_supported_with_proofs(logic):
+    result = False
+    if logic in [
```
4tXJ7f: Can just write `return logic in [ ... ]` instead of the `if` statement :)

test/regress/run_regression.py
```diff
@@ -205,6 +233,8 @@ def run_regression(unsat_cores, proofs, dump, use_skip_return_code, wrapper,
             command_lines.append(line[len(COMMAND_LINE):])
         elif line.startswith(REQUIRES):
             requires.append(line[len(REQUIRES):].strip())
+        elif line.startswith(LOGIC):
+            logic = line[len(LOGIC):-1].strip()
```
4tXJ7f: I have a slight preference to deal with this the same as with the `status_regex`. I would define a `logic_regex` and then use it to extract the logic. This seems more consistent since setting the logic depends on the file format.

test/regress/run_regression.py
```diff
@@ -223,6 +251,8 @@ def run_regression(unsat_cores, proofs, dump, use_skip_return_code, wrapper,
             sys.exit('Cannot determine status of "{}"'.format(benchmark_path))
     if expected_exit_status is None:
         expected_exit_status = 0
+    if logic_regex:
+        logic = re.findall(logic_regex, benchmark_content)
```
4tXJ7f: This should be:

```suggestion
        logic_match = re.search(logic_regex, benchmark_content)
        if logic_match:
            logic = logic_match[1]
```

I think right now the `logic_supported_with_proofs` always fails because we are giving it a list of matches instead of a string.

test/regress/run_regression.py
```diff
@@ -79,6 +80,29 @@ def get_cvc4_features(cvc4_binary):
     return features
 
 
+def logic_supported_with_proofs(logic):
+    return logic in [
```
4tXJ7f: Maybe add an `assert isinstance(logic, str)` since we want to make sure that we are not returning `False` just because `logic` has the wrong type.

test/regress/run_regression.py
```diff
@@ -172,6 +196,9 @@ def run_regression(unsat_cores, proofs, dump, use_skip_return_code, wrapper,
         sys.exit('"{}" must be *.cvc or *.smt or *.smt2 or *.p or *.sy'.format(
             benchmark_basename))
 
+    logic_regex = None
+    if benchmark_ext == '.smt2':
+        logic_regex = r'\(\s*set-logic\s*(.*)\)'
```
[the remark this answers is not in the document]
yoni206: Combined.

test/regress/run_regression.py
yoni206: Added.

test/regress/run_regression.py
```diff
@@ -79,6 +79,30 @@ def get_cvc4_features(cvc4_binary):
     return features
 
 
+def logic_supported_with_proofs(logic):
+    assert logic == None or isinstance(logic, str)
```
4tXJ7f:
```suggestion
    assert logic is None or isinstance(logic, str)
```

PR #3335, opened 2019-10-02
[SMT2 Parser] Move code of `sygusCommand`

4tXJ7f:
This commit moves the code in `sygusCommand` in the SMT2 parser to the
`Smt2` class. The original code was pushing and popping the current
scope inline. This commit adds a class `SynthFunFactory` that takes care
of that upon creation and destruction.

### DIFF
--- a/src/parser/smt2/Smt2.g
+++ b/src/parser/smt2/Smt2.g
@@ -553,17 +553,14 @@ command [std::unique_ptr<CVC4::Command>* cmd]
 
 sygusCommand [std::unique_ptr<CVC4::Command>* cmd]
 @declarations {
-  std::string name, fun;
-  std::vector<std::string> names;
   Expr expr, expr2;
   Type t, range;
-  std::vector<Expr> terms;
-  std::vector<Expr> sygus_vars;
+  std::vector<std::string> names;
   std::vector<std::pair<std::string, Type> > sortedVarNames;
-  Type sygus_ret;
-  Expr synth_fun;
-  Type sygus_type;
+  std::unique_ptr<Smt2::SynthFunFactory> synthFunFactory;
+  std::string name, fun;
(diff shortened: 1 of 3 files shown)


4tXJ7f:
It may be worth talking about whether the solution with the factory class is something that we want. I'm happy to consider other designs.

src/parser/smt2/Smt2.g
```diff
@@ -589,99 +586,43 @@ sygusCommand [std::unique_ptr<CVC4::Command>* cmd]
...
+      // `grammar` specifies the required grammar for the function to
+      // synthesize, expressed as a type
+      sygusGrammarV1[grammar, synthFunFactory->getSygusVars(), fun]
     )?
     {
-      PARSER_STATE->popScope();
-      Debug("parser-sygus") << "...read synth fun " << fun << std::endl;
-      cmd->reset(
-          new SynthFunCommand(fun, synth_fun, sygus_type, isInv, sygus_vars));
+      *cmd = synthFunFactory->mkCommand(grammar);
```
aniemetz: I have a slight preference for using `reset` here.

4tXJ7f: Hmm, it doesn't look like you can use `reset` when passing in a `std::unique_ptr` (see [this](https://en.cppreference.com/w/cpp/memory/unique_ptr/operator%3D) for a description of the relationship between assignment and `reset`).

aniemetz: Isn't it sufficient to assign it without dereferencing? This would trigger a move anyways, right?

4tXJ7f: I've made `cmd` a return value instead of passing it as a pointer to an `std::unique_ptr`. PTAL

PR #3346, opened 2019-10-02
[CVC Parser] Add support for regular expressions

4tXJ7f:
This commit adds support for regular expressions in the CVC parser along
with a simple regression that illustrates and tests the use of the
regular expression operators.


4tXJ7f:
Any opinions on the syntax?

cc @barrettcw @tinelli

4tXJ7f:
We've discussed this offline during last week's meeting and we've decided that the syntax is OK for now and that we can add syntactic sugar if needed later.

ajreynol:
Regarding syntax, `STRING_IN_REGEXP` might be better as infix to be more in line with sets membership (infix `IS_IN`).  Otherwise, I agree the syntax is fine for now.

4tXJ7f:
I've overloaded `IS_IN` to be `STRING_IN_REGEXP` for non-sets. PTAL and let me know what you think. Do you prefer a separate keyword for the operator?

PR #3355, opened 2019-10-04
OpTerm Refactor: Allow retrieving OpTerm used to create Term in public C++ API

makaimann:
## Summary
This pull request refactors the way `OpTerms` are used and created. The main goal is to support retrieving the `OpTerm` used to create a Term through the public API. One important use-case is traversing an expression, possibly making some changes, and rebuilding.

For this to work, we need an `OpTerm` to represent any possible operator. Before this change, there were `Kind` operators, `OpTerm` operators, and function `Terms`. Based on discussions with maintainers, the proposed changes use a HOL-style semantics for function `Terms` and extend `OpTerm` to also wrap `Kinds`. 

## Specific Changes
* Function terms are treated just like any other Term. This means that the children of `(f x y)` are `f`, `x`, and `y` in the public API
  * This is a HOL-style interpretation of functions
  * This is convenient, because now we can always rebuild this Term by using `APPLY_UF` applied to the children.
* `OpTerm` is renamed to `Op`
* `Op` now holds a `Kind` and (optionally) an `Expr`. If it holds only a kind, then it represents a non-indexed operator. Indexed operators also hold an `Expr` representing the underlying operator.
  * This is important, because now we can retrieve an `Op` and apply it, even for non-indexed operators
* As a side effect of the change above, we can also simplify the `mkTerm` interface and remove all `*_OP` kinds from the public API. This is possible because the `Op` can always store the `Kind` needed in `mkTerm`.

   Previously you would use OpTerms like this:
    ```
    OpTerm ext = slv.mkOpTerm(BITVECTOR_EXTRACT_OP, 2, 0);
    Term exta = slv.mkTerm(BITVECTOR_EXTRACT, ext, a);
    ```
    
    The refactor would change the behavior to:   
    ```
    Op ext = slv.mkOp(BITVECTOR_EXTRACT, 2, 0);
    Term exta = slv.mkTerm(ext, a);
    ```

## A Few Flags for Reviewers

* The refactored `Op` stores an `Expr` and an _external_ `Kind`. Hopefully this isn't too confusing, I only did this so that the public API header file does not need to include internal kinds
* There is one place where an API function calls another API function. I could change that if you prefer, but I thought [this](https://github.com/makaimann/CVC4/blob/f4292fbe320a8148f4da2f07f26e798824bd0fcb/src/api/cvc4cpp.cpp#L2849) was cleaner
* Datatypes need to be special-cased
  * Constructor data types internally have kind `SKOLEM` or `BUILTIN`. However, we need the `Kind` stored in a constructor `Op` to be `APPLY_CONSTRUCTOR` for it to be able to make terms correctly.
  * [example 1](https://github.com/makaimann/CVC4/blob/f4292fbe320a8148f4da2f07f26e798824bd0fcb/src/api/cvc4cpp.cpp#L1704)
  * [example 2](https://github.com/makaimann/CVC4/blob/f4292fbe320a8148f4da2f07f26e798824bd0fcb/src/api/cvc4cpp.cpp#L1054)
* In previous discussions, we talked about refactoring so that indexed operators are basically just a fancy enum. Instead, I decided to keep the `Op` wrapping an `Expr`. My reasoning for this is that if you're traversing a formula and get an operator, you don't want to translate it to the enum representation, and then back to the `Expr` representation every time.

    e.g. the issue would look something like this
    ```
    // Op is just an enum with indices (Rust style)
    // Get operator from Term t
    // under the hood, getOp first retrieves an Expr, extracts the indices and creates an Op
    Op o = t.getOp();
    // Rebuild term
    // under the hood, the Op o is examined, and used to create an Expr 
    //   which is then used to build a new Expr that is then wrapped in a Term
    t = slv.mkTerm(o, std::vector(t.begin(), t.end());
    ```

## Final Thoughts

Additionally, we could actually remove all of the `mkTerm(Kind kind, ...` methods. Instead, we would just rely on the `mkTerm(Op op, ...` versions. This would implicitly construct an `Op` from a kind with the `Op(Kind k)` constructor. This would later be unpacked of course, but it would reduce code redundancy.

Please let me know your thoughts! I'd like to have several people look at this because it's somewhat significant change to the interface.


src/api/cvc4cpp.cpp
```diff
@@ -1062,6 +1232,35 @@ Sort Term::getSort() const
...
+
+OpTerm Term::getOpTerm() const
+{
+  CVC4_API_CHECK_NOT_NULL;
+  CVC4_API_CHECK(d_expr->hasOperator())
+      << "Expecting Term to have an OpTerm when calling getOpTerm()";
+  CVC4::Expr op = d_expr->getOperator();
+  CVC4::Type t = op.getType();
+
+  // special cases for Datatype operators
```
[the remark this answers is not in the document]
makaimann: Thanks for catching this! Addressed with 8c7645bd0261a7e60a227fb47861e9714689cada.

makaimann:
@aina, thanks for the initial feedback! I created a helper function to avoid calling an API function from another API function in 9f43a442d9956c6f33097756b2395846dd6edc2e. Then, I decided to do the OpTerm renaming in this PR with 0d476502ed6be37f3fa29ab3c7827f62bc9ca959 and d5fc69222ebaa9bbfcaea719b4df01c36d77b65f.

src/api/cvc4cpp.h
```diff
@@ -523,6 +523,122 @@ struct CVC4_PUBLIC SortHashFunction
...
 };
 
+/* -------------------------------------------------------------------------- */
+/* Op                                                                     */
+/* -------------------------------------------------------------------------- */
+
+/**
+ * A CVC4 operator term.
+ * An operator term is a term that represents certain operators, instantiated
+ * with its required parameters, e.g., a term of kind BITVECTOR_EXTRACT.
```
aniemetz: This still says `operator term` -- should we change this to `operator`?

src/api/cvc4cpp.h
```diff
@@ -523,6 +523,122 @@ struct CVC4_PUBLIC SortHashFunction
...
+
+  // !!! This is only temporarily available until the parser is fully migrated
+  // to the new API. !!!
+  CVC4::Expr getExpr(void) const;
+
+ private:
+  /* The kind of this operator term. */
+  Kind d_kind;
+
+  /** True iff this Op is indexed
```
aniemetz:
```suggestion
  /**
   * True iff this Op is indexed
```

src/api/cvc4cpp.h
```diff
@@ -590,6 +706,17 @@ class CVC4_PUBLIC Term
...
   Sort getSort() const;
 
+  /**
+   * @return true iff this term has an operator
+   */
+  bool hasOp() const;
+
+  /**
+   * @return the Op used to create this term
+   * safe to call when hasOp() returns true
```
aniemetz:
```suggestion
   * Note: This is safe to call when hasOp() returns true.
```

src/api/cvc4cpp.h
```diff
@@ -523,6 +523,122 @@ struct CVC4_PUBLIC SortHashFunction
...
+
+  /**
+   * Get the indices used to create this Op.
+   * Supports the following template arguments:
+   *   - string
+   *   - Kind
+   *   - uint32_t
+   *   - pair<uint32_t, uint32_t>
+   * Check the Op Kind with getKind() to determine which argument to use.
+   * @return the indices used to create this Op
```
aniemetz: What happens if the Op doesn't have indices?

src/api/cvc4cpp.h
```diff
@@ -523,6 +523,122 @@ struct CVC4_PUBLIC SortHashFunction
...
+  template <typename T>
+  T getIndices() const;
+
+  /**
+   * @return a string representation of this operator term
+   */
+  std::string toString() const;
+
+  // !!! This is only temporarily available until the parser is fully migrated
+  // to the new API. !!!
```
aniemetz: What happens if the Op doesn't have an Expr?

src/theory/quantifiers/first_order_model.h
```diff
@@ -103,7 +103,7 @@ class FirstOrderModel : public TheoryModel
   /** is model basis term */
   bool isModelBasisTerm(Node n);
   /** get model basis term for op */
-  Node getModelBasisOpTerm(Node op);
+  Node getModelBasisOp(Node op);
```
aniemetz: Was this change intentional?

src/api/cvc4cpp.h
```diff
@@ -730,9 +862,10 @@ class CVC4_PUBLIC Term
 
    private:
     /* The internal expression iterator wrapped by this iterator. */
-    void* d_iterator;
-    /* Constructor. */
-    explicit const_iterator(void*);
+    // void* d_iterator;
+    std::shared_ptr<CVC4::Expr> orig_expr;
```
aniemetz: Prefix with `_d` (see https://github.com/CVC4/CVC4/wiki/Code-Style-Guidelines#members).

src/api/cvc4cpp.h
```diff
@@ -730,9 +862,10 @@ class CVC4_PUBLIC Term
...
+    /* Keeps track of the iteration position */
+    uint32_t pos;
```

src/api/cvc4cpp.h
```diff
@@ -730,9 +862,10 @@ class CVC4_PUBLIC Term
 
    private:
     /* The internal expression iterator wrapped by this iterator. */
-    void* d_iterator;
-    /* Constructor. */
-    explicit const_iterator(void*);
+    // void* d_iterator;
```
aniemetz: Remove

src/api/cvc4cpp.h
makaimann: Oops, yep we should -- missed it with `sed`. Fixed with d19d4b3a2465d44ca8372af2676a0e902a47f654

src/api/cvc4cpp.h
makaimann: Also fixed with a332a47a5880592beb687dae8c90aa870eed316c. If it's only holding a Kind, it prints the Kind's toString. Otherwise it uses the Expr. Those should be the only two valid cases, and otherwise it throws an exception (e.g. if it's a null Op). Does this seem right to you?

In particular, I'm not sure what it should return if it only holds a Kind. Should it print something like `Op(PLUS)` or something like that so it doesn't look exactly the same as a Kind?

src/theory/quantifiers/first_order_model.h
makaimann: It sure wasn't -- sloppy `ag`/`sed` on my part. Thanks for catching that -- should have looked at the changed files. Fixed with d61f876d59aef39563da053b15a42fcc9aee0122 and a10fc0ce7b45befe1a67850156ec38005802fdbe.

src/api/cvc4cpp.h
makaimann: My bad -- fixed with 1407a11df2e54820bdf28adde350299ea90f6c58.

src/api/cvc4cpp.h
makaimann: Also fixed with 1407a11df2e54820bdf28adde350299ea90f6c58.

src/api/cvc4cpp.h
makaimann: Done with 4915d2a55c808a5525a58fe41bbe51a4e6f05a13.

test/unit/api/solver_black.h
```diff
@@ -911,6 +892,73 @@ void SolverBlack::testDefineFunsRec()
...
 }
 
+void SolverBlack::testUFIteration()
+{
+  Sort intSort = d_solver->getIntegerSort();
+  Sort funSort = d_solver->mkFunctionSort({intSort, intSort}, intSort);
+  Term x = d_solver->mkConst(intSort, "x");
+  Term y = d_solver->mkConst(intSort, "y");
+  Term f = d_solver->mkConst(funSort, "f");
+  Term fxy = d_solver->mkTerm(APPLY_UF, f, x, y);
```
ajreynol: Shouldn't this be `mkTerm(Op(APPLY_UF, f), x, y )` ?

test/unit/api/solver_black.h
```diff
@@ -911,6 +892,73 @@ void SolverBlack::testDefineFunsRec()
...
+
+  // Expecting the uninterpreted function to be one of the children
+  Term expected_children[3] = {f, x, y};
```
ajreynol: This would be `{ x, y }` I guess if the above is changed.

test/unit/api/solver_black.h
makaimann: Not anymore, no. In the previous API, it would be more like that. Part of this refactor is to treat functions just as terms -- more of HOL perspective. One nice side effect of this, is that if you're traversing a formula and rebuilding it, you don't need to special-case functions. You just call `mkTerm` with `APPLY_UF` and the children.

ajreynol: Ah, I understand now.

APPLY_UF might be a misnomer then, as `f` could be an arbitrary term whose type is a function.  Should we change this to APPLY?

src/api/cvc4cpp.h
aniemetz: I think we should, otherwise users can't test for this before calling `getIndices`.

src/api/cvc4cpp.h
aniemetz: Sorry, my comment was referencing the wrong line (even though it did also apply to `toString()`, which I would have missed otherwise ;-)

What I actually wanted to ask is what happens in the call to `getExpr()` if the Expr is null -- then we simply return the null Expr, right? This might be a problem.

aniemetz: > Does this seem right to you?
yes

src/parser/cvc/Cvc.g
```diff
@@ -1683,12 +1683,12 @@ uminusTerm[CVC4::Expr& f]
   unsigned minusCount = 0;
 }
     /* Unary minus */
-  : (MINUS_TOK { ++minusCount; })* bvBinaryOpTerm[f]
+  : (MINUS_TOK { ++minusCount; })* bvBinaryOp[f]
```
aniemetz: Can you revert the changes in this file? They don't have anything to do with your changes.

src/api/cvc4cpp.h
makaimann: Ahh that's right, we would just return a null Expr. Maybe an assertion failure would be good there as well? What do you think?

test/unit/api/solver_black.h
makaimann: Ahh that's an interesting point! It would still map to the internal kind `CVC4::Kind::APPLY_UF` though. What other kinds of functions would you expect? Can you define functions through the API?

src/parser/cvc/Cvc.g
makaimann: Oops, sure! Done with a0f690b64f5f9fd921cf3f8313e980134accda8e. I assumed it was talking about the same "OpTerm" -- my bad.

test/unit/api/solver_black.h
makaimann: Okay, so we do have `defineFunction` in the API. So that’s a good point. Under the hood, would you still use `APPLY_UF` for a defined function? My understanding was that it’s essentially a macro in CVC4, but I’m not sure where the substitution happens. Also, are there other cases besides UF and defined functions?

ajreynol: A defined function (via define-fun) I would still consider to be an uninterpreted function for uniformity.

However, when we do higher-order SMT, three cases of function applications whose heads are not uninterpreted functions that come to mind are:
(1) A lambda expression, e.g. `(APPLY (lambda ((x Int)) (+ x 1)) c)`,
(2) A term ITE, e.g. `(APPLY (ite B f g) c)`,
(3) A universal variable e.g. `(APPLY x c)` in `(forall ((x (-> Int Bool))) (APPLY x c))`.

makaimann: Ahh I see, thanks for the explanation! Changing it to APPLY makes sense to me. But maybe we should do it in a separate PR? Seems slightly out of the scope of this one to me -- because this changes the semantics of term iteration and how OpTerms are handled. This APPLY_UF confusion already exists in master. Would you recommend renaming the internal kind as well? It seems strange to have `APPLY` map to `APPLY_UF` here:
https://github.com/CVC4/CVC4/blob/cbd86eb4ed8bafc17f28244b746a376a019462f1/src/api/cvc4cpp.cpp#L90

src/api/cvc4cpp.h
makaimann: Great, did this with b39abd6ea9c01aaaad7de26421d58e3ee1734c6e. As we discussed on slack, I used a helper function to avoid calling API functions from other API functions.

src/api/cvc4cpp.h
aniemetz: Assertion failure sounds good, this is a temporary available function for the parser migration and will be internal in the future.

src/parser/cvc/Cvc.g
aniemetz: I guess this was coincidence :-)

It should in the future, but we shouldn't include it in this PR (but rather do this explicitly when migrating the parser).

test/unit/api/solver_black.h
ajreynol: Yes, I think it may be out of scope for this PR.

ajreynol:
I agree with the spirit of this PR. Having the operator as the first child makes certain traversal operations easier, as traversals can be applied to children only. Whereas in the Node structure, we have to special case for operators, see e.g.:
https://github.com/CVC4/CVC4/blob/master/src/expr/node.h#L1362
https://github.com/CVC4/CVC4/blob/master/src/expr/term_canonize.cpp#L182
https://github.com/CVC4/CVC4/blob/master/src/expr/node_algorithm.cpp#L214

The remaining things that are a bit strange to me are:
(1) The AST structure of the Node/Term is now out of sync (the first child of Term does not correspond to the first child of Node).
(2) There is a distinction of builtin operators, e.g. bit-vector extract, that are not added as children.

makaimann:
Thanks for the feedback @ajreynol!

For (2), I think this actually makes sense (to me at least). Builtin operators (indexed or not) are the operator itself. Whereas a function is a Term, so the operator is `APPLY_UF` (possibly soon that will just be `APPLY` as we discussed earlier). I suppose we could turn operators into terms as well, but then we'd basically only have one operator "`APPLY`" which is maybe a bit too abstract. I just think of indexed operators as "indexed enums" -- similar to the capability in Rust. They're no different than any other operator (e.g. PLUS), they just happen to have an index.

As for your point in (1) -- I agree this is strange. I discussed this offline with @aniemetz, @barrettcw and @mpreiner before making this PR. The idea was that the API would be cleaner, without having to change all the internals. But, maybe we should consider changing the Node structure at some point as well. I'm not sure if that's something we should prioritize now or not. I'm assuming it would be a pretty massive undertaking?

ajreynol:
Yes, it would be a massive undertaking.  I agree its lower priority. We should just be aware of this discrepancy.

HanielB:
I think it makes sense to have the API with a different design given that we consider it better. The internal data-structure could be simplified at a later point.

HanielB:
Regarding the discussion about having a single `APPLY` operator: wouldn't uniformity be better? These indexed operators are just partially applied functions.

makaimann:
That's true -- so would the semantics be that every term building call uses `APPLY`? Then we could basically drop the operator, because there's only one. Now, the first child is always the operator? Builtin operators (indexed or not) would be treated the same. So it'd look something like:
```
Term x = slv.mkConst(bitvector32, "x");
Term y = slv.mkConst(bitvector32, "y");
// this looks exactly the same as before but with an implied APPLY operator
Term xpy = slv.mkTerm(BITVECTOR_PLUS, x, y);
// this iterates over BITVECTOR_PLUS, x, y
for (Term child : xpy) {
   cout << child << endl;
}
```

Did I get that right?

HanielB:
Yeah, I think that would be nice. So in the case of extract or the ones Andy mentions for terms of functional type I think we would require first creating the term and then having it as first argument of `mkTerm`, as you had in your example with the builtin kind.

barrettcw:
I'm not enthusiastic about the idea of changing every kind to APPLY.  I like having static kinds to make it quick and easy to dispatch on kinds for common cases.

PR #3360, opened 2019-10-07
New C++ API: Add Term::getId().

aniemetz:
Plus:
+ use explicit types in NodeValue
+ add unit test for Term::isParameterized()

Note: I moved `Term::isParameterized()` down, below `isNull()`, for consistency. Nothing else changed in this function.

Labelling 'major' since I need it for the MBT.

### DIFF
--- a/src/api/cvc4cpp.h
+++ b/src/api/cvc4cpp.h
@@ -576,10 +576,25 @@ class CVC4_PUBLIC Term
   bool operator!=(const Term& t) const;
 
   /**
+   * @return the id of this term
+   */
+  uint64_t getId() const;
+
+  /**
    * @return the kind of this term
    */
   Kind getKind() const;
-  
+
+  /**
+   * @return the sort of this term
+   */
+  Sort getSort() const;
+
+  /**
+   * @return true if this Term is a null term
+   */
+  bool isNull() const;
+
   /**
    * @return true if this expression is parameterized.
    *
@@ -596,16 +611,6 @@ class CVC4_PUBLIC Term
    *   mkTerm(t.getKind(), b1, ..., bn )
    */
   bool isParameterized() const;
-  
-  /**
-   * @return the sort of this term
-   */
-  Sort getSort() const;
-
-  /**
-   * @return true if this Term is a null term
-   */
-  bool isNull() const;
 
   /**
    * Boolean negation.
--- a/src/expr/node_value.h
+++ b/src/expr/node_value.h
@@ -79,10 +79,10 @@ class NodeValue {
 
   /** Maximum reference count possible.  Used for sticky
    *  reference-counting.  Should be (1 << num_bits(d_rc)) - 1 */
-  static const uint32_t MAX_RC = (1u << NBITS_REFCOUNT) - 1;
+  static const uint32_t MAX_RC = (((uint32_t) 1) << NBITS_REFCOUNT) - 1;
 
   /** A mask for d_kind */
-  static const uint32_t kindMask = (1u << NBITS_KIND) - 1;
+  static const uint32_t kindMask = (((uint32_t) 1) << NBITS_KIND) - 1;
 
   // This header fits into 96 bits
 
--- a/src/expr/node_value.h
+++ b/src/expr/node_value.h
@@ -70,18 +70,19 @@ namespace expr {
  * This is a NodeValue.
  */
 class NodeValue {
-
-  static const unsigned NBITS_REFCOUNT = CVC4__EXPR__NODE_VALUE__NBITS__REFCOUNT;
-  static const unsigned NBITS_KIND = CVC4__EXPR__NODE_VALUE__NBITS__KIND;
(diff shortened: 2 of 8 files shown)


src/api/cvc4cpp.h
```diff
@@ -575,11 +575,26 @@ class CVC4_PUBLIC Term
    */
   bool operator!=(const Term& t) const;
 
+  /**
+   * @return the id of this term
+   */
+  uint64_t getId() const;
```
4tXJ7f: We should add a comment that this operation is not allowed on a null term.

src/expr/node_value.h
```diff
@@ -70,32 +70,33 @@ namespace expr {
...
-  uint64_t d_rc        : NBITS_REFCOUNT;
+  uint32_t d_rc : NBITS_REFCOUNT;
 
   /** Kind of the expression */
-  uint64_t d_kind      : NBITS_KIND;
+  uint32_t d_kind : NBITS_KIND;
 
   /** Number of children */
-  uint64_t d_nchildren : NBITS_NCHILDREN;
+  uint32_t d_nchildren : NBITS_NCHILDREN;
```
4tXJ7f: Nit: the changes from `uint64_t` to `uint32_t` don't really change anything, so I am not sure if we should do it.

src/expr/node_value.h
```diff
@@ -252,13 +253,16 @@ class NodeValue {
...
+
   Kind getKind() const { return dKindToKind(d_kind); }
+
   kind::MetaKind getMetaKind() const { return kind::metaKindOf(getKind()); }
-  unsigned getNumChildren() const {
-    return (getMetaKind() == kind::metakind::PARAMETERIZED)
-      ? d_nchildren - 1
-      : d_nchildren;
+
+  uint32_t getNumChildren() const
```
4tXJ7f: I am wondering if we should be more conservative here and return `uint64_t` or `size_t`?

src/expr/node_value.h
aniemetz: We only use 26 bits for the number of children and I suppose we will never support more than 32. In any case, we should be consistent. It was `uint64_t` here and `unsigned` as a return value for `getNumChildren()`.

src/expr/node_value.h
aniemetz: See above

src/api/cvc4cpp.h
aniemetz: We currently don't have this for any of the member functions where we guard against a null Term (or Sort, for that matter). What is your preferred way to document this?

src/expr/node_value.h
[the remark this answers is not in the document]
makaimann: Saw you addressed it on @4tXJ7f's comment

4tXJ7f:
We discussed this offline and came to the conclusion that `uint32_t` is fine.

test/unit/api/term_black.h
```diff
@@ -57,6 +59,14 @@ void TermBlack::testEq()
   TS_ASSERT(x != z);
 }
 
+void TermBlack::testGetId()
+{
+  Term n;
+  TS_ASSERT_THROWS(n.getId(), CVC4ApiException&);
+  Term x = d_solver.mkVar(d_solver.getIntegerSort(), "x");
+  TS_ASSERT_THROWS_NOTHING(x.getId());
+}
```
makaimann:
```suggestion
Term y = x;
TS_ASSERT_EQUALS(x.getId(), y.getId());

Term z = d_solver.mkVar(d_solver.getIntegerSort(), "z");
TS_ASSERT_DIFFERS(x.getId(), z.getId());
}
```

src/api/cvc4cpp.h
aniemetz: We decided offline that this is out of scope for this PR. We postpone this since documentation about error cases is in general missing for API classes and their functions:
https://github.com/CVC4/CVC4/issues/3361

test/unit/api/term_black.h
aniemetz: Good catch!

test/unit/api/term_black.h
```diff
@@ -65,6 +65,11 @@ void TermBlack::testGetId()
   TS_ASSERT_THROWS(n.getId(), CVC4ApiException&);
   Term x = d_solver.mkVar(d_solver.getIntegerSort(), "x");
   TS_ASSERT_THROWS_NOTHING(x.getId());
+Term y = x;
+TS_ASSERT_EQUALS(x.getId(), y.getId());
+
+Term z = d_solver.mkVar(d_solver.getIntegerSort(), "z");
+TS_ASSERT_DIFFERS(x.getId(), z.getId());
```
makaimann: Looks like the formatting wasn't right -- that's my bad. Otherwise it looks good to me!

aniemetz: done

PR #3369, opened 2019-10-08
Fix non-termination in datatype type enumerator

### LINKED ISSUE #3316: crash on problem
When the following smt2 input
```
(set-info :smt-lib-version 2.5)
(set-logic ALL)
(declare-datatypes () ((a0(c0$0)(c0$1(c0$1$0 String)(c0$1$1 Int))(c0$2(c0$2$0 Bool)(c0$2$1 Int)(c0$2$2 String)))
                       (a1(c1$0)(c1$1)(c1$2))
                       (a2(c2$0(c2$0$0 Int)(c2$0$1 a0)(c2$0$2 String)(c2$0$3 a3)(c2$0$4 String)(c2$0$5 Bool)))
                       (a3(c3$0(c3$0$0 a7)(c3$0$1 a1)(c3$0$2 a5)(c3$0$3 a6)(c3$0$4 Int)(c3$0$5 Bool)(c3$0$6 Bool))(c3$1(c3$1$0 String)))
                       (a4(c4$0(c4$0$0 String)(c4$0$1 a2)(c4$0$2 String))(c4$1(c4$1$0 a0)(c4$1$1 a4)(c4$1$2 a4)(c4$1$3 a7))(c4$2))
                       (a5(c5$0(c5$0$0 a2))(c5$1)(c5$2)(c5$3(c5$3$0 a0))(c5$4)(c5$5(c5$5$0 a4)(c5$5$1 Int))(c5$6))
                       (a6(c6$0(c6$0$0 a7)(c6$0$1 a7)(c6$0$2 String))(c6$1)(c6$2)(c6$3)(c6$4)(c6$5)(c6$6))
                       (a7(c7$0(c7$0$0 a2)(c7$0$1 Int))(c7$1(c7$1$0 a4)(c7$1$1 Int)(c7$1$2 Bool)))))
(define-funs-rec ((f3((v4 Bool))a7)
                  (f2()a6)
                  (f1((v1 a3)(v2 a1)(v3 Bool))String)
                  (f0((v0 a6))a4))
                  ((c7$0 (c2$0 0 c0$0 "" (c3$1 "") "" true) 0) 
                  c6$2 
                  "" 
                  c4$2))
(check-sat)
``` 
is offered to cvc4 as follows
```
cvc4 --fmf-fun-rlv <filename>
```  
it crashes on windows!

without the `--fmf-fun-rlv` flag it just returns unknown.

My version of cvc4 is
```
> cvc4 --version
This is CVC4 version 1.8-prerelease
compiled with GCC version 5.3.1 20160211
on Sep 16 2019 11:19:28
```

ajreynol:
This PR fixes a nasty corner case of datatype enumeration.  Previously, the datatype enumerator chose a "zero constructor", i.e. the constructor of the first term in its enumeration.  This constructor was chosen based on constructing a default term via `mkGroundTerm` (which traverses the datatype to find a finite term) and taking the topmost constructor. However, this led to infinite loops in some rare cases of datatypes like:

```
C -> cons_cd(D) | cons_ce(E)
D -> cons_dc(C) | cons_de(E)
E -> zero
```

If the default term chosen for `C` is `cons_cd(cons_de(zero))` and the default term chosen for `D` is `cons_dc(cons_ce(zero))`, then the type enumerator for C was non-terminating: we would try to construct a value `cons_cd( cons_dc( cons_cd( cons_dc( .... )...)`.

The solution is to use the entire resulting term of (a variant of) `mkGroundTerm` as the first enumerated term and enumerate as normal, then skipping that term when it is encountered.

Since `mkGroundTerm` returns a term and not necessarily a value, this PR extends the type/datatype interface with a `mkGroundValue` function.

This fixes #3316.


ajreynol:
I require this to be merged first to continue towards refactoring the datatype for C++ API / parser migration.

src/theory/datatypes/type_enumerator.cpp
```diff
@@ -166,59 +192,125 @@ Node DatatypesEnumerator::getTermEnum( TypeNode tn, unsigned i ){
...
+   {
+     // find the "zero" term via mkGroundTerm
      Debug("dt-enum-debug") << "make ground term..." << std::endl;
-     Node t = d_type.mkGroundTerm();
-     Debug("dt-enum-debug") << "done : " << t << std::endl;
-     Assert(t.getKind() == kind::APPLY_CONSTRUCTOR);
-     // start with the constructor for which a ground term is constructed
-     d_zeroCtor = datatypes::utils::indexOf(t.getOperator());
+     // Start with the ground term constructed via mkGroundTerm, which does
+     // a traversal over the structure of the datatype to find a finite term.
```
HanielB: The comment refers to `mkGroundTerm` but you are calling `mkGroundValue`

src/theory/datatypes/type_enumerator.cpp
```diff
@@ -166,59 +192,125 @@ Node DatatypesEnumerator::getTermEnum( TypeNode tn, unsigned i ){
...
+     d_zeroTerm = Node::fromExpr(d_datatype.mkGroundValue(d_type.toType()));
```
HanielB: From what I could gather `mkGroundValue` itself seems to depend on the type enumerator... I assume that when you are not well founded this will be captured downstream and this will all break, so this call could throw an exception? I think it'd be good to say it here.

src/theory/datatypes/type_enumerator.cpp
```diff
@@ -166,59 +192,125 @@ Node DatatypesEnumerator::getTermEnum( TypeNode tn, unsigned i ){
...
      }
    }
    d_size_limit = 0;
-   //set up initial conditions (should always succeed)
-   ++*this; //increment( d_ctor );
+   if (!d_zeroTermActive)
+   {
+     // set up initial conditions (should always succeed)
+     ++*this;
+   }
```
HanielB: What is happening here?

src/theory/datatypes/type_enumerator.cpp
```diff
@@ -166,59 +192,125 @@ Node DatatypesEnumerator::getTermEnum( TypeNode tn, unsigned i ){
...
+         else
+         {
+           return *this;
+         }
+       }
+     }
+     // Here, we need to step from the current constructor to the next one
 
+     // Find the next constructor (only complicated by the notion of the "zero"
+     // constructor
```
HanielB: Could you clarify?

src/theory/datatypes/type_enumerator.cpp
ajreynol: I've added more clarification, the argument for non-termination here is independent of this change.

src/theory/datatypes/type_enumerator.cpp
ajreynol: The process for constructing the next term in the enumeration is being called. This is necessary to do since setting up the first term is analogous to setting up each subsequent term.

src/theory/datatypes/type_enumerator.cpp
ajreynol: That comment was outdated, I removed.

PR #3372, opened 2019-10-08
Ackermann Task2: Support ackermannization on uninterpreted sorts in BV

Ying1123:
Support ackermannization on uninterpreted sorts in BV. For uninterpreted sorts, we create a new Bit Vector sort to replace it. We guarantee that the new Bit Vector will have a fresh size, so that different sorts will be translated into different Bit Vector Sorts. Besides, if for an uninterpreted sort S, the number of variables within sort S is n, the replacing Bit Vector will have size at least (log n)+1.


src/preprocessing/passes/ackermann.cpp
```diff
@@ -187,9 +187,117 @@ void collectFunctionsAndLemmas(FunctionToArgsMap& fun_to_args,
...
+  for (Node& a : assertions->ref())
+  {
+    if (seen.find(a) == seen.end())
+    {
+      to_process.push_back(a);
+      seen.insert(a);
+    }
+  }
+  TNode term;
+  for (unsigned i = 0; i < to_process.size(); ++i)
```
4tXJ7f: Personally, I find it confusing when `for` loops change the range that they iterate over :) I find it easier to read when we have a `while (!toProcess.empty())` loop and then remove the last element of `toProcess` in each iteration (should also be a bit more memory efficient than keeping all the nodes in `toProcess`).

src/preprocessing/passes/ackermann.cpp
```diff
@@ -187,9 +187,117 @@ void collectFunctionsAndLemmas(FunctionToArgsMap& fun_to_args,
...
+/* Update the statistics for each uninterpreted sort */
+void updateUSortsCardinality(USortToBVSizeMap& usort_cardinality, TNode term)
+{
+  TypeNode type = term.getType();
+  if (type.isSort())
+  {
+    if (usort_cardinality.find(type) == usort_cardinality.end())
+    {
+      usort_cardinality.insert(make_pair(type, make_pair(0, 0)));
+    }
```
4tXJ7f: The above `if` statement is not necessary: When you do `usort_cardinality[type]` and the key `type` does not exist, C++ will create that element for you, initialized with the default value for the type (which happens to be a pair of two zeros).

src/preprocessing/passes/ackermann.h
```diff
@@ -38,6 +40,8 @@ namespace passes {
 using TNodeSet = std::unordered_set<TNode, TNodeHashFunction>;
 using FunctionToArgsMap =
     std::unordered_map<TNode, TNodeSet, TNodeHashFunction>;
+using USortToBVSizeMap = std::
+    unordered_map<TypeNode, pair<unsigned, unsigned>, TypeNode::HashFunction>;
```
4tXJ7f:
```suggestion
    unordered_map<TypeNode, pair<size_t, size_t>, TypeNode::HashFunction>;
```

src/preprocessing/passes/ackermann.cpp
```diff
@@ -216,6 +324,23 @@ PreprocessingPassResult Ackermann::applyInternal(
         i, d_funcToSkolem.apply((*assertionsToPreprocess)[i]));
   }
 
+  /* the current version only support BV for removing uninterpreted sorts */
+  if (d_logic.isTheoryEnabled(theory::THEORY_BV))
```
4tXJ7f: We should consider just always enabling `THEORY_BV` when Ackermannization is used. @yoni206 what do you think?

src/preprocessing/passes/ackermann.cpp
```diff
@@ -187,9 +187,117 @@ void collectFunctionsAndLemmas(FunctionToArgsMap& fun_to_args,
...
+    if (type.isSort())
+    {
+      unsigned size = usort_cardinality[type].second;
+      if (size == 0)
+      {
+        size = log2(usort_cardinality[type].first) + 1;
+        while (used.find(size) != used.end())
+        {
+          ++size;
+        }
```
4tXJ7f: I am not quite sure if it is really necessary to find unique sizes for all the different sorts. If the input is well-typed, we should never be mixing different sort. Maybe I am missing something?

cc @yoni206

src/preprocessing/passes/ackermann.cpp
```diff
@@ -187,9 +187,117 @@ void collectFunctionsAndLemmas(FunctionToArgsMap& fun_to_args,
...
+  }
+
+  for (TNode a : to_process)
+  {
+    TypeNode type = a.getType();
+    if (type.isBitVector())
+    {
+      used.insert(type.getBitVectorSize());
+    }
+  }
```
4tXJ7f: Couldn't we combine this loop with the loop above? It may also not be necessary, depending on whether unique bit-vector sizes are really needed or not.

src/preprocessing/passes/ackermann.cpp
```diff
@@ -187,9 +187,117 @@ void collectFunctionsAndLemmas(FunctionToArgsMap& fun_to_args,
...
+      to_process.push_back(a);
+      seen.insert(a);
+    }
+  }
+  TNode term;
+  for (unsigned i = 0; i < to_process.size(); ++i)
+  {
+    term = to_process[i];
+    AlwaysAssert(term.getKind() != kind::STORE,
+                 "Cannot use ackermannization on QF_ABV formula with stores");
```
4tXJ7f: I think that we already check this somewhere else, right? If yes, this may not be necessary.

src/preprocessing/passes/ackermann.cpp
```diff
@@ -187,9 +187,117 @@ void collectFunctionsAndLemmas(FunctionToArgsMap& fun_to_args,
...
+    {
+      to_process.push_back(a);
+      seen.insert(a);
+    }
+  }
+  TNode term;
+  for (unsigned i = 0; i < to_process.size(); ++i)
+  {
+    term = to_process[i];
+    AlwaysAssert(term.getKind() != kind::STORE,
```
mpreiner: If we keep this check, use `CVC4_CHECK(term.getKind() != kind::STORE) << "..."` instead of `AlwaysAssert`.

src/preprocessing/passes/ackermann.cpp
yoni206: If your input two (uninterpreted) constants `s1` and `s2` of (corresponding uninterpreted) sorts `S1` and `S2`, then `s1` and `s2` are implicitly distinct.
Shouldn't `S1` and `S2` be mapped to different BV sorts? Otherwise, you might end up constructing a model in which `s1` and `s2` are equal, as this is only an implicit constraint, not an explicit one. Or is it I that is missing something?

src/preprocessing/passes/ackermann.cpp
yoni206: Good question.
First, I think BV is only needed if Ackermannization is used *and* uninterpreted sorts are used.
Second, always translating uninterpreted sorts to BV sorts is a possibility. Another possibility, e.g., for QF_UFLIA, is to translate uninterpreted sorts to `Int` and enforce `distinct` constraints on all variables of that sort. The former has less constraints but involves theory combination, while the latter is the other way around. Do you think there is a clear winner? do you think it is interesting to try both?

src/preprocessing/passes/ackermann.cpp
yoni206: For readability, I would either include the `if` statement or explain this default `c++` behavior inline.

src/preprocessing/passes/ackermann.cpp
```diff
@@ -187,9 +187,117 @@ void collectFunctionsAndLemmas(FunctionToArgsMap& fun_to_args,
...
+        {
+          ++size;
+        }
+        usort_cardinality[type].second = size;
+        used.insert(size);
+      }
+      Node skolem = nm->mkSkolem("BVSKOLEM$$",
+                                 nm->mkBitVectorType(size),
+                                 "a variable created by the ackermannization "
+                                 "preprocessing pass for theory BV");
```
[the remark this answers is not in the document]
yoni206: Doesn't it represent an uninterpreted constant (and not a whole sort)?

src/preprocessing/passes/ackermann.h
```diff
@@ -64,6 +68,14 @@ class Ackermann : public PreprocessingPass
   /* Map each function term to the new Skolem variable created by
    * ackermannization */
   theory::SubstitutionMap d_funcToSkolem;
+  /* Map each uninterpreted sort to the new Skolem variable created by
+   * ackermannization */
+  theory::SubstitutionMap d_sortsToSkolem;
```
yoni206: Do we map each sort to a variable? Or each variable of a sort to a BV variable?

src/preprocessing/passes/ackermann.cpp
4tXJ7f: I haven't thought very deeply about model generation but we would anyway want to map the bit-vectors back to representatives of the corresponding uninterpreted sorts (note our current models doesn't actually include the representatives for uninterpreted sorts, see #2851). To reconstruct the representatives, you just have to look at the bit-vector values of all terms of a given uninterpreted sort and assign a unique representative to each bit-vector value that you see. `s1` and `s2` cannot be equal in that scenario because you are assigning different representatives based on the sort of the original (pre-ackermannization) sort and a term has a unique sort.

yoni206: How can you be sure that `s1` and `s2` will be assigned different representatives? Say `s1` and `s2` are the only variables in the formula. If we map both `S1` and `S2` to `BV[1]`, then `s1` and `s2` will be assigned to variables of the same BV sort. Why are they distinct?

src/preprocessing/passes/ackermann.cpp
yoni206: > Should we output an error if uninterpreted sorts are used but not eliminated by Ackermannization or should we just silently keep them?

Well... If the user explicitly turned on Ackermannization, then I would go for exception. But if Ackermannization was turned on by `bitblast=eager`, then I think we should silently keep them. Does this make sense?

> I think that it is sufficient to just turn the uninterpreted sorts into Ints, actually (since the cardinality of an uninterpreted sort is infinite).

In this example, there is a single uninterpreted sort `U`. But if you had `U1` and `U2`, shouldn't you assert that `u1 != u2` for every `u1 in U1` and `u2 in U2`?

src/preprocessing/passes/ackermann.h
Ying1123: Oh, we map each variable of a sort to a BV variable. Maybe I need to reconsider for a better name.

src/preprocessing/passes/ackermann.cpp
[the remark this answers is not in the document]
Ying1123: Can I say uninterpreted sort term?

src/preprocessing/passes/ackermann.cpp
Ying1123: Using non-unique size may not change the satisfiability. But for underlying theory, it could be confusing if we replace two different sorts with a same new sort. First, I think the model generation is possible to give s1 and s2 same value. It maybe safe for satisfiability but could be misleading. And Second, consider for further development, converting S1 and S2 to same sort may mislead the optimization algorithm. I mean current method may be safe, but we don't know in the future if there is a smart mechanism that deduce a rule s1=s2 which should not be true. Carefully design may could avoid mistakes, but still, we will cause the system be vulnerable.

Ying1123: It could also cause a problem that, the number of sorts and the cardinality for each sort will be ambiguous.

src/preprocessing/passes/ackermann.cpp
4tXJ7f: > But if Ackermannization was turned on by `bitblast=eager`, then I think we should silently keep them.

If that's the case, then we implicitly have BV because that's what the eager bit-blaster operates on :)

> In this example, there is a single uninterpreted sort U. But if you had U1 and U2, shouldn't you assert that u1 != u2 for every u1 in U1 and u2 in U2?

I guess that's the same argument that we have below about the different bit-widths. Assuming that you are right about the argument below, you'd have to divide `Int`s into ranges belonging to different sorts, I think.

src/preprocessing/passes/ackermann.cpp
4tXJ7f: Alright, so I think that we all agree that it doesn't change the satisfiability. If the satisfiablity is preserved, then any optimizations that we do on the transformed problem should be safe to do, otherwise it is not equi-satisfiable. What remains is the question whether we need the unique bit-width to distinguish between different sorts. I am arguing that you can maintain a map of transformed terms to the original sort and maintain the mapping that way. I am quite certain that such a map is sufficient to identify the sort of the original term (i.e. there shouldn't be any collisions between different terms in that map) or at least I can't think of a case where it wouldn't be.

So for @yoni206's example:
```
(declare-sort S1 0)
(declare-sort S2 0)
(declare-const s11 S1)
(declare-const s12 S1)
(declare-const s21 S2)
(declare-const s22 S2)
(assert (= s11 s12))
(assert (not (= s21 s22)))
```

This would get turned into:
```
(declare-const bv11 (_ BitVec 1))
(declare-const bv12 (_ BitVec 1))
(declare-const bv21 (_ BitVec 1))
(declare-const bv22 (_ BitVec 1))
(assert (= bv11 bv12))
(assert (not (= bv21 bv22)))
```

And we would have a map:
```
bv11 --> S1
bv12 --> S1
bv21 --> S2
bv22 --> S2
```

Now let's say that we get the model:
```
bv11 = 0
bv12 = 0
bv21 = 0
bv22 = 1
```

We can use the map above to decide that `bv11` and `bv21` are representatives of different sorts, so we could get something like:
```
bv11 = r11
bv12 = r11
bv21 = r21
bv22 = r22
```

Can you think of an example where this wouldn't work?

4tXJ7f: To give some additional intuition: We always replace terms of uninterpreted sorts with skolems (https://github.com/CVC4/CVC4/pull/3372/files#diff-4c3f5767ca5f0a1b1951e86d68f8afaeR221), skolems are unique w.r.t. the original sort that they belong to, so the map above should not have any collisions.

src/preprocessing/passes/ackermann.cpp
yoni206: Doing `distinct` between integers that come from different sorts is a more general solution than dividing to ranges I think.

src/preprocessing/passes/ackermann.cpp
yoni206: I think you are right and this is possible. Intuitively, it is like using different "copies" of `BV[1]` for `S1` and `S2`, right? Without quantifiers, I also can't think of an example where this wouldn't work (but that doesn't mean much). @Ying1123 , I think this would complicate the correctness proof, but just a little bit.

4tXJ7f: Yes, exactly. Instead of mapping the bit-width to the original sorts, we would be mapping the skolems themselves to the original sorts. I think the only additional piece that you need is to show that you can uniquely identify the original sort of each term in the formula and of course the formula needs to be well-typed (well-sorted? Not sure what's the correct terminology here).

ajreynol:
I'm not sure how you are dealing with quantifiers in this PR.  You should be careful to either disallow this preprocessing pass when quantifiers are enabled or ensure that the terms you consider don't have universal variables (of kind `BOUND_VARIABLE`).

yoni206:
I agree with Andy, this is just for qf.

src/preprocessing/passes/ackermann.cpp
```diff
@@ -187,9 +188,124 @@ void collectFunctionsAndLemmas(FunctionToArgsMap& fun_to_args,
...
+        res.push_back(a);
+        seen.insert(a);
+      }
+    }
+    ++i;
+  }
+  return res;
+}
+
+/* This is the top level of converting uninterpreted sorts to bit vectors.
```
4tXJ7f:
```suggestion
/* This is the top level of converting uninterpreted sorts to bit-vectors.
```

src/preprocessing/passes/ackermann.cpp
```diff
@@ -187,9 +188,124 @@ void collectFunctionsAndLemmas(FunctionToArgsMap& fun_to_args,
...
+{
+  if (term.getType().isSort() && !(term.getKind() == kind::BOUND_VARIABLE))
+    return true;
+  return false;
+}
+
+/* Given the lowest capacity requirements for each uninterpreted sorts, assign
+ * a sufficient bit-vector size. Get the converting map */
+void collectUSortsToBV(USortToBVSizeMap& usortCardinality,
+                       vector<TNode>& vec,
```
4tXJ7f:
```suggestion
                       const vector<TNode>& vec,
```

We are not modifying `vec` (as far as I can tell), so this should be a `const` reference. Also, can we choose a more descriptive variable name? If that's the case, I would also move `vec` to be the first argument (to have a separation between arguments that are just inputs and arguments that are modified), although that's subjective.

src/preprocessing/passes/ackermann.cpp
```diff
@@ -187,9 +188,124 @@ void collectFunctionsAndLemmas(FunctionToArgsMap& fun_to_args,
 
 /* -------------------------------------------------------------------------- */
 
+bool needsReplace(TNode term)
+{
+  if (term.getType().isSort() && !(term.getKind() == kind::BOUND_VARIABLE))
```
4tXJ7f: I think that we would be better off to just not allow this preprocessing pass to run when quantifiers are enabled (because we remove UF from the logic but we may not eliminate all UFs if we have bound variables). Also, I am not sure if we'd need to do more work to make the pass work with quantifiers in general. What do you think @yoni206 ?

src/preprocessing/passes/ackermann.cpp
```diff
@@ -187,9 +188,124 @@ void collectFunctionsAndLemmas(FunctionToArgsMap& fun_to_args,
...
+      if (size == 0)
+      {
+        size = log2(usortCardinality[type].first) + 1;
+        usortCardinality[type].second = size;
+      }
+      Node skolem = nm->mkSkolem(
+          "BVSKOLEM$$",
+          nm->mkBitVectorType(size),
+          "a variable created by the ackermannization "
+          "preprocessing pass, representing term with uninterpreted sorts.");
```
4tXJ7f:
```suggestion
          "preprocessing pass, representing a term with uninterpreted sort.");
```

src/preprocessing/passes/ackermann.cpp
```diff
@@ -187,9 +188,124 @@ void collectFunctionsAndLemmas(FunctionToArgsMap& fun_to_args,
...
+  std::vector<TNode> toProcess = getListOfTerms(assertions);
+  bool hasUninterpretedSorts = false;
+  for (TNode term : toProcess)
+  {
+    if (needsReplace(term))
+    {
+      hasUninterpretedSorts = true;
+      TypeNode type = term.getType();
+      /* Update the statistics for each uninterpreted sort */
+      // For non-existing key, C++ will create a new element for it, which has
```
4tXJ7f:
```suggestion
      // For non-existing keys, C++ will create a new element for it, which has
```

src/preprocessing/passes/ackermann.cpp
```diff
@@ -187,9 +188,124 @@ void collectFunctionsAndLemmas(FunctionToArgsMap& fun_to_args,
...
+{
+  std::vector<TNode> toProcess = getListOfTerms(assertions);
+  bool hasUninterpretedSorts = false;
+  for (TNode term : toProcess)
+  {
+    if (needsReplace(term))
+    {
+      hasUninterpretedSorts = true;
+      TypeNode type = term.getType();
+      /* Update the statistics for each uninterpreted sort */
```
4tXJ7f: Make the style of the comment consistent with the comment below (`/* ... */` vs. `// ...`).

src/preprocessing/passes/ackermann.cpp
```diff
@@ -187,9 +188,124 @@ void collectFunctionsAndLemmas(FunctionToArgsMap& fun_to_args,
...
+ * different terms for each uninterpreted sort. Then for each sort, we will
+ * assign a new bit-vector type with a sufficient size.
+ * The size is calculated to have enough capacity, that can accommodate the
+ * variables occured in the original formula. */
+void usortsToBitVectors(const LogicInfo& d_logic,
+                        AssertionPipeline* assertions,
+                        USortToBVSizeMap& usortCardinality,
+                        SubstitutionMap& sortsToSkolem)
+{
+  std::vector<TNode> toProcess = getListOfTerms(assertions);
```
4tXJ7f: As far as I can tell, we later only care about terms in `toProcess` that `needsReplace()`. If that is the case, we shouldn't add them to `toProcess` in the first place (they can already be filtered out in `getListOfTerms()`. This will save time and memory.

src/preprocessing/passes/ackermann.cpp
yoni206: Ackermannization assumes no quantifiers, so yes I support not allowing the pass in the presence of quantifiers.

src/preprocessing/passes/ackermann.cpp
yoni206: Looking again at the PR, I think we should *not* throw an exception if BV is not enabled, but just leave the uninterpreted sorts.
This PP is called `ackermannization`, not `uninterpreted-sort-elimination`. If a user has a QF_UFLIA problem with uninterpreted sorts, she should be able to perform Ackermannization in order to eliminate function symbols, leaving uninterpreted sorts intact.
This makes me think that maybe uninterpreted-sort-elimination should be a separate preprocessing pass @4tXJ7f what do you think about that?

Either way, I think we should eliminate uninterpreted sorts only if BV is enabled. Otherwise, leave them in the formula.

src/preprocessing/passes/ackermann.h
```diff
@@ -64,6 +66,15 @@ class Ackermann : public PreprocessingPass
   /* Map each function term to the new Skolem variable created by
    * ackermannization */
   theory::SubstitutionMap d_funcToSkolem;
+  /* Map each uninterpreted sort term to the new Skolem variable created by
+   * ackermannization */
+  theory::SubstitutionMap d_sortsToSkolem;
+  /* Map each uninterpreted sort term to a pair of integers.
+   * The first value is the lowest capacity that the targeting BV should have
+   * The second value is the size of the BV to which the sort will actually be
```
yoni206: https://github.com/CVC4/CVC4/pull/3372#discussion_r334044329
Didn't we agree that the first number is enough here? That is, we can always assign bit-vectors of that size, without needing to increasing it for each distinct sort?
If this is the case, we don't need `second` here.

src/preprocessing/passes/ackermann.h
```diff
@@ -64,6 +66,15 @@ class Ackermann : public PreprocessingPass
   /* Map each function term to the new Skolem variable created by
```
yoni206:
```suggestion
  /* Map each function-application term to the new Skolem variable created by
```

src/preprocessing/passes/ackermann.cpp
```diff
@@ -187,9 +188,124 @@ void collectFunctionsAndLemmas(FunctionToArgsMap& fun_to_args,
...
+          "a variable created by the ackermannization "
+          "preprocessing pass, representing a term with uninterpreted sort.");
+      sortsToSkolem.addSubstitution(term, skolem);
+    }
+  }
+}
+
+/* This function returns the list of terms in formula represented by assertions.
+ * We use a BFS to get all terms without duplications. */
+std::vector<TNode> getListOfTerms(AssertionPipeline* assertions)
```
yoni206: Maybe you can use `getSymbols` from `node_algorithm.h` instead of this function?
https://github.com/CVC4/CVC4/blob/cbd86eb4ed8bafc17f28244b746a376a019462f1/src/expr/node_algorithm.h#L99

src/preprocessing/passes/ackermann.cpp
```diff
@@ -187,9 +188,124 @@ void collectFunctionsAndLemmas(FunctionToArgsMap& fun_to_args,
...
+  }
+  return res;
+}
+
+/* This is the top level of converting uninterpreted sorts to bit-vectors.
+ * We use BFS to get all terms without duplications, and count the number of
+ * different terms for each uninterpreted sort. Then for each sort, we will
+ * assign a new bit-vector type with a sufficient size.
+ * The size is calculated to have enough capacity, that can accommodate the
+ * variables occured in the original formula. */
```
yoni206: When we eliminate uninterpreted sorts, all terms of uninterpreted sorts are actually variables, right? If so, this should be clarified in the comments.

src/preprocessing/passes/ackermann.cpp
```diff
@@ -187,9 +188,124 @@ void collectFunctionsAndLemmas(FunctionToArgsMap& fun_to_args,
...
+}
+
+/* -------------------------------------------------------------------------- */
+
 Ackermann::Ackermann(PreprocessingPassContext* preprocContext)
     : PreprocessingPass(preprocContext, "ackermann"),
-      d_funcToSkolem(preprocContext->getUserContext())
+      d_funcToSkolem(preprocContext->getUserContext()),
+      d_sortsToSkolem(preprocContext->getUserContext()),
+      d_logic(preprocContext->getLogicInfo())
```
yoni206: Shouldn't `d_usortCardinality ` be initialized here as well?

src/preprocessing/passes/ackermann.cpp
```diff
@@ -187,9 +188,124 @@ void collectFunctionsAndLemmas(FunctionToArgsMap& fun_to_args,
...
+      // the value initialized with a pair of two zeros.
+      usortCardinality[type].first = usortCardinality[type].first + 1;
+    }
+  }
+
+  if (hasUninterpretedSorts)
+  {
+    /* the current version only supports BV for removing uninterpreted sorts */
+    AlwaysAssert(d_logic.isTheoryEnabled(theory::THEORY_BV))
+        << "Cannot use Ackermannization on formula with uninterpreted "
```
yoni206: Shouldn't this change according to @mpreiner 's #3366 ?

Ying1123: This change is because of been aligned with that change #3366. What's your concern here? Was I do in a wrong way?

mpreiner: Looks perfectly fine ;-)

yoni206: my bad, sorry

src/preprocessing/passes/ackermann.h
Ying1123: I was thinking of including that information in the class, because I don't want to assume the user knows the algorithm of assigning lengths. Do you still think we should remove it?

src/preprocessing/passes/ackermann.cpp
Ying1123: Thanks for pointing. But getSymbols is get symbols from a term, while here getList is getting symbols for a formula. Seems they are different?

Ying1123: Maybe I can use getVariables() for each assertion instead?

src/preprocessing/passes/ackermann.cpp
Ying1123: We are not expecting initialization of d_usortCardinality from input and preprocContext. It should be just empty. Should we initialize every member data structure, so as d_funcToArgs?

src/preprocessing/passes/ackermann.h
yoni206: If you remove the second value and only include the first value, why does this require anything from the users? Don't you compute `first` yourself?

src/preprocessing/passes/ackermann.cpp
yoni206: https://github.com/CVC4/CVC4/blob/cbd86eb4ed8bafc17f28244b746a376a019462f1/src/expr/node_algorithm.cpp#L342

I think that `getSymbols` will work for every node. 
But I guess you can decide which one to use. Please make sure to look at their code (not only signature and comments) and also to make sure that you are not getting bound variables (there are not supposed to be any in your formulas).

src/preprocessing/passes/ackermann.h
Ying1123: Ying:"The first value is the number of occurrences, the second value is the length of bitvector we assigned. So if the first value is 10, the second value is 4. I was trying to say that, if we only include the first value, when other class want to call the class of ackermannization, it doesn’t know what length we assigned. "

Yoni:"Hmmm. Well… if you save pairs, but the value of the second is a function of the first, then it could be error-prone to save both, because you are counting on the fact that they have this relation, and maybe someplace you forget to update it.
On the other hand, if you really need to access both of them a lot of times, then maybe it makes sense to save both."

Ying:"It is not supposed to be an explicit function, considering for future development.
If we throw one, the information is lost forever. If we really want to keep only one, I would suggest the second one"

Yoni:"My only concern is that a future developer sees that there is a connection between first and second, relies on this in the code, but doesn’t bother to maintain this connection."

@4tXJ7f We want to hear your opinion about this issue. Should we just keep the first number, or both, or just the second?

src/preprocessing/passes/ackermann.cpp
4tXJ7f: No, it is fine to leave out members for which the default constructor should be called.

src/preprocessing/passes/ackermann.h
4tXJ7f: I tend to agree with @yoni206 here. Since the second number is a simple function of the first, it would be better to just define a function that looks up the value in the map and transforms it to a bit-width. If we ever need a more complicated mapping, we can still re-introduce the second value. In general, I would advise not to over-engineer for future development if it adds complexity (this is just a rule of thumb of course and your mileage may vary :) ).

yoni206: "If we throw one, the information is lost forever."
Maybe a way to keep this information is to have it in the documentation of the function that @4tXJ7f suggests to add.

src/preprocessing/passes/ackermann.cpp
```diff
@@ -187,9 +188,110 @@ void collectFunctionsAndLemmas(FunctionToArgsMap& fun_to_args,
 
 /* -------------------------------------------------------------------------- */
 
+/* Given a minimum capacity for an uninterpreted sort, return the size of the
+ * new BV type */
+size_t getBVSkolemSize(size_t capacity) { return (size_t)log2(capacity) + 1; }
+
+/* Given the lowest capacity requirements for each uninterpreted sorts, assign
```
4tXJ7f:
```suggestion
/* Given the lowest capacity requirements for each uninterpreted sort, assign
```

src/preprocessing/passes/ackermann.cpp
```diff
@@ -187,9 +188,110 @@ void collectFunctionsAndLemmas(FunctionToArgsMap& fun_to_args,
...
+    Node skolem = nm->mkSkolem(
+        "BVSKOLEM$$",
+        nm->mkBitVectorType(size),
+        "a variable created by the ackermannization "
+        "preprocessing pass, representing a term with uninterpreted sort.");
+    sortsToSkolem.addSubstitution(term, skolem);
+  }
+}
+
+/* This function returns the list of terms with uninterpreted sort in formula
```
4tXJ7f:
```suggestion
/* This function returns the list of terms with uninterpreted sort in the formula
```

src/preprocessing/passes/ackermann.cpp
```diff
@@ -187,9 +188,110 @@ void collectFunctionsAndLemmas(FunctionToArgsMap& fun_to_args,
...
+ * represented by assertions. We use a BFS to get all terms without
+ * duplications. */
```
4tXJ7f: Remove comment about BFS, since we don't care about the implementation of `expr::getVariables()` here.

src/preprocessing/passes/ackermann.cpp
```diff
@@ -187,9 +188,110 @@ void collectFunctionsAndLemmas(FunctionToArgsMap& fun_to_args,
...
+        res.push_back(var);
+      }
+    }
+  }
+
+  return res;
+}
+
+/* This is the top level of converting uninterpreted sorts to bit-vectors.
+ * We use BFS to get all variables without duplications, and count the number of
```
4tXJ7f: Remove comment about BFS, since we don't care about the implementation of `expr::getVariables()` here.

src/preprocessing/passes/ackermann.cpp
```diff
@@ -187,9 +188,110 @@ void collectFunctionsAndLemmas(FunctionToArgsMap& fun_to_args,
...
+  for (Node& assertion : assertions->ref())
+  {
+    std::unordered_set<TNode, TNodeHashFunction> vars;
+    expr::getVariables(assertion, vars);
+
+    for (const TNode& var : vars)
+    {
+      if (var.getType().isSort() && seen.find(var) == seen.end())
+      {
+        seen.insert(var);
```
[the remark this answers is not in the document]
Ying1123: vars is a set for single assertion, while seen is over all assertions.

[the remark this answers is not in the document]
4tXJ7f: Ah yes, that is a good point. How about just returning the set instead of the vector? We just need a collection that we can iterate over, so there is no need to compute two representations. We can also reuse the set that we pass to `expr::getVariables()` since that function does not clear the set, so this whole method becomes something like:

```
  TNodeSet vars;
  for (const Node& assertion : assertions->ref())
  {
    expr::getVariables(assertion, vars);
  }
  return vars;
```

src/preprocessing/passes/ackermann.h
```diff
@@ -61,9 +65,18 @@ class Ackermann : public PreprocessingPass
  private:
   /* Map each function to a set of terms associated with it */
   FunctionToArgsMap d_funcToArgs;
-  /* Map each function term to the new Skolem variable created by
+  /* Map each function-application term to the new Skolem variable created by
    * ackermannization */
   theory::SubstitutionMap d_funcToSkolem;
+  /* Map each uninterpreted sort term to the new Skolem variable created by
+   * ackermannization */
+  theory::SubstitutionMap d_sortsToSkolem;
```
yoni206: We don't map sorts to skolems, but variables of uninterpreted sorts to variables of bv sorts. So I suggest renaming this field to something like
`d_usVarsToBVVars` (or something nicer with the same meaning).

src/preprocessing/passes/ackermann.h
```diff
@@ -61,9 +65,18 @@ class Ackermann : public PreprocessingPass
...
+  /* Map each uninterpreted sort term to a pair of integers.
+   * The first value is the lowest capacity that the targeting BV should have
+   * The second value is the size of the BV to which the sort will actually be
+   * converted into */
```
yoni206: documentation should be updated (no more second and first).

src/preprocessing/passes/ackermann.cpp
```diff
@@ -187,9 +188,110 @@ void collectFunctionsAndLemmas(FunctionToArgsMap& fun_to_args,
...
 
+/* Given a minimum capacity for an uninterpreted sort, return the size of the
+ * new BV type */
+size_t getBVSkolemSize(size_t capacity)
+{
+  return static_cast<size_t>(log2(capacity)) + 1;
+}
+
+/* Given the lowest capacity requirements for each uninterpreted sort, assign
+ * a sufficient bit-vector size. Get the converting map */
```
yoni206: It is not clear what is the converting map. Could you please clarify in the comment?

yoni206: We are no longer assigning a sufficient bv size, right? It is just the lowest capacity itself I think.

src/preprocessing/passes/ackermann.cpp
```diff
@@ -187,9 +188,110 @@ void collectFunctionsAndLemmas(FunctionToArgsMap& fun_to_args,
...
+void collectUSortsToBV(const vector<TNode>& terms,
```
yoni206: There are actually variables, not arbitrary terms. This is crucial for the correctness of the minimum capacity, so I would rename this variable to `vars`, and make sure that these variables are not called terms throughout the conversion.

src/preprocessing/passes/ackermann.cpp
```diff
@@ -187,9 +188,110 @@ void collectFunctionsAndLemmas(FunctionToArgsMap& fun_to_args,
...
+
+  for (TNode term : terms)
+  {
+    TypeNode type = term.getType();
+    size_t size = getBVSkolemSize(usortCardinality[type]);
+    Node skolem = nm->mkSkolem(
+        "BVSKOLEM$$",
+        nm->mkBitVectorType(size),
+        "a variable created by the ackermannization "
+        "preprocessing pass, representing a term with uninterpreted sort.");
```
yoni206: I think it could be useful to say which original uninterpreted sort was it in this message (when a user does `--dump-assertions`, this information will appear and can be useful for them.

src/preprocessing/passes/ackermann.cpp
```diff
@@ -187,9 +188,110 @@ void collectFunctionsAndLemmas(FunctionToArgsMap& fun_to_args,
...
+  {
+    /* the current version only supports BV for removing uninterpreted sorts */
+    AlwaysAssert(d_logic.isTheoryEnabled(theory::THEORY_BV))
+        << "Cannot use Ackermannization on formula with uninterpreted "
+           "sorts without BV logic";
+
+    for (TNode term : toProcess)
+    {
+      /* ackermannizatino assumes no quantifiers */
+      AlwaysAssert(term.getKind() != kind::BOUND_VARIABLE)
```
yoni206: I think this check should be done at the beginning of the entire Ackermannization preprocessing pass.

src/preprocessing/passes/ackermann.cpp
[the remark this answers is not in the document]
Ying1123: Good point to change the vector to set. But is this a bit hack? Actually, it maybe confusing about the name of the function. Just as my experiences, usually, getXXX() should return the value we want to get.
But this maybe a historical thing, I will just change it as you suggested.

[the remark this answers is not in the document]
4tXJ7f: Maybe I am misunderstanding but you could argue that the value that want to get is a set of variables with uninterpreted sorts, so the name is not wrong :) `getVarsWithUSorts` doesn't say anything about whether the variables are in a set or in a vector, IMHO.

src/preprocessing/passes/ackermann.cpp
4tXJ7f: I'd suggest that we just check in `SmtEngine` whether the logic is quantifier-free or not. What do you think about that?

src/preprocessing/passes/ackermann.cpp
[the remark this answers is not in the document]
Ying1123: Oh, and if I want to return a list with only U sort terms, it is better to have separate them.

src/preprocessing/passes/ackermann.cpp
Ying1123: it is sufficient, because 2^size may greater than the lowest capacity

src/preprocessing/passes/ackermann.cpp
```diff
@@ -187,9 +188,110 @@ void collectFunctionsAndLemmas(FunctionToArgsMap& fun_to_args,
...
+/* Given a minimum capacity for an uninterpreted sort, return the size of the
+ * new BV type */
+size_t getBVSkolemSize(size_t capacity)
+{
+  return static_cast<size_t>(log2(capacity)) + 1;
+}
+
+/* Given the lowest capacity requirements for each uninterpreted sort, assign
+ * a sufficient bit-vector size.
+ * Get the map from variables with uninterpreted sort to the fresh skolem BV
```
yoni206:
```suggestion
 * Populate usVarsToBVVars so that it maps variables with uninterpreted sort to the fresh skolem BV variables.
```

src/preprocessing/passes/ackermann.cpp
```diff
@@ -187,9 +188,110 @@ void collectFunctionsAndLemmas(FunctionToArgsMap& fun_to_args,
...
+                        USortToBVSizeMap& usortCardinality,
+                        SubstitutionMap& usVarsToBVVars)
+{
+  std::unordered_set<TNode, TNodeHashFunction> toProcess =
+      getVarsWithUSorts(assertions);
+
+  if (toProcess.size() > 0)
+  {
+    /* the current version only supports BV for removing uninterpreted sorts */
+    AlwaysAssert(d_logic.isTheoryEnabled(theory::THEORY_BV))
```
yoni206: I just discussed this with @4tXJ7f and we think that it is better in this case not to abort but to leave uninterpreted sorts intact. So if BV is not enabled, just don't change the assertions at all. One way to do it would be (in addition to deleting the following line and formatting) is:

```suggestion
    if (d_logic.isTheoryEnabled(theory::THEORY_BV)) { return;}
```

src/preprocessing/passes/ackermann.cpp
```diff
@@ -187,9 +188,110 @@ void collectFunctionsAndLemmas(FunctionToArgsMap& fun_to_args,
...
+        << "Cannot use Ackermannization on formula with uninterpreted "
+           "sorts without BV logic";
+
+    for (TNode term : toProcess)
+    {
+      TypeNode type = term.getType();
+      /* Update the statistics for each uninterpreted sort.
```
yoni206: statistics is kind of a "reserved" word in CVC4 for solving statistics. Suggestion:

```suggestion
      /* Update the counts for each uninterpreted sort.
```

src/preprocessing/passes/ackermann.cpp
```diff
@@ -187,9 +188,110 @@ void collectFunctionsAndLemmas(FunctionToArgsMap& fun_to_args,
...
+       * For non-existing keys, C++ will create a new element for it, which has
+       * the value initialized with a pair of two zeros. */
```
yoni206: Is it still a pair of two zeros? I think now it is only a single zero.
And actually right there you increment it by 1, and then the result is 1, not 0. so I suggest:

```suggestion
       * a default 0 value, before incrementing by 1. */
```

src/preprocessing/passes/ackermann.cpp
```diff
@@ -187,9 +188,110 @@ void collectFunctionsAndLemmas(FunctionToArgsMap& fun_to_args,
...
+{
+  return static_cast<size_t>(log2(capacity)) + 1;
+}
+
+/* Given the lowest capacity requirements for each uninterpreted sort, assign
+ * a sufficient bit-vector size.
+ * Get the map from variables with uninterpreted sort to the fresh skolem BV
+ * variables */
+void collectUSortsToBV(const unordered_set<TNode, TNodeHashFunction>& vars,
+                       USortToBVSizeMap& usortCardinality,
```
4tXJ7f: It looks like this can be `const USortToBVSizeMap& usortCardinality` (use `usortCardinality->at(type)` instead of `usortCardinality[type]`).

src/preprocessing/passes/ackermann.h
```diff
@@ -38,12 +38,16 @@ namespace passes {
...
     std::unordered_map<TNode, TNodeSet, TNodeHashFunction>;
+using USortToBVSizeMap =
+    std::unordered_map<TypeNode, size_t, TypeNode::HashFunction>;
 
 class Ackermann : public PreprocessingPass
 {
  public:
   Ackermann(PreprocessingPassContext* preprocContext);
 
+  size_t getBVSkolemSize(size_t capacity);
```
4tXJ7f: This is a helper function like `collectUSortsToBV`, no need to declare it in the header.

PR #3374, opened 2019-10-09
Reorder NodeValue class according to our code style guidelines. 

aniemetz:
Reorder is in commit https://github.com/CVC4/CVC4/commit/26200d08e47b0a6ea2677f9b21e6a28728887a92.
Commit https://github.com/CVC4/CVC4/commit/61bb192e67f9dc82bfffe090da845fc290d845b6 the return type of a member function.
Commit https://github.com/CVC4/CVC4/commit/3a9703faa27edaf52e22db5ce2bbbc64209e541f replaces all `typedef`s in the NodeValue class with `using` statements.


src/expr/node_value.cpp
```diff
@@ -78,5 +78,23 @@ void NodeValue::printAst(std::ostream& out, int ind) const {
...
+    NodeValue::iterator<NodeTemplate<true> >::difference_type p,
+    NodeValue::iterator<NodeTemplate<true> > i)
+{
+  return i + p;
+}
+
+/**
+ * Provides a symmetric addition operator to that already defined in
+ * the iterator class.
+ */
```
4tXJ7f: This comment is already in the header, I think that we can remove it from here.

src/expr/node_value.h
```diff
@@ -281,87 +221,144 @@ class NodeValue {
...
+  static const uint32_t NBITS_ID = CVC4__EXPR__NODE_VALUE__NBITS__ID;
+  static const uint32_t NBITS_NCHILDREN =
+      CVC4__EXPR__NODE_VALUE__NBITS__NCHILDREN;
+
+  /** Maximum reference count possible.  Used for sticky
+   *  reference-counting.  Should be (1 << num_bits(d_rc)) - 1 */
+  static const uint32_t MAX_RC = (((uint32_t)1) << NBITS_REFCOUNT) - 1;
+
+  /** A mask for d_kind */
+  static const uint32_t kindMask = (((uint32_t)1) << NBITS_KIND) - 1;
```
4tXJ7f: Data members actually go to the end of the class according to our ordering.

aniemetz: Good catch! I did the reordering twice in the end since I had to narrow down compilation issues and this slipped through somehow.

aniemetz: Our guidelines say that this does not apply to constants, they come after types and nested structs and classes: https://github.com/CVC4/CVC4/wiki/Code-Style-Guidelines#declaration-order

4tXJ7f: Ah true, I was focusing on them being static data members.

PR #3379, opened 2019-10-09
Make order of theories explicit in the source code.

### LINKED ISSUE #2517: Make order of theories explicit in the source code.
Until now the build system was used to ensure that the order of theories is correct (`src/Makefile.theories`). We should make this order explicit in the source code (maybe `src/theory/theory_engine.h`) and should not rely on the build system.

aniemetz:
Fixes #2517.

This makes the order of theories explicit in the source code rather than relying on the order defined via the build system. Previously, the build system ensured the order of the theories via the KINDS_FILES variable, which is a list of kinds files that is fed to code generation scripts (`mkkind`, `mkmetakind`, `mkrewriter`, `mktheorytraits`). The generated code critical to the order of theories w.r.t. soundess is the `TheoryId enum`, and the `CVC4_FOR_EACH_THEORY` macro. Ideally, we would want to get rid of the latter (ugly and error prone), which is not possible in the current configuration, and to be discussed in the future.

This PR moves the TheoryID enum and related functions to `theory/theory_id.h`, and the `CVC4_FOR_EACH_THEORY` macro to `theory/theory_engine.cpp`, the only place where it is used.

I ran it on whole SMT-LIB (non-incremental and incremental) and did not encounter any soundness issues. The only issue that did occur is not related to these changes, non-critical and known: https://github.com/CVC4/CVC4/issues/2993

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -682,6 +682,8 @@ libcvc4_add_sources(
   theory/theory.h
   theory/theory_engine.cpp
   theory/theory_engine.h
+  theory/theory_id.cpp
+  theory/theory_id.h
   theory/theory_model.cpp
   theory/theory_model.h
   theory/theory_model_builder.cpp
@@ -717,25 +719,21 @@ include_directories(include)
 include_directories(. ${CMAKE_CURRENT_BINARY_DIR})
 
 #-----------------------------------------------------------------------------#
-# IMPORTANT: The order of the theories is important. It affects the order in
-# which theory solvers are called/initialized internally. For example, strings
-# depends on arith, quantifiers needs to come as the very last.
-# See issue https://github.com/CVC4/CVC4/issues/2517 for more details.
 
 set(KINDS_FILES
+  ${PROJECT_SOURCE_DIR}/src/theory/arrays/kinds
+  ${PROJECT_SOURCE_DIR}/src/theory/arith/kinds
   ${PROJECT_SOURCE_DIR}/src/theory/builtin/kinds
   ${PROJECT_SOURCE_DIR}/src/theory/booleans/kinds
-  ${PROJECT_SOURCE_DIR}/src/theory/uf/kinds
-  ${PROJECT_SOURCE_DIR}/src/theory/arith/kinds
   ${PROJECT_SOURCE_DIR}/src/theory/bv/kinds
-  ${PROJECT_SOURCE_DIR}/src/theory/fp/kinds
-  ${PROJECT_SOURCE_DIR}/src/theory/arrays/kinds
   ${PROJECT_SOURCE_DIR}/src/theory/datatypes/kinds
+  ${PROJECT_SOURCE_DIR}/src/theory/idl/kinds
+  ${PROJECT_SOURCE_DIR}/src/theory/fp/kinds
+  ${PROJECT_SOURCE_DIR}/src/theory/quantifiers/kinds
   ${PROJECT_SOURCE_DIR}/src/theory/sep/kinds
   ${PROJECT_SOURCE_DIR}/src/theory/sets/kinds
   ${PROJECT_SOURCE_DIR}/src/theory/strings/kinds
-  ${PROJECT_SOURCE_DIR}/src/theory/quantifiers/kinds
-  ${PROJECT_SOURCE_DIR}/src/theory/idl/kinds)
+  ${PROJECT_SOURCE_DIR}/src/theory/uf/kinds)
 
 #-----------------------------------------------------------------------------#
 # Add subdirectories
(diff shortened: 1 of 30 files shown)


src/CMakeLists.txt
```diff
@@ -717,10 +719,21 @@ include_directories(include)
...
+#  ${PROJECT_SOURCE_DIR}/src/theory/booleans/kinds
+#  ${PROJECT_SOURCE_DIR}/src/theory/bv/kinds
+#  ${PROJECT_SOURCE_DIR}/src/theory/datatypes/kinds
+#  ${PROJECT_SOURCE_DIR}/src/theory/idl/kinds
+#  ${PROJECT_SOURCE_DIR}/src/theory/fp/kinds
+#  ${PROJECT_SOURCE_DIR}/src/theory/quantifiers/kinds
+#  ${PROJECT_SOURCE_DIR}/src/theory/sep/kinds
+#  ${PROJECT_SOURCE_DIR}/src/theory/sets/kinds
+#  ${PROJECT_SOURCE_DIR}/src/theory/strings/kinds
+#  ${PROJECT_SOURCE_DIR}/src/theory/uf/kinds)
```
4tXJ7f: Remove?

src/theory/theory_engine.cpp
```diff
@@ -59,6 +59,34 @@ using namespace CVC4::theory;
...
+  CVC4_FOR_EACH_THEORY_STATEMENT(CVC4::theory::THEORY_UF)        \
+  CVC4_FOR_EACH_THEORY_STATEMENT(CVC4::theory::THEORY_ARITH)     \
+  CVC4_FOR_EACH_THEORY_STATEMENT(CVC4::theory::THEORY_BV)        \
+  CVC4_FOR_EACH_THEORY_STATEMENT(CVC4::theory::THEORY_FP)        \
+  CVC4_FOR_EACH_THEORY_STATEMENT(CVC4::theory::THEORY_ARRAYS)    \
+  CVC4_FOR_EACH_THEORY_STATEMENT(CVC4::theory::THEORY_DATATYPES) \
+  CVC4_FOR_EACH_THEORY_STATEMENT(CVC4::theory::THEORY_SEP)       \
+  CVC4_FOR_EACH_THEORY_STATEMENT(CVC4::theory::THEORY_SETS)      \
+  CVC4_FOR_EACH_THEORY_STATEMENT(CVC4::theory::THEORY_STRINGS)   \
+  CVC4_FOR_EACH_THEORY_STATEMENT(CVC4::theory::THEORY_QUANTIFIERS)
```
4tXJ7f: Couldn't we make this a for loop that iterates over the `TheoryId`s? Like that we only have the order of the enum and don't need to replicate it here.

src/expr/kind_template.cpp
```diff
@@ -101,22 +90,11 @@ TheoryId kindToTheoryId(::CVC4::Kind k) {
 TheoryId typeConstantToTheoryId(::CVC4::TypeConstant typeConstant) {
   switch(typeConstant) {
 ${type_constant_to_theory_id}
-#line 105 "${template}"
-  case LAST_TYPE:
-    break;
+#line 94 "${template}"
+case LAST_TYPE: break;
```
4tXJ7f: Wrong indentation.

src/theory/theory_id.h
```diff
@@ -0,0 +1,46 @@
+#include "cvc4_public.h"
+
+#ifndef CVC4__THEORY_ID_H
```
4tXJ7f: This should be `CVC4__THEORY__THEORY_ID_H`.

src/theory/theory_id.h
```diff
@@ -0,0 +1,46 @@
...
+  THEORY_UF,
+  THEORY_ARITH,
+  THEORY_BV,
+  THEORY_FP,
+  THEORY_ARRAYS,
+  THEORY_DATATYPES,
+  THEORY_SEP,
+  THEORY_SETS,
+  THEORY_STRINGS,
+  THEORY_QUANTIFIERS,
```
4tXJ7f: What about `THEORY_IDL`?

src/theory/theory_engine.cpp
aniemetz: No, because we need the theory to instantiate a template.

src/expr/kind_template.cpp
aniemetz: Interesting. Clang-format screwed that up. I formatted the whole function now.

src/theory/theory_engine.cpp
4tXJ7f: Ah, right :/ That's a bit unfortunate. Oh well... we may be able to change it in the future. It looks like it is mainly a template because of the theory traits (which could also be implemented using virtual method dispatch).

PR #3384, opened 2019-10-11
Check that logic is set when synth-fun command is encountered

ajreynol:
This ensures we assume the default ALL logic when `synth-fun` is encountered before `set-logic` instead of aborting due to undefined symbols. This behavior was introduced in #3335; this PR adds back a line of code that was deleted in that PR.


4tXJ7f:
Do you have a simple example that causes issues?

ajreynol:
I don't think its possible to have a regression currently, since we don't have a way of saying what is the expected output on stderr which is what the error about no logic is printed on.

ajreynol:
nvm, i think i figured it out

PR #3385, opened 2019-10-11
Ensure lemmas from sygus repair const are guarded

ajreynol:
Previously we could spuriously answer "unsat" due to excluding the entire range of values for a datatype.  This was caused by not guarding a refinement lemma from the repair constant module was guarded by the "feasible guard" of the sygus conjecture.

### DIFF
--- a/test/regress/regress1/sygus/repair-const-unk.sy
+++ b/test/regress/regress1/sygus/repair-const-unk.sy
@@ -1,0 +1,17 @@
+; EXPECT: unknown
+; COMMAND-LINE: --sygus-out=status --sygus-repair-const --lang=sygus2 --no-sygus-pbe
+(set-logic LIA)
+(synth-fun f ((x Int) (y Int)) Int
+  ((Start Int) (CInt Int))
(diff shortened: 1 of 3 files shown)


test/regress/regress1/sygus/repair-const-unk.sy
```diff
@@ -0,0 +1,17 @@
+; EXPECT: unknown
+; COMMAND-LINE: --sygus-out=status --sygus-repair-const --lang=sygus2 --no-sygus-pbe
```
HanielB: Why do you need `--no-sygus-pbe` here?

ajreynol: That just happened to be the options I was using. This isn't important really.

HanielB: I think it's best to have the minimal set of options that leads to desired behavior so that we avoid confusion in the future. If you can remove `--no-sygus-pbe` I think it's best to do it.

PR #3386, opened 2019-10-11
Apply sygus repair constant techniques restricted to refinement lemmas

ajreynol:
This supplements the techniques for repairing constants in sygus solutions via ensuring the constants supplied to the current skeleton are a *candidate* solution, that is, a solution for the current set of CEGIS refinement lemmas. Previously, we enforced that the repair constant module ensured the constants supplied to the current skeleton are a solution to the entire synthesis conjecture.

The current fairness scheme is naive (each skeleton is repaired exactly once). In the future, we could consider repairing skeletons multiple times.

This is work towards improving predicate generation for Unif+PI.

### DIFF
--- a/src/theory/quantifiers/sygus/cegis.cpp
+++ b/src/theory/quantifiers/sygus/cegis.cpp
@@ -215,13 +215,9 @@ Node Cegis::getRefinementLemmaFormula()
   {
     ret = nm->mkConst(true);
   }
-  else if (conj.size() == 1)
-  {
-    ret = conj[0];
-  }
   else
   {
-    ret = nm->mkNode(AND, conj);
+    ret = conj.size() == 1 ? conj[0] : nm->mkNode(AND, conj);
   }
   return ret;
 }
@@ -269,15 +265,11 @@ bool Cegis::constructCandidates(const std::vector<Node>& enums,
         return true;
       }
       Node rl = getRefinementLemmaFormula();
-      bool ret = false;
       // try to solve for the refinement lemmas only
-      if (src->repairSolution(rl, candidates, fail_cvs, candidate_values))
-      {
-        // We will exclude the skeleton as well; this means that we have one
-        // chance to repair the solution.
-        ret = true;
-      }
-      // repair solution didn't work, exclude this solution
+      bool ret = src->repairSolution(rl, candidates, fail_cvs, candidate_values);
+      // Even if ret is true, we will exclude the skeleton as well; this means
+      // that we have one chance to repair each skeleton. It is possible however
+      // that we might want to repair the same skeleton multiple times.
       std::vector<Node> exp;
       for (unsigned i = 0, size = enums.size(); i < size; i++)
       {
--- a/src/theory/quantifiers/sygus/cegis.cpp
+++ b/src/theory/quantifiers/sygus/cegis.cpp
@@ -269,10 +269,13 @@ bool Cegis::constructCandidates(const std::vector<Node>& enums,
         return true;
       }
       Node rl = getRefinementLemmaFormula();
+      bool ret = false;
       // try to solve for the refinement lemmas only
       if (src->repairSolution(rl,candidates, fail_cvs, candidate_values))
       {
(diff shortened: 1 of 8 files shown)


src/theory/quantifiers/sygus/cegis.cpp
```diff
@@ -198,6 +198,34 @@ bool Cegis::addEvalLemmas(const std::vector<Node>& candidates,
...
+    ret = nm->mkConst(true);
+  }
+  else if (conj.size() == 1)
+  {
+    ret = conj[0];
+  }
+  else
+  {
+    ret = nm->mkNode(AND, conj);
+  }
```
HanielB: I suppose you'd not like to inline this to ret = cond1? ... : cond2? ... : ...?

src/theory/quantifiers/sygus/cegis.cpp
```diff
@@ -235,10 +263,20 @@ bool Cegis::constructCandidates(const std::vector<Node>& enums,
...
       {
         return true;
       }
+      Node rl = getRefinementLemmaFormula();
+      bool ret = false;
+      // try to solve for the refinement lemmas only
+      if (src->repairSolution(rl, candidates, fail_cvs, candidate_values))
+      {
+        // We will exclude the skeleton as well; this means that we have one
+        // chance to repair the solution.
```
HanielB: This is a bit confusing. I assume this comment refers to the `repairSolution` call above taking `rl` right?

src/theory/quantifiers/sygus/cegis.cpp
```diff
@@ -235,10 +263,20 @@ bool Cegis::constructCandidates(const std::vector<Node>& enums,
...
+        ret = true;
```
HanielB: Why do you have this in an `if` block rather than `ret = src->repairSolution(rl, candidates, fail_cvs, candidate_values)`?

src/theory/quantifiers/sygus/cegis.cpp
```diff
@@ -235,10 +263,20 @@ bool Cegis::constructCandidates(const std::vector<Node>& enums,
     {
       std::vector<Node> fail_cvs = enum_values;
       Assert(candidates.size() == fail_cvs.size());
+      // try to solve entire problem?
       if (src->repairSolution(candidates, fail_cvs, candidate_values))
       {
         return true;
```
HanielB: This PR made me think that it's spurious to do the solution check upstream when we know, in this point here, that `candidate_values` will necessarily work. Maybe we should make `constructCandidates` return an enum rather than a bool to account for fail / candidate / solution?

src/theory/quantifiers/sygus/cegis.cpp
ajreynol: It gets a little ugly if its highly nested, I'll merge the last 2 blocks though.

src/theory/quantifiers/sygus/cegis.cpp
ajreynol: The comment refers to the fact that we aren't returning here and falling through.  I'll reorganize this a bit to make it clearer.

src/theory/quantifiers/sygus/cegis.cpp
ajreynol: I've refactored.

src/theory/quantifiers/sygus/cegis.cpp
ajreynol: Although it is spurious, I think it's best to re-verify the solution upstream.  This isn't very expensive and makes things more uniform.

ajreynol:
Btw, I also removed the debugging code for primed variables, since I realized sygus v2 invariant problems always give a warning for invariant synthesis problems otherwise.

src/theory/quantifiers/sygus/cegis.cpp
HanielB: Yes... I was not too keen on the micro-optimization.

PR #3387, opened 2019-10-11
Minor refactor in strings rewriter

ajreynol:


src/theory/strings/theory_strings_rewriter.cpp
```diff
@@ -631,30 +631,37 @@ Node TheoryStringsRewriter::rewriteConcat(Node node)
...
-          if(tmpNode[0].isConst()) {
-            preNode = NodeManager::currentNM()->mkConst( preNode.getConst<String>().concat( tmpNode[0].getConst<String>() ) );
-            node_vec.push_back( preNode );
-          } else {
-            node_vec.push_back( preNode );
-            node_vec.push_back( tmpNode[0] );
-          }
-          preNode = Node::null();
-          ++j;
+  for (const Node& nc : node)
```
4tXJ7f: Could just do `for (Node tmp : node)`, then we don't need the next line.

src/theory/strings/theory_strings_rewriter.cpp
```diff
@@ -631,30 +631,37 @@ Node TheoryStringsRewriter::rewriteConcat(Node node)
...
+          node_vec.push_back(tmpNode[0]);
         }
-        tmpNode = tmpNode[j];
+        preNode = Node::null();
+        ++j;
+      }
+      for (; j < tmpNode.getNumChildren() - 1; ++j)
+      {
+        node_vec.push_back(tmpNode[j]);
       }
```
4tXJ7f: Could do: `node_vec.insert(node_vec.end(), tmpNode.begin() + j, tmpNode.end() - 1);` (untested)

ajreynol: There were a few issues with doing this, but I've now refactored so there isn't a for loop.

PR #3392, opened 2019-10-15
Universe set cardinality for finite types with finite cardinality

mudathirmahgoub:
This pull request relies on the pull requests [3381](https://github.com/CVC4/CVC4/pull/3381) and [3377](https://github.com/CVC4/CVC4/pull/3377) which are merged now.

This pull request partially fixes the issue [1124](https://github.com/CVC4/CVC4/issues/1124) by adding the following lemmas for each finite type `t` and for each Set `S`:

1- `(=> true (<= (card (as univset t)) n)` where `n` is the cardinality of `t`
2- `(=> true (subset S (as univset t))`
3- `(=> (not (member negativeMember S))) (member negativeMember (as univset t)))`

- All the tests in `regress1/sets/finite-type/` are solved by these lemmas except [sets-card-neg-mem-union-2.smt2](https://github.com/CVC4/CVC4/compare/master...mudathirmahgoub:universeCardinality?expand=1#diff-5b2c0ca0be71cbb82c18fe36a492e3dd) which times out after generating a lot of conflicts.

- Cardinality for uninterpreted sorts with `finite-model-find` will generate an assertion error 
`The cardinality beth[0] of the finite type Atom is not supported yet.`

- Cardinality for relations or universe sets of infinite types will return `unknown` for now.


4tXJ7f:
Quick question: It seems to me that this PR relies on #3377, not #3381. Am I missing something?

test/regress/regress1/sets/finite-type/sets-card-neg-mem-union-1.smt2
```diff
@@ -0,0 +1,30 @@
+(set-logic QF_ALL)
+;(set-info :status unsat)
```
4tXJ7f: This seems to be the same file as test/regress/regress1/sets/finite-type/sets-card-neg-mem-union-2.smt2 except for this commented out line and this file not appearing in the list of regressions, so we can probably remove this file from the PR.

src/theory/sets/cardinality_extension.cpp
```diff
@@ -57,20 +65,105 @@ void CardinalityExtension::registerTerm(Node n)
...
+        d_im.assertInference(subset, d_state.d_true, "univset is a super set", 1);
+      }
+
+      // negative members are members in the universe set
+      const std::map<Node, Node>& negativeMembers = d_state.getNegativeMembers(representative);
+
+      for (const std::pair<Node,  Node> & negativeMember: negativeMembers)
+      {
+        Node member = nm->mkNode(MEMBER, negativeMember.first, univ);
+        // negativeMember.second is the reason for the negative membership and has kind MEMBER. So We
```
4tXJ7f:
```suggestion
        // negativeMember.second is the reason for the negative membership and has kind MEMBER. So we
```

src/theory/sets/cardinality_extension.cpp
```diff
@@ -57,20 +65,105 @@ void CardinalityExtension::registerTerm(Node n)
...
+    proxy = it->second;
+  }
+
+  // get all equivalent classes of type t
+  vector<Node> representatives = d_state.getSetsEqClasses(t);
+  // get the cardinality of the finite type t
+  Cardinality card = t.getCardinality();
+  if(!card.isFinite())
+  {
+    //ToDo: support uninterpreted types with --finite-model-find
```
4tXJ7f: TODOs should refer to an issue (if the issue doesn't exist yet, please create one).

src/theory/sets/cardinality_extension.cpp
```diff
@@ -57,20 +65,105 @@ void CardinalityExtension::registerTerm(Node n)
...
+    }
+  }
+}
+
+void CardinalityExtension::checkFiniteType(TypeNode & t)
+{
+  Assert(t.isInterpretedFinite());
+  NodeManager* nm = NodeManager::currentNM();
+  // get the universe set of type t
+  // ToDo: investigate refactoring nm->mkSetType
```
4tXJ7f: TODOs should refer to an issue (if the issue doesn't exist yet, please create one).

src/theory/sets/solver_state.h
```diff
@@ -200,10 +203,12 @@ class SolverState
   /** debug print set */
   void debugPrintSet(Node s, const char* c) const;
 
- private:
   /** constants */
   Node d_true;
   Node d_false;
+
+private:
```
4tXJ7f: Not a huge fan of making data members public. Maybe define getters?

src/theory/sets/cardinality_extension.cpp
```diff
@@ -57,20 +65,105 @@ void CardinalityExtension::registerTerm(Node n)
...
+      subset = Rewriter::rewrite(subset);
+      if(! d_state.isEntailed(subset, true))
+      {
+        d_im.assertInference(subset, d_state.d_true, "univset is a super set", 1);
+      }
+
+      // negative members are members in the universe set
+      const std::map<Node, Node>& negativeMembers = d_state.getNegativeMembers(representative);
+
+      for (const std::pair<Node,  Node> & negativeMember: negativeMembers)
```
4tXJ7f: Apply ClangFormat.

mudathirmahgoub:
> Quick question: It seems to me that this PR relies on #3377, not #3381. Also, sets-card-neg-mem-union-2.smt2 seems to be solved in less than a second. Am I missing something?

Actually it depends on both of them. For `sets-card-neg-mem-union-2.smt2`, I copied the wrong file. Will change. The original one did not finish after 10 hours !!

test/regress/regress1/sets/finite-type/sets-card-neg-mem-union-1.smt2
mudathirmahgoub: Yes. It is fixed now.

src/theory/sets/cardinality_extension.cpp
mudathirmahgoub: This is related to issue 1123 which is not totally fixed. I wonder if we need to split that issue. For now I would just refer to [1123](https://github.com/CVC4/CVC4/issues/1123).

src/theory/sets/cardinality_extension.cpp
mudathirmahgoub: Actually I don't think it needs refactoring any more. It was weird to me that I need to get a `(Set t)` out of type `t` to get the universe set. But now I think it is more consistent with SMT-LIB  `(as univset (Set t))`. So I removed this `ToDo` comment

src/theory/sets/solver_state.h
mudathirmahgoub: Done.  I wonder if there is a nice way to globally access these 2 constants which are common. @ajreynol  pointed out before using static objects to hold these constants afftects the performance.

src/theory/sets/cardinality_extension.cpp
mudathirmahgoub: Done

src/theory/sets/solver_state.h
4tXJ7f: Yeah, we should probably add them to the `NodeManager` (in another PR). This shouldn't negatively affect performance but we should double check.

src/theory/sets/cardinality_extension.cpp
```diff
@@ -57,20 +65,113 @@ void CardinalityExtension::registerTerm(Node n)
...
+    proxy = it->second;
+  }
+
+  // get all equivalent classes of type t
+  vector<Node> representatives = d_state.getSetsEqClasses(t);
+  // get the cardinality of the finite type t
+  Cardinality card = t.getCardinality();
+  if (!card.isFinite())
+  {
+    // ToDo (#1123): support uninterpreted sorts with --finite-model-find
```
ajreynol: We usually write "TODO" not "ToDo".

src/theory/sets/cardinality_extension.cpp
```diff
@@ -57,20 +65,113 @@ void CardinalityExtension::registerTerm(Node n)
...
+            << " is not supported yet." << endl;
+    Assert(false) << message.str().c_str();
+  }
+
+  Node typeCardinality = nm->mkConst(Rational(card.getFiniteCardinality()));
+
+  Node cardUniv = nm->mkNode(kind::CARD, proxy);
+  Node leq = nm->mkNode(kind::LEQ, cardUniv, typeCardinality);
+
+  /** (=> true (<= (card (as univset t)) cardUniv) */
```
ajreynol: Nit: as a general rule here, I prefer comments with `//` instead of `/** */` since the latter makes it harder to do other block comments when debugging.

src/theory/sets/cardinality_extension.cpp
```diff
@@ -57,20 +65,113 @@ void CardinalityExtension::registerTerm(Node n)
...
+  if (!d_state.isEntailed(leq, true))
+  {
+    d_im.assertInference(leq, d_state.getTrue(), "finite type cardinality", 1);
+  }
+
+  // add subset lemmas for sets and membership lemmas for negative members
+  for (Node& representative : representatives)
+  {
+    if (representative
```
ajreynol: Nit: can you put the comment on the line before so that clang-format is prettier here?

src/theory/sets/solver_state.h
```diff
@@ -204,6 +209,9 @@ class SolverState
   /** debug print set */
   void debugPrintSet(Node s, const char* c) const;
 
+  /** get the constant true */
+  const Node& getTrue() const { return d_true; }
```
ajreynol: Can we remove this?  If you want to use true somewhere, either recreate it on the spot or cache it local to the class that is using it.

src/theory/sets/cardinality_extension.cpp
mudathirmahgoub: > The code looks pretty good to me.
> 
> I have some minor comments. Also, I noticed that the entirety of TheorySetsPrivate was formatted. In the future, do try to avoid formatting code that was not changed.

Reviewing code changes and format together is a bit harder. How about having pull requests only for formatting large files to unify the format in cvc4?

src/theory/sets/cardinality_extension.cpp
mudathirmahgoub: Done. What does `Nit` mean by the way?

src/theory/sets/solver_state.h
mudathirmahgoub: Done.

src/theory/sets/cardinality_extension.cpp
ajreynol: "nitpick", it just means its a really minor comment.

src/theory/sets/cardinality_extension.cpp
ajreynol: We've been progressively formatting the files in CVC4. You can submit PRs to format separately (and to update the old code to new coding guidelines).  Since the code base of CVC4 is very large, this can't be done easily, so this is a low-priority task in general.

PR #3396, opened 2019-10-16
Update overflow check to handle negative numbers

makaimann:
Previously, the overflow check in `mkBVFromStrHelper` did not consider negative values. This pull request simply handles that case.

### DIFF
--- a/src/api/cvc4cpp.cpp
+++ b/src/api/cvc4cpp.cpp
@@ -2023,9 +2023,19 @@ Term Solver::mkBVFromStrHelper(uint32_t size,
       << "base 2, 10, or 16";
 
   Integer val(s, base);
-  CVC4_API_CHECK(val.modByPow2(size) == val)
-      << "Overflow in bitvector construction (specified bitvector size " << size
-      << " too small to hold value " << s << ")";
+
+  if (val.strictlyNegative())
+  {
+    CVC4_API_CHECK(val >= -Integer("2", 10).pow(size - 1))
+        << "Overflow in bitvector construction (specified bitvector size "
+        << size << " too small to hold value " << s << ")";
+  }
+  else
+  {
+    CVC4_API_CHECK(val.modByPow2(size) == val)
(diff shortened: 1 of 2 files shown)


makaimann:
I'm not sure if branching around the `CVC4_API_CHECK`s is the right way to handle this. After a cursory search, I didn't see anything to compare to. I can do the entire check within the argument if that's preferred.

src/api/cvc4cpp.cpp
```diff
@@ -2023,9 +2023,19 @@ Term Solver::mkBVFromStrHelper(uint32_t size,
       << "base 2, 10, or 16";
 
   Integer val(s, base);
-  CVC4_API_CHECK(val.modByPow2(size) == val)
-      << "Overflow in bitvector construction (specified bitvector size " << size
-      << " too small to hold value " << s << ")";
+
+  if (val.strictlyNegative())
+  {
+    CVC4_API_CHECK(val >= -Integer("2", 10).pow(size - 1))
```
aniemetz: The condition for `CVC4_API_CHECK` is the one you want to hold for the check to *not* fail -- looks like this condition is the fail condition? Also, I'm missing a corresponding unit test.

makaimann: Oops, yeah forgot to add unit tests. Added with e7193b6d28ff143693267d85ad610b4c15dd08d2. Regarding the API check, it's correct as is. The value is negative so it needs to be greater than the minimum bound. It's tested in the added unit test.

aniemetz: Ah right, my bad.

aniemetz:
> I'm not sure if branching around the `CVC4_API_CHECK`s is the right way to handle this. After a cursory search, I didn't see anything to compare to. I can do the entire check within the argument if that's preferred.
> 
> Alternatively, we could go a different route and not support creating negative bit-vectors directly at all -- in which case, I could change this to check for that.

I think that's fine, else we have to do the `strictlyNegative()` check twice.

PR #3403, opened 2019-10-23
Fix global-declarations support

### LINKED ISSUE #3402: bug with global-declarations and declare-datatypes
The following script demonstrates that CVC4 forgets datatype declarations after a pop even when the `global-declarations` option is set.

```
$ cat cvc4bug.smt2
(set-option :global-declarations true)
(set-logic ALL_SUPPORTED)

(push 1)
(declare-datatype Struct1 (par (T0) ((mk-struct1 (struct1-proj0 T0)))))
(declare-fun x () (Struct1 Bool))
(pop 1)

;(assert (= x false))

(define-fun y () (Struct1 Bool) (mk-struct1 true))

$ cvc4 -i cvc4bug.smt2
(error "Parse Error: cvc4bug.smt2:11.31: Symbol 'Struct1' not declared as a type

  (define-fun y () (Struct1 Bool) (mk-struct1 true))
                    ^
")

$ cvc4 --version
This is CVC4 version 1.7-prerelease [git master 1955e4b5]
```

I'm using homebrew CVC4 for OS X, which is pinned to a prerelease for some reason... I've had trouble compiling CVC4 from source in the past.

Note that the constant `x` remains in scope, even though it's type becomes inaccessable.  Morever, CVC4 remembers its type: if you uncomment the assertion above, you get the following error:
```
(error "Subexpressions must have a common base type:
Equation: (= x false)
Type 1: (Struct1 Bool)
Type 2: Bool
")
```

4tXJ7f:
Fixes #3402. Previously, we were removing datatype and uninterpreted
sort declarations when popping the user context even with
`global-declarations` set to true. However, the SMT-LIB standard says
that ":global-declarations [...] makes all definitions and declarations
global and not removable by pop operations". This commit fixes those
issues by making sure that the uninterpreted sorts as well as the
datatype sorts, constructors, and testers are inserted at level zero if
global declarations are enabled.

### DIFF
--- a/src/parser/parser.h
+++ b/src/parser/parser.h
@@ -527,11 +527,15 @@ public:
                  bool levelZero = false, bool doOverload = false);
 
   /** Create a new type definition. */
-  void defineType(const std::string& name, const Type& type);
+  void defineType(const std::string& name,
+                  const Type& type,
+                  bool levelZero = false);
 
   /** Create a new (parameterized) type definition. */
   void defineType(const std::string& name,
-                  const std::vector<Type>& params, const Type& type);
+                  const std::vector<Type>& params,
+                  const Type& type,
+                  bool levelZero = false);
 
   /** Create a new type definition (e.g., from an SMT-LIBv2 define-sort). */
   void defineParameterizedType(const std::string& name,
--- a/src/parser/parser.h
+++ b/src/parser/parser.h
@@ -526,12 +526,27 @@ public:
   void defineVar(const std::string& name, const Expr& val,
                  bool levelZero = false, bool doOverload = false);
 
-  /** Create a new type definition. */
+  /**
+   * Create a new type definition.
+   *
+   * @param name The name of the type
+   * @param type The type that should be associated with the name
+   * @param levelZero If true, the type definition is considered global and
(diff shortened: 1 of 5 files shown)


src/parser/parser.h
```diff
@@ -527,11 +527,15 @@ class CVC4_PUBLIC Parser {
                  bool levelZero = false, bool doOverload = false);
 
   /** Create a new type definition. */
```
ajreynol: Can you add documentation for the arguments (especially the added one).

4tXJ7f: Good point, done.

src/parser/parser.h
```diff
@@ -527,11 +527,15 @@ class CVC4_PUBLIC Parser {
...
-  void defineType(const std::string& name, const Type& type);
+  void defineType(const std::string& name,
+                  const Type& type,
+                  bool levelZero = false);
 
   /** Create a new (parameterized) type definition. */
   void defineType(const std::string& name,
-                  const std::vector<Type>& params, const Type& type);
+                  const std::vector<Type>& params,
```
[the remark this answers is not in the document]
4tXJ7f: Good point, done.

PR #3409, opened 2019-10-27
Fix collect model info for higher-order

### LINKED ISSUE #3405: Assertion failure -- CVC4 HO with finite-model-find
```
./cvc4 --uf-ho --finite-model-find tmp.smt2
(error "Assertion failure.
void CVC4::TheoryEngine::check(CVC4::theory::Theory::Effort)
/home/irfan/work/projects/CVC4/src/theory/theory_engine.cpp:678:

  d_curr_model->isBuiltSuccess()
")
```
[tmp.zip](https://github.com/CVC4/CVC4/files/3774502/tmp.zip)

ajreynol:
This ensures we add lemmas when collect model info fails for the higher order extension of UF.  This fixes #3405 (that benchmark now answers unknown).

### DIFF
--- a/src/theory/theory_model_builder.cpp
+++ b/src/theory/theory_model_builder.cpp
@@ -301,14 +301,17 @@ bool TheoryEngineModelBuilder::buildModel(Model* m)
                          << std::endl;
   if (!d_te->collectModelInfo(tm))
   {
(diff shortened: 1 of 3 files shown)


4tXJ7f:
Add the regression? :)

ajreynol:
I generally don't like adding regressions that are "unknown".

PR #3410, opened 2019-10-27
Fix for non-linear models

### LINKED ISSUE #3407: Invalid model on NRA formula
Hi,
For this formula:
```
(set-logic NRA)
(declare-fun a () Real)
(declare-fun b () Real)
(assert (= (* a b) 1))
(check-sat)
(get-model)
```
CVC4 will give an invalid model:
```
(model
(define-fun a () Real (/ (- 3) 2))
(define-fun b () Real (/ (- 1) 2))
)
```
which does not satisfy the formula.

OS: Ubuntu 18.04
Revision: 64cef99

ajreynol:
This fixes several issues that were leading to wrong models for non-linear inputs:
(1) We were not clearing the model cache when we add an approximation to the model,
(2) We were not recording approximations after a last call effort check with no lemmas.
(3) We were not recording exact values that we are using during non-linear's check model. We now use the same infrastructure for approximations, e.g. the approximation for variable x with exact value c is `choice x. x = c`, which simplifies to `c`.

Fixes #3407.


src/theory/arith/nonlinear_extension.cpp
```diff
@@ -2508,6 +2492,61 @@ void NonlinearExtension::check(Theory::Effort e) {
...
+      {
+        pred = Rewriter::rewrite(pred);
+        m->recordApproximation(v, pred);
+      }
+    }
+    // Also record the exact values we used. An exact value can be seen as a
+    // special kind approximation of the form (choice x. x = exact_value).
+    // Notice that the above term gets rewritten such that the choice function
+    // is eliminated.
+    for (unsigned i = 0, num = d_check_model_vars.size(); i < num; i++)
```
ahmed-irfan: we should use size_t instead of unsigned

src/theory/arith/nonlinear_extension.cpp
```diff
@@ -2508,6 +2492,61 @@ void NonlinearExtension::check(Theory::Effort e) {
...
+
+  // Did we internally determine a model exists? If so, we need to record some
+  // information in the theory engine's model class.
+  if (d_builtModel.get())
+  {
+    if (e == Theory::EFFORT_FULL)
+    {
+      // don't need to build the model yet
+      return;
+    }
```
ahmed-irfan: what about the other effort calls?

ajreynol: Effort here is either FULL or LAST_CALL, where we only need to build a model if we are at LAST_CALL. I'll make the condition here more inclusive (if the effort is less than LAST_CALL.

ahmed-irfan: I see

ahmed-irfan:
I left some comments.
Overall LGTM

ajreynol:
I've addressed the review.

PR #3415, opened 2019-10-28
Fix integer division rewrite

### LINKED ISSUE #3412: CVC4 gives UNSAT on SAT formula
Hi,
For this formula:
```
(declare-fun a () Int)
(declare-fun c () Int)
(assert (= (div a c) (- 1)))
(check-sat)
(get-model)
```
CVC4 reports `unsat`. While Z3 reports this model:
```
(define-fun a () Int
    1)
(define-fun c () Int
    (- 1))
```
If I feed this model to the formula, CVC4 still reports `unsat`. Is it expected?

OS: Ubuntu 18.04
Revision: 3af2dfe

4tXJ7f:
Fixes #3412. This commit fixes the issue introduced by #2929. When
rewriting division, the rewrite only added the unary minus for
`INTS_DIVISION` but not `INTS_DIVISION_TOTAL`. It also returned the node
without the unary minus (`nn` instead of `ret`).


4tXJ7f:
Good point, done.

PR #3417, opened 2019-10-29
Rename datatypes sygus solver

ajreynol:
This renames the datatypes sygus solver to `SygusExtension` to be in line with the standard naming scheme.


src/theory/datatypes/theory_datatypes.cpp
```diff
@@ -142,9 +140,9 @@ void TheoryDatatypes::check(Effort e) {
   d_addedLemma = false;
   
   if( e == EFFORT_LAST_CALL ){
-    Assert( d_sygus_sym_break );
+    Assert( d_sygusExtension!=nullptr );
```
HanielB: clang-format

PR #3418, opened 2019-10-29
Eagerly beta reduce during sygus to builtin term conversion

ajreynol:
The motivation for this PR is twofold:
(1) User-facing API functions involving SyGuS (`get-abduct` is an example of this) are currently returning SyGuS solutions with non-beta-reduced lambda applications in them, e.g. `((lambda ((x Int)) x) 0)`. This was reported by Daniel Larraz. This PR fixes this issue, and ensures the term we return through API calls indeed matches the input grammar.
(2) The performance of CVC4 should improve, since we do not construct intermediate unreduced lambda applications in most contexts.


src/theory/quantifiers/sygus/term_database_sygus.h
```diff
@@ -223,15 +223,20 @@ class TermDbSygus {
...
   Node mkGeneric(const Datatype& dt,
                  unsigned c,
                  std::map<TypeNode, int>& var_count,
-                 std::map<int, Node>& pre);
+                 std::map<int, Node>& pre,
+                 bool doBetaRed=true);
   /** same as above, but with empty var_count */
-  Node mkGeneric(const Datatype& dt, int c, std::map<int, Node>& pre);
+  Node mkGeneric(const Datatype& dt, int c, std::map<int, Node>& pre,
+                 bool doBetaRed=true);
```
HanielB: clang-format

src/theory/datatypes/theory_datatypes_utils.cpp
```diff
@@ -138,26 +139,41 @@ Node mkSygusTerm(const Datatype& dt,
...
+  Kind ok = op.getKind();
+  if (ok != BUILTIN)
   {
-    schildren.push_back(op);
+    if (ok==LAMBDA && doBetaReduction)
+    {
+      // do immediate beta reduction
+      std::vector< Node > vars{op[0].begin(),op[0].end()};
+      Assert(vars.size()==children.size());
+      Node ret = op[1].substitute(vars.begin(),vars.end(),children.begin(), children.end());
```
HanielB: shouldn't you be doing the capture avoiding substitution?

ajreynol: No, the terms we are using to substitute (those generated by sygus grammars) do not have quantifiers in them, nor does the body of op have quantifiers, so it is safe to use the normal substitution.

HanielB: I see. Could you add a comment over that line to clarify that?

ajreynol: Done.

PR #3419, opened 2019-10-29
Split some generic utilities from the non-linear extension

ajreynol:
This is work towards correcting some issues with model generation in the non-linear solver.

This moves some of the utility functions out of the solver so that the code is more flexible to further refactor.

There are no significant code changes in this commit apart from added documentation.

### DIFF
--- a/src/theory/arith/arith_utilities.cpp
+++ b/src/theory/arith/arith_utilities.cpp
@@ -14,14 +14,12 @@
 
 #include "arith_utilities.h"
 
+using namespace CVC4::kind;
+
 namespace CVC4 {
 namespace theory {
 namespace arith {
 
-/** Join kinds, where k1 and k2 are arithmetic relations returns an
- * arithmetic relation ret such that
- * if (a <k1> b) and (a <k2> b), then (a <ret> b).
- */
 Kind joinKinds(Kind k1, Kind k2)
 {
   if (k2 < k1)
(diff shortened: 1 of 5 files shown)


src/theory/arith/arith_utilities.h
```diff
@@ -302,6 +301,31 @@ inline Node mkPi()
   return NodeManager::currentNM()->mkNullaryOperator(
       NodeManager::currentNM()->realType(), kind::PI);
 }
+/** Join kinds, where k1 and k2 are arithmetic relations returns an
+ * arithmetic relation ret such that
+ * if (a <k1> b) and (a <k2> b), then (a <ret> b).
+ */
+Kind joinKinds(Kind k1, Kind k2);
```
ahmed-irfan: What it is doing -- What is the contract of this function?

ahmed-irfan: If I understood correctly from the code, Is it getting rid of redundant relation?

src/theory/arith/arith_utilities.cpp
```diff
@@ -0,0 +1,196 @@
...
+  Assert(isRelationOperator(k2));
+  if (k1 == EQUAL)
+  {
+    return k2;
+  }
+  else if (k1 == LT)
+  {
+    if (k2 == LEQ)
+    {
+      return k1;
```
ahmed-irfan: the behavior seems similar to joinKinds. I wonder if that is intentional

src/theory/arith/arith_utilities.cpp
```diff
@@ -0,0 +1,196 @@
...
+    }
+  }
+  else if (k1 == GT)
+  {
+    if (k2 == GEQ)
+    {
+      return k1;
```
ahmed-irfan: ditto

src/theory/arith/arith_utilities.h
ajreynol: This is used in the lemma schema here: https://github.com/CVC4/CVC4/blob/master/src/theory/arith/nonlinear_extension.h#L828.

Basically, if you have two relations between related terms, you need to know what the resulting relation is.

src/theory/arith/arith_utilities.cpp
ajreynol: The two are similar but trans applies to 3 terms whereas join applies to 2, for example `a = b ^ b >= c` imples` a >= c` whereas `a=b ^ a>=b` implies `a=b`.

ahmed-irfan: I see

PR #3432, opened 2019-11-04
Make check synth solution robust to auxiliary assertions

ajreynol:
This makes checkSynthSolution robust to auxiliary assertions. In particular, getSynthSolutions now tracks which synthesis conjecture is being solved. This is then used to determine which of our assertions in the SmtEngine are auxiliary assertions and which correspond to synthesis conjectures. This also makes us robust to the case where there are multiple simultaneous synthesis conjectures (although no known use case of this exists yet).

This is in preparation for sygus PBE + recursive functions.

This is going to have merge conflicts with #3306, let's merge that first.


src/smt/smt_engine.cpp
```diff
@@ -4862,33 +4873,39 @@ void SmtEngine::checkSynthSolution()
...
     // Apply any define-funs from the problem.
     {
-      unordered_map<Node, Node, NodeHashFunction> cache;
+      std::unordered_map<Node, Node, NodeHashFunction> cache;
       conj = d_private->expandDefinitions(conj, cache);
     }
     Notice() << "SmtEngine::checkSynthSolution(): -- expands to " << conj << endl;
     Trace("check-synth-sol") << "Expanded assertion " << conj << "\n";
-    if (conj.getKind() != kind::FORALL)
+    if (conjs.find(conj) == conjs.end())
```
HanielB: In retrospect `conj` was a poor choice for naming the local variable with the assertion. Could you rename it for "assertion" or something?

src/smt/smt_engine.cpp
```diff
@@ -4862,33 +4873,39 @@ void SmtEngine::checkSynthSolution()
     Node conj = Node::fromExpr(*i);
     // Apply any define-funs from the problem.
     {
-      unordered_map<Node, Node, NodeHashFunction> cache;
+      std::unordered_map<Node, Node, NodeHashFunction> cache;
       conj = d_private->expandDefinitions(conj, cache);
```
HanielB: We don't expand definitions when doing SyGuS, is that this is necessary? For all assertions?

src/smt/smt_engine.cpp
ajreynol: Sure.

src/smt/smt_engine.cpp
ajreynol: Expand definitions is always called on all assertions (https://github.com/CVC4/CVC4/blob/master/src/smt/smt_engine.cpp#L3241) regardless of what kind of check is run.

HanielB: So doing it again here is redundant?

ajreynol: No, it is run on assertions prior to giving them to the solver. However, we maintain the assertions in their original form because they are the ground truth of what the input was.  Thus, we redo it here again.  This is analogous to the code for `check-model`.

HanielB: I see. Didn't remember that. :)

PR #3433, opened 2019-11-04
Support for SyGuS PBE + recursive functions

ajreynol:
This enables the SyGuS solver to track recursive function definitions. This enables us to use a specialized evaluator to handle PBE inputs where the grammar includes recursive terminating functions. This feature is currently guarded by a new option `--sygus-rec-fun` but will be enabled by default in the near future.

The feature is fully functional, however, it uncovered bugs in the checkSynthSolution code. Thus, this PR is waiting on #3432.


test/regress/regress1/sygus/rec-fun-sygus.sy
```diff
@@ -0,0 +1,24 @@
+; EXPECT: unsat
+; COMMAND-LINE: --sygus-out=status --sygus-rec-fun --lang=sygus2
+
+; FIXME: does not pass check-synth due to not being robust to auxiliary assertions.
```
HanielB: I imagine you can remove this now?

src/theory/quantifiers/sygus/synth_conjecture.cpp
```diff
@@ -539,7 +539,8 @@ bool SynthConjecture::doCheck(std::vector<Node>& lems)
     return false;
   }
 
-  lem = Rewriter::rewrite(lem);
+  // simplify the lemma based on the term database sygus utility
+  lem = d_tds->rewriteNode(lem);
```
HanielB: Shouldn't you already be replacing other instances of lemma rewriting within the sygus module? There are several others that you have not rewritten.

src/theory/quantifiers/sygus/term_database_sygus.cpp
```diff
@@ -725,6 +726,28 @@ SygusTypeInfo& TermDbSygus::getTypeInfo(TypeNode tn)
...
+      // If recursive functions are enabled, then we use the recursive function
+      // evaluation utility.
+      Node fres = d_funDefEval->evaluate(res);
+      if (!fres.isNull())
+      {
+        return fres;
+      }
+      // It may have failed, in which case there are undefined symbols in res.
+      // In this case, we revert to the result of rewriting in the return
+      // statement below.
```
HanielB: I'm a bit confused by this. Why do you discard the result of performing evaluation if the whole term is not reduced to a constant? Wouldn't you have simplified at least part of it via the evaluations?

test/regress/regress1/sygus/rec-fun-sygus.sy
ajreynol: Yes, will do.

src/theory/quantifiers/sygus/synth_conjecture.cpp
ajreynol: That is a good point, let's do this in a separate PR once we resolve the new default behavior of the sygus solver. I think many places will change (everywhere that calls the extended rewriter).

src/theory/quantifiers/sygus/term_database_sygus.cpp
ajreynol: It's currently all or nothing. It isn't totally clear if partially evaluating has any use here: we are only using this in the context of evaluation. A partially evaluated node may actually be worse (think of a case that a function expands/terminates with some huge expression that we don't want to keep around).

src/theory/quantifiers/sygus/synth_conjecture.cpp
HanielB: OK

src/theory/quantifiers/sygus/term_database_sygus.cpp
HanielB: Yeah... I wonder. I think that the evaluator should be akin to the rewriter, in the sense that we always use it and in general it helps. If it does not it's an issue of the user. It might difficult for us to determine a priori whether things well help or not, so we assume they will (like we do with quantifiers).

PR #3434, opened 2019-11-04
Avoid non-well-founded sygus grammars

### LINKED ISSUE #3423: SyGuS "not well founded" error
The below SyGuS problem:
```
(set-logic ALL)
(synth-fun f ((x0 Bool)) Bool ((B Bool) (I Int))
	(
		(B Bool ((Variable Bool) (Constant Bool) (= I I) )) 
		(I Int ((Constant Int) (+ I I)))
	)
)
(constraint (= (f false) false))
(check-synth)
```
fails with the error:
```
Parse Error: dt_f_I is not well-founded
```

However, if `(+ I I)` is removed from the grammar:
```
(set-logic ALL)
(synth-fun f ((x0 Bool)) Bool ((B Bool) (I Int))
	(
		(B Bool ((Variable Bool) (Constant Bool) (= I I) )) 
		(I Int ((Constant Int)))
	)
)
(constraint (= (f false) false))
(check-synth)
```
CVC4 is able to synthesize:
```
unsat
(define-fun f ((x0 Bool)) Bool x0)
```

This might in some way be related to [https://github.com/CVC4/CVC4/pull/3301](https://github.com/CVC4/CVC4/pull/3301)

Edit: This is on 43ab3f4, the latest version of HEAD.

ajreynol:
This corrects an issue where we were generating non-well-founded sygus grammars for legal user inputs. This fixes #3423.

### DIFF
--- a/src/parser/smt2/Smt2.g
+++ b/src/parser/smt2/Smt2.g
@@ -1054,21 +1054,21 @@ sygusGrammar[CVC4::Type & ret,
       // grammar. This results in the error below.
       // We can also be in a case where the only rule specified was
       // (Constant T), in which case we have not yet added a constructor. We
-      // ensure an arbitrary constant is added in this case.
+      // ensure an arbitrary constant is added in this case. We additionally
+      // add a constant if the grammar allows it regardless of whether the
+      // datatype has other constructors, since this ensures the datatype is
+      // well-founded (see 3423).
+      if (aci)
+      {
+        Expr c = btt.mkGroundTerm();
+        PARSER_STATE->addSygusConstructorTerm(datatypes[i], c, ntsToUnres);
+      }
       if (datatypes[i].getNumConstructors() == 0)
       {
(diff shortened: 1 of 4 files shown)


src/parser/smt2/Smt2.g
```diff
@@ -1054,21 +1054,21 @@ sygusGrammar[CVC4::Type & ret,
...
+      // ensure an arbitrary constant is added in this case. We additionally
+      // add a constant if the grammar allows it regardless of whether the
+      // datatype has other constructors, since this ensures the datatype is
+      // well-founded (see 3423).
+      if (aci)
+      {
+        Expr c = btt.mkGroundTerm();
+        PARSER_STATE->addSygusConstructorTerm(datatypes[i], c, ntsToUnres);
+      }
       if (datatypes[i].getNumConstructors() == 0)
```
HanielB: this could be an `else if`

PR #3435, opened 2019-11-04
Fix ho extensionality in collect model info

### LINKED ISSUE #3422: Fatal Failure in CVC4 HO with finite-model-find
```
./cvc4 --uf-ho --finite-model-find tmp.smt2
Fatal failure within void CVC4::TheoryEngine::check(CVC4::theory::Theory::Effort) at /home/irfan/work/projects/CVC4/src/theory/theory_engine.cpp:678
Check failure

 d_curr_model->isBuiltSuccess()

Aborted (core dumped)
```
[tmp.zip](https://github.com/CVC4/CVC4/files/3795005/tmp.zip)

ajreynol:
This corrects a bug in the model construction for higher-order in which the witness to extensionality was using constants that were not fresh. This can lead to model inconsistencies when combing theories.

Fixes #3422.  The benchmark in that issue now times out.

### DIFF
--- a/src/theory/uf/ho_extension.cpp
+++ b/src/theory/uf/ho_extension.cpp
@@ -50,43 +50,49 @@ Node HoExtension::expandDefinition(Node node)
   return node;
 }
 
-Node HoExtension::getExtensionalityDeq(TNode deq)
+Node HoExtension::getExtensionalityDeq(TNode deq, bool isCached)
 {
   Assert(deq.getKind() == NOT && deq[0].getKind() == EQUAL);
   Assert(deq[0][0].getType().isFunction());
-  std::map<Node, Node>::iterator it = d_extensionality_deq.find(deq);
-  if (it == d_extensionality_deq.end())
+  if (isCached)
   {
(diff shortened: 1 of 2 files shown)


ajreynol:
The witness k for extensionality can be such that k occurs in lemmas; like Skolemizing existential quantifiers, we can assume we always use the same k.

For higher-order, the model construction at times delays witnessing extensionality until model construction, i.e. we don't need to witness extensionality for infinite types since we know we can always find new terms. However, if `k` was already witnessed during solving, we may have constraints on `k`.

In practice, this can occur if we witnessed extensionality for a function disequality, backtracked, and then found a model in a context where extensionality was not applied.

HanielB:
I see. This might be less confusing however if the cache was context-dependent?

ajreynol:
We don't want the cache to be context-dependent.  When solving, we want to use the same constants always, otherwise we will pollute the context with unecessary skolems.

HanielB:
OK. Could you then add your explanation above to the documentation of `getExtensionalityDeq`? In was really helpful for understanding the issue.

PR #3439, opened 2019-11-04
Refactor type matcher utility

ajreynol:
This is work towards a Node-level API for the Datatype class. This refactors a utility so that some of its interfaces do not use `Type`. It also documents the class and moves it to a more appropriate file name.

Further refactoring will be done to this class to eliminate uses of `Type` (which will lead to more substantial changes throughout the code).


src/expr/type_matcher.cpp
```diff
@@ -0,0 +1,129 @@
...
+      // recognize subtype relation
+      if (!tnn.isNull())
+      {
+        d_match[index] = tnn;
+        return true;
+      }
+      else
+      {
+        return false;
+      }
```
mpreiner: Remove `else` and just `return false;`

src/expr/type_matcher.cpp
```diff
@@ -0,0 +1,129 @@
...
+    }
+    else
+    {
+      d_match[i - d_types.begin()] = tn;
```
mpreiner:
```suggestion
      d_match[index] = tn;
```

PR #3442, opened 2019-11-05
Remove a few deprecated functions from Datatype

ajreynol:
There are several functions that are deprecated artifacts of very early design decisions on datatypes. This removes two of them.


ajreynol:
Actually, on second thought, this PR is unnecessary (it will be a consequence of removing the old API which we will do once and for all). I'm going to close this for now.

PR #3446, opened 2019-11-07
cmake: Disable C++ GNU extensions.

### LINKED ISSUE #971: Removing ## for '-std=c++11' (Bugzilla #752)
**Imported from Bugzilla**
Reporter: *Tim King*
Status: *NEW*
Severity: *normal*
Assigned to: *Caleb Donovick*
Component: *{Internals} Utilities*
Milestone: *1.6*
Version: *master*
Platform: *PC*
OS: *Linux*

Original attachment names and IDs:
- _[macro.cpp](http://cvc4.cs.stanford.edu/bugzilla_attachments/423_macro.cpp)_ (Example solution using macros to count VA_ARGS, ID 423)

On 2016-09-18 23:29:36 -0700, Tim King wrote:
> To support the '\-std\=c\+\+11' flag, CVC4 needs to remove usages of the GNU extension \#\# from variadiac macros in the pre\-processor. Some examples are from src/base/cvc4\_assert.h:
> 
> \#define InternalError(msg...)                                         \
>   throw ::CVC4::InternalErrorException(\_\_PRETTY\_FUNCTION\_\_, \_\_FILE\_\_, \
>                                        \_\_LINE\_\_, \#\#msg)
> 
> One plan is to split this into a 0\-argument version and a 1\+ argument version. Such as:
> 
> \#define InternalError()                                               \
>   throw ::CVC4::InternalErrorException(\_\_PRETTY\_FUNCTION\_\_, \_\_FILE\_\_, \
>                                        \_\_LINE\_\_)
> \#define InternalErrorWithMessage(msg...)                              \
>   throw ::CVC4::InternalErrorException(\_\_PRETTY\_FUNCTION\_\_, \_\_FILE\_\_, \
>                                        \_\_LINE\_\_, msg)

On 2017-04-03 16:00:14 -0700, Caleb Donovick wrote:
> This could be avoided by using \-std\=gnu\+\+11.  Seeing as \_\_PRETTY\_FUNCTION\_\_ is a gcc extension supported by clang it seems reasonable to use the \#\# to swallow the trailing comma (also supported by clang).  Regardless we should move from,
> 
> \#define InternalError(msg...)                                         \
>   throw ::CVC4::InternalErrorException(\_\_PRETTY\_FUNCTION\_\_, \_\_FILE\_\_, \
>                                        \_\_LINE\_\_, \#\#msg)
> 
> to
> 
> \#define InternalError(...)                                            \
>   throw ::CVC4::InternalErrorException(\_\_PRETTY\_FUNCTION\_\_, \_\_FILE\_\_, \
>                                        \_\_LINE\_\_, \#\#\_\_VA\_ARGS\_\_) 
> 
> 
> To move to fully portable code would require the following changes:
> 
> Move from \_\_PRETTY\_FUNCTION\_\_ to a portable solution e.g.,
> \#ifdef compiler and use \_\_PRETTY\_FUNCTION\_\_ for GCC/clang, \_\_FUNCSIG\_\_ for MSVC and \_\_func\_\_ otherwise.  Note \_\_func\_\_ is part of C\+\+11 standard (Â§8.4.1) but provides less detail than \_\_PRETTY\_FUNCTION\_\_ / \_\_FUNCSIG\_\_
> 
> As for the variadic macro issue:
> As proposed by Tim, we can have a 0\-argument version and a 1\+ argument version however this is clunky in my opinion.
> 
> Macro magic can be used to count \_\_VA\_ARGS\_\_ up to some set number of arguments (more arguments than are support will break the macro).  Further, such solutions tend to be fragile and have cases for which they are incorrect.
> 
> Perhaps the best solution is to use templates tricks or constexpr to count the number of arguments and dispatch accordingly.  However, these solutions require  C\+\+11.

On 2017-04-03 17:33:53 -0700, Caleb Donovick wrote:
> Created attachment 423
> Example solution using macros to count VA\_ARGS
> 
> PRINT\_MACRO\_CHOOSER will fail if its passed more than 15 arguments as it will return its 16th argument instead of PRINT\_ARG.  This could be trivially extend to any arbitrary number of arguments, however, there will always be some maximum. It may break if given macros as arguments as they might not be properly expanded.  The exact rules for macro expansion always allude me so I can't give a specific case for which it fail.

mpreiner:
This commit removes the last variadic macro (was unused). All other variadic macros have been eliminated in #3366.

Fixes #971.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -26,6 +26,7 @@ endif()
 set(CMAKE_MODULE_PATH ${PROJECT_SOURCE_DIR}/cmake)
 set(CMAKE_C_STANDARD 99)
 set(CMAKE_CXX_STANDARD 11)
+set(CMAKE_CXX_EXTENSIONS OFF)
 
 # Generate compile_commands.json, which can be used for various code completion
 # plugins.
--- a/src/prop/cnf_stream.h
+++ b/src/prop/cnf_stream.h
@@ -88,25 +88,6 @@ class CnfStream {
   /** Pointer to the proof corresponding to this CnfStream */
   CnfProof* d_cnfProof;
 
-  /**
-   * How many literals were already mapped at the top-level when we
-   * tried to convertAndAssert() something.  This
-   * achieves early detection of units and leads to fewer
-   * clauses.  It's motivated by the following pattern:
-   *
-   *   ASSERT  BIG FORMULA => x
-   *   (and then later...)
-   *   ASSERT  BIG FORMULA
-   *
-   * With the first assert, BIG FORMULA is clausified, and a literal
-   * is assigned for the top level so that the final clause for the
-   * implication is "lit => x".  But without "fortunate literal
-   * detection," when BIG FORMULA is later asserted, it is clausified
-   * separately, and "lit" is never asserted as a unit clause.
-   */
-  // KEEP_STATISTIC(IntStat, d_fortunateLiterals,
-  //               "prop::CnfStream::fortunateLiterals", 0);
-
   /** Remove nots from the node */
   TNode stripNot(TNode node) {
     while (node.getKind() == kind::NOT) {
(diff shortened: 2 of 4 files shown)


mpreiner:
Adding do-not-merge until @4tXJ7f was able to review.

Issue #2153, opened 2018-07-05
model doesn't satisfy


akgperson:
Error message:
```
(error "Internal error detected
void CVC4::TheoryEngine::checkTheoryAssertionsWithModel(bool)
../../../../src/theory/theory_engine.cpp:2275:

  
THEORY_ARITH has an asserted fact that the model doesn't satisfy.
The fact: (= kf_10 (+ (* (- 1.0) (* r0 r15)) (* (/ 2000 681247) (* r0 r0 r15)) (* (/ (- 1000) 681247) (* r0 r5 r15)) (* (/ (- 1000) 681247) (* r0 r15 r15)) (* (/ 1000 681247) (* r0 r15 r16))))
Model value: false
")

```
Input File:
```
(set-logic ALL)
(set-option :incremental true)
(declare-const v0 Bool)
(declare-const v1 Bool)
(declare-const v2 Bool)
(declare-const v3 Bool)
(declare-const v4 Bool)
(declare-const v5 Bool)
(declare-const i1 Int)
(declare-const i5 Int)
(declare-const i6 Int)
(declare-const i7 Int)
(declare-const r0 Real)
(declare-const r2 Real)
(declare-const r3 Real)
(declare-const r4 Real)
(declare-const r5 Real)
(declare-const r7 Real)
(declare-const r10 Real)
(declare-const r11 Real)
(declare-const r13 Real)
(declare-const r14 Real)
(declare-sort S0 0)
(declare-const v6 Bool)
(declare-const v7 Bool)
(declare-const i10 Int)
(declare-const r15 Real)
(declare-const v8 Bool)
(declare-const i11 Int)
(declare-const i12 Int)
(declare-const v9 Bool)
(declare-const v10 Bool)
(declare-const r16 Real)
(declare-const v11 Bool)
(declare-const r17 Real)
(declare-const v12 Bool)
(declare-const v13 Bool)
(declare-const r18 Real)
(declare-const v14 Bool)
(declare-const i13 Int)
(declare-const r19 Real)
(declare-const r20 Real)
(declare-const v15 Bool)
(declare-const v16 Bool)
(declare-const v17 Bool)
(declare-const r21 Real)
(declare-sort S1 0)
(declare-const v18 Bool)
(declare-const r22 Real)
(declare-const v19 Bool)
(declare-const r23 Real)
(declare-const v20 Bool)
(declare-const S1-0 S1)
(declare-const v21 Bool)
(declare-const r24 Real)
[...]

PR #3452, opened 2019-11-11
Refactor non-linear extension for model-based refinement

ajreynol:
This PR does some additional refactoring of non-linear extension and nl model classes to be more flexible, in preparation for addressing #2153.

There are no major behavior changes in this PR; the biggest changes were to move a block of code to its own function `(NonlinearExtension::modelBasedRefinement)` and to make the function `NlModel::getModelRepair` more flexible.


src/theory/arith/nl_model.cpp
```diff
@@ -1225,12 +1203,37 @@ bool NlModel::getApproximateSqrt(Node c, Node& l, Node& u, unsigned iter) const
   return true;
 }
 
-void NlModel::recordApproximations()
+void NlModel::printModelValue(const char* c, Node n, unsigned prec) const
+{
+  if (Trace.isOn(c))
+  {
+    Trace(c) << "  " << n << " -> ";
+    for (unsigned i = 0; i < 2; i++)
```
ahmed-irfan: running the loop in reverse order will make the code more readable.
```
for (unsigned i = 1; i >= 0; --i)
{
   std::map<Node, Node>::const_iterator it = d_mv[i].find(n);
   Assert(it != d_mv[i].end());
```

PR #3454, opened 2019-11-11
Distinguish unknown status for model printing

ajreynol:
This distinguishes whether a model corresponds to a "sat" or "unknown" query. This is required to be compliant for TPTP printing, which does not expect models to be printed unless the solver is sure there is a model.


src/printer/tptp/tptp_printer.cpp
```diff
@@ -55,11 +55,22 @@ void TptpPrinter::toStream(std::ostream& out, const CommandStatus* s) const
 
 void TptpPrinter::toStream(std::ostream& out, const Model& m) const
 {
-  out << "% SZS output start FiniteModel for " << m.getInputName() << endl;
+  std::stringstream statusName;
```
4tXJ7f: Nit: This doesn't need to be an `std::stringstream`, it can just be an `std::string`. E.g. `std::string statusName(m.isKnownSat() ? "FiniteModel" : "CandidateFiniteModel")`.

PR #3462, opened 2019-11-15
Introduce the Node-level Datatypes API

ajreynol:
This adds classes corresponding to the Node-level Datatype API "DType", which is a specification for a datatype type.  It does not enable the use of this layer yet. A followup PR will update the Expr-level Datatype to use the Node-level code, which is currently verified to be functional on this branch: https://github.com/ajreynol/CVC4/tree/dtype-integrate. Futher PRs will make the internal (Node-level) code forgo the use of the Expr-layer datatype, which will then enable the Expr-layer to be replaced by the Term-layer datatype.

Most of the documentation for the methods in DType/DTypeConstructor/DTypeSelector was copied from Datatype/DatatypeConstructor/DatatypeConstructorArg.


src/expr/dtype_cons.cpp
```diff
@@ -0,0 +1,669 @@
...
+using namespace CVC4::theory;
+
+namespace CVC4 {
+
+DTypeConstructor::DTypeConstructor(std::string name)
+    :  // We don't want to introduce a new data member, because eventually
+       // we're going to be a constant stuffed inside a node.  So we stow
+       // the tester name away inside the constructor name until
+       // resolution.
+      d_name(name + '\0' + "is_" + name),  // default tester name is "is_FOO"
```
mpreiner: Isn't there a cleaner way? This is quite hacky.

src/expr/dtype_cons.cpp
```diff
@@ -0,0 +1,669 @@
...
+}
+
+DTypeConstructor::DTypeConstructor(std::string name,
+                                   std::string tester,
+                                   unsigned weight)
+    :  // We don't want to introduce a new data member, because eventually
+       // we're going to be a constant stuffed inside a node.  So we stow
+       // the tester name away inside the constructor name until
+       // resolution.
+      d_name(name + '\0' + tester),
```
mpreiner: Ditto.

src/expr/dtype_cons.cpp
```diff
@@ -0,0 +1,669 @@
...
+    {
+      return (int)its->second;
+    }
+  }
+  else
+  {
+    unsigned sindex = DType::indexOf(sel);
+    if (getNumArgs() > sindex && d_args[sindex]->getSelector() == sel)
+    {
+      return (int)sindex;
```
mpreiner:
```suggestion
      return static_cast<int>(sindex);
```

src/expr/dtype_cons.cpp
```diff
@@ -0,0 +1,669 @@
...
+  size_t index = 0;
+  std::vector<TypeNode> argTypes;
+  for (std::shared_ptr<DTypeSelector> arg : d_args)
+  {
+    std::string argName = arg->d_name;
+    TypeNode range;
+    if (arg->d_selector.isNull())
+    {
+      // the unresolved type wasn't created here; do name resolution
+      std::string typeName = argName.substr(argName.find('\0') + 1);
```
mpreiner: If there is no other way than storing additional information in the string, at least create helper functions do extract the corresponding part of the string.

src/expr/dtype.h
```diff
@@ -0,0 +1,600 @@
...
+#include "expr/dtype_cons.h"
+#include "expr/dtype_selector.h"
+#include "expr/node.h"
+#include "expr/node_manager_attributes.h"
+#include "expr/type_node.h"
+
+namespace CVC4 {
+
+// ----------------------- datatype attributes
+/** TODO */
```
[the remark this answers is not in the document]
ajreynol: Forgot some documentation, will add.

src/expr/dtype.h
```diff
@@ -0,0 +1,600 @@
...
+   * for parametric datatypes, where t is an instantiated
+   * parametric datatype type whose datatype is this class.
+   */
+  Cardinality getCardinality(TypeNode t) const;
+  Cardinality getCardinality() const;
+
+  /**
+   * Return true iff this DType has finite cardinality. If the
+   * datatype is not well-founded, this method returns false. The
+   * DType must be resolved or an exception is thrown.
```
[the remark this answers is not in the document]
ajreynol: Ok, changing to "or an assertion is violated".

src/expr/dtype_cons.cpp
ajreynol: I agree this is hacky. I can make some attempts to clean this up.

ajreynol:
@mpreiner , I managed to remove the tester name hacks from DTypeConstructor. It turns out that this is only important for parsing and thus shouldn't be in the Node-level API.   This change was confirmed to work on: https://github.com/ajreynol/CVC4/tree/dtype-integrate.

src/expr/dtype_cons.cpp
```diff
@@ -0,0 +1,647 @@
...
+    else
+    {
+      ctype = d_constructor.getType();
+    }
+    Assert(ctype.isConstructor());
+    Assert(ctype.getNumChildren() - 1 == getNumArgs());
+    // compute the shared selectors
+    const DType& dt = DType::datatypeOf(d_constructor);
+    std::map<TypeNode, unsigned> counter;
+    for (unsigned j = 0; j < ctype.getNumChildren() - 1; j++)
```
4tXJ7f:
```suggestion
    for (size_t j = 0; j < ctype.getNumChildren() - 1; j++)
```

src/expr/dtype_cons.cpp
```diff
@@ -0,0 +1,647 @@
...
+  std::vector<TypeNode> children;
+  for (TypeNode::const_iterator i = range.begin(), iend = range.end();
+       i != iend;
+       ++i)
+  {
+    origChildren.push_back((*i));
+    children.push_back(
+        doParametricSubstitution((*i), paramTypes, paramReplacements));
+  }
+  for (unsigned i = 0; i < paramTypes.size(); ++i)
```
4tXJ7f:
```suggestion
  for (size_t i = 0; i < paramTypes.size(); ++i)
```

src/expr/dtype_cons.h
```diff
@@ -0,0 +1,320 @@
...
+   */
+  int getSelectorIndexInternal(Node sel) const;
+
+  /** involves external type
+   *
+   * Get whether this constructor has a subfield
+   * in any constructor that is not a datatype type.
+   */
+  bool involvesExternalType() const;
+  /** involves external type
```
4tXJ7f:
```suggestion
  /** involves uninterpreted type
```

src/expr/dtype.h
```diff
@@ -0,0 +1,620 @@
...
+  /** whether this datatype has been resolved */
+  bool d_resolved;
+  /** self type */
+  mutable TypeNode d_self;
+  /** cache for involves external type */
+  bool d_involvesExt;
+  /** cache for involves uninterpreted type */
+  bool d_involvesUt;
+  /** the builtin type that this sygus type encodes */
+  TypeNode d_sygus_type;
```
4tXJ7f: Use camelCase.

src/expr/dtype.h
```diff
@@ -0,0 +1,620 @@
...
+  /** the variable list for the sygus function to synthesize */
+  Node d_sygus_bvl;
```
4tXJ7f: Use camelCase.

src/expr/dtype.h
```diff
@@ -0,0 +1,620 @@
...
+  /** whether all constants are allowed as solutions */
+  bool d_sygus_allow_const;
```
4tXJ7f: Use camelCase.

src/expr/dtype.h
```diff
@@ -0,0 +1,620 @@
...
+  /** whether all terms are allowed as solutions */
+  bool d_sygus_allow_all;
```
4tXJ7f: Use camelCase.

src/expr/dtype.h
```diff
@@ -0,0 +1,620 @@
...
+
+  /** is this type a recursive singleton type?
+   * The range of this map stores
+   * 0 if the field has not been computed,
+   * 1 if this datatype is a recursive singleton type,
+   * -1 if this datatype is not a recursive singleton type.
+   * For definition of (co)recursive singleton, see
+   * Section 2 of Reynolds et al. CADE 2015.
+   */
+  mutable std::map<TypeNode, int> d_card_rec_singleton;
```
4tXJ7f: Use camelCase.

src/expr/dtype.h
```diff
@@ -0,0 +1,620 @@
...
+  /** if d_card_rec_singleton is true,
+   * This datatype has infinite cardinality if at least one of the
+   * following uninterpreted sorts having cardinality > 1.
+   */
+  mutable std::map<TypeNode, std::vector<TypeNode> > d_card_u_assume;
```
4tXJ7f: Use camelCase.

src/expr/dtype.h
```diff
@@ -0,0 +1,620 @@
...
+  /** cache of whether this datatype is well-founded */
+  mutable int d_well_founded;
```
4tXJ7f: Use camelCase.

src/expr/dtype.h
```diff
@@ -0,0 +1,620 @@
...
+  /** cache of ground term for this datatype */
+  mutable std::map<TypeNode, Node> d_ground_term;
```
4tXJ7f: Use camelCase.

src/expr/dtype.h
```diff
@@ -0,0 +1,620 @@
...
+  /** cache of ground values for this datatype */
+  mutable std::map<TypeNode, Node> d_ground_value;
```
4tXJ7f: Use camelCase.

src/expr/dtype.h
```diff
@@ -0,0 +1,620 @@
...
+{
+  size_t operator()(const DType& dt) const
+  {
+    return std::hash<std::string>()(dt.getName());
+  }
+  size_t operator()(const DType* dt) const
+  {
+    return std::hash<std::string>()(dt->getName());
+  }
+  size_t operator()(const DTypeConstructor& dtc) const
```
4tXJ7f: Wouldn't it make more sense to define the hash functions for `DTypeConstructor` and `DTypeConstructor*` in a separate struct that lives in `dtype_cons.h`?

src/expr/dtype.h
```diff
@@ -0,0 +1,620 @@
...
+  }
+  bool operator<=(const DTypeIndexConstant& uc) const
+  {
+    return d_index <= uc.d_index;
+  }
+  bool operator>(const DTypeIndexConstant& uc) const { return !(*this <= uc); }
+  bool operator>=(const DTypeIndexConstant& uc) const { return !(*this < uc); }
+
+ private:
+  const unsigned d_index;
```
4tXJ7f:
```suggestion
  const size_t d_index;
```

src/expr/dtype.h
```diff
@@ -0,0 +1,620 @@
...
+   * allow_const : whether all constants are (implicitly) allowed by the
+   * datatype
+   * allow_all : whether all terms are (implicitly) allowed by the datatype
+   *
+   * Notice that allow_const/allow_all do not reflect the constructors
+   * for this datatype, and instead are used solely for relaxing constraints
+   * when doing solution reconstruction (Figure 5 of Reynolds et al.
+   * CAV 2015).
+   */
+  void setSygus(TypeNode st, Node bvl, bool allow_const, bool allow_all);
```
4tXJ7f: Use camelCase for arguments.

src/expr/dtype.h
```diff
@@ -0,0 +1,620 @@
...
+  /** Get the name of this DType. */
+  std::string getName() const;
+
+  /** Get the number of constructors (so far) for this DType. */
+  size_t getNumConstructors() const;
+
+  /** Is this datatype parametric? */
+  bool isParametric() const;
+
+  /** Get the nubmer of type parameters */
```
4tXJ7f:
```suggestion
  /** Get the number of type parameters */
```

src/expr/dtype.h
```diff
@@ -0,0 +1,620 @@
...
+   * for parametric datatypes, where t is an instantiated
+   * parametric datatype type whose datatype is this class.
+   */
+  bool isFinite(TypeNode t) const;
+  bool isFinite() const;
+
+  /**
+   * Return true iff this  DType is finite (all constructors are
+   * finite, i.e., there  are finitely  many ground terms) under the
+   * assumption unintepreted sorts are finite. If the
```
4tXJ7f:
```suggestion
   * assumption that uninterpreted sorts are finite. If the
```

src/expr/dtype.h
4tXJ7f: Comment on the different meanings of -1, 0, 1.

src/expr/dtype.cpp
```diff
@@ -0,0 +1,743 @@
...
+}
+
+DType::~DType() {}
+
+std::string DType::getName() const { return d_name; }
+size_t DType::getNumConstructors() const { return d_constructors.size(); }
+
+bool DType::isParametric() const { return d_params.size() > 0; }
+size_t DType::getNumParameters() const { return d_params.size(); }
+TypeNode DType::getParameter(unsigned int i) const
```
4tXJ7f:
```suggestion
TypeNode DType::getParameter(size_t i) const
```

src/expr/dtype.cpp
```diff
@@ -0,0 +1,743 @@
...
+  if (isSygus())
+  {
+    // all datatype constructors should be sygus and have sygus operators whose
+    // free variables are subsets of sygus bound var list.
+    std::unordered_set<Node, NodeHashFunction> svs;
+    for (const Node& sv : d_sygus_bvl)
+    {
+      svs.insert(sv);
+    }
+    for (unsigned i = 0, ncons = d_constructors.size(); i < ncons; i++)
```
4tXJ7f:
```suggestion
    for (size_t i = 0, ncons = d_constructors.size(); i < ncons; i++)
```

src/expr/dtype.cpp
```diff
@@ -0,0 +1,743 @@
...
+          << "Sygus datatype contains a non-sygus constructor";
+      std::unordered_set<Node, NodeHashFunction> fvs;
+      expr::getFreeVariables(sop, fvs);
+      for (const Node& v : fvs)
+      {
+        if (svs.find(v) == svs.end())
+        {
+          // Print a warning and return false, indicating we should abort,
+          // since this datatype is not well formed.
+          Warning() << "Sygus constructor has an operator with a free variable "
```
4tXJ7f: Shouldn't this be an error?

src/expr/dtype.cpp
```diff
@@ -0,0 +1,743 @@
...
+  return d_card_u_assume[t].size();
+}
+
+unsigned DType::getNumRecursiveSingletonArgTypes() const
+{
+  Assert(!isParametric());
+  return getNumRecursiveSingletonArgTypes(d_self);
+}
+
+TypeNode DType::getRecursiveSingletonArgType(TypeNode t, unsigned i) const
```
4tXJ7f:
```suggestion
TypeNode DType::getRecursiveSingletonArgType(TypeNode t, size_t i) const
```

src/expr/dtype.cpp
```diff
@@ -0,0 +1,743 @@
...
+{
+  Assert(d_card_rec_singleton.find(t) != d_card_rec_singleton.end());
+  Assert(isRecursiveSingleton(t));
+  return d_card_u_assume[t][i];
+}
+
+TypeNode DType::getRecursiveSingletonArgType(unsigned i) const
```
4tXJ7f:
```suggestion
TypeNode DType::getRecursiveSingletonArgType(size_t i) const
```

src/expr/dtype.cpp
```diff
@@ -0,0 +1,743 @@
...
+  return d_self.instantiateParametricDatatype(params);
+}
+
+const DTypeConstructor& DType::operator[](size_t index) const
+{
+  Assert(index < getNumConstructors());
+  return *d_constructors[index];
+}
+
+Node DType::getSharedSelector(TypeNode dtt, TypeNode t, unsigned index) const
```
4tXJ7f:
```suggestion
Node DType::getSharedSelector(TypeNode dtt, TypeNode t, size_t index) const
```

src/expr/dtype_cons.h
```diff
@@ -0,0 +1,320 @@
...
+   * We store this information for (possibly multiple)
+   * datatype types dtt, since this constructor may be
+   * for a parametric datatype, where dtt is an instantiated
+   * parametric datatype.
+   */
+  mutable std::map<TypeNode, std::vector<Node> > d_shared_selectors;
+  /** for each type, a cache mapping from shared selectors to
+   * its argument index for this constructor.
+   */
+  mutable std::map<TypeNode, std::map<Node, unsigned> > d_shared_selector_index;
```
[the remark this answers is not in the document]
ajreynol: I do agree that these "mutable" annotations are annoying.
However, removing the is not straightforward to do.  The problem is that parametric datatypes may involve infinitely many unique calls to computing shared selectors (think (List T) for (List Int), (List Bool), (List Real)...).  This means that we can't just make a single call during resolve() and be finished.
As a result, making these immutable would require a number of the core functions (e.g. getSelectorInternal) non-const.

src/expr/dtype_selector.h
```diff
@@ -0,0 +1,98 @@
...
+  /** the name of this selector */
+  std::string d_name;
+  /** the selector expression */
+  Node d_selector;
+  /**
+   * The constructor associated with this selector. This field is initialized
+   * by the constructor of this selector during a call to
+   * DTypeConstructor::resolve.
+   */
+  Node d_constructor;
```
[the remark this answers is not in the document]
ajreynol: The problem with this is it makes the code even more confusing in my opinion, since it gives the impression that `setConstructor()` is part of the interface for this class.  Instead, the interaction documented here pertains to how this class is initialized; `setConstructor()` should definitely never be called in any other context.

src/expr/dtype_cons.cpp
```diff
@@ -0,0 +1,647 @@
...
+  return c;
+}
+
+bool DTypeConstructor::isFinite(TypeNode t) const
+{
+  Assert(isResolved());
+
+  TNode self = d_constructor;
+  // is this already in the cache ?
+  if (self.getAttribute(DTypeFiniteComputedAttr()))
```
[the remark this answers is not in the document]
ajreynol: No, this does not work for Boolean attributes. This code pattern is used in several places, see e.g. https://github.com/CVC4/CVC4/blob/master/src/expr/type_node.cpp#L76.

src/expr/dtype_cons.cpp
```diff
@@ -0,0 +1,647 @@
...
+  self.setAttribute(DTypeFiniteAttr(), true);
+  return true;
+}
+
+bool DTypeConstructor::isInterpretedFinite(TypeNode t) const
+{
+  Assert(isResolved());
+  TNode self = d_constructor;
+  // is this already in the cache ?
+  if (self.getAttribute(DTypeUFiniteComputedAttr()))
```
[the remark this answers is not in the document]
ajreynol: See above.

src/expr/dtype.h
ajreynol: Sure, this was copied over from datatype.h. I'm not sure how important these hash functions are anyways. I don't think we ever hash on datatypes.

src/expr/dtype.cpp
ajreynol: The error should be propagated upwards to the user level so we can give a proper (pretty) exception.

src/expr/dtype_cons.cpp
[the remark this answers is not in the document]
4tXJ7f: Thanks, that's good to know.

ajreynol:
I've addressed @4tXJ7f's review in my latest commit.

src/expr/dtype.cpp
4tXJ7f: Hmm, so why print a warning at all in that case?

src/expr/dtype.h
4tXJ7f: It would be great if we could change this to a `size_t` before merging.

src/expr/dtype_cons.h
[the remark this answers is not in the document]
4tXJ7f: I see, thanks for the explanation.

ajreynol:
I suppose the warning can be deleted.

Issue #3461, opened 2019-11-15
Does the SyGuS solver support custom datatypes and records?


tjknoth:
```
(set-logic ALL_SUPPORTED)

(declare-datatype Doc ((D (owner Int) (body Int))))

(declare-datatype Policy 
  ((p (principal Int)) 
   (por (left Policy) (right Policy))))

(synth-fun mkPolicy ((d Doc)) Policy
  ((Start Policy (Q))
   (Q Policy ((p 0) (p 1) (por Q Q))))
)

(check-synth)
```

Should something like this synthesize? I see a "operator does not have function type" error. It seems like there might be some issues around synthesizing functions over custom data types. In the future we would like to be able to include the record selectors in the grammar as well.

ajreynol:
We have recently improved our support for synthesis of datatype expressions in CVC4's SyGuS solver, so yes this should be handled.

There is an issue with the sygus version 1 parser that is causing the spurious error.  I will submit a fix for this soon.

As a workaround, you can use the sygus version 2 format using command line option `--lang=sygus2`. Notice this format is planned to be the default format accepted by cvc4 in its next stable release.

Here is the version of your benchmark in sygus version 2 format (almost identical except a "predeclaration" is added prior to specifying grammar):

```
(set-logic ALL_SUPPORTED)

(declare-datatype Doc ((D (owner Int) (body Int))))

(declare-datatype Policy 
  ((p (principal Int)) 
   (por (left Policy) (right Policy))))

(synth-fun mkPolicy ((d Doc)) Policy
  ((Start Policy) (Q Policy))
  ((Start Policy (Q))
   (Q Policy ((p 0) (p 1) (por Q Q))))
)

(check-synth)
```

CVC4 answers quickly for this benchmark using `--lang=sygus2`:
```
unsat
(define-fun mkPolicy ((d Doc)) Policy (p 0))
```

For more details on the new format, see https://sygus.org/language/.

tjknoth:
Perfect, thanks!

PR #3465, opened 2019-11-15
Introduce SyGuS datatype API

ajreynol:
This introduces a "SyGuS datatype" class which is a wrapper on top of the Expr-layer Datatype class.  This class is responsible for initializing the sygus constructors and sygus-specific information in a datatype. This is a first step towards a future SyGuS API.

Currently, only SyGuS grammar normalization uses this class, although the Parser, the default grammar constructor, and other custom SyGuS algorithms within CVC4 will be updated to use this interface. 

This PR can be merged independently of #3461.  This PR will make it easier to swap out the Expr-level Datatype implementation when the Node-level Datatypes API is ready.

### DIFF
--- a/src/expr/sygus_datatype.cpp
+++ b/src/expr/sygus_datatype.cpp
@@ -20,21 +20,15 @@ using namespace CVC4::kind;
 
 namespace CVC4 {
 
-SygusDatatype::SygusDatatype(const std::string& name)
-    : d_dt(Datatype(name))
-{
-}
+SygusDatatype::SygusDatatype(const std::string& name) : d_dt(Datatype(name)) {}
 
-std::string SygusDatatype::getName() const
-{
-  return d_dt.getName();
-}
+std::string SygusDatatype::getName() const { return d_dt.getName(); }
 
 void SygusDatatype::addConstructor(Node op,
-                                       const std::string&  name,
-                                       std::shared_ptr<SygusPrintCallback> spc,
-                                       int weight,
-                                       const std::vector<TypeNode>& consTypes)
+                                   const std::string& name,
+                                   std::shared_ptr<SygusPrintCallback> spc,
+                                   int weight,
+                                   const std::vector<TypeNode>& consTypes)
 {
   d_ops.push_back(op);
   d_cons_names.push_back(std::string(name));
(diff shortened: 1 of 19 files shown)


src/expr/sygus_datatype.h
```diff
@@ -0,0 +1,114 @@
...
+typedef expr::Attribute<SygusAnyConstAttributeId, bool> SygusAnyConstAttribute;
+
+/**
+ * Keeps the necessary information for initializing a sygus datatype, which
+ * includes the operators, names, print callbacks and list of argument types
+ * for each constructor.
+ *
+ * It also maintains a datatype to represent the structure of the type node.
+ *
+ * Notice that is class is only responsible for setting SyGuS-related
```
HanielB: is class -> this class

src/expr/sygus_datatype.h
```diff
@@ -0,0 +1,114 @@
...
+ * information regarding the datatype. It is still required that the user
```
HanielB: that the user -> from the user

src/expr/sygus_datatype.cpp
```diff
@@ -0,0 +1,97 @@
...
+                                   const std::vector<TypeNode>& consTypes,
+                                   bool prepend)
+{
+  if (prepend)
+  {
+    d_ops.push_back(op);
+    d_cons_names.push_back(std::string(name));
+    d_pc.push_back(spc);
+    d_weight.push_back(weight);
+    d_consArgs.push_back(consTypes);
```
HanielB: this is not prepending :)

src/theory/quantifiers/sygus/sygus_grammar_norm.cpp
```diff
@@ -155,98 +156,10 @@ Node SygusGrammarNorm::TypeObject::eliminatePartialOperators(Node n)
...
-  sygus_norm->d_dt_all.push_back(d_dt);
-  sygus_norm->d_unres_t_all.insert(d_unres_tn.toType());
-  Trace("sygus-grammar-normalize") << "---------------------------------\n";
-}
-
 void SygusGrammarNorm::TransfDrop::buildType(SygusGrammarNorm* sygus_norm,
-                                             TypeObject& to,
+                                             TypeNode orig,
+                                             TypeNode unres,
+                                             SygusDatatype& to,
```
HanielB: change the variable name to reflect the new type (this and other occurrences)

ajreynol: After my last commit, this no longer applies (grammar norm still operates on TypeObjects).

PR #3473, opened 2019-11-17
Improve interface for sygus datatype, fix utilities

ajreynol:
This is in preparation for customizing grammars with e.g. coefficients for LIA.

It improves the interface for sygus datatype and fixes a few utilities that were not robust to grammars involving "any constant" constructors.


ajreynol:
Need to make a minor additional change, marking do not merge for now.

PR #3478, opened 2019-11-19
Fix reduction of `sqrt`

### LINKED ISSUE #3475: CVC4 check unsat cores faliure on NRA formula
Hi,
For this formula,
```
(set-logic ALL)
(declare-fun x () Real)
(assert (< x 0))
(assert (not (= (/ (sqrt x) (sqrt x)) x)))
(check-sat)
```
CVC4 reports `unsat`.  
This formula should be `sat` in real arithmathic, since `(/ (sqrt x) (sqrt x))` can be rewrited to `1`.
If I enable `check-unsat-cores`, CVC4 will throw out a fatal failure:
```
Fatal failure within void CVC4::SmtEngine::checkUnsatCore() at 
/home/CVC4/src/smt/smt_engine.cpp:4613
Internal error detectedSmtEngine::checkUnsatCore(): produced core was satisfiable.
```

OS: Ubuntu 18.04
Revision: 17f0468

4tXJ7f:
Fixes #3475. This commit fixes two issues related to the reduction of
`sqrt`. Previously, we reduced `sqrt(x)` to `choice y. y = x * x`. The
problem is that this reduction makes a problem `unsat` as soon as there
is a square root of a negative number and it does not maintain the
invariant that `x = y => sqrt(x) = sqrt(y)` (i.e. that `sqrt` behaves
like a function). This commit changes the reduction of `sqrt(x)` to
`choice y. ite(x >= 0.0, y * y = x ^ Uf(x), Uf(x))`, meaning that we
interpret the square root of negative numbers as an undefined value
(similar to divisions by zero in the theories of integer/real
arithmetic).

### DIFF
--- a/src/theory/arith/theory_arith_private.h
+++ b/src/theory/arith/theory_arith_private.h
@@ -828,11 +828,16 @@ private:
 
   Statistics d_statistics;
 
-  enum ArithSkolemId
+  enum class ArithSkolemId
   {
-    arith_skolem_div_by_zero,
-    arith_skolem_int_div_by_zero,
-    arith_skolem_mod_by_zero,
+    /* an uninterpreted function f s.t. f(x) = x / 0.0 (real division) */
+    DIV_BY_ZERO,
+    /* an uninterpreted function f s.t. f(x) = x / 0 (integer division) */
+    INT_DIV_BY_ZERO,
+    /* an uninterpreted function f s.t. f(x) = x mod 0 */
+    MOD_BY_ZERO,
+    /* an uninterpreted function f s.t. f(x) = sqrt(x) */
+    SQRT,
   };
 
   /**
--- a/src/theory/arith/theory_arith_private.cpp
+++ b/src/theory/arith/theory_arith_private.cpp
@@ -4973,7 +4973,7 @@ Node TheoryArithPrivate::expandDefinition(LogicRequest &logicRequest, Node node)
       if (!den.isConst() || den.getConst<Rational>().sgn() == 0)
       {
         Node divByZeroNum =
-            getArithSkolemApp(logicRequest, num, arith_skolem_div_by_zero);
+            getArithSkolemApp(logicRequest, num, ArithSkolemId::DIV_BY_ZERO);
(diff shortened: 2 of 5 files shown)


src/theory/arith/theory_arith_private.h
```diff
@@ -828,11 +828,16 @@ class TheoryArithPrivate {
 
   Statistics d_statistics;
 
-  enum ArithSkolemId
+  enum class ArithSkolemId
```
ajreynol: Why is the "class" keyword here?

4tXJ7f: [class enums](https://en.cppreference.com/w/cpp/language/enum) make the constants scoped, i.e. you _have_ to refer to them as `ArithSkolemId::*` (so you don't need to prefix your constants). At the same time it disallows implicit conversions, so you can't accidentally compare an `ArithSkolemId` to an integer or another `enum`. The [C++ Core Guidlines](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Renum-class) recommend the use of class enums over "plain" enums for this reason.

ahmed-irfan:
I think it is wrong to interpret that the `sqrt` result of a negative number is undefined. It is defined as a complex number that does not exist in the Real numbers. So, the UNSAT answer to the query ` sqrt(x) = x /\ x < 0` is correct.

ajreynol:
I had talked to Andres offline about this.

The semantics committed by Andres are correct.  The issue is that the domain/range of `sqrt` is declared to be "Real". Thus, even though the result of the sqrt for a negative number should be complex, it cannot be defined in the range of Real, thus we must leave it undefined.  
While the example `sqrt(x) = x /\ x < 0` being SAT is unintutitive, notice that there are other contexts where the globally unsat semantics would be even more unintuitive. Say for example:
```
(declare-fun P (Real) Bool)
(assert (P (sqrt (- 1)))
(check-sat)
```
This would be unsat by the semantics you mention.

In short, we are defining "sqrt" to be the behavior of square root, restricted to its Real -> Real behavior. If a user wanted to model the behavior of sqrt for complex numbers, this would have to be a layer on top of this.

PR #3483, opened 2019-11-20
Evaluation unfolding for symbolic SyGuS constructors

ajreynol:
This updates our evaluation unfolding techniques to fully integrate symbolic constructors (e.g. the "any constant" constructor). 

This is work towards combining symbolic sygus constructors with unification algorithms.

### DIFF
--- a/src/theory/quantifiers/sygus/sygus_eval_unfold.cpp
+++ b/src/theory/quantifiers/sygus/sygus_eval_unfold.cpp
@@ -260,8 +260,8 @@ Node SygusEvalUnfold::unfold(Node en,
     Trace("sygus-eval-unfold-debug")
         << "...it is an any-constant constructor" << std::endl;
     Assert(dt[i].getNumArgs() == 1);
-    // always abstract the any constant, regardless of whether this condition
-    // holds? TODO
+    // If the argument to evaluate is itself concrete, then we use its
+    // argument; otherwise we return its selector.
     if (en[0].getKind() == APPLY_CONSTRUCTOR)
     {
       Trace("sygus-eval-unfold-debug")
--- a/src/theory/quantifiers/sygus/sygus_eval_unfold.cpp
+++ b/src/theory/quantifiers/sygus/sygus_eval_unfold.cpp
@@ -239,9 +239,10 @@ Node SygusEvalUnfold::unfold(Node en,
     args.push_back(en[i]);
   }
 
-  Type headType = en[0].getType().toType();
+  TypeNode headType = en[0].getType();
+  Type headTypeT = headType.toType();
   NodeManager* nm = NodeManager::currentNM();
-  const Datatype& dt = static_cast<DatatypeType>(headType).getDatatype();
+  const Datatype& dt = headType.getDatatype();
   unsigned i = datatypes::utils::indexOf(ev.getOperator());
   if (track_exp)
   {
@@ -271,7 +272,7 @@ Node SygusEvalUnfold::unfold(Node en,
     else
     {
       Node ret = nm->mkNode(
-          APPLY_SELECTOR_TOTAL, dt[i].getSelectorInternal(headType, 0), en[0]);
+          APPLY_SELECTOR_TOTAL, dt[i].getSelectorInternal(headTypeT, 0), en[0]);
       Trace("sygus-eval-unfold-debug")
           << "...return (from constructor) " << ret << std::endl;
       return ret;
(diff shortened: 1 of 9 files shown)


src/theory/quantifiers/sygus/sygus_eval_unfold.cpp
```diff
@@ -158,7 +164,10 @@ void SygusEvalUnfold::registerModelValue(Node a,
               eval_children.end(), it->second[i].begin(), it->second[i].end());
           Node eval_fun = nm->mkNode(DT_SYGUS_EVAL, eval_children);
           eval_children.resize(1);
-          res = d_tds->unfold(eval_fun, vtm, exp);
+          // If we explicitly asked to unfold, we use single step, otherwise
+          // we use multi step.
+          res = unfold(eval_fun, vtm, exp, true, !do_unfold);
```
HanielB: So Booleans with anyconst will always be one-step unfolded? Why shouldn't anyconst subsume that? (I'm not sure it should, but this makes the case it definitely shouldn't).

src/theory/quantifiers/sygus/sygus_eval_unfold.cpp
```diff
@@ -193,6 +204,129 @@ void SygusEvalUnfold::registerModelValue(Node a,
...
+  AlwaysAssert(ev.getKind() == APPLY_CONSTRUCTOR);
+  std::vector<Node> args;
+  for (unsigned i = 1, nchild = en.getNumChildren(); i < nchild; i++)
+  {
+    args.push_back(en[i]);
+  }
+
+  Type headType = en[0].getType().toType();
+  NodeManager* nm = NodeManager::currentNM();
+  const Datatype& dt = static_cast<DatatypeType>(headType).getDatatype();
```
HanielB: update

src/theory/quantifiers/sygus/sygus_eval_unfold.cpp
```diff
@@ -193,6 +204,129 @@ void SygusEvalUnfold::registerModelValue(Node a,
...
+    Trace("sygus-eval-unfold-debug")
+        << "...it is an any-constant constructor" << std::endl;
+    Assert(dt[i].getNumArgs() == 1);
+    // If the argument to evaluate is itself concrete, then we use its
+    // argument; otherwise we return its selector.
+    if (en[0].getKind() == APPLY_CONSTRUCTOR)
+    {
+      Trace("sygus-eval-unfold-debug")
+          << "...return (from constructor) " << en[0][0] << std::endl;
+      return en[0][0];
```
HanielB: for anyconst shouldn't we always be in this first case? Why the `else`?

src/theory/quantifiers/sygus/sygus_eval_unfold.cpp
ajreynol: Why would anyconst subsume this?
If we are using symbolic constructors, we do the limited unfolding (traverse and unfold all but the any constant) in the case where we otherwise would do complete unfolding. This means that our policy on when the one step folding happens remains unchanged.

src/theory/quantifiers/sygus/sygus_eval_unfold.cpp
ajreynol: There are some rare cases where we ask this function to evaluate a partially specified datatype term (I've added some comments in this PR about that).
In other words, we generally use this function to unfold eval(e,C) for shared selector chain e, but in some cases we use it for eval( d, C) where d is an application of a constructor. In the case the we are asking this function to unfold e.g. eval( plus( anyconstant(5), e.2), C), we would evaluate the any constant here to 5. Whereas the more common case is that we use this function to unfold eval( e, C ) where e has a model value that we are evaluating wrt.

src/theory/quantifiers/sygus/sygus_eval_unfold.cpp
HanielB: The consequence of this is that we will never do full unfolding of Boolean operators, isn't it? Shouldn't we have this as an option?

ajreynol: The option `options::sygusEvalUnfoldBool()` determines whether we do single-step or full unfolding of Bool operators. This is preserved by this PR.

HanielB: I see now.

PR #3484, opened 2019-11-21
[proof] ITE translation fix

alex-ozdemir:
Our proof machinery has two notions of boolean-ness. The first is `formula`, which is used at the SMT/SAT level, for logical formulas that might be true or false. The second is the sort `Bool`, which is an SMT sort, and is used for user-defined booleans. Our proof machinery lifts ITEs over `Bool` (called `ite`) so that they are instead over `formula` (called `ifte`).

Unfortunately, we messed up this lifting. To do the lift, both arms of the ITE must be lifted to `formula`. However, we forgot to do this.

Determining whether a  `Bool` to `formula` lift is necessary requires some coordination between the theory and the global proof, since the theory controls whether `Bool` or `formula` is used by default. That was broken for arithmetic as well.

This PR includes a test which verifies that we're doing this correctly.

This bug was found within a failing test for #2855, but that test also fails for other reasons, which is why a new test is required.


src/proof/arith_proof.cpp
```diff
@@ -1188,4 +1188,11 @@ void LFSCArithProof::printAliasingDeclarations(std::ostream& os, std::ostream& p
   // Nothing to do here at this point.
 }
 
+bool LFSCArithProof::printsAsBool(const Node& n)
+{
+  // Our boolean variables and constants print as sort Bool.
+  // All complex booleans print as formulas.
+  return n.getType().isBoolean() and (n.isVar() or n.isConst());
```
yoni206: Is this peculiar to arithmetic proofs? Or can be moved to the original function in `theory_proof.cpp`?
If we really need to override, maybe document why this is different than the general case?

src/proof/theory_proof.cpp
```diff
@@ -869,6 +869,23 @@ void LFSCTheoryProofEngine::printBoundTerm(Expr term, std::ostream& os, const Pr
   printTheoryTerm(term, os, map);
 }
 
+void LFSCTheoryProofEngine::printBoundFormula(Expr term,
+                                              std::ostream& os,
+                                              const ProofLetMap& map)
+{
+  Assert(term.getType().isBoolean() or term.getType().isPredicate());
+  bool wrapWithBoolToPred = term.getType().isBoolean() and printsAsBool(term);
```
yoni206: https://github.com/CVC4/CVC4/blob/596fe8c79106dd9b7764df2ddce6b2d3344fea34/src/proof/theory_proof.cpp#L1113

Isn't `wrapWithBoolToPred` always false? What am I missing here?

src/proof/theory_proof.cpp
```diff
@@ -879,17 +896,28 @@ void LFSCTheoryProofEngine::printCoreTerm(Expr term, std::ostream& os, const Pro
...
-    if (booleanCase && printsAsBool(term[0])) os << "(p_app ";
-    printBoundTerm(term[0], os, map);
-    if (booleanCase && printsAsBool(term[0])) os << ")";
+    bool usePredicateType = term.getType().isBoolean();
+    os << (usePredicateType ? "(ifte " : "(ite _ ");
 
+    printBoundFormula(term[0], os, map);
     os << " ";
-    printBoundTerm(term[1], os, map);
+    if (usePredicateType)
```
yoni206: Shouldn't this be updated to `term[1].getType().isBoolean()` in case the results of the ITE are booleans?

src/proof/arith_proof.cpp
alex-ozdemir: This rule is specific for arithmetic proofs. Note that `theory_proof.cpp` does contain a default implementation, but the function is virtual, because ultimately, since theories print the theory-proof, it is theories that control the type of a printed proof term.

How about I expand the documentation for `TheoryProof::printsAsBool`, elaborating on why theories may need to override?

src/proof/theory_proof.cpp
alex-ozdemir: A term of sort `Bool` may, in the proof, be printed with type `formula`, in which case `printsAsBool` would return false, and `wrapWithBoolToPred` would also be false.

src/proof/theory_proof.cpp
alex-ozdemir: I believe that our typing rules dictate that that the arms of an ITE have the same type as the ITE expression itself. Thus, it suffices to check the type of the ITE itself.

I'll add an assertion for this though!

alex-ozdemir: Revisiting this, I was wrong: the arms can actually be strict subtypes of the ITE.

However, we still want to use `usePredicateType`, redubbed `useFormulaType`, because if we choose to use `ifte`, the arms have to be formulas.

test/regress/regress0/arith/ite-lift.smt2
```diff
@@ -0,0 +1,15 @@
...
+(set-info :difficulty "0")
+(set-logic QF_LRA)
+
+(declare-fun x_0 () Real)
+(declare-fun x_1 () Real)
+(declare-fun b_f () Bool)
+(assert (<= x_0 0))
+(assert (<= x_1 0))
+(assert (not b_f))
+(assert (ite b_f b_f (>= (+ x_0 x_1) 1)))
```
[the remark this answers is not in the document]
alex-ozdemir: Yep!

PR #3485, opened 2019-11-21
Introduce a hard limit for rec-fun evaluation in SyGuS

HanielB:
This avoids infinite loops in the evaluation of recursive functions when their arguments can change at each iteration (such as in an evaluation of a construct simulating a loop). Currently we use a predefined limit of 1000 evaluations of a recursive function per round. This can be set via an option.


HanielB:
No, I have to do another one that allows us to claim the recursive function for SyGuS

PR #3493, opened 2019-11-22
 Update ownership policy for dynamic quantifiers splitting

### LINKED ISSUE #3481: unproved VC with only one quantifier instance needed
Hi CVC4 team,
while working on some SPARK code, I have encountered a VC which seemed provable to me but for which cvc4 1.7.1 (and I think cvc4 master too) returns unknown almost instantly. The VC attached may look a bit complicated because it is generated code, but really, only the last 2 asserts are of interest. The first one is a universally quantified formula, and the other is the negation of an instance of this formula (if I have not done any mistake). I would be interested in understanding what is causing the issue for CVC4, and if there would be a (systematic) way to help the prover here.
Best regards,
Claire
[cvc4_unproved.txt](https://github.com/CVC4/CVC4/files/3869738/cvc4_unproved.txt)

ajreynol:
Fixes #3481. 

We now allow both the original and split form of quantified formulas from dynamic splitting to be processed in parallel by E-matching/conflict-based/enumerative instantiation.


ajreynol:
Yes, for UNSAT I'd expect this to consistently be better.

PR #3496, opened 2019-11-25
Better front-end type checking for SyGuS

ajreynol:
In the version 1 SyGuS parser, we were unable to type check SyGuS constructors on the front end in a reasonable way due to the complexity of the parser.  This is no longer the case in the version 2 parser.  This adds a call to getType(true) at the proper place in parser/smt2/smt2.cpp and updates the documentation in sygus/type_info.cpp.  This ensures that we throw a proper type checking error instead of segfaulting on SyGuS inputs with terms that are not well-typed within SyGuS grammars.

### DIFF
--- a/src/parser/smt2/smt2.cpp
+++ b/src/parser/smt2/smt2.cpp
@@ -1435,6 +1435,10 @@ void Smt2::addSygusConstructorTerm(Datatype& dt,
                                    std::map<Expr, Type>& ntsToUnres) const
 {
   Trace("parser-sygus2") << "Add sygus cons term " << term << std::endl;
+  // Ensure that we do type checking here to catch sygus constructors with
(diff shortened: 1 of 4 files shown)


src/parser/smt2/smt2.cpp
```diff
@@ -1435,6 +1435,10 @@ void Smt2::addSygusConstructorTerm(Datatype& dt,
                                    std::map<Expr, Type>& ntsToUnres) const
 {
   Trace("parser-sygus2") << "Add sygus cons term " << term << std::endl;
+  // Ensure that we do type checking here to catch sygus constructors with
+  // malformed builtin operators. The argument "true" to getType here forces
+  // a recursive well-typedness check.
+  Type eType = term.getType(true);
```
HanielB: Could you change this for just the function call rather than the assignment?

PR #3502, opened 2019-11-27
Enable sygusRecFun by default and fixes SyGuS+RecFun+HO issues

HanielB:
Now the option to use defined recursive functions in a SyGuS problem to reduce candidate solutions is on by default. This is achieved by claiming these functions in the SynthEngine. 

Doing the above uncovered a bug in the combination of  SyGuS+RecFun+HO, which was only working because the instantiation of recursive functions was making up for refinement lemmas being generated with recursive functions (what should not happen). This has been fixed.


src/theory/quantifiers/sygus/synth_engine.cpp
```diff
@@ -276,8 +276,20 @@ void SynthEngine::registerQuantifier(Node q)
...
+  if (d_quantEngine->getQuantAttributes()->isFunDef(q))
+  {
+    Assert(options::sygusRecFun());
+    // If it is a recursive function definition, add it to the function
+    // definition evaluator class.
+    Trace("cegqi") << "Registering function definition : " << q << "\n";
+    FunDefEvaluator* fde = d_tds->getFunDefEvaluator();
+    fde->assertDefinition(q);
+    return;
+  }
```
ajreynol: There is an indentation issue here, right?

test/regress/CMakeLists.txt
```diff
@@ -1743,6 +1743,7 @@ set(regress_1_tests
   regress1/sygus/qe.sy
   regress1/sygus/qf_abv.smt2
   regress1/sygus/real-grammar.sy
+  regress1/sygus/rec-fun-swap.sy
```
ajreynol: I don't see this regression committed.

PR #3509, opened 2019-11-29
Fix case of higher-order + sygus inference

### LINKED ISSUE #3507: CVC4 --sygus-inference check failure on LIA formula at CVC4/src/preprocessing/passes/sygus_inference.cpp:202
Hi,
For this formula,
```
(set-logic ALL)
(declare-fun f (Int) Bool)
(declare-fun g (Int) Bool)
(assert (and (distinct f g) (g 0)))
(assert (exists ((x Int)) (f x)))
(check-sat)
```
CVC4 with --sygus-inference will throw out a check failure:
```
Fatal failure within bool CVC4::preprocessing::passes::SygusInference::solveSygus(std::vector<CVC4::NodeTemplate<true> >&, std::vector<CVC4::NodeTemplate<true> >&, std::vector<CVC4::NodeTemplate<true> >&) at /home/CVC4/src/preprocessing/passes/sygus_inference.cpp:202
Check failure
 std::find(free_functions.begin(), free_functions.end(), cur) == free_functions.end()
Aborted
``` 

OS: Ubuntu 18.04
Revision: 5e2d39c

ajreynol:
Fixes #3507.

### DIFF
--- a/src/preprocessing/passes/sygus_inference.cpp
+++ b/src/preprocessing/passes/sygus_inference.cpp
@@ -197,10 +197,13 @@ bool SygusInference::solveSygus(std::vector<Node>& assertions,
         }
         else if (cur.isVar() && cur.getKind() != BOUND_VARIABLE)
         {
-          // a free variable is a 0-argument function to synthesize
-          Assert(std::find(free_functions.begin(), free_functions.end(), cur)
-                 == free_functions.end());
-          free_functions.push_back(cur);
+          // A free variable that is not in a higher-order context is a 
+          // 0-argument function-to-synthesize.
+          if (std::find(free_functions.begin(), free_functions.end(), cur)
+                 == free_functions.end())
+          {
+            free_functions.push_back(cur);
+          }
         }
         else if (cur.isClosure())
         {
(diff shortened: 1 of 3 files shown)


src/preprocessing/passes/sygus_inference.cpp
```diff
@@ -197,10 +197,13 @@ bool SygusInference::solveSygus(std::vector<Node>& assertions,
         }
         else if (cur.isVar() && cur.getKind() != BOUND_VARIABLE)
         {
-          // a free variable is a 0-argument function to synthesize
-          Assert(std::find(free_functions.begin(), free_functions.end(), cur)
-                 == free_functions.end());
-          free_functions.push_back(cur);
+          // A free variable that is not in a higher-order context is a
+          // 0-argument function-to-synthesize.
```
HanielB: Please clarify that you can ignore the HO functions here because that are accounted for in a different part of the code

ajreynol: I'm not ignoring them, I'm ensuring we don't add duplicates to the vector.

HanielB: Yes, you are ignoring the duplicate. If one looks only to the comment one can be confused, as I was, that they are duplicates. It'd be to say why they are duplicates.

ajreynol: I've made the fix more robust, see my latest commit.

PR #3518, opened 2019-12-02
Improve flexibility of lemma output in non-linear solver

ajreynol:
Work towards #2153.  This PR makes it so that lemma output is not hardcoded into the core procedure of the non-linear solver.


src/theory/arith/nonlinear_extension.h
```diff
@@ -327,18 +347,19 @@ class NonlinearExtension {
...
+  /** singleton version of above */
+  unsigned filterLemma(Node lem, std::vector<Node>& out);
 
-  /** Potentially sends lemmas to the output channel and clears lemmas. Returns
-   * the number of lemmas sent to the output channel.
+  /**
+   * Send lemmas in out on the output channel of theory of arithmetic.
    */
-  int flushLemmas(std::vector<Node>& lemmas);
+  void sendLemmas(const std::vector<Node>& out, bool preprocess = false);
```
ahmed-irfan: what is preprocess?

ajreynol: It corresponds to this flag: https://github.com/CVC4/CVC4/blob/master/src/theory/output_channel.h#L116

ahmed-irfan: i see

PR #3522, opened 2019-12-02
Add mkOp for a single Kind

makaimann:
I realized I never added the `mkOp(Kind kind)` method. This is not necessary, because you can always call `mkTerm` with the raw `Kind`. However, I was thinking that for uniformity it should be there.


src/api/cvc4cpp.cpp
```diff
@@ -3048,6 +3048,13 @@ Term Solver::mkTuple(const std::vector<Sort>& sorts,
 /* Create operators                                                           */
 /* -------------------------------------------------------------------------- */
 
+Op Solver::mkOp(Kind kind) const
```
aniemetz: Can you add a unit test?

makaimann: Definitely! I'll add that now.

I actually just realized there's an issue here. Adding this makes it kinda easy to do stupid things, such as: `mkOp(BITVECTOR_EXTRACT)`. How would you recommend guarding against that? Should we just list all indexed operators and check that it's not one of those? Or would it be better to just not have this `mkOp` with a single `Kind` functionality?

aniemetz: Good question. If we keep it, we should guard against using it with indexed operators. We don't add new operators frequently, so I think that should be fine.

makaimann: Sounds good. Made that change with 9c8f7e7d186c08db8f190556cf5797d1a677f081 (+ a small fix in the subsequent commit)

src/api/cvc4cpp.cpp
```diff
@@ -527,6 +527,30 @@ const static std::unordered_map<CVC4::Kind, Kind, CVC4::kind::KindHashFunction>
         {CVC4::Kind::LAST_KIND, LAST_KIND},
     };
 
+/* Set of kinds for indexed operators */
+const static std::unordered_set<Kind, KindHashFunction> indexed_kinds(
```
```suggestion
const static std::unordered_set<Kind, KindHashFunction> s_indexed_kinds(
```

src/api/cvc4cpp.h
```diff
@@ -1890,9 +1890,18 @@ class CVC4_PUBLIC Solver
                const std::vector<Term>& terms) const;
 
   /* .................................................................... */
-  /* Create Operator Terms                                                */
+  /* Create Operators                                                     */
   /* .................................................................... */
 
+  /**
+   * Create an operator for a builtin Kind
```
aniemetz: Can you add a line that the kind may not be an indexed kind?

makaimann: Good idea! Updated with **fc9989c8e73989b41e6f4c030c8ada23a9357283**

Issue #3441, opened 2019-11-05
Crash with --check-models 


wintered:
Hi all, 

for the following formula, CVC4 will crash with `--check-models` option enabled

```
(set-logic QF_NIA)                                                              
(declare-fun a () Int)                                                          
(declare-fun b () Int)                                                          
(declare-fun c () Int)                                                          
(declare-fun d () Int)                                                          
(assert (and (>= (+ (* b c (- (-  4) d)) (- 1)) 0 (div a b))))                  
(check-sat)  
```

```
Fatal failure within void CVC4::TheoryEngine::checkTheoryAssertionsWithModel(bool) at /home/windomin/solver_binaries/cache/CVC4/src/theory/theory_engine.cpp:2229
Internal error detectedTHEORY_ARITH has an asserted fact that the model doesn't satisfy.
The fact: (= kf_6 (+ (* 4 b) (* b d)))
Model value: false

Aborted
```
OS: Ubuntu 18.04
Revision: c547bd5

ajreynol:
This is a duplicate of #2153. The solution is in progress (see #3426).

Issue #3411, opened 2019-10-27
Assertion failure when checking the model of NRA formula


muchang:
Hi,
For this formula,
```
(set-logic NRA)
(declare-fun a () Real)
(declare-fun b () Real)
(assert (= (/ 0 (+ 1 (* a a b))) 0))
(check-sat)
```
CVC4 will throw out an assertion violation:
```
(error "Assertion failure
void CVC4::theory::TheoryEngineModelBuilder::debugCheckModel(CVC4::theory::TheoryModel*)
/home/suz/software/CVC4/src/theory/theory_model_builder.cpp:878:
  tm->getValue(*eqc_i) == rep
run with -d check-model::rep-checking for details")
```
when check-models is on:
```
cvc4 --check-models file.smt2
```

OS: Ubuntu 18.04
Revision: 2493601

ajreynol:
Thanks for the benchmark, this is a duplicate of #2153.

Currently the models produced by the non-linear solver violate some of the assertions in CVC4's model validating code.

Issue #3003, opened 2019-05-15
--check-models fails on QF_NRA benchmark


4tXJ7f:
On the following benchmark, CVC4 fails with `--check-models`:

```
(set-logic QF_NRA)                                                                                                                                                  
(declare-fun _substvar_15_ () Real)                                                                                                                                 
(declare-fun _substvar_17_ () Real)                                                                                                                                 
(assert (let ((?v_1 (<= 0.0 _substvar_15_))) (and ?v_1 (and ?v_1 (and ?v_1 (= (* _substvar_15_ _substvar_15_) (+ 1 (* _substvar_17_ (* _substvar_17_ (- 1)))))))))) 
(check-sat)                                                                                                                                                         
```

ajreynol:
This appears to be due to an approximation of an irrational value, if you run --dump-models:
```
(model
(approximation (and (>= _substvar_15_ (/ 209533 524288)) (>= (* (- 1.0) _substvar_15_) (/ (- 15086389) 37748736))))
(define-fun _substvar_15_ () Real (/ 15086389 37748736))
(define-fun _substvar_17_ () Real (/ 150994931 150994944))
)
```
Notice the approximation. 
CVC4 uses these kinds of approximations in a "SAT" response when it, e.g. simultaneously knows that x must be equal to sqrt(2) and must be greater than 1.4. If it can compute (by binary search) that sqrt(2) is greater than 1.4, then it can answer "SAT" without having an exact model.
Issue https://github.com/CVC4/CVC4/issues/2008 is related to this.

ajreynol:
Summarizing here: this isn't a critical failure, I'm going to lower this to "minor".  Notice z3 also says "sat" on this one.

ajreynol:
This is a duplicate of #2153.

Issue #2008, opened 2018-05-29
Generate model error instead of assertion failure for approximate models


4tXJ7f:
Currently, running with `--check-models` on QF_NRA/meti-tarski/sin/cos/sin-cos-346-b-chunk-0169.smt2 results in an assertion failure:

```
(error "Internal error detected
void CVC4::TheoryEngine::checkTheoryAssertionsWithModel(bool)
../../../../src/theory/theory_engine.cpp:2275:


THEORY_ARITH has an asserted fact that the model doesn't satisfy.
The fact: (= (* skoSQ3 skoSQ3) 3.0)
Model value: false
")
```

Since the models may be approximate, it can happen that the model does not satisfy the constraints but this should be reported as a model error not an assertion failure (maybe together with the caveat about approximate models).

ajreynol:
@4tXJ7f, can you say more about what a "model error" is? Is there a particular kind of exception for this? Or should this be a warning?

ajreynol:
This is a duplicate of #2153.

ajreynol:
Reopening since this issue actually pertains to a feature that should be fixed: we should throw an error instead of an assertion failure when approximations are used in the model.

PR #3525, opened 2019-12-04
Make nonlinear solver intercept model assignments from the linear arithmetic solver

ajreynol:
This fixes consistency issues for models generated for non-linear arithmetic. 

Currently, the models generated by nonlinear arithmetic have "internal inconsistencies", e.g. the equality engine of the TheoryModel object contains assertions that it does not satisfy. This is caused by the following chain of events:

(1) The linear arithmetic solver decides on a set of model assignments and adds them to the current candidate model object, call it M. In the implementaiton, this is communicated in the form of a set of equalities v1 = c1 ^ ... ^ vn = cn which are added to the TheoryModel object during collectModelInfo.
(2) During a last call effort, the nonlinear solver may infer that the current assertions are in fact satisfied by an alternative model assignment, that is, for some different v1 = c'1 ^ ... ^ vn = c'n.  It then "overwrites" these values by the recordApproximation feature.  
(3) The system (soundly) answers "SAT". The model is in fact guaranteed to be correct wrt getModelValue calls. However, the model is "internally inconsistent" since the original equalities v1 = c1 ^ ... ^ vn = cn are not retracted from the model. Thus, we encounter assertion failures in debug as well as `--check-model` failures in either production and debug (these check model failures only occur for internally generated assertions).

This commit refactors the interplay between the linear and nonlinear solvers so that the nonlinear solver "intercepts" the model assignment from the linear solver and modifies it as necessary *before* it is sent to TheoryModel during collectModelInfo.  This ensures that no assertions are later invalidated when the nonlinear solver decides to change model values based on nonlinear reasoning a la Cimatti et al TACAS 2017.

This PR also removes the use of "skolem atoms" which was a way of avoiding checkModel failures due to internally introduced atoms. This is no longer necessary as we can correctly generate models now.

This fixes #2531 (that benchmark times out now).  We now generate internally consistent models for the benchmarks on #2008, #3003, #3441, #3411, which are added as regressions by this commit.
Notice that `--check-model` fails on some benchmarks still, since we do not have the proper infrastructure in place for aborting checkModel calls when approximations are used. I've reopened #2008 to track this issue.


4tXJ7f:
I am adding @alex-ozdemir as a secondary reviewer since he is interested in starting to review PRs.

ahmed-irfan:
Shouldn't there be some regression tests with the option `--check-model`?

ajreynol:
`--check-model` is run for all regression tests by default :)
See https://github.com/CVC4/CVC4/blob/master/test/regress/run_regression.py#L309

test/regress/regress1/nl/issue3441.smt2
```diff
@@ -0,0 +1,8 @@
+(set-logic QF_NIA)
```
ahmed-irfan: I am not sure, but did you miss the headers 
```
; COMMAND-LINE
; EXPECT
```

ajreynol: This isn't necessary if there aren't any specific options you are using or output you are expecting (beyond a status).

ahmed-irfan:
> `--check-model` is run for all regression tests by default :)
> See https://github.com/CVC4/CVC4/blob/master/test/regress/run_regression.py#L309

I didn't know :)
Thanks

PR #3529, opened 2019-12-04
Optimize the rewriter for DT_SYGUS_EVAL

ajreynol:
This makes it so that we don't construct intermediate unfoldings of applications of DT_SYGUS_EVAL, which wastes time in node construction. It makes the sygusToBuiltin utility in TermDbSygus use this implementation.


ajreynol:
@4tXJ7f this may interest you

src/theory/datatypes/datatypes_rewriter.cpp
```diff
@@ -120,34 +120,16 @@ RewriteResponse DatatypesRewriter::postRewrite(TNode in)
     if (ev.getKind() == APPLY_CONSTRUCTOR)
     {
       Trace("dt-sygus-util") << "Rewrite " << in << " by unfolding...\n";
```
HanielB: shouldn't you update this trace?

src/theory/datatypes/theory_datatypes_utils.h
```diff
@@ -178,12 +178,36 @@ Node applySygusArgs(const Datatype& dt,
...
+ * It does so by (1) converting constant subterms of n to builtin terms and
+ * evaluating them on the arguments args, (2) unfolding non-constant
+ * applications of sygus constructors in n with respect to args and (3)
+ * converting all other non-constant subterms of n to applications of
+ * DT_SYGUS_EVAL.
+ *
+ * For example, if
+ *   n = C_+( C_*( C_x(), C_y() ), n' ), and args = { 3, 4 }
+ * where n' is a variable, then this method returns:
+ *   12 + (DT_SYGUS_EVAL n' 3 4)
```
HanielB: Currently we can't have that `n'` has a different type, right?

src/theory/datatypes/theory_datatypes_utils.cpp
```diff
@@ -377,6 +378,196 @@ bool checkClash(Node n1, Node n2, std::vector<Node>& rew)
...
+        // non-datatypes are themselves
+        visited[cur] = cur;
+      }
+    }
+    else if (it->second.isNull())
+    {
+      Node ret = cur;
+      Assert(cur.getKind() == APPLY_CONSTRUCTOR);
+      const Datatype& dt = cur.getType().getDatatype();
+      // non sygus-datatype terms are also themselves
```
HanielB: Wouldn't it be better to treat them in the `else` above then? Not sure if they should be going in the cache as well, but if you think they should then you could just duplicate the attribute setting below.

src/theory/datatypes/theory_datatypes_utils.cpp
```diff
@@ -377,6 +378,196 @@ bool checkClash(Node n1, Node n2, std::vector<Node>& rew)
...
+        }
+        Assert(args.size() == svars.size());
+        // try evaluation if we have constant arguments
+        Node ret = constArgs ? eval.eval(bt, svars, args) : Node::null();
+        if (ret.isNull())
+        {
+          // if evaluation was not available, use a substitution
+          ret = bt.substitute(
+              svars.begin(), svars.end(), args.begin(), args.end());
+        }
```
HanielB: can `eval.eval` return `null` as well? Otherwise this is a bit of a strange way of constructing this assignment. Determine the result of `ret` directly by `constArgs`

src/theory/datatypes/theory_datatypes_utils.cpp
```diff
@@ -377,6 +378,196 @@ bool checkClash(Node n1, Node n2, std::vector<Node>& rew)
...
+          visited[cur] = nm->mkNode(DT_SYGUS_EVAL, eargs);
+        }
+      }
+    }
+    else if (it->second.isNull())
+    {
+      Node ret = cur;
+      Assert(cur.getKind() == APPLY_CONSTRUCTOR);
+      const Datatype& dt = cur.getType().getDatatype();
+      // non sygus-datatype terms are also themselves
```
HanielB: again you could have solved this guy already above, no?

src/theory/datatypes/datatypes_rewriter.cpp
ajreynol: It's still technically unfolding, it just is multi-step unfolding :)

src/theory/datatypes/theory_datatypes_utils.h
ajreynol: n' can have arbitrary type wrt n.

src/theory/datatypes/theory_datatypes_utils.cpp
ajreynol: My reason for doing this was avoiding several things, the main being that I don't want to access the datatype of the node in two places (hence twice for each constructor node).  So I did intentionally do it this way.

Btw, it is a super rare case that we ever pay a penalty for the non-sygus datatype case (this would be any-constants over user-defined datatypes).  Hence we should optimize for the case where there is no non-sygus datatypes.

src/theory/datatypes/theory_datatypes_utils.cpp
ajreynol: Yes, eval can return null (if a kind is unimplemented).

src/theory/datatypes/theory_datatypes_utils.cpp
ajreynol: Similar reason as above, should optimize for the non-sygus datatype case.

src/theory/datatypes/theory_datatypes_utils.h
HanielB: Isn't it the case then that `n'` can be applied to only some of the arguments that `n` was applied to?

src/theory/datatypes/theory_datatypes_utils.cpp
HanielB: Agreed. You could add a comment to have the reason for that choice documented in the code.

src/theory/datatypes/theory_datatypes_utils.h
ajreynol: No, `DT_SYGUS_EVAL` is a parametric operator; the key is that n and n' are terms pertaining to the same function to synthesize by construction, hence their evaluation operators (on paper) take the same set of arguments.

HanielB: Well, then it's not arbitrary. I thought this was the case, since we cannot define functions-to-synthesize in term of others.

PR #3534, opened 2019-12-05
contrib: Setup all dependencies in deps/ directory.

mpreiner:
Every dependency set up via the corresponding get-* script now downloads
the sources to deps/ and is installed into deps/install. This
makes the cmake/Find* modules simpler and allows us to cache deps/install
for our CI builds.

### DIFF
--- a/contrib/get-cryptominisat
+++ b/contrib/get-cryptominisat
@@ -20,7 +20,7 @@ cmake -DENABLE_PYTHON_INTERFACE=OFF \
       -DCMAKE_INSTALL_PREFIX="$INSTALL_DIR" \
       ..
 
-make libcryptominisat5 -j$(nproc) install
+make install -j$(nproc)
 
 echo
 echo ===================== Now configure CVC4 with =====================
--- a/contrib/get-cryptominisat
+++ b/contrib/get-cryptominisat
@@ -2,28 +2,25 @@
 #
 source "$(dirname "$0")/get-script-header.sh"
 
-if [ -e cryptominisat5 ]; then
-  echo 'error: file or directory "cryptominisat5" exists; please move it out of the way.' >&2
-  exit 1
-fi
-
+CMS_DIR="$DEPS_DIR/cryptominisat5"
 version="5.6.3"
 
-webget https://github.com/msoos/cryptominisat/archive/$version.tar.gz cryptominisat-$version.tar.gz
-tar xfvz cryptominisat-$version.tar.gz
-rm cryptominisat-$version.tar.gz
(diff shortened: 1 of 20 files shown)


contrib/get-cryptominisat
```diff
@@ -2,28 +2,25 @@
...
-
 cmake -DENABLE_PYTHON_INTERFACE=OFF \
       -DSTATICCOMPILE=ON \
       -DNOM4RI=ON \
+      -DONLY_SIMPLE=ON \
+      -DCMAKE_INSTALL_PREFIX="$INSTALL_DIR" \
       ..
 
-make libcryptominisat5 -j$(nproc)
+make libcryptominisat5 -j$(nproc) install
```
4tXJ7f: Won't the install build everything? I.e. wouldn't it be the same as just writing `make install -j$(nproc)`? Can we still install CMS without Boost?

contrib/get-script-header.sh
```diff
@@ -2,7 +2,15 @@
 #
 set -e -o pipefail
 
-cd "$(dirname "$0")/.."
+[ ! -e contrib ] && echo "$0 not called from base directory" && exit 1
```
4tXJ7f:
```suggestion
[ ! -d contrib ] && echo "$0 not called from base directory" && exit 1
```

contrib/get-script-header.sh
```diff
@@ -36,3 +44,52 @@ if [ -z "$PYTHON" ]; then
...
+  cp "$1" "$INSTALL_LIB_DIR"
+}
+
+function install_includes
+{
+  include="$1"
+  subdir="$2"
+  echo "Copying $1 to $INSTALL_INCLUDE_DIR/$subdir"
+  [ ! -d "$INSTALL_INCLUDE_DIR" ] && mkdir -p "$INSTALL_INCLUDE_DIR"
+  [ -n "$subdir" ] && mkdir -p "$INSTALL_INCLUDE_DIR/$subdir"
```
4tXJ7f:
```suggestion
  [ -n "$subdir" ] && [ ! -d "$INSTALL_INCLUDE_DIR/$subdir" ] && mkdir -p "$INSTALL_INCLUDE_DIR/$subdir"
```

For consistency?

contrib/get-cryptominisat
mpreiner: With `-DONLY_SIMPLE=ON` it builds CMS without boost. (https://github.com/msoos/cryptominisat/blob/master/CMakeLists.txt#L429)

PR #3535, opened 2019-12-06
Add ExprManager as argument to Datatype

ajreynol:
This makes the Expr-level Datatype take an ExprManager pointer as an argument.  Notice that this is similar to how Expr are currently handled (https://github.com/CVC4/CVC4/blob/master/src/expr/expr_template.h#L231).

This further requires making the Term-level DatatypeDecl class take a Solver as an argument.

This is required to set up an ExprManagerScope during destruction of Datatypes, since these will soon use the Node-level Datatype as a backend.


src/api/cvc4cpp.h
```diff
@@ -1104,29 +1105,38 @@ class CVC4_PUBLIC DatatypeDecl
  public:
   /**
    * Constructor.
+   * @param s the solver that created this datatype declaration
    * @param name the name of the datatype
    * @param isCoDatatype true if a codatatype is to be constructed
    * @return the DatatypeDecl
    */
-  DatatypeDecl(const std::string& name, bool isCoDatatype = false);
+  DatatypeDecl(const Solver* s,
```
4tXJ7f: These constructors should be private, users should use the `mk*` methods of the solver.

src/api/cvc4cpp.h
```diff
@@ -2292,6 +2302,41 @@ class CVC4_PUBLIC Solver
...
+   */
+  DatatypeDecl mkDatatypeDecl(const std::string& name,
+                              bool isCoDatatype = false);
+
+  /**
+   * Create a datatype declaration.
+   * Create sorts parameter with Solver::mkParamSort().
+   * @param name the name of the datatype
+   * @param param the sort parameter
+   * @param isCoDatatype true if a codatatype is to be constructed
```
4tXJ7f: Describe return value.

src/api/cvc4cpp.h
```diff
@@ -2292,6 +2302,41 @@ class CVC4_PUBLIC Solver
...
+  DatatypeDecl mkDatatypeDecl(const std::string& name,
+                              Sort param,
+                              bool isCoDatatype = false);
+
+  /**
+   * Create a datatype declaration.
+   * Create sorts parameter with Solver::mkParamSort().
+   * @param name the name of the datatype
+   * @param params a list of sort parameters
+   * @param isCoDatatype true if a codatatype is to be constructed
```
4tXJ7f: Describe return value.

src/api/cvc4cpp.h
```diff
@@ -1140,6 +1138,7 @@ class CVC4_PUBLIC DatatypeDecl
                const std::vector<Sort>& params,
                bool isCoDatatype = false);
 
+ public:
```
4tXJ7f: The private methods need to move after the public ones :)

PR #3540, opened 2019-12-06
Activate node-level datatype API

ajreynol:
This commit makes it so that the Node-level datatypes API is used as a backend to the Expr-level datatypes API.

The next step will be to update the code base in src/theory/ to access the Node-level datatype functions directly.


src/expr/datatype.cpp
```diff
@@ -979,44 +574,33 @@ Type DatatypeConstructor::getSpecializedConstructorType(Type returnType) const {
...
+  ExprManagerScope ems(d_constructor);
+  return d_internal->isSygusIdFunc();
 }
 
 unsigned DatatypeConstructor::getWeight() const
 {
   PrettyCheckArgument(
       isResolved(), this, "this datatype constructor is not yet resolved");
-  return d_weight;
+  return d_internal->getWeight();
```
4tXJ7f: Add `ExprManagerScope ems(d_constructor);` to be on the safe side.

src/expr/datatype.cpp
```diff
@@ -1029,214 +613,26 @@ std::shared_ptr<SygusPrintCallback> DatatypeConstructor::getSygusPrintCallback()
...
-    if( t.isDatatype() ){
-      const Datatype& dt = ((DatatypeType)t).getDatatype();
-      if( !dt.computeWellFounded( processing ) ){
-        return false;
-      }
-    }
-  }
-  return true;
+  TypeNode tn = TypeNode::fromType(t);
+  return d_internal->getCardinality(tn);
```
4tXJ7f: Add `ExprManagerScope ems(d_constructor);` to be on the safe side.

src/expr/datatype.cpp
```diff
@@ -1248,49 +644,41 @@ const DatatypeConstructorArg& DatatypeConstructor::operator[](std::string name)
...
 Expr DatatypeConstructor::getSelector(std::string name) const {
-  return (*this)[name].getSelector();
+  return (*this)[name].d_selector;
 }
 
 Type DatatypeConstructor::getArgType(unsigned index) const
 {
   PrettyCheckArgument(index < getNumArgs(), index, "index out of bounds");
-  return static_cast<SelectorType>((*this)[index].getType()).getRangeType();
+  return d_internal->getArgType(index).toType();
```
4tXJ7f: Add `ExprManagerScope ems(d_constructor);` to be on the safe side.

src/expr/datatype.cpp
```diff
@@ -1248,49 +644,41 @@ const DatatypeConstructorArg& DatatypeConstructor::operator[](std::string name)
...
 }
 
 bool DatatypeConstructor::involvesExternalType() const{
-  for(const_iterator i = begin(); i != end(); ++i) {
-    if(! SelectorType((*i).getSelector().getType()).getRangeType().isDatatype()) {
-      return true;
-    }
-  }
-  return false;
+  return d_internal->involvesExternalType();
```
4tXJ7f: Add `ExprManagerScope ems(d_constructor);` to be on the safe side.

src/expr/datatype.cpp
```diff
@@ -1248,49 +644,41 @@ const DatatypeConstructorArg& DatatypeConstructor::operator[](std::string name)
...
 }
 
 bool DatatypeConstructor::involvesUninterpretedType() const{
-  for(const_iterator i = begin(); i != end(); ++i) {
-    if(SelectorType((*i).getSelector().getType()).getRangeType().isSort()) {
-      return true;
-    }
-  }
-  return false;
+  return d_internal->involvesUninterpretedType();
```
4tXJ7f: Add `ExprManagerScope ems(d_constructor);` to be on the safe side.

src/expr/datatype.cpp
```diff
@@ -1248,49 +644,41 @@ const DatatypeConstructorArg& DatatypeConstructor::operator[](std::string name)
...
+    : d_internal(nullptr)
 {
   PrettyCheckArgument(name != "", name, "cannot construct a datatype constructor arg without a name");
+  d_internal = std::make_shared<DTypeSelector>(name, Node::fromExpr(selector));
 }
 
 std::string DatatypeConstructorArg::getName() const
 {
-  string name = d_name;
+  string name = d_internal->getName();
```
4tXJ7f: Add `ExprManagerScope ems(d_constructor);` to be on the safe side.

src/expr/datatype.cpp
```diff
@@ -1306,42 +694,27 @@ Expr DatatypeConstructorArg::getSelector() const {
...
   PrettyCheckArgument(isResolved(), this,
                 "cannot get a associated constructor for argument of an unresolved datatype constructor");
-  return d_constructor;
+  ExprManagerScope ems(d_selector);
+  return d_internal->getConstructor().toExpr();
 }
 
 SelectorType DatatypeConstructorArg::getType() const {
-  return getSelector().getType();
+  return d_selector.getType();
```
4tXJ7f: Add `ExprManagerScope ems(d_constructor);` to be on the safe side.

src/expr/datatype.cpp
```diff
@@ -1306,42 +694,27 @@ Expr DatatypeConstructorArg::getSelector() const {
...
+  ExprManagerScope ems(d_constructor);
+  Node seln = Node::fromExpr(sel);
+  return d_internal->getSelectorIndexInternal(seln);
 }
 
 Expr DatatypeConstructorArg::getConstructor() const {
   PrettyCheckArgument(isResolved(), this,
                 "cannot get a associated constructor for argument of an unresolved datatype constructor");
-  return d_constructor;
+  ExprManagerScope ems(d_selector);
```
4tXJ7f: In the other methods, we seem to use `d_constructor` for the `ExprManagerScope`. I'd guess it doesn't matter that we get it from `d_selector` here but maybe we should change it to `d_constructor` for consistency?

src/expr/datatype.cpp
```diff
@@ -1350,7 +723,26 @@ Type DatatypeConstructorArg::getRangeType() const {
...
+  // HOWEVER, this causes problems in ExprManager tear-down, because
+  // the attributes are removed before the pool is purged.  When the
+  // pool is purged, this triggers an equality test between Datatypes,
+  // and this triggers a call to isResolved(), which breaks because
+  // d_selector has no type after attributes are stripped.
+  //
+  // This problem, coupled with the fact that this function is called
+  // _often_, means we should just use a boolean flag.
+  //
+  return d_internal->isResolved();
```
4tXJ7f: Add `ExprManagerScope ems(d_constructor);` to be on the safe side.

src/expr/datatype.cpp
```diff
@@ -1350,7 +723,26 @@ Type DatatypeConstructorArg::getRangeType() const {
 
 bool DatatypeConstructorArg::isUnresolvedSelf() const
 {
-  return d_selector.isNull() && d_name.size() == d_name.find('\0') + 1;
+  std::string name = getName();
+  return d_selector.isNull() && name.size() == name.find('\0') + 1;
```
4tXJ7f: Is the second conjunct of this condition still relevant?

src/expr/datatype.cpp
```diff
@@ -1448,4 +841,93 @@ std::ostream& operator<<(std::ostream& out, const DatatypeIndexConstant& dic) {
   return out << "index_" << dic.getIndex();
 }
 
+std::string Datatype::getName() const { return d_internal->getName(); }
```
4tXJ7f: Add `ExprManagerScope ems(*d_em);` to be on the safe side.

src/expr/datatype.cpp
```diff
@@ -1448,4 +841,93 @@ std::ostream& operator<<(std::ostream& out, const DatatypeIndexConstant& dic) {
...
+size_t Datatype::getNumConstructors() const { return d_constructors.size(); }
+
+bool Datatype::isParametric() const { return d_internal->isParametric(); }
```
4tXJ7f: Add `ExprManagerScope ems(*d_em);` to be on the safe side.

src/expr/datatype.cpp
```diff
@@ -1448,4 +841,93 @@ std::ostream& operator<<(std::ostream& out, const DatatypeIndexConstant& dic) {
...
+size_t Datatype::getNumParameters() const
+{
+  return d_internal->getNumParameters();
```
4tXJ7f: Add `ExprManagerScope ems(*d_em);` to be on the safe side.

src/expr/datatype.cpp
```diff
@@ -1448,4 +841,93 @@ std::ostream& operator<<(std::ostream& out, const DatatypeIndexConstant& dic) {
...
+}
+Type Datatype::getParameter(unsigned int i) const
+{
+  CheckArgument(isParametric(),
+                this,
+                "Cannot get type parameter of a non-parametric datatype.");
+  CheckArgument(i < getNumParameters(),
+                i,
+                "Type parameter index out of range for datatype.");
+  return d_internal->getParameter(i).toType();
```
4tXJ7f: Add `ExprManagerScope ems(*d_em);` to be on the safe side.

src/expr/datatype.cpp
```diff
@@ -1448,4 +841,93 @@ std::ostream& operator<<(std::ostream& out, const DatatypeIndexConstant& dic) {
...
+}
+
+std::vector<Type> Datatype::getParameters() const
+{
+  CheckArgument(isParametric(),
+                this,
+                "Cannot get type parameters of a non-parametric datatype.");
+  std::vector<Type> params;
+  std::vector<TypeNode> paramsn = d_internal->getParameters();
```
4tXJ7f: Add `ExprManagerScope ems(*d_em);` to be on the safe side.

src/expr/datatype.cpp
```diff
@@ -1448,4 +841,93 @@ std::ostream& operator<<(std::ostream& out, const DatatypeIndexConstant& dic) {
...
+  // convert to type
+  for (unsigned i = 0, nparams = paramsn.size(); i < nparams; i++)
+  {
+    params.push_back(paramsn[i].toType());
+  }
+  return params;
+}
+
+bool Datatype::isCodatatype() const { return d_internal->isCodatatype(); }
```
4tXJ7f: Add `ExprManagerScope ems(*d_em);` to be on the safe side.

src/expr/datatype.cpp
```diff
@@ -1448,4 +841,93 @@ std::ostream& operator<<(std::ostream& out, const DatatypeIndexConstant& dic) {
...
+
+bool Datatype::isSygus() const { return d_internal->isSygus(); }
```
4tXJ7f: Add `ExprManagerScope ems(*d_em);` to be on the safe side.

src/expr/datatype.cpp
```diff
@@ -1448,4 +841,93 @@ std::ostream& operator<<(std::ostream& out, const DatatypeIndexConstant& dic) {
...
+
+bool Datatype::isTuple() const { return d_internal->isTuple(); }
```
4tXJ7f: Add `ExprManagerScope ems(*d_em);` to be on the safe side.

src/expr/datatype.cpp
```diff
@@ -1448,4 +841,93 @@ std::ostream& operator<<(std::ostream& out, const DatatypeIndexConstant& dic) {
...
+
+bool Datatype::isRecord() const { return d_isRecord; }
+
+Record* Datatype::getRecord() const { return d_record; }
+bool Datatype::operator!=(const Datatype& other) const
+{
+  return !(*this == other);
+}
+
+bool Datatype::isResolved() const { return d_internal->isResolved(); }
```
4tXJ7f: Add `ExprManagerScope ems(*d_em);` to be on the safe side.

src/expr/datatype.cpp
```diff
@@ -1448,4 +841,93 @@ std::ostream& operator<<(std::ostream& out, const DatatypeIndexConstant& dic) {
...
+}
+
+Datatype::const_iterator Datatype::end() const
+{
+  return const_iterator(d_constructors, false);
+}
+
+bool DatatypeConstructor::isResolved() const
+{
+  return d_internal->isResolved();
```
4tXJ7f: Add `ExprManagerScope ems(*d_em);` to be on the safe side.

src/expr/datatype.cpp
```diff
@@ -150,46 +123,81 @@ size_t Datatype::cindexOfInternal(Expr item)
...
+    replacementsn.push_back(TypeNode::fromType(t));
+  }
+  for (const Type& t : paramTypes)
+  {
+    paramTypesn.push_back(TypeNode::fromType(t));
+  }
+  for (const Type& t : paramReplacements)
+  {
+    paramReplacementsn.push_back(TypeNode::fromType(t));
+  }
```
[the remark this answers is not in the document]
ajreynol: I don't think this is worthwhile. If we are planning to remove the Expr-level API, I would prefer not adding new infrastructure for Expr/Type.

src/expr/datatype.cpp
ajreynol: We use `d_constructor` for `DatatypeConstructor` and `d_selector` for `DatatypeConstructorArg`.

src/expr/datatype.cpp
ajreynol: There are numerous of these functions that are necessary to be called *before* resolution (when d_constructor is created).  Thus, this suggestion would lead to a segfault since `d_constructor` is NULL. 

Thus we can't just do this to be on the safe side, it's in fact less safe since it assumes that d_constructor is always non-null.

src/expr/datatype.cpp
ajreynol: This would cause segfaults for reasons mentioned above.

src/expr/datatype.cpp
```diff
@@ -275,360 +253,160 @@ void Datatype::addSygusConstructor(Expr op,
...
   PrettyCheckArgument(isResolved(), this, "this datatype is not yet resolved");
-  return mkGroundTermInternal(t, false);
+  ExprManagerScope ems(d_self);
+  TypeNode tn = TypeNode::fromType(t);
+  Node gterm = d_internal->mkGroundTerm(tn);
+  if (gterm.isNull())
+  {
+    IllegalArgument(
+        *this, "datatype is not well-founded, cannot construct a ground term!");
+  }
```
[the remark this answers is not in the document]
ajreynol: I can change this, but in general I don't think we should care about refactoring code at the Expr layer that will be deleted soon.

src/expr/datatype.cpp
ajreynol: I don't think so, I'll remove.

src/expr/datatype.cpp
```diff
@@ -200,52 +208,24 @@ void Datatype::resolve(ExprManager* em,
...
-  }
 }
 
 void Datatype::addConstructor(const DatatypeConstructor& c) {
-  PrettyCheckArgument(!d_resolved, this,
-                "cannot add a constructor to a finalized Datatype");
+  Trace("dt-debug") << "Datatype::addConstructor" << std::endl;
+  PrettyCheckArgument(
+      !isResolved(), this, "cannot add a constructor to a finalized Datatype");
   d_constructors.push_back(c);
```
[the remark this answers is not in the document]
ajreynol: Right, that is something to consider.

src/expr/datatype.cpp
```diff
@@ -132,8 +105,8 @@ size_t Datatype::indexOfInternal(Expr item)
   if( item.getKind()==kind::APPLY_TYPE_ASCRIPTION ){
     return indexOf( item[0] );
   }else{
-    Assert(n.hasAttribute(DatatypeIndexAttr()));
-    return n.getAttribute(DatatypeIndexAttr());
+    Assert(n.hasAttribute(DTypeIndexAttr()));
+    return n.getAttribute(DTypeIndexAttr());
```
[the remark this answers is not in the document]
ajreynol: This index handles a corner case (of APPLY_TYPE_ASCRIPTION nodes generated at the user-level) that does not apply to the internal representation. So, this indexOf is more general.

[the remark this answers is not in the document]
ajreynol: FYI, this is something that I am in the process of refactoring, but is not included in this PR.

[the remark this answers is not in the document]
4tXJ7f: That sounds good, thanks.

src/expr/datatype.cpp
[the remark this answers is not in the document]
4tXJ7f: Yes, you are right.

src/expr/datatype.cpp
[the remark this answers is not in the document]
4tXJ7f: Ok, that's fair.

src/expr/datatype.cpp
4tXJ7f: Ah yes, I managed to confuse myself, sorry.

src/expr/datatype.cpp
4tXJ7f: Hmm, so we should consider also storing a pointer to the `ExprManager` in `DatatypeConstructorArg` (at least for the new API, doesn't matter that much for the `Expr` layer).

ajreynol: Right, that's something I wanted to avoid for Expr since it would be even more hassle.

ajreynol:
Yes, we should discuss this at some point.

PR #3546, opened 2019-12-08
[Regressions] Require proof support for abduction

4tXJ7f:
PR #3255 introduced a new feature for more scalable
abduction/interpolation that relies on unsat cores. The regressions
added in the PR were not marked as requiring proof support, however,
leading to failing builds if proof support was not enabled.


4tXJ7f:
This should fix the nightlies.

PR #3548, opened 2019-12-09
Avoid calling rewriter from type checker

### LINKED ISSUE #3536: Fatal failure within CVC4::theory::RewriteResponse (QF_ABVFP formula)
Hi, 
Consider the following formula: 

```
(declare-fun symbolic_0_double_12660 () (Array (_ BitVec 32) (_ BitVec 8)))
(assert
 (let ((?x12 (concat (select symbolic_0_double_12660 (_ bv1 32)) (select symbolic_0_double_12660 (_ bv0 32)))))
(let ((?x18 (concat (select symbolic_0_double_12660 (_ bv3 32)) (concat (select symbolic_0_double_12660 (_ bv2 32)) ?x12))))
(let ((?x24 (concat (select symbolic_0_double_12660 (_ bv5 32)) (concat (select symbolic_0_double_12660 (_ bv4 32)) ?x18))))
(let ((?x30 (concat (select symbolic_0_double_12660 (_ bv7 32)) (concat (select symbolic_0_double_12660 (_ bv6 32)) ?x24))))
(let (($x38 (= true (fp.geq ((_ to_fp 11 53) ?x30) ((_ to_fp 11 53) (_ bv13831004815617530266 64))) (fp.leq ((_ to_fp 11 53) ?x30) ((_ to_fp 11 53) (_ bv4607632778762754458 64))))))
(not (distinct true $x38))))))))
(check-sat)
```
CVC4 crashes on it with the following error message:

```
Fatal failure within CVC4::theory::RewriteResponse CVC4::theory::fp::rewrite::removed(CVC4::TNode, bool) at /home/windomin/project_share/solver_binaries/cache/CVC4/src/theory/fp/theory_fp_rewriter.cpp:157
Unreachable code reachedkind (FLOATINGPOINT_TO_FP_GENERIC) should have been removed?
```
Revision: 46eeb6a
OS: Ubuntu 18.04

4tXJ7f:
Fixes #3536. The type checker for the chain operator was calling the
rewriter. However, the floating-point rewriter was expecting
`TheoryFp::expandDefinition()` to be applied before rewriting. If the
chain operator had subterms that were supposed to be removed by
`TheoryFp::expandDefinition()`, the FP rewriter was throwing an
exception. This commit fixes the issue by not calling the full rewriter
in the type checker but by just expanding the chain operator. This is a
bit less efficient than before because the rewriter does not cache the
result of expanding the chain operator anymore but assuming that there
are no long chains, the performance impact should be negligible. It also
seemed like a reasonable assumption that the rewriter can expect to run
after `expandDefinition()` because otherwise the rewriter has to expand
definitions, which may be too restrictive.


4tXJ7f:
@ajreynol, do you agree that it makes sense to assume that `expandDefinition()` is applied before the rewriter. @aniemetz, you may be interested in this since you've been looking into the FP solver.

ajreynol:
I don't think its safe to assume that `expandDefinition` is called before rewriting.  I also don't think that the type checker should ever have to call the rewriter.

I think this PR is a step in the right direction (for the latter of these concerns).  For the former, we should additionally ensure that type checking rules are given for the FP operators that are causing the issues.  Do you happen to know which operators are causing issues?  And is it due to missing type rules?

4tXJ7f:
> For the former, we should additionally ensure that type checking rules are given for the FP operators that are causing the issues. Do you happen to know which operators are causing issues? And is it due to missing type rules?

So the issue is not due to missing type rules but due to the rewriter asserting that certain kinds do not occur, e.g. [`FLOATINGPOINT_TO_FP_GENERIC`](https://github.com/CVC4/CVC4/blob/master/src/theory/fp/theory_fp_rewriter.cpp#L1047). Probably the easiest fix is to just ignore those kinds instead of failing an assertion.

ajreynol:
Ah yes, that makes sense to me.  The rewriter should never assert a kind doesn't exist.  I would change these to identity.  If FP wants to ensure that those kinds don't exist, the proper time I believe would be during preRegisterTerm (https://github.com/CVC4/CVC4/blob/master/src/theory/fp/theory_fp.h#L43).

4tXJ7f:
@ajreynol I have implemented your suggestion.

4tXJ7f:
@aniemetz would you do a run with assertions enabled?

aniemetz:
@4tXJ7f maybe both, and compare the production run without these changes to get a feel for a possible performance degression.

4tXJ7f:
Alright, I did three runs with a short timeout (60s) on non-incremental FP benchmarks (default options):
```
  DIRECTORY |                                           fp_fix3536 |                                fp_fix3536_assertions |                                            fp_master |
  Benchmark |  Stat  Slvd   Tot  To Mo  Err  Cpu[s]   Mem[MB] Uniq |  Stat  Slvd   Tot  To Mo  Err  Cpu[s]   Mem[MB] Uniq |  Stat  Slvd   Tot  To Mo  Err  Cpu[s]   Mem[MB] Uniq |
      ABVFP |    ok     1     1   0  0    0     1.9      94.3    0 |    ok     1     1   0  0    0     4.0      97.8    0 |    ok     1     1   0  0    0     1.8      96.0    0 |
       BVFP |    ok     8    24  16  0    0   968.3    1406.0    0 |    ok     8    24  16  0    0   971.2    1387.5    0 |    ok     8    24  16  0    0   967.7    1392.8    0 |
         FP |    ee  1131  2476 540  0  805 37722.0  672721.1    0 |    ee  1114  2476 557  0  805 39114.6  623741.8    0 |    ee  1131  2476 540  0  805 37779.9  670435.5    0 |
   QF_ABVFP |    ee 17942 18129  35  0  152  7166.6  251257.8    3 |    ee 17936 18129  41  0  152  9230.9  274551.0    0 |    ee 17939 18129  38  0  152  7344.0  251673.0    0 |
    QF_BVFP |    ee 17129 17215  22  0   64  5573.4  240877.6    0 |    ee 17127 17215  24  0   64  7166.1  259668.3    0 |    ee 17129 17215  22  0   64  5679.3  241098.5    0 |
 QF_BVFPLRA |    to     0     1   1  0    0    60.1     903.0    0 |    to     0     1   1  0    0    60.1     885.8    0 |    to     0     1   1  0    0    60.1     891.0    0 |
      QF_FP |    ok 40186 40300 114  0    0  8336.4   96116.3    0 |    ok 40185 40300 115  0    0  8592.1   89664.1    0 |    ok 40186 40300 114  0    0  8386.5   93884.9    0 |
   QF_FPLRA |    ok     3    13  10  0    0   662.0   12152.0    0 |    ok     3    13  10  0    0   681.6   12033.5    0 |    ok     3    13  10  0    0   659.9   12154.6    0 |
   __totals |    ee 76400 78159 738  0 1021 60490.6 1275528.1    3 |    ee 76374 78159 764  0 1021 65820.5 1262029.8    0 |    ee 76397 78159 741  0 1021 60879.2 1271626.3    0 |
```

As we can see, fp_fix3536 (this branch) and fp_master (current master) show the same performance whereas the fp_fix3536_assertions (assertion build) shows no additional errors.

PR #3549, opened 2019-12-09
Incorporate rewriting on demand in the evaluator

ajreynol:
This incorporates the rewriter on demand in the evaluator. This is required for terms that mix datatypes which cannot be evaluated and other builtin terms which can.  This is highly useful for terms involving applications of DT_SYGUS_EVAL.

This makes some initial benchmarking on SyGuS with optimization objectives 2.5x faster.

Also adds a few missing cases to the evaluator.


src/theory/evaluator.cpp
```diff
@@ -568,6 +665,22 @@ EvalResult Evaluator::evalInternal(TNode n,
           results[currNode] = EvalResult(res);
           break;
         }
+        case kind::BITVECTOR_UDIV:
```
4tXJ7f: Hmm, technically, this is not correct if the second operand is zero. Maybe we should return `EvalResult()` in that case?

src/theory/evaluator.cpp
```diff
@@ -568,6 +665,22 @@ EvalResult Evaluator::evalInternal(TNode n,
...
+        case kind::BITVECTOR_UDIV_TOTAL:
+        {
+          BitVector res = results[currNode[0]].d_bv;
+          res = res.unsignedDivTotal(results[currNode[1]].d_bv);
+          results[currNode] = EvalResult(res);
+          break;
+        }
+        case kind::BITVECTOR_UREM:
```
4tXJ7f: Hmm, technically, this is not correct if the second operand is zero. Maybe we should return `EvalResult()` in that case?

src/theory/evaluator.cpp
```diff
@@ -190,11 +281,12 @@ EvalResult Evaluator::evalInternal(TNode n,
...
         // requires different substitutions
-        results[currNode] = evalInternal(op[1], lambdaArgs, lambdaVals);
+        results[currNode] =
+            evalInternal(op[1], lambdaArgs, lambdaVals, evalAsNode);
         if (results[currNode].d_tag == EvalResult::INVALID)
         {
-          // evaluation was invalid, we fail
-          return results[currNode];
+          // evaluation was invalid, we take the node of op[1] as the result
+          evalAsNode[currNode] = evalAsNode[op[1]];
```
4tXJ7f: I think that there is an issue here if we have the same lambda twice invoked with different arguments and both evaluations are invalid. Maybe `evalAsNode` should be a `CDHashMap` and we push/pop the context around evaluating lambdas??

src/theory/evaluator.cpp
```diff
@@ -140,30 +158,103 @@ EvalResult Evaluator::evalInternal(TNode n,
...
+            itn = evalAsNode.find(currNodeChild);
+            if (itn != evalAsNode.end())
+            {
+              // take cached value
+              echildren.push_back(itn->second);
+            }
+            else
+            {
+              // must take original
+              echildren.push_back(currNodeChild);
```
4tXJ7f: Aren't we guaranteed that a child is either in `results` or `evalAsNode`?

src/theory/evaluator.cpp
```diff
@@ -140,30 +158,103 @@ EvalResult Evaluator::evalInternal(TNode n,
...
+                           << currNodeVal << std::endl;
+        if (currNodeVal.getNumChildren() > 0)
+        {
+          // We may continue with a valid EvalResult at this point only if
+          // we have no children. We must otherwise fail here since some of
+          // our children may not have successful evaluations.
+          results[currNode] = EvalResult();
+          evalAsNode[currNode] = currNodeVal;
+          continue;
+        }
```
4tXJ7f: Could you add a comment that we try to turn a `currNodeVal` with no children into an `EvalResult`? I was a bit confused by why we continue at first.

src/theory/evaluator.cpp
ajreynol: Yes, technically this is the case (although this is super limited right now: this is if the user forces a pre-smt-lib 2.6 semantics).  I'll make this a special case.

src/theory/evaluator.cpp
ajreynol: Yes, good catch. I'll fix.

src/theory/evaluator.cpp
ajreynol: Yes, I can turn this into an Assert.

PR #3550, opened 2019-12-10
Allow unsat cores with sygus inference

ajreynol:
Fixes the nightlies.


ajreynol:
I initially did but then I realized that block is actually specific to `unconstrainedSimp`.

ajreynol:
Also, btw I just added another necessary fix to a regression.

Issue #3527, opened 2019-12-04
CVC4 incorrect model on UF formula with distinct operator


muchang:
Hi,
For this formula
```
(declare-fun f (Int) Bool)
(declare-fun g (Int) Bool)
(assert (distinct f g))
(check-sat)
(get-model)
```
CVC4 gives an incorrect model:
```
(model
(define-fun f ((BOUND_VARIABLE_294 Int)) Bool false)
(define-fun g ((BOUND_VARIABLE_294 Int)) Bool false)
)
```

OS: Ubuntu 18.04
Revision: f3d40cc

ajreynol:
This file involves higher-order reasoning, you should use `--uf-ho` for problems with higher-order reasoning. With this option, we produce this model:
```
sat
(model
(define-fun f ((BOUND_VARIABLE_313 Int)) Bool false)
(define-fun g ((BOUND_VARIABLE_313 Int)) Bool (= BOUND_VARIABLE_313 0))
)
```
The issue here is we are not throwing an exception when higher-order constraints are used without `--uf-ho`.  We are in the process of fixing this.

ajreynol:
Fixed in #3551.

Issue #3506, opened 2019-11-29
CVC4 --sygus-inference check failure on LIA formula at CVC4/src/theory/theory_model_builder.cpp:65


muchang:
Hi, 
For this formula,
```
(set-logic ALL)
(declare-fun f (Int) Bool)
(declare-fun g (Int) Bool)
(assert (= f g))
(assert (exists ((x Int)) (f x)))
(check-sat)
```
CVC4 with --sygus-inference will throw out a check failure:
```
Fatal failure within bool CVC4::theory::TheoryEngineModelBuilder::isAssignable(CVC4::TNode) at /home/CVC4/src/theory/theory_model_builder.cpp:65
Check failure
 !n.getType().isFunction()
Aborted
```

OS: Ubuntu 18.04
Revision: 5e2d39c

ajreynol:
Since this problem involves equality between functions, you should use `--uf-ho` for this problem. This avoids the assertion failure.

ajreynol:
Fixed in #3551.

PR #3551, opened 2019-12-10
Fix ufho issues

HanielB:
This addresses issues #3506 and #3527


src/parser/smt2/smt2.cpp
```diff
@@ -1825,6 +1824,18 @@ Expr Smt2::applyParseOp(ParseOp& p, std::vector<Expr>& args)
     const Datatype& dt = ((DatatypeType)t).getDatatype();
     return em->mkExpr(kind::APPLY_SELECTOR, dt[0][n].getSelector(), args);
   }
+  else if (!em->getOptions().getUfHo()
+           && (kind == kind::EQUAL || kind == kind::DISTINCT))
```
ajreynol: Will this catch everything? What about functions as arguments to other functions?  Or as arguments to other interpreted functions?

HanielB: I think those will be caught by the type checker. This code catches the case for equality, which is parametric so it's fine in terms of types.

PR #3554, opened 2019-12-10
Fix spurious parse error for rational real array constants

ajreynol:
Currently we can't parse constant arrays that store real values that are given as rationals `(/ n m)`. We throw a spurious parse error for `((as const (Array Int Real)) (/ 1 3))`, indicating that the argument of the array is not constant. This is caused by the fact that `(/ 1 3)` is parsed as a *division* term not a rational value.

This adds a special case to constant array construction so that we compute the result of a constant division instead of using the division term `(/ n m)` when constructing an array constant.

I am open to suggestions on improving this.


ajreynol:
The problem actually applies in other cases:
```
(error "Parse Error: array-rational.smt2:3.50: expected constant term inside array constant, but found nonconstant term:

  (assert (= a ((as const (Array Int Real)) (/ 1 3))))
                                                 ^

the term: (/ 1 3)")
```
We probably need to call some kind of rewriter to ensure the constant is normalized.

PR #3558, opened 2019-12-12
Fix Unif+PI algorithm with symbolic unfolding

HanielB:


HanielB:
I actually feel guilty because I reviewed and approved it with this bug!

PR #3561, opened 2019-12-12
Generate code for options with modes.

mpreiner:
This commit adds support for code generation of options with modes (enums). From now on option enums can be specified in the corresponding *.toml files without the need of extra code.

Further, this moves all enums to the options namespace and shortens the names of the enum values if possible.

@mudathirmahgoub this may be of interest to you for extracting the option modes for the webapp.


src/options/quantifiers_modes.h
```diff
@@ -26,345 +26,9 @@ namespace CVC4 {
 namespace theory {
```
ajreynol: I'm assuming this file will be deleted?

src/decision/decision_engine.cpp
```diff
@@ -53,8 +52,11 @@ void DecisionEngine::init()
   Trace("decision-init") << " * options->decisionStopOnly: "
                          << options::decisionStopOnly() << std::endl;
 
-  if(options::decisionMode() == decision::DECISION_STRATEGY_INTERNAL) { }
-  if(options::decisionMode() == decision::DECISION_STRATEGY_JUSTIFICATION) {
+  if (options::decisionMode() == options::DecisionMode::INTERNAL)
```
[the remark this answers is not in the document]
mpreiner: Done

src/options/bv_options.toml
```diff
@@ -6,46 +6,73 @@ header = "options/bv_options.h"
...
   category   = "expert"
   long       = "bv-optimize-sat-proof=MODE"
-  type       = "CVC4::theory::bv::BvOptimizeSatProof"
-  default    = "CVC4::theory::bv::BITVECTOR_OPTIMIZE_SAT_PROOF_FORMULA"
-  handler    = "stringToBvOptimizeSatProof"
-  predicates = ["satSolverEnabledBuild"]
-  includes   = ["options/bv_bitblast_mode.h"]
+  type       = "BvOptimizeSatProof"
+  default    = "FORMULA"
+#  predicates = ["satSolverEnabledBuild"]
```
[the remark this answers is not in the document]
mpreiner: I added the check here: c8569719fed5588446b3067224ff7d854cd8200a

src/options/decision_options.toml
```diff
@@ -50,8 +60,16 @@ header = "options/decision_options.h"
...
   category   = "expert"
   long       = "decision-weight-internal=HOW"
-  type       = "decision::DecisionWeightInternal"
-  default    = "decision::DECISION_WEIGHT_INTERNAL_OFF"
-  handler    = "stringToDecisionWeightInternal"
+  type       = "DecisionWeightInternal"
+  default    = "OFF"
   read_only  = true
-  help       = "computer weights of internal nodes using children: off, max, sum, usr1 (meaning evolving)"
+  help       = "compute weights of internal nodes using children: off, max, sum, usr1 (meaning evolving)"
```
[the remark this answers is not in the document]
mpreiner: I'm not sure what this was supposed to mean :)

src/options/quantifiers_modes.h
mpreiner: I've deleted it.

src/options/decision_options.toml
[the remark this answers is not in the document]
aniemetz: I think his comment is regarding the `(meaning evolving)` :D

src/theory/rewriter.cpp
```diff
@@ -29,7 +30,7 @@ namespace CVC4 {
 namespace theory {
 
 static TheoryId theoryOf(TNode node) {
-  return Theory::theoryOf(THEORY_OF_TYPE_BASED, node);
+  return Theory::theoryOf(options::TheoryOfMode::THEORY_OF_TYPE_BASED, node);
```
ajreynol: FYI you'll have a merge conflict here (this code does not depend on options::TheoryOfMode::THEORY_OF_TYPE_BASED anymore, keep the code in master unchanged).

src/options/mkoptions.py
```diff
@@ -226,6 +228,64 @@
...
+TPL_IMPL_MODE_CASE = \
+"""
+    case {type}::{enum}:
+      out << "{enum}";
+      break;"""
+
+TPL_DECL_MODE_HANDLER = \
+"""
+{type}
+stringTo{type}(std::string option, std::string optarg);"""
```
4tXJ7f: Nit: Can't we make the `std::string` arguments `const std::string&` arguments here?

src/options/mkoptions.py
```diff
@@ -1437,9 +1448,9 @@ def mkoptions_main():
     # long options are available).
     for module in modules:
         for option in module.options:
-            check_links(option.filename, option.lineno, option.links)
+            check_links(option.filename, 0, option.links)
         for alias in module.aliases:
-            check_links(alias.filename, alias.lineno, alias.links)
+            check_links(alias.filename, 0, alias.links)
```
4tXJ7f: I think the two calls to `check_links` above are the only ones. Should we maybe remove the `lineno` argument from the function?

src/options/mkoptions.py
```diff
@@ -1428,7 +1439,7 @@ def mkoptions_main():
                     m[0].alternate = False
                     alias.alternate_for = m[0]
                     del g_long_cache[alias.long]
-            check_long(alias.filename, alias.lineno, alias.long)
+            check_long(alias.filename, 0, alias.long)
```
4tXJ7f: I think the two calls to `check_long` changed in this PR are the only ones. Should we maybe remove the lineno argument from the function?

src/options/mkoptions.py
```diff
@@ -1235,122 +1319,50 @@ def parse_module(filename, file):
...
-        alias.filename = filename
-        res.aliases.append(alias)
+    if 'option' in module:
+        for attribs in module['option']:
+            lineno = 0
+            check_attribs(filename, lineno,
+                          OPTION_ATTR_REQ, OPTION_ATTR_ALL, attribs, 'option')
+            option = Option(attribs)
+            if option.mode and not option.help_mode:
+                perr(filename, lineno,
```
4tXJ7f: IMHO we should change `perr` to take `lineno` as an optional argument (and not print out the line number in that case). No line number is better than a wrong one.

mpreiner: Totally forgot about that argument. I actually wanted to change the error reporting since we don't have the line number information anymore. Now you can optionally pass an option/alias object from which a name/long option name is extracted for error reporting.

src/options/mkoptions.py
mpreiner: See comment below

src/options/mkoptions.py
mpreiner: Ditto

src/theory/rewriter.cpp
mpreiner: Done, thanks for the info!

ajreynol:
This looks great to me.  I don't see any issues with the option updates.  I'll defer to Andres for final approval.

src/options/mkoptions.py
```diff
@@ -346,8 +346,19 @@ def die(msg):
...
+        if isinstance(option_or_alias, Option):
+            msg_prefix = 'option '
+            if option_or_alias.name:
+                msg_prefix = "{} '{}' ".format(msg_prefix, option_or_alias.name)
+            else:
+                msg_prefix = "{} '{}' ".format(msg_prefix, option_or_alias.long)
+        else:
+            assert isinstance(option_or_alias, Alias)
+            msg_prefix = "alias '{}' ".format(option_or_alias.long)
+    die('parse error in {}: {}{}'.format(filename, msg, msg_prefix))
```
4tXJ7f: The `msg_prefix` looks more like a `msg_suffix` to me :)

mpreiner: Haha true, it was a prefix when I started but then made a suffix out of it :-)

PR #3562, opened 2019-12-12
Eliminate Expr-level calls in TypeNode

ajreynol:
This eliminates expression-level calls from TypeNode, namely `getDatatype`, `isRecord` and `getRecord`.

It adds TypeNode to Type calls in a few places where it is necessary to access the expression level.  

The next steps will be to update NodeManager/ExprManager so that DType/Datatype are stored at the right levels.

There are also a number of places (e.g. SygusGrammarNorm) where is not yet possible to convert to the Node-level due to some missing utilities (e.g. sygus print callbacks) in the node layer.  This will need to be addressed in later PRs.

### DIFF
--- a/src/expr/type.cpp
+++ b/src/expr/type.cpp
@@ -99,7 +99,7 @@ bool Type::isFunctionLike() const
 Expr Type::mkGroundTerm() const {
   NodeManagerScope nms(d_nodeManager);
   Expr ret = d_typeNode->mkGroundTerm().toExpr();
-  PrettyCheckArgument(ret != NULL, this, "Cannot construct ground term!");
+  PrettyCheckArgument(ret.isNull(), this, "Cannot construct ground term!");
   return ret;
 }
 
--- a/src/expr/type.cpp
+++ b/src/expr/type.cpp
@@ -99,7 +99,10 @@ bool Type::isFunctionLike() const
 Expr Type::mkGroundTerm() const {
   NodeManagerScope nms(d_nodeManager);
   Expr ret = d_typeNode->mkGroundTerm().toExpr();
-  PrettyCheckArgument(ret.isNull(), this, "Cannot construct ground term!");
+  if (ret.isNull())
+  {
+    IllegalArgument(this, "Cannot construct ground term!");
+  }
   return ret;
 }
 
--- a/src/expr/type.cpp
+++ b/src/expr/type.cpp
@@ -579,7 +579,7 @@ const Datatype& DatatypeType::getDatatype() const {
     return d_nodeManager->getDatatypeForIndex(dic.getIndex());
   }
   Assert(d_typeNode->getKind() == kind::PARAMETRIC_DATATYPE);
-  return static_cast<DatatypeType>((*d_typeNode)[0].toType()).getDatatype();
+  return DatatypeType((*d_typeNode)[0].toType()).getDatatype();
 }
 
 bool DatatypeType::isParametric() const {
--- a/src/expr/type.cpp
+++ b/src/expr/type.cpp
@@ -98,7 +98,9 @@ bool Type::isFunctionLike() const
 
 Expr Type::mkGroundTerm() const {
   NodeManagerScope nms(d_nodeManager);
-  return d_typeNode->mkGroundTerm().toExpr();
+  Expr ret = d_typeNode->mkGroundTerm().toExpr();
+  PrettyCheckArgument(ret != NULL, this, "Cannot construct ground term!");
+  return ret;
 }
 
 Expr Type::mkGroundValue() const
--- a/src/expr/type.cpp
+++ b/src/expr/type.cpp
@@ -332,7 +332,7 @@ bool Type::isTuple() const {
 bool Type::isRecord() const {
   NodeManagerScope nms(d_nodeManager);
   return d_typeNode->getKind() == kind::DATATYPE_TYPE
-         && static_cast<DatatypeType>(*this).getDatatype().isRecord();
+         && DatatypeType(*this).getDatatype().isRecord();
 }
 
 /** Is this a symbolic expression type? */
--- a/src/expr/type.cpp
+++ b/src/expr/type.cpp
@@ -326,7 +326,8 @@ bool Type::isTuple() const {
 /** Is this a record type? */
 bool Type::isRecord() const {
   NodeManagerScope nms(d_nodeManager);
-  return d_typeNode->isRecord();
+  return d_typeNode->getKind() == kind::DATATYPE_TYPE
+         && static_cast<DatatypeType>(*this).getDatatype().isRecord();
 }
 
 /** Is this a symbolic expression type? */
@@ -566,7 +567,14 @@ std::vector<Type> ConstructorType::getArgTypes() const {
 
 const Datatype& DatatypeType::getDatatype() const {
   NodeManagerScope nms(d_nodeManager);
-  return d_typeNode->getDatatype();
+  Assert(isDatatype());
+  if (d_typeNode->getKind() == kind::DATATYPE_TYPE)
+  {
+    DatatypeIndexConstant dic = d_typeNode->getConst<DatatypeIndexConstant>();
+    return d_nodeManager->getDatatypeForIndex(dic.getIndex());
+  }
+  Assert(d_typeNode->getKind() == kind::PARAMETRIC_DATATYPE);
+  return static_cast<DatatypeType>((*d_typeNode)[0].toType()).getDatatype();
 }
 
 bool DatatypeType::isParametric() const {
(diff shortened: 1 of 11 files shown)


src/theory/quantifiers/sygus/sygus_grammar_norm.cpp
```diff
@@ -599,7 +599,7 @@ TypeNode SygusGrammarNorm::normalizeSygusRec(TypeNode tn)
     return tn;
   }
   /* Collect all operators for normalization */
-  const Datatype& dt = tn.getDatatype();
+  const Datatype& dt = static_cast<DatatypeType>(tn.toType()).getDatatype();
```
4tXJ7f: Strictly speaking, this is undefined behavior because `tn.toType()` is not of type `DatatypeType` (UBSan should complain about this). The correct way is:

```suggestion
  const Datatype& dt = DatatypeType(tn.toType()).getDatatype();
```

src/expr/type.cpp
```diff
@@ -326,7 +328,8 @@ bool Type::isTuple() const {
 /** Is this a record type? */
 bool Type::isRecord() const {
   NodeManagerScope nms(d_nodeManager);
-  return d_typeNode->isRecord();
+  return d_typeNode->getKind() == kind::DATATYPE_TYPE
+         && static_cast<DatatypeType>(*this).getDatatype().isRecord();
```
4tXJ7f: Strictly speaking, this is undefined behavior because `*this` is not of type `DatatypeType` (UBSan should complain about this). The correct way is:

```suggestion
         && DatatypeType(*this).getDatatype().isRecord();
```

src/expr/type.cpp
```diff
@@ -566,7 +569,14 @@ std::vector<Type> ConstructorType::getArgTypes() const {
...
   NodeManagerScope nms(d_nodeManager);
-  return d_typeNode->getDatatype();
+  Assert(isDatatype());
+  if (d_typeNode->getKind() == kind::DATATYPE_TYPE)
+  {
+    DatatypeIndexConstant dic = d_typeNode->getConst<DatatypeIndexConstant>();
+    return d_nodeManager->getDatatypeForIndex(dic.getIndex());
+  }
+  Assert(d_typeNode->getKind() == kind::PARAMETRIC_DATATYPE);
+  return static_cast<DatatypeType>((*d_typeNode)[0].toType()).getDatatype();
```
4tXJ7f: Strictly speaking, this is undefined behavior because `(*d_typeNode)[0].toType()` is not of type `DatatypeType` (UBSan should complain about this). The correct way is:

```suggestion
  return DatatypeType((*d_typeNode)[0].toType()).getDatatype();
```

src/printer/cvc/cvc_printer.cpp
```diff
@@ -373,8 +373,10 @@ void CvcPrinter::toStream(
           if( n.getNumChildren()==1 ){
             out << "TUPLE";
           }
-        }else if( t.isRecord() ){
-          const Record& rec = t.getRecord();
+        }
+        else if (t.toType().isRecord())
+        {
+          const Record& rec = static_cast<DatatypeType>(t.toType()).getRecord();
```
4tXJ7f: (see other comments about `static_cast<DatatypeType>`)

```suggestion
          const Record& rec = static_cast<DatatypeType>(t.toType()).getRecord();
```

src/printer/cvc/cvc_printer.cpp
```diff
@@ -404,7 +408,12 @@ void CvcPrinter::toStream(
     case kind::APPLY_SELECTOR_TOTAL: {
         TypeNode t = n[0].getType();
         Node opn = n.getOperator();
-        if (t.isTuple() || t.isRecord())
+        if (!t.isDatatype())
+        {
+          toStream(op, opn, depth, types, false);
+        }
+        else if (t.isTuple()
+                 || static_cast<DatatypeType>(t.toType()).isRecord())
```
4tXJ7f: (see other comments about `static_cast<DatatypeType>`)

```suggestion
                 || DatatypeType(t.toType()).isRecord())
```

src/theory/datatypes/theory_datatypes.cpp
```diff
@@ -631,8 +631,9 @@ Node TheoryDatatypes::expandDefinition(LogicRequest &logicRequest, Node n) {
       }
       else
       {
-        Assert(t.isRecord());
-        const Record& record = t.getRecord();
+        Assert(t.toType().isRecord());
+        const Record& record =
+            static_cast<DatatypeType>(t.toType()).getRecord();
```
4tXJ7f: (see other comments about `static_cast<DatatypeType>`)

```suggestion
            DatatypeType(t.toType()).getRecord();
```

src/theory/datatypes/theory_datatypes_type_rules.h
```diff
@@ -287,11 +287,13 @@ struct RecordUpdateTypeRule {
...
     if (check) {
-      if (!recordType.isRecord()) {
+      if (!recordType.toType().isRecord())
+      {
         throw TypeCheckingExceptionPrivate(
             n, "Record-update expression formed over non-record");
       }
-      const Record& rec = recordType.getRecord();
+      const Record& rec =
+          static_cast<DatatypeType>(recordType.toType()).getRecord();
```
4tXJ7f: (see other comments about `static_cast<DatatypeType>`)

```suggestion
          DatatypeType(recordType.toType()).getRecord();
```

src/theory/quantifiers/sygus/sygus_grammar_cons.cpp
```diff
@@ -1175,14 +1175,16 @@ void CegGrammarConstructor::mkSygusDefaultGrammar(
...
       cargsTester.push_back(unres_types[iuse]);
       for (unsigned k = 0, size_k = dt.getNumConstructors(); k < size_k; ++k)
       {
-        Trace("sygus-grammar-def") << "...for " << dt[k].getTesterName() << std::endl;
-        sdtBool.addConstructor(
-            dt[k].getTester(), dt[k].getTesterName(), cargsTester);
+        Trace("sygus-grammar-def")
+            << "...for " << dt[k].getTester() << std::endl;
+        std::stringstream sst;
+        sst << dt[k].getTester();
```
4tXJ7f: This is a bit ugly. Maybe we should also have `DTypeConstructor::getTesterName()`?

src/theory/quantifiers/sygus/sygus_grammar_norm.cpp
ajreynol: It seems our coding guidelines say to use `static_cast` (https://github.com/CVC4/CVC4/wiki/Code-Style-Guidelines#other-c-features). Everywhere a cast of this form is done, I ensure that `tn.toType()` is of `DatatypeType`.

src/theory/quantifiers/sygus/sygus_grammar_cons.cpp
ajreynol: Tester names are a user-level piece of information to support deprecated (pre SMT-LIB 2.6) custom tester names. We removed this field from the Node-level constructor, I would not want to readd this.

src/theory/quantifiers/sygus/sygus_grammar_norm.cpp
4tXJ7f: `TypeNode::toType()` returns an object of type `Type` (i.e. it is not a `DatatypeType` anymore even if the original was because it has been sliced). The cast would be ok if it were on a `Type*` (`static_cast<DatatypeType*>`). I am suggesting to use the `DatatypeType` constructor for the "cast" because that's how our `Type` hierarchy is designed (I don't necessarily agree with the design but that's how it is). See also PR #3420 where I fixed existing casts because UBSan complained.

src/theory/quantifiers/sygus/sygus_grammar_cons.cpp
4tXJ7f: Hmm, that's fair but how is this going to work when we remove the `Expr` layer? Who will maintain that information and how do we access it?

ajreynol: I think we should remove that field.  Testers no longer have indepdent names in SMT-LIB, instead their syntax is e.g. `(_ is cons)`.

src/theory/quantifiers/sygus/sygus_grammar_norm.cpp
ajreynol: I see, this is indeed confusing.  It's annoying that there is an ambiguity between cast/constructor.  
In light of this, I agree with your changes; will update.

src/theory/quantifiers/sygus/sygus_grammar_cons.cpp
4tXJ7f: That sounds good to me.

ajreynol:
Marking "do not merge" since there seems to be a strange issue with github where this PR is already merged.

PR #3566, opened 2019-12-13
Improve polynomial anyterm grammar

HanielB:
This changes the grammar construction in the case of anyterm + polynomial grammar so that the predicates `EQUAL` and `LEQ` take as arguments the tuple `(ANYTERM, ZEROINT)` rather then `(ANYTERM, ANYTERM)`, in which `ZEROINT` is a type solely with the 0 constant.

This avoid enumerating spurious terms and leads to a 10x improvement in one regression.


src/theory/quantifiers/sygus/sygus_grammar_cons.cpp
```diff
@@ -1079,6 +1085,20 @@ void CegGrammarConstructor::mkSygusDefaultGrammar(
       // This ensures that ( c1*x + c2*y >= 0 ) has the same weight as
       // e.g. ( x >= 0 ) or ( y >= 0 ).
       sdts[iat].d_sdt.addConstructor(op, "polynomial", cargsAnyTerm, spc, 0);
+      // define zero integer dtype
```
ajreynol: Does this have to be a zero type?  I thought of doing this but I think we can just build `0` as part of the sygus operator for the Boolean.  As in, its sygus constructor becomes (lambda cx cy.  (cx*x + cy*y >= 0) instead of (lambda cx cy czero.  cx*x + cy*y >= czero).

HanielB: You know, I spent a ridiculous amount of time remembering and getting right the thing of adding new types / datatypes / initializing them properly and then as I submitted I thought "maybe I could just have done a lambda"? You're right, I'll fix.

HanielB: You know, I spent a ridiculous amount of time remembering and getting right the thing of adding new types / datatypes / initializing them properly and then as I submitted I thought "maybe I could just have done a lambda"? You're right, I'll fix.

ajreynol: Haha I know the feeling :)

ajreynol:
Any updates on this?

src/theory/quantifiers/sygus/sygus_grammar_cons.cpp
```diff
@@ -1132,6 +1167,9 @@ void CegGrammarConstructor::mkSygusDefaultGrammar(
...
```
ajreynol, line 1170: Remove this comment (about type).

src/theory/quantifiers/sygus/sygus_grammar_cons.h
```diff
@@ -248,8 +248,21 @@ class CegGrammarConstructor
...
+                                         Node templ_arg,
+                                         TypeNode templ_arg_sygus_type,
+                                         Node bvl,
+                                         const std::string& fun,
+                                         unsigned& tcount);
+
+  /**
+   * Given a kind k, create a lambda operator with the given builtin input type
+   * and an extra zero argument of that same type. Also creates a print
+   * callback since we do not want to print the lambda.
```
ajreynol, line 261: Can you give a quick example here?

PR #3567, opened 2019-12-13
Update sygus grammar normalization to use node-level datatype.

ajreynol:
This updates the sygus grammar normalization class to use the node-level datatype.

There is one exception: since sygus print callbacks are Expr-layer, we need extract it from Datatype in one place.


src/theory/quantifiers/sygus/sygus_grammar_norm.cpp
```diff
@@ -208,31 +210,29 @@ void SygusGrammarNorm::TypeObject::addConsInfo(SygusGrammarNorm* sygus_norm,
   }
 
   std::vector<TypeNode> consTypes;
-  for (const DatatypeConstructorArg& arg : cons)
+  const std::vector<std::shared_ptr<DTypeSelector> >& args = cons.getArgs();
+  for (const std::shared_ptr<DTypeSelector>& arg : args)
```
HanielB: You changed because it is more efficient to do it like this?

ajreynol: No, the previous form of iteration is no longer available (we are switching Expr -> Node here), so this was necessary to make it compile.

PR #3568, opened 2019-12-13
Move Datatype management to ExprManager

ajreynol:
This is further work towards decoupling the Expr layer from the Node layer.

This PR makes it so that ExprManager does memory management for Datatype while NodeManager maintains a list of DType.

As a reminder, the ownership policy (and level of indirection through DatatypeIndex) is necessary due to not being able to store Datatype within Node since this leads to circular dependencies in the Node AST.

### DIFF
--- a/src/expr/expr_manager_template.h
+++ b/src/expr/expr_manager_template.h
@@ -80,7 +80,7 @@ private:
   ExprManager(const ExprManager&) = delete;
   ExprManager& operator=(const ExprManager&) = delete;
 
-  /** A list of datatypes owned by this node manager. */
+  /** A list of datatypes owned by this expr manager. */
   std::vector<std::unique_ptr<Datatype> > d_ownedDatatypes;
 
  public:
--- a/src/expr/expr_manager_template.h
+++ b/src/expr/expr_manager_template.h
@@ -573,6 +573,11 @@ private:
   /** Returns the maximum arity of the given kind. */
   static unsigned maxArity(Kind kind);
 
+  /**
+   * Return the datatype at the given index owned by this class. Type nodes are
+   * associated with datatypes through the DatatypeIndexConstant class. The
+   * argument index is intended to be a value taken from that class.
+   */
   const Datatype& getDatatypeForIndex(unsigned index) const;
 };/* class ExprManager */
 
--- a/src/expr/expr_manager_template.h
+++ b/src/expr/expr_manager_template.h
@@ -84,7 +84,6 @@ private:
   std::vector<std::unique_ptr<Datatype> > d_ownedDatatypes;
 
  public:
-
   /**
    * Creates an expression manager with default options.
    */
--- a/src/expr/expr_manager_template.h
+++ b/src/expr/expr_manager_template.h
@@ -79,8 +79,11 @@ private:
   // undefined, private copy constructor and assignment op (disallow copy)
   ExprManager(const ExprManager&) = delete;
   ExprManager& operator=(const ExprManager&) = delete;
-  
-public:
+
+  /** A list of datatypes owned by this node manager. */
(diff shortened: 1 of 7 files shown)


src/printer/cvc/cvc_printer.cpp
```diff
@@ -179,7 +180,9 @@ void CvcPrinter::toStream(
       break;
 
     case kind::DATATYPE_TYPE: {
-      const Datatype& dt = (NodeManager::currentNM()->getDatatypeForIndex( n.getConst< DatatypeIndexConstant >().getIndex() ));
+      const Datatype& dt =
```
4tXJ7f: Couldn't we change this code to use the `Node` layer?

src/expr/node_manager.h
```diff
@@ -178,7 +178,7 @@ class NodeManager {
   std::vector<NodeManagerListener*> d_listeners;
 
   /** A list of datatypes owned by this node manager. */
-  std::vector<Datatype*> d_ownedDatatypes;
+  std::vector<std::shared_ptr<DType> > d_ownedDTypes;
```
4tXJ7f: I don't quite see a reason for this to be a `shared_ptr`: We only seem to be returning references to the `DType`s owned by this vector, so no one else can actually increment the reference count of the `shared_ptr`s (so we might as well just use `unique_ptr`s).

src/expr/node_manager.cpp
```diff
@@ -248,23 +240,17 @@ NodeManager::~NodeManager() {
   d_options = NULL;
 }
 
-unsigned NodeManager::registerDatatype(Datatype* dt) {
-  unsigned sz = d_ownedDatatypes.size();
-  d_ownedDatatypes.push_back( dt );
+unsigned NodeManager::registerDatatype(std::shared_ptr<DType> dt)
```
4tXJ7f: Since we are anyway modifying this function, can we switch to `size_t` instead of `unsigned`?

src/expr/expr_manager_template.h
```diff
@@ -79,9 +79,11 @@ class CVC4_PUBLIC ExprManager {
   // undefined, private copy constructor and assignment op (disallow copy)
   ExprManager(const ExprManager&) = delete;
   ExprManager& operator=(const ExprManager&) = delete;
-  
-public:
 
+  /** A list of datatypes owned by this node manager. */
```
4tXJ7f:
```suggestion
  /** A list of datatypes owned by this expr manager. */
```

src/printer/cvc/cvc_printer.cpp
ajreynol: No, we have to consider whether the datatype was declared to be a `Record`, which is a user-level only field.

src/expr/node_manager.h
ajreynol: I'm not sure how I could change this currently.  NodeManager actually doesn't do the creation, ExprManager does (and in fact the DType is a shared_ptr within Datatype).  My understanding is one can't convert shared_ptr to unique_ptr.  Or more concretely, what change are you suggesting that would allow this?

src/printer/cvc/cvc_printer.cpp
4tXJ7f: Ah right, `isRecord()` has been removed from the Node layer.

src/expr/node_manager.h
4tXJ7f: We've discussed this quite a bit offline and we've come to the conclusion that it is easiest to leave it as it is now (there is a shared ownership between `Datatype::d_internal` and `NodeManager::d_ownedDTypes`). Ideally, registered `DType`s would only be owned by the `NodeManager` but this will require some additional refactoring that is out-of-scope.

PR #3575, opened 2019-12-16
Fix evaluator for non-evaluatable nodes

ajreynol:
This ensures that the Evaluator always returns the result of substitution + rewriting for constant substitutions.

This requires a few further extensions to the code, namely:
(1) Applying substutitions to operators,
(2) Reconstructing all nodes that fail to evaluate by taking into account evaluation of their children.

### DIFF
--- a/src/theory/evaluator.cpp
+++ b/src/theory/evaluator.cpp
@@ -212,7 +212,7 @@ EvalResult Evaluator::evalInternal(
         // Reconstruct the node with a combination of the children that
         // successfully evaluated, and the children that did not.
         Trace("evaluator") << "Evaluator: collect arguments" << std::endl;
-        currNodeVal = reconstruct(currNode, results, evalAsNode);
+        currNodeVal = reconstruct(currNodeVal, results, evalAsNode);
         Trace("evaluator") << "Evaluator: now after substitution + rewriting: "
                            << currNodeVal << std::endl;
         if (currNodeVal.getNumChildren() > 0)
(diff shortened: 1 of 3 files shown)


src/theory/evaluator.cpp
```diff
@@ -121,7 +121,21 @@ Node Evaluator::eval(TNode n,
   Trace("evaluator") << "Evaluating " << n << " under substitution " << args
                      << " " << vals << std::endl;
   std::unordered_map<TNode, Node, NodeHashFunction> evalAsNode;
-  return evalInternal(n, args, vals, evalAsNode).toNode();
+  Node ret = evalInternal(n, args, vals, evalAsNode).toNode();
+  // if we failed to evaluate
+  if (ret.isNull())
+  {
+    // maybe it was stored in the evaluation-as-node map
```
4tXJ7f:
```suggestion
    // should be stored in the evaluation-as-node map
```

src/theory/evaluator.cpp
```diff
@@ -235,24 +236,40 @@ EvalResult Evaluator::evalInternal(
...
         {
+          // variable with no substitution is itself
           evalAsNode[currNode] = currNode;
           results[currNode] = EvalResult();
           continue;
         }
         ptrdiff_t pos = std::distance(args.begin(), it);
         currNodeVal = vals[pos];
+        // Don't need to reconstruct since range of substitution should already
+        // be normalized.
```
4tXJ7f: Maybe we should add an assertion `vals[pos] == Rewriter::rewrite(vals[pos])`?

PR #3576, opened 2019-12-16
Use the evaluator utility in the function definition evaluator

ajreynol:
Improves performance on ground conjectures with recursive functions.  We use the evalutator to (partially) evaluate bodies of recursive functions, instead of relying on substitution+rewriting.


ajreynol:
It's simple because of the work that the Evaluator is doing underneath now :)

PR #3577, opened 2019-12-17
Increment Taylor degree for tangent and secant plane inferences for transcendentals

### LINKED ISSUE #3564: Sygus inference, exp soundness bug
Hi,
for the following formula CVC4 with `--sygus-inference` return `unsat` but `sat` without that option.  

```
(assert (or (< 60.3 (exp 4.1) 60.4) (< (exp 5.1) 164.1)))
(check-sat)
```
Revision: 8afa8a8
OS: Ubuntu 18.04

ajreynol:
A previous fix ensured we increment the Taylor degree to ensure that bounds on `exp` were sound when computing model via `getPolynomialApproximationBoundForArg`.  This PR extends the fix to when tangent/secant plane inferences are computed.  This fixes #3564.


src/theory/arith/nonlinear_extension.cpp
```diff
@@ -3042,21 +3042,22 @@ bool NonlinearExtension::checkTfTangentPlanesFun(Node tf,
...
+  Node c = d_model.computeAbstractModelValue(tf[0]);
+  int csign = c.getConst<Rational>().sgn();
+  Assert(csign == 1 || csign == -1);
+
   // Figure 3: P_l, P_u
   // mapped to for signs of c
   std::map<int, Node> poly_approx_bounds[2];
   std::vector<Node> pbounds;
-  getPolynomialApproximationBounds(k, d, pbounds);
+  getPolynomialApproximationBoundForArg(k, c, d, pbounds);
```
ahmed-irfan: A side remark: It is not related to this PR, but the name of this function is quite confusing. 
I understood after going through the function body.

PR #3578, opened 2019-12-18
Define all options modified by ENABLE_BEST using cvc4_option

no-preserve-root:
The `ENABLE_BEST` option in the build system does not work correctly. `USE_CADICAL` and `USE_CRYPTOMINISAT` are defined by `option()`, not `cvc4_option()` and are therefore not tri-state. The `ENABLE_BEST` then assumes the variable has been manually set and leaves the options disabled.

This patch uses `cvc4_option()` for both.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -142,19 +142,19 @@ option(ENABLE_PROFILING        "Enable support for gprof profiling")
 # >> 3-valued: IGNORE ON OFF
 #    > allows to detect if set by user (default: IGNORE)
 #    > only necessary for options set for ENABLE_BEST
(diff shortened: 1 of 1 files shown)


4tXJ7f:
@no-preserve-root Thank you for your contribution. Could you please also [sign the commit with `git commit -s`](https://github.com/CVC4/CVC4/blob/master/CONTRIBUTING.md#contributing) because if we have to rebase the commit, your GPG signature gets lost (we don't use merge commits).

no-preserve-root:
@4tXJ7f the commit is now Signed-off.

PR #3581, opened 2019-12-23
Initial support for string reverse

ajreynol:
Type rules, parsing and printing, basic rewriting including constant evaluation, reduction for string reverse (`str.rev`).

Also improves support in a few places for tolower/toupper.


test/regress/regress0/strings/str-rev-simple.smt2
```diff
@@ -0,0 +1,11 @@
...
+(set-info :status sat)
+(declare-fun x () String)
+(declare-fun y () String)
+(declare-fun z () String)
+(assert (= (str.rev "ABC") "CBA"))
+(assert (= (str.rev "") ""))
+(assert (= (str.rev "A") x))
+(assert (= (str.rev (str.++ x y)) "BCA"))
+(assert (= (str.rev "BC") y))
+(check-sat)
```
4tXJ7f: Maybe I am missing something but this looks like essentially the same benchmark as test/regress/regress1/strings/str-rev-simple-s.smt2 ? Is there a reason for having both? Also, shouldn't we require `--strings-exp` for `str.rev`?

src/theory/strings/theory_strings_rewriter.cpp
```diff
@@ -3217,6 +3227,39 @@ Node TheoryStringsRewriter::rewriteStrConvert(Node node)
...
+  if (x.isConst())
+  {
+    std::vector<unsigned> nvec = node[0].getConst<String>().getVec();
+    std::reverse(nvec.begin(), nvec.end());
+    Node retNode = nm->mkConst(String(nvec));
+    return returnRewrite(node, retNode, "str-conv-const");
+  }
+  else if (x.getKind() == STRING_CONCAT)
+  {
+    std::vector<Node> children;
```
4tXJ7f: Could use `NodeBuilder` here.

src/theory/strings/theory_strings_preprocess.cpp
```diff
@@ -514,7 +514,40 @@ Node StringsPreprocess::simplify( Node t, std::vector< Node > &new_nodes ) {
...
+    Node i = nm->mkBoundVar(nm->integerType());
+    Node bvi = nm->mkNode(BOUND_VAR_LIST, i);
+
+    Node revi = nm->mkNode(
+        MINUS, nm->mkNode(STRING_LENGTH, x), nm->mkNode(PLUS, i, d_one));
+    Node ssr = nm->mkNode(STRING_SUBSTR, r, i, d_one);
+    Node ssx = nm->mkNode(STRING_SUBSTR, x, revi, d_one);
+
+    Node bound =
+        nm->mkNode(AND, nm->mkNode(LEQ, d_zero, i), nm->mkNode(LEQ, i, lenr));
```
4tXJ7f:
```suggestion
        nm->mkNode(AND, nm->mkNode(LEQ, d_zero, i), nm->mkNode(LT, i, lenr));
```

src/theory/strings/theory_strings_preprocess.cpp
```diff
@@ -514,7 +514,40 @@ Node StringsPreprocess::simplify( Node t, std::vector< Node > &new_nodes ) {
...
+    Node rangeA = nm->mkNode(
+        FORALL, bvi, nm->mkNode(OR, bound.negate(), ssr.eqNode(ssx)));
+    // assert:
+    //   len(r) = len(x) ^
+    //   forall i. 0 <= i < len(r) =>
+    //     substr(r,i,1) == substr(x,len(x)-(i+1),1)
```
4tXJ7f:
```suggestion
    //     substr(r,i,1) = substr(x,len(x)-(i+1),1)
```

(consistency)

test/regress/regress0/strings/str-rev-simple.smt2
ajreynol: The last assertion is different: here, we say y is the reverse of "BC", in the other we say that y is some string whose reverse is "BC".  The latter is more difficult at the moment (without using an injectivity argument) and requires --string-exp.

src/theory/strings/theory_strings_rewriter.cpp
ajreynol: I did not use it since we reverse the child vector below.  Is there a way to reverse the child vector of a NodeBuilder?

4tXJ7f: Oops, you are right of course.

test/regress/regress0/strings/str-rev-simple.smt2
4tXJ7f: I see. Thanks for the explanation.

PR #3583, opened 2019-12-23
Update any-constant and normalization policies for sygus grammars

### LINKED ISSUE #3580: --sygus-steam prevents finding solution
The following SyGuS problem does not seem to be handled well when running with `--sygus-stream`.  Removing `f` (which, trivially, must always just return `true`) results in drastically better performance:
```
(set-logic ALL)
(synth-fun f 
	() Bool 
	((B Bool)) 
	(
		(B Bool (true))
	)
)
(synth-fun g 
	((r Int)) Bool 
	((B Bool) (I Int) (IConst Int)) 
	(
		(B Bool ((= I I) (=> B B)))
		(I Int (r 0 (mod I IConst)))
		(IConst Int ((Constant Int)))
	)
) 
(constraint (g 0))
(constraint (not (g 1)))
(constraint (g 2))
(constraint f)
(check-synth)
```
with only the flag `--lang=sygus2`, CVC4 finds the following solution in <.1s:
```
(define-fun f () Bool true)
(define-fun g ((r Int)) Bool (= 0 (mod r 2)))
```

When the  `--sygus-stream` flag is added CVC4 does not return a solution within 10 minutes.

**Different performance with changes:**
Removing `=>` from the grammar of `g` results in `--sygus-stream` returning `unknown`.  Removing `f`:
```
(set-logic ALL)
(synth-fun g 
	((r Int)) Bool 
	((B Bool) (I Int) (IConst Int)) 
	(
		(B Bool ((= I I) (=> B B)))
		(I Int (r 0 (mod I IConst)))
		(IConst Int ((Constant Int)))
	)
) 
(constraint (g 0))
(constraint (not (g 1)))
(constraint (g 2))
(check-synth)
```
results in `--sygus-stream` finding several solution almost instantly:
```
time cvc4 test.sy --lang=sygus2 --sygus-stream
(define-fun g ((r Int)) Bool (= 0 (mod r 2)))
(define-fun g ((r Int)) Bool (= 0 (mod (mod r 2) 2)))
(define-fun g ((r Int)) Bool (= 0 (mod (mod r 2) 3)))
(define-fun g ((r Int)) Bool (= 0 (mod (mod r 3) 2)))
(define-fun g ((r Int)) Bool (=> (= r (mod r 2)) (= r 0)))
(define-fun g ((r Int)) Bool (= 0 (mod (mod r 2) 4)))
(define-fun g ((r Int)) Bool (= 0 (mod (mod r 2) 5)))
(define-fun g ((r Int)) Bool (= 0 (mod (mod r 2) 6)))
(define-fun g ((r Int)) Bool (= 0 (mod (mod r 2) 7)))
(define-fun g ((r Int)) Bool (= 0 (mod (mod r 2) 8)))
(define-fun g ((r Int)) Bool (= 0 (mod (mod r 2) 9)))
(define-fun g ((r Int)) Bool (= 0 (mod (mod r 2) 10)))
(define-fun g ((r Int)) Bool (= 0 (mod (mod r 2) 11)))
(define-fun g ((r Int)) Bool (= 0 (mod (mod r 2) 12)))
(define-fun g ((r Int)) Bool (= 0 (mod (mod r 2) 13)))
(define-fun g ((r Int)) Bool (= 0 (mod (mod r 2) 14)))
(define-fun g ((r Int)) Bool (= 0 (mod (mod r 2) 15)))
(define-fun g ((r Int)) Bool (= 0 (mod (mod r 4) 2)))
(define-fun g ((r Int)) Bool (= 0 (mod (mod r 5) 2)))
(define-fun g ((r Int)) Bool (= 0 (mod (mod r 6) 2)))
```

ajreynol:
Previously, we had a configurable flag "useSymbolicCons" indicating, per enumerator, whether we want to use any-constant constructors.  For some strategies e.g. PBE, this flag would be disabled by default.  The problem is that in certain circumstances this can make the overall conjecture infeasible if any-constant was user-provided and *necessary* for a solution. This was the case in #3580.

After this PR, we always use the any-constant constructor if it is user-provided, for any enumerator.  For default grammars, the use of any-constant constructors is determined by `--sygus-grammar-cons` (they are currently added if `--sygus-grammar-cons!=simple`).  To ensure this is the case, the policy for grammar normalization has been updated to not apply to internally constructed grammars.  This is due to the fact that normalization *adds* any-constants which previously were blocked by symmetry breaking lemmas when useSymbolicCons=false; now they are not added in the first place for internally generated grammars when `--sygus-grammar-cons=simple` (default).

This fixes #3580.


HanielB:
I think there is an argument to be made that using `Constant` in the grammar does not necessarily entail using the symbolic constructor. We could do as we did before, using default constants and operations to generate more. What do you think?

ajreynol:
Yes I agree we should be able to do this as a heuristic.

Note that it's a bit tricky to do what you said properly: if you have a grammar with one production rule (Constant Int) then strictly speaking (+ 1 1)  does not meet the spec so we'd have to normalize to 2 as a postprocessing step.  This is a bit more involved and could be done on another PR.

HanielB:
OK, that sounds good.

PR #3597, opened 2020-01-10
Fix printing of models of uninterpreted sorts

### LINKED ISSUE #3596: Invalid syntax in generated model for type of cardinality one
[model.smt2.txt](https://github.com/CVC4/CVC4/files/4045155/model.smt2.txt)
On the attached file, cvc4 generates a model which contains lines of this form:
```
; cardinality of integer is 1
(declare-sort integer 0)
(declare-fun  () integer)
```
The last line looks like incorrect syntax, and indeed cvc4 doesn't accept this syntax itself.

This is on current master.

4tXJ7f:
Fixes #3596. When printing the representatives of uninterpreted sorts in
a model, CVC4 would print lines such as:

```
(declare-fun  () integer)
```

This was due to two issues:

- Representatives of uninterpreted sorts were not quoted correctly
  (empty symbol names should be quoted)
- The `quoteSymbol` function in `smt2_printer.cpp` was not turning a
  node into a string before quoting it but was instead just generating
  an empty string

This commit fixes both issues and unifies the implementations for
quoting symbols that we had in `printer/smt2/smt2_printer.cpp` and
`util/smt2_quote_string.cpp`.


ajreynol:
Nice catch(es).  LGTM.

PR #3603, opened 2020-01-13
New C++ API: Add nullary constructor for Result.

aniemetz:
+ Add unit tests for cvc4::api::Result
+ Fix satisfiability vs validitiy result in SmtEngine

Labeling this as `major` because I need it for the mbt.


test/unit/api/result_black.h
```diff
@@ -0,0 +1,118 @@
...
+void ResultBlack::testEq()
+{
+  Sort u_sort = d_solver->mkUninterpretedSort("u");
+  Term x = d_solver->mkVar(u_sort, "x");
+  d_solver->assertFormula(x.eqTerm(x));
+  Result res;
+  Result res2 = d_solver->checkSat();
+  Result res3 = d_solver->checkSat();
+  res = res2;
+  TS_ASSERT (res == res2);
```
4tXJ7f:
```suggestion
  TS_ASSERT_EQUALS(res, res2);
```

test/unit/api/result_black.h
```diff
@@ -0,0 +1,118 @@
...
+  d_solver->setLogic("QF_NIA");
+  d_solver->setOption("incremental", "false");
+  d_solver->setOption("solve-int-as-bv", "32");
+  Sort int_sort = d_solver->getIntegerSort();
+  Term x = d_solver->mkVar(int_sort, "x");
+  d_solver->assertFormula(x.eqTerm(x).notTerm());
+  Result res = d_solver->checkValid();
+  TS_ASSERT (!res.isValid());
+  TS_ASSERT (res.isValidUnknown());
+  TS_ASSERT_EQUALS(res.getUnknownExplanation(), "UNKNOWN_REASON");
```
4tXJ7f: Inconsistent formatting (space between `TS_ASSERT (res.isValidUnknown())` but not `TS_ASSERT_EQUALS(res.getUnknownExplanation(), "UNKNOWN_REASON")`).

aniemetz: Oh shoot, that happened because I forgot to add the file initially (when I clang-formatted). Will fix.

aniemetz:
@4tXJ7f @makaimann would it make sense to have a `isNull()` function in `cvc::api::Result`?

makaimann:
> @4tXJ7f @makaimann would it make sense to have a `isNull()` function in `cvc::api::Result`?

I was thinking it would, but what do you think? Is it unnecessary? It doesn't seem like it'd be possible to return a null result as-is. But maybe it'd be better to have an assertion for that.

4tXJ7f:
> @4tXJ7f @makaimann would it make sense to have a `isNull()` function in `cvc::api::Result`?

I am fine with adding that. Although it may go away again if we switch to the result representation as discussed yesterday (i.e. not distinguishing between sat and valid results) because we could just say that the default result is "unknown".

aniemetz:
Hm, no. A `null` Result maybe should not be the same as an `unknown` Result. A null result would be `Result res;`, an unknown result would be a result that is an actual result return from a check sat call that is `unknown`. Does that make sense? @4tXJ7f

aniemetz:
Also, @barrettcw what's your opinion?

aniemetz:
> My only question is whether there should be checks for null results before returning them.
Nope since we only have null results on the API level, and there we know what we create.

4tXJ7f:
> Hm, no. A `null` Result maybe should not be the same as an `unknown` Result. A null result would be `Result res;`, an unknown result would be a result that is an actual result return from a check sat call that is `unknown`. Does that make sense? @4tXJ7f

Hmm, I don't have a strong preference. I'd be fine to essentially have four states: null, sat, unsat, unknown. One question that we should ask ourselves: instead of having a null result, shouldn't we let the client deal with that case? A client of the API can use `std::optional` or heap allocated objects for representing the null case.

barrettcw:
Rather than calling it Null, maybe we call it NoResult and add an isNoResult tester?

aniemetz:
> One question that we should ask ourselves: instead of having a null result, shouldn't we let the client deal with that case?

The only case where we can get a null/no result is when you pre-declare a result, i.e., `Result res;`.
I would add a tester so that we can have assertions on the API level.

@barrettcw we call it null for Ops, Terms and Sorts -- for the sake of consistency we might want to also use `null` here. Also, 'no result' could be confused with 'unknown', which this definitely isn't. It's just a Result object we created that doesn't not wrap an actual check-sat result yet.

aniemetz:
I'll go ahead and merge this for now, I can rename stuff if necessary in an extra PR.

PR #3605, opened 2020-01-13
Generalize example-based sym breaking to conjectures with constant function apps

ajreynol:
This commit generalizes our example-based symmetry breaking to conjectures with constant function applications. This means for a conjecture like:
`exists f. f( 45 ) > 0 ^ f( 99 ) > 0`
we only built a set of terms that are pairwise unique over inputs (45), (99).

Previously, example-based symmetry breaking was limited to PBE, but this limitation was too strict.


src/theory/quantifiers/sygus/sygus_pbe.h
```diff
@@ -153,7 +153,18 @@ class SygusPbe : public SygusModule
...
   bool isPbe() { return d_is_pbe; }
-  /** is the enumerator e associated with I/O example pairs? */
+  /**
+   * Is the enumerator e associated with examples? This is true if the
+   * function-to-synthesize associated with e is only applied to concrete
+   * arguments. Notice that the conjecture need not be in PBE form for this
+   * to be the case. For example, f has examples in:
+   *   exists f. f( 1 ) = 3 ^ f( 2 ) = 4
+   *   exists f. f( 45 ) > 0 ^ f( 99 ) > 0
+   *   exists f. forall x. ( x > 5 => f( 4 ) < x )
```
HanielB: How is the utility helpful in this example?

ajreynol: Added more documentation.

PR #3611, opened 2020-01-17
Use axioms when checking goal entailment for abduction algorithm

ajreynol:
This PR utilitizes the axioms of an abduction query while checking whether the current solution entails the goal.

This both improves the quality of our counterexample points and should make the algorithm converge on solutions much faster.

### DIFF
--- a/src/theory/quantifiers/sygus/cegis_core_connective.h
+++ b/src/theory/quantifiers/sygus/cegis_core_connective.h
@@ -139,8 +139,8 @@ class VariadicTrie
  *   {
  *     D += { d' }
  *     if D is false for all v in pts(B)
- *       if D => B
- *         Let U be a subset of D such that U ^ ~B is unsat.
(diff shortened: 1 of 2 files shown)


src/theory/quantifiers/sygus/cegis_core_connective.h
```diff
@@ -337,13 +337,14 @@ class CegisCoreConnective : public Cegis
    * Assuming smt has just been called to check-sat and returned "UNSAT", this
    * method get the unsat core and adds it to uasserts.
    *
-   * If query is non-null, then it is excluded from uasserts. If query was
-   * in the unsat core, then this method returns true. Otherwise, this method
-   * returns false. It also returns false if query was null.
+   * The assertions in queryAsserts, are excluded from uasserts. If one
```
HanielB: Clarify what `queryAsserts` is

PR #3612, opened 2020-01-17
LIRA proof: Arithmetic predicates & reification thereof

alex-ozdemir:
Declare a mixed (Int >= Real) predicate type and a function for reifying it.

### DIFF
--- a/proofs/signatures/th_lira.plf
+++ b/proofs/signatures/th_lira.plf
@@ -60,11 +60,7 @@
 ; Technically it's a real literal on the right, but we don't assume that here.
 (program reify_arith_pred ((p formula)) formula
   (match p
-         ((not p') (match p'
-                          ((>=_Real x y) (not (>=_Real (reify_real_term x) (reify_real_term y))))
-                          ((>=_Int x y) (not (>=_Real (reify_int_term x) (reify_int_term y))))
-                          ((>=_IntReal x y) (not (>=_Real (reify_int_term x) (reify_real_term y))))
-                          (default (fail formula))))
+         ((not p') (not (reify_arith_pred p)))
          ((>=_Real x y) (>=_Real (reify_real_term x) (reify_real_term y)))
          ((>=_Int x y) (>=_Real (reify_int_term x) (reify_int_term y)))
          ((>=_IntReal x y) (>=_Real (reify_int_term x) (reify_real_term y)))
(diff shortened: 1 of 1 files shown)


proofs/signatures/th_lira.plf
```diff
@@ -33,7 +33,7 @@
     ((u-_Int x) (u-_Real (reify_int_term x)))
     ((*_Int x y) (*_Real (reify_int_term x) (reify_int_term y)))
     ; Reifying integer division is an error, since it changes the value!
-    ((/_Int x y) (fail Real))
```
alex-ozdemir: A typo in my last PR.

proofs/signatures/th_lira.plf
```diff
@@ -52,3 +52,21 @@
     ((/_Real x y) (/_Real (reify_real_term x) (reify_real_term y)))
   ))
 
+; Predicates of the form (term Integer) >= (term Real) or negations thereof
+(declare >=_IntReal (! x (term Int) (! y (term Real) formula)))
```
yoni206: Given that we have conversions above, we would we need this predicate? Can't we use the above conversions and then use >=_Real?

proofs/signatures/th_lira.plf
```diff
@@ -52,3 +52,21 @@
...
+
+; From an arith predicate, compute the equivalent real predicate
+; All arith predicates are (possibly negated) >='s with a real on the right.
+; Technically it's a real literal on the right, but we don't assume that here.
+(program reify_arith_pred ((p formula)) formula
+  (match p
+         ((not p') (match p'
```
yoni206: Maybe I am too naive here but can't we do
```
((not p') (not (reify_arith_pred p'))
```
?

alex-ozdemir: Well yes. That is a good idea :P.

proofs/signatures/th_lira.plf
alex-ozdemir: Good question. Sometimes we need to know that one side of the inequality is integral so that we can tighten the bound (e.g. go from `x < 4.5` to `x <= 4`). If we don't have a theorem of the form "the (integral) left side is less than the right", its not clear to to write a sound tightening axiom.

yoni206: Thanks for the explanation, makes sense.

PR #3627, opened 2020-01-21
Types & side-conditions for linear and affine fns

alex-ozdemir:
This PR introduces types for linear combinations of arith variables
along with side conditions for their arithmetic.

It does the same for affine functions.

These primitives are ultimately used in our machinery for Farkas proofs.

This machinery is just a stream-lined version of what we already have in `th_lra.plf`.

### DIFF
--- a/proofs/signatures/th_lira.plf
+++ b/proofs/signatures/th_lira.plf
@@ -192,3 +192,94 @@
     (! pf_real_bound (th_holds (not (>=_IntReal t (term_int_to_real (a_int old_bound)))))
       (! sc_neg (^ (negate_linear_int_term t) neg_t)
         (th_holds (>=_IntReal neg_t (term_int_to_real (a_int neg_int_bound))))))))))))
+
+;; ======================================== ;;
+;; Linear Combinations and Affine functions ;;
+;; ======================================== ;;
+
+; Unifying type for both kinds of arithmetic variables
+(declare arith_var type)
+(declare av_from_int (! v int_var arith_var))
+(declare av_from_real (! v real_var arith_var))
+
+; Total order type -- return value for the comparison of two things
+(declare ord type)
+(declare ord_lt ord)
+(declare ord_eq ord)
+(declare ord_gt ord)
+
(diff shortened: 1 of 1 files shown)


proofs/signatures/th_lira.plf
```diff
@@ -192,3 +192,94 @@
...
+                (mp_ifzero c
+                  (lc_add l1' l2')
+                  (lc_cons c v1 (lc_add l1' l2')))))
+            (ord_gt (lc_cons c2 v2 (lc_add l1 l2')))))))))
+
+; Scaling a linear combination
+(program lc_mul_c ((l lc) (c mpq)) lc
+  (match l
+    (lc_null l)
+    ((lc_cons c' v' l') (lc_cons (mp_mul c c') v' (lc_mul_c l' c)))))
```
yoni206: Don't you need a special case for zero? Same question for `aff_mul_c` below.

alex-ozdemir: I don't believe that you need a special case, because it is okay to store a 0 coefficient in the linear combination. We could special-case it though. What do you think?

proofs/signatures/th_lira.plf
```diff
@@ -192,3 +192,94 @@
...
+        (lc_null l1)
+        ((lc_cons c2 v2 l2')
+          (match (arith_var_cmp v1 v2)
+            (ord_lt (lc_cons c1 v1 (lc_add l1' l2)))
+            (ord_eq
+              (let c (mp_add c1 c2)
+                (mp_ifzero c
+                  (lc_add l1' l2')
+                  (lc_cons c v1 (lc_add l1' l2')))))
+            (ord_gt (lc_cons c2 v2 (lc_add l1 l2')))))))))
```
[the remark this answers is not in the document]
alex-ozdemir: It can, because the two variables are from different lists, not the same list.

proofs/signatures/th_lira.plf
yoni206: OK, I did no that. In that case no special case is needed.

PR #3631, opened 2020-01-21
Types and side conditions for affine bounds

alex-ozdemir:
Bounds (being positive, non-negative) actually have an arithmetic.
This PR defines that.

Useful b/c Farkas proofs are basically just sums of bounded affine
functions.

### DIFF
--- a/proofs/signatures/th_lira.plf
+++ b/proofs/signatures/th_lira.plf
@@ -353,3 +353,49 @@
     (! is_affy (is_aff y aff_y)
       (! a (^ (aff_mul_c aff_y x) aff_z)
         (is_aff (*_Real y (a_real x)) aff_z))))))))
+
+;; ========================== ;;
+;; Bounds on Affine Functions ;;
+;; ========================== ;;
+
+; Bounds that an affine function might satisfy
+(declare bound type)
+(declare bound_pos bound)           ; > 0
+(declare bound_non_neg bound)       ; >= 0
+
+; formulas over affine functions
+(declare bounded_aff (! a aff (! b bound formula)))
+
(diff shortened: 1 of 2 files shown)


proofs/signatures/th_lira.plf
```diff
@@ -353,3 +353,49 @@
...
+;; Bounds on Affine Functions ;;
+;; ========================== ;;
+
+; Bounds that an affine function might satisfy
+(declare bound type)
+(declare bound_pos bound)           ; > 0
+(declare bound_non_neg bound)       ; >= 0
+
+; formulas over affine functions
+(declare bounded_aff (! a aff (! b bound formula)))
```
yoni206: Maybe document something like: Given `a` and `b`, what does `f` represent?

proofs/signatures/th_lira.plf
```diff
@@ -353,3 +353,49 @@
...
+    (bound_pos bound_pos)
+    (bound_non_neg b2)))
+
+; The implication of `a1` satisfying `b` and `a2` satisfying `b2`, obtained by
+; summing the inequalities.
+(program bounded_aff_add ((a1 aff) (b bound) (a2 aff) (b2 bound)) formula
+  (bounded_aff (aff_add a1 a2) (bound_add b b2)))
+
+
+; Is this rational positive?
```
yoni206: consider moving to `th_real.plf`.

proofs/signatures/th_lira.plf
alex-ozdemir: Good idea. Will do!

proofs/signatures/th_lira.plf
alex-ozdemir: Sounds good to me!

PR #3632, opened 2020-01-21
Axioms for affine function bounds. Tests.

alex-ozdemir:
This PR is much larger, because it includes tests for the whole proof system.
You can run `lfscc sat.plf smt.plf lrat.plf drat.plf er.plf th_base.plf th_int.plf th_real.plf th_lira.plf th_lira_test.plf` from the `proofs/signatures` directory to run the tests (assuming `lfscc` is on your path).

This is the last signature PR!

### DIFF
--- a/proofs/signatures/th_lira_test.plf
+++ b/proofs/signatures/th_lira_test.plf
@@ -1,0 +1,339 @@
+; Depends On: th_lra.plf
+;; Proof (from predicates on linear polynomials) that the following imply bottom
+;
+; -x - 1/2 y + 2 >= 0
+;  x +     y - 8 >= 0
+;  x -     y + 0 >= 0
+;
+(check
+  ; Variables
+  (% x real_var
+  (% y real_var
+  ; linear combinations
+  (@ m1 (lc_cons (~ 1/1) (av_from_real x) (lc_cons (~ 1/2) (av_from_real y) lc_null))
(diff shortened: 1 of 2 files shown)


proofs/signatures/th_lira_test.plf
```diff
@@ -0,0 +1,339 @@
+; Depends On: th_lra.plf
```
yoni206:
```suggestion
; Depends On: th_lira.plf
```

proofs/signatures/th_lira_test.plf
```diff
@@ -0,0 +1,339 @@
...
+           (bounded_aff_add _ _ _ _ _
+             (bounded_aff_mul_c _ _ _ 1/1 n3)
+             bounded_aff_ax_0_>=_0)))))
+  )))
+  )))
+  )))
+  ))
+)
+
+;; Term proof, 2 (>=), one (not >=)
```
yoni206: can you clarify this line? I think it is enough to do that once and the rest will be clear.

alex-ozdemir: Yeah, that is opaque. I've expanded this phrase.

alex-ozdemir:
Thanks for reviewing, @yoni206! I've made changes to address the shortcomings you point out.

yoni206:
Thanks for a great and fun-to-review sequence of PRs, that conforms with https://github.com/CVC4/CVC4/wiki/Developer-Guide#how-to-break-up-large-code-reviews !

PR #3639, opened 2020-01-22
Avoid PLUS with one child for bv2nat elimination

### LINKED ISSUE #3621: Fatal failure at CVC4/src/./expr/node_builder.h:961
Hi,
For this formula,
```
(declare-fun a () (_ BitVec 1))
(assert (< (bv2nat a) 1))
(check-sat)
```
```
Fatal failure within CVC4::expr::NodeValue* CVC4::NodeBuilder<N>::constructNV() [with unsigned int nchild_thresh = 10] at CVC4/src/./expr/node_builder.h:961
Check failure
 getNumChildren() >= kind::metakind::getLowerBoundForKind(getKind())
Nodes with kind PLUS must have at least 2 children (the one under construction has 1)
Aborted
```

OS: Ubuntu 18.04
Revision: c3bc4ac

ajreynol:
Fixes #3621.

### DIFF
--- a/src/theory/bv/theory_bv.cpp
+++ b/src/theory/bv/theory_bv.cpp
@@ -630,9 +630,10 @@ int TheoryBV::getReduction(int effort, Node n, Node& nr)
           nm->mkNode(kind::EQUAL,
                      nm->mkNode(nm->mkConst(BitVectorExtract(bit, bit)), n[0]),
                      bvone);
-      children.push_back( nm->mkNode(kind::ITE, cond, nm->mkConst(Rational(i)), z));
+      children.push_back(
+          nm->mkNode(kind::ITE, cond, nm->mkConst(Rational(i)), z));
     }
-    nr = children.size()==1 ? children[0] : nm->mkNode(kind::PLUS,children);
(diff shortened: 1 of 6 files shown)


src/theory/bv/theory_bv.cpp
```diff
@@ -622,17 +622,18 @@ int TheoryBV::getReduction(int effort, Node n, Node& nr)
...
       Node cond =
           nm->mkNode(kind::EQUAL,
                      nm->mkNode(nm->mkConst(BitVectorExtract(bit, bit)), n[0]),
                      bvone);
-      result << nm->mkNode(kind::ITE, cond, nm->mkConst(Rational(i)), z);
+      children.push_back(
+          nm->mkNode(kind::ITE, cond, nm->mkConst(Rational(i)), z));
     }
-    nr = Node(result);
+    nr = children.size() == 1 ? children[0] : nm->mkNode(kind::PLUS, children);
```
mpreiner: How about using `RewriteRule<BVToNatEliminate>::apply` instead of duplicating the code?

ajreynol: That function is private so I've added a bit-vector utility that both call now.

PR #3660, opened 2020-01-27
Fix header installation on MacOS.

mpreiner:
On MacOS sed -i requires a suffix to be set.

@4tXJ7f did you have any issues on MacOS/homebrew with this?


src/fix-install-headers.sh
```diff
@@ -2,4 +2,4 @@
 
 dir=$1
 find "$dir/include/cvc4/" -type f | \
-  xargs sed -i 's/include.*"\(.*\)"/include <cvc4\/\1>/'
+  xargs sed -i'' 's/include.*"\(.*\)"/include <cvc4\/\1>/'
```
4tXJ7f: To make it work on the current version of macOS, I needed to do:

```suggestion
  xargs sed -i'' -e 's/include.*"\(.*\)"/include <cvc4\/\1>/'
```

This should work on 10.9+ [according to StackOverflow](https://stackoverflow.com/questions/4247068/sed-command-with-i-option-failing-on-mac-but-works-on-linux).

Side note: instead of piping to `xargs`, we should just use the `-exec` option of `find` here.

mpreiner: Done.

Issue #3662, opened 2020-01-28
theory/sets/cardinality_extension.cpp:113; cardinality not supported yet.


wintered:
```
$cvc4 --fmf-fun --quiet bug.smt2
Fatal failure within void CVC4::theory::sets::CardinalityExtension::checkFiniteType(CVC4::TypeNode&) at /net5p/su.fs.inf.ethz.ch/export/su/project/solver_binaries/cache/CVC4/src/theory/sets/cardinality_extension.cpp:113
Check failure

 false
The cardinality beth[0] of the finite type a is not supported yet.

[1]    109806 abort      cvc4/build-2020-01-25-bde056b/cvc4 --incremental --fmf-fun bug.smt2

$ cat bug.smt2 
(declare-sort a 0)
(declare-fun b () (Set a))
(assert (= (card b) 0))
(check-sat)
```

ajreynol:
We should throw a proper exception instead of an assertion failure for this.

mudathirmahgoub:
A Logical exception is now thrown instead of an assertion failure. Supporting this feature is a long term goal. Thanks Dominik for reporting this issue.

ajreynol:
Closing this since we now throw a LogicException.

PR #3663, opened 2020-01-28
Fix cardinality of uninterpreted types when univset is not used

mudathirmahgoub:
This fixes the error that is thrown when the `cardinality` operator is used on sets of `uninterpreted sorts` with `finite model find` option. Computing the cardinality of the `univset` is not yet implemented, but if the `univset` term is not explicitly used in the input, no error (like  issue #3662)  should be thrown.

This PR only fixes this case, and it is not a permanent solution. It does not fix the following example , which returns `unknown` for `check-sat`, although the cardinality of the `univset` is not constrained. 
```smt2
(set-logic ALL)
(set-option :sets-ext true)
(declare-sort a 0)
(declare-const b (Set a))
(assert (= (card b) 1))
(assert (subset b (as univset (Set a))))
(check-sat)
(get-model)
```
If the cardinality is zero the result is `sat`
```smt2
(set-logic ALL)
(set-option :sets-ext true)
(declare-sort a 0)
(declare-const b (Set a))
(assert (= (card b) 0))
(assert (subset b (as univset (Set a))))
(check-sat)
(get-model)
```
and the returned model is 
```smt2
(model
; cardinality of a is 1
(declare-sort a 0)
(declare-fun groundTerm_2 () a)
(define-fun b () (Set a) (as emptyset (Set a)))
)
```


ajreynol:
In general, I'm not sure if we can currently handle examples like `bug3663.smt2`.  Isn't it the case that we may run out of elements during model construction?  Thus I think this PR is actually allowing for cases where CVC4 would incorrectly report "sat" (which is worse behavior than that which is reported in #3663).

mudathirmahgoub:
Was cardinality operator supported for uninterpreted sorts + FMF before we added cardinality for finite types? If no, then, we don't need this PR and it should be closed. 

If yes, then I think we can support `bug3663.smt2`, because the rules for cardinality of `univset` is not used and not needed. If it is used, then this PR would throw an assertion error. 

The problem is when the term `univset` is used like the first example above, the returned answer is not `sat` but `unknown`. For some reason, it returns the correct model `empty set` for the second example.

ajreynol:
Can you try more examples, like (card a) > 100 with finite model finding where a is of type (Set U)? I am interested if you get check-model failures.

PR #3665, opened 2020-01-29
expectedType in proof-printing code

alex-ozdemir:
To print lemma proofs in theories that use multiple sorts that have a
subtype relationship, we need to increase communication between the
TheoryProofEngine and the theory proofs themselves.

This commit add an (optional) argument `expectedType` to many
term-printing functions in TheoryProofEngine and TheoryProof.

Right now it is unused, so always takes on the default value of "null"
(meaning no type expectation), but in the future the TheoryProofEngine
will use it to signal TheoryProof about what type is expected to be
printed.


src/proof/theory_proof.h
```diff
@@ -234,8 +258,12 @@ class TheoryProof {
    *
    * @param term expression representing term
    * @param os output stream
+   * @param expectedType The type that this is expected to have in a parent node.
+   *                      Null if there are no such requirements.
+   *                      This is useful for requesting type conversions from the theory.
+   *                      e.g. in (5.5 == 4) the right-hand-side should be converted to a real.
    */
-  virtual void printOwnedTerm(Expr term, std::ostream& os, const ProofLetMap& map) = 0;
+  virtual void printOwnedTerm(Expr term, std::ostream& os, const ProofLetMap& map, Type expectedType = Type()) = 0;
```
yoni206: I might be reading [this](https://github.com/CVC4/CVC4/wiki/Code-Style-Guidelines#default-arguments) wrong, but aren't you in one of the "banned" cases here?

yoni206: Can `expectedType` be a reference to `Type` and not `Type` (in this virtual function but also in each instance of it in the different classes)?

yoni206: Do you expect to usually use a `Type` or a `TypeNode`?

alex-ozdemir: Good catch! You are correct. Andres and I cooked up an alternative. I'll implement it and update the PR.

alex-ozdemir: From talking to Andres: `TypeNode` is to `Type` as `Node` is to `Expr`.
I should be using `TypeNode`. I'll update the PR.

alex-ozdemir: Given that I change to `TypeNode`, I think its better to stay with a value b/c nodes are cheap anyway.

PR #3666, opened 2020-01-29
Refactor relevance vectors for asserted quantifiers

### LINKED ISSUE #3664: --fmf-fun-rlv --sygus-inference fatal failure at CVC4/src/theory/quantifiers/first_order_model.cpp:61
Hi,
For this formula,
```
(set-logic QF_NRA)
(declare-fun a () Real)
(assert (= (/ a a) 1.0))
(check-sat)
```
`cvc4 --fmf-fun-rlv --sygus-inference` throws out a fatal failure:
```
Fatal failure within CVC4::Node CVC4::theory::quantifiers::FirstOrderModel::getAssertedQuantifier(unsigned int, bool) at CVC4/src/theory/quantifiers/first_order_model.cpp:61
Check failure

 d_forall_rlv_assert.size() == d_forall_asserts.size()

Aborted
```
OS: Ubuntu 18.04
Revision: bde056b

ajreynol:
Ensures we don't access the list of relevant quantifiers if it has not been computed.

Fixes #3664.


src/theory/quantifiers/first_order_model.cpp
```diff
@@ -55,12 +56,11 @@ unsigned FirstOrderModel::getNumAssertedQuantifiers() {
...
 
 Node FirstOrderModel::getAssertedQuantifier( unsigned i, bool ordered ) { 
-  if( !ordered ){
+  if( !ordered || !d_forallRlvComputed ){
     return d_forall_asserts[i]; 
-  }else{
-    Assert(d_forall_rlv_assert.size() == d_forall_asserts.size());
-    return d_forall_rlv_assert[i];
   }
+  Assert(d_forall_rlv_assert.size() == d_forall_asserts.size());
```
HanielB: Why not use this test rather than introduce a new flag?

ajreynol: Using the flag makes it explicit that the data is not stale.  Otherwise we are relying on a hidden invariant that this condition is equivalent to the fact that d_forall_rlv_assert is not stale.

HanielB: The actual issue is the invariant being hidden hehe The flag did not help me much, to understand what was going on I still had to look over the code to discover that invariant. Maybe just saying it explicitly above this line would suffice.

PR #3670, opened 2020-01-29
Example inference utility

ajreynol:
This is work towards generalizing our strategies for "symmetry breaking up to examples" to non-PBE conjectures (from this branch: https://github.com/ajreynol/CVC4/tree/optExamplesSygus).

This introduces a self-contained utility that infers whether a conjecture is in "examples form".  The code is taken from the current SygusPbe module.  That code will be updated to use this module.

### DIFF
--- a/src/theory/quantifiers/sygus/example_infer.cpp
+++ b/src/theory/quantifiers/sygus/example_infer.cpp
@@ -174,12 +174,15 @@ bool ExampleInfer::collectExamples(Node n,
         else
         {
           Assert(n_output.isConst());
+          // finished processing this node if it was an I/O pair
+          return true;
         }
-        // finished processing this node
-        return true;
       }
-      d_examples_invalid[eh] = true;
-      d_examplesOut_invalid[eh] = true;
+      else
+      {
+        d_examples_invalid[eh] = true;
+        d_examplesOut_invalid[eh] = true;
+      }
     }
   }
   for (unsigned i = 0, nchild = n.getNumChildren(); i < nchild; i++)
--- a/src/theory/quantifiers/sygus/example_infer.cpp
+++ b/src/theory/quantifiers/sygus/example_infer.cpp
@@ -41,8 +41,7 @@ bool ExampleInfer::initialize(Node n, const std::vector<Node>& candidates)
     d_examplesOut[v].clear();
     d_examplesTerm[v].clear();
   }
-
-  std::map<Node, bool> visited;
+  std::map<std::pair<bool,bool>, std::unordered_set<Node, NodeHashFunction>> visited;
   // n is negated conjecture
   if (!collectExamples(n, visited, true, false))
   {
@@ -85,16 +84,17 @@ bool ExampleInfer::initialize(Node n, const std::vector<Node>& candidates)
 }
 
 bool ExampleInfer::collectExamples(Node n,
-                                   std::map<Node, bool>& visited,
+                                   std::map<std::pair<bool,bool>, std::unordered_set<Node, NodeHashFunction>>& visited,
                                    bool hasPol,
                                    bool pol)
 {
-  if (visited.find(n) != visited.end())
+  std::pair<bool,bool> cacheIndex = std::pair<bool,bool>(hasPol, pol );
+  if (visited[cacheIndex].find(n) != visited[cacheIndex].end())
   {
     // already visited
     return true;
   }
-  visited[n] = true;
+  visited[cacheIndex].insert(n);
   NodeManager* nm = NodeManager::currentNM();
   Node neval;
   Node n_output;
@@ -120,6 +120,7 @@ bool ExampleInfer::collectExamples(Node n,
           n_output = n[1 - r];
         }
         neval_is_evalapp = true;
+        break;
       }
     }
   }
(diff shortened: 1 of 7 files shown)


src/theory/quantifiers/sygus/example_infer.h
```diff
@@ -0,0 +1,158 @@
...
+namespace CVC4 {
+namespace theory {
+namespace quantifiers {
+
+/** Example Inference
+ *
+ * This class determines whether a formula "has examples", for details
+ * see the method hasExamples below. This is important for certain
+ * optimizations in enumerative SyGuS, include example-based symmetry breaking
+ * (discarding terms that equivalent to a previous one up to examples).
```
HanielB: scratch "that"

src/theory/quantifiers/sygus/example_infer.h
```diff
@@ -0,0 +1,158 @@
...
+                       bool pol);
+  /** Pointer to the sygus term database */
+  TermDbSygus* d_tds;
+  /** is this an examples conjecture for all functions-to-synthesize? */
+  bool d_isExamples;
+  /** for each candidate variable f (a function-to-synthesize), whether the
+   * conjecture is purely PBE for that variable
+   * In other words, all occurrences of f are guarded by equalities that
+   * constraint its arguments to constants.
+   */
```
HanielB: update this to reflect it being used not only for PBE

src/theory/quantifiers/sygus/example_infer.h
```diff
@@ -0,0 +1,158 @@
...
+   * another example is:
+   *   exists f. forall x. f( 0 ) = 2 V f( 3 ) = 3
+   * since the formula is not a conjunction (the example values are not
+   * entailed). However, the domain of f in both cases is finite, which can be
+   * used for search space pruning.
+   */
+  std::map<Node, bool> d_examplesOut_invalid;
+  /**
+   * For each candidate variable (function-to-synthesize), input of I/O
+   * examples
```
HanielB: update

src/theory/quantifiers/sygus/example_infer.h
```diff
@@ -0,0 +1,158 @@
...
+   */
+  std::map<Node, std::vector<std::vector<Node>>> d_examples;
+  /**
+   * For each candidate variable (function-to-synthesize), output of I/O
+   * examples.
```
HanielB: add a disclaimer about cases when examples have no output

src/theory/quantifiers/sygus/example_infer.h
```diff
@@ -0,0 +1,158 @@
...
+   */
+  std::map<Node, std::vector<Node>> d_examplesOut;
+  /** the list of example terms
+   * For the example [EX#1] above, this is f( 0 ), f( 5 ), f( 6 )
+   */
+  std::map<Node, std::vector<Node>> d_examplesTerm;
+  /**
+   * Map from example input terms to their output, for example [EX#1] above,
+   * this is { f( 0 ) -> 2, f( 5 ) -> 7, f( 6 ) -> 8 }.
```
HanielB: ditto

src/theory/quantifiers/sygus/example_infer.cpp
```diff
@@ -0,0 +1,266 @@
...
+      if (n[r].getKind() == DT_SYGUS_EVAL)
+      {
+        neval = n[r];
+        if (n[1 - r].isConst())
+        {
+          n_output = n[1 - r];
+        }
+        neval_is_evalapp = true;
+      }
+    }
```
HanielB: This does not seem robust for `f(c1) = f(c2)`. How is that being handled?

src/theory/quantifiers/sygus/example_infer.h
```diff
@@ -131,21 +132,22 @@ class ExampleInfer
...
-   * examples
+   * For each function-to-synthesize f, the list of concrete inputs to f.
    */
   std::map<Node, std::vector<std::vector<Node>>> d_examples;
   /**
-   * For each candidate variable (function-to-synthesize), output of I/O
-   * examples.
+   * For each function-to-synthesize f, the list of outputs according to the
+   * I/O specification for f.
+   * The vector d_examplesOut[f] valid only if d_examplesOut_invalid[f]=true.
```
HanielB: is valid

src/theory/quantifiers/sygus/example_infer.cpp
ajreynol: You're right, there is a bug (in the existing code as well)
Will fix.

ajreynol: If we are not equality but not an I/O pair (one side constant), we now always traverse to the subterms, this means we will collect both c1 and c2.

HanielB: I don't get it. :)

ajreynol: This code sets neval_is_evalapp to true if we are f( s ) = t for any s, t.  It may be that s, t are not constants.  It doesnt matter if t is a non-constant (x) or another application of a function to synthesize, the check for isConst on line 118 here will fail.

HanielB: I see. Another thing: in the case of `f(c1) = f(c2)` isn't the code gonna add the example `c2` twice? Once in the code block starting at line 127 and then when it visits the child `f(c2)` from the recursive call at line 193.

ajreynol: I've updated the code so that this is explicitly guarded against.  The previous code was indeed unclear whether this could happen.

HanielB: Cool, thanks!

PR #3675, opened 2020-01-29
Minor updates to string utilities

ajreynol:
This final preparation for splitting the core solver for word equations out of TheoryStrings.

### DIFF
--- a/src/theory/strings/inference_manager.cpp
+++ b/src/theory/strings/inference_manager.cpp
@@ -630,6 +630,7 @@ void InferenceManager::setIncomplete() { d_out.setIncomplete(); }
 
 void InferenceManager::markCongruent(Node a, Node b)
 {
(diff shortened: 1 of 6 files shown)


src/theory/strings/inference_manager.cpp
```diff
@@ -625,6 +626,16 @@ void InferenceManager::explain(TNode literal,
     }
   }
 }
+void InferenceManager::setIncomplete() { d_out.setIncomplete(); }
+
+void InferenceManager::markCongruent(Node a, Node b)
+{
+  ExtTheory* eth = d_parent.getExtTheory();
+  if (eth->hasFunctionKind(a.getKind()))
```
4tXJ7f: Shouldn't we also check the kind of `b`?

ajreynol: They should be the same, I'll add an assertion.

PR #3678, opened 2020-01-30
Move disequality list to solver state in strings

ajreynol:
Towards splitting the core solver for the theory of strings.

Also moves the equality notification callback code into solver state.

### DIFF
--- a/src/theory/strings/solver_state.cpp
+++ b/src/theory/strings/solver_state.cpp
@@ -201,14 +201,9 @@ bool SolverState::areDisequal(Node a, Node b) const
 
 eq::EqualityEngine* SolverState::getEqualityEngine() const { return &d_ee; }
 
-void SolverState::getDisequalityList(std::vector<Node>& diseqs) const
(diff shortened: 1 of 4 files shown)


src/theory/strings/solver_state.cpp
```diff
@@ -200,6 +201,62 @@ bool SolverState::areDisequal(Node a, Node b) const
 
 eq::EqualityEngine* SolverState::getEqualityEngine() const { return &d_ee; }
 
+void SolverState::getDisequalityList(std::vector<Node>& diseqs) const
```
4tXJ7f: Instead of copying the values to a vector, I think that we should just return a `const` reference to the `NodeList` here. Also note that `CDList` defines `begin()` and `end()`, so we can iterate over it using a range-based for loop, i.e. `for (const Node& n : d_eeDisequalities)`.

ajreynol: Yeah you're right, I'll update.

PR #3680, opened 2020-01-30
Split base solver from the theory of strings

ajreynol:
This splits reasoning steps for basic term indexing and inferences for inferring when an equivalence class is entailed to be a constant.

The next step will be to split additional components (e.g. the core solver for word equations).

The code was moved, documented, and updated to meet the style guidelines when applicable. There are no major behavior changes in this PR.


src/theory/strings/base_solver.cpp
```diff
@@ -0,0 +1,427 @@
...
+              std::vector<Node> c;
+              Node nc = d_termIndex[k].add(n, 0, d_state, d_emptyString, c);
+              if (nc != n)
+              {
+                // check if we have inferred a new equality by removal of empty
+                // components
+                if (n.getKind() == STRING_CONCAT && !d_state.areEqual(nc, n))
+                {
+                  std::vector<Node> exp;
+                  unsigned count[2] = {0, 0};
```
4tXJ7f:
```suggestion
                  size_t count[2] = {0, 0};
```

src/theory/strings/base_solver.cpp
```diff
@@ -0,0 +1,427 @@
...
+void BaseSolver::checkInit()
+{
+  // build term index
+  d_eqcToConst.clear();
+  d_eqcToConstBase.clear();
+  d_eqcToConstExp.clear();
+  d_termIndex.clear();
+  d_stringsEqc.clear();
+
+  std::map<Kind, unsigned> ncongruent;
```
4tXJ7f:
```suggestion
  std::map<Kind, size_t> ncongruent;
```

src/theory/strings/base_solver.cpp
```diff
@@ -0,0 +1,427 @@
...
+  std::map<Kind, unsigned> congruent;
```
4tXJ7f:
```suggestion
  std::map<Kind, size_t> congruent;
```

src/theory/strings/base_solver.cpp
```diff
@@ -0,0 +1,427 @@
...
+          << "  Terms[" << it->first << "] = " << ncongruent[it->first] << "/"
+          << (congruent[it->first] + ncongruent[it->first]) << std::endl;
+    }
+  }
+}
+
+void BaseSolver::checkConstantEquivalenceClasses()
+{
+  // do fixed point
+  unsigned prevSize;
```
4tXJ7f:
```suggestion
  size_t prevSize;
```

src/theory/strings/base_solver.cpp
```diff
@@ -0,0 +1,427 @@
...
+        Trace("strings-debug")
+            << "Constant eqc : " << c << " for " << n << std::endl;
+        Trace("strings-debug") << "  ";
+        for (const Node& v : vecc)
+        {
+          Trace("strings-debug") << v << " ";
+        }
+        Trace("strings-debug") << std::endl;
+      }
+      unsigned count = 0;
```
4tXJ7f:
```suggestion
      size_t count = 0;
```

src/theory/strings/base_solver.cpp
```diff
@@ -0,0 +1,427 @@
...
+      unsigned countc = 0;
```
4tXJ7f:
```suggestion
      size_t countc = 0;
```

src/theory/strings/base_solver.h
```diff
@@ -0,0 +1,194 @@
...
+
+/** The base solver for the theory of strings
+ *
+ * This implements techniques for inferring when terms are congruent in the
+ * current context, and techniques for inferring when equivalence classes
+ * are equivalent to constants.
+ */
+class BaseSolver
+{
+  friend class InferenceManager;
```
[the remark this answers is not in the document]
ajreynol: It doesn't accidentally carried this over from somewhere else.

src/theory/strings/base_solver.cpp
ajreynol: This is a (debug) count of the number of congruent classes, not an index.  Thus I dont think it should change.

src/theory/strings/base_solver.cpp
ajreynol: Ditto.

src/theory/strings/base_solver.cpp
4tXJ7f: Yeah, I was actually hesitating to suggest it in this case. We should probably still make it `uint32_t` to conform [to our guidelines](https://github.com/CVC4/CVC4/wiki/Code-Style-Guidelines#integer-types).

PR #3684, opened 2020-01-30
Refactor sygus stats

ajreynol:
This makes the sygus statistics class standalone.  This will make it possible to define further statistics without accessing data in the SynthEngine.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -592,6 +592,8 @@ libcvc4_add_sources(
   theory/quantifiers/sygus/sygus_process_conj.h
   theory/quantifiers/sygus/sygus_repair_const.cpp
   theory/quantifiers/sygus/sygus_repair_const.h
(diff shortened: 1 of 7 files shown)


HanielB:
I don't quite understand the motivation. Could you elaborate?

ajreynol:
I would like to add more statistics for sygus, however its statistics class lives within SynthEngine.  Instead, I would like to have the SygusStatistics class be standalone and pass it to all the submodules.  In theory this could be done currently, but I'd have to pass around a statistics object that is defined within SynthEngine, which looks really ugly.

HanielB:
Doesn't everybody have access to SynthEngine already? Passing a statistics module everywhere does not seem very clean to me... this is the kind of information I'd expect to be centralized in the driver of the component.

ajreynol:
It does live there, we are passing a reference.  Otherwise everything has to have access to the *SynthEngine* itself.  You can already see an example of this in SynthConjecture.

PR #3686, opened 2020-01-30
Allow PBE symmetry breaking with sygus stream

ajreynol:
This led to an unexpected performance decrease in pLTL synthesis benchmarks.

There is no reason to disable this kind of symmetry breaking when sygus-stream is enabled.

### DIFF
--- a/src/options/quantifiers_options.toml
+++ b/src/options/quantifiers_options.toml
@@ -1138,7 +1138,6 @@ header = "options/quantifiers_options.h"
   long       = "sygus-unif-pi=MODE"
   type       = "SygusUnifPiMode"
   default    = "NONE"
-  read_only  = true
   help       = "mode for synthesis via piecewise-indepedent unification"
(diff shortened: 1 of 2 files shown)


HanielB:
And why are we disabling PBE itself in this case?

ajreynol:
Since PBE (decision tree learning) shouldnt be applied with sygus-stream (or otherwise many similar decision trees would be generated).

HanielB:
Please add a comment saying that in the point where you disable PBE. Also probably best to disable Unif+PI as well.

ajreynol:
sygusUnif includes Unif+PI.  We do have comments talking about "basic sygus" already:  https://github.com/CVC4/CVC4/blob/master/src/smt/smt_engine.cpp#L2004

ajreynol:
Actually, nvm about SyGuS+PI, I've added code to disable this now, good catch.

PR #3691, opened 2020-01-31
Handle `expectedType` in TheoryProofEngine

alex-ozdemir:
`TheoryProofEngine` now uses the `expectedType` optional argument.
  * When printing terms, it sets this for theories that it dispatches too
  * It occasionally asks theories for help determining the `expectedType` using `equalityType`, which has a sensible default implementation.
  * It is mindful of `expectedType` when using the let map.

I also moved to hpp function implementations into the cpp.

### DIFF
--- a/src/proof/theory_proof.cpp
+++ b/src/proof/theory_proof.cpp
@@ -905,10 +905,12 @@ void LFSCTheoryProofEngine::printBoundTermAsType(Expr term,
       if (count > LET_COUNT)
       {
         os << "let" << id;
+        Debug("pf::tp::letmap") << "Using let map for " << term << std::endl;
         return;
       }
     }
   }
+  Debug("pf::tp::letmap") << "Skipping let map for " << term << std::endl;
 
   printTheoryTerm(term, os, map, expectedType);
 }
--- a/src/proof/theory_proof.cpp
+++ b/src/proof/theory_proof.cpp
@@ -176,18 +176,19 @@ void TheoryProofEngine::printConstantDisequalityProof(std::ostream& os, Expr c1,
   getTheoryProof(theory::Theory::theoryOf(c1))->printConstantDisequalityProof(os, c1, c2, globalLetMap);
 }
 
-  void TheoryProofEngine::printTheoryTerm(Expr term,
-                       std::ostream& os,
-                       const ProofLetMap& map,
-                       TypeNode expectedType)
-  {
-    this->printTheoryTermAsType(term, os, map, expectedType);
-  }
+void TheoryProofEngine::printTheoryTerm(Expr term,
+                                        std::ostream& os,
+                                        const ProofLetMap& map,
+                                        TypeNode expectedType)
+{
+  this->printTheoryTermAsType(term, os, map, expectedType);
+}
 
 TypeNode TheoryProofEngine::equalityType(const Expr& left, const Expr& right)
 {
   Assert(theory::Theory::theoryOf(left) == theory::Theory::theoryOf(right));
-  return getTheoryProof(theory::Theory::theoryOf(left))->equalityType(left, right);
+  return getTheoryProof(theory::Theory::theoryOf(left))
+      ->equalityType(left, right);
 }
 
 void TheoryProofEngine::registerTerm(Expr term) {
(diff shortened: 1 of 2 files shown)


src/proof/theory_proof.cpp
```diff
@@ -866,18 +881,27 @@ void LFSCTheoryProofEngine::printBoundTermAsType(Expr term,
...
   Debug("pf::tp") << "LFSCTheoryProofEngine::printBoundTerm( " << term << " ) " << std::endl;
 
-  ProofLetMap::const_iterator it = map.find(term);
-  if (it != map.end()) {
-    unsigned id = it->second.id;
-    unsigned count = it->second.count;
+  // Since let-abbreviated terms are abbreviated with their default type, only
+  // use the let map if there is no expectedType or the expectedType matches
+  // the default.
+  if (expectedType.isNull()
```
yoni206: Is it possible to add at this point a regression test in which this condition does not hold? Intuitively it feels to me like there should be other places in the code that are affected by this decision (like closing parenthesis of `let` etc.), and a regression that reaches here can prove me wrong. If the current infrastructure cannot yet print complete proofs with expected types, then could you please make sure this does not affect other places?

alex-ozdemir: > Intuitively it feels to me like there should be other places in the code that are affected by this decision (like closing parenthesis of let etc.) ... If the current infrastructure cannot yet print complete proofs with expected types, then could you please make sure this does not affect other places?

Choosing not to use the let map is always fine.
The let map provides shorter names for many terms, but using it is optional, and printing the term is always a safe alternative.

> Is it possible to add at this point a regression test in which this condition does not hold?
The let map does not contain some terms in pretty much every proofs.
I've added the debug tag "pf::tp::letmap" which you can enable to see it in action. For example, the `regress0/arith/ackermann.real.smt2` regression (which is the first one I checked) produces the following output using the flag:

```
Skipping let map for true
Skipping let map for (= (g b) (f b))
Skipping let map for (g b)
Skipping let map for b
Skipping let map for (f b)
Skipping let map for b
Skipping let map for (= (f a) b)
Skipping let map for (f a)
Skipping let map for a
Skipping let map for b
Skipping let map for (let ((_let_0 (f a))) (= (f (g (f _let_0))) (f (g _let_0))))
Skipping let map for (f (g (f (f a))))
Skipping let map for (g (f (f a)))
Skipping let map for (f (f a))
Skipping let map for (f a)
Skipping let map for a
Skipping let map for (f (g (f a)))
Skipping let map for (g (f a))
Skipping let map for (f a)
Skipping let map for a
Skipping let map for (= (g a) (f a))
Skipping let map for (g a)
Skipping let map for a
Skipping let map for (f a)
Skipping let map for a
Skipping let map for (= (f b) a)
Skipping let map for (f b)
Skipping let map for b
Skipping let map for a
Skipping let map for (not (= a b))
Skipping let map for (= a b)
Skipping let map for a
Skipping let map for b
Skipping let map for a
Skipping let map for (f a)
Using let map for a
Skipping let map for (f (f a))
Using let map for (f a)
Skipping let map for (g (f (f a)))
Using let map for (f (f a))
Skipping let map for (f (g (f (f a))))
Using let map for (g (f (f a)))
Skipping let map for (g (f a))
Using let map for (f a)
Skipping let map for (f (g (f a)))
Using let map for (g (f a))
Skipping let map for (let ((_let_0 (f a))) (= (f (g (f _let_0))) (f (g _let_0))))
Using let map for (f (g (f (f a))))
Using let map for (f (g (f a)))
Skipping let map for b
Skipping let map for (= a b)
Using let map for a
Using let map for b
Skipping let map for (= b (f a))
Using let map for b
Using let map for (f a)
Skipping let map for (f b)
Using let map for b
Skipping let map for (= a (f b))
Using let map for a
Using let map for (f b)
Skipping let map for (g a)
Using let map for a
Skipping let map for (= (f a) (g a))
Using let map for (f a)
Using let map for (g a)
Skipping let map for (g b)
Using let map for b
Skipping let map for (= (f b) (g b))
Using let map for (f b)
Using let map for (g b)
Using let map for b
Using let map for (f a)
Using let map for (f a)
Using let map for (g a)
Using let map for (f (g (f (f a))))
Using let map for (f (g (f a)))
Using let map for (f b)
Using let map for (g b)
Using let map for (= a b)
Using let map for a
Using let map for (f b)
Using let map for (let ((_let_0 (f a))) (= (f (g (f _let_0))) (f (g _let_0))))
Using let map for (= a b)
Using let map for (= b (f a))
Using let map for (= a (f b))
Using let map for (= (f a) (g a))
Using let map for (= (f b) (g b))
Skipping let map for f
Skipping let map for g
Skipping let map for f
Skipping let map for f
Skipping let map for g
Skipping let map for f
Skipping let map for f
Skipping let map for f
Skipping let map for g
Skipping let map for f
Skipping let map for f
Skipping let map for g
Skipping let map for f
Skipping let map for f
Skipping let map for f
Skipping let map for f
Skipping let map for g
Skipping let map for f
Skipping let map for f
Skipping let map for g
Skipping let map for f
Skipping let map for f
Skipping let map for f
Skipping let map for g
Skipping let map for f
Skipping let map for f
Skipping let map for g
Skipping let map for f
Skipping let map for f
Skipping let map for f
Skipping let map for f
Skipping let map for f
Skipping let map for f
Skipping let map for g
Skipping let map for f
Skipping let map for f
Skipping let map for f
Skipping let map for g
Skipping let map for f
Skipping let map for f
Skipping let map for g
Skipping let map for g
Skipping let map for f
Skipping let map for g
Skipping let map for g
unsat

```

PR #3698, opened 2020-02-03
Example evaluation cache utility

ajreynol:
Adds a separate utility for managing the storage of terms on concrete inputs for examples conjectures.

This code is taken from SygusPbe; that code will be removed when this code is activated.

### DIFF
--- a/src/theory/quantifiers/sygus/example_eval_cache.cpp
+++ b/src/theory/quantifiers/sygus/example_eval_cache.cpp
@@ -51,18 +51,17 @@ Node ExampleEvalCache::addSearchVal(Node bv)
     return Node::null();
   }
   std::vector<Node> vals;
-  evaluateVec(bv, vals, true);
+  evaluateVec(bv, vals, false);
   Trace("sygus-pbe-debug") << "Add to trie..." << std::endl;
   Node ret = d_trie.addOrGetTerm(bv, vals);
   Trace("sygus-pbe-debug") << "...got " << ret << std::endl;
-  // Clean up the cache data if necessary: if the enumerated term
+  // Only save the cache data if necessary: if the enumerated term
   // is redundant, its cached data will not be used later and thus should
-  // be cleared now.
-  if (ret != bv)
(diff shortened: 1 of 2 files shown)


src/theory/quantifiers/sygus/example_eval_cache.cpp
```diff
@@ -0,0 +1,123 @@
...
+  Trace("sygus-pbe-debug") << "...got " << ret << std::endl;
+  // Clean up the cache data if necessary: if the enumerated term
+  // is redundant, its cached data will not be used later and thus should
+  // be cleared now.
+  if (ret != bv)
+  {
+    // immediately clear it
+    Trace("sygus-pbe-debug") << "...clear example cache" << std::endl;
+    clearEvaluationCache(bv);
+  }
```
HanielB: You could change this so that you just add to the cache if you have different results. It'd save doing this erasure operation. You will have the term and the evaluations anyway...

src/theory/quantifiers/sygus/example_eval_cache.h
```diff
@@ -0,0 +1,152 @@
...
+  /** evaluate builtin
+   *
+   * This returns the evaluation of bn on the i^th example for the
+   * function-to-synthesis associated with enumerator e. If there are not at
+   * least i examples, it returns the rewritten form of bn. For example, if bn =
+   * x+5, e is an enumerator for f in the example [EX#1] from SygusPbe, then
+   *   evaluateBuiltin( tn, bn, e, 0 ) = 7
+   *   evaluateBuiltin( tn, bn, e, 1 ) = 9
+   *   evaluateBuiltin( tn, bn, e, 2 ) = 10
+   * This
```
HanielB: ?

src/theory/quantifiers/sygus/example_eval_cache.cpp
ajreynol: Not sure it makes a difference, but I've changed it now.

PR #3700, opened 2020-02-03
Articulate proof-related debug statements in arith

alex-ozdemir:
Checking in a few debugging changes I made for the arith proof work.
Others might find them useful.

A list of changes:
  * centralize lemma reporting through an `outputLemma` member function
  which has a debug statement
  * Add debug statements for the other `output`* member functions.
  * implement (and use in a place or two) a `printProofTree` member
  function for the contraint class.
  * refine the debug tags for a number of debug statements (e.g. go from
  "pf::arith" to "pf::arith::explain") to allow for more fine-grained
  control of debug output.
  * Add some integer-related debug statements

### DIFF
--- a/src/theory/arith/constraint.cpp
+++ b/src/theory/arith/constraint.cpp
@@ -528,6 +528,7 @@ bool Constraint::hasTrichotomyProof() const {
 
 void Constraint::printProofTree(std::ostream& out, size_t depth) const
 {
+#if IS_PROOFS_BUILD
   const ConstraintRule& rule = getConstraintRule();
   out << std::string(2 * depth, ' ') << "* " << getVariable() << " [";
   if (hasLiteral())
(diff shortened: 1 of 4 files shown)


alex-ozdemir:
Fixed: the `Constraint::printProofTree` function (understandably) uses the `d_farkasCoefficients` member. However, the member is only defined in proof builds. Now, the `Constraint::printProofTree` function does something else in non-proof builds.

src/theory/arith/constraint.cpp
```diff
@@ -526,6 +526,50 @@ bool Constraint::hasTrichotomyProof() const {
   return getProofType() == TrichotomyAP;
 }
 
+void Constraint::printProofTree(std::ostream& out, size_t depth) const
+{
+#if IS_PROOFS_BUILD
```
yoni206: Maybe it would be better to also check if we are in debug and that the relevant debug/trace flag was turned on, as you do in some functions below. Otherwise an un-careful call to this function may lead to slower running times on debug (because of the loop inside).

alex-ozdemir: Interesting idea. I figured that the caller is responsible for gating this on the correct debug/trace flag, since this function wouldn't know the identity of the correct debug/trace flag: this function just gets an ostream.

PR #3701, opened 2020-02-03
Regression tests for arithmetic proofs.

alex-ozdemir:
These tests are designed to test interesting cases of arithmetic proofs,
such as mixing integers and reals, equalities, and tightening bounds.

Right now, they have the --no-check-proofs flag set, which prevents them
from testing the proof machinery. However, once we check that machinery
into master, we'll remove that flag, thus enabling the full effect of
the tests.

### DIFF
--- a/test/regress/regress0/arith/arith-eq.smt2
+++ b/test/regress/regress0/arith/arith-eq.smt2
@@ -1,0 +1,15 @@
+; COMMAND-LINE: --no-check-models  --no-check-unsat-cores --no-check-proofs
+; EXPECT: unsat
+(set-logic QF_LRA)
+(set-info :smt-lib-version 2.0)
+
+(declare-fun x () Real)
+(declare-fun y () Real)
+(declare-fun z () Real)
+
+(assert (= z 0))
+(assert (= (* 3 x) y))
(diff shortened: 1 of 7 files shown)


test/regress/regress0/arith/arith-eq.smt2
```diff
@@ -0,0 +1,15 @@
+; COMMAND-LINE: --no-check-models  --no-check-unsat-cores --no-check-proofs
```
yoni206: Is `--no-check-models` necessary?

alex-ozdemir: In haiku:
copy-paste strikes again

yoni206: :)
In that case, I also don't find the smtlib version that important either (I am assuming it is there for the same reason).

yoni206: BTW you have 6 syllables in the third line...

alex-ozdemir:
> I ... would like to verify -- do all these tests terminate quickly

Yes. They take less than a second on my machine

alex-ozdemir:
And sounds good. I'll remove those too.

PR #3707, opened 2020-02-03
Datatypes with nested recursion are not handled in TheoryDatatypes unless option is set

ajreynol:
Towards experimental support for non-simply recursive datatypes (https://github.com/ajreynol/CVC4/tree/dtNonSimpleRec). Builds a check for non-simple recursion in the DType class.  If a term of a datatype type is registered to TheoryDatatypes for a datatype that has nested recursion, we throw a LogicException unless the option dtNestedRec is set to true.  Also includes a bug discovered in the TypeMatcher utility and another in expr::getComponentTypes.

It also adds a unit test using the new API for a simple parametric datatype example as well, not related to nested recursion, as this was previously missing.


src/expr/dtype.cpp
```diff
@@ -471,21 +473,77 @@ bool DType::isInterpretedFinite() const
...
+  std::vector<TypeNode> processing;
+  if (!computeWellFounded(processing))
+  {
+    // not well-founded since no ground term can be constructed
+    Trace("datatypes-init")
+        << "DType::isWellFounded: false due to no ground terms." << std::endl;
+    d_wellFounded = -1;
+    return false;
+  }
+  // If we do not permit non-simple recursion for datatypes, we must
```
barrettcw: Can we avoid the double-negative here?  This is confusing.

ajreynol:
The code in this PR needs to be refactored to catch cases where the non-simple recursion occurs as the result of mutually recursive datatypes:
```
      DATATYPE
        list = cons(car: ns, cdr: list) | nil
        ns = elem(data: set(list))
     END;
``

**Comment by ajreynol** [comment]
This PR should be ready to go now.  I've added the unit tests and refactored the code to catch several missing cases of nested recursion.

File: src/expr/dtype.h [hunk 2]
```diff
@@ -489,6 +499,47 @@ class DType
    * Helper for mkGroundTerm and mkGroundValue above.
    */
   Node mkGroundTermInternal(TypeNode t, bool isValue) const;
+  /**
+   * This method is used to get strict subfield types of this datatype.
+   *
+   * A subfield type T of a datatype type D is a type such that a value of
```
yoni206, line 505:
```suggestion
   * A subfield type T of a datatype type D is a type such that a term of
```

?

src/expr/dtype.h
```diff
@@ -489,6 +499,47 @@ class DType
...
+   * type T may appear as a subterm of a value of type D.
```
yoni206, line 506:
```suggestion
   * type T may appear as a subterm of a term of type D.
```
?

src/expr/dtype.h
```diff
@@ -489,6 +499,47 @@ class DType
...
+   *
+   * A *strict* subfield type T of a datatype type D is a type such that a
+   * value v of type T may appear as a subterm of a value of D, and moreover
+   * v occurs as a strict subterm of a non-datatype term in that value.
+   *
+   * For example, the strict subfield types of:
```
yoni206, line 512:
```suggestion
   * For example, the strict subfield types of `T` in:
```
?

And simiarly for other examples?

src/expr/dtype.h
```diff
@@ -489,6 +499,47 @@ class DType
...
+   * definition due to the term (Container (singleton (cons 0 nil))), where
+   * 0 occurs as a subterm of (singleton (cons 0 nil)).
+   *
+   * For example, the strict subfield types of:
+   *   T -> Emp | Container(s : List)
+   *   List -> nil | cons( head : (Set T), tail: List)
+   * are { T, List }. Notice that T is a strict subfield type of itself since
+   * List is a subfield type of T and T is a strict subfield type of List.
+   *
+   * For example, the strict subfield types of:
```
yoni206, line 527: Thanks for all these examples. Could you also specify the non-strict subfield types in the examples?

src/expr/dtype.h
ajreynol: Value is more precise here. Notice that values of datatypes have term structure.  We don't want to say "term" here since this would include arbitrary ways of constructing terms of datatype type.

src/expr/dtype.h
ajreynol: Yes, I'll change.

src/expr/dtype.h
ajreynol: Sure.

src/expr/dtype.cpp
```diff
@@ -557,6 +565,133 @@ Node DType::mkGroundTermInternal(TypeNode t, bool isValue) const
...
+          // where D is the uninstantiated datatype type.  We should not view D
+          // as a strict subfield type of tn. Thus, we need a special case here
+          // which ignores the first child, when isStrictC is false.
+          for (unsigned i = 1, nchild = tn.getNumChildren(); i < nchild; i++)
+          {
+            expr::getComponentTypes(tn[i], types);
+          }
+          continue;
+        }
+      }
```
barrettcw, line 612: To make this code less confusing, add a comment here to indicate that we are now in the non-datatype case, meaning all proper subtypes need to be added to the list

src/expr/dtype.cpp
```diff
@@ -557,6 +565,133 @@ Node DType::mkGroundTermInternal(TypeNode t, bool isValue) const
...
+      bool hasTn = types.find(tn) != types.end();
+      Trace("datatypes-init")
+          << "Collect subfield types " << tn << ", hasTn=" << hasTn
+          << ", isStrictC=" << isStrictC << std::endl;
+      expr::getComponentTypes(tn, types);
+      if (!isStrictC && !hasTn)
+      {
+        // the top-level type is not a strict subfield type
```
barrettcw, line 620: Extend comment: the top-level type is added by getComponentTypes, so remove it if it was not already listed in types

src/expr/dtype.cpp
```diff
@@ -557,6 +565,133 @@ Node DType::mkGroundTermInternal(TypeNode t, bool isValue) const
...
+        Assert(types.find(tn) != types.end());
+        types.erase(tn);
+      }
+    }
+  }
+  // now, go back and add all strict subfield types from datatypes if
+  // not done so already
```
barrettcw, line 627: This is because getComponentTypes does not recurse into datatype definitions?  Explain that here.

src/expr/dtype.h
```diff
@@ -287,6 +287,16 @@ class DType
    * violated.
    */
   bool isWellFounded() const;
+  /**
+   * Does this datatype have nested recursion? This is true if this datatype
+   * definition does not contain itself as a strict subfield type, or a variant
```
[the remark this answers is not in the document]
ajreynol: Ah yes, good catch (I switched polarity at some point and missed this).

src/expr/dtype.h
```diff
@@ -489,6 +499,51 @@ class DType
    * Helper for mkGroundTerm and mkGroundValue above.
    */
   Node mkGroundTermInternal(TypeNode t, bool isValue) const;
+  /**
+   * This method is used to get strict subfield types of this datatype.
+   *
+   * A subfield type T of a datatype type D is a type such that a value of
+   * type T may appear as a subterm of a value of type D.
+   *
+   * A *strict* subfield type T of a datatype type D is a type such that a
```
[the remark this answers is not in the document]
ajreynol: Sounds good.

PR #3398, opened 2019-10-17
Move more string utility functions

ajreynol:
This is work towards splitting a "core solver" object from TheoryStrings.

This moves global functions from TheoryStrings to InferenceManager/SolverState, making them accessible in the future by modules that have references to these objects.

It also corrects an issue where we were maintaining two `d_conflict` fields.

PR #3711, opened 2020-02-04
Revert semantic change from refactoring

4tXJ7f:
In PR #3398, a minor change was introduced: One of the checks for
whether there is a conflict or not was flipped
(https://github.com/CVC4/CVC4/pull/3398/files#diff-62fffb0d9df0385909621e424bdcef72R1652). This commit reverts that change.


ajreynol:
Wow, thanks.

Issue #1123, opened 2017-09-21
Support set cardinality for sets of finite element types


ajreynol:
Currently we throw an error when set cardinality is applied to a set with finite type:

https://github.com/CVC4/CVC4/blob/master/src/theory/sets/theory_sets_private.cpp#L625

We should support this combination.

PR #3712, opened 2020-02-04
Univeset Cardinality constraints for infinite types

mudathirmahgoub:
This pull request adds cardinality  constraints for infinite types if the universe sets are explicitly. The constraints are similar to finite types except that the cardinality of `univset` now is a variable (`skolem int`) instead of a constant. 

With this pull request the user can write a constraint like 
```smt2
(assert (< (card (as univset (Set Int)) 10)))
```
This PR is somehow related to issue #1123


src/theory/sets/cardinality_extension.cpp
```diff
@@ -121,22 +126,42 @@ void CardinalityExtension::checkFiniteType(TypeNode& t)
...
 
-  Node typeCardinality = nm->mkConst(Rational(card.getFiniteCardinality()));
+  Node typeCardinality;
+  if (t.isInterpretedFinite())
+  {
+    typeCardinality = nm->mkConst(Rational(card.getFiniteCardinality()));
+  }
+  else
+  {
+    std::map<TypeNode, Node>::iterator skolemIt =
```
ajreynol: Is this skolem necessary? Can't we just leave typeCardinality null in this case and then not send the lemma?

src/theory/sets/cardinality_extension.h
```diff
@@ -324,18 +324,22 @@ class CardinalityExtension
   void checkNormalForm(Node eqc, std::vector<Node>& intro_sets);
 
   /**
-   * add cardinality lemmas for each finite type
+   * add cardinality lemmas for each type except uninterpreted sorts with
+   * finite-model-find
    */
-  void checkFiniteTypes();
+  void checkUnivsetCardinality();
```
ajreynol: We should have a better name for this function.  Maybe `checkCardinalityExtended`  for checking cardinality extended (handling cases that go beyond Bansal et al 2016).

src/theory/sets/cardinality_extension.cpp
mudathirmahgoub: Right. Nice catch 👍

src/theory/sets/cardinality_extension.h
mudathirmahgoub: Sure

src/theory/sets/cardinality_extension.cpp
```diff
@@ -121,22 +126,26 @@ void CardinalityExtension::checkFiniteType(TypeNode& t)
...
 
-  // add subset lemmas for sets, and membership lemmas for negative members
+  // add subset lemmas for sets and membership lemmas for negative members
   for (Node& representative : representatives)
   {
-    // the universe set is a subset of itself
-    if (representative != d_ee.getRepresentative(univ))
+    if (representative
+        != d_ee.getRepresentative(
+            univ))  // the universe set is a subset of itself
```
ajreynol: Nit: can you put the comment on the previous line so that this condition is not ugly after clang-format?

src/theory/sets/cardinality_extension.h
```diff
@@ -324,18 +324,22 @@ class CardinalityExtension
   void checkNormalForm(Node eqc, std::vector<Node>& intro_sets);
 
   /**
-   * add cardinality lemmas for each finite type
+   * add cardinality lemmas for each type except uninterpreted sorts with
```
ajreynol: Update this documentation, perhaps refer to the documentation below.

PR #3716, opened 2020-02-05
Add `ArithProof::{printInteger,getLfscFunction}`

alex-ozdemir:
* `printInteger` will be useful for printing integers in LFSC proofs
* `getLfscFunction` will be useful for printing terms more succinctly.


alex-ozdemir:
Note: `getLfscFunction` is currently dead code, and is targeting the LIRA signature, hence the inclusion of comparisons with suffix `IntReal`.

src/proof/arith_proof.cpp
```diff
@@ -870,6 +870,56 @@ void LFSCArithProof::printOwnedSort(Type type, std::ostream& os) {
...
+      Unreachable() << "Tried to get the operator for a non-operator kind: " << n.getKind();
+  }
+  std::string typeString;
+  if (n.getType().isInteger()) {
+    typeString = "Int";
+  } else if (n.getType().isReal()) {
+    typeString = "Real";
+  } else { // Boolean
+    if (n[0].getType().isInteger()) {
+      typeString = "IntReal";
```
yoni206: Can't this ever be "Int"? Like in ">=_Int"?

src/proof/arith_proof.h
```diff
@@ -97,6 +106,15 @@ class LFSCArithProof : public ArithProof {
...
   static void printRational(std::ostream& o, const Rational& r);
 
+  /**
+   * Print an integer in LFSC format.
+   *        e.g. 5 or (~ 1)
+   *
+   * @param o ostream to print to.
+   * @param i the integer to print
+   */
+  static void printInteger(std::ostream& o, const Integer& i);
```
yoni206: I am not sure these two functions need to be static, maybe they can be functions that are external to the class (I am following [these](https://github.com/CVC4/CVC4/wiki/Code-Style-Guidelines#nonmember-static-member-and-global-functions) instructions here). On the other hand, the rest of the similar printing functions are static so it would be weird and not uniform to change only this one. I'll leave it for you to decide, just wanted to mention this in case you haven't thought of that.

alex-ozdemir: That's a good question. I hadn't noticed that. I would prioritize uniformity over style guide compliance.

Perhaps I'll submit a PR fixing the style later.

src/proof/arith_proof.cpp
alex-ozdemir: Actually no. If the constant on the right is an integer, we still lift it to a Real.

alex-ozdemir: Really good question though. The refinement that you suggest would be sound, but it doesn't allow any new reasoning to be applied.

yoni206: How do you know that it is a constant? Maybe assert it?

alex-ozdemir: Good idea. Do you think that assertion would be better located here, or in `printOwnedTermAsType` (which will call this soon)?

yoni206: Personally I would probably put it in both in fear of future mis-use of both (my suggestion is an assertion that will run only on debug, not on production). But this might be an overkill.  Completely up to you.

alex-ozdemir: Sounds good. I feel weird about putting it here because there's nothing about getting a predicates name that implies the right hand side is constant. The constant-ness is pretty unrelated. What I'll do it document the fact that `Int` predicates are never returned.

PR #3717, opened 2020-02-05
Propagate expected types through UF arguments

alex-ozdemir:


src/proof/uf_proof.cpp
```diff
@@ -649,11 +649,13 @@ void LFSCUFProof::printOwnedTermAsType(Expr term,
...
   }
   os << func << " ";
+  Assert(func.getType().isFunction());
+  FunctionType funcType = static_cast<FunctionType>(func.getType());
   for (unsigned i = 0; i < term.getNumChildren(); ++i) {
 
     bool convertToBool = (term[i].getType().isBoolean() && !d_proofEngine->printsAsBool(term[i]));
     if (convertToBool) os << "(f_to_b ";
-    d_proofEngine->printBoundTerm(term[i], os, map);
+    d_proofEngine->printBoundTerm(term[i], os, map, TypeNode::fromType(funcType.getArgTypes()[i]));
```
yoni206: How about computing `funcType.getArgTypes()` before the loop?

src/proof/uf_proof.cpp
```diff
@@ -649,11 +649,13 @@ void LFSCUFProof::printOwnedTermAsType(Expr term,
     os << "(apply _ _ ";
   }
   os << func << " ";
+  Assert(func.getType().isFunction());
+  FunctionType funcType = static_cast<FunctionType>(func.getType());
```
yoni206: In another PR you eventually preferred to use a node representation instead of an expr representation of types.
This is also possible here, by using `fromExpr` function of `Node` on `term`, and then use `getOperator().getType().getArgTypes()`. It could be better than a static cast, though I am not sure. It would slightly diminish the usage of Expr but not entirely. Your call, and maybe @4tXJ7f has an opinion here?

alex-ozdemir: Since this is a small-scope local variable I'm not immediately worried about using `Expr`. That being said, I defer to Andres. I don't actually know how these structures work.

src/proof/uf_proof.cpp
alex-ozdemir: This shouldn't be necessary. Its a const method and the compiler should be able to lift it automatically.

yoni206: I seem to recall discussing something similar with @aniemetz and would be interested in her opinion on this! I am not sure what assumptions do we adopt about the compiler.

src/proof/uf_proof.cpp
alex-ozdemir: I'll just change it. Thanks for the suggestion!

aniemetz: Internally, we should never use Expr, always Node. Expr is/was the old API layer, and should not be used in the solver internals.

src/proof/uf_proof.cpp
aniemetz: We don't want (never) to rely on the compiler here (we don't have control over the compiler, and which compiler is used). And even if it is static, you could have side effects that make it impossible to lift. It's best practice to move such function calls to outside of the loop.

alex-ozdemir:
I think I've addressed the comments. Let me know if you see anything else!

src/proof/uf_proof.cpp
4tXJ7f: Yes, we should not use `Expr`-layer classes internally as @aniemetz said. I would also like to add that this code actually results in undefined behavior: You can't cast an object of type `Type` to `FunctionType`. It kind of works in practice because `FunctionType` has no additional data members nor virtual methods (see also #3420) but is dangerous.

src/proof/uf_proof.cpp
4tXJ7f: I agree with @aniemetz. It seems unlikely that the compiler could do this here and even if it could, it's best not to rely on it.

src/proof/uf_proof.cpp
alex-ozdemir: Good to know. I know less about class hierarchies in C++ than I would like to.

This is gone now.

PR #3722, opened 2020-02-06
Generalize containsQuantifiers to hasClosure

ajreynol:
This generalizes the containsQuantifiers function to hasClosure and moves it to expr/node_algorithm.


src/expr/node_algorithm.cpp
```diff
@@ -229,6 +229,43 @@ bool hasFreeVar(TNode n)
...
+    bool hasC = false;
+    if (n.isClosure())
+    {
+      hasC = true;
+    }
+    else
+    {
+      for (auto i = n.begin(); i != n.end() && !hasC; ++i)
+      {
+        hasC = hasBoundVar(*i);
```
HanielB: Should this have `hasClosure`?

ajreynol: Yes.

PR #3723, opened 2020-02-06
Refactor check-model handling in SmtEngine

### LINKED ISSUE #3718: --check-models Fatal Faliure at CVC4/src/smt/smt_engine.cpp:4831
Hi,
For this formula,
```
(assert (= (sqrt 0) 0))
(check-sat)
```
CVC4 `--check-models` throws out a fatal failure:
```
Fatal failure within void CVC4::SmtEngine::checkModel(bool) at CVC4/src/smt/smt_engine.cpp:4831
Check failure
 quantifiers::QuantifiersRewriter::containsQuantifiers(n)
Aborted
```

OS: Ubuntu 18.04
Revision: ea43556

ajreynol:
This PR refactors the way SmtEngine processes assertions in check-model. 

It removes special cases for quantifiers / (quantified) rewrite rules, since other theories (e.g. non-linear with transcendental functions) have similar limitations for checking assertions.

The changes in behavior:
- Quantifiers of kind REWRITE_RULE are no longer skipped and instead processed uniformly,
- We do not throw assertion failures related to the expectations on the classes of nodes that can be checked,
- We throw a warning instead of an error for assertions that are non-constant, regardless of whether the logic is quantifier-free.

This fixes #3718.


src/smt/smt_engine.cpp
```diff
@@ -4821,36 +4810,56 @@ void SmtEngine::checkModel(bool hardFailure) {
...
+    if (!n.isConst())
+    {
+      // Not constant, print a less severe warning message here.
+      Warning() << "Warning : SmtEngine::checkModel(): cannot check simplified "
+                   "assertion : "
+                << n << endl;
+      continue;
+    }
+    // Assertions that simplify to false result in an InternalError or
+    // Warning being thrown below (when hardFailure is true).
```
4tXJ7f:
```suggestion
    // Warning being thrown below (when hardFailure is false).
```

PR #3727, opened 2020-02-07
Split strings finite model finding strategy

ajreynol:
Moves strings finite model finding strategy to its own file.


src/theory/strings/strings_fmf.cpp
```diff
@@ -0,0 +1,124 @@
...
+  Trace("strings-dstrat-reg")
+      << "presolve: register decision strategy." << std::endl;
+  std::vector<Node> inputVars;
+  for (NodeSet::const_iterator itr = d_inputVars.begin();
+       itr != d_inputVars.end();
+       ++itr)
+  {
+    inputVars.push_back(*itr);
+  }
+  d_sslds->initialize(inputVars);
```
4tXJ7f: Why isn't `StringsFmf::StringSumLengthDecisionStrategy::initialize()` taking a `const NodeSet&` as a parameter making this conversion to a vector unnecessary?

ajreynol: I'd prefer it having a universal interface (conversely it were later used in a context where we had a vector, the opposite translation is a lot less natural).

PR #3733, opened 2020-02-08
Use example evaluation cache instead of sygus PBE 

ajreynol:
This updates various parts of the code (sygus datatypes extension, cegis, sygus unif io) use example evaluation cache instead of the sygus PBE module.

This PR reduces recomputation of evaluations for conjectures that are in examples form.  It also refactors CEGIS for fast enumerators to use a separate specialized function (checkRefinementLemmas) instead of being a special case of getRefinementEvalLemmas.

It also reverts a change to example evaluation cache which was incorrect (it was not robust to the case where the evaluation was already cached due to another source).


src/theory/quantifiers/sygus/sygus_invariance.cpp
```diff
@@ -91,15 +91,15 @@ void EquivSygusInvarianceTest::init(
 {
   // compute the current examples
   d_bvr = bvr;
-  if (aconj->getPbe()->hasExamples(e))
+  Assert(tds != nullptr);
+  Node f = tds->getSynthFunForEnumerator(e);
```
HanielB: this is not being used

src/theory/quantifiers/sygus/sygus_invariance.cpp
```diff
@@ -149,9 +149,11 @@ bool EquivSygusInvarianceTest::invariant(TermDbSygus* tds, Node nvn, Node x)
     if (!d_enum.isNull())
     {
       bool ex_equiv = true;
+      ExampleEvalCache* eec = d_conj->getExampleEvalCache(d_enum);
+      Assert(eec != nullptr);
       for (unsigned j = 0; j < d_exo.size(); j++)
```
HanielB: cache size, since you're already here...

src/theory/quantifiers/sygus/example_eval_cache.cpp
```diff
@@ -51,17 +51,16 @@ Node ExampleEvalCache::addSearchVal(Node bv)
...
   Trace("sygus-pbe-debug") << "...got " << ret << std::endl;
   // Only save the cache data if necessary: if the enumerated term
   // is redundant, its cached data will not be used later and thus should
   // be discarded.
-  if (ret == bv)
+  if (ret != bv)
   {
-    std::vector<Node>& eocv = d_exOutCache[bv];
-    eocv.insert(eocv.end(), vals.begin(), vals.end());
+    clearEvaluationCache(bv);
```
HanielB: Can you comment on why you had to revert this?

ajreynol: It's in the PR description.  The issue is two-fold:
- If it was already cached prior to entering this function and is not redundant, we should not add to the cache again (we were doubling the size of  the example out vector, leading to segfaults),
- If it was already cached prior to entering this function and is redundant, we should clear the cache immediately.

HanielB: I see now. I think you should add these comments to the code. It's an insight that is hard to gain otherwise.

PR #3736, opened 2020-02-10
Refactor `CoreSolver::processSimpleNEq()`

4tXJ7f:
This commit refactors and documents `CoreSolver::processSimpleNEq()`.
This method processes equalities between normal forms.

### DIFF
--- a/src/theory/strings/core_solver.cpp
+++ b/src/theory/strings/core_solver.cpp
@@ -1014,425 +1014,515 @@ void CoreSolver::processNEqc(std::vector<NormalForm>& normal_forms)
 }
 
 void CoreSolver::processSimpleNEq(NormalForm& nfi,
-                                     NormalForm& nfj,
-                                     unsigned& index,
-                                     bool isRev,
-                                     unsigned rproc,
-                                     std::vector<InferInfo>& pinfer)
+                                  NormalForm& nfj,
+                                  unsigned& index,
+                                  bool isRev,
+                                  unsigned rproc,
+                                  std::vector<InferInfo>& pinfer)
 {
-  std::vector<Node>& nfiv = nfi.d_nf;
-  std::vector<Node>& nfjv = nfj.d_nf;
   NodeManager* nm = NodeManager::currentNM();
   eq::EqualityEngine* ee = d_state.getEqualityEngine();
+
+  const std::vector<Node>& nfiv = nfi.d_nf;
+  const std::vector<Node>& nfjv = nfj.d_nf;
   Assert(rproc <= nfiv.size() && rproc <= nfjv.size());
-  bool success;
-  do {
-    success = false;
-    //if we are at the end
-    if (index == (nfiv.size() - rproc) || index == (nfjv.size() - rproc))
+  while (true)
+  {
+    bool lhsDone = (index == (nfiv.size() - rproc));
+    bool rhsDone = (index == (nfjv.size() - rproc));
+    if (lhsDone && rhsDone)
     {
-      if (index == (nfiv.size() - rproc) && index == (nfjv.size() - rproc))
+      // We are done with both normal forms
+      break;
+    }
+    else if (lhsDone || rhsDone)
+    {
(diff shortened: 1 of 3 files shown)


src/theory/strings/core_solver.cpp
```diff
@@ -1014,425 +1014,513 @@ void CoreSolver::processNEqc(std::vector<NormalForm>& normal_forms)
...
+      if (!isRev)
+      {  // FIXME
+        NormalForm::getExplanationForPrefixEq(nfi, nfj, -1, -1, info.d_ant);
+        ProcessLoopResult plr =
+            processLoop(lhsLoopIdx != -1 ? nfi : nfj,
+                        lhsLoopIdx != -1 ? nfj : nfi,
+                        lhsLoopIdx != -1 ? lhsLoopIdx : rhsLoopIdx,
+                        index,
+                        info);
+        if (plr == ProcessLoopResult::INFERENCE)
```
ajreynol: Should we be returning for ProcessLoopResult::CONFLICT?  (The code a few lines below used to check for ProcessLoopResult::SKIPPED.

src/theory/strings/core_solver.cpp
```diff
@@ -1014,425 +1014,513 @@ void CoreSolver::processNEqc(std::vector<NormalForm>& normal_forms)
...
+        info.d_conc = nc.eqNode(isRev ? utils::mkNConcat(sk, firstChar)
+                                      : utils::mkNConcat(firstChar, sk));
+        info.d_new_skolem[LENGTH_SPLIT].push_back(sk);
+        info.d_id = INFER_SSPLIT_CST;
+        pinfer.push_back(info);
+        break;
+      }
+      else
+      {
+        break;
```
ajreynol: Can you comment what this break corresponds to?

4tXJ7f: This is a bit tricky. It corresponds to the case where `isRev` is `true`. In this case, we would skip the following block in the original code: https://github.com/CVC4/CVC4/blob/499f1c70dbc869e2b1dd5f8a7b98d45bee2d29cc/src/theory/strings/core_solver.cpp#L1252-L1353 . After skipping this block, we essentially end up at the outer while loop and break out of it because `success` is `false`. In the refactored code, we have to break because we don't want to do a V-Split when `isRev` is `true`.

src/theory/strings/core_solver.cpp
4tXJ7f: Good point, I've fixed this.

src/theory/strings/core_solver.cpp
ajreynol: Ah I see, can you add a comment to the code that says this?  I do remember this was the case but it looks a bit scary when we break without an inference.

4tXJ7f: I've added the comment, PTAL.

4tXJ7f:
For completeness here are the benchmark results comparing master against this PR (300s timeout, `--strings-exp --re-elim-agg`):
```
          DIRECTORY |                                     master0210 |                       refactorProcessSimpleNEq | 
          Benchmark |  Stat  Slvd   Tot  To Mo Err   Cpu[s]  Mem[MB] |  Stat  Slvd   Tot  To Mo Err   Cpu[s]  Mem[MB] | 
               pyex |    ok 25161 25421 260  0   0 152022.2 458681.1 |    ok 25165 25421 256  0   0 143421.0 456616.7 | 
 str_int_benchmarks |    ok 21504 21573  69  0   0  53038.2 180430.0 |    ok 21507 21573  66  0   0  52120.3 180890.2 | 
        transformed |    ee  8494  9183 251  0 438  78343.4  42186.1 |    ee  8496  9183 249  0 438  77948.2  42051.2 | 
           __totals |    ee 55159 56177 580  0 438 283403.8 681297.2 |    ee 55168 56177 571  0 438 273489.5 679558.1 | 
```

Overall, this PR performs slightly better.

PR #3741, opened 2020-02-11
Switch to th_lira.plf

alex-ozdemir:
Switches arith_proof.cpp from th_lra to th_lira.

Changes:

  * Eliminate the `d_realMode` hack.
     * instead: modify `printOwnedTermAsType` prints as integers OR
       reals, depending on `expectedType`.
     * simultaneously: write `printOwnedTermAsType` more concisely
     * also: reimplement `printOwnedSort`.
  * Change to the LIRA axioms:
     * Because they reason about bound types using side conditions, we
       no longer need to worry about choosing the correct strictness for
       our axiom.
        * This allows us to cut out a lot of code, rewriting & shrinking
          `printTheoryLemmaProof`.
     * They also have different names.
        * This requires us to change a lot of string literals
   * enable proof-checking for many tests.


alex-ozdemir:
Also, a note: this is a big PR:

I could shrink it by making dead versions of the 2 large changed methods one at a time, and then enabling them.

If you prefer that, I'll do so.

proofs/signatures/CMakeLists.txt
```diff
@@ -16,7 +16,7 @@ set(core_signature_files
     th_bv_rewrites.plf
     th_real.plf
     th_int.plf
-    th_lra.plf
+    th_lira.plf
```
yoni206: I wonder if `th_lra.plf` should be now deleted? or maybe safer to write a comment at its first line?

src/proof/arith_proof.cpp
```diff
@@ -656,7 +658,7 @@ Node ProofArith::toStreamRecLFSC(std::ostream& out,
 }
 
 ArithProof::ArithProof(theory::arith::TheoryArith* arith, TheoryProofEngine* pe)
-  : TheoryProof(arith, pe), d_recorder(), d_realMode(false)
+  : TheoryProof(arith, pe), d_recorder()
```
yoni206: Seems like you can remove this field from the .h file, right?

test/regress/regress0/arith/bug569.smt2
```diff
@@ -1,4 +1,4 @@
-; COMMAND-LINE: --no-check-unsat-cores --no-check-proofs
+; COMMAND-LINE: --no-check-unsat-cores
```
yoni206: I suggest checking coverage of these tests w.r.t. new proof code to make sure that all new derivation rules are exercised. 
https://github.com/CVC4/CVC4/wiki/Developer-Guide#code-coverage-of-unitsystemregression-tests

Alternatively, you could manually generate and check the proof or temporarilly plant a bug. The reason I am suggesting this is that sometimes proofs pass because they are not even generated because of some flag that was turned on in the smt engine or something like that.

test/regress/regress0/arith/integers/ackermann5.smt2
```diff
@@ -1,4 +1,4 @@
-; COMMAND-LINE: --ackermann --no-check-models --no-check-proofs --no-check-unsat-cores
+; COMMAND-LINE: --ackermann --no-check-models --no-check-unsat-cores
```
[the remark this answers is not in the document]
alex-ozdemir: Sounds good!

proofs/signatures/CMakeLists.txt
alex-ozdemir: I generally try to avoid checking commented-out code into version control, preferring to let the history keep it.

I'll remove `th_lra.plf` in a subsequent PR.

src/proof/arith_proof.cpp
alex-ozdemir: You are correct! Nice catch! Thanks!

test/regress/regress0/arith/bug569.smt2
alex-ozdemir: Thanks! I've done this. Some paths are actually being missed. Specifically, I'm having a hard time getting the `IntReal` tightening rules to go off. I'll keep working on this.

test/regress/regress0/arith/arith-tighten-1.smt2
```diff
@@ -1,4 +1,4 @@
-; COMMAND-LINE: --no-check-unsat-cores --no-check-proofs
+; COMMAND-LINE: --no-check-unsat-cores
 ; EXPECT: unsat
 (set-logic QF_LIRA)
```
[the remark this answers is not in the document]
alex-ozdemir: Done, as `tighten-arith-2.smt2`.

alex-ozdemir:
I haven't quite responded to all of your comments yet. I'll let you know when I have.

test/regress/regress0/arith/bug569.smt2
alex-ozdemir: Update: I've realized that the rewriter purifies `IntReal` terms into `IntInt` terms, so the former kind of term will never occur in proofs. I've removed the code for the applicable rules, and have added assertions stating that indeed no `IntReal` terms occur in proofs.

I didn't modify that the term-printing code, because they can occur in the input, and must therefor be printable.

alex-ozdemir:
Okay, I think that I've addressed your comments now, @yoni206!

test/regress/regress0/arith/bug569.smt2
alex-ozdemir: Addition: I'll remove the axioms for `IntReal` in a later PR.
I'm not doing it now because this will require me to change some of the tests in `th_lira_test.plf`, and this PR is already big enough.

PR #3742, opened 2020-02-11
Introduce tables in the rewriter

4tXJ7f:
This commit adds tables in the rewriter that store which function should
be used to rewrite which kind. We have separate tables for `EQUAL`
because every theory has its own equality rewriter.


4tXJ7f:
FYI, a refactored rewriter looks as follows: https://github.com/4tXJ7f/CVC4/commit/ea626c181d7e11fc16536b4baf407f2ab2e18d66.

src/theory/rewriter.cpp
```diff
@@ -81,15 +81,45 @@ struct RewriteStackElement {
...
     return node;
   }
-  Rewriter& rewriter = getInstance();
-  return rewriter.rewriteTo(theoryOf(node), node);
+  return getInstance().rewriteTo(theoryOf(node), node);
+}
+
+void Rewriter::registerPreRewrite(
+    Kind k, std::function<RewriteResponse(RewriteEnvironment*, TNode)> fn)
+{
```
mpreiner: What if k is EQUAL here? Should we add an assertion `Assert(k != EQUAL) << "Register pre-rewrites for EQUAL with registerPreRewriteEqual."`

src/theory/rewriter.cpp
```diff
@@ -81,15 +81,45 @@ struct RewriteStackElement {
...
+  d_preRewriters[k] = fn;
+}
+
+void Rewriter::registerPostRewrite(
+    Kind k, std::function<RewriteResponse(RewriteEnvironment*, TNode)> fn)
+{
```
mpreiner: Ditto.

src/theory/rewriter.cpp
aniemetz: Why do we have two different functions to register a pre- and post-rewrite anyways?

4tXJ7f: Huh, I somehow missed your comment @mpreiner. That's a good point, will do. @aniemetz some rewriters perform different rewrites in pre- and post-rewrites (e.g. the BV rewriter).

aniemetz: No, what I meant is, why do we have a different function to register EQ and the rest?

4tXJ7f: Ah because equalities are rewritten by all theories (depending on the operands) whereas other kinds are only associated with a single theory.

aniemetz: We could still have a special handling in the register function for equalities no? I still don't get why we need 2 functions for this.

mpreiner: It's more convenient this way. Otherwise, we would have to pass the theory to all rewrites, which is overkill.

PR #3747, opened 2020-02-11
Fix approximate bounds for transcendental functions whose model values rewrite

### LINKED ISSUE #3725: Fatal failure at CVC4/src/theory/arith/nl_model.cpp:1040
Hi,
For this formula,
```
(declare-fun a () Real)
(assert (distinct (sin a) 0.0))
(check-sat)
```
```
Fatal failure within bool CVC4::theory::arith::NlModel::simpleCheckModelMsum(const std::map<CVC4::NodeTemplate<true>, CVC4::NodeTemplate<true> >&, bool) at CVC4/src/theory/arith/nl_model.cpp:1040
Check failure
 false
Aborted
```

OS: Ubuntu 18.04
Revision: c98585d

ajreynol:
We weren't setting approximate values for some transcendental functions whose model values rewrite. This weakened our ability to answer "sat", caught by an assertion failure.

Fixes #3725 (that benchmark now times out without an assertion failure).


src/theory/arith/nonlinear_extension.cpp
```diff
@@ -793,15 +793,40 @@ bool NonlinearExtension::checkModel(const std::vector<Node>& assertions,
...
+      }
+      if (!bl.isNull() && !bu.isNull())
+      {
+        // We have rewritten an application of a transcendental function
+        // based on the current model values.It could be that the model value
+        // rewrites sin(x) ---> sin(-c) ---> -sin(c), thus we need
+        // to negate the bounds in this case.
+        if (atf.getKind() != tf.getKind())
+        {
+          if (atf.getKind() == MULT && atf.getNumChildren() == 2
```
ahmed-irfan: shouldn't we do this kind of reasoning inside the function `getTfModelBounds`
(https://github.com/CVC4/CVC4/blob/3e91545ab5914fe65f57544d0d00a7dd9e897e64/src/theory/arith/nonlinear_extension.cpp#L3770)

ajreynol: I had to refresh my memory on this.  It turns out we don't have to.

In that function, we compute the model value of `f(x)`, viewing it as a term, whereas in the code I changed above, we are computing the value of `f(x^M)` where x^M is the model value of x, which can rewrite to `-sin(c)` for instance.  This isn't necessary here.

PR #3748, opened 2020-02-11
Forcing rewrite pass rather than asserting if formula has been rewritten

HanielB:
This ensures that users or developers don't accidentally break the solver either via options or changing the SMT engine flow so that the formula is not rewritten up to a given point.


src/smt/smt_engine.cpp
```diff
@@ -3342,13 +3342,8 @@ void SmtEnginePrivate::processAssertions() {
...
 
-  // Assertions ARE guaranteed to be rewritten by this point
-#ifdef CVC4_ASSERTIONS
-  for (unsigned i = 0; i < d_assertions.size(); ++i)
-  {
-    Assert(Rewriter::rewrite(d_assertions[i]) == d_assertions[i]);
-  }
-#endif
+  // Assertions MUST BE guaranteed to be rewritten by this point
+  d_passes["rewrite"]->apply(&d_assertions);
```
4tXJ7f: We can remove the then branch from the if statement just above this (to avoid calling the rewriting pass twice when we have unsat cores enabled).

HanielB: Cool, will do. Do you know why `apply-subst` is incompatible with unsat-cores so I can say it in the new `if`?

4tXJ7f: Likely just because we lack proper dependency tracking for the substitutions (that's pretty much the only work that's required for unsat cores).

Issue #3565, opened 2019-12-13
make test failure


kostis:
[Apologies if this is a known issue or completely uninteresting...]

The installation instructions recommend building CVC4 with
`./configure.sh && cd build && make && make check`
to see if the build has succeeded, but my fingers are used to typing `make test` instead and I often do that (instead or in addition to make check).

Turns out that `make test` does do something related to what its name implies (runs some regression tests, although it takes considerably longer than `make check`), and it sort of works, but not quite.

One of the tests consistently fails, at least for me -- I am running on Debian 9 with gcc version 6.3.0 20170516 (Debian 6.3.0-18+deb9u1), if it makes any difference.  The `make test` command finishes with:
<pre>
99% tests passed, 1 tests failed out of 1934

Label Time Summary:
regress0    =  61.05 sec (1098 tests)
regress1    = 100.21 sec (717 tests)
regress2    = 100.20 sec (92 tests)
regress3    = 591.34 sec (14 tests)
regress4    = 959.80 sec (6 tests)
system      =   0.05 sec (7 tests)

Total Test time (real) = 1813.27 sec

The following tests FAILED:
        1925 - regress4/comb2.shuffled-as.sat03-420.smtv1.smt2 (Failed)
Errors while running CTest
Makefile:72: recipe for target 'test' failed
make: *** [test] Error 8
</pre>

Is this a known issue?  If that particular test does not fail just for me, can it be fixed or removed if it's not interesting/relevant anymore?  I find it a pity to have a test suite of 1934 tests and not being able to reliably run it simply because one of its tests fails...

Thanks for a wonderful repo/tool!

4tXJ7f:
This is very likely due to the test timing out (tests at higher regression levels tend to take much longer to run). `make check` only tests up to regression level 2, which is what we generally use for testing. The test timeout can be changed by setting the environment variable `TEST_TIMEOUT` to the number of seconds that an invocation of CVC4 may take (default is 600s).

kostis:
Yep, this was it.

With the default timeout value the three tests that take longest to run are:
<pre>
   regress3/pp-regfile.smtv1.smt2 ............................... Passed  564.12 sec
   regress4/comb2.shuffled-as.sat03-420.smtv1.smt2 .............. ***Failed  655.89 sec
   regress4/instance_1151.smtv1.smt2 ............................ Passed  228.85 sec
</pre>

while, on my machine, if I run with `TEST_TIMEOUT=1000 make test` all tests pass:
[...]

PR #3750, opened 2020-02-11
run_regression: Distinguish between timeout and failure.

mpreiner:
If --use-skip-return-code is enabled and a regression test times out it
will return EXIT_SKIP instead of EXIT_FAILURE.


4tXJ7f:
Arguably, this also touches on #3565.

PR #3753, opened 2020-02-12
[Python] Properly destroy CVC4 object

### LINKED ISSUE #1523: [Python] Properly destroy CVC4 solver object
See https://github.com/pysmt/pysmt/issues/417.

4tXJ7f:
Fixes #1523. The issue was that our Python proxy classes did not have
`self.thisown` set to 1, so the C++ classes were not destroyed when the
proxy classes were garbage collected. This was introduced in commit
7b3aaa55f2e544b752640f99b3516292bda181b7, presumably because `SmtEngine`
holds a reference to `ExprManager` that Python does not know about. This
commit fixes the issue by adding a member to the `SmtEngine` proxy class
that signals to Python that it cannot destroy the `ExprManager` before
the `SmtEngine`.


4tXJ7f:
I've asked the pySMT devs if they could test the fix from their side. It is probably a good idea to wait with merging this PR until they had a chance to test it.

PR #3759, opened 2020-02-14
Add temporary global API conversion utilities.

ajreynol:
We will need temporary conversion utilities to be accessible globally for the parser migration.  This also includes conversions between internal and external kinds.

This is currently a blocker for further PRs related to parser migration, so marking as "major".


src/api/cvc4cpp.h
```diff
@@ -2689,6 +2690,20 @@ class CVC4_PUBLIC Solver
   std::unique_ptr<Random> d_rng;
 };
 
+// !!! These only temporarily public until the parser is fully migrated to the
```
4tXJ7f:
```suggestion
// !!! Only temporarily public until the parser is fully migrated to the
```

src/api/cvc4cpp.h
```diff
@@ -2689,6 +2690,20 @@ class CVC4_PUBLIC Solver
...
+// new API. !!!
+std::vector<Expr> termVectorToExprs(const std::vector<Term>& terms);
+std::vector<Type> sortVectorToTypes(const std::vector<Sort>& sorts);
+std::vector<Term> exprVectorToTerms(const std::vector<Expr>& terms);
+std::vector<Sort> typeVectorToSorts(const std::vector<Type>& sorts);
+std::set<Type> sortSetToTypes(const std::set<Sort>& sorts);
+
 }  // namespace api
+
+// !!! These only temporarily public until the parser is fully migrated to the
```
4tXJ7f:
```suggestion
// !!! Only temporarily public until the parser is fully migrated to the
```

PR #3764, opened 2020-02-17
Using ParseOp in TPTP

HanielB:
Makes the TPTP parser use ParseOp rather than a myriad of calls to the expression manager. This will help us in the parser migration.

This also adds a missing check in the SMT2 parsing relative to HOL.

### DIFF
--- a/src/parser/tptp/Tptp.g
+++ b/src/parser/tptp/Tptp.g
@@ -1061,10 +1061,8 @@ thfLogicFormula[CVC4::Expr& expr]
         }
       }
     | // N-ary and &
-      ( {
-  args.push_back(expr); }
-        ( AND_TOK thfUnitaryFormula[expr] {
-  args.push_back(expr); } )+
+      ( { args.push_back(expr); }
+        ( AND_TOK thfUnitaryFormula[expr] { args.push_back(expr); } )+
         {
           expr = MK_EXPR_ASSOCIATIVE(kind::AND, args);
         }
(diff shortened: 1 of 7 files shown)


src/parser/tptp/Tptp.g
```diff
@@ -892,30 +1046,25 @@ thfLogicFormula[CVC4::Expr& expr]
...
+          case tptp::NA_REVAND:
+            expr = MK_EXPR(kind::NOT, MK_EXPR(kind::AND, expr, expr2));
+            break;
         }
       }
     | // N-ary and &
-      ( { args.push_back(expr); }
-        ( AND_TOK thfUnitaryFormula[expr] { args.push_back(expr); } )+
+      ( {
+  args.push_back(expr); }
```
ajreynol: Fix the clang-formatting here?

src/parser/tptp/Tptp.g
```diff
@@ -883,7 +1034,10 @@ thfLogicFormula[CVC4::Expr& expr]
         {
           // TODO create whatever lambda
         }
-        expr = MK_EXPR(kind::EQUAL, expr, expr2);
+        args.push_back(expr);
+        args.push_back(expr2);
+        ParseOp p(kind::EQUAL);
+        expr = PARSER_STATE->applyParseOp(p, args);
```
ajreynol: Nit: this is a bit over-refactored, this line could have stayed the same.  (Same applies to a few other places).

HanielB: In `applyParseOp` there are checks for using HOL stuff. We'd have to move them here otherwise.

ajreynol: I see, sounds good.

Issue #2846, opened 2019-02-28
JVM crash in finalizer


peterzeller:
The finalizers of Expr and Type make the program crash by calling `SmtEngine.dlRef(em)`.

Version
---------

Version: built from repository with tag `1.6`

Example
-----------

(Updated: Changed `mkBoundVar` to `mkVar`)

```
package test;

import edu.nyu.acsys.CVC4.*;

class Cvc4Test {

    public static void main(String... args) throws InterruptedException {
        System.loadLibrary("cvc4jni");
        var em = new ExprManager();
        var smt = new SmtEngine(em);
        smt.setOption("produce-models", new SExpr(true));
        var x = em.mkVar("x", em.integerType());
        var y = em.mkVar("y", em.integerType());
        var z = em.mkVar("z", em.integerType());
        gc("a");
        var a1 = em.mkExpr(Kind.GT, x, em.mkExpr(Kind.PLUS, y, z));
        gc("b");
        smt.assertFormula(a1);
        gc("c");
        var a2 = em.mkExpr(Kind.LT, y ,z);
        gc("d");
        smt.assertFormula(a2);
        gc("e");
        System.out.println(a1);
        System.out.println(a2);
        gc("f");
        var res = smt.checkSat();
        gc("g");
        System.out.println("Res = " + res);
        System.out.println("x =  " + smt.getValue(x));
        System.out.println("y =  " + smt.getValue(y));
        System.out.println("z =  " + smt.getValue(z));
    }

    private static void gc(String msg) throws InterruptedException {
        System.out.println("calling gc " + msg);
        Thread.sleep(100);
        System.gc();
    }

}
```

This example calls `System.gc` to trigger garbage collection and thus calling of the finalizers. Since `System.gc` is just a hint to the runtime, the test won't fail every time.

When it fails the output is something like the following:

```
calling gc a
calling gc b
#
# A fatal error has been detected by the Java Runtime Environment:
#
#  SIGSEGV (0xb) at pc=0x00007fb4a926bfa0, pid=13580, tid=13602
#
# JRE version: OpenJDK Runtime Environment (11.0.1+13) (build 11.0.1+13-Ubuntu-3ubuntu3.18.10.1)
# Java VM: OpenJDK 64-Bit Server VM (11.0.1+13-Ubuntu-3ubuntu3.18.10.1, mixed mode, sharing, tiered, compressed oops, g1 gc, linux-amd64)
# Problematic frame:
[...]

PR #3767, opened 2020-02-17
Remove front-end support for Chain

ajreynol:
Towards parser migration.

Notice this PR makes it so that the new API is included by the parser (this and subsequent changes to the Parser will be in terms of api::Term not Expr). I have included a previous suggestion by @4tXJ7f to fix this, however it seems to be broken Travis CI.


4tXJ7f:
I will submit a PR to properly disable the parser components in our bindings as part of fixing #2846, which should take care of the issues with the bindings in this PR as well.

PR #3768, opened 2020-02-17
Split extended functions solver in strings

ajreynol:
Moves techniques for handling extended functions to their own class in theory of strings.

No behavior changes intended in this PR.


src/theory/strings/extf_solver.h
```diff
@@ -0,0 +1,198 @@
...
+ ** directory for licensing information.\endverbatim
+ **
+ ** \brief Solver for extended functions of theory of strings.
+ **/
+
+#include "cvc4_private.h"
+
+#ifndef CVC4__THEORY__STRINGS__EXTF_SOLVER_H
+#define CVC4__THEORY__STRINGS__EXTF_SOLVER_H
+
```
4tXJ7f: Include `<vector>` and `<map>`.

Issue #3770, opened 2020-02-18
API allows you to set a non-included SAT solver via bv-sat-solver


andrewvaughanj:
Using the "legacy" Python bindings (i.e., SWIG), it seems that it is possible to "enable" a SAT solver via `bv-sat-solver`, which was not compiled-in.

Example (for a `_CVC4.so` that does not have CMS compiled-in):

```
#!/usr/bin/env python

from __future__ import print_function, unicode_literals

import CVC4

if __name__ == "__main__":
    em = CVC4.ExprManager()
    smt = CVC4.SmtEngine(em)

    smt.setLogic(str("QF_AUFBV"))
    smt.setOption(str("produce-models"), CVC4.SExpr(str("true")))
    smt.setOption(str("incremental"), CVC4.SExpr(str("true")))
    smt.setOption(str("output-language"), CVC4.SExpr(str("smt2")))

    sat_solver = "CaDiCal"
    built_with_attr = getattr(
        CVC4, "Configuration_isBuiltWith{:s}".format(sat_solver.title())
    )
    assert built_with_attr()
    smt.setOption(str("bv-sat-solver"), CVC4.SExpr(str(sat_solver.lower())))
    enabled = smt.getOption(str("bv-sat-solver")).toString().split("::")[1]
    assert enabled.lower() == sat_solver.lower()

    #
    # Notice this assertion is negated!
    #
    sat_solver = "Cryptominisat"
    built_with_attr = getattr(
        CVC4, "Configuration_isBuiltWith{:s}".format(sat_solver.title())
    )
    assert not built_with_attr()
    smt.setOption(str("bv-sat-solver"), CVC4.SExpr(str(sat_solver.lower())))
    enabled = smt.getOption(str("bv-sat-solver")).toString().split("::")[1]
    assert enabled.lower() == sat_solver.lower()

# EOF
```

Is this an expected behaviour? It seems that there is some level of "options validation" (because if I try to set `bv-sat-solver` to be something stupid, CVC4 rejects it), but it would be nice if CVC4 could reject options that are not compiled-in.

I'd be happy to take at fixing this, if some pointers could be given. However, I think that this information comes from the `toml`, which probably isn't "build option" aware.

andrewvaughanj:
[...]

PR #3771, opened 2020-02-18
Adding validation to the  option, to check if non-compiled-in solvers are selected #3770

andrewvaughanj:
Resolves https://github.com/CVC4/CVC4/issues/3770


src/options/options_handler.cpp
```diff
@@ -147,6 +147,22 @@ void OptionsHandler::abcEnabledBuild(std::string option, std::string value)
 
 void OptionsHandler::checkBvSatSolver(std::string option, SatSolverMode m)
 {
+#if !defined(CVC4_USE_CRYPTOMINISAT)
+  if (m == SatSolverMode::CRYPTOMINISAT)
+  {
+    std::stringstream ss;
+    ss << "option `" << option << "' requires a Cryptominisat build of CVC4; this binary was not built with Cryptominisat support";
```
mpreiner:
```suggestion
    ss << "option `" << option << "' requires a CryptoMiniSat build of CVC4; this binary was not built with CryptoMiniSat support";
```

andrewvaughanj:
> IMHO, it would be better to use `Configuration::isBuiltWithCryptominisat()` and `Configuration::isBuiltWithCadical()` instead of the `#if`s.

All done!

PR #3772, opened 2020-02-18
resource manager: Add statistic for every resource.

### LINKED ISSUE #3751: resource limit and steps increments
I'm trying to tune cvc4 resource limits using the various *step switches on the command line:
```
  --bitblast-step=N      
  --bv-sat-conflict-step=N
  --cnf-step=N           
  --decision-step=N      
  --lemma-step=N         
  --parse-step=N         
  --preprocess-step=N    
  --quantifier-step=N    
  --restart-step=N       
  --rewrite-step=N       
  --sat-conflict-step=N  
  --theory-check-step=N  
```
I am assuming these switches influence how the internal counter is incremented for each of these submodules. This internal counter is output with the `--stats` switch as `smt::SmtEngine::resourceUnitsUsed`. (correct?)

To improve our tuning of the resource limits, I would like to know how many times each of these separate increments was hit (e.g. how many times the `rewrite_step` increment was done). Is this information available in the output of `--stats`? If not, could it be added? Otherwise it seems impossible to use these switches other than guessing and checking the outcome.

mpreiner:
This commit adds statistics for all resource steps. A resource statistic is incremented by 1 if the resource is spent (via `spendResource`).

Fixes #3751.


src/util/resource_manager.h
```diff
@@ -74,123 +77,165 @@ class CVC4_PUBLIC Timer {
...
+   IntStat d_numCnfStep;
+   IntStat d_numDecisionStep;
+   IntStat d_numLemmaStep;
+   IntStat d_numParseStep;
+   IntStat d_numPreprocessStep;
+   IntStat d_numQuantifierStep;
+   IntStat d_numRestartStep;
+   IntStat d_numRewriteStep;
+   IntStat d_numSatConflictStep;
+   IntStat d_numTheoryCheckStep;
```
4tXJ7f: Could arguably also be a map from `Resource` to `IntStat` (especially if the list grows). I am fine with this for now though.

src/prop/bvminisat/core/Solver.cc
```diff
@@ -1482,11 +1483,11 @@ void ClauseAllocator::reloc(CRef& cr,
...
 
 void Solver::setNotify(Notify* toNotify) { d_notify = toNotify; }
-bool Solver::withinBudget(uint64_t amount) const
+bool Solver::withinBudget(ResourceManager::Resource r) const
 {
   AlwaysAssert(d_notify);
-  d_notify->spendResource(amount);
-  d_notify->safePoint(0);
+  d_notify->spendResource(r);
+  d_notify->safePoint(r);
```
4tXJ7f: Shouldn't the two lines above be just `d_notify->safePoint(r);` since `safePoint()` calls `spendResource` and the original code did not spend the `amount` twice.

src/util/resource_manager.cpp
```diff
@@ -199,6 +203,72 @@ void ResourceManager::spendResource(unsigned amount)
...
+    case Resource::RewriteStep:
+      amount = d_options[options::rewriteStep];
+      ++d_statistics.d_numRewriteStep;
+      break;
+    case Resource::SatConflictStep:
+      amount = d_options[options::satConflictStep];
+      ++d_statistics.d_numSatConflictStep;
+      break;
+    case Resource::TheoryCheckStep:
+      amount = d_options[options::rewriteStep];
```
4tXJ7f:
```suggestion
      amount = d_options[options::theoryCheckStep];
```

src/util/resource_manager.h
mpreiner: I'm not sure if you gain much with a map, since you have to initialize each statistic with a name anyways in the Statistics constructor.

src/prop/bvminisat/core/Solver.cc
mpreiner: Yes you are right. `safePoint()` will call `spendResource` in https://github.com/CVC4/CVC4/blob/master/src/theory/theory_engine.h#L284 . I wonder why it was not `d_notify->safePoint(amount);` from the beginning.

PR #3773, opened 2020-02-18
Change datatype selector/constructor/tester to terms

makaimann:
Previously the `Expr` for the selector/constructor/tester was stored in the `Op`. This change moves it to a `Term`. Then to apply it, use the Kind/Op `APPLY_<SELECTOR/CONSTRUCTOR/TESTER>` followed by the corresponding Term and then the arguments.

The one hiccup we've run into is that the arity changes. For example, a selector should have a single argument but now it has two (the selector and the datatype to select from). We did not run into this disparity for uninterpreted functions, because the arity checker doesn't check the exact arity of the function (only the operator -- APPLY)

### DIFF
--- a/src/api/cvc4cpp.cpp
+++ b/src/api/cvc4cpp.cpp
@@ -627,6 +627,7 @@ uint32_t maxArity(Kind k)
   // special cases for API level
   // higher-order logic perspective at API
   // functions/constructors/selectors/testers are terms
+  // functions/constructors don't have a max so they don't need to be handled
   if (k == APPLY_SELECTOR)
   {
     max++;
@@ -1389,12 +1390,6 @@ Op Term::getOp() const
 }
 
 bool Term::isNull() const { return isNullHelper(); }
-
-bool Term::isParameterized() const
-{
-  CVC4_API_CHECK_NOT_NULL;
-  return d_expr->isParameterized();
-}
 
 Term Term::notTerm() const
 {
--- a/src/api/cvc4cpp.cpp
+++ b/src/api/cvc4cpp.cpp
@@ -589,20 +589,21 @@ const static std::unordered_set<Kind, KindHashFunction> s_indexed_kinds(
      FLOATINGPOINT_TO_FP_UNSIGNED_BITVECTOR,
      FLOATINGPOINT_TO_FP_GENERIC});
 
-/** Set of kinds where API term structure differs from internal structure
+namespace {
+
+bool isDefinedKind(Kind k) { return k > UNDEFINED_KIND && k < LAST_KIND; }
+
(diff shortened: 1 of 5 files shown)


src/api/cvc4cpp.h
```diff
@@ -1218,9 +1218,9 @@ class CVC4_PUBLIC DatatypeSelector
 
   /**
    * Get the selector operator of this datatype selector.
-   * @return the selector operator
+   * @return the selector Term
```
ajreynol: Term -> term

makaimann: Good catch, thanks! Fixed in 8777f854d5b024519116300a8ee0ac45411507ad.

makaimann:
All right, so we had some offline discussion and @ajreynol noticed that `getOp` was a bit strange, because it always took the `Expr` `operator` even when it was not an indexed operator. To fix this, we only build an `Op` with an `Expr` for parameterized terms. 

However, because the API takes a "higher-order" perspective, there are some special cases. Applications of uninterpreted functions and datatype constructors/selectors/testers are considered parameterized. But at the API level, we want the Op to just be `APPLY_*` and the first argument is the function/constructor/selector/tester. Thus there are special cases for this in `getOp` as well as term iteration. 

The one other place we need a special case is when checking the max arity. For functions and constructors, this is not a problem because they don't have a max. Selectors and testers are only supposed to have one argument, but at the API level they have two (the selector/tester itself, and the argument). This is special-cased here: https://github.com/makaimann/CVC4/blob/db9521b3cbffae0fb227e3f7f332603be3faa584/src/api/cvc4cpp.cpp#L631.

In the future, we might not check the arities until type checking.

src/api/cvc4cpp.cpp
```diff
@@ -589,6 +589,17 @@ const static std::unordered_set<Kind, KindHashFunction> s_indexed_kinds(
      FLOATINGPOINT_TO_FP_UNSIGNED_BITVECTOR,
      FLOATINGPOINT_TO_FP_GENERIC});
 
+/** Set of kinds where API term structure differs from internal structure
+ *  The API takes a "higher-order" perspective and treats functions as well
+ *  as datatype constructors/selectors/testers as terms
+ *  but interally they are not
+ */
+const static std::unordered_set<CVC4::Kind, CVC4::kind::KindHashFunction>
```
ajreynol: I'd prefer if this were a utility function not data.

makaimann: Cool, just made that change with ee9583187c5e420b3b20bd6a9a6bef0fd338bc01. I also made it so that all the special casing uses this new utility function. Should be cleaner that way.

PR #3778, opened 2020-02-19
Change Record to shared_ptr in Expr-level Datatype

ajreynol:
Fixes segfaults due to the memory corruption when it is used in the new API.


ajreynol:
Fixes https://github.com/CVC4/cvc4-projects/issues/112.

PR #3780, opened 2020-02-19
Use side effect utility for non-linear lemmas

### LINKED ISSUE #3647: --sygus-inference Fatal failure at CVC4/src/theory/arith/nonlinear_extension.cpp:3224
Hi,
For this formula,
```
(assert (distinct (sin 1.0) 0.0))
(check-sat)
```
`cvc4 --sygus-inference` throws out a fatal failure:
```
Fatal failure within bool CVC4::theory::arith::NonlinearExtension::checkTfTangentPlanesFun(CVC4::Node, unsigned int, std::vector<CVC4::NodeTemplate<true> >&) at /CVC4/src/theory/arith/nonlinear_extension.cpp:3224
Check failure
 std::find( d_secant_points[tf][d].begin(), d_secant_points[tf][d].end(), c) == d_secant_points[tf][d].end()
Aborted
```

OS: Ubuntu 18.04
Revision: fd88b18

ajreynol:
Fixes #3647.

Previously, when doing incremental linearization for transcedental functions, we would add points to the list of secant points at the time when linearization lemmas were generated. However, our strategy has been updated such that lemmas may be abandoned (say in the case that a higher priority lemma is found).  Thus, our list of secant points had spurious points corresponding to lemmas that weren't sent. This led to assertion failures, and likely led to gaps in our linearization, hindering our ability to say sat/unsat.

This PR introduces a "lemma side effect" class to ensure that modifications to the state of the nonlinear solver are in sync with the lemmas we send out.


ahmed-irfan:
@ajreynol there are conflicts

ajreynol:
Resolved

PR #3787, opened 2020-02-20
Remove IntReal tightening axioms from th_lira.plf

alex-ozdemir:
A closer inspection revealed that the theory of arithmetic's rewriter
prevented such terms from being the hypotheses of lemmas/conflicts in
the theory of arithmetic. Thus, we do not need axioms for them at this
time.

### DIFF
--- a/proofs/signatures/th_lira_test.plf
+++ b/proofs/signatures/th_lira_test.plf
@@ -249,7 +249,7 @@
   (@ f2 (>=_Real
           (+_Real (term_real_var x) (term_int_to_real (term_int_var y)))
           (a_real 0/1))
-  (@ f3 (not (>=_IntReal (*_Int (a_int 1) (term_int_var y)) (a_real 0/1)))
+  (@ f3 (not (>=_IntReal (*_Int (a_int 1) (term_int_var y)) (term_int_to_real (a_int 0))))
 
   ; Normalization
   ; proof of real predicates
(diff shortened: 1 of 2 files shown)


proofs/signatures/th_lira_test.plf
```diff
@@ -249,7 +249,7 @@
   (@ f2 (>=_Real
           (+_Real (term_real_var x) (term_int_to_real (term_int_var y)))
           (a_real 0/1))
-  (@ f3 (not (>=_IntReal (*_Int (a_int 1) (term_int_var y)) (a_real 0/1)))
```
alex-ozdemir: These changes align the example statements in this test with the statements that arithmetic will actually produce.
It will print the literal as an integer, and then convert it to a real higher up.

proofs/signatures/th_lira_test.plf
```diff
@@ -292,48 +292,3 @@
...
-      (bounded_aff_contra _ _
-       (bounded_aff_add _ _ _ _ _
-         (bounded_aff_mul_c _ _ _ 1/1 n1)
-         (bounded_aff_add _ _ _ _ _
-           (bounded_aff_mul_c _ _ _ 1/1 n2)
-           bounded_aff_ax_0_>=_0)))
-  )))
-  )))
-  ))
-)
```
alex-ozdemir: This test centers around tightening a bound of the form `[int] >= [rational]`. However, the rewriter does this tightening automatically, and the proof system just trusts the rewriter, so there is no need to do this ourselves.

Since we're removing the corresponding axiom, we must also remove the test.

alex-ozdemir:
You bet!

In PR #3787 I've added those assertions. In this PR I remove the dead axioms.

Since the axioms are in fact dead, the two PRs are actually orthogonal, and can be merged in any order without issue.

yoni206:
I am assuming you mean #3741 :) And yes, I now see it there. Great!

PR #3797, opened 2020-02-22
Utilities for words

ajreynol:
This adds a utility file for handling operations over constant "words" (this will eventually generalize string or sequence constants).  This is work towards https://github.com/CVC4/cvc4-projects/issues/23.

Also improves documentation in regexp.h.


src/theory/strings/word.h
```diff
@@ -0,0 +1,119 @@
...
+#define CVC4__THEORY__STRINGS__WORD_H
+
+#include <vector>
+
+#include "expr/node.h"
+
+namespace CVC4 {
+namespace theory {
+namespace strings {
+namespace word {
```
4tXJ7f: Why introduce a separate namespace?

src/theory/strings/word.h
```diff
@@ -0,0 +1,119 @@
...
+Node substr(TNode x, std::size_t i, std::size_t j);
+
+/** Return the prefix of x of size at most i */
+Node prefix(TNode x, std::size_t i);
+
+/** Return the suffix of x of size at most i */
+Node suffix(TNode x, std::size_t i);
+
+/**
+ * Checks if there is any overlap between string x and another string y. This
```
4tXJ7f:
```suggestion
 * Checks if there is any overlap between word x and another word y. This
```

src/theory/strings/word.h
```diff
@@ -0,0 +1,119 @@
...
+ * corresponds to checking whether one string contains the other and whether a
```
4tXJ7f:
```suggestion
 * corresponds to checking whether one word contains the other and whether a
```

src/theory/strings/word.h
```diff
@@ -0,0 +1,119 @@
...
+ * @param x The first string
+ * @param y The second string
+ * @return True if there is an overlap, false otherwise
+ */
+bool noOverlapWith(TNode x, TNode y);
+
+/** overlap
+ *
+ * if overlap returns m>0,
+ * then the maximal suffix of this string that is a prefix of y is of length m.
```
4tXJ7f:
```suggestion
 * then the maximal suffix of this word that is a prefix of y is of length m.
```

src/theory/strings/word.h
```diff
@@ -0,0 +1,119 @@
...
+ * x.overlap("ab") = 0
+ * x.overlap("d") = 0
+ * x.overlap("bcdefdef") = 5
+ */
+std::size_t overlap(TNode x, TNode y);
+
+/** reverse overlap
+ *
+ * if roverlap returns m>0,
+ * then the maximal prefix of this string that is a suffix of y is of length m.
```
4tXJ7f:
```suggestion
 * then the maximal prefix of this word that is a suffix of y is of length m.
```

src/theory/strings/word.h
```diff
@@ -0,0 +1,119 @@
...
+/** Return the prefix of x of size at most i */
+Node prefix(TNode x, std::size_t i);
+
+/** Return the suffix of x of size at most i */
+Node suffix(TNode x, std::size_t i);
+
+/**
+ * Checks if there is any overlap between string x and another string y. This
+ * corresponds to checking whether one string contains the other and whether a
+ * substring of one is a prefix of the other and vice-versa.
```
4tXJ7f:
```suggestion
 * substring/subsequence of one is a prefix of the other and vice-versa.
```

src/theory/strings/word.cpp
```diff
@@ -0,0 +1,229 @@
...
+
+Node mkEmptyWord(Kind k)
+{
+  NodeManager* nm = NodeManager::currentNM();
+  if (k == CONST_STRING)
+  {
+    std::vector<unsigned> vec;
+    return nm->mkConst(String(vec));
+  }
+  Assert(false);
```
4tXJ7f: I'd suggest that we turn those `Assert(false)` into:

```suggestion
  Unimplemented();
```

src/theory/strings/word.h
```diff
@@ -0,0 +1,119 @@
...
+/** Returns true if y is a prefix of x */
+bool hasPrefix(TNode x, TNode y);
+
+/** Returns true if y is a suffix of x */
+bool hasSuffix(TNode x, TNode y);
+
+/** Replace the first occurrence of y in x with t */
+Node replace(TNode x, TNode y, TNode t);
+
+/** Return the substring of x starting at index i */
```
4tXJ7f:
```suggestion
/** Return the substring/subsequence of x starting at index i */
```

src/theory/strings/word.h
```diff
@@ -0,0 +1,119 @@
...
+Node substr(TNode x, std::size_t i);
+
+/** Return the substring of x starting at index i with size at most j */
```
4tXJ7f:
```suggestion
/** Return the substring/subsequence of x starting at index i with size at most j */
```

src/theory/strings/word.h
ajreynol: Same reason as we have `util::`.  These functions would otherwise pollute the strings namespace significantly.  Also, I think it's nice to have this so that there is a consistent naming scheme when calling them.

ajreynol: An alternative would be to make them static members of a class?

ajreynol:
Added unit tests.

src/theory/strings/word.h
4tXJ7f: I think that I'd prefer that, yes. I like to keep the namespaces and directories in sync.

PR #3801, opened 2020-02-22
Embed mkAssociative utilities within the API.

ajreynol:
Towards parser/API migration.


4tXJ7f:
I thought that the decision was to move these helpers into the API and make them transparent for the user s.t. the number of supported arguments for each operator matches the SMT-LIB standard. Do I remember this wrong?

PR #3804, opened 2020-02-24
Fix bugs related to printing Sygus commands

abdoo8080:
With this commit, most Sygus problems should print correctly. The current printing functionality was tested on 158 Sygus regress files (0, 1, and 2) and 153 of them were printed in Sygus2 format and contained "(check-synth)". The printing functionality was tested again on the generated files and gave almost the same results. The details are listed below:

### Parsing Errors:
regress1/rr-verify/fp-arith.sy

    CVC4 Error:
    Parse Error: ./test/regress/regress1/rr-verify/fp-arith.sy:9.50: Expected CVC4 to be compiled with SymFPU support

regress1/rr-verify/fp-bool.sy

    CVC4 Error:
    Parse Error: ./test/regress/regress1/rr-verify/fp-bool.sy:9.50: Expected CVC4 to be compiled with SymFPU support

regress1/sygus/enum-test.sy

    CVC4 Error:
    Parse Error: ./test/regress/regress1/sygus/enum-test.sy:5.23: Unexpected token: '('.

### Fatal Failures:
regress1/sygus/issue3461.sy

    Fatal failure within virtual void CVC4::printer::SygusExprPrintCallback::toStreamSygus(const CVC4::Printer *, std::ostream &, CVC4::Expr) const at ../src/printer/sygus_print_callback.cpp:85
    Check failure

    e.getNumChildren() == d_args.size()

regress1/sygus/temp_input_to_synth_ic-error-121418.sy

    Fatal failure within void CVC4::FloatingPointLiteral::unfinished() const at ../src/util/floatingpoint.cpp:462
    Unimplemented code encounteredFloating-point literals not yet implemented.

regress2/sygus/lustre-real.sy

    CVC4 interrupted by user.

### Failed Tests:
1793 - regress1/sygus/issue3461.sy (Failed): likely because of parse error above

1966 - regress2/sygus/lustre-real.sy (Failed): likely because it was aborted in run 1

### Parse Errors:
regress2/sygus/lustre-real_autogen.sy: likely because it was aborted in run 1

    CVC4 Error:
    Parse Error: ./test/regress/regress2/sygus/lustre-real_autogen.sy:8.0: Unexpected token: '<<EOF>>'.

### Fatal Failures:
regress1/sygus/temp_input_to_synth_ic-error-121418_autogen.sy: same error as in run 1

    Fatal failure within void CVC4::FloatingPointLiteral::unfinished() const at ../src/util/floatingpoint.cpp:462
    Unimplemented code encounteredFloating-point literals not yet implemented.

### Failed Tests:
1793 - regress1/sygus/issue3461.sy (Failed): because of parse error in run 1

1808 - regress1/sygus/logiccell_help.sy (Failed)

    CVC4 Error:
    Parse Error: ./test/regress/regress1/sygus/logiccell_help.sy:15.23: Expected <invalid>, found ')'.

    ((Start (_ BitVec 3)) (Start_1_arg_0 (_ BitVec 1)) (Start_1_arg_1 (_ B...

1966 - regress2/sygus/lustre-real.sy (Failed): likely because it was aborted in run 1


ajreynol:
The SYMFPU failures are due to not configuring with `--symfpu`, do add this option to your configure line.  You will also have to install symfpu with `./contrib/get-symfpu` (https://github.com/CVC4/CVC4/blob/master/contrib/get-symfpu).

src/smt/smt_engine.cpp
```diff
@@ -1130,6 +1118,27 @@ void SmtEngine::setLogic(const std::string& s)
   SmtScope smts(this);
   try {
     setLogic(LogicInfo(s));
+
+    // dump out a set-logic command
+    if (Dump.isOn("benchmark"))
```
ajreynol: Can you comment on the change(s) in behavior of moving this code?

abdoo8080: I had a problem where declare-var command sometimes gets printed before the set-logic command (which is printed in SmtEngine::finishInit function). Because of that, CVC4, printed warnings about not setting the logic and some tests failed. I moved the code from finishInit to this function to ensure that set-logic command is printed as soon as it's encountered.

ajreynol: Right, (I think) I agree with this change however it does effect the default behavior for this and other use cases.

For example, another consequence is that we don't print any `set-logic` when dumping `raw-benchmark` for benchmarks without a set-logic command, whereas previously the code always force-printed a `(set-logic ALL)`.

ajreynol: After discussion on devs, this block of code should be broken into two: 
- The case where `raw-benchmark` is true, should be moved here, as you did,
- The case where `raw-benchmark` is false should remain where it was (move it back to its original place).

ajreynol:
LGTM

PR #3809, opened 2020-02-24
Convert parser input interface to api::Term

ajreynol:
In preparation for parser migration.

This PR updates the Input classes to return api::Term instead of Expr.

It patches the nra_examples but does not address https://github.com/CVC4/cvc4-projects/issues/108.


src/parser/parser.cpp
```diff
@@ -661,14 +661,14 @@ Command* Parser::nextCommand()
...
-Expr Parser::nextExpression()
+api::Term Parser::nextExpression()
 {
   Debug("parser") << "nextExpression()" << std::endl;
   d_resourceManager->spendResource(ResourceManager::Resource::ParseStep);
   Expr result;
   if (!done()) {
     try {
-      result = d_input->parseExpr();
+      result = d_input->parseExpr().getExpr();
```
4tXJ7f: Why not just make the `result` a `Term`?

ajreynol: It is on my branch, this change touches fewer lines (will have fewer merge conflicts this way).

ajreynol: Actually i guess its even 2 vs 2 haha. Anyways that was the reason.

ajreynol: Actually nevermind, this way has fewer merge conflicts :)  Anyways that was supposely the reason.

src/parser/parser.cpp
```diff
@@ -665,10 +665,10 @@ api::Term Parser::nextExpression()
 {
   Debug("parser") << "nextExpression()" << std::endl;
   d_resourceManager->spendResource(ResourceManager::Resource::ParseStep);
-  Expr result;
+  api::Term result;
   if (!done()) {
     try {
-      result = d_input->parseExpr().getExpr();
+      result = d_input->parseExpr();
```
4tXJ7f: The `return api::Term(result);` statement below can be replaced by just `return result;`, I think.

PR #3812, opened 2020-02-25
More fixes for printing sygus commands

ajreynol:
Towards v1 -> v2 sygus conversion.

This makes several fixes and improvements related to printing sygus commands:
(1) The logic is extended with LIA, DT, UF internally during setDefaults instead of during parsing.  This is the correct place for this extension of the logic since it should be applied regardless of the parser.  5 existing logic bugs were discovered as a result of this in regressions, which are fixed by this PR.
(2) Ensures that terms in sygus grammars are printed without letification, since this is prohibited in sygus. Notice the formulas printed by constraints need to be letified (otherwise we can't convert large lustre benchmarks). Thus, the letification threshold should determine this but always be overridden for grammar terms.
(3) Ensures final options are set for all sygus-specific commands, which follows the standards prescribed by sygus v2 (all set-* commands come before other commands).


ajreynol:
FYI @abdoo8080

src/printer/sygus_print_callback.cpp
```diff
@@ -78,8 +78,9 @@ void SygusExprPrintCallback::toStreamSygus(const Printer* p,
     sbody =
         sbody.substitute(vars.begin(), vars.end(), subs.begin(), subs.end());
 
+    // print to stream without letification
     std::stringstream body_out;
-    body_out << sbody;
+    p->toStream(body_out, sbody, -1, false, 0);
```
HanielB: I think this would be better handled if, by default, when doing SyGuS, the letification threshold would be 0 (by assigning the option to 0), but if the user were to set it to be bigger than we'd use that. Just because the default should be without lets doesn't mean a user can't want to print with them.

ajreynol: We tried this actually.  The issue is you want the formulas printed by constraints to be letified (otherwise we can't convert large lustre benchmarks).  Thus, the letification threshold should determine this but always be overridden for grammar terms.

HanielB: I see. Please add this info to the PR description.

Issue #677, opened 2017-08-19
Syntactic equality guard (Bugzilla #458)


mpreiner:
**Imported from Bugzilla**
Reporter: *Wei Wang*
Status: *RESOLVED*
Severity: *enhancement*
Assigned to: *Morgan Deters*
Component: *RewriteRules*
Milestone: *1.4*
Version: *master*
Platform: *All*
OS: *Mac OS*

On 2012-11-19 15:26:47 -0800, Wei Wang wrote:
> As an example, for one of the inference rules of reachability theory
> 
> R(x1, x2), R(x2, x3)
> \-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-
>     R(x1, x3)
> 
> 
> In instantiation, we don't the candidate of x1 is same as that of x2, and same requirement for x2 and x3, which would introduce many redundant lemmas. Thus, a syntactic equality guard like !(x1 \=\= x2 OR x2 \=\= x3) would be helpful, and "\=\=" is syntactic equal.

On 2014-06-21 03:19:22 -0700, Morgan Deters wrote:
> Rewrite rules theory decommissioned in favor of quantifiers.

PR #3816, opened 2020-02-25
Sets & Relations Java example

mudathirmahgoub:
This PR is related to [(Bugzilla #677)](https://github.com/CVC4/cvc4-projects/issues/25)


examples/api/java/Relations.java
```diff
@@ -0,0 +1,215 @@
...
+ ** in the top-level source directory) and their institutional affiliations.
+ ** All rights reserved.  See the file COPYING in the top-level source
+ ** directory for licensing information.\endverbatim
+ **
+ ** \brief Reasoning about relations with CVC4 via Java API.
+ **
+ ** A simple demonstration of reasoning about strings with CVC4 via Jave API.
+ **/
+
+import edu.nyu.acsys.CVC4.*;
```
ajreynol: I think this should say "stanford" not "nyu".

examples/api/java/Relations.java
```diff
@@ -0,0 +1,215 @@
...
+
+/*
```
ajreynol: Can you clarify why this is being written here?  Like "This file uses the API to make a sat call equivalent to the following benchmark:".

examples/api/java/Relations.java
mudathirmahgoub: I remembered it was mentioned in the meeting, but my version was a bit behind master. 
It is fixed to `edu.standfaord.CVC4.*` What about the other name called  `uiowa` :)

examples/api/java/Relations.java
mudathirmahgoub: > I believe you need to add the example to the tests makefile here: https://github.com/CVC4/CVC4/blob/master/examples/api/java/CMakeLists.txt
> 
> Then, verify that it passes Travis.

I missed this because I searched for `Strings.java` and not `Strings`. Fixed

ajreynol:
LGTM, as a note this fixes https://github.com/CVC4/cvc4-projects/issues/25.

ajreynol:
LGTM, as a note this fixes https://github.com/CVC4/cvc4-projects/issues/25.

PR #3818, opened 2020-02-25
Add a new arith constraint proof rule: IntTightenAP

alex-ozdemir:
This rule is used when a bound on an integer expression is tightened
because of integer reasoning.

Before this rule was subsumed by IntHoleAP, a catch-all rule for integer
reasoning. We are now articulating IntTightenAP separately, because we
can produce proofs for it.

For IntHoleAP, we will have to omit a hole.


src/theory/arith/constraint.cpp
```diff
@@ -1247,6 +1252,19 @@ bool Constraint::allHaveProof(const ConstraintCPVec& b){
   return true;
 }
 
+void Constraint::impliedByIntTighten(ConstraintCP a, bool nowInConflict){
```
yoni206: I suggest including debug messages similar to `impliedByIntHole `.

yoni206: (otherwise LGTM).

alex-ozdemir: Sounds good to me. I've done so in d0bbc29fa223f5cea0943b8727573ea1799b285c.

PR #3819, opened 2020-02-25
Some initial work on using words

ajreynol:
Towards support for sequences (CVC4/cvc4-projects#23.)


src/theory/strings/word.h
```diff
@@ -29,50 +30,53 @@ namespace strings {
 class Word
 {
 public:
-  /** make empty constant of kind k */
-  static Node mkEmptyWord(Kind k);
+ /** make empty constant of type tn */
```
4tXJ7f: It looks like the formatting somehow got messed up (`public` is not indented at all and the `static` method only by one space. Can you ClangFormat the whole file?

src/theory/strings/theory_strings_rewriter.cpp
```diff
@@ -2042,13 +2052,11 @@ Node TheoryStringsRewriter::rewriteContains( Node node ) {
...
     {
-      CVC4::String t = node[1].getConst<String>();
-      Node ret =
-          NodeManager::currentNM()->mkConst(s.find(t) != std::string::npos);
+      Node ret = nm->mkConst(Word::find(node[0], node[1]) != std::string::npos);
       return returnRewrite(node, ret, "ctn-const");
     }else{
       Node t = node[1];
-      if (s.size() == 0)
+      if (Word::getLength(node[0]) == 0)
```
```suggestion
      if (Word::isEmpty(node[0]))
```

src/theory/strings/theory_strings_rewriter.cpp
```diff
@@ -970,8 +980,7 @@ Node TheoryStringsRewriter::rewriteStarRegExp(TNode node)
     return returnRewrite(node, node[0], "re-star-nested-star");
   }
   else if (node[0].getKind() == STRING_TO_REGEXP
-           && node[0][0].getKind() == CONST_STRING
-           && node[0][0].getConst<String>().isEmptyString())
+           && node[0][0].getKind() == CONST_STRING && Word::isEmpty(node[0][0]))
```
4tXJ7f: We should have a method `Word::isWord(Node n)` that returns true for string/sequence constants and one for string/sequence regexps.

ajreynol: Actually, we already do: `.isConst()` :)  That is the preferred way of checking this, indeed I should fix the kind comparison here.

ajreynol: Btw, we will eventually need utility function for `isStringLike( TypeNode tn )` or something like that.

PR #3820, opened 2020-02-25
Move fix for vacuous sygus types out of the parser

ajreynol:
This moves a fix for vacuous sygus types out of the parser and into the Expr-level datatype.

This is a temporary fix so that further progress can be made on parser migration (and to declutter the parser).

This will be refactored when an API for SyGuS is established (https://github.com/CVC4/cvc4-projects/issues/38).


ajreynol:
FYI @abdoo8080

PR #3822, opened 2020-02-26
quantifier macros

kovach:
wip: TODOs in `src/preprocessing/passes/quantifier_macro_elimination.cpp`
@ajreynol

### DIFF
--- a/src/preprocessing/passes/quantifier_macro_elimination.cpp
+++ b/src/preprocessing/passes/quantifier_macro_elimination.cpp
@@ -1,5 +1,5 @@
 /*********************                                                        */
-/*! \file foo.cpp
+/*! \file quantifier_macro_elimination.cpp
  ** \verbatim
  ** Top contributors (to current version):
  **   Scott Kovach, Andrew Reynolds
@@ -11,7 +11,7 @@
  **
  ** \brief TODO add summary
  **
- ** TODO question: should I rename this pass?
+ ** TODO
  **    main TODOs:
  **    - don't recurse in rewriting function
  **    - fix unsat-core issue
--- a/src/preprocessing/passes/quantifier_macro_elimination.cpp
+++ b/src/preprocessing/passes/quantifier_macro_elimination.cpp
@@ -45,9 +45,6 @@ bool getApplicationVariableArgumentsIfUnique(Node application,
                                              vector<Node>& arguments)
 {
   std::set<Node> seenVars;
-  // TODO question: is this necessary because we're not sure whether type
-  // checking has occurred at this stage?
-  TypeNode operatorType = application.getOperator().getType();
   for (size_t i = 0; i < application.getNumChildren(); ++i)
   {
     Node arg = application[i];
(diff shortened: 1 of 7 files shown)


src/preprocessing/passes/quantifier_macro_elimination.cpp
```diff
@@ -0,0 +1,249 @@
...
+  Trace("QME") << "definitions:\n\n" << d_macroDefinitions << endl << endl;
+  Trace("QME") << "definition bindings:\n" << d_macroBoundVariables << endl;
+  // Apply macro substitutions:
+  if (d_macroDefinitions.size() == 0)
+  {
+    return PreprocessingPassResult::NO_CONFLICT;
+  }
+  for (Node& assertion : assertions)
+  {
+    // TODO currently 10 quantifier regression tests fail when check-unsat-cores
```
kovach: I think Andres pointed me to this being the point where something should be done to resolve this issue but I haven't looked into it yet

src/preprocessing/passes/quantifier_macro_elimination.cpp
```diff
@@ -0,0 +1,249 @@
...
+
+QuantifierMacroElimination::~QuantifierMacroElimination() {}
+
+// Check that arguments of application are unique bound variables and return
+// them in arguments. Assumes application is an APPLY_UF node.
+bool getApplicationVariableArgumentsIfUnique(Node application,
+                                             vector<Node>& arguments)
+{
+  std::set<Node> seenVars;
+  // TODO question: is this necessary because we're not sure whether type
```
ajreynol: Don't need to check welltypedness here.

src/preprocessing/passes/quantifier_macro_elimination.cpp
```diff
@@ -0,0 +1,249 @@
...
+  // traverse definition and check for occurrence of predicate
+  vector<Node> stack;
+  stack.push_back(definition);
+  while (stack.size() > 0)
+  {
+    Node current = stack.back();
+    stack.pop_back();
+    if (current.getKind() == APPLY_UF)
+    {
+      auto currentOperator = current.getOperator();
```
ajreynol: I generally don't like using `auto` anywhere since it is less descriptive.

src/preprocessing/passes/quantifier_macro_elimination.h
```diff
@@ -0,0 +1,57 @@
...
+#include "preprocessing/preprocessing_pass_context.h"
+#include "theory/logic_info.h"
+#include "theory/substitutions.h"
+
+namespace CVC4 {
+namespace preprocessing {
+namespace passes {
+using namespace std;
+
+// TODO: are these already defined somewhere else?
```
ajreynol: No.

I generally don't like defining these since it makes the code harder to follow in my opinion (I tend to be paranoid and don't want to have to follow definitions of whether the helper function is defined as expected).

src/preprocessing/passes/quantifier_macro_elimination.cpp
ajreynol: Yes, you can look at the old version of the macros preprocessing pass to see how to do this:  https://github.com/CVC4/CVC4/blob/master/src/preprocessing/passes/quantifier_macros.cpp#L104

kovach:
unsat core issue is fixed, but bug822 is still timing out

aniemetz:
Closed until this is picked up again.

PR #3825, opened 2020-02-26
Refactor type ascriptions in the parser

ajreynol:
Towards parser migration.

This consolidates two blocks of code (cvc/smt2) that do type ascriptions to a utility function in the parser.  It updates this function to use the new API.

This code will be further refactored when the interface for parametric datatype constructors is further developed in the new API.


src/parser/parser.cpp
```diff
@@ -515,6 +515,71 @@ Expr Parser::mkHoApply(Expr expr, std::vector<Expr>& args)
...
+  {
+    t = d_solver->mkUniverseSet(s);
+  }
+  else if (k == api::SEP_NIL)
+  {
+    t = d_solver->mkSepNil(s);
+  }
+  else if (k == api::APPLY_CONSTRUCTOR)
+  {
+    std::vector<api::Term> children;
```
mpreiner:
```suggestion
    std::vector<api::Term> children(t.begin(), t.end());
```

src/parser/cvc/Cvc.g
```diff
@@ -1836,23 +1836,8 @@ postfixTerm[CVC4::Expr& f]
...
-          f = MK_CONST(CVC4::EmptySet(t));
-        } else if(f.getKind() == CVC4::kind::UNIVERSE_SET && t.isSet()) {
-          f = EXPR_MANAGER->mkNullaryOperator(t, kind::UNIVERSE_SET);
-        } else {
-          if(f.getType() != t) {
-            PARSER_STATE->parseError("Type ascription not satisfied.");
-          }
-        }
+      {
+        f= PARSER_STATE->applyTypeAscription(f,t).getExpr();
```
4tXJ7f:
```suggestion
        f = PARSER_STATE->applyTypeAscription(f,t).getExpr();
```

PR #3826, opened 2020-02-26
Fix -Wshadow warnings in common headers

4tXJ7f:


ajreynol:
Adding @mpreiner since he had some wshadow changes too.  Otherwise LGTM.

PR #3828, opened 2020-02-26
Add support for is_digit and regular expression difference

ajreynol:
Towards support for the strings standard.  This adds support for str.is_digit and re.diff, which both can be eliminated eager during preprocessing.


src/api/cvc4cpp.cpp
```diff
@@ -263,13 +263,15 @@ const static std::unordered_map<Kind, CVC4::Kind, KindHashFunction> s_kinds{
     {STRING_LEQ, CVC4::Kind::STRING_LEQ},
     {STRING_PREFIX, CVC4::Kind::STRING_PREFIX},
     {STRING_SUFFIX, CVC4::Kind::STRING_SUFFIX},
+    {STRING_ISDIGIT, CVC4::Kind::STRING_ISDIGIT},
```
4tXJ7f: Wouldn't it be more consistent to call this kind `STRING_IS_DIGIT` (to match `STRING_TO_REGEXP`/`STRNG_IN_REGEXP`?

ajreynol: Sure, ill update

ajreynol: We should really consider changing STRING_STRREPL -> STRING_REPLACE too then :)

4tXJ7f: I agree, we should make everything consistent now that the standard exists (also names like `STRING_CONTAINS`, `STRING_FROM_INT`, `STRING_TO_INT`).

ajreynol: Let's do this on a separate PR.

PR #3829, opened 2020-02-26
Add support for str.from_code

4tXJ7f:
This commit adds support for `str.from_code`. This is work towards
supporting the new strings standard. The code currently just does an
eager expansion of the operator. The commit also renames `STRING_CODE`
to `STRING_TO_CODE` to better reflect the names of the operators in the
new standard.


src/theory/strings/theory_strings.cpp
```diff
@@ -595,6 +595,25 @@ void TheoryStrings::preRegisterTerm(TNode n) {
 
 Node TheoryStrings::expandDefinition(LogicRequest &logicRequest, Node node) {
   Trace("strings-exp-def") << "TheoryStrings::expandDefinition : " << node << std::endl;
+
+  if (node.getKind() == STRING_FROM_CODE)
+  {
+    // str.from_code(t) ---> choice k. ite(0 <= t < |A|, t = str.to_code(k), "")
```
ajreynol: the else branch should be `k = ""`.

4tXJ7f: Good catch, fixed.

PR #3832, opened 2020-02-26
Fixes for bounds on transcendental functions

### LINKED ISSUE #3783: Fatal failure at CVC4/src/theory/arith/nl_model.cpp:1043
Hi,
For this formula:
```
(declare-fun a () Real)
(assert (< (cos a) 1.0))
(check-sat)
```
```
Fatal failure within bool CVC4::theory::arith::NlModel::simpleCheckModelMsum(const std::map<CVC4::NodeTemplate<true>, CVC4::NodeTemplate<true> >&, bool) at /CVC4/src/theory/arith/nl_model.cpp:1043
Check failure
 false
Aborted
```

OS: Ubuntu 18.04
Commit: 508ecb3

ajreynol:
This PR refactors and fixes how bounds are set for transcendental functions.  The new code ensures that all transcendental function applications are given bounds.  (Our previous failures to do so were hindering our ability to say "sat", due to NlModel::checkModel failures).

There were previously two issues on why transcendental function applications were not being assigned bounds:
- "Slave" transcendental functions (e.g. those that we reduce via `sin(t) = sin(y) ^ -pi <= y <= pi ^ y + 2*pi*N = t`) were not being given bounds explicitly,
- Transcendental functions that are congruent to others (e.g. f(x) where f(y) exists and x=y in the current context) were being ignored and hence not bound.

This PR clarifies the master/slave relationship that tracks which transcendental function applications have been purified, and furthermore tracks congruence classes.

The setting of bounds and the check-model is further simplified by setting bounds on the original *terms*, whereas the current code sets bounds on the model values of terms.  In other words, previously if we had term `sin(y)` and `y^M = c`, then we'd set bounds for sin(c), whereas the new code sets the bound on `sin(y)` directly.

Fixes #3783.  We answer unknown without an assertion failure on that benchmark now.  Further work based on ignoring literals from internally generated lemmas is necessary for solving it sat.


src/theory/arith/nonlinear_extension.h
```diff
@@ -511,13 +511,21 @@ class NonlinearExtension {
   
   //transcendental functions
   /**
-   * Maps arguments of SINE applications to a fresh skolem. This is used for
-   * ensuring that the argument of SINE we process are on the interval
-   * [-pi .. pi].
+   * Some transcendental functions f(t) are "purified", e.g. we add
+   * t = y ^ f(t) = f(y) where y is a fresh varaible. Those that are not
```
ahmed-irfan: typo: variable

PR #3834, opened 2020-02-27
Convert more uses of string to word

ajreynol:
Towards theory of sequences.


src/theory/strings/theory_strings_rewriter.cpp
```diff
@@ -3436,21 +3455,22 @@ Node TheoryStringsRewriter::rewriteStringCode(Node n)
 
 Node TheoryStringsRewriter::splitConstant( Node a, Node b, int& index, bool isRev ) {
   Assert(a.isConst() && b.isConst());
-  index = a.getConst<String>().size() <= b.getConst<String>().size() ? 1 : 0;
-  unsigned len_short = index==1 ? a.getConst<String>().size() : b.getConst<String>().size();
+  size_t lenA = Word::getLength(a);
+  size_t lenB = Word::getLength(b);
+  index = lenA <= lenB ? 1 : 0;
+  unsigned len_short = index == 1 ? lenA : lenB;
```
```suggestion
  size_t len_short = index == 1 ? lenA : lenB;
```

Might as well make that change.

PR #3841, opened 2020-02-28
Replace conditional rewrite pass in quantifiers with the extended rewriter

### LINKED ISSUE #3839: sygus-inference incorrect answer on QF_NIA formula
Hi,
For this formula:
```
(declare-fun a () Int)
(declare-fun b () Int)
(assert (xor (> a 0) (not (and (ite (= a b) (> (* 4 a b) 1) true) (> (* a a) 0)))))
(assert (= a b))
(assert (> (* a b) 0))
(check-sat)
```
CVC4 --sygus-inference incorrectly gives unsat, while this formula should be sat.

OS: Ubuntu 18.04
Commit: 5c844f0

ajreynol:
Fixes #3839.

Previously, the quantifiers rewriter had a rewriting step that was an ad-hoc version of some of the rewrites that have been incorporated into the extended rewriter.  Moreover, the code for that pass was buggy.

This eliminates the previous conditional rewriting step from the "term process" rewrite pass in quantifiers.  It additional adds an optional (disabled by default) rewriting pass that calls the extended rewriter on the body of quantified formulas.  This subsumes the previous behavior and should not be buggy.

Notice that the indentation in computeProcessTerms changed and subsequently has been updated to the new coding standards.

This PR relies on https://github.com/CVC4/CVC4/pull/3840.

### DIFF
--- a/src/theory/quantifiers/quantifiers_rewriter.h
+++ b/src/theory/quantifiers/quantifiers_rewriter.h
@@ -151,7 +151,6 @@ class QuantifiersRewriter : public TheoryRewriter
                              Node ipl);
   static Node computeProcessTerms2(Node body,
                                    std::map<Node, Node>& cache,
-                                   std::map<Node, Node>& icache,
                                    std::vector<Node>& new_vars,
                                    std::vector<Node>& new_conds,
                                    bool elimExtArith);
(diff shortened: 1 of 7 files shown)


src/theory/quantifiers/quantifiers_rewriter.h
```diff
@@ -201,12 +198,42 @@ class QuantifiersRewriter : public TheoryRewriter
                                const std::vector<Node>& args,
                                QAttributes& qa);
   //-------------------------------------end conditional splitting
+  //------------------------------------- process terms
+  /** compute process terms
+   *
+   * This takes as input a quantified formula q with attributes qa whose
+   * body is body.
+   *
+   * This rewrite steps eliminates problematic terms from the bodies of
```
HanielB: Remove "steps"?

src/theory/quantifiers/quantifiers_rewriter.h
```diff
@@ -201,12 +198,42 @@ class QuantifiersRewriter : public TheoryRewriter
...
+  static Node computeProcessTerms(Node body,
+                                  std::vector<Node>& new_vars,
+                                  std::vector<Node>& new_conds,
+                                  Node q,
+                                  QAttributes& qa);
+  //------------------------------------- end process terms
+  //------------------------------------- extended rewrite
+  /** compute extended rewrite
+   *
+   * This returns the result of apply the extended rewriter on the body
```
HanielB: apply -> applying

ajreynol:
On SMT-LIB quantified logics UF, `--ext-rewrite-quant` is +134-198 over 23k benchmarks, so best not to enable by default.

PR #3845, opened 2020-02-28
Postorder traversal iter

alex-ozdemir:
Implement an iterator for postorder traversals.

I believe that this will be useful in pre-processing passes, many of
which do postorder traversals that they implement by hand.

Right now this iterator does not support modification of the traversal
pattern, but we could add this later on, if we want it.

My plan for now is to see how much I can simplify preprocessing passes with just this.

This iterator aims to be STL compliant, but I'm not an expert. I would *love* feedback along these lines.


alex-ozdemir:
I wasn't sure who to assign for review, so I put both of you on it. You guys can decide who, or re-assign.

src/smt_util/node_traversal.h
```diff
@@ -0,0 +1,102 @@
...
+  bool operator==(const NodePostorderIterator&) const;
+  // Equals
+  bool operator!=(const NodePostorderIterator&) const;
+
+ private:
+  // The next thing on the stack is the next thing to visit.
+  // If it is associated with `false`, then its children haven't been visited
+  // yet, and we visit them first. If it is associated with `true`, then we
+  // visit it. Invariant: the top (back) node is always associated with `true`.
+  std::vector<std::pair<bool, TNode>> d_stack;
```
alex-ozdemir: One downside of this invariant is that the `begin` member function is actually quite expensive.

I could eliminate this invariant, pushing that work fully into the increment operation.

src/smt_util/node_traversal.h
```diff
@@ -0,0 +1,102 @@
...
+
+  // Preincrement
+  NodePostorderIterator& operator++();
+  // Postincrement
+  NodePostorderIterator operator++(int);
+  // Dereference
+  reference operator*();
+  // Equals
+  bool operator==(const NodePostorderIterator&) const;
+  // Equals
```
mpreiner:
```suggestion
  // Not Equal
```

src/smt_util/node_traversal.h
```diff
@@ -0,0 +1,102 @@
...
+  bool operator!=(const NodePostorderIterator&) const;
+
+ private:
+  // The next thing on the stack is the next thing to visit.
```
mpreiner: What's the next thing on the stack? Do you mean back()?

src/smt_util/node_traversal.cpp
```diff
@@ -0,0 +1,72 @@
+#include "node_traversal.h"
+
+namespace CVC4 {
+
+NodePostorderIterator::NodePostorderIterator(TNode n) : d_stack{{false, n}}
+{
+  enqueueChildrenUntilLeaf();
```
mpreiner: I would not do that on construction.

src/smt_util/node_traversal.cpp
```diff
@@ -0,0 +1,72 @@
...
+      d_stack.back().first = true;
+      TNode back = d_stack.back().second;
+      // use integer underflow wrapping behavior to reverse-iterate over the
+      // children,
+      for (size_t i = back.getNumChildren() - 1; i < back.getNumChildren(); --i)
+      {
+        // and enqueue them.
+        d_stack.emplace_back(false, back[i]);
+      }
+    }
```
mpreiner: This looks like you are doing a tree traversal here, since you don't prevent nodes from being pushed multiple times onto `d_stack`. You need an additional cache for this that keeps track of visited nodes (which should make the bool flag on the stack obsolete). You usually don't want to do tree traversals on DAGs.

test/unit/smt_util/node_traversal_black.h
```diff
@@ -0,0 +1,150 @@
...
+    Node eb = d_nodeManager->mkConst(false);
+    Node cnd = d_nodeManager->mkNode(XOR, tb, eb);
+    Node top = d_nodeManager->mkNode(XOR, cnd, cnd);
+
+    auto traversal = NodePostorderIterable(top);
+
+    size_t count = std::count_if(traversal.begin(),
+                                 traversal.end(),
+                                 [](TNode n) { return n.isConst(); });
+    TS_ASSERT(count == 4);
```
mpreiner: See comment about tree traversals.

test/unit/smt_util/node_traversal_black.h
```diff
@@ -0,0 +1,150 @@
...
+    Node cnd = d_nodeManager->mkNode(XOR, tb, eb);
+    Node top = d_nodeManager->mkNode(XOR, cnd, cnd);
+    std::vector<TNode> expected = {tb, eb, cnd, tb, eb, cnd, top};
+
+    auto traversal = NodePostorderIterable(top);
+
+    std::vector<TNode> actual;
+    std::copy(traversal.begin(), traversal.end(), std::back_inserter(actual));
+    TS_ASSERT(actual == expected);
+  }
```
mpreiner: Ditto.

src/smt_util/node_traversal.h
alex-ozdemir: Changed in 5b4bcc498

src/smt_util/node_traversal.h
alex-ozdemir: Done in 5b4bcc498

src/smt_util/node_traversal.h
alex-ozdemir: Exactly. Clarified in 5b4bcc498

src/smt_util/node_traversal.cpp
alex-ozdemir: Sounds good. Changed in 5b4bcc498

src/smt_util/node_traversal.cpp
alex-ozdemir: Thanks! You're completely right. I forgot that nodes are DAGs. Fixed in 5b4bcc498.

I believe that just a visit set is insufficient, and I used a map instead, but I'd love to find a way to do without the map.

test/unit/smt_util/node_traversal_black.h
alex-ozdemir: Fixed in 5b4bcc498

test/unit/smt_util/node_traversal_black.h
alex-ozdemir: Fixed in 5b4bcc498

src/smt_util/node_traversal.h
```diff
@@ -0,0 +1,117 @@
...
+#include <iterator>
+#include <vector>
+
+#include "expr/node.h"
+
+namespace CVC4 {
+
+// Iterator for traversing a node in post-order
+// It does DAG-traversal, so indentical sub-nodes will be visited once only.
+class NodePostorderIterator
```
mpreiner: How about making the iterator more general. E.g., a `NodeDfsIterator` that can do both pre-order and post-order traversal. The constructor can take a bool flag that indicates whether you want pre or post-order traversals.

alex-ozdemir: A good idea, and what I plan to do.

I thought that'd I'd start with something simple (just post-order), look for use-sites, discover which generalizations would be useful (e.g. pre-order traversal), expand the interface, and then rinse-and-repeat.

test/unit/smt_util/node_traversal_black.h
```diff
@@ -0,0 +1,150 @@
...
+    Node eb = d_nodeManager->mkConst(false);
+    Node cnd = d_nodeManager->mkNode(XOR, tb, eb);
+    Node top = d_nodeManager->mkNode(XOR, cnd, cnd);
+    std::vector<TNode> expected = {tb, eb, cnd, top};
+
+    auto traversal = NodePostorderIterable(top);
+
+    std::vector<TNode> actual;
+    std::copy(traversal.begin(), traversal.end(), std::back_inserter(actual));
+    TS_ASSERT(actual == expected);
```
4tXJ7f:
```suggestion
    TS_ASSERT_EQUALS(actual, expected);
```

test/unit/smt_util/node_traversal_black.h
```diff
@@ -0,0 +1,150 @@
...
+    Node eb = d_nodeManager->mkConst(false);
+    Node cnd = d_nodeManager->mkNode(XOR, tb, eb);
+    Node top = d_nodeManager->mkNode(XOR, cnd, cnd);
+
+    auto traversal = NodePostorderIterable(top);
+
+    size_t count = std::count_if(traversal.begin(),
+                                 traversal.end(),
+                                 [](TNode n) { return n.isConst(); });
+    TS_ASSERT(count == 2);
```
4tXJ7f:
```suggestion
    TS_ASSERT_EQUALS(count, 2);
```

test/unit/smt_util/node_traversal_black.h
```diff
@@ -0,0 +1,150 @@
...
+
+    size_t count = 0;
+    for (auto i : NodePostorderIterable(cnd))
+    {
+      if (i.isConst())
+      {
+        ++count;
+      }
+    }
+    TS_ASSERT(count == 2);
```
4tXJ7f:
```suggestion
    TS_ASSERT_EQUALS(count, 2);
```

test/unit/smt_util/node_traversal_black.h
```diff
@@ -0,0 +1,150 @@
...
+    Node tb = d_nodeManager->mkConst(true);
+    Node eb = d_nodeManager->mkConst(false);
+    Node cnd = d_nodeManager->mkNode(XOR, tb, eb);
+
+    size_t count = 0;
+    for (auto i : NodePostorderIterable(cnd))
+    {
+      ++count;
+    }
+    TS_ASSERT(count == 3);
```
4tXJ7f:
```suggestion
    TS_ASSERT_EQUALS(count, 3);
```

test/unit/smt_util/node_traversal_black.h
```diff
@@ -0,0 +1,150 @@
...
+    Node tb = d_nodeManager->mkConst(true);
+    Node eb = d_nodeManager->mkConst(false);
+    Node cnd = d_nodeManager->mkNode(XOR, tb, eb);
+
+    NodePostorderIterator i = NodePostorderIterable(cnd).begin();
+    NodePostorderIterator end = NodePostorderIterable(cnd).end();
+    TS_ASSERT(*(i++) == tb);
+    TS_ASSERT(*(i++) == eb);
+    TS_ASSERT(*(i++) == cnd);
+    TS_ASSERT(i == end);
```
4tXJ7f:
```suggestion
    TS_ASSERT_EQUALS(*(i++), tb);
    TS_ASSERT_EQUALS(*(i++), eb);
    TS_ASSERT_EQUALS(*(i++), cnd);
    TS_ASSERT_EQUALS(i, end);
```

test/unit/smt_util/node_traversal_black.h
```diff
@@ -0,0 +1,150 @@
...
+#include "expr/node_builder.h"
+#include "expr/node_manager.h"
+#include "expr/node_value.h"
+#include "smt_util/node_traversal.h"
+
+using namespace CVC4;
+using namespace CVC4::kind;
+using namespace std;
+
+class NodeBlack : public CxxTest::TestSuite
```
4tXJ7f:
```suggestion
class NodeTraversalBlack : public CxxTest::TestSuite
```

test/unit/smt_util/node_traversal_black.h
```diff
@@ -0,0 +1,150 @@
...
+    TS_ASSERT(*i == tb);
+    TS_ASSERT(i != end);
+    ++i;
+    TS_ASSERT(*i == eb);
+    TS_ASSERT(i != end);
+    ++i;
+    TS_ASSERT(*i == cnd);
+    TS_ASSERT(i != end);
+    ++i;
+    TS_ASSERT(i == end);
```
4tXJ7f:
```suggestion
    TS_ASSERT_EQUALS(*i, tb);
    TS_ASSERT_NOT_EQUALS(i, end);
    ++i;
    TS_ASSERT_EQUALS(*i, eb);
    TS_ASSERT_NOT_EQUALS(i, end);
    ++i;
    TS_ASSERT_EQUALS(*i, cnd);
    TS_ASSERT_NOT_EQUALS(i, end);
    ++i;
    TS_ASSERT_EQUALS(i, end);
```

src/smt_util/node_traversal.h
```diff
@@ -0,0 +1,117 @@
+/*********************                                                        */
+/*! \file node_traversal.h
```
4tXJ7f: I think that the appropriate location for this file is in `src/expr` (similar to `node_algorithm`) because it is not really connected to SMT solving per se. Opinions @mpreiner @aniemetz?

src/smt_util/node_traversal.h
```diff
@@ -0,0 +1,117 @@
...
+  using difference_type = std::ptrdiff_t;
+
+  // Move/copy construction and assignment. Destructor.
+  NodePostorderIterable(NodePostorderIterable&&) = default;
+  NodePostorderIterable& operator=(NodePostorderIterable&&) = default;
+  NodePostorderIterable(NodePostorderIterable&) = default;
+  NodePostorderIterable& operator=(NodePostorderIterable&) = default;
+  ~NodePostorderIterable() = default;
+
+  NodePostorderIterator begin();
```
4tXJ7f:
```suggestion
  NodePostorderIterator begin() const;
```

src/smt_util/node_traversal.h
```diff
@@ -0,0 +1,117 @@
...
+  NodePostorderIterator end();
```
4tXJ7f:
```suggestion
  NodePostorderIterator end() const;
```

src/smt_util/node_traversal.h
```diff
@@ -0,0 +1,117 @@
...
+class NodePostorderIterable
+{
+ public:
+  NodePostorderIterable(TNode n);
+
+  // STL type definitions for an iterable
+  using iterator = NodePostorderIterator;
+  using value_type = TNode;
+  using reference = TNode&;
+  using difference_type = std::ptrdiff_t;
```
4tXJ7f: Also define the `iterator_category`. I believe that this is a `forward_iterator_tag`?

src/smt_util/node_traversal.cpp
```diff
@@ -0,0 +1,101 @@
+#include "node_traversal.h"
```
4tXJ7f: Add our usual header with copyright info.

src/smt_util/node_traversal.h
```diff
@@ -0,0 +1,117 @@
...
+  // While we could expand the stack, adding children until the top node is a
+  // leaf, we do not do so, because this is expensive, and we want construction
+  // to be cheap.
+  std::vector<TNode> d_stack;
+
+  // Whether (and how) we've visited a node.
+  // Absent if we haven't visited it.
+  // Set to `false` if we've pre-visited it (enqueued its children).
+  // Set to `true` if we've also post-visited it.
+  std::unordered_map<TNode, bool, TNodeHashFunction> d_visited;
```
4tXJ7f: [Data members go below member functions.](https://github.com/CVC4/CVC4/wiki/Code-Style-Guidelines#declaration-order)

src/smt_util/node_traversal.h
mpreiner: Yes, should be the same as node_algorithm.

src/smt_util/node_traversal.h
alex-ozdemir: Done in f90e06131

src/smt_util/node_traversal.h
alex-ozdemir: Done in f90e06131

src/smt_util/node_traversal.h
alex-ozdemir: Done in f90e06131

test/unit/smt_util/node_traversal_black.h
alex-ozdemir: Done in 0a0781f1e

test/unit/smt_util/node_traversal_black.h
alex-ozdemir: Done in 0a0781f1e

test/unit/smt_util/node_traversal_black.h
alex-ozdemir: Done in 0a0781f1e

test/unit/smt_util/node_traversal_black.h
alex-ozdemir: Done in 0a0781f1e

test/unit/smt_util/node_traversal_black.h
alex-ozdemir: Done in 0a0781f1e

test/unit/smt_util/node_traversal_black.h
alex-ozdemir: Done in 0a0781f

src/smt_util/node_traversal.h
alex-ozdemir: Done in 19a0080ef

alex-ozdemir: We'll probably want to move [the node visitor](https://github.com/CVC4/CVC4/blob/master/src/smt_util/node_visitor.h) at some point too.

src/smt_util/node_traversal.cpp
alex-ozdemir: Done in b3eb1e2bf

src/smt_util/node_traversal.h
alex-ozdemir: Done in c934b96ec and 8016712f4

src/smt_util/node_traversal.h
alex-ozdemir: Hmm. Since this isn't actually an iterator (it is a range-expression), I thought that defining `iterator_category` would be inappropriate. Am I mistaken?

My goal was to support range-based for loops. Now that I look at the [requirements](https://en.cppreference.com/w/cpp/language/range-for) though, I don't believe I even need these declarations. How about I get rid of all of them?

test/unit/smt_util/node_traversal_black.h
alex-ozdemir: Done in c934b96ec

I actually created two test classes, one for pre-order and one for post-order.

test/unit/smt_util/node_traversal_black.h
alex-ozdemir: Actually, I used `TS_ASSERT_DIFFERS` in place of `TS_ASSERT_NOT_EQUALS`. I don't believe the latter is defined.

src/smt_util/node_traversal.h
4tXJ7f: Oh right, yes. We should only define them on the `NodePostorderIterator` itself and not the `NodePostorderIterable`. I got the classes mixed up.

src/expr/node_traversal.h
```diff
@@ -0,0 +1,148 @@
...
+ ** in the top-level source directory) and their institutional affiliations.
+ ** All rights reserved.  See the file COPYING in the top-level source
+ ** directory for licensing information.\endverbatim
+ **
+ ** \brief Iterators for traversing nodes.
+ **/
+
+#include "cvc4_private.h"
+
+#ifndef CVC4__SMT_UTIL__NODE_TRAVERSAL_H
```
4tXJ7f: Update the include guard.

src/expr/node_traversal.h
```diff
@@ -0,0 +1,148 @@
...
+  NodeDfsIterator(TNode n, bool postorder);
+  // Construct an end-of-traversal iterator
+  NodeDfsIterator(bool postorder);
+
+  // STL type definitions for an iterator
+  using value_type = TNode;
+  using pointer = TNode*;
+  using reference = TNode&;
+  using iterator_category = std::forward_iterator_tag;
+  using difference_type = std::ptrdiff_t;
```
4tXJ7f: [Typedefs go before constructors.](https://github.com/CVC4/CVC4/wiki/Code-Style-Guidelines#declaration-order)

src/expr/node_traversal.h
alex-ozdemir: Done in c394cea1a

src/expr/node_traversal.h
alex-ozdemir: Done in c394cea1a

src/expr/node_traversal.h
```diff
@@ -0,0 +1,141 @@
...
+  std::vector<TNode> d_stack;
+
+  // Whether (and how) we've visited a node.
+  // Absent if we haven't visited it.
+  // Set to `false` if we've already pre-visited it (enqueued its children).
+  // Set to `true` if we've also already post-visited it.
+  std::unordered_map<TNode, bool, TNodeHashFunction> d_visited;
+
+  // Whether this is a post-order iterator (the alternative is pre-order)
+  //
```
4tXJ7f:
```suggestion
```

src/expr/node_traversal.cpp
```diff
@@ -0,0 +1,182 @@
...
+  }
+}
+
+void NodeDfsIterator::advanceUntilVisitIfJustConstructed()
+{
+  if (d_stack.size() == 1 && d_visited.size() == 0)
+  {
+    advanceUntilVisit();
+  }
+}
```
mpreiner: How about we rename this to `initialize` and have `d_initialized` flag that is set in this method?
```
if (!d_initialized)
{
    advanceUntilVisit();
    d_initialized = true;
}
```

src/expr/node_traversal.cpp
```diff
@@ -0,0 +1,182 @@
...
+    finishPostVisit();
+  }
+}
+
+void NodeDfsIterator::advanceUntilVisit()
+{
+  // While a node is enqueued ..
+  while (d_postorder ? !atPostVisit() : !atPreVisit())
+  {
+    advance();
```
mpreiner: Since `advance()` is only used here, I would inline the code. It makes the traversal less confusing. You should be also able to inline `atPostVisit()` and `atPreVisit()` then you don't have to find `visitEntry` multiple times and you can directly break out of the loop.

src/expr/node_traversal.cpp
```diff
@@ -0,0 +1,182 @@
...
+}
+
+NodeDfsIterator::NodeDfsIterator(bool postorder)
+    : d_stack(), d_visited(), d_postorder(postorder)
+{
+}
+
+NodeDfsIterator& NodeDfsIterator::operator++()
+{
+  std::cout << "post-increment: " << d_stack << std::endl;
```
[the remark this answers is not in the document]
alex-ozdemir: Oh. Awkward. I wish I could test for this.

src/expr/node_traversal.cpp
alex-ozdemir: Sure. Done in a049c682f

src/expr/node_traversal.cpp
```diff
@@ -0,0 +1,182 @@
...
+NodeDfsIterator NodeDfsIterator::operator++(int)
+{
+  NodeDfsIterator copyOfOld(*this);
+  ++*this;
+  return copyOfOld;
+}
+
+TNode& NodeDfsIterator::operator*()
+{
+  std::cout << "deref         : " << d_stack << std::endl;
```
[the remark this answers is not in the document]
alex-ozdemir: Done in a049c682f

src/expr/node_traversal.cpp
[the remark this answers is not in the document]
alex-ozdemir: Done in a049c682f.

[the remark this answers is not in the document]
alex-ozdemir: I removed `<iostream>`. I guess that's a simple enough check :)

src/expr/node_traversal.cpp
alex-ozdemir: Done in a049c682f

alex-ozdemir:
Thanks for the ideas, Mathias. I've implemented them, and also inlined `finishPostVisit` to its two call-sites in order to avoid an extra lookup in the main loop.

I did not inline `finishPreVisit`, since its lookup is not extra and it is also used in two locations,

src/expr/node_traversal.cpp
```diff
@@ -0,0 +1,160 @@
...
+{
+  Assert(!d_stack.empty());
+  TNode back = d_stack.back();
+  d_visited[back] = false;
+  // Use integer underflow to reverse-iterate
+  for (size_t n = back.getNumChildren(), i = n - 1; i < n; --i)
+  {
+    d_stack.push_back(back[i]);
+  }
+}
```
mpreiner: This method can be removed now.

src/expr/node_traversal.cpp
```diff
@@ -0,0 +1,160 @@
+/*********************                                                        */
+/*! \file node_traversal.cpp
+ ** \verbatim
+ ** Top contributors (to current version):
+ **   Alex Ozdemir
+ ** This file is part of the CVC4 project.
+ ** Copyright (c) 2009-2019 by the authors listed in the file AUTHORS
```
mpreiner:
```suggestion
 ** Copyright (c) 2009-2020 by the authors listed in the file AUTHORS
```

test/unit/expr/node_traversal_black.h
```diff
@@ -0,0 +1,281 @@
+/*********************                                                        */
+/*! \file node_traversal_black.h
+ ** \verbatim
+ ** Top contributors (to current version):
+ **   Alex Ozdemir
+ ** This file is part of the CVC4 project.
+ ** Copyright (c) 2009-2019 by the authors listed in the file AUTHORS
```
mpreiner:
```suggestion
 ** Copyright (c) 2009-2020 by the authors listed in the file AUTHORS
```

src/expr/node_traversal.h
```diff
@@ -0,0 +1,131 @@
+/*********************                                                        */
+/*! \file node_traversal.h
+ ** \verbatim
+ ** Top contributors (to current version):
+ **   Alex Ozdemir
+ ** This file is part of the CVC4 project.
+ ** Copyright (c) 2009-2019 by the authors listed in the file AUTHORS
```
mpreiner:
```suggestion
 ** Copyright (c) 2009-2020 by the authors listed in the file AUTHORS
```

src/expr/node_traversal.cpp
```diff
@@ -0,0 +1,160 @@
...
+{
+}
+
+NodeDfsIterator& NodeDfsIterator::operator++()
+{
+  // If we were just constructed, advance to first visit
+  initializeIfUninitialized();
+
+  // Void our current visit, and advance to the next one
+  d_current = TNode();
```
mpreiner: Is this really necessary?

src/expr/node_traversal.h
```diff
@@ -0,0 +1,131 @@
...
+
+  // Whether this is a post-order iterator (the alternative is pre-order)
+  bool d_postorder;
+
+  // Whether this iterator has been initialized (advanced to its first
+  // visit)
+  bool d_initialized;
+
+  // Current referent node. Valid if `d_initialized` and we're not at the end.
+  TNode d_current;
```
mpreiner: Can `d_current` be used to replace `d_initialized`, i.e., if `d_current.isNull()` we have to initialize.

src/expr/node_traversal.cpp
```diff
@@ -0,0 +1,160 @@
...
+      d_postorder(postorder),
+      d_initialized(true),
+      d_current(TNode())
+{
+}
+
+NodeDfsIterator& NodeDfsIterator::operator++()
+{
+  // If we were just constructed, advance to first visit
+  initializeIfUninitialized();
```
mpreiner: Why do we still have to call `initializeIfUninitialized` here if we later call `advanceToNextVisit` anyways?

src/expr/node_traversal.h
alex-ozdemir: As discussed in slack: Yes! Great idea!

src/expr/node_traversal.cpp
alex-ozdemir: Yeah. This is a bit ugly, but necessary. Short reason: if we're unintialized then we need to call `advanceToNextVisit` twice.

Why? If we're uninitialized, and `operator++` is called, then we need to (a) go to the first visit, and then (b) go to the next one.

src/expr/node_traversal.cpp
alex-ozdemir: Right now it is needed, but there are some other options.

The problem is that the end iterator has a null `d_current`, so a fully-incremented iterator should also have a null `d_current`.

The idea here is that we nullify `d_current`, and `advanceToNextVisit` will populate `d_current` if there is another node to visit. If not it will leave it as the null node.

A few other options:
* have `advanceToNextVisit` write in a null if needed
* do something fancy with `operator==` that I haven't thought of.

I'm certainly open to any of the above options, and others, if you have ideas.

src/expr/node_traversal.cpp
alex-ozdemir: Done in 246dd569e

src/expr/node_traversal.h
alex-ozdemir: Done in 246dd569e

src/expr/node_traversal.cpp
```diff
@@ -0,0 +1,145 @@
...
+  Assert(!d_current.isNull());
+
+  return d_current;
+}
+
+bool NodeDfsIterator::operator==(const NodeDfsIterator& other) const
+{
+  // The stack and current node uniquely represent traversal state. We need not
+  // use the scheduled node set. We also ignore the order: users should not
+  // compare nodes of different order.
```
4tXJ7f: Maybe assert that the orders are the same?

src/expr/node_traversal.h
```diff
@@ -0,0 +1,131 @@
...
+class NodeDfsIterable
+{
+ public:
+  NodeDfsIterable(TNode n);
+
+  // Modifying the traversal order
+  // Modify this iterable to be in post-order (default)
+  NodeDfsIterable& in_postorder();
+  // Modify this iterable to be in pre-order
+  NodeDfsIterable& in_preorder();
```
4tXJ7f: Use camelCase.

src/expr/node_traversal.h
```diff
@@ -0,0 +1,131 @@
...
+
+// Node wrapper that is iterable in DAG post-order
+class NodeDfsIterable
+{
+ public:
+  NodeDfsIterable(TNode n);
+
+  // Modifying the traversal order
+  // Modify this iterable to be in post-order (default)
+  NodeDfsIterable& in_postorder();
```
4tXJ7f: Use camelCase.

src/expr/node_traversal.cpp
4tXJ7f: Couldn't we just set `d_current` to `TNode()` after the while loop in `advanceToNextVisit`? That would seem quite natural to me (i.e. `advanceToNextVisit` would always update `d_current`).

src/expr/node_traversal.h
```diff
@@ -0,0 +1,131 @@
...
+  TNode d_current;
+};
+
+// Node wrapper that is iterable in DAG post-order
+class NodeDfsIterable
+{
+ public:
+  NodeDfsIterable(TNode n);
+
+  // Modifying the traversal order
```
4tXJ7f: Instead of having the methods below, why not just have the order as an argument in the constructor?

src/expr/node_traversal.cpp
alex-ozdemir: Yeah, I think that's the first option in the list above. I would support it. What do  you think, Mathias?

src/expr/node_traversal.cpp
alex-ozdemir: Done in 814f73316

src/expr/node_traversal.h
alex-ozdemir: Done in 5a28e541a

src/expr/node_traversal.h
alex-ozdemir: Done in 5a28e541a

src/expr/node_traversal.h
alex-ozdemir: Readability: It's immediately clear what `NodeDfsIteratable(n).inPreorder()` does, but `NodeDfsIterable(n, false)` wouldn't be clear until you read the docs.
I'd be happy to create an enum class, and pass that into the constructor (`NodeDfsIterable(n, Order::PRE)`). What do you think of that?

src/expr/node_traversal.cpp
mpreiner: Sounds good!

src/expr/node_traversal.cpp
mpreiner: OK, can you please add more documentation to make this clear?

src/expr/node_traversal.cpp
alex-ozdemir: Done in 836575c38

src/expr/node_traversal.cpp
alex-ozdemir: Sure. I added a bit of explanation inline in 836575c38. Let me know what you think.

src/expr/node_traversal.cpp
```diff
@@ -0,0 +1,150 @@
...
+      d_visited(),
+      d_postorder(postorder),
+      d_current(TNode())
+{
+}
+
+NodeDfsIterator& NodeDfsIterator::operator++()
+{
+  // If we were just constructed, advance to first visit, **before**
+  // advancing past it to the enxt visit (below).
```
mpreiner:
```suggestion
  // advancing past it to the next visit (below).
```

alex-ozdemir: Good catch! Thanks!

src/expr/node_traversal.h
4tXJ7f: Ah, yes. Sorry, I didn't realize which pattern you were shooting for here. I am fine with both solutions. I'd say that we just leave it as-is for.

src/expr/node_traversal.cpp
4tXJ7f: Oh, I was actually thinking about a literal `Assert()` since we don't want people to compare iterators of different orders (i.e. it would indicate a bug in the code). What do you think?

alex-ozdemir: Ah! That's what you meant. Yes, that's better.

alex-ozdemir: Done in a0eaf3e11

test/unit/expr/node_traversal_black.h
```diff
@@ -0,0 +1,281 @@
...
+ ** \verbatim
+ ** Top contributors (to current version):
+ **   Alex Ozdemir
+ ** This file is part of the CVC4 project.
+ ** Copyright (c) 2009-2020 by the authors listed in the file AUTHORS
+ ** in the top-level source directory) and their institutional affiliations.
+ ** All rights reserved.  See the file COPYING in the top-level source
+ ** directory for licensing information.\endverbatim
+ **
+ ** \brief Black box testing of note traversal iterators.
```
mpreiner:
```suggestion
 ** \brief Black box testing of node traversal iterators.
```

alex-ozdemir:
Thanks guys, for putting so much effort into this. It has certainly improved a lot thanks to both of you :tada:.

PR #3847, opened 2020-02-28
 Split collect model info by types in strings

ajreynol:
Towards a theory of sequences.

We will need to do similar splits per type for most of the functions throughout strings.


src/theory/strings/type_enumerator.h
```diff
@@ -125,6 +120,21 @@ class StringEnumeratorLength {
   }
 
   bool isFinished() { return d_curr.isNull(); }
+
+ private:
+  /** The type we are enumerating */
+  TypeNode d_type;
+  /** The cardinality of the alphabet */
+  uint32_t d_cardinality;
```
4tXJ7f: It feels weird to have a cardinality here. Shouldn't this be a property of the type? E.g. sequences of integers have an infinite alphabet.

src/theory/strings/type_enumerator.h
```diff
@@ -125,6 +120,21 @@ class StringEnumeratorLength {
...
+  /** The data (index to members) */
+  std::vector<unsigned> d_data;
```
4tXJ7f: I think that this should become a vector of nodes (single char strings for strings) and `mkCurr` should call `mkWord()`.

src/theory/strings/type_enumerator.h
ajreynol: Right, this will be generalized when sequences are supported.  For now, the EnumeratorLength class is guarded by only supporting `String`.  I can try to make more progress on this in this PR, but I wanted to keep the changes small.

src/theory/strings/type_enumerator.h
ajreynol: I think this will actually stay a vector of indices, where the difference is that a Node vector of characters will be generated on demand.  E.g. for sequences, there will be a level of indirection.

Again, this is something I could make more progress towards but it will require more substantial changes to this class.

PR #3852, opened 2020-03-02
Support unicode internal representation and escape sequences

ajreynol:
Work towards support for the strings standard. 

This updates the string solver and parser such that:
- The internal representation of strings is vectors of code points,
- Generation of the previous internal representation of strings has been relegated to the type enumerator. This is the code that ensures that "A" is the first character chosen for values of strings in models,
- The previous ad-hoc escape sequence handling is moved from the String class to the parser.  It will live there for at least one version of CVC4, until we no longer support non-smt-lib complaint escape sequences or non-printable characters in strings,
- Handle unicode escape sequences according to the SMT-LIB standard in String,
- Simplify a number of calls to String utility functions, since the conversion between the previous internal format and code points is now unnecessary,
- Fixed a bug in the handling of TO_CODE: it should be based on the alphabet cardinality, not the number of internal code points.


src/theory/quantifiers/sygus_sampler.cpp
```diff
@@ -561,7 +561,7 @@ Node SygusSampler::getRandomValue(TypeNode tn)
       {
         d_rstring_alphabet.push_back(ch);
         Trace("sygus-sample-str-alpha")
-            << " \"" << String::convertUnsignedIntToChar(ch) << "\"";
+            << " \"" << static_cast<char>(ch) << "\"";
```
4tXJ7f: Not a huge deal but this cast will potentially lose some bits.

src/util/regexp.cpp
```diff
@@ -265,62 +275,20 @@ std::size_t String::roverlap(const String &y) const {
...
-          t = t.substr(t.size() - 2, 2);
-          s = "\\x" + t;
-          // std::string s2 = static_cast<std::ostringstream*>(
-          // &(std::ostringstream() << (int)c) )->str();
-        }
-      }
-      str += s;
+    if (isPrintable(d_str[i]) || !useEscSequences)
+    {
+      str << static_cast<char>(d_str[i]);
```
4tXJ7f: Do we really need the `useEscSequences` argument? This cast potentially truncates the character value.

src/theory/quantifiers/sygus_sampler.cpp
ajreynol: Ill change to the \uN format.

src/util/regexp.cpp
ajreynol: In this case it won't since we are also checking that d_str[i] is printable.  It is important to print this way since I believe this is the code that prints e.g. (define-fun t () String *"AAAA"*) to a user.

PR #3883, opened 2020-03-02
Fixed shadow warnings for batch number 150

alex-ozdemir:


src/theory/theory_engine.cpp
```diff
@@ -470,17 +470,19 @@ void TheoryEngine::printAssertions(const char* tag) {
...
+            {
+              Trace(tag) << "[" << i << "]: ";
+            }
+            else
+            {
+              Trace(tag) << "(" << i << "): ";
+            }
+            Trace(tag) << (*it).d_assertion << endl;
           }
-          Trace(tag) << (*it).d_assertion << endl;
```
alex-ozdemir: Inserted a scope.

alex-ozdemir:
Thanks! I found a bug in the process, so I'm **really** glad you pointed that out. Fixed now.

PR #3893, opened 2020-03-02
Fix Wshadow warnings for batch 146.

mpreiner:


src/theory/strings/regexp_operation.cpp
```diff
@@ -826,15 +826,15 @@ void RegExpOpr::firstChars(Node r, std::set<unsigned> &pcset, SetNodes &pvset)
 
   if(Trace.isOn("regexp-fset")) {
     Trace("regexp-fset") << "END FSET(" << mkString(r) << ") = {";
-    for (std::set<unsigned>::const_iterator itr = pcset.begin();
-         itr != pcset.end();
-         ++itr)
+    for (std::set<unsigned>::const_iterator it = pcset.begin();
```
aniemetz: Convert into a for each loop?

src/theory/strings/regexp_operation.cpp
```diff
@@ -1495,24 +1494,25 @@ Node RegExpOpr::intersectInternal( Node r1, Node r2, std::map< PairNodes, Node >
         }
         if(Trace.isOn("regexp-int-debug")) {
           Trace("regexp-int-debug") << "Try CSET(" << cset.size() << ") = {";
-          for (std::vector<unsigned>::const_iterator itr = cset.begin();
-               itr != cset.end();
-               ++itr)
+          for (std::vector<unsigned>::const_iterator it = cset.begin();
```
aniemetz: Ditto

src/theory/strings/regexp_operation.cpp
```diff
@@ -1495,24 +1494,25 @@ Node RegExpOpr::intersectInternal( Node r1, Node r2, std::map< PairNodes, Node >
...
-            Trace("regexp-int-debug") << ( *itr );
+            Trace("regexp-int-debug") << (*it);
           }
           Trace("regexp-int-debug") << std::endl;
         }
         std::map< PairNodes, Node > cacheX;
-        for (std::vector<unsigned>::const_iterator itr = cset.begin();
-             itr != cset.end();
-             ++itr)
+        for (std::vector<unsigned>::const_iterator it = cset.begin();
```
aniemetz: Ditto

src/theory/strings/regexp_operation.cpp
aniemetz: Ah right, not possible, nevermind.

src/theory/strings/regexp_operation.cpp
aniemetz: Same

src/theory/strings/regexp_operation.cpp
aniemetz: Same

PR #3896, opened 2020-03-03
fix shadow warnings, group 67

4tXJ7f:


src/prop/minisat/core/Solver.cc
```diff
@@ -130,74 +130,92 @@ CRef Solver::TCRef_Undef = CRef_Undef;
 CRef Solver::TCRef_Lazy = CRef_Lazy;
 
 class ScopedBool {
```
aniemetz: Can you format this line, too?

4tXJ7f:
We discussed this offline and decided that we should go with minimal changes instead of changing the data members to start with `d_`.

PR #3905, opened 2020-03-03
Allow bv2int only for sublogics of QF_UFBVNIRA

yoni206:
This PR adds a check to the `smt_engine` that allows to use it only for sub-logics of `QF_UFBV`.


ajreynol:
Can you make this not depend on the logic?  What if I have a BV problem but did not specify a logic?

yoni206:
My recent commit allows arithmetic (changed the title as well). Regarding unspecified logic, I think the best place to abort is here:
https://github.com/CVC4/CVC4/blob/18fe192c29a9a2c37d1925730af01e906b9888c5/src/preprocessing/passes/bv_to_int.cpp#L319

But what kind of exception should I throw?

ajreynol:
I'm saying that I want to apply this for any logic. I would not expect your code to have to check what the logic is.  Ideally, your code should work when it encounters something that is not BV.

yoni206:
Well, theory combination could get problematic. For example, I don't know
how to handle arrays, and so I want to not allow using this flag when
arrays are present.

src/smt/smt_engine.cpp
```diff
@@ -1216,6 +1216,11 @@ void SmtEngine::setDefaults() {
 
   if (options::solveBVAsInt() > 0)
   {
+    if (!(d_logic <= LogicInfo("QF_UFBVNIRA"))) {
+      throw OptionException(
+          "--solve-bv-as-int=X only supported for QF_BV and QF_UFBV.");
```
ajreynol: Update

src/preprocessing/passes/bv_to_int.cpp
```diff
@@ -315,9 +315,8 @@ Node BVToInt::bvToInt(Node n)
             }
             else
             {
-              // Boolean variables are left unchanged.
               AlwaysAssert(current.getType() == d_nm->booleanType()
-                           || current.getType().isSort());
+                           || current.getType().isSort() || current.getType() == d_nm->integerType() || current.getType() == d_nm->realType() );
```
ajreynol: Can you generalize this?  Shouldn't it be that all types besides bit-vector are preserved?

yoni206: All types in the supported logics besides bit-vectors, except UFs, that need to change also. So I can change it to `current type not in [bitvector, UF]`,  but that would rely on the fact that we are in a supported logic. A bit less safe, and not a lot clearer in my opinion, but if u prefer it that way i can change.

ajreynol: As I mentioned, we should generalize to handle other logics.

yoni206: as i responded, that could be problematic :)
https://github.com/CVC4/CVC4/pull/3905#issuecomment-594108026

yoni206:
Talking offline with Andy, we come up with a better plan for increasing the scope of this pp. Details are in:
https://github.com/CVC4/cvc4-projects/issues/129

PR #3906, opened 2020-03-03
Wshfix

ntsis:


4tXJ7f:
The PR should be created against the `wshadow` branch and not `master`. I think we have to close this one and create a new one @ntsis .

Issue #3430, opened 2019-11-04
Check failure with --solve-int-as-bv=64 


wintered:
Hi all,

On the following formula CVC4 crashes with a Check failure when using `--solve-int-as-bv=64`. Without that option CVC4 returns `sat` which is the correct result. 

```
(set-logic QF_NIA)
(declare-fun WnRElIU_new () Int)
(declare-fun shifted1_a__4 () Int)
(declare-fun shifted2_lam3n3 () Int)
(assert (= shifted2_lam3n3 (div WnRElIU_new shifted1_a__4)))
(check-sat)
```

```
 getNumChildren() >= kind::metakind::getLowerBoundForKind(getKind())
Nodes with kind APPLY_UF must have at least 1 children (the one under construction has 0)
[3]    22210 abort      /home/windomin/solver_binaries/cvc4/build-2019-11-01-c547bd5/cvc4
```

Configuration: ` --solve-int-as-bv=64`
OS: Ubuntu 18.04
Revision: c547bd5

nafur:
With d19b800ac00feb44bfc6302f02695c8700e15c12 the output changed to
```
(error "Cannot translate to BV: (div WnRElIU_new shifted1_a__4)")
```
Then with 52082c72d78eee219e3049285d5df559dacac8b5 the output changed to
```(error "Illegal argument detected
bool CVC4::LogicInfo::isLinear() const

  `*this' is a bad argument; expected isTheoryEnabled(theory::THEORY_ARITH) to hold
Arithmetic not used in this LogicInfo; cannot ask whether it's linear")
```
which is still the output on current master.
The problem seems to be that `--solve-int-as-bv` forces the theory to `QF_BV` (and disables arithmetic). However, we then end up at https://github.com/CVC4/CVC4/blob/master/src/smt/process_assertions.cpp#L743 which identifies the theory of the division as arithmetic and then yields to calling `OperatorElim` that calls `isLinear()` on a non-arithmetic `LogicInfo`.

ajreynol:
We now output `(error "Cannot translate to BV: (div WnRElIU_new shifted1_a__4)")` again, this is expected behavior.

PR #3907, opened 2020-03-03
int-to-bv: Support uninterpreted functions and integer division

yoni206:
The current int-to-bv preprocessing pass does not support integer division and modulus. This PR adds support for these operations. Since internally they are represented using uninterpreted functions (for the case where the second argument is 0), this actually requires supporting uninterpreted functions in this preprocessing pass.

Solves #3430.


yoni206:
Semantics of div and mod is more complicated, and is different between bv and arith. Working on a correction. Closing the PR in order to avoid clogging travis.

PR #3920, opened 2020-03-04
Split string-specific operators from TheoryStringsRewriter

ajreynol:
Organization towards theory of sequences.

The motivation of this PR is to ensure that string-specific operators in the rewriter are in their own file; thus the use of `mkConst<String>` / `getConst<String>` is allowable in `rewriter_str.cpp`.


src/theory/strings/rewriter_str.h
```diff
@@ -0,0 +1,88 @@
...
+
+namespace CVC4 {
+namespace theory {
+namespace strings {
+
+/**
+ * An extension of TheoryStringsRewriter that handles operators that
+ * are specific to strings (and cannot be applied to sequences).
+ */
+class RewriterStr : public TheoryStringsRewriter
```
4tXJ7f: How about calling them `TheorySequenceRewriter` (for what is currently called `TheoryStringsRewriter`) and `TheoryStringsRewriter` (for what is currently `RewriterStr`)?

src/theory/strings/theory_strings_rewriter.cpp
```diff
@@ -1691,44 +1692,23 @@ RewriteResponse TheoryStringsRewriter::postRewrite(TNode node) {
...
     retNode = rewriteMembership(node);
   }
   else if (nk == STRING_TO_CODE)
   {
-    retNode = rewriteStringToCode(node);
+    retNode = RewriterStr::rewriteStringToCode(node);
+  }
+  else if (nk == STRING_FROM_CODE)
+  {
+    retNode = RewriterStr::rewriteStringFromCode(node);
```
4tXJ7f: Instead of having those here, the object-oriented way to do this would be to override `postRewrite` in `RewriterStr` and then to call the original `postRewrite` and add those cases in addition to that (otherwise the inheritance on `RewriterStr` doesn't seem to get us much).

src/theory/strings/rewriter_str.h
ajreynol: Sure, I can update this.

src/theory/strings/theory_strings_rewriter.cpp
ajreynol: This actually isn't possible currently since static functions can't be virtual.  (Is this changing in the future?)

PR #3921, opened 2020-03-05
Convert more uses of strings to words

ajreynol:
Towards theory of sequences.

Also adds documentation to strncmp/rstrncmp and adds them to the Word interface.


src/theory/strings/core_solver.cpp
```diff
@@ -1137,13 +1137,14 @@ void CoreSolver::processSimpleNEq(NormalForm& nfi,
       // Constants in both normal forms.
       //
       // E.g. "abc" ++ ... = "ab" ++ ...
-      String c1 = x.getConst<String>();
-      String c2 = y.getConst<String>();
+      size_t lenX = Word::getLength(x);
+      size_t lenY = Word::getLength(y);
```
4tXJ7f: Nit: It is a bit confusing to have both `lenX` and `xLen` in this function. Can we maybe rename `xLen` to `xLenN` (length of `x` as a node) and `lenX` to `xLen`? I prefer `xLen` over `lenX` because the different casing of `x` makes things a bit weird.

src/theory/strings/word.h
```diff
@@ -45,6 +45,22 @@ class Word
   /** Return true if x is empty */
   static bool isEmpty(TNode x);
 
+  /** string compare
+   *
+   * Returns true if x is equal to y for their first n characters.
+   * If n is larger than the length of x or y, this method returns
+   * true if and only if x is equal to y.
+   */
+  static bool strncmp(TNode x, TNode y, std::size_t n);
```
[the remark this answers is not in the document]
ajreynol: I think i prefer this name; notice that we also have "substr".  I think its fine if some of the names are string-inspired.

src/theory/strings/core_solver.cpp
ajreynol: I will rename xLen/yLen to xLenTerm/yLenTerm.

PR #3922, opened 2020-03-05
Added unit-cube-like test for branch and bound

amaleewilson:
Changed branch and bound to give preference to a simple rounding of the real solution.


src/theory/arith/theory_arith_private.cpp
```diff
@@ -3938,15 +3938,36 @@ Node TheoryArithPrivate::branchIntegerVariable(ArithVar x) const {
...
+    Node literal = d_containing.getValuation().ensureLiteral(eq);
+    d_containing.getOutputChannel().requirePhase(literal, true);
+    Node lem = NodeManager::currentNM()->mkNode(kind::OR, literal, lem);
+  }
+  else
+  {
+    //Node eq = Rewriter::rewrite(NodeManager::currentNM()->mkNode(kind::EQUAL, var, mkRationalNode(floor_d+1)));
+    //Node diseq = eq.notNode();
+    Node ub = Rewriter::rewrite(NodeManager::currentNM()->mkNode(kind::LEQ, var, mkRationalNode(floor_d)));
+    Node lb = ub.notNode();
```
ahmed-irfan: @amaleewilson i see this is not your code. Nevertheless, it seems wrong. 
It should have been `Node lb = mkNode(kind::GEQ, var, mkRationalNode(floor_d + 1))`

@ajreynol What do you think?

ahmed-irfan:
@amaleewilson tests failed...

src/theory/arith/theory_arith_private.cpp
```diff
@@ -3938,15 +3938,36 @@ Node TheoryArithPrivate::branchIntegerVariable(ArithVar x) const {
...
 
-  //Node eq = Rewriter::rewrite(NodeManager::currentNM()->mkNode(kind::EQUAL, var, mkRationalNode(floor_d+1)));
-  //Node diseq = eq.notNode();
+  if (options::cubeTest())
+  {
+    Trace("integers") << "cube test enabled"  << endl;
+    Integer ceil_d = d.ceiling();
+    Rational f = r - floor_d;
+    // Multiply by -1 to get abs value.
+    Rational c = (r - ceil_d) * (-1);
```
ahmed-irfan: simplified: `c = ceil_d - r`

src/theory/arith/theory_arith_private.cpp
```diff
@@ -3938,15 +3938,36 @@ Node TheoryArithPrivate::branchIntegerVariable(ArithVar x) const {
...
+  if (options::cubeTest())
+  {
+    Trace("integers") << "cube test enabled"  << endl;
+    Integer ceil_d = d.ceiling();
+    Rational f = r - floor_d;
+    // Multiply by -1 to get abs value.
+    Rational c = (r - ceil_d) * (-1); 
+ 
+    Integer nearest;
+    (c > f) ? nearest = floor_d : nearest = ceil_d;
```
ahmed-irfan: (A matter of style)

I would prefer to use:
``` Integer nearest = (c > f) ? floor_d : ceil_d; ```

File: src/theory/arith/theory_arith_private.cpp [hunk 4]
```diff
@@ -3938,15 +3938,36 @@ Node TheoryArithPrivate::branchIntegerVariable(ArithVar x) const {
...
+    Node lb = Rewriter::rewrite(NodeManager::currentNM()->mkNode(kind::GEQ, var, mkRationalNode(nearest + 1)));
+    Node lem = NodeManager::currentNM()->mkNode(kind::OR, ub, lb);
+    Node eq = Rewriter::rewrite(NodeManager::currentNM()->mkNode(kind::EQUAL, var, mkRationalNode(nearest)));
+    Node literal = d_containing.getValuation().ensureLiteral(eq);
+    d_containing.getOutputChannel().requirePhase(literal, true);
+    Node lem = NodeManager::currentNM()->mkNode(kind::OR, literal, lem);
+  }
+  else
+  {
+    //Node eq = Rewriter::rewrite(NodeManager::currentNM()->mkNode(kind::EQUAL, var, mkRationalNode(floor_d+1)));
```
ajreynol: You can remove this commented out code.

src/theory/arith/theory_arith_private.cpp
ajreynol: I agree on this style point.

src/theory/arith/theory_arith_private.cpp
ajreynol: I think these are equivalent.  x >= c+1 is equivalent to ~(x <= c ) right (for integers)?  I prefer this way since its clearer that this is indeed a proper split.

src/theory/arith/theory_arith_private.cpp
```diff
@@ -3938,15 +3938,36 @@ Node TheoryArithPrivate::branchIntegerVariable(ArithVar x) const {
...
+    Integer nearest;
+    (c > f) ? nearest = floor_d : nearest = ceil_d;
 
-  Node ub = Rewriter::rewrite(NodeManager::currentNM()->mkNode(kind::LEQ, var, mkRationalNode(floor_d)));
-  Node lb = ub.notNode();
+    Node ub = Rewriter::rewrite(NodeManager::currentNM()->mkNode(kind::LEQ, var, mkRationalNode(nearest - 1)));
+    Node lb = Rewriter::rewrite(NodeManager::currentNM()->mkNode(kind::GEQ, var, mkRationalNode(nearest + 1)));
+    Node lem = NodeManager::currentNM()->mkNode(kind::OR, ub, lb);
+    Node eq = Rewriter::rewrite(NodeManager::currentNM()->mkNode(kind::EQUAL, var, mkRationalNode(nearest)));
+    Node literal = d_containing.getValuation().ensureLiteral(eq);
```
ajreynol: Can you give a few comments here (one or two lines would suffice) to state the intuition behind this code.

src/theory/arith/theory_arith_private.cpp
ahmed-irfan: > I think these are equivalent. x >= c+1 is equivalent to ~(x <= c ) right (for integers)?
I agree it is right for integers. However, the branch and bound lemmas are for LRA solver.
And actually we are not creating a split but removing a rational (non-integral) of the solution space.

For example, if LRA solver returned the solution x = 3.5 then the branch-and-bound lemma is going to tell the LRA solver that  x <=3 \/ x >= 4.

ajreynol: The rewriter will take care of this.  Assuming x's type is integer, the rewriter will turn either constraint into the same thing (a normalized LEQ atom).

ahmed-irfan: i see. Then I agree with you.

Thanks for clarifying :)

src/theory/arith/theory_arith_private.cpp
amaleewilson: I knew something felt wrong when I wrote this. Thanks!

src/theory/arith/theory_arith_private.cpp
amaleewilson: This commented out code was there when I checked out this file, which is why I left it. I'm happy to remove it, but not sure if someone wanted it to remain there.

src/theory/arith/theory_arith_private.cpp
amaleewilson: > I think these are equivalent. x >= c+1 is equivalent to ~(x <= c ) right (for integers)? I prefer leaving the code unchanged here since its clearer that this is indeed a proper split.

These comments and this code is unchanged from the original.

src/theory/arith/theory_arith_private.cpp
amaleewilson: ` error: no match for ‘operator-’ (operand types are ‘CVC4::Integer’ and ‘const CVC4::Rational’)
     Rational c = ceil_d - r;
                  ~~~~~~~^~~
`

src/theory/arith/theory_arith_private.cpp
ajreynol: I think its fine to remove (its been our policy to not have any dead/commented out code in master).

src/theory/arith/theory_arith_private.cpp
```diff
@@ -3937,15 +3937,33 @@ Node TheoryArithPrivate::branchIntegerVariable(ArithVar x) const {
   TNode var = d_partialModel.asNode(x);
   Integer floor_d = d.floor();
 
-  //Node eq = Rewriter::rewrite(NodeManager::currentNM()->mkNode(kind::EQUAL, var, mkRationalNode(floor_d+1)));
-  //Node diseq = eq.notNode();
-
-  Node ub = Rewriter::rewrite(NodeManager::currentNM()->mkNode(kind::LEQ, var, mkRationalNode(floor_d)));
-  Node lb = ub.notNode();
-
+  Node lem;
```
mpreiner: Add `NodeManger *nm = NodeManager::currentNM();` here and use `nm` instead of calling `NodeManager::currentNM()` each time.

src/options/didyoumean_test.cpp
```diff
@@ -751,6 +751,7 @@ set<string> getOptionStrings() {
       "no-sets-propagate",
       "sets-eager-lemmas",
       "no-sets-eager-lemmas",
+      "cube-test",
```
mpreiner: You don't have to add `cube-test` here. This is just a test file and horribly outdated. We should probably remove the file in the future.

test/regress/regress1/arith/arith-brab-test.smt2
```diff
@@ -0,0 +1,21 @@
+; EXPECT: sat
```
ajreynol: Will need to additionally add `; COMMAND-LINE: --arith-brab` before the `; EXPECT` line.

src/options/arith_options.toml
```diff
@@ -551,3 +551,12 @@ header = "options/arith_options.h"
   default    = "true"
   read_only  = true
   help       = "whether to increment the precision for irrational function constraints"
+
+[[option]]
+  name       = "brabTest"
+  category   = "regular"
+  long       = "arith-brab"
+  type       = "bool"
+  default    = "false"
```
ahmed-irfan: should we turn it on by default?
@ajreynol @amaleewilson what do you think?

ajreynol: Yes, we should enable by default.

ajreynol:
Ok, last remaining item is to add `--no-arith-brab` to one of the regressions, then LGTM.

As mentioned, would be okay to add this to some benchmark where branch and bound kicks in but `--arith-brab` is not necessary for solving.  (To test this, you can try just adding an `AlwaysAssert(false)` to the branch and bound code, and then see which regressions fail.)

amaleewilson:
> Ok, last remaining item is to add `--no-arith-brab` to one of the regressions, then LGTM.
> 
> As mentioned, would be okay to add this to some benchmark where branch and bound kicks in but `--arith-brab` is not necessary for solving. (To test this, you can try just adding an `AlwaysAssert(false)` to the branch and bound code, and then see which regressions fail.)

Can I copy the test that already

aniemetz:
regress0/nl/ext-rew-aggr-test.smt2 fails on Travis

ajreynol:
I believe @amaleewilson  said `regress0/nl/ext-rew-aggr-test.smt2 ` can be solved by adding `--nl-ext-tplanes`.

amaleewilson:
> I believe @amaleewilson said `regress0/nl/ext-rew-aggr-test.smt2 ` can be solved by adding `--nl-ext-tplanes`.

Yes. Should I add this option to that test for this PR?

ajreynol:
Yes, do add that option. We will need Travis CI to pass before merging.

PR #3927, opened 2020-03-06
Set assertion in `CnfStream::ensureLiteral()`

### LINKED ISSUE #3814: --check-unsat-cores fatal failure at /CVC4/src/proof/cnf_proof.cpp:211
Hi,
For this formula:
```
(declare-fun a () (Array Bool Bool))
(declare-fun b () (Array Bool Bool))
(declare-fun c () Bool)
(declare-fun d () Bool)
(declare-fun e () Bool)
(assert (= a (store b c d)))
(assert (= e (select a (select b d))))
(check-sat)
```
CVC4 --check-unsat-cores throws out a fatal failure:
```
Fatal failure within CVC4::Node CVC4::CnfProof::getCurrentAssertion() at /home/suz/software/CVC4/src/proof/cnf_proof.cpp:211
Check failure
 d_currentAssertionStack.size()
Aborted
```

OS: Ubuntu 18.04
Commit: 6e17dd6

4tXJ7f:
Fixes #3814. `CnfProof` has a stack of assertions that are being
converted to clauses. `CnfStream::ensureLiteral()` can result in clauses
being added to the SAT solver. When adding a clause, we require an
assertion that can be associated with the clause
(https://github.com/CVC4/CVC4/blob/ba6ade0fc3f4cd339885652bb9bf5c87113c498d/src/prop/minisat/core/Solver.cc#L471-L476).
However, in the issue that was reported, the stack was empty, resulting
in an assertion failure. This commit fixes the issue by setting the
current assertion to be the null node when a literal is being ensured
(and changing the proof code to update the assertion associated with a
literal if it is currently null). This should be ok since the clauses
are not inputs or lemmas (if they are, the assertion associated with the
clause will be updated).


src/proof/cnf_proof.cpp
```diff
@@ -135,8 +135,11 @@ void CnfProof::setClauseAssertion(ClauseId clause, Node expr) {
   // case we keep the first assertion. For example asserting a /\ b
   // and then b /\ c where b is an atom, would assert b twice (note
   // that since b is top level, it is not cached by the CnfStream)
-  if (d_clauseToAssertion.find(clause) != d_clauseToAssertion.end())
+  const auto& it = d_clauseToAssertion.find(clause);
```
yoni206: Can u document this null node trick in the comments above this line?

src/prop/cnf_stream.cpp
```diff
@@ -160,8 +160,13 @@ void TseitinCnfStream::ensureLiteral(TNode n, bool noPreregistration) {
     // If we were called with something other than a theory atom (or
     // Boolean variable), we get a SatLiteral that is definitionally
     // equal to it.
+
```
yoni206: same

test/regress/regress0/arrays/issue3814.smt2
```diff
@@ -0,0 +1,12 @@
+; REQUIRES: proof
+; COMMAND-LINE: --produce-unsat-cores
+; EXPECT: sat
```
yoni206: sat and unsat core?

src/proof/cnf_proof.cpp
4tXJ7f: Good point, added comment.

src/prop/cnf_stream.cpp
4tXJ7f: Good point, added comment.

test/regress/regress0/arrays/issue3814.smt2
4tXJ7f: Yeah, it is a bit funny. The input that the issue was reported on is actually `sat`. The assertion failure ocured during bookkeeping that we do when we have the production of unsat cores enabled (they used `--check-unsat-cores`). Rather than trying to come up with an example that triggers the bug, I've opted to reuse the original example and simply enabling unsat core production.

PR #3929, opened 2020-03-06
Remove tester name from APIs

ajreynol:
This removes the field "tester name" from the Expr-level and Term-level APIs.  This field is an artifact of parsing and thus should be handled in the parsers.

This refactor uncovered an issue in our regressions, namely our smt version >= 2.6 was not strictly complaint, since the symbol `is-cons` was being automatically defined for testers of constructors `cons`. This disables this behavior when strict mode is enabled.  It updates the regressions with this issue.

This is work towards parser migration.

### DIFF
--- a/src/parser/smt2/smt2.cpp
+++ b/src/parser/smt2/smt2.cpp
@@ -507,14 +507,11 @@ api::Term Smt2::getExpressionForNameAndType(const std::string& name,
 
 bool Smt2::getTesterName(api::Term cons, std::string& name)
 {
-  if (v2_6())
+  if (v2_6() && strictModeEnabled())
   {
     // 2.6 or above uses indexed tester symbols, if we are in strict mode,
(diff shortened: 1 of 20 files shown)


src/parser/smt2/smt2.cpp
```diff
@@ -505,6 +505,23 @@ api::Term Smt2::getExpressionForNameAndType(const std::string& name,
   return Parser::getExpressionForNameAndType(name, t);
 }
 
+bool Smt2::getTesterName(api::Term cons, std::string& name)
+{
+  if (v2_6())
+  {
+    // 2.6 or above uses indexed tester symbols, if we are in strict mode,
+    // we do not automatically define is-cons for constructor cons.
+    if (strictModeEnabled())
```
4tXJ7f: Nit: Combine the if-statements above?

PR #3934, opened 2020-03-07
Explicit end marker for models printed in the CVC language

Ying1123:
When communicating with CVC4 using pipes and the CVC language, it was not possible to determine when all the lines of a model have been printed.

This change adds begin and end markers as the example below:
```
MODEL BEGIN
x : INT = -3; 
y : INT = 0;  
z : INT = 0;
MODEL END;
```


src/printer/cvc/cvc_printer.cpp
```diff
@@ -1173,6 +1173,23 @@ void DeclareFunctionCommandToStream(std::ostream& out,
...
+  std::string ln;
+  while (std::getline(c, ln))
+  {
+    out << "; " << ln << std::endl;
+  }
+
+  // print the model
+  out << "MODEL BEGIN" << std::endl;
+  this->Printer::toStream(out, m);
+  out << "MODEL END" << std::endl;
```
4tXJ7f:
```suggestion
  out << "MODEL END;" << std::endl;
```

test/regress/regress0/bv/ackermann4.smt2
```diff
@@ -12,4 +12,5 @@
 (assert (= (f (f (f v0))) (g (f v0))))
 
 (check-sat)
+(get-model)
```
4tXJ7f: The expected output of this method needs to be adjusted to include the model. Note that you want a test with a unique solution for this (because otherwise the test fails when CVC4 happens to pick a different model), so it would probably be better to create a separate test.

Ying1123: Thanks! changed to regress0/print_lambda.cvc

test/regress/regress0/print_lambda.cvc
```diff
@@ -1,7 +1,9 @@
 % SCRUBBER: sed -e 's/f : (INT) -> INT = (LAMBDA(.*:INT): 0);$/f : (INT) -> INT = (LAMBDA(VAR:INT): 0);/'
 % COMMAND-LINE: --produce-models
 % EXPECT: sat
+% EXPECT: MODEL BEGIN
 % EXPECT: f : (INT) -> INT = (LAMBDA(VAR:INT): 0);
```
yoni206: I am not completely sure CVC4 will always generate this value for `f`. It is definitely not the only possibility for it.
Maybe add a much simpler test, something like `x=5` over the integers where `x` is the only variable. Then there is a single model.

Ying1123: Interesting, I didn't notice that. So how the old code can pass this test case?

yoni206: i guess CVC4 always returns this model, but you are right in the sense that Andres' and my concerns apply to the situation before this PR... So I guess I would be OK with leaving this test as is.

Ying1123: new one added

yoni206:
@4tXJ7f should this be enabled/disabled by an option or turned on by default? I can imagine users who have scripts that parse models without these markers, and these users might have a problem upgrading.

4tXJ7f:
Yes, I think that this should be the default and we don't need an option to disable it. We should probably mention it in the [NEWS](https://github.com/CVC4/CVC4/blob/master/NEWS) file though. The reason for introducing this is actually to make parsing possible when you are communicating with CVC4 over pipes and doing incremental solving.

4tXJ7f:
Note: When merging this, I'd change the title to exclude #9. And instead say in the commit description "Fixes https://github.com/CVC4/cvc4-wishues/issues/9." This should automatically close the issue when merging it.

PR #3985, opened 2020-03-09
Fix -Wshadow warning in test/unit/context/context_mm_black.

aniemetz:


test/unit/context/context_mm_black.h
```diff
@@ -46,11 +46,15 @@ class ContextBlack : public CxxTest::TestSuite {
     unsigned chunkSizeBytes = 16384;
     unsigned maxFreeChunks = 100;
     unsigned piecesPerChunk = 13;
-    unsigned len = chunkSizeBytes / piecesPerChunk; // Length of the individual block
-    unsigned N = maxFreeChunks*piecesPerChunk;
+    unsigned len;
+    unsigned N;
+
+    len = chunkSizeBytes / piecesPerChunk;  // Length of the individual block
+    N = maxFreeChunks * piecesPerChunk;
```
4tXJ7f: Why move the variable initializations? [I prefer to always initialize variables when declaring them.](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Res-always)

test/unit/context/context_mm_black.h
```diff
@@ -70,17 +74,19 @@ class ContextBlack : public CxxTest::TestSuite {
...
     // Push, then allocate, then pop all at once
-    for(unsigned p = 0; p < 5; ++ p) {
+    for (unsigned p = 0; p < 5; ++p)
+    {
       d_cmm->push();
-      for(unsigned i = 1; i < N; ++i) {
-        unsigned len = i * factor;
+      for (unsigned i = 1; i < N; ++i)
+      {
+        len = i * factor;
```
4tXJ7f: This is a unit test, so I don't really care. However, I think that in general, it's better not to reuse a variable for two different things.

test/unit/context/context_mm_black.h
aniemetz: Because `len` is reused and that's why *I* prefer to have it declared without assignment (if it is a simple type), because it is easier to see it that way.

test/unit/context/context_mm_black.h
aniemetz: I could have moved the initialization to the head of the `for` loop, it's really in two separate blocks, and honestly, I don't share your sentiment. But for better readability I didn't move it to the for loop header. I actually don't really want to change this.

PR #3988, opened 2020-03-09
Fix issue with reset-assertions.

aniemetz:
Calling (reset-assertions) in start mode was not handled correctly.
Additionally, when calling (check-sat) after (reset-assertions) after a
(check-sat) call that answered unsat, we answered unsat instead of sat.
This cleans up and fixes reset-assertions) handling.


src/smt/smt_engine.cpp
```diff
@@ -5560,24 +5565,46 @@ void SmtEngine::reset()
...
+  // (see solver execution modes in the SMT-LIB standard)
   Assert(d_userLevels.size() == 0 && d_userContext->getLevel() == 1);
   d_context->popto(0);
   d_userContext->popto(0);
   DeleteAndClearCommandVector(d_modelGlobalCommands);
   d_userContext->push();
   d_context->push();
+
+  // Create new PropEngine.
+  d_propEngine.reset(nullptr);
```
4tXJ7f:
```suggestion
```
This line should not be necessary.

test/regress/regress0/smtlib/reset-assertions.smt2
```diff
@@ -0,0 +1,11 @@
...
+; EXPECT: sat
+(set-logic QF_BV)
+(declare-fun v0 () (_ BitVec 4))
+(set-option :produce-models true)
+(reset-assertions)
+(set-option :incremental true)
+(assert (and (= v0 (_ bv0 4)) (distinct v0 (_ bv0 4))))
+(check-sat)
+(reset-assertions)
+(check-sat)
```
4tXJ7f: Can we have an additional unit test that tests `global-declarations` with `reset-assertion`? Something like:

```
(set-option :global-declarations true)
(set-logic QF_BV)
(declare-const x (_ BitVec 4))
(assert (= x (bvneg x)))
(check-sat)
(reset-assertions)
(assert (= x x))
(check-sat)
```

src/smt/smt_engine.cpp
aniemetz: It is. We have to make sure that the PropEngine is destroyed before the new object is constructed, else we have issues with the order in how statistics are registered.

test/regress/regress0/smtlib/reset-assertions.smt2
aniemetz: I added a new test (a different one from above), and I also moved an existing test into `smtlib`. https://github.com/CVC4/CVC4/pull/3988/commits/871f43634aa3602e8e5d46da043a813168249f04

src/smt/smt_engine.cpp
4tXJ7f: I see, that sucks. Thanks for the explanation.

aniemetz: I only documented this on line 939 above, do you want me to copy that down to here, too?

aniemetz: Done

PR #4014, opened 2020-03-10
Refactor enumerator for strings

ajreynol:
Towards theory of sequences.  This separates out a virtual base class and utility class for the strings enumerator, which will be extended for sequences later.


ajreynol:
With sequences, there will be two new cases: `SequenceEnumLen` which inherits from `SEnumLen`, and a type enumerator for sequences.  

I kept WordIter separate from SEnumLen for two reasons: (1) it is potentially for general use (not only related to string enumeration), (2) I don't want to make SEnumLen do too many things.

PR #4034, opened 2020-03-11
Make handling of illegal internal representatives in quantifiers engine more robust

### LINKED ISSUE #4002: cbqi-all + fmf-fun: fatal failure at /theory/quantifiers/instantiate.cpp:189
Hi,
For this formula:
~~~~
(set-logic UFNRA)
(set-option :cbqi-all true)
(set-option :fmf-fun true)
(declare-sort S0 0)
(declare-const S0-0 S0)
(check-sat)
(assert (forall ((q2 S0)) (not (distinct q2 S0-0))))
(check-sat)
~~~~

CVC4 incremental mode throws out a fatal failure:
~~~~
sat
Fatal failure within bool CVC4::theory::quantifiers::Instantiate::addInstantiation(CVC4::Node, std::vector<CVC4::NodeTemplate<true> >&, bool, bool, bool) at /home/peisen/test/tofuzz/CVC4/src/theory/quantifiers/instantiate.cpp:189
Check failure

 false
~~~~

OS: Ubuntu 16.04
Commit: 7d3df3b642

ajreynol:
Fixes #4002 (that benchmark is now unknown).

The experimental option` --cbqi-all` previously had some issues when combined with finite model finding.  When these two options are used simultaneously, it may be the case that certain equivalence classes are "illegal" since they contain only terms that are ineligible for instantiation. 

The previous code threw a warning when this occurred which in extreme cases allowed for potentially ineligible terms for instantiation.  The new code is more conservative: we never choose illegal internal representatives and instead set the incomplete flag in finite model finding when this occurs.

A block of code changed indentation in this PR, which was updated to the new standards.


src/theory/quantifiers/equality_query.cpp
```diff
@@ -111,74 +111,74 @@ Node EqualityQueryQuantifiersEngine::getInternalRepresentative(Node a,
...
     }
   }
+  TypeNode v_tn = q.isNull() ? a.getType() : q[0][index].getType();
   if (options::quantRepMode() == options::QuantRepMode::EE)
   {
-    return r;
+    int score = getRepScore(r, q, index, v_tn);
+    if (score >= 0)
+    {
+      // if we are not a valid representative, try to select one below
```
HanielB: it's a bit confusing that you say this but then exits the method. Please rephrase to make clear you are referring to the negative case, when you did not enter the `if`

PR #4037, opened 2020-03-11
New C++ API: Remove support for (reset).

aniemetz:
Supporting SMT-LIB's `(reset)` command on the API level is dangerous and not required -- it's sufficient to just destroy the solver object and create a new one if necessary. It's dangerous because invalidated objects can be passed in after a reset, and we would need to find a clean way to guard against this. We want to guard against this in the future, but for now it is cleaner to make it explicit (by not having this functionality in the API but forcing the user to destroy and recreate the solver object) that these objects can't be reused.

### DIFF
--- a/src/main/command_executor.cpp
+++ b/src/main/command_executor.cpp
@@ -24,7 +24,6 @@
 #include <string>
 #include <vector>
 
-#include "api/cvc4cpp.h"
 #include "main/main.h"
 #include "smt/command.h"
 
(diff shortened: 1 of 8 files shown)


src/main/command_executor.cpp
```diff
@@ -49,14 +48,17 @@ void setNoLimitCPU() {
...
+    : d_solver(nullptr),
+      d_smtEngine(nullptr),
       d_options(options),
       d_stats("driver"),
       d_result(),
-      d_replayStream(NULL)
-{}
+      d_replayStream(nullptr)
+{
+  d_solver.reset(new api::Solver(&options));
```
4tXJ7f: Nit: this could actually be move to the member initializer list above.

Issue #789, opened 2017-08-19
Cyclic substitution produced by Arith ITE utils


mpreiner:
**Imported from Bugzilla**
Reporter: *Morgan Deters*
Status: *NEW*
Severity: *normal*
Assigned to: *Clark Barrett*
Component: *Miscellaneous*
Milestone: *1.6*
Version: *master*
Platform: *PC*
OS: *Linux*

Original attachment names and IDs:
- _[memory.smt2](http://cvc4.cs.stanford.edu/bugzilla_attachments/230_memory.smt2)_ (memory hog, ID 230)

On 2014-06-20 20:52:38 -0700, Morgan Deters wrote:
> Created attachment 230
> memory hog
> 
> Attached testcase eats gigs and gigs of memory really quickly.  Investigate.

On 2014-06-20 21:07:34 -0700, Morgan Deters wrote:
> Derived from a scrambling of 20\-vars/cut\_lemma\_01\_006.smt2.

ajreynol:
This is related to --ite-simp.  The arithmetic ITE utility is generating a cyclic substitution which is causing the SubstitutionMap to infinite loop.

PR #4038, opened 2020-03-11
Do not enable some SMT-COMP specific options by default

ajreynol:
Moves SMT-COMP-specific options to the SMT-COMP script.  Both of these options have led to issues (segfaults or infinite loops).

Issue #789 can be downgraded to "minor" after this PR.

Btw, I did not add these specialized options to the "incremental" script of SMT-COMP, since I'm assuming they should not be used there.


ajreynol:
FYI @barrettcw

PR #4040, opened 2020-03-11
Simplifications to the Datatypes API

ajreynol:
Removes DatatypeSelectorDecl and DatatypeDeclSelfSort.  Add selectors is now inlined. A special case is added for the "self selector", instead of using a class as a dummy argument.

I updated the Python files, although would be helpful to double check this is correct.


ajreynol:
Thanks for the review and fix, marking "do not merge" so @aina can review as well.

PR #4044, opened 2020-03-12
Perform theory widening eagerly

### LINKED ISSUE #3971: Fatal failure for QF_NIA instance at theory_arith_private.cpp:4099
Hi,
For this QF_NIA formula:
[theory_arith_private4096.txt](https://github.com/CVC4/CVC4/files/4306955/theory_arith_private4096.txt)

Without any other options, CVC4 incremental mode throws out a fatal failure:
~~~~
sat
Fatal failure within CVC4::Node CVC4::theory::arith::TheoryArithPrivate::explain(CVC4::TNode) at /home/peisen/test/tofuzz/CVC4/src/theory/arith/theory_arith_private.cpp:4096
Check failure

 !c->isAssumption()
~~~~

OS: Ubuntu 16.04
Commit: 7d3df3b642aa1b

### LINKED ISSUE #3991: dump-unsat-cores: fatal failure at theory/theory_engine.cpp:1728
Hi,
For this formula:
[theoryengine1728.txt](https://github.com/CVC4/CVC4/files/4310194/theoryengine1728.txt)

---minimized input----
~~~~
(set-logic NIA)
(set-option :dump-unsat-cores true)
(declare-const i7 Int)
(check-sat)
(declare-const v18 Bool)
(assert (and (= i7 93) (or (> 19 i7) v18) v18))
(check-sat)
(check-sat)
(assert (> 19 i7))
(assert (> (div i7 0) 0))
(check-sat)
~~~~

CVC4 incremental mode throws out a fatal failure:
~~~~
sat
sat
sat
Fatal failure within CVC4::Node CVC4::TheoryEngine::getExplanationAndRecipe(CVC4::TNode, CVC4::LemmaProofRecipe*) at /home/peisen/test/tofuzz/CVC4/src/theory/theory_engine.cpp:1728
Check failure

 d_propagationMap.find(toExplain) != d_propagationMap.end()
~~~~

OS: Ubuntu 16.04
Commit: 7d3df3b642

4tXJ7f:
Fixes #3971 and fixes #3991. In incremental mode, the logic can change from one
`(check-sat)` call to another. In the reported issue, we start with `QF_NIA`
but then switch to `QF_UFNIA` because there is a `div` term (which has a UF in
its expanded form). Dealing with this issue is challenging in general. As a
result, we have decided not to allow theory widening in
`Theory::expandDefinitions()` anymore but instead to do it eagerly in
`SmtEngine::setDefaults()`.


ajreynol:
Thanks for tracking this down.  I think the more robust solution is to *always* to expand to UFNIRA when using NIRA, and not have arithmetic be responsible for widening the logic. 

This would then mean that sharing is always enabled when non-linear arithmetic is.  This is similar to what is done for strings here:  https://github.com/CVC4/CVC4/blob/master/src/smt/smt_engine.cpp#L1590.

That way hopefully some of this LogicRequest stuff can be cleaned up.

4tXJ7f:
As discussed offline, I benchmarked the impact of this on QF_NIA and QF_FP (production build with assertions, 60s timeout):
```
 DIRECTORY |                                                 master |                                                fix3991 |
 Benchmark |  Stat  Slvd   Tot    To Mo Err   Cpu[s]   Mem[MB] Uniq |  Stat  Slvd   Tot    To Mo Err   Cpu[s]   Mem[MB] Uniq |
        FP |    ok 40186 40300   114  0   0   8502.7   92199.4   30 |    ok 40173 40300   127  0   0   9148.5   95739.2   17 |
       NIA |    ok  6522 23876 11586  0   0 826473.7  980869.3 1448 |    ok  6481 23876 11723  0   0 836255.0 1036156.4 1407 |
  __totals |    ok 46708 64176 11700  0   0 834976.4 1073068.7 1478 |    ok 46654 64176 11850  0   0 845403.4 1131895.6 1424 |
```

src/smt/smt_engine.cpp
```diff
@@ -1603,12 +1605,57 @@ void SmtEngine::setDefaults() {
...
       log.enableTheory(THEORY_UF);
       d_logic = log;
       d_logic.lock();
     }
   }
+  if (d_logic.isTheoryEnabled(THEORY_ARITH) && !d_logic.isLinear())
+  {
+    // Non-linear arithmetic requires UF to deal with division/mod because
+    // their expansion introduces UFs for the division/mod-by-zero case.
+    if (!d_logic.isTheoryEnabled(THEORY_UF))
```
ajreynol: Can we introduce a flag `needsUf` and then only do this Logic setting once, also with a string for the reason?
This is similar to what I did for disabling models: https://github.com/CVC4/CVC4/blob/master/src/smt/smt_engine.cpp#L2283

src/theory/theory_engine.cpp
```diff
@@ -1297,8 +1297,8 @@ void TheoryEngine::assertToTheory(TNode assertion, TNode originalAssertion, theo
     return;
   }
 
-  // If sharing is disabled, things are easy
```
ajreynol: You can keep this comment

PR #4054, opened 2020-03-12
Convert most instances of dataypes in parsers to the new API

ajreynol:
I am still accessing Expr-level Datatypes for the sygus v1/v2 parsers (within smt2). This will be addressed in two ways in the future:
(1) The sygus v1 parser will be deleted,
(2) The sygus v2 parser will be updated to use a "Grammar" object as an extension of the new API, which will hide all calls to the underlying datatype.  (See https://github.com/abdoo8080/CVC4/tree/sygus-api).  FYI @abdoo8080 .

Note I've renamed "mkMutualDatatypeTypes" to "bindMutualDatatypeTypes" to be more accurate and follow the updated name conventions in the parser.

The next step will be to handle parametric datatypes, which are not specifically addressed by this PR.


src/parser/parser.h
```diff
@@ -592,19 +592,16 @@ class CVC4_PUBLIC Parser {
...
-   *  then if doOverload is true, we create overloaded operators.
-   *  else if doOverload is false, the existing expression is shadowed by the new expression.
+   * Creates and binds sorts of a list of mutually-recursive datatype
+   * declarations.
    *
-   * flags specify information about the datatype, e.g. whether it should be
-   * printed out as a definition in models or not
-   *   (see enum in expr_manager_template.h).
+   * For each symbol defined by the datatype, if a symbol with name already
+   * exists, then if doOverload is true, we create overloaded operators. else if
```
mpreiner:
```suggestion
   * exists, then if doOverload is true, we create overloaded operators. Else, if
```

PR #4058, opened 2020-03-13
Bv2int fail on demand

yoni206:
This PR postpones as much as possible failures in the bv-to-int preprocessing pass.
Only when a node is (re-)constructed and the operator is not supported by the translation performed by the preprocessing pass, we throw an exception. Otherwise, the node stays the same.
Adresses this: https://github.com/CVC4/cvc4-projects/issues/129


src/preprocessing/passes/bv_to_int.cpp
```diff
@@ -729,38 +730,57 @@ Node BVToInt::bvToInt(Node n)
...
-              break;
+              else {
+                translated_children.insert(translated_children.begin(), intUF);
+                // Insert the term to the cache
+                d_bvToIntCache[current] =
+                    d_nm->mkNode(kind::APPLY_UF, translated_children);
+                /**
+                 * Add range constraints if necessary.
+                 * If the original range was a BV sort, the current application of
+                 * the fucntion Must be within the range determined by the
```
ajreynol: fucntion -> function

PR #4078, opened 2020-03-13
DecisionEngine: Use single unique pointer for ITE strategy .

aniemetz:
Previously, DecisionEngine maintained a vector of ITE strategies.
However, only one was ever created. This uses a single unique_ptr member
for the ITE strategy instead of a vector.

### DIFF
--- a/src/decision/decision_engine.cpp
+++ b/src/decision/decision_engine.cpp
@@ -26,12 +26,12 @@ using namespace std;
 namespace CVC4 {
 
 DecisionEngine::DecisionEngine(context::Context* sc, context::UserContext* uc)
-    : d_enabledITEStrategies(),
+    : d_enabledITEStrategy(nullptr),
       d_needIteSkolemMap(),
-      d_relevancyStrategy(NULL),
+      d_relevancyStrategy(nullptr),
       d_assertions(uc),
-      d_cnfStream(NULL),
-      d_satSolver(NULL),
+      d_cnfStream(nullptr),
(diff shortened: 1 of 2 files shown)


src/decision/decision_engine.cpp
```diff
@@ -65,11 +65,22 @@ void DecisionEngine::shutdown()
...
-  d_enabledITEStrategies.clear();
+  d_enabledITEStrategy.reset(nullptr);
   d_needIteSkolemMap.clear();
 }
 
+SatLiteral DecisionEngine::getNext(bool& stopSearch)
+{
+  NodeManager::currentResourceManager()->spendResource(
+      ResourceManager::Resource::DecisionStep);
+  Assert(d_cnfStream != NULL) << "Forgot to set cnfStream for decision engine?";
```
4tXJ7f:
```suggestion
  Assert(d_cnfStream != nullptr) << "Forgot to set cnfStream for decision engine?";
```

src/decision/decision_engine.cpp
```diff
@@ -65,11 +65,22 @@ void DecisionEngine::shutdown()
...
+  Assert(d_satSolver != NULL) << "Forgot to set satSolver for decision engine?";
```
4tXJ7f:
```suggestion
  Assert(d_satSolver != nullptr) << "Forgot to set satSolver for decision engine?";
```

4tXJ7f:
LGTM, two minor suggestions for consistency with the other changes.

PR #4079, opened 2020-03-13
Int2BV fail on demand

yoni206:
This PR delays error on unsupported symbols as much as possible, by only throwing the error when actually constructing the node.


ajreynol:
Btw it looks like travis is failing due to the added regression.

ajreynol:
Can we proceed with this? I would recommend turning off eager bitblasting in this and in the SMT COMP current script for now.  https://github.com/CVC4/CVC4/blob/master/contrib/competitions/smt-comp/run-script-smtcomp-current#L42

yoni206:
@ajreynol :
done. looks ok?

PR #4080, opened 2020-03-13
Expand the definition of a "simple" farkas proof.

alex-ozdemir:
Before, a "simple farkas proof" was one that just applied the farkas
lemma.

Now it allows for the antecedents to (optionally) be tightened.

Note that hasSimpleFarkasProof was (and is) dead code.

We will use it to decide whether to print a proof or a hole.

### DIFF
--- a/src/proof/arith_proof.cpp
+++ b/src/proof/arith_proof.cpp
@@ -1156,8 +1156,10 @@ void LFSCArithProof::printTheoryLemmaProof(std::vector<Expr>& lemma,
   }
   else
   {
+    AlwaysAssert(false) << "arith hole!";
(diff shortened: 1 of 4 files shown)


yoni206:
Is there any way to test this? Ideally, one test for an assumption, one for tightened assumption, and one where we print a hole (or one test with all three)?

alex-ozdemir:
We're not quite at the stage where we can print a hole, but the others we can test using assertions.
I'll make this change tomorrow morning!

alex-ozdemir:
I put in an assertion which failed on holes, and have now removed it. If everthing passes, we're good to go.

PR #4111, opened 2020-03-18
Enable configuration of Python path

4tXJ7f:
This commit adds the argument `--python-dir=PATH` to the `configure.sh`
script. This allows users to specify where to look for the Python
binary. This is useful when users want to use Python installed in a
virtualenv.


4tXJ7f:
We have discussed this offline and came to the conclusion that it is better to change the Homebrew formula to prepand all `system` calls with activating the virtual environment first. Closing this PR for now.

PR #4121, opened 2020-03-19
Fix variable shadowing issue in sygus-inference

### LINKED ISSUE #4083: miniscope-quant sygus-inference var-ineq-elim-quant fatal failure at src/expr/node_algorithm.cpp:338
Hi,
For this formula:
```
(set-option :miniscope-quant true)
(set-option :sygus-inference true)
(set-option :var-ineq-elim-quant false)
(declare-fun b ( Int ) Bool)
(assert (forall (( c Int ) ( d Int )) (and (> d 3 ) (xor (>= c d) (b c)))))  
(check-sat)             
```
```
Fatal failure within bool CVC4::expr::getFreeVariables(CVC4::TNode, std::unordered_set<CVC4::NodeTemplate<true>, CVC4::NodeHashFunction>&, bool) at src/expr/node_algorithm.cpp:338
Check failure

 bound_var.find(cn) == bound_var.end()

Aborted
```

OS: Ubuntu 18.04
Commit: 442ab0c

ajreynol:
This makes the `sygus-inference` preprocessing pass avoid variable shadowing, which technically could happen by forcing unexpected options.

Fixes #4083.


ajreynol:
`(set-info :status unsat)` is enough actually.  This suffices if you don't have command line options or non-standard output, where notice the regression is using options via commands.

PR #4127, opened 2020-03-20
Collect statistics about normal form inferences

4tXJ7f:
This commit adds code to count the number of inferences made of each
inference type for normal form inferences. It extends the `Inference`
enum in `infer_info.h` and adds two new `sendInference()` methods in the
`InferenceManager` to send and count inferences that have a corresonding
entry in the `Inference` enum.


ajreynol:
This looks great. 

Can you delete the old statistics class from TheoryStrings while we're at it?  https://github.com/CVC4/CVC4/blob/master/src/theory/strings/theory_strings.h#L372

One minor question about the design: wondering if the statistics class should be a standalone utility that is passed to InferenceManager and lives in TheoryStrings. This is simliar to what I did recently for SyGuS, here is the standalone class: https://github.com/CVC4/CVC4/blob/master/src/theory/quantifiers/sygus/sygus_stats.h 

Concerning global stats class vs local one,as an advantage, all statistics would be done in one place, and as a disadvantage there is less encapsulation.  

I tend to favor the global option (since we don't have to make little stats classes everywhere), although it's debatable which is best in this case.

ajreynol:
Can you post an example output of the stats?

src/theory/strings/core_solver.cpp
```diff
@@ -952,7 +952,7 @@ void CoreSolver::processNEqc(std::vector<NormalForm>& normal_forms)
         {
           return;
         }
-        else if (!pinfer.empty() && pinfer.back().d_id == 1)
+        else if (!pinfer.empty() && pinfer.back().d_id == Inference::INFER_EMP)
```
ajreynol: There is an issue here, this is actually an optimization that says effectively "If the inference has a minimal id, then we cannot find a better inference, so break". This changes the sematics of this optimization. Maybe we can just delete this or otherwise fix this in a clear way.

src/theory/strings/core_solver.cpp
```diff
@@ -981,7 +981,7 @@ void CoreSolver::processNEqc(std::vector<NormalForm>& normal_forms)
   bool set_use_index = false;
   Trace("strings-solve") << "Possible inferences (" << pinfer.size()
                          << ") : " << std::endl;
-  unsigned min_id = 9;
+  Inference min_id = Inference::NONE;
```
ajreynol: We should perhaps move NONE to the end of the enum just to make sure its clear that the heuristic on order inferences is consistent.  Right now, its weird setting min_id to NONE=0 initially.

4tXJ7f:
> Can you post an example output of the stats?

For `regress2/strings/non_termination_regular_expression6.smt2` the output is:

`theory::strings::inferences, "[(N_Unify : 2), (N_EndpointEq : 15), (S-Split(VAR)-prop : 1), (Len-Split(Len) : 9), (Len-Split(Emp) : 11), (S-Split(CST-P) : 18), (S-Split(VAR) : 1)]"`

I'll address the other points.

src/theory/strings/core_solver.cpp
4tXJ7f: As discussed offline, I've investigated the impact of this and it seems negligible. I have submitted #4129 to remove this heuristic.

src/theory/strings/core_solver.cpp
4tXJ7f: I have moved it to the end (note that this does not change the semantics of the code because we have `set_use_index` to detect whether or not the values have been set).

4tXJ7f:
After an offline discussion, I have moved the statistics to a separate file and made them a member of `TheoryStrings` as suggested.

src/theory/strings/inference_manager.cpp
```diff
@@ -186,13 +186,33 @@ void InferenceManager::sendInference(const std::vector<Node>& exp,
...
 }
 
-void InferenceManager::sendInference(const InferInfo& i)
+void InferenceManager::sendInference(const std::vector<Node>& exp,
+                                     const std::vector<Node>& exp_n,
+                                     Node eq,
+                                     Inference infer,
+                                     bool asLemma)
+{
+  d_parent.getStatistics()->d_inferences << infer;
```
ajreynol: Actually I'd like to move away from using parent, can we pass a reference to stats and then use it directly (similar to how we pass other utilities like EqualityEngine/SolverState/etc.?).

4tXJ7f: Sounds good, done.

PR #4128, opened 2020-03-20
Do not require that function sorts are first class internally

ajreynol:
This PR removes the requirement in the NodeManager that argument types to the function sort are first class. 

Notice that the new API does this check (as it should): https://github.com/CVC4/CVC4/blob/master/src/api/cvc4cpp.cpp#L2633

Moreover, SyGuS v2 internally requires constructing function types having arguments that are not first class (e.g. regular expression type).  This is required to update the regression https://github.com/CVC4/CVC4/blob/master/test/regress/regress1/sygus/re-concat.sy to SyGuS v2.

FYI @abdoo8080 .


HanielB:
Aren't there any places in the code that are calling `mkFunctionType` and not doing this check themselves?

ajreynol:
Yes, that is the point of this PR.  Those places in the code should not do this check.  This should only apply to user-defined sorts e.g. coming from `declare-fun`.

PR #4131, opened 2020-03-21
Restrict cases of sygus grammar reduction based on argument variants

ajreynol:
An assertion is triggered for some V2 versions of SyGuS V1 benchmarks during sygus grammar reduction.  This PR updates this technique so that it only applies to sygus constructors whose sygus operators are lambdas of the form 
`lambda x1 ... xn. f( x1 ... xn )`.

FYI @abdoo8080 .


src/theory/quantifiers/sygus/sygus_grammar_red.cpp
```diff
@@ -60,7 +60,16 @@ void SygusRedundantCons::initialize(QuantifiersEngine* qe, TypeNode tn)
     std::vector<Node> glist;
     if (sop.isConst() || sop.getKind() == LAMBDA)
     {
-      Assert(g.getNumChildren() == dt[i].getNumArgs());
+      // g should have the same number of children as there are arguments to
+      // the datatype, if the sygus operator is a lambda.
+      // This is not the case for constants, since those that are represented by
+      // AST (e.g. datatypes) are such that g may have >0 children but dt[i]
+      // has no arguments.
```
HanielB: This is not very clear. Could you expand on the example?

ajreynol: I've revised the policy to be clearer now.  (The optimization I was trying to preserve only was applicable to grammars coming from V1 inputs).

HanielB: Looks good, thanks!

PR #4140, opened 2020-03-22
Infer when sygus operators are equivalent to builtin kinds

ajreynol:
The V2 parser always turns sygus operators into lambdas for consistency. This PR ensures that we nevertheless infer when a sygus operator is equivalent to a builtin operator, e.g.
`(lambda x y. (+ x y))` is equivalent to `+`.

The main reason this is required is to ensure that solution reconstruction heuristics work optimally when we make the switch SyGuS V1 -> V2 (see 5 failing benchmarks due to `--cegqi-si=all` on https://github.com/CVC4/CVC4/pull/4136).


src/theory/quantifiers/sygus/type_info.cpp
```diff
@@ -128,7 +122,32 @@ void SygusTypeInfo::initialize(TermDbSygus* tds, TypeNode tn)
             << "In sygus datatype " << dt.getName()
             << ", argument to a lambda constructor is not " << lat << std::endl;
       }
-      if (sop[0].getKind() == ITE)
+      // See if it is a builtin kind, possible if the operator is of the form:
+      // lambda x1 ... xn. f( x1, ..., xn ) and f is not a parametrized kind
+      // (e.g. APPLY_UF).
```
HanielB: clarify that APPLY_UF is a parameterized kind. Also fix "parametrized" above.

PR #4144, opened 2020-03-24
Restrict partial triggers to standard quantified formulas

### LINKED ISSUE #4143: cbqi partial-triggers pure-th-triggers relational-triggers sygus-inference fatal failure at src/theory/quantifiers/sygus/ce_guided_single_inv.cpp:390
Hi,
For this formula:
```
(set-option :cbqi false)
(set-option :partial-triggers true)
(set-option :pure-th-triggers true)
(set-option :relational-triggers true)
(set-option :sygus-inference true)
(declare-const b Int)
(declare-const c Int)
(assert (or (= (- b c) 2) (= b 3)))
(check-sat)
```
```
Fatal failure within bool CVC4::theory::quantifiers::CegSingleInv::solve() at src/theory/quantifiers/sygus/ce_guided_single_inv.cpp:390
Check failure

 qs.size() <= 1

Aborted
```

OS: Ubuntu 18.04
Commit: c9b7c3d

ajreynol:
Fixes #4143.

### DIFF
--- a/src/theory/quantifiers/ematching/inst_strategy_e_matching.cpp
+++ b/src/theory/quantifiers/ematching/inst_strategy_e_matching.cpp
@@ -368,7 +368,9 @@ void InstStrategyAutoGenTriggers::generateTriggers( Node f ){
       // quantified formulas (those with only two children), since this
       // technique should not be used for e.g. quantifiers marked for
       // quantifier elimination.
-      if (options::partialTriggers() && f.getNumChildren() == 2)
+      QAttributes qa;
+      QuantAttributes::computeQuantAttributes( f, qa );
+      if (options::partialTriggers() && qa.isStandard())
       {
         std::vector< Node > vcs[2];
         for( unsigned i=0; i<f[0].getNumChildren(); i++ ){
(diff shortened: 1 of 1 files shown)


src/theory/quantifiers/ematching/inst_strategy_e_matching.cpp
```diff
@@ -360,7 +360,16 @@ void InstStrategyAutoGenTriggers::generateTriggers( Node f ){
...
-      if( options::partialTriggers() ){
+      // Invoke partial trigger strategy: partition variables of quantified
+      // formula into (X,Y) where X are contained in a trigger and Y are not.
+      // We then force a split of the quantified formula so that it becomes:
+      //   forall X. forall Y. P( X, Y )
+      // and hence is treatable by E-matching. We only do this for "standard"
+      // quantified formulas (those with only two children), since this
+      // technique should not be used for e.g. quantifiers marked for
+      // quantifier elimination.
+      if (options::partialTriggers() && f.getNumChildren() == 2)
```
HanielB: It'd be nice if you could abstract this test of being an "standard quantifier", what do you think? It might improve the code's legibility.

ajreynol: Actually this has been done already: https://github.com/CVC4/CVC4/blob/master/src/theory/quantifiers/quantifiers_attributes.h#L157
although I need to double check what cases a quantified formula can be standard while having 3 children.  (This code was being overly cautious/simple).

ajreynol: I addressed this in my latest commits.

HanielB: Great, thanks!

PR #4150, opened 2020-03-25
Care graphs based on polymorphic operators in strings

ajreynol:
Towards theory of sequences.

To compute a care graphs, we need to group function applications by the string type they are associated with.  This PR introduces a utility to function to get the "owning string type" for a function application, and updates the care graph computation so that it partitions function applications according to this type.


src/theory/strings/theory_strings.cpp
```diff
@@ -889,16 +891,19 @@ void TheoryStrings::computeCareGraph(){
       }
     }
     if( has_trigger_arg ){
-      index[op].addTerm( f1, reps );
+      TypeNode ft = utils::getOwnerStringType(f1);
+      std::pair<TypeNode, Node> ikey = std::pair<TypeNode, Node>(ft, op);
```
4tXJ7f: Just a minor simplification.

```suggestion
      std::pair<TypeNode, Node> ikey(ft, op);
```

PR #4156, opened 2020-03-26
Split transcendental solver to its own file

ajreynol:
Towards refactoring non-linear arithmetic.  This moves all things related to transcendental functions to their own file.  It also moves common utils to the utility file.

There are no major code changes in this commit.


src/theory/arith/nonlinear_extension.cpp
```diff
@@ -1008,126 +796,19 @@ int NonlinearExtension::checkLastCall(const std::vector<Node>& assertions,
...
   }
 
+  // initialize the trancendental function solver
+  std::vector<Node> lemmas;
+  d_trSlv.initLastCall(assertions, false_asserts, xts, lemmas, lemsPp);
+
+  // process lemmas that may have been generated by the transcendental solver
   filterLemmas(lemmas, lems);
-  if (!lems.empty())
+  if (!lems.empty() || !lemsPp.empty())
```
ahmed-irfan: why is it different now?

src/theory/arith/nonlinear_extension.cpp
```diff
@@ -1008,126 +796,19 @@ int NonlinearExtension::checkLastCall(const std::vector<Node>& assertions,
...
-    Trace("nl-ext-lemma") << "NonlinearExtension::Lemma : purify : " << lem
-                          << std::endl;
-    lemsPp.push_back(lem);
-  }
-  if (!lemsPp.empty())
-  {
-    Trace("nl-ext") << "  ...finished with " << lemsPp.size()
-                    << " new lemmas SINE phase shifting." << std::endl;
-    return lemsPp.size();
+    return lems.size() + lemsPp.size();
```
ahmed-irfan: why are we summing them up? aren't lems and lemsPp sematically the same lemmas -- the only difference is the preprocessing, right?

ajreynol: `lems` and `lemsPp` are two independent sets of lemmas.

The code was modified slightly here. Previously, transcendental function handling in this block of code generated lems and then sent out lemmas if any were generated. If no lemmas were generated, it proceeded to generate a new set of lemmas in lemsPp.  Since the transcendental function solver does not have access to the output channel, these two steps are combined into one return case here.

ahmed-irfan: ok, got it

PR #4157, opened 2020-03-26
Disable slow regression

ajreynol:
Should fix timeout in asan build.


test/regress/CMakeLists.txt
```diff
@@ -2388,6 +2387,8 @@ set(regression_disabled_tests
   regress1/sygus/array_search_2.sy
   regress1/sygus/array_sum_2_5.sy
   regress1/sygus/crcy-si-rcons.sy
+  # currently slow
```
mpreiner: Maybe add a commit to document since when it's slow? The nigthlies started failing on the 24th.

PR #4158, opened 2020-03-26
Convert more uses of string-specific functions

ajreynol:
Towards theory of sequences.


src/theory/strings/word.cpp
```diff
@@ -76,6 +76,29 @@ size_t Word::getLength(TNode x)
...
+    NodeManager* nm = NodeManager::currentNM();
+    std::vector<unsigned> ccVec;
+    const std::vector<unsigned>& cvec = x.getConst<String>().getVec();
+    for (unsigned i = 0, size = cvec.size(); i < size; i++)
+    {
+      ccVec.clear();
+      ccVec.insert(ccVec.end(), cvec.begin() + i, cvec.begin() + i + 1);
+      Node ch = nm->mkConst(String(ccVec));
+      ret.push_back(ch);
+    }
```
4tXJ7f: It seems to me as if this could be simplified to:

```suggestion
    for (unsigned chVal : x.getConst<String>().getVec())
    {
      ccVec.clear();
      ccVec.push_back(chVal);
      Node ch = nm->mkConst(String(ccVec));
      ret.push_back(ch);
    }
```

src/theory/strings/word.cpp
```diff
@@ -76,6 +76,29 @@ size_t Word::getLength(TNode x)
...
+    return ret;
+  }
+  Unimplemented();
+  std::vector<Node> ret;
+  return ret;
```
4tXJ7f: Strictly speaking, those two lines are not needed because `Unimplemented()` never returns (and is marked as such). Personal preference though.

PR #4159, opened 2020-03-27
Fix issues with unsat cores and reset-assertions

### LINKED ISSUE #4151: Fatal failure with :produce-unsat-assumptions and reset-assertions.
CVC4 crashes with the following error message on the input below:
```
sat
Fatal failure within void CVC4::ProofManager::initSatProof(CVC4::Minisat::Solver*) at /path/to/src/proof/proof_manager.cpp:145
Check failure

 currentPM()->d_satProof == __null

Aborted (core dumped)
```
Input:
```
(set-logic ALL)
(set-option :produce-unsat-assumptions true)
(check-sat)
(reset-assertions)
```

4tXJ7f:
Fixes #4151. Commit e9f4cec2cad02e270747759223090c16b9d2d44c fixed how
`(reset-assertions)` is handled by destroying and recreating the
`PropEngine` owned by `SmtEngine`. When unsat cores are enabled,
creating a `PropEngine` triggers the creation of a SAT proof and a CNF
proof. In the `ProofManager`, we had assertions that checked that those
kinds of proofs were only created once, which is not true anymore. This
commit removes the assertions, cleans up the memory management in
`ProofManager` to use `std::unique_ptr` and makes all the
`ProofManager::init*` methods non-static for consistency.

The commit also fixes an additional issue that I encountered while
testing the fix: When creating the new `PropEngine`, we were not
asserting `true` and `(not false)`, which lead to an error if we tried
to get the unsat core after a `(reset-assertion)` command and we had
asserted `(assert false)`. The commit fixes this by asserting `true` and
`(not false)` in the constructor of `PropEngine`.

The regression test is an extension of the example in #4151 and covers
both issues.


src/proof/proof_manager.cpp
```diff
@@ -89,26 +85,29 @@ const Proof& ProofManager::getProof(SmtEngine* smt)
...
   }
   return *(currentPM()->d_fullProof);
 }
 
-CoreSatProof* ProofManager::getSatProof() {
+CoreSatProof* ProofManager::getSatProof()
+{
   Assert(currentPM()->d_satProof);
-  return currentPM()->d_satProof;
+  return currentPM()->d_satProof.get();
```
aniemetz: I'd prefer having a function `ProofManager::getSatProof` as a wrapper around this.

src/proof/proof_manager.cpp
```diff
@@ -89,26 +85,29 @@ const Proof& ProofManager::getProof(SmtEngine* smt)
...
 }
 
-CnfProof* ProofManager::getCnfProof() {
+CnfProof* ProofManager::getCnfProof()
+{
   Assert(currentPM()->d_cnfProof);
-  return currentPM()->d_cnfProof;
+  return currentPM()->d_cnfProof.get();
```
aniemetz: Ditto

src/proof/proof_manager.cpp
```diff
@@ -89,26 +85,29 @@ const Proof& ProofManager::getProof(SmtEngine* smt)
...
 }
 
-TheoryProofEngine* ProofManager::getTheoryProofEngine() {
+TheoryProofEngine* ProofManager::getTheoryProofEngine()
+{
   Assert(currentPM()->d_theoryProof != NULL);
-  return currentPM()->d_theoryProof;
+  return currentPM()->d_theoryProof.get();
```
aniemetz: Ditto

src/proof/proof_manager.h
```diff
@@ -180,9 +180,8 @@ class ProofManager {
 
   // initialization
   void         initSatProof(Minisat::Solver* solver);
```
aniemetz: Can you format this line?

PR #4163, opened 2020-03-27
Refactor: Fix -Wshadow warnings in Cython-generated python binding code

makaimann:
There are several code patterns in Cython that produce code that gives a `-Wshadow` warning. This refactor avoids those patterns. Some patterns that seem to cause this warning include:
* the use of `yield` -- used in `__iter__` to return a generator
* format strings with arguments that are not strings or ints
* using deque from python collections
* passing a list to a function and returning a modified list (couldn't find a workaround for this one so I inlined it)

For other developers editing the Cython bindings, I added this information to the Style guidelines in comments at the top of `cvc4.pxi`

### DIFF
--- a/src/api/python/CMakeLists.txt
+++ b/src/api/python/CMakeLists.txt
@@ -27,7 +27,7 @@ add_custom_target(
   DEPENDS
     genkinds.py
   COMMENT
-    "Generate cvc4kinds.{pxd,pyx}"
+    "Generate cvc4kinds.{pxd,pxi}"
 )
 
 add_cython_target(pycvc4 CXX)
--- a/src/api/python/cvc4.pxi
+++ b/src/api/python/cvc4.pxi
@@ -1074,7 +1074,6 @@ cdef class Sort:
 
 cdef class Term:
     cdef c_Term cterm
-    # cdef deque[c_Term] children
     def __cinit__(self):
         # cterm always set in the Solver object
         pass
--- a/src/api/python/cvc4.pxi
+++ b/src/api/python/cvc4.pxi
@@ -403,7 +403,6 @@ cdef class Solver:
         sort.csort = self.csolver.mkParamSort(symbolname.encode())
         return sort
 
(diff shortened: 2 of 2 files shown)


makaimann:
@aniemetz would you mind building this and checking for `-Wshadow` warnings? When you brought it up with me, you gave a trace like this:
```
../../../../src/api/python/cvc4.pxi: In function 'PyObject* __pyx_pf_6pycvc4_6Solver_50mkReal(__pyx_obj_6pycvc4_Solver*, PyObject*, PyObject*)':
../../../../src/api/python/cvc4.pxi:1201:53: warning: declaration of '__pyx_tstate' shadows a previous local [-Wshadow]
../../../../src/api/python/cvc4.pxi:558:13: note: in expansion of macro '__Pyx_PyThreadState_declare'
  558 |             except Exception as e:
      |             ^~~~~~~~~~~~~~~~~~~~~~     
../../../../src/api/python/cvc4.pxi:1201:53: note: shadowed declaration is here
../../../../src/api/python/cvc4.pxi:556:7: note: in expansion of macro '__Pyx_PyThreadState_declare'
  556 |             try:
      |       ^     ~~~~
```

Unfortunately, I don't get that trace on either my laptop or desktop. Instead what I got (and this PR fixes) is this:
```
cython_utility:1741:19: warning: declaration shadows a local variable [-Wshadow]
        PyObject *args[2] = {arg1, arg2};
                  ^
cython_utility:1738:15: note: previous declaration is here
    PyObject *args, *result = NULL;
```

mpreiner:
I don't think that we have to get rid of the `-Wshadow` warnings in the generated Cython code. We can't really control what code it generates. How about we disable the `-Wshadow` flag for the code generated by Cython?

aniemetz:
I agree. This is also more robust in the future (since we don't have control over it).

makaimann:
Sounds good to me, how about I just open a separate PR for that.

PR #4166, opened 2020-03-27
turning class into namespace

yoni206:
The abduction module is based on a class that has no fields and includes a single function which is static. 
According to my understanding of [the relevant part of the coding guidelines](https://github.com/CVC4/CVC4/wiki/Code-Style-Guidelines#nonmember-static-member-and-global-functions), this module should be in a namespace rather than a class. 
Am I correct?

### DIFF
--- a/src/theory/quantifiers/sygus/sygus_abduct.h
+++ b/src/theory/quantifiers/sygus/sygus_abduct.h
@@ -80,9 +80,9 @@ Node mkAbductionConjecture(const std::string& name,
                                   const std::vector<Node>& axioms,
                                   TypeNode abdGType);
 
+}  // namespace sygus_abduct
 }  // namespace quantifiers
 }  // namespace theory
 }  // namespace CVC4
-}  // namespace sygus_abduct
 
 #endif /* CVC4__THEORY__QUANTIFIERS__SYGUS_ABDUCT_H */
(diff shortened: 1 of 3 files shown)


src/theory/quantifiers/sygus/sygus_abduct.h
```diff
@@ -53,38 +55,34 @@ namespace quantifiers {
...
+ */
+Node mkAbductionConjecture(const std::string& name,
+                                  const std::vector<Node>& asserts,
+                                  const std::vector<Node>& axioms,
+                                  TypeNode abdGType);
 
 }  // namespace quantifiers
 }  // namespace theory
 }  // namespace CVC4
+}  // namespace sygus_abduct
```
Ying1123: Shouldn't this be before the line }  // namespace quantifiers?

yoni206: fixed in latest commit. thanks!

ajreynol:
@4tXJ7f  has argued for the opposite, for instance this PR : https://github.com/CVC4/CVC4/pull/3797 we decided to change a namespace to a class.  Perhaps @4tXJ7f has an opinion here?
Personally, I prefer the class over the namespace.

aniemetz:
@ajreynol I disagree. In this particular case, the class doesn't even have members, I like this solution better.

mpreiner:
@yoni206 @ajreynol @aniemetz We can soon start baking a birthday cake for this PR. What's the status?

yoni206:
I'm fine with closing it. From the discussion, it seems like differences in
preference between @aniemetz and @ajreynol about classes vs. namespaces
when there is a single method.

PR #4167, opened 2020-03-28
Support indexed operators re.loop and re.^

### LINKED ISSUE #4161: Fatal failure at theory/strings/sequences_rewriter.cpp:1199 & 1217
```
% cvc4-fa2ba76 --quiet bug.smt2
Fatal failure within static CVC4::Node 
CVC4::theory::strings::SequencesRewriter::rewriteLoopRegExp(CVC4::TNode) at  
/src/theory/strings/sequences_rewriter.cpp:1199
Check failure
 n1.getConst<Rational>().sgn() >= 0
Negative integer in string REGEXP_LOOP (1)
Aborted
% cat bug.smt2
(declare-fun a () String)  
(assert (str.in.re a (re.loop (str.to.re "") (- 4))))      
(check-sat)
```
OS: Ubuntu 18.04
Commit: fa2ba76

ajreynol:
Towards support for the strings standard.

This modifies our interface so that we accept the SMT-LIB standard versions of re.loop and re.^.  This means re.loop no longer accepts 3 arguments but 1 (with 2 indices).

This means we no longer accept re.loop with only a lower bound and no upper bound on the number of repetitions.

Also fixes #4161.


src/util/regexp.h
```diff
@@ -0,0 +1,95 @@
...
+  {
+    return d_repeatAmount == r.d_repeatAmount;
+  }
+};
+
+struct CVC4_PUBLIC RegExpLoop
+{
+  unsigned d_loopAmountLo;
+  unsigned d_loopAmountHi;
+  RegExpLoop(unsigned l, unsigned h) : d_loopAmountLo(l), d_loopAmountHi(h) {}
```
4tXJ7f:
```suggestion
  uint32_t d_loopAmountLo;
  uint32_t d_loopAmountHi;
  RegExpLoop(uint32_t l, uint32_t h) : d_loopAmountLo(l), d_loopAmountHi(h) {}
```

src/util/regexp.h
```diff
@@ -0,0 +1,95 @@
...
+{
+  return os << r.d_repeatAmount;
+}
+
+inline std::ostream& operator<<(std::ostream& os,
+                                const RegExpLoop& bv) CVC4_PUBLIC;
+inline std::ostream& operator<<(std::ostream& os, const RegExpLoop& r)
+{
+  return os << "[" << r.d_loopAmountLo << ":" << r.d_loopAmountHi << "]";
+}
```
4tXJ7f:
```suggestion
std::ostream& operator<<(std::ostream& os,
                                const RegExpRepeat& bv) CVC4_PUBLIC;
std::ostream& operator<<(std::ostream& os, const RegExpRepeat& r)
{
  return os << r.d_repeatAmount;
}

std::ostream& operator<<(std::ostream& os,
                                const RegExpLoop& bv) CVC4_PUBLIC;
std::ostream& operator<<(std::ostream& os, const RegExpLoop& r)
{
  return os << "[" << r.d_loopAmountLo << ":" << r.d_loopAmountHi << "]";
}
```

Can we move the implementation of those to the source file? I am actually a bit surprised that this compiles when just including `iosfwd`.

src/theory/strings/theory_strings_utils.h
```diff
@@ -153,6 +153,14 @@ bool isStringKind(Kind k);
  */
 TypeNode getOwnerStringType(Node n);
 
+/* Get the number of repetitions for a regexp repeat node */
+unsigned getRepeatAmount(TNode node);
+
+/* Get the upper index of given regexp loop node. */
+unsigned getLoopHigh(TNode node);
+/* Get the lower index of given regexp loop node. */
+unsigned getLoopLow(TNode node);
```
4tXJ7f:
```suggestion
/* Get the number of repetitions for a regexp repeat node */
uint32_t getRepeatAmount(TNode node);

/* Get the upper index of given regexp loop node. */
uint32_t getLoopHigh(TNode node);
/* Get the lower index of given regexp loop node. */
uint32_t getLoopLow(TNode node);
```

src/theory/strings/theory_strings_utils.cpp
```diff
@@ -293,6 +293,24 @@ TypeNode getOwnerStringType(Node n)
...
+
+unsigned getLoopHigh(TNode node)
+{
+  Assert(node.getKind()==REGEXP_LOOP);
+  return node.getOperator().getConst<RegExpLoop>().d_loopAmountHi;
+}
+
+unsigned getLoopLow(TNode node)
+{
+  Assert(node.getKind()==REGEXP_LOOP);
```
4tXJ7f: Clang-Format.

src/theory/strings/theory_strings_utils.h
4tXJ7f: I am not quite sure about the naming of `getLoopHigh`/`getLoopLow`. What do you think about `getLoopMaxOccurrences`/`getLoopMinOccurrences` or `getLoopMaxAmount`/`getLoopMinAmount` (and renaming the corresponding members of the operator accordingly)?

src/theory/strings/sequences_rewriter.cpp
```diff
@@ -1192,61 +1192,35 @@ Node SequencesRewriter::rewriteLoopRegExp(TNode node)
...
-      TypeNode rtype = nm->regExpType();
-      for (unsigned j = l; j < u; j++)
-      {
-        vec_nodes.push_back(r);
-        n = utils::mkConcat(vec_nodes, rtype);
-        vec2.push_back(n);
-      }
-      retNode = nm->mkNode(REGEXP_UNION, vec2);
-    }
+    retNode = n;
```
4tXJ7f: That's not quite correct according to the standard: If `u < l` then this should be rewritten to `re.none`.

src/theory/strings/sequences_rewriter.cpp
```diff
@@ -1192,61 +1192,35 @@ Node SequencesRewriter::rewriteLoopRegExp(TNode node)
...
-                  ? rest
-                  : vec_nodes.size() == 1
-                        ? nm->mkNode(REGEXP_CONCAT, r, rest)
-                        : nm->mkNode(REGEXP_CONCAT,
-                                     nm->mkNode(REGEXP_CONCAT, vec_nodes),
-                                     rest);
+    std::vector<Node> vec2;
+    vec2.push_back(n);
+    TypeNode rtype = nm->regExpType();
+    for (unsigned j = l; j < u; j++)
```
4tXJ7f:
```suggestion
    for (uint32_t j = l; j < u; j++)
```

src/theory/strings/regexp_operation.cpp
```diff
@@ -686,16 +680,16 @@ Node RegExpOpr::derivativeSingle( Node r, CVC4::String c ) {
         break;
       }
       case kind::REGEXP_LOOP: {
-        if(r[1] == r[2] && r[1] == d_zero) {
+        unsigned l = utils::getLoopLow(r);
+        unsigned u = utils::getLoopHigh(r);
```
```suggestion
        uint32_t l = utils::getLoopLow(r);
        uint32_t u = utils::getLoopHigh(r);
```

src/theory/strings/regexp_operation.cpp
```diff
@@ -501,18 +495,18 @@ int RegExpOpr::derivativeS( Node r, CVC4::String c, Node &retNode ) {
...
```
```suggestion
        uint32_t l = utils::getLoopLow(r);
        uint32_t u = utils::getLoopHigh(r);
```

src/theory/strings/regexp_operation.cpp
```diff
@@ -260,7 +252,9 @@ int RegExpOpr::delta( Node r, Node &exp ) {
         break;
       }
       case kind::REGEXP_LOOP: {
-        if(r[1] == d_zero) {
+        unsigned lo = utils::getLoopLow(r);
```
```suggestion
        uint32_t lo = utils::getLoopLow(r);
```

src/parser/cvc/Cvc.g
```diff
@@ -2072,8 +2072,11 @@ stringTerm[CVC4::api::Term& f]
     { f = MK_TERM(CVC4::api::REGEXP_OPT, f); }
   | REGEXP_RANGE_TOK LPAREN formula[f] COMMA formula[f2] RPAREN
     { f = MK_TERM(CVC4::api::REGEXP_RANGE, f, f2); }
-  | REGEXP_LOOP_TOK LPAREN formula[f] COMMA formula[f2] COMMA formula[f3] RPAREN
-    { f = MK_TERM(CVC4::api::REGEXP_LOOP, f, f2, f3); }
+  | REGEXP_LOOP_TOK LPAREN formula[f] COMMA lo=numeral COMMA hi=numeral RPAREN
+    {
+      api::Op lop = SOLVER->mkOp(CVC4::api::REGEXP_LOOP,lo,hi);
```
4tXJ7f:
```suggestion
      api::Op lop = SOLVER->mkOp(CVC4::api::REGEXP_LOOP, lo, hi);
```

src/theory/strings/theory_strings_utils.h
ajreynol: Sure.

ajreynol:
Strangely, the issue with the wrong semantics was present before this commit.  I will fix this.

src/util/regexp.h
```diff
@@ -19,31 +19,28 @@
 
 #include <cstdint>
 #include <iosfwd>
+#include <ostream>
```
4tXJ7f: This include should be moved to the source file for faster compilation.

PR #4168, opened 2020-03-28
Stop printing datatype declaration for Sygus V1 grammar.

abdoo8080:


src/parser/parser.h
```diff
@@ -273,6 +273,11 @@ class CVC4_PUBLIC Parser {
     return d_input;
   }
 
+  /** Get unresolved sorts */
+  inline std::set<api::Sort>& getUnresolvedSorts() {
```
ajreynol: Will need to clang-format this.

PR #4169, opened 2020-03-28
Use traversal iterators in IntToBv

alex-ozdemir:
This PR rips the traversal machinery out of Int-to-Bv, replacing it with traversal iterators.

Also, cleaned `childrenTypesChanged` a bit.

While basically I just cut out some lines, the diff is rather messy (I think the diffing tool doesn't like indentation changes).


alex-ozdemir:
Tip for viewing a nicer diff: use `git diff --ignore-space-change`, e.g.

    git diff --ignore-space-change $(git merge-base master int-to-bv-refac)

alex-ozdemir:
Ah! Apparently the github UI allows for this to. In the all-file diff view, you'll find an option in the gear menu:

image

src/preprocessing/passes/int_to_bv.cpp
```diff
@@ -147,30 +98,15 @@ Node intToBV(TNode n, NodeMap& cache)
...
-    intToBV_stack_element& stackHead = toVisit.back();
-    TNode current = stackHead.d_node;
-
-    // If node is already in the cache we're done, pop from the stack
-    NodeMap::iterator find = cache.find(current);
-    if (find != cache.end())
-    {
-      toVisit.pop_back();
-      continue;
-    }
```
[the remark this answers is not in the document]
yoni206: same question for the similar place in the binarizer.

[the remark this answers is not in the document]
alex-ozdemir: Ah. I was wondering about this. I'll put it back in.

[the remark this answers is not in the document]
alex-ozdemir: Yoni and I talked about this offline and realized that if the traversal encounters an previously seen node, this optimization doesn't just avoid rebuilding it, it avoids traversing its DAG altogether.

The traversal iterator (right now) doesn't allow you to avoid traversing parts of the DAG. It does allow you to avoid rebuilding previously seen nodes, or skip the traversal altogether, if the whole DAG has been seen before, but that's the best it can do.

If this optimization is important, I should delay this PR and add a "skip traversal" feature to the iterator (or something like that) first. If this optimization isn't that important, I can approximate it as described above, and continue the PR.

My guess is that it is important, but I don't really know this part of the code very well. What do you think, @4tXJ7f?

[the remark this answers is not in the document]
mpreiner: The pass should avoid traversing parts of the DAG that already have been processed. This is a general problem with the iterator and preprocessing passes.

src/preprocessing/passes/int_to_bv.cpp
```diff
@@ -147,30 +99,16 @@ Node intToBV(TNode n, NodeMap& cache)
...
   AlwaysAssert(!options::incrementalSolving());
 
-  vector<intToBV_stack_element> toVisit;
   NodeMap binaryCache;
   Node n_binary = intToBVMakeBinary(n, binaryCache);
-  toVisit.push_back(TNode(n_binary));
 
-  while (!toVisit.empty())
+  for (TNode current : NodeDfsIterable(n).inPostorder().skipIf(
+           [&cache](TNode nn) { return cache.count(nn) > 0; }))
```
alex-ozdemir, line 106: This is where we skip nodes already in the cache.

src/preprocessing/passes/int_to_bv.cpp
```diff
@@ -36,107 +37,58 @@ using NodeMap = std::unordered_map<Node, Node, NodeHashFunction>;
...
 
 Node intToBVMakeBinary(TNode n, NodeMap& cache)
 {
-  // Do a topological sort of the subexpressions and substitute them
-  vector<intToBV_stack_element> toVisit;
-  toVisit.push_back(n);
-
-  while (!toVisit.empty())
+  for (TNode current : NodeDfsIterable(n).inPostorder().skipIf(
+           [&cache](TNode nn) { return cache.count(nn) > 0; }))
```
alex-ozdemir, line 55: This is where we skip nodes already in the cache.

src/preprocessing/passes/int_to_bv.cpp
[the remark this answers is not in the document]
alex-ozdemir: Now that the `skipIf` traversal modifier is on `master`, I've updated this PR. Why don't you take another look, Yoni and Mathias?

src/preprocessing/passes/int_to_bv.cpp
```diff
@@ -147,30 +99,16 @@ Node intToBV(TNode n, NodeMap& cache)
   AlwaysAssert(size > 0);
   AlwaysAssert(!options::incrementalSolving());
 
-  vector<intToBV_stack_element> toVisit;
   NodeMap binaryCache;
   Node n_binary = intToBVMakeBinary(n, binaryCache);
-  toVisit.push_back(TNode(n_binary));
 
-  while (!toVisit.empty())
+  for (TNode current : NodeDfsIterable(n).inPostorder().skipIf(
```
yoni206, line 105: should `n` be `n_binary`?

src/expr/node_traversal.cpp
```diff
@@ -63,18 +63,19 @@ TNode& NodeDfsIterator::operator*()
...
 
-bool NodeDfsIterator::operator==(const NodeDfsIterator& other) const
+bool NodeDfsIterator::operator==(const NodeDfsIterator& other)
 {
   // The stack and current node uniquely represent traversal state. We need not
   // use the scheduled node set.
   //
   // Users should not compare iterators for traversals of different nodes, or
   // traversals with different skipIfs.
+  initializeIfUninitialized();
```
yoni206, line 73: Do the comments above this line correspond to it or to the line after it?

src/expr/node_traversal.h
```diff
@@ -58,9 +58,9 @@ class NodeDfsIterator
   // Dereference
   reference operator*();
   // Equals
-  bool operator==(const NodeDfsIterator&) const;
+  bool operator==(const NodeDfsIterator&);
```
yoni206, line 61: feels a bit confusing. Is there a clean way to keep these operators `const`? If not, perhaps this is worth documenting in a comment? Or am I missing something standard?

src/expr/node_traversal.h
```diff
@@ -58,9 +58,9 @@ class NodeDfsIterator
...
   // Not equals
-  bool operator!=(const NodeDfsIterator&) const;
+  bool operator!=(const NodeDfsIterator&);
```
yoni206, line 63: ditto.

src/expr/node_traversal.h
alex-ozdemir: Unfortunately, I believe that this is required, and is the cost of have `begin()` be constant-time.
I'll write about it.

src/expr/node_traversal.cpp
alex-ozdemir: The line after. I'll clean this up.

src/preprocessing/passes/int_to_bv.cpp
alex-ozdemir: Yes! Good catch!

src/expr/node_traversal.h
alex-ozdemir: Thanks for this comment! While writing the docs, I realized that I forgot to initialize `other`.

alex-ozdemir: Per an offline discussion with Yoni, we'd like you to take a look at this, @mpreiner. Do you think there might be some way to keep the `operator==` implementation `const`, with allowing for uninitialized iterators?

I can imagine one or two, but they're pretty bad.

alex-ozdemir: Okay, Mathias and I talked about this.

He thinks that this weirdness should be documented, but may be unavoidable.

alex-ozdemir:
Okay, @yoni206. I believe that I've addressed all of your concerns. Could you revisit this?

alex-ozdemir:
Paging @mpreiner to review the `skipIf` implementation, if he would like.

alex-ozdemir:
Actually, I just realized that `skipIf` is **not** in this PR. I believe that we've reviewed everything and answered all questions, so I'll go ahead and merge this.

PR #4172, opened 2020-03-28
Convert the last few Sygus benchmarks to V2.

abdoo8080:


test/regress/regress1/sygus/issue3461.sy
```diff
@@ -1,17 +1,17 @@
...
-(declare-datatype Policy 
-  ((p (principal Int)) 
+(declare-datatype Policy
+  ((p (principal Int))
    (por (left Policy) (right Policy))))
 
 (synth-fun mkPolicy ((d Doc)) Policy
-  ((Start Policy (Q))
-   (Q Policy ((p 0) (p 1) (por Q Q))))
+  ((Q Policy))
```
ajreynol: nit: Technically we should have 2 non-terminal symbols, Start and Q here.

abdoo8080: Addressed.

PR #4175, opened 2020-03-29
Frontend support for the choice operator

mudathirmahgoub:
Added the operator `choice` to `Smt2.g` and `Cvc.g`.
Removed the unused parameter `hasBoundVars` from ` TheoryModel::getModelValue`


test/regress/regress0/parser/choice.smt2
```diff
@@ -0,0 +1,10 @@
+(set-logic ALL)
+(set-info :status sat)
+(declare-fun a () Int)
+(declare-fun b () Int)
+(declare-fun c () Int)
+(assert (= (choice ((x Int)) (= x a)) 1))
+(assert (= (choice ((x Int)) (= x b)) 2))
+;(assert (let ((x (choice ((x Int)) true))) (and (distinct a b x)(= x c))))
```
mudathirmahgoub: `get-model` throws an [assertion](https://github.com/CVC4/CVC4/blob/master/src/smt/smt_engine.cpp#L3041) error when this commented line is enabled. 
I know excluding nodes with kind `CHOICE` in this assertion does not work because of wrong models like this one
```smt2
(define-fun c () Int (choice ((x Int)) true) ) 
```
So how to solve this issue?

src/parser/smt2/Smt2.g
```diff
@@ -2682,6 +2683,7 @@ ATTRIBUTE_INST_LEVEL : ':quant-inst-max-level';
 // operators (NOTE: theory symbols go here)
 EXISTS_TOK        : 'exists';
 FORALL_TOK        : 'forall';
+CHOICE_TOK        : 'choice';
```
4tXJ7f: This token should only be enabled when we are in non-strict parsing
```suggestion
CHOICE_TOK        : { !PARSER_STATE->strictModeEnabled() }? 'choice';
```

test/regress/regress0/parser/choice.smt2
ajreynol: This is an issue with some other operators as well (e.g. `(define c () Real (sin c'))`).  The issue is mostly just related to throwing the proper warnings instead of giving assertion failures. We can address on future PRs.

PR #4176, opened 2020-03-29
Use traversal interators in BvToInt

alex-ozdemir:
Replaces 2 of the 3 manual traversals in the bv-to-int preprocessing pass with `NodeDfsIterable(...).inPostorder()`.

Basically is just puts in a `for (TNode current : NodeDfsIterable(n).inPostorder())` in place of a while loop, and strips out a bunch of machinery.

**_Important_**: When viewing the diff, use the [setting](https://github.blog/2018-05-01-ignore-white-space-in-code-review/) that ignores whitespace changes! This is critical, because the indentation of a lot of code has changed. 

The last traversal does work during both pre- and post-visits, which the traversal iterator doesn't support yet. I did not modify this traversal.

@ahmed-irfan : I've assigned you because you have the most recent commit for this file, but feel free to re-assign.


aniemetz:
I suggest @mpreiner to review this before it gets merged (even if the other 2 approve it first), he was the one who mainly reviewed the related iterator PR (and @4tXJ7f ).

src/preprocessing/passes/bv_to_int.cpp
```diff
@@ -91,73 +92,44 @@ Node BVToInt::modpow2(Node n, uint64_t exponent)
  */
 Node BVToInt::makeBinary(Node n)
 {
-  vector<Node> toVisit;
-  toVisit.push_back(n);
-  while (!toVisit.empty())
+  for (TNode current : NodeDfsIterable(n).inPostorder())
```
ahmed-irfan: is it a DAG traversal?

src/preprocessing/passes/bv_to_int.cpp
```diff
@@ -269,560 +241,541 @@ Node BVToInt::bvToInt(Node n)
...
-  while (!toVisit.empty())
+  for (Node current : NodeDfsIterable(n).inPostorder())
   {
-    Node current = toVisit.back();
     uint64_t currentNumChildren = current.getNumChildren();
-    if (d_bvToIntCache.find(current) == d_bvToIntCache.end())
+    if (d_bvToIntCache.find(current) != d_bvToIntCache.end())
     {
-      // This is the first time we visit this node and it is not in the cache.
-      d_bvToIntCache[current] = Node();
```
ahmed-irfan: @yoni206 why were you doing this?
And how do the changes capture that?

src/preprocessing/passes/bv_to_int.cpp
alex-ozdemir: Yes.

src/preprocessing/passes/bv_to_int.cpp
alex-ozdemir: In this kind of traversal, there are three states for nodes:

1. Not yet previsited
2. Previsited (children added) by not post-visited
3. Postvisited

I believe state (1) was represented as "not in the cache", (2) was represented as "in the cache, mapped to Null", and (3) was represented as "in the cache, mapped to something non-null".

The post-order iterator handles all of this for you. Now the cache just stores the mapping from nodes to what they're processed into.

alex-ozdemir:
I marked this PR as do-not-merge pending the resolution of the discussion [here](https://github.com/CVC4/CVC4/pull/4169#discussion_r4025789760).

alex-ozdemir:
I've updated this PR to skip nodes already in the cache. This removes the blocker mentioned in the comment above.

src/preprocessing/passes/bv_to_int.cpp
```diff
@@ -269,560 +242,542 @@ Node BVToInt::bvToInt(Node n)
...
   uint64_t granularity = options::solveBVAsInt();
 
-  while (!toVisit.empty())
+  for (TNode current : NodeDfsIterable(n).inPostorder().skipIf(
+           [this](TNode nn) { return d_bvToIntCache.count(nn) > 0; }))
   {
-    Node current = toVisit.back();
     uint64_t currentNumChildren = current.getNumChildren();
-    if (d_bvToIntCache.find(current) == d_bvToIntCache.end())
+    if (d_bvToIntCache.find(current) != d_bvToIntCache.end())
```
yoni206, line 251: Is this condition ever `false` given the function given to `skipIf`?

mpreiner: Yes, this shouldn't be needed anymore since skipIf takes care of that, @alex-ozdemir?

alex-ozdemir: `skipIf` does take care of this. I'll change this accordingly.

Note though, that the `skipIf` PR hasn't landed yet, and is waiting on your review @mpreiner.

mpreiner:
@alex-ozdemir, @yoni206   what's the status of this PR?

mpreiner (approved):

alex-ozdemir:
IIRC, this was good to go, pending your review. What do you recall, @yoni206?

At any rate, I'll resolve the merge conflict later today.

alex-ozdemir:
Oof, the conflict was worse than I expected. I'll confer with Yoni to figure out what's going on first.

alex-ozdemir:
Okay, finally found some time to get back to this.
There are two phases of the bv-to-int pass which are refactored by this PR. One (`makeBinary`) can still be refactored as before, but the other (`bvToInt`) has become a little more complex due to how UF function symbols interact with its traversal. I'm going to close this PR, create a new one for `makeBinary`, and consider `bvToInt` after.

PR #4177, opened 2020-03-29
Remove undocumented/uncommon aliases

4tXJ7f:
This commit removes aliases that are either undocumented or associated
with uncommon options.

Note: The removal of `--statistics`, `--language`, and `--output-language` may be controversial. The options were undocumented previously and I am in favor of simplifying the options as much as possible, however. Please let me know what you think.


src/options/expr_options.toml
```diff
@@ -30,16 +25,6 @@ header = "options/expr_options.h"
   read_only  = true
   help       = "dagify common subexprs appearing > N times (1 == default, 0 == don't dagify)"
 
-[[alias]]
-  category   = "undocumented"
-  long       = "dag-thresh=N"
-  links      = ["--default-dag-thresh=N"]
```
aniemetz: why don't we rename the original one to `dag-thresh`?

src/options/expr_options.toml
```diff
@@ -67,8 +52,3 @@ header = "options/expr_options.h"
   default    = "DO_SEMANTIC_CHECKS_BY_DEFAULT"
   read_only  = true
   help       = "never type check expressions"
-
-[[alias]]
-  category   = "undocumented"
-  long       = "no-type-checking"
-  links      = ["--no-eager-type-checking"]
```
aniemetz: same here

ajreynol:
I am addressing the option names on another branch.

PR #4178, opened 2020-03-30
Rewrites for all remaining return statements in strings rewriter

ajreynol:
Towards proofs for string rewrites.  All return statements all now associated with an enum value.

An indentation in a block of code changed in rewriteMembership.


src/theory/strings/sequences_rewriter.cpp
```diff
@@ -1662,132 +1766,125 @@ Node SequencesRewriter::rewriteMembership(TNode node)
...
           }
         }
-        if (retNode != node)
-        {
-          Trace("regexp-ext-rewrite") << "Regexp star : rewrite " << node
-                                      << " -> " << retNode << std::endl;
-          break;
-        }
+      }
+      if (retNode != node)
```
4tXJ7f: If I am not mistaken, we should add a rewrite enum for this case and return instead of break.

ajreynol: Good catch, fixed.

PR #4190, opened 2020-03-31
Switch to GitHub actions for CI

mpreiner:
This commit enables CI with GitHub actions and adds macOS builds.


mpreiner:
I've disabled Travis in this PR. That's the reason why it's still waiting for the status.

PR #4191, opened 2020-03-31
Rename checkValid/query to checkEntailed.

aniemetz:
This renames api::Solver::checkValidAssuming to checkEntailed and
removes api::Solver::checkValid. Internally, SmtEngine::query is renamed
to SmtEngine::checkEntailed, and these changes are further propagated to
the Result class.

@makaimann to keep you in the loop


aniemetz:
> LGTM, I assume that refactoring of the Result class will happen in a separate PR?
@ajreynol yes

PR #4195, opened 2020-03-31
Introduce enums for all string inferences, excluding the core solver

ajreynol:
Introduce enum values for all calls to sendInference outside of the core solver.

This included some minor refactoring.


src/theory/strings/infer_info.h
```diff
@@ -94,6 +119,73 @@ enum class Inference : uint32_t
   //        for fresh u, u1, u2.
   // This is the rule F-Loop from Figure 5 of Liang et al CAV 2014.
   FLOOP,
+  //-------------------------------------- end core solver
+  //-------------------------------------- regexp solver
+  // regular expression normal form conflict
+  //   ( x in R ^ x = y ^ rewrite((str.in_re y R)) = false ) => false
+  // where y is the normal form computed for y.
```
4tXJ7f, line 126:
```suggestion
  // where y is the normal form computed for x.
```

PR #4197, opened 2020-04-01
Initialize theory rewriters in theories

4tXJ7f:
Until now, the `Rewriter` was responsible for creating `TheoryRewriter`
instances. This commit adds a method `mkTheoryRewriter()` that theories
override to create an instance of their corresponding theory rewriter.
The advantage is that the theories can pass additional information to
their theory rewriter (e.g. a statistics object).


test/unit/theory/theory_engine_white.h
```diff
@@ -91,118 +92,140 @@ struct RewriteItem {
...
+{
+ public:
+  /**
+   * Overrides TheoryRewriter::preRewrite().  This "fake theory" version
+   * ensures that this actual, observed pre-rewrite call matches the next
+   * "expected" call set up by the test.
+   */
+  RewriteResponse preRewrite(TNode n)
+  {
+    //    if(false) { //s_expected.empty()) {
```
ajreynol, line 105: Do we need this? Perhaps delete this and the other comment blocks.  I assume this is broken code?

PR #4201, opened 2020-04-02
Remove links field in all toml files

ajreynol:
This includes:
- All options pertaining to SMTEngine are now handled at the top of setDefaults.
- smtlibStrict was deleted in favor of a script.
- statsEveryQuery enables stats by modifying a public option function.  This is a slight hack but this code will likely get refactored as well soon.

A few other changes:
- Fix a bug in SMTEngine: defineFunction should finalize options.
- Call setDefaults before initilizing the TheoryEngine and ProofManager. This is necessary so that the PROOF(...) macro works earlier during initialization.

The next step will be to remove the links infrastructure for the options infrastructure. This will enable further detangling of our options dependencies.


contrib/cvc4_strict_smtlib
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+cvc4=./cvc4
```
4tXJ7f, line 3: How about something like:
```suggestion
cvc4="${CVC4_HOME}/cvc4"
```

src/smt/set_defaults.cpp
```diff
@@ -45,6 +45,49 @@ namespace smt {
...
+  if (options::bitvectorEqualitySlicer.wasSetByUser())
+  {
+    Trace("smt") << "SmtEngine: setting bitvectorEqualitySolver" << std::endl;
+    options::bitvectorEqualitySolver.set(true);
+  }
+  if (options::bitvectorAlgebraicBudget.wasSetByUser())
+  {
+    Trace("smt") << "SmtEngine: setting bitvectorAlgebraicSolver" << std::endl;
+    options::bitvectorAlgebraicSolver.set(true);
+  }
```
4tXJ7f, line 89: We generally use `Notice()` for these so that users can be informed about option changes with `-v` even if they use a production build.

```suggestion
  if (options::checkModels() || options::dumpModels())
  {
    Notice() << "SmtEngine: setting produceModels" << std::endl;
    options::produceModels.set(true);
  }
  if (options::checkModels())
  {
    Notice() << "SmtEngine: setting produceAssignments" << std::endl;
    options::produceAssignments.set(true);
  }
  if (options::dumpUnsatCoresFull())
  {
    Notice() << "SmtEngine: setting dumpUnsatCores" << std::endl;
    options::dumpUnsatCores.set(true);
  }
  if (options::checkUnsatCores() || options::dumpUnsatCores()
      || options::unsatAssumptions())
  {
    Notice() << "SmtEngine: setting unsatCores" << std::endl;
    options::unsatCores.set(true);
  }
  if (options::checkProofs() || options::dumpProofs())
  {
    Notice() << "SmtEngine: setting proof" << std::endl;
    options::proof.set(true);
  }
  if (options::bitvectorAigSimplifications.wasSetByUser())
  {
    Notice() << "SmtEngine: setting bitvectorAig" << std::endl;
    options::bitvectorAig.set(true);
  }
  if (options::bitvectorEqualitySlicer.wasSetByUser())
  {
    Notice() << "SmtEngine: setting bitvectorEqualitySolver" << std::endl;
    options::bitvectorEqualitySolver.set(true);
  }
  if (options::bitvectorAlgebraicBudget.wasSetByUser())
  {
    Notice() << "SmtEngine: setting bitvectorAlgebraicSolver" << std::endl;
    options::bitvectorAlgebraicSolver.set(true);
  }
```

contrib/cvc4_strict_smtlib
```diff
@@ -0,0 +1,8 @@
...
+
+# This is the set of command line arguments that is required to be strictly
+# complaint with the input and output requirements of the current SMT-LIB
+# standard.
+$cvc4 --lang=smt2 --output-lang=smt2 --strict-parsing --default-expr-depth=-1 --print-success --incremental --abstract-values $@
```
4tXJ7f, line 8:
```suggestion
"$cvc4" --lang=smt2 --output-lang=smt2 --strict-parsing --default-expr-depth=-1 --print-success --incremental --abstract-values $@
```

src/smt/set_defaults.cpp
```diff
@@ -45,6 +45,49 @@ namespace smt {
...
+  {
+    Trace("smt") << "SmtEngine: setting proof" << std::endl;
+    options::proof.set(true);
+  }
+  if (options::bitvectorAigSimplifications.wasSetByUser())
+  {
+    Trace("smt") << "SmtEngine: setting bitvectorAig" << std::endl;
+    options::bitvectorAig.set(true);
+  }
+  if (options::bitvectorEqualitySlicer.wasSetByUser())
```
4tXJ7f, line 80: What if the user explicitly disables the equality slicer?

ajreynol: Right now I'm preserving the previous behavior.  

I would prefer delaying addressing these issues until we have infrastructure for resolving these things.

4tXJ7f: Yeah, I think that's fair.

PR #4202, opened 2020-04-03
Update theory rewriter ownership, add stats to strings

4tXJ7f:
This commit adds statistics for string rewrites. This is work towards proof     
support in the string solver. At a high level, this commit adds a pointer to a  
`SequenceStatistics` in the rewriters and modifies                              
`SequencesRewriter::returnRewrite()` to count the rewrites done. In practice,   
to make this work requires a couple of changes, some of them temporary:         
                                                                                
- We can't have a single `Rewriter` instance shared between different           
  `SmtEngine` instances anymore. Thus the `Rewriter` is now owned by the        
  `SmtEngine` and calling the rewriter retrieves the rewriter associated with   
  the current `SmtEngine`. This is a temporary workaround before we get rid of  
  singletons.                                                                   
- Methods in the `SequencesRewriter` and the `StringsRewriter` are made         
  non-`static` because they need access to the statistics instance.             
- `StringsEntail` now has non-`static` methods because it needs a reference to  
  the sequences rewriter that it can call.                                      
- The interaction between the `StringsRewriter` and the `SequencesRewriter`     
  changed: the `StringsRewriter` is now a proper `TheoryRewriter` that inherits 
  from `SequencesRewriter` and calls its `postRewrite()` before applying its    
  own rewrites (this is essentially a reversal of roles from before: the        
  `SequencesRewriter` used to call `static` methods in the `StringsRewriter`).  
- The theory rewriters are now owned by the individual theories. This design    
  mirrors the `EqualityEngine`s owned by the individual theories.


ajreynol:
Will need to update the PR description.

src/theory/strings/extf_solver.cpp
```diff
@@ -620,7 +621,8 @@ void ExtfSolver::checkExtfInference(Node n,
   if (inferEqr.getKind() == EQUAL)
   {
     // try to use the extended rewriter for equalities
-    inferEqrr = SequencesRewriter::rewriteEqualityExt(inferEqr);
+    inferEqrr =
+        StringsRewriter(&d_statistics.d_rewrites).rewriteEqualityExt(inferEqr);
```
ajreynol, line 625: Can we actually just pass the instance of the rewriter here as a reference to the constructor instead of creating a new instance?

src/theory/strings/extf_solver.h
```diff
@@ -180,6 +182,8 @@ class ExtfSolver
   InferenceManager& d_im;
   /** cache of all skolems */
   SkolemCache& d_skCache;
+  /** The theory rewriter for this theory. */
+  StringsRewriter d_rewriter;
```
ajreynol, line 186:
```suggestion
  StringsRewriter& d_rewriter;
```

PR #4207, opened 2020-04-03
Renamed operator CHOICE to WITNESS

mudathirmahgoub:
Renamed operator CHOICE to WITNESS, and removed it from the front end


mudathirmahgoub:
Should we change this [option](https://github.com/CVC4/CVC4/blob/e24e6f3620996ee9e5010d30fefc51247cc55fdc/src/options/smt_options.toml#L411) to just `model-witness` instead of `model-witness-choice`?

src/api/cvc4cppkind.h
```diff
@@ -134,15 +134,15 @@ enum CVC4_PUBLIC Kind : int32_t
    */
   LAMBDA,
   /**
-   * Hilbert choice (epsilon) expression.
+   * Hilbert witness (epsilon) expression.
```
ajreynol, line 137: Just write "Witness expression".

src/api/cvc4cppkind.h
```diff
@@ -134,15 +134,15 @@ enum CVC4_PUBLIC Kind : int32_t
...
    * Parameters: 2
    *   -[1]: BOUND_VAR_LIST
-   *   -[2]: Hilbert choice body
+   *   -[2]: Hilbert witness body
```
ajreynol, line 140: Witness body.

src/theory/builtin/kinds
```diff
@@ -298,7 +298,7 @@ operator SEXPR 0: "a symbolic expression (any arity)"
 
 operator LAMBDA 2 "a lambda expression; first parameter is a BOUND_VAR_LIST, second is lambda body"
 
-operator CHOICE 2 "a Hilbert choice (epsilon) expression; first parameter is a BOUND_VAR_LIST, second is the Hilbert choice body"
+operator WITNESS 2 "a Hilbert witness (epsilon) expression; first parameter is a BOUND_VAR_LIST, second is the Hilbert witness body"
```
ajreynol, line 301: Remove Hilbert.

src/smt/term_formula_removal.cpp
```diff
@@ -138,9 +138,9 @@ Node RemoveTermFormulas::run(TNode node, std::vector<Node>& output,
       }
     }
   }
-  else if (node.getKind() == kind::CHOICE)
+  else if (node.getKind() == kind::WITNESS)
   {
-    // If a Hilbert choice function, witness the choice.
+    // If a Hilbert choice function, witnesses the choice.
```
ajreynol, line 143: Just witness choice.

ajreynol:
I would rename `model-witness-choice` to `model-witness-value`.

src/parser/tptp/Tptp.g
```diff
@@ -926,7 +926,10 @@ thfQuantifier[CVC4::api::Kind& kind]
   : FORALL_TOK { kind = api::FORALL; }
   | EXISTS_TOK { kind = api::EXISTS; }
   | LAMBDA_TOK { kind = api::LAMBDA; }
-  | WITNESS_TOK { kind = api::WITNESS; }
+  | WITNESS_TOK
```
HanielB, line 929: Please call it CHOICE_TOK, to avoid confusion

src/api/cvc4cppkind.h
```diff
@@ -134,15 +134,15 @@ enum CVC4_PUBLIC Kind : int32_t
    */
   LAMBDA,
   /**
-   * Hilbert choice (epsilon) expression.
+   * Witness expression.
```
HanielB, line 137: I think it'd be good to document what exactly is the "Witness" operator, i.e. how it differs from Hilbert choice.

HanielB: I don't understand the comment on the extra operand. And I think what would be good to document is that the "witness" operator or however we will call it is like Hilbert's choice except that it does not respect the axiom that forall x. F \equiv G -> choice x. F = choice x. G

ajreynol:
I believe the consensus is to use `witness` / `WITNESS`.

I'd like to proceed on this soon and have this in before the release.  @mudathirmahgoub, can you update the documentation in the cvc4cppkinds.h file to address Haniel's comment?

src/api/cvc4cppkind.h
```diff
@@ -134,15 +134,31 @@ enum CVC4_PUBLIC Kind : int32_t
...
-   * Hilbert choice (epsilon) expression.
+   * The syntax of a witness term is similar to a quantified formula except that
+   * only one bound variable is allowed.
+   * The term (witness ((x T)) F) returns an element x of type T
+   * and asserts F.
+   *
+   * The witness operator behaves like the description operator
+   * (see https://planetmath.org/hilbertsvarepsilonoperator) if there is no x
+   * that satisfies F. But if such x exists, the witness operator does not
+   * respect the uniqueness formula:
```
ajreynol, line 145:
```suggestion
   * enforce the axiom that ensures uniqueness up to logical equivalence:
```

src/api/cvc4cppkind.h
```diff
@@ -134,15 +134,31 @@ enum CVC4_PUBLIC Kind : int32_t
...
+   * forall x. F \equiv G => witness x. F =  witness x. G
+   *
+   * For example if there are 2 elements of type T that satisfy F, then the
+   * following expression is true:
+   * (distinct
+   *    (witness ((x Int)) (F x))
```
ajreynol, line 151:
```suggestion
   *    (witness ((x Int)) F)
```

src/api/cvc4cppkind.h
```diff
@@ -134,15 +134,31 @@ enum CVC4_PUBLIC Kind : int32_t
...
+   *    (witness ((x Int)) (F x)))
```
ajreynol, line 152:
```suggestion
   *    (witness ((x Int)) F))
```

src/api/cvc4cppkind.h
```diff
@@ -134,15 +134,31 @@ enum CVC4_PUBLIC Kind : int32_t
...
+   * and asserts F.
+   *
+   * The witness operator behaves like the description operator
+   * (see https://planetmath.org/hilbertsvarepsilonoperator) if there is no x
+   * that satisfies F. But if such x exists, the witness operator does not
+   * respect the uniqueness formula:
+   * forall x. F \equiv G => witness x. F =  witness x. G
+   *
+   * For example if there are 2 elements of type T that satisfy F, then the
+   * following expression is true:
```
ajreynol, line 149:
```suggestion
   * following formula is satisfiable:
```

src/api/cvc4cppkind.h
```diff
@@ -134,15 +134,31 @@ enum CVC4_PUBLIC Kind : int32_t
...
+   * that satisfies F. But if such x exists, the witness operator does not
+   * enforce the axiom that ensures uniqueness up to logical equivalence:
+   * forall x. F \equiv G => witness x. F =  witness x. G
+   *
+   * For example if there are 2 elements of type T that satisfy F, then the
+   * following formula is satisfiable:
+   * (distinct
+   *    (witness ((x Int)) F)
+   *    (witness ((x Int)) F))
+   *
```
ajreynol, line 153:
```suggestion
   * This kind is primarily used internally, but may be returned in models (e.g. for arithmetic terms in non-linear queries). However, it is not supported by the parser. Moreover, the user of the API should be cautious when using this operator. In general, all witness terms (witness ((x Int)) F) should be such that (exists ((x Int)) F) is a valid formula. If this is not the case, then the semantics in formulas that use witness terms may be intuitive.  For example, the following formula is unsatisfiable:
      (or (= (witness ((x Int)) false) 0) (not (= (witness ((x Int)) false) 0))
whereas notice that (or (= z 0) (not (= z 0))) is true for any z.
```

src/parser/tptp/Tptp.g
```diff
@@ -1627,7 +1630,7 @@ NOT_TOK        : '~';
 FORALL_TOK     : '!';
 EXISTS_TOK     : '?';
 LAMBDA_TOK     : '^';
-CHOICE_TOK     : '@+';
+WITNESS_TOK     : '@+';
```
ajreynol, line 1633: As Haniel said, this should be CHOICE_TOK.

test/regress/regress0/parser/witness.cvc
```diff
@@ -0,0 +1,10 @@
+% EXPECT: sat
```
ajreynol, line 1: I believe this benchmark should be deleted (we do not support WITNESS in parsers).

test/regress/regress0/parser/witness.smt2
```diff
@@ -0,0 +1,10 @@
+(set-logic ALL)
```
ajreynol, line 1: Ditto, delete.

PR #4222, opened 2020-04-07
Add skip predicate to node traversal.

alex-ozdemir:
Sometime you want to skip specific sub-DAGs when traversing a node. For example, you might be doing a transformation with a cache, and want to skip sub-DAGs that you've already processed.

This PR would add a `skipIf` builder method to `NodeDfsIterable`, which allows the user to provide a predicate according to which nodes will be omitted from the subsequent traversal.


mpreiner:
Marking do-not-merge until typos are fixed.

PR #4270, opened 2020-04-10
Ensure exported sygus solutions match grammar

ajreynol:
Previously we were doing rewriting/expand definitions during grammar normalization, which overwrote the original sygus operators. The connection to the original grammar was maintained via the SygusPrintCallback utility, which ensured that a sygus term printed in a way that matched the grammar.

We now have several use cases where solutions from SyGuS will be directly exported to the user, including the current use of `get-abduct`. This means that the terms must match the grammar, and we cannot simply rely on the print callback.

This moves the code to normalize sygus operators to datatypes utils, where the conversion between sygus and builtin terms takes place.  This allows a version of this function where isExternal = true, which constructs terms matching the original grammar.

This PR enables the SyGuS API to have an accurate getSynthSolution method.  It also will eliminate the need for SygusPrintCallback altogether, once the v1 parser is deleted.


src/theory/datatypes/theory_datatypes_utils.h
```diff
@@ -146,17 +146,31 @@ bool checkClash(Node n1, Node n2, std::vector<Node>& rew);
...
  */
 Kind getOperatorKindForSygusBuiltin(Node op);
+/**
+ * Returns the total version of Kind k if it is a partial operator, or
+ * otherwise k itself.
+ */
+Kind getEliminateKind(Kind k);
+/**
+ * Returns a version of n where all partial functions such as bvudiv
+ * have been replaced by their total versions like bvudiv_total.
```
yoni206, line 156: is this example up to date? isn't `bvudiv` already total by now?

ajreynol: bvudiv still exists (to support legacy version of smt-lib), although this will be eliminated in the future.

test/regress/regress1/sygus/yoni-true-sol.smt2
```diff
@@ -0,0 +1,20 @@
...
+(assert (and (>= n 0) (>= m 0)))
+(assert (< n i))
+(assert (< (+ i j) m))
+(get-abduct A
+  (<= n m)
+  ((GA Bool) (GJ Int) (GI Int))
+  (
+  (GA Bool ((>= GJ GI)))
+  (GJ Int ( j))
+  (GI Int ( i))
```
yoni206, line 18:
```suggestion
  (GI Int (mod i 1))
```

This way you are more sure that the solution is different from default grammar.

test/regress/regress1/sygus/yoni-true-sol.smt2
```diff
@@ -0,0 +1,20 @@
+; COMMAND-LINE: --produce-abducts
+; EXPECT: (define-fun A () Bool (>= j i))
+(set-logic QF_LIA)
```
yoni206, line 3:
```suggestion
(set-logic QF_NIA)
```

test/regress/regress1/sygus/yoni-true-sol.smt2
```diff
@@ -0,0 +1,20 @@
+; COMMAND-LINE: --produce-abducts
+; EXPECT: (define-fun A () Bool (>= j i))
```
yoni206, line 2:
```suggestion
; EXPECT: (define-fun A () Bool (>= j (mod i 1)))
```

PR #4298, opened 2020-04-12
Abort if in conflict in enumerative instantiation

### LINKED ISSUE #4293: Fatal failure at src/theory/quantifiers/instantiate.cpp:101
```
[760] % cvc4-1.7 -q small.smt2
unknown
[761] % cvc4 -q small.smt2
Fatal failure within bool CVC4::theory::quantifiers::Instantiate::addInstantiation(CVC4::Node, std::vector<CVC4::NodeTemplate<true> >&, bool, bool, bool) at /src/theory/quantifiers/instantiate.cpp:101
Check failure
 !d_qe->inConflict()
Aborted
[762] % 
[762] % cat small.smt2
(set-option :inst-when-tc-first false)
(set-option :arith-rewrite-equalities true)
(set-option :var-elim-quant false)
(set-option :quant-cf false)
(set-option :fs-interleave true)
(declare-fun b () Int)
(declare-fun c (Int) Bool)
(declare-fun d (Int) Bool)
(declare-fun e (Int Int) Bool)
(assert (forall ((f Int)) (= (c f) (= (distinct f b) (d f)))))
(assert (d b))
(assert
 (or (exists ((i Int)) (and (c i) (= i b)))
 (exists ((j Int) (f Int))
  (distinct (distinct j f)
  (= (e j f) (distinct (distinct f b) (e f j)))))))
(check-sat)
[763] %
```

OS: Ubuntu 18.04
Commit: 0073464

ajreynol:
In very rare cases, quantifiers engine can be the first to detect a quantifier-free conflict while constructing term indices.  When this occurs, instantiation modules can quit immediately.  This was not happening in a case of enumerative instantiation.

Fixes #4293.


ajreynol:
In very rare cases, quantifiers engine can be the first to detect a quantifier-free conflict while constructing term indices. When this occurs, instantiation modules can quit immediately. This was not happening in a case of enumerative instantiation.

Fixes #4293.

PR #4311, opened 2020-04-14
Add ProofNode data structure

ajreynol:
This is the core data structure for proofs in the new proofs infrastructure.  PfRule is a global enumeration of ids of proof nodes (analogous to Kind for Nodes).

Marking "complex" since would be good to have both reviews.


HanielB:
Can you change ProofStep to ProofRule? I think it is a more intuitive name.

src/expr/proof_step.h
```diff
@@ -0,0 +1,65 @@
...
+namespace CVC4 {
+
+/**
+ * An enumeration for proof steps. This enumeration is analogous to Kind for
+ * Node objects. In the documentation below, P:F denotes a ProofNode that
+ * proves formula F.
+ */
+enum class ProofStep : uint32_t
+{
+  //================================================= CORE
```
HanielB, line 31: What does CORE stand for?

src/expr/proof_step.h
```diff
@@ -0,0 +1,65 @@
...
+  // --------------
+  // Conclusion F
+  ASSUME,
+
+  //================================================= UNKNOWN
+  UNKNOWN,
+};
+
+/**
+ * Converts an proof step to a string. Note: This function is also used in
```
HanielB, line 44: an -> a

src/expr/proof_step.h
```diff
@@ -0,0 +1,65 @@
...
+ * `safe_print()`. Changing this functions name or signature will result in
```
HanielB, line 45: functions -> function

src/expr/proof_step.h
```diff
@@ -0,0 +1,65 @@
...
+ * `safe_print()` printing "<unsupported>" instead of the proper strings for
+ * the enum values.
+ *
+ * @param id The proof step
+ * @return The name of the proof step
+ */
+const char* toString(ProofStep id);
+
+/**
+ * Writes an proof step name to a stream.
```
HanielB, line 55: an -> a

src/expr/proof_node.h
```diff
@@ -0,0 +1,91 @@
...
+ *
+ * A ProofNode represents a single step in a proof. It contains:
+ * (1) d_id, an identifier indicating the type of inference,
+ * (2) d_children, the child ProofNode objects indicating its premises,
+ * (3) d_args, additional arguments used to determine the conclusion,
+ * (4) d_proven, cache of the formula that this ProofNode proves.
+ *
+ * Overall, a ProofNode and its children form a directed acyclic graph.
+ *
+ * A ProofNode is intended to be mutable in that (1), (2) and (3) can be
```
HanielB, line 37: in that -> so that?

src/expr/proof_node.h
```diff
@@ -0,0 +1,91 @@
...
+  /** Get children */
+  const std::vector<std::shared_ptr<ProofNode>>& getChildren() const;
+  /** Get arguments */
+  const std::vector<Node>& getArguments() const;
+  /** get what this node proves, or the null node if this is an invalid proof */
+  Node getResult() const;
+  /** Get assumptions
+   *
+   * This adds to the vector assump all formulas that are "assumptions" of the
+   * given proof. An assumption is a formula that is the argument of a
```
HanielB, line 64: Better not to use "argument" to avoid confusion with `d_args`. Also it'd be good to clarify that you are "recursively" traversing the proof to get its ASSUME leafs.

src/expr/proof_node.h
```diff
@@ -0,0 +1,91 @@
...
+ * Overall, a ProofNode and its children form a directed acyclic graph.
+ *
+ * A ProofNode is intended to be mutable in that (1), (2) and (3) can be
+ * modified. An example is when a ProofNode is established to be a "hole"
+ * for something to be proven later. However, (4) is intended to be immutable.
+ *
+ * The method setValue is private and can be called by objects that manage
+ * ProofNode objects in trusted ways that ensure that the node maintains
+ * the invariant above. Furthermore, notice that this class is not responsible
+ * for setting d_proven; this is done externally by a ProofNodeManager class.
```
HanielB, line 44: I'm confused. What is the intended usage of `setValue` if it is private?

ajreynol: It will be called by ProofNodeManager, which will be a friend class.

HanielB: OK.

src/expr/proof_step.h
ajreynol: I'll change to "core" (it was stylize like a theory id).

HanielB: But it'd be good to also clarify what does it mean for a rule to be a "core rule"

ajreynol: Added

ajreynol:
@HanielB  unfortunately the name "ProofRule" is taken: https://github.com/CVC4/CVC4/blob/master/src/proof/proof_manager.h#L105, and I'd rather not touch this.  I do like the name proof rule. I've renamed to "PfRule".  What do you think?

HanielB:
Let's use PfRule for now and when the old proof code disappears we can take the name back.

src/expr/proof_node.h
HanielB: I'd still prefer if you did not use "argument". Maybe "premise"?

ajreynol: Updated the doc

HanielB: What you are collecting is the conclusion of the ASSUME rules, isn't it? That was the origin of my confusion... maybe we should consider changing the documentation of that rule. The assumption is the conclusion of the ASSUME rule and we need no arguments in the proof node. I like the definition you wrote of arguments:

" * (3) d_args, additional arguments used to determine the conclusion,"

Since in the ASSUME rule we don't need an argument to determine the conclusion there is no point in putting the formula as an argument.

ajreynol: As discussed offline, this is an optimization which I believe is premature at this point.  The field d_proven is intended to be modified only by a checker. The ProofNode should explicitly represent what it proves, which is why the conclusion lives in d_args.  We can revisit this point as a performance optimization later if needed.

PR #4312, opened 2020-04-15
Split TermRegistry object from TheoryStrings

ajreynol:
This consolidates functionalities from TheoryStrings and InferenceManager related to registering terms, including sending "preregistration lemmas" for them.  The main purpose of this PR is to detangle this module from InferenceManager so that these two modules have exactly one call to OutputChannel::lemma each.

For the purposes of the theory solvers, TermRegistry contains the official SkolemCache of TheoryStrings, and can be seen as subsuming the previous interface for this class.

This PR is needed for further progress on strings proofs, marking as major since this will be a blocker shortly for this project.

A few things were cleaned in this PR.  One function changed name InferenceManager::registerTerm --> TermRegistry::getRegisterTermLemma.  No major behavior changes.

### DIFF
--- a/src/theory/strings/term_registry.h
+++ b/src/theory/strings/term_registry.h
@@ -52,7 +52,7 @@ class TermRegistry
   ~TermRegistry();
   /**
    * Preregister term, called when TheoryStrings::preRegisterTerm(n) is called.
-   * 
+   *
    * If we are using strings finite model finding (options::stringsFmf),
    * this determines if the term n should be added to d_inputVars, the set
    * of terms of type string whose length we are minimizing with its decision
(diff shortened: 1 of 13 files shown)


src/theory/strings/term_registry.h
```diff
@@ -0,0 +1,261 @@
...
+   *
+   * This method is called on non-constant string terms n that are "atomic"
+   * with respect to length. That is, the rewritten form of len(n) is itself.
+   *
+   * It sends a lemma on the output channel that ensures that the length n
+   * satisfies its assigned status (given by argument s).
+   *
+   * If the status is LENGTH_ONE, we send the lemma len( n ) = 1.
+   *
+   * If the status is LENGTH_GEQ, we send a lemma n != "" ^ len( n ) > 0.
```
4tXJ7f, line 100:
```suggestion
   * If the status is LENGTH_GEQ_ONE, we send a lemma n != "" ^ len( n ) > 0.
```

PR #4315, opened 2020-04-15
Eliminate remaining references to parent TheoryStrings object

ajreynol:
This PR makes it so that the module dependencies in the theory of strings are acyclic.  This is important for further organization towards proofs for strings.

The main change in this PR is to ensure that InferenceManager has a pointer to ExtTheory, which is needed for several of its methods.  This required changing several solvers into unique_ptr to properly initialize.

### DIFF
--- a/src/theory/strings/inference_manager.cpp
+++ b/src/theory/strings/inference_manager.cpp
@@ -382,12 +382,14 @@ void InferenceManager::assertPendingFact(Node atom, bool polarity, Node exp)
   {
     // we must ensure these terms are registered
     Trace("strings-pending-debug") << "  Register term" << std::endl;
-    for (unsigned j = 0; j < 2; j++)
+    for (const Node& t : atom)
     {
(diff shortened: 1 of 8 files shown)


src/theory/strings/inference_manager.cpp
```diff
@@ -30,16 +28,16 @@ namespace CVC4 {
 namespace theory {
 namespace strings {
 
-InferenceManager::InferenceManager(TheoryStrings& p,
-                                   context::Context* c,
+InferenceManager::InferenceManager(context::Context* c,
                                    context::UserContext* u,
                                    SolverState& s,
                                    TermRegistry& tr,
+                                   ExtTheory* e,
```
4tXJ7f, line 35: Why is this a pointer instead of a reference like the other arguments?

ajreynol: Had followed another part of the code, this is now updated to be consistent everywhere.

PR #4317, opened 2020-04-15
Add ProofNodeManager and ProofChecker

ajreynol:
Further work on adding core utilities for ProofNode objects, in support of the new proof infrastructure.

ProofNodeManager is analogous to NodeManager. It is a trusted way of generating only "well-formed" ProofNode pointers (according to a checker).

ProofChecker is analogous to TypeChecker.  It is intended to be infrastructure for our internal proof checker.

This PR (and 1 more) is required to get to a place where Haniel and I can collaborate on further development for proofs.

### DIFF
--- a/src/expr/proof_checker.cpp
+++ b/src/expr/proof_checker.cpp
@@ -1,0 +1,58 @@
+/*********************                                                        */
+/*! \file proof_checker.cpp
+ ** \verbatim
+ ** Top contributors (to current version):
+ **   Andrew Reynolds
+ ** This file is part of the CVC4 project.
+ ** Copyright (c) 2009-2019 by the authors listed in the file AUTHORS
+ ** in the top-level source directory) and their institutional affiliations.
+ ** All rights reserved.  See the file COPYING in the top-level source
+ ** directory for licensing information.\endverbatim
+ **
+ ** \brief Implementation of proof checker
+ **/
+
+#include "expr/proof_checker.h"
+
+namespace CVC4 {
+
+Node ProofChecker::check(ProofNode* pn, Node expected)
+{
+  return check(pn->getId(), pn->getChildren(), pn->getArguments(), expected);
+}
+
+Node ProofChecker::check(
+    PfRule id,
+    const std::vector<std::shared_ptr<ProofNode>>& children,
+    const std::vector<Node>& args,
+    Node expected)
+{
+  std::map<PfRule, ProofRuleChecker*>::iterator it = d_checker.find(id);
+  if (it == d_checker.end())
+  {
+    return Node::null();
+  }
+  // check it with the corresponding checker
+  Node res = it->second->check(id, children, args);
+  if (!expected.isNull() && res != expected)
+  {
+    // it did not match the given expectation, fail
+    return Node::null();
+  }
+  return res;
+}
+
+void ProofChecker::registerChecker(PfRule id, ProofRuleChecker* psc)
+{
+  std::map<PfRule, ProofRuleChecker*>::iterator it = d_checker.find(id);
+  if (it != d_checker.end())
+  {
+    // checker is already provided
+    Notice() << "ProofChecker::registerChecker: checker already exists for "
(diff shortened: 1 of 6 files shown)


src/expr/proof_checker.h
```diff
@@ -0,0 +1,92 @@
...
+/** A class for checking proofs */
+class ProofChecker
+{
+ public:
+  ProofChecker() {}
+  ~ProofChecker() {}
+  /**
+   * Return the formula that is proven by proof node pn, or null if pn is not
+   * well-formed. If expected is non-null, then we return null if pn does not
+   * prove expected.
```
HanielB, line 59: I think it's good to emphasize that the check is not done recursively, it's indeed only the node that is checked.

src/expr/proof_checker.cpp
```diff
@@ -0,0 +1,80 @@
...
+  // check it with the corresponding checker
+  std::vector<Node> cchildren;
+  for (const std::shared_ptr<ProofNode>& pc : children)
+  {
+    Assert(pc != nullptr);
+    Node cres = pc->getResult();
+    if (cres.isNull())
+    {
+      Trace("pfcheck") << "ProofChecker::check: child proof was invalid"
+                       << std::endl;
```
HanielB, line 47: Maybe good to say that it is invalid because it has no conclusion (otherwise it may sound like it failed to check or something).

src/expr/proof_node_manager.h
```diff
@@ -0,0 +1,116 @@
...
+namespace CVC4 {
+
+/**
+ * A manager for proof node objects. This is a trusted interface for creating
+ * and updating ProofNode objects.
+ *
+ * In more detail, we say a ProofNode is "well-formed (with respect to checker
+ * X)" if its d_proven field is non-null, and corresponds to the formula that
+ * the ProofNode proves according to X. The ProofNodeManager class constructs
+ * and update nodes that well-formed with respect to its underlying checker.
```
HanielB, line 34: that are

src/expr/proof_node_manager.h
```diff
@@ -0,0 +1,116 @@
...
+ */
+class ProofNodeManager
+{
+ public:
+  ProofNodeManager(ProofChecker* pc = nullptr);
+  ~ProofNodeManager() {}
+  /**
+   * This constructs a ProofNode with the given arguments. The expected
+   * argument, when provided, indicates the formula that the returned node
+   * is expected to prove. If we find that it does not based on the underlying
```
HanielB, line 58: not, based

src/expr/proof_node_manager.h
```diff
@@ -0,0 +1,116 @@
...
+ private:
+  /** The (optional) proof checker */
+  ProofChecker* d_checker;
+  /** Check internal
+   *
+   * This returns the result of proof checking a ProofNode with the provided
+   * arguments with an (optional) expected conclusion.
+   *
+   * This throws an assertion error if we fail to check such a proof node, or
+   * if expected is provided (non-null) is different what is proven by the
```
HanielB, line 105: and is different

src/expr/proof_node_manager.h
```diff
@@ -0,0 +1,116 @@
...
+                  const std::vector<std::shared_ptr<ProofNode>>& children,
+                  const std::vector<Node>& args);
+
+ private:
+  /** The (optional) proof checker */
+  ProofChecker* d_checker;
+  /** Check internal
+   *
+   * This returns the result of proof checking a ProofNode with the provided
+   * arguments with an (optional) expected conclusion.
```
HanielB, line 102: rather than optional say that it may be null (otherwise it can be confusing that `expected` does not have a default value)

src/expr/proof_checker.cpp
ajreynol: Sounds good.   Notice this case should never happen if all ProofNode are generated by ProofNodeManager. I've added an `Assert` failure here.

PR #4323, opened 2020-04-16
Add (context-dependent) Proof

ajreynol:
A (context-dependent) collection of proof steps that are linked to together to form a ProofNode dag.

Note that the name "Proof" is currently taken. I am calling this class "CDProof", although it is optionally context dependent.


src/expr/proof.h
```diff
@@ -0,0 +1,178 @@
...
+ *
+ * It is important to note that the premises of proof steps given to this class
+ * are *Node* not *ProofNode*. In other words, the user of this class does
+ * not have to manage ProofNode pointers while using this class. Instead,
+ * ProofNode is the final product of this class, via the getProof interface,
+ * which returns a ProofNode object that has linked together the proof steps
+ * added to this object.
+ *
+ * Its main interface is the function addStep, which registers a single
+ * step in the proof. These steps can be provided in any order.
```
HanielB, line 45: So the next example is easier to understand say the inputs of `addStep` here.

src/expr/proof.h
```diff
@@ -0,0 +1,178 @@
...
+ * call:
+ * - addStep( A, ID_A, { B, C }, {} )
+ * - addStep( B, ID_B, { D }, {} )
+ * - addStep( E, ID_E, {}, {} )
+ * with default options, then getProof( A ) returns the ProofNode of the form:
+ *   ID_A( ID_B( ASSUME( D ) ), ASSUME( C ) )
+ * Notice that the above calls to addStep can be made in either order. The
+ * proof step for E was not involved in the proof of A.
+ *
+ * If the user wants to combine proofs, then a addProof interface is
```
HanielB, line 57: a addProof -> an addProof

src/expr/proof.h
```diff
@@ -0,0 +1,178 @@
...
+ * available. The method addProof can be seen as syntax sugar for making
+ * multiple calls to addStep. Continuing the above example, if we call:
+ * - addProof( F, ID_F( ASSUME( A ), ASSUME( E ) ) )
+ * is the same as calling steps corresponding the steps in the provided proof
+ * bottom-up, starting from the leaves:
+ * --- addStep( A, ASSUME, {}, {}, true )
+ * --- addStep( E, ASSUME, {}, {}, true )
+ * --- addStep( F, ID_F, { A, E }, {}, true )
```
HanielB, line 65: clarify the role of fourth argument to `addStep` in the example

src/expr/proof.h
```diff
@@ -0,0 +1,178 @@
...
+ * --- addStep( B, ID_B2, {}, {}, true )
+ * --- addStep( C, ID_C, {}, {}, true )
+ * --- addStep( A, ID_A2, {B, C}, {}, true )
+ * --- addStep( D, ID_D, {A}, {}, true )
+ * Afterwards, getProof( D ) returns:
+ *   ID_D( ID_A1( ID_B1(), ID_C() ) )
+ * Notice that the steps with ID_A1 and ID_B1 were not overwritten by this call,
+ * whereas the assumption of C was overwritten by the proof ID_C(). Notice that
+ * the proof of D was completely traversed, despite encountering formulas, A and
+ * B, that were already given proof steps.
```
HanielB, line 89: It'd be good to say how you would overwrite if you wanted to.

src/expr/proof.h
```diff
@@ -0,0 +1,178 @@
...
+ * where assume ID_B2() and ID_C() prove B and C respectively. This call is
+ * equivalent to calling:
+ * --- addStep( B, ID_B2, {}, {}, true )
+ * --- addStep( C, ID_C, {}, {}, true )
+ * --- addStep( A, ID_A2, {B, C}, {}, true )
+ * --- addStep( D, ID_D, {A}, {}, true )
+ * Afterwards, getProof( D ) returns:
+ *   ID_D( ID_A1( ID_B1(), ID_C() ) )
+ * Notice that the steps with ID_A1 and ID_B1 were not overwritten by this call,
+ * whereas the assumption of C was overwritten by the proof ID_C(). Notice that
```
HanielB, line 87: Emphasize that assumptions are viewed as holes that may be filled, reason why they are by default overwritten.

src/expr/proof.h
```diff
@@ -0,0 +1,178 @@
...
+ * the proof of D was completely traversed, despite encountering formulas, A and
```
HanielB, line 88: Clarify that you are referring to the behavior of addProof, which will always lead to the recursive traversal.

src/expr/proof.h
```diff
@@ -0,0 +1,178 @@
...
+   */
+  bool addStep(Node expected,
+               PfRule id,
+               const std::vector<Node>& children,
+               const std::vector<Node>& args,
+               bool ensureChildren = false,
+               bool forceOverwrite = false);
+  /** Add proof
+   *
+   * @param fact The intended conclusion of the proof.
```
HanielB, line 152: I think we can do away with passing expected here. The proof node has a conclusion and the expected node is passed just to check that the conclusion of the proof is *already* equal to it.

src/expr/proof.h
```diff
@@ -0,0 +1,178 @@
...
+   * @param pn The proof of the given fact.
+   * @param forceOverwrite Whether we wish to force overwriting if a step was
```
HanielB, line 154: emphasize that this is the overwriting of non-assumption steps

ajreynol:
I've improved the documentation and simplified the interface to addProof.

src/expr/proof.h
ajreynol: Done above.

PR #4331, opened 2020-04-18
Track inference id for pending facts in strings

ajreynol:
This improves our tracking of pending inferences in strings so that we record pending inferences as a triple (id, fact, exp).  This will ensure that proof steps can be constructed at the time we decide to process facts.  It also inlines the `sendInfer` method into `sendInference` for simplicity.

This also improves the policy for reference counting facts and their explanations: we add them to d_keep when they are added to the equality engine. Previously, we were adding them when they were registered as pending.  This means we would collect facts in this pending set that were added but later abandoned, which is unnecessary.


src/theory/strings/inference_manager.h
```diff
@@ -35,6 +35,29 @@ namespace CVC4 {
 namespace theory {
 namespace strings {
 
+/**
+ * A pending inference. This is a helper class to track a unprocessed call to
```
4tXJ7f, line 39:
```suggestion
 * A pending inference. This is a helper class to track an unprocessed call to
```

PR #4363, opened 2020-04-19
Refactor inference manager in strings to be amenable to proofs

ajreynol:
This is a key refactoring towards proofs for strings.

This ensures that InferInfo is maintained until just before facts, lemmas and conflicts are processed.  This allows us both to:
(1) track more accurate stats and debug information,
(2) have enough information to ensure that proofs can be constructed at the moment facts, lemmas, and conflicts are processed.

Changes in this PR:
- PendingInfer and InferInfo were merged, CoreInferInfo is now the previous equivalent of InferInfo, extended with core-solver-specific information, which is only used by CoreSolver.
- sendInference( const InferInfo& info, ... ) is now the central method for sending inferences which is called by the other variants, not the other way around.
- sendLemma is inlined into sendInference(...) for clarity. 
- doPendingLemmas and doPendingFacts are extended to process the side effects of the inference infos.

There is actually a further issue with pending inferences related to eagerly processing the side effect of CoreInferInfo before we know the lemma is sent. I believe this issue does not occur in practice based on our strategy. Further refactoring could tighten this, e.g. virtual void InferInfo::processSideEffect. I will do this in another PR.

Further refactoring can address whether `asLemma` should be a field of InferInfo (it probably should), and whether we should explicitly check for AND in conclusions instead of making it the responsibility of the user of this class.

### DIFF
--- a/src/theory/strings/core_solver.cpp
+++ b/src/theory/strings/core_solver.cpp
@@ -2289,7 +2289,7 @@ void CoreSolver::checkLengthsEqc() {
   }
 }
 
-bool CoreSolver::processInferInfo(const CoreInferInfo& ii)
+bool CoreSolver::processInferInfo(CoreInferInfo& ii)
 {
   // rewrite the conclusion, ensure non-trivial
   ii.d_conc = Rewriter::rewrite(ii.d_conc);
--- a/src/theory/strings/core_solver.cpp
+++ b/src/theory/strings/core_solver.cpp
@@ -30,6 +30,8 @@ namespace CVC4 {
 namespace theory {
 namespace strings {
 
+CoreInferInfo::CoreInferInfo() : d_index(0), d_rev(false) {}
+
 CoreSolver::CoreSolver(context::Context* c,
                        context::UserContext* u,
                        SolverState& s,
@@ -963,16 +965,17 @@ void CoreSolver::processNEqc(std::vector<NormalForm>& normal_forms,
   unsigned max_index = 0;
   for (unsigned i = 0, size = pinfer.size(); i < size; i++)
   {
-    Trace("strings-solve") << "#" << i << ": From " << pinfer[i].d_i << " / "
-                           << pinfer[i].d_j << " (rev=" << pinfer[i].d_rev
+    CoreInferInfo& ipii = pinfer[i];
+    Trace("strings-solve") << "#" << i << ": From " << ipii.d_i << " / "
+                           << ipii.d_j << " (rev=" << ipii.d_rev
                            << ") : ";
-    Trace("strings-solve") << pinfer[i].d_conc << " by " << pinfer[i].d_id
+    Trace("strings-solve") << ipii.d_conc << " by " << ipii.d_id
                            << std::endl;
-    if (!set_use_index || pinfer[i].d_id < min_id
-        || (pinfer[i].d_id == min_id && pinfer[i].d_index > max_index))
+    if (!set_use_index || ipii.d_id < min_id
+        || (ipii.d_id == min_id && ipii.d_index > max_index))
     {
-      min_id = pinfer[i].d_id;
-      max_index = pinfer[i].d_index;
+      min_id = ipii.d_id;
+      max_index = ipii.d_index;
       use_index = i;
       set_use_index = true;
     }
@@ -1174,7 +1177,6 @@ void CoreSolver::processSimpleNEq(NormalForm& nfi,
     info.d_i = nfi.d_base;
     info.d_j = nfj.d_base;
     info.d_rev = isRev;
-    InferInfo& iinfo = info.d_infer;
     Assert(index < nfiv.size() - rproc && index < nfjv.size() - rproc);
     if (!d_state.areDisequal(xLenTerm, yLenTerm) && !d_state.areEqual(xLenTerm, yLenTerm)
         && !x.isConst()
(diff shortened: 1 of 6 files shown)


src/theory/strings/core_solver.cpp
```diff
@@ -2286,29 +2291,27 @@ void CoreSolver::checkLengthsEqc() {
...
-       ii.d_new_skolem)
-  {
-    for (const Node& n : sks.second)
-    {
-      d_termReg.registerTermAtomic(n, sks.first);
-    }
-  }
-}
+  // send the inference, which is a lemma
+  d_im.sendInference(ii, true);
```
4tXJ7f, line 2311: Note that `ii` will be sliced here. If we intend to have virtual methods in `InferInfo`, this will not work and we will instead have to pass a pointer. We could make the vectors holding inferences/lemmas `std::vector<std::unique_ptr<InferInfo>>`.

src/theory/strings/inference_manager.cpp
```diff
@@ -320,23 +274,25 @@ void InferenceManager::doPendingFacts()
...
+                            << ")) ; fact " << ii.d_id << std::endl;
+    // only keep stats if we process it here
+    Trace("strings-lemma") << "Strings::Fact: " << fact << " from " << exp
+                           << " by " << ii.d_id << std::endl;
+    d_statistics.d_inferences << ii.d_id;
+    // assert it as a pending fact
+    bool polarity = fact.getKind() != NOT;
+    TNode atom = polarity ? fact : fact[0];
+    // no double negation or double (conjunctive) conclusions
+    Assert(atom.getKind() != NOT && atom.getKind() != AND);
```
4tXJ7f, line 294: Should we maybe add these conditions to `InferInfo::isFact()`? I find it a bit confusing that something can be a fact but then can't be sent out as a fact.

src/theory/strings/inference_manager.cpp
```diff
@@ -349,20 +305,72 @@ void InferenceManager::doPendingFacts()
...
-  if (!d_state.isInConflict())
+  if (d_state.isInConflict())
+  {
+    // just clear the pending vectors, nothing else to do
+    d_pendingLem.clear();
+    d_pendingReqPhase.clear();
+    return;
+  }
+  NodeManager* nm = NodeManager::currentNM();
+  for (unsigned i = 0, psize = d_pendingLem.size(); i < psize; i++)
```
4tXJ7f, line 316: I don't see the index `i` being used. Should we maybe simply do `for (InferInfo& ii : d_pendingLem)`?

src/theory/strings/inference_manager.cpp
```diff
@@ -349,20 +305,72 @@ void InferenceManager::doPendingFacts()
...
+                           << std::endl;
+    // only keep stats if we process it here
+    d_statistics.d_inferences << ii.d_id;
+    ++(d_statistics.d_lemmasInfer);
+    d_out.lemma(lem);
+
+    // Process the side effects of the inference info.
+    // [1] Register the new skolems from this inference. We register them here
+    // (lazily), since this is the moment when we have decided to use the
+    // inference at use_index that involves them.
```
4tXJ7f, line 353:
```suggestion
    // (lazily)
```

src/theory/strings/inference_manager.cpp
ajreynol: Yes, I'll do that.

src/theory/strings/core_solver.cpp
ajreynol: I think the solution (which I had an earlier version of this PR) is to make CoreInferInfo a wrapper of InferInfo.

Generally, I think we should keep this core data structure simple. After some thought, I think I would not like virtual methods in InferInfo.

4tXJ7f: Yeah, I think I like this approach.

src/theory/strings/inference_manager.cpp
ajreynol: There is a distinction here about whose responsibility it is to use the class correctly.  Since this data structure is check is done millions of times, it would be nice to have some basic contracts on what is allowed to be a valid infer info.  The first check here is definitely a sanity check: no one should make a conclusion of the form (not (not F)).  The second is debatable, an earlier version of this PR had this check for AND in isFact.  I'll think about it a bit more.

src/theory/strings/inference_manager.cpp
```diff
@@ -349,20 +305,72 @@ void InferenceManager::doPendingFacts()
...
+      sendPhaseRequirement(pp.first, pp.second);
+    }
+  }
+  // process the pending require phase calls
+  for (const std::pair<const Node, bool>& prp : d_pendingReqPhase)
+  {
+    Trace("strings-pending") << "Require phase : " << prp.first
+                             << ", polarity = " << prp.second << std::endl;
+    d_out.requirePhase(prp.first, prp.second);
   }
```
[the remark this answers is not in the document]
ajreynol: The issue with this is that someone may wish to set a phase without having a lemma.

I'm going to do a different simplification:  InferInfo does not contain d_pendingPhase.  It actually doesn't make sense to have it in InferInfo since:
(1) our inferences with pending phase requirements are restricted to lemma infer infos,
(2) pending lemmas and pending phase requirements are always processed at the same time. 
Thus, the called on sendInference / sendPhaseRequirement knows they are handled in sync.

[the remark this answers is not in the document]
4tXJ7f: Yeah, that works, too. I am happy as long as we only have a single mechanism for setting phase requirements.

src/theory/strings/inference_manager.cpp
ajreynol: Actually its used, we store the use_index based on it.

src/theory/strings/inference_manager.cpp
4tXJ7f: We've discussed the handling of facts/lemmas offline and will potentially revisit it in a future PR.

PR #4373, opened 2020-04-21
Allow eager bitblasting with solve bv as int in QF_NIA

ajreynol:
This also updates the SMT COMP script to revert to our previous behavior.

This is required for SMT COMP. It should be beneficial for satisfiable QF_NIA instances. I will revisit/test this independently.


src/smt/set_defaults.cpp
```diff
@@ -133,8 +133,19 @@ void setDefaults(SmtEngine& smte, LogicInfo& logic)
...
   if (options::solveIntAsBV() > 0)
   {
+    // not compatible with incremental
+    if (options::incrementalSolving())
+    {
+      throw OptionException(
+          "solving integers as bitvectors is currently not supported "
+          "when solving incrementally.");
+    }
+    // Int to BV current always eliminates arithmetic completely (or otherwise
```
yoni206, line 143:
```suggestion
    // Int to BV currently always eliminates arithmetic completely (or otherwise
```

ajreynol:
I am still not certain what is the best configuration for SMT-COMP, will test more today.  For now, I will merge, and then update the SMT COMP script later if needed.

PR #4378, opened 2020-04-22
Support the SMT-LIB Unicode string standard by default

ajreynol:
This PR merges --lang=smt2.6.1 and --lang=smt2.6 (default). It makes it so that 2.6 always expects the syntax of the string standard http://smtlib.cs.uiowa.edu/theories-UnicodeStrings.shtml.

I've updated the regressions so that the 2.6 benchmarks are now compliant with the standard. Some of the <=2.5 benchmarks I've updated to 2.6. Others I have left for now, in particular the ones that rely on special characters or ad-hoc escape sequences.  The old formats will be supported in the release but removed shortly afterwards.

This PR is a prerequisite for the release, but not necessarily SMT-COMP (which will use --lang=smt2.6.1 if needed).  Notice that we still do not have parsing support for str.replace_re or str.replace_re_all. This is required to be fully compliant.


ajreynol:
Marking do not merge; we should either:
(1) merge after SMT-COMP, or
(2) change the SMT-COMP script to not use --lang=smt2.6.1.

PR #4390, opened 2020-04-25
Add sygus datatype substitution utility method

ajreynol:
This makes the method for substiutiton and generalization of sygus datatypes a generic utility method. It updates the abduction method to use it. Interpolation is another target user of this utility.

This is not required for the release.


src/theory/datatypes/theory_datatypes_utils.h
```diff
@@ -245,6 +245,46 @@ Node sygusToBuiltin(Node c, bool isExternal = false);
...
+/** Get free symbols in a sygus datatype type
+ *
+ * Add the free symbols (expr::getSymbols) in terms that can be generated by
+ * sygus datatype sdt to the set syms. For example, given sdt encodes the
+ * grammar:
+ *   G -> a | +( b, G ) | c | e
+ * We have that { a, b, c, e } are added to syms. Notice that expr::getSymbols
+ * excludes variables whose kind is BOUND_VARIABLE.
+ */
+void getFreeSymbolsSygusType(TypeNode sdt,
```
yoni206, line 257: Can you use it in the abduction module?

ajreynol: I think I could potentially use this as an optimization.  This will require further changes though. I'd prefer doing this in a separate PR.

src/theory/datatypes/theory_datatypes_utils.h
```diff
@@ -245,6 +245,46 @@ Node sygusToBuiltin(Node c, bool isExternal = false);
...
+                             std::unordered_set<Node, NodeHashFunction>& syms);
+
+/** Substitute and generalize a sygus datatype type
+ *
+ * This transforms a sygus datatype sdt into another one sdt' that generates
+ * terms t such that t * { vars -> syms } is generated by sdt.
+ *
+ * The arguments syms and vars should be vectors of the same size and types.
+ * It is recommended that the arguments in syms and vars should be variables
+ * (return true for .isVar()) but this is not required.
```
HanielB, line 267: say why

src/theory/datatypes/theory_datatypes_utils.h
```diff
@@ -245,6 +245,46 @@ Node sygusToBuiltin(Node c, bool isExternal = false);
...
+ *
+ * The variables in vars of type BOUND_VARIABLE are added to the
+ * formal argument list of t. Other symbols are not.
```
HanielB, line 270: I think this need to be justified. First thing that comes to mind is that this should be optional, but there is the question of whether the synthesis module would even work if this was not done, right? Probably good to say this.

src/theory/datatypes/theory_datatypes_utils.cpp
```diff
@@ -711,6 +712,223 @@ Node sygusToBuiltinEval(Node n, const std::vector<Node>& args)
...
+  ssutn0 << sdtd.getName() << "_s";
+  TypeNode abdTNew =
+      nm->mkSort(ssutn0.str(), ExprManager::SORT_FLAG_PLACEHOLDER);
+  unres.insert(abdTNew.toType());
+  dtProcessed[sdt] = abdTNew;
+
+  // We must convert all symbols in the sygus datatype type sdt to
+  // apply the substitution { syms -> vars }, where syms is the free
+  // variables of the input problem, and vars is the formal argument list
+  // of the function-to-synthesize.
```
HanielB, line 811: update this comment

PR #4402, opened 2020-04-28
Update cardinality in strings to unicode standard

ajreynol:
This updates the default cardinality in strings to match the Unicode standard, 196608.

This avoids a check-model failure from 25 benchmarks in SMT-LIB, which were related to a split due to insufficient constants being required during collectModelInfo. 

This also makes a few places throw an AlwaysAssert(false) that otherwise would lead to incorrect models.  These regardless should never throw, but it would be better to have an assertion failure.


src/theory/strings/theory_strings.cpp
```diff
@@ -491,6 +498,11 @@ bool TheoryStrings::collectModelInfoType(
         processed[eqc] = c;
         if (!m->assertEquality(eqc, c, true))
         {
+          // this should never happen due to the model soundness argument
+          // for strings
+          AlwaysAssert(false)
```
4tXJ7f, line 503:
```suggestion
          Unreachable()
```

src/theory/strings/theory_strings.cpp
```diff
@@ -534,6 +546,11 @@ bool TheoryStrings::collectModelInfoType(
       processed[nodes[i]] = cc;
       if (!m->assertEquality(nodes[i], cc, true))
       {
+        // this should never happen due to the model soundness argument
+        // for strings
+        AlwaysAssert(false) << "TheoryStrings::collectModelInfoType: "
```
4tXJ7f, line 551:
```suggestion
        Unreachable() << "TheoryStrings::collectModelInfoType: "
```

src/theory/strings/theory_strings.cpp
```diff
@@ -264,6 +264,8 @@ bool TheoryStrings::collectModelInfo(TheoryModel* m)
   // assert the (relevant) portion of the equality engine to the model
   if (!m->assertEqualityEngine(&d_equalityEngine, &termSet))
   {
+    Assert(false);
```
4tXJ7f, line 267: It may be a good idea to make this an `Unreachable()` as well.

PR #4418, opened 2020-04-29
Fix an incorrect limit in conversion from real to float

martin-cs:
This error is a bit inexplicable but very very definitely wrong.
A test case from the original bug report is included.


martin-cs:
Rebased and the test corrected.

src/util/floatingpoint.cpp
```diff
@@ -920,15 +921,21 @@ static FloatingPointLiteral constructorHelperBitVector(
...
       Integer signedSignificand(sign * significand);
-      
-      // Only have pow(uint32_t) so we should check this.
-      Assert(this->t.significand() <= 32);
+
+      // We only have multiplyByPow(uint32_t) so we can't convert all numbers.
+      // Note that multipling by 2^n requires n bits of space (worst case)
+      // so, in effect, these tests limit us to cases where the resultant
+      // number requires up to 512 megabyte to represent.
+      Assert(sizeof(unsigned int) >= sizeof(uint32_t));
```
aniemetz, line 929: What do you need this assertion for?

martin-cs: `multiplyByPow2` takes a `uint32_t`.
The best fit for a conversion from an `Integer` is to `unsigned int`
If it overflowed it would silently give the wrong result, so I want to check that won't happen.

BUT

I've got it the wrong way around, haven't I?  It should be `sizeof(uint32_t) >= sizeof(unsigned int)`, also it is arguably redundant given the later Asserts.

martin-cs: Actually, no, I'm an idiot who doesn't understand his code.

martin-cs: I am checking `exp` is `numeric_limits<uint32_t>::max()` which means it is fine for the `multiplyByPow2` as long as the conversion via `unsigned int` doesn't loose anything.  So `unsigned int` has to be at least `uint32_t` which is not guaranteed by the C standard.

aniemetz: OK. We'll move to fixed width types eventually, and then `exp.toUnsignedInt()` will return uint32_t or uint64_t and this will become obsolete. But for now, all is good.

src/util/floatingpoint.cpp
```diff
@@ -920,15 +921,21 @@ static FloatingPointLiteral constructorHelperBitVector(
...
+      Integer shiftLimit(std::numeric_limits<uint32_t>::max());
```
aniemetz, line 930:
```suggestion
#ifdef CVC4_ASSERTIONS
      Integer shiftLimit(std::numeric_limits<uint32_t>::max());
#endif
```

src/util/floatingpoint.cpp
4tXJ7f: Since this doesn't depend on runtime values, I'd suggest that we use a `static_assert` here.

aniemetz: Good point.

martin-cs: I have changed to a `static_assert` and updated the comments to include more of this reasoning.

PR #4435, opened 2020-05-03
Bv to int elimination bugfix

yoni206:
fix 1:
------
The wrong flag was checked in the traversal, causing an assertion error [here](https://github.com/CVC4/CVC4/blob/8236d7f9bff3aef4f7b37a15d509b8a11551401f/src/preprocessing/passes/bv_to_int.cpp#L247)
This is fixed in this PR. A test was added as well.

fix 2:
------
It is desirable that bv-to-bool runs before bv-to-int, but this was not the case, and is fixed in this PR.

Do not merge until after competition release (label added).


src/preprocessing/passes/bv_to_int.cpp
```diff
@@ -184,43 +184,39 @@ Node BVToInt::eliminationPass(Node n)
...
+                                  RewriteRule<SgeEliminate>,
+                                  RewriteRule<ShlByConst>,
+                                  RewriteRule<LshrByConst> >::apply(current);
+      // save in the cache
+      d_eliminationCache[current] = currentEliminated;
+      // put the eliminated node in the rebuild cache, but mark that it hasn't
+      // yet been rebuilt by assigning null.
+      d_rebuildCache[currentEliminated] = Node();
+      // Push the eliminated node to the stack
+      toVisit.push_back(currentEliminated);
```
alex-ozdemir, line 214: Would it be correct to map `currentEliminated` to itself in `d_eliminationCache`?

I worry that not doing this results in a non-trivial amount of redundant work. Currently, if `a` is eliminated to `b`, then `b` is put in the stack to be visited, but, since it is not in the elimination cache, we will re-eliminate it (a no-op), and push its children onto the stack, and process them again (as no-ops, since by this point they've been fully processed).

Let me know if I'm misunderstanding something!

src/smt/set_defaults.cpp
```diff
@@ -350,6 +350,18 @@ void setDefaults(SmtEngine& smte, LogicInfo& logic)
...
+  {
+    /**
+     * Operations on 1 bits are better handled as Boolean operations
+     * than as integer operations.
+     * Therefore, we enable bv-to-bool, which runs before
+     * the translation to integers.
+     */
+    options::bitvectorToBool.set(true);
+  }
+
```
alex-ozdemir, line 364: Is pulling this out of the unsat-core/proof conditional intentional?

yoni206: yes, this was there by mistake. bv-to-bool should be turned on regardless of proofs/unsat cores.

src/preprocessing/passes/bv_to_int.cpp
yoni206: What if the elimination introduces a new child?
say `a` is `((_ zero_extend 4) x)`. Then `b` will be `(concat (_ bv0 4) x)`. If you put `b` in `d_eliminationCache` and map it to itself right at that point, then when will you add `(_ bv0 4)` to the stack to be visited?

alex-ozdemir: We talked about this offline, and we think that the suggested change can be made and would be an improvement. Yoni's gonna go ahead with it.

PR #4438, opened 2020-05-04
Make SolveEq and PlusCombineLikeTerms idempotent

### LINKED ISSUE #3692: Fatal failure at CVC4/src/./theory/bv/theory_bv_rewrite_rules_normalization.h:829
Hi,
For this formula,
```
(declare-fun a () (_ BitVec 32))
(assert (exists ((b (_ BitVec 32)))
  (forall ((c (_ BitVec 32)))
    (exists ((d (_ BitVec 32)))
      (let ((a!1 (bvmul (bvadd (bvsmod (bvneg #x00000042) d)
                               (bvmul (bvneg #x00000007) c))
                        (bvand (bvnot #x0000003d) a)))
            (a!2 (distinct (bvsdiv (bvxnor (bvsub #x0000000b d)
                                           (bvmul #x00000021 c))
                                   (bvsdiv #x00000005 b))
                           (bvnot #x00000055))))
        (or (distinct a!1 #x00000033) (not a!2)))))))
(check-sat)
```
```
Fatal failure within static CVC4::Node CVC4::theory::bv::RewriteRule<rule>::apply(CVC4::TNode) [with CVC4::theory::bv::RewriteRuleId rule = (CVC4::theory::bv::RewriteRuleId)135; CVC4::Node = CVC4::NodeTemplate<true>; CVC4::TNode = CVC4::NodeTemplate<false>] at CVC4/src/./theory/bv/theory_bv_rewrite_rules_normalization.h:829
Check failure
 (newLeft == left && newRight == right) || Rewriter::rewrite(newLeft) != left || Rewriter::rewrite(newRight) != right
Aborted
```

OS: Ubuntu 18.04
Revision: 7583e03

4tXJ7f:
Fixes #3692 and an assertion failure that came up during the test runs
for SMT-COMP. The bit-vector rewrites `SolveEq` and
`PlusCombineLikeTerms` were not always idempotent. At a high level,
`SolveEq` combines common terms from two sides of an equality and
`PlusCombineLikeTerms` combines common terms within an addition.
However, in doing so, these rewrites were reordering the operands of the
bit-vector addition based on the node ids of the terms that were
multiplied with their coefficients. Consider the addition `3 * x * y + 5 * y * z`
(the bit-width does not matter). `PlusCombineLikeTerms` would
reorder this addition to `5 * y * z + 3 * x * y` if the node id of `y * z`
was smaller than the node id of `x * y`. The issue is that node ids
are not fixed for a given term: If we have a term `x * y` and that term
reaches ref count 0, we may get a different id for that same term if we
recreate it later on. When terms reach ref count 0, we don't immediately
delete them but add them to our set of zombies to be deleted whenever
the list of zombies grows larger than some fixed size. When applying
`SolveEq` and `PlusCombineLikeTerms` multiple times (even in direct
succession without doing anything else), the node order may change
because some of the terms like `x * y` may be zombies while others have
been deleted and get new ids, leading to the relative order of node ids
changing. I suspect that we could construct a case where we get into an
infinite rewrite loop.

This commit addresses the issue as follows: It does not perform the
rewrites `SolveEq` and `PlusCombineLikeTerms` if none of the operands
change. This makes the rewrites idempotent. Note however that we are
still not guaranteeing that a term has the same rewritten form
throughout an execution because the node ids may change if the term has
been freed in the meantime. However, this limitation is consistent with
other rewrites such as the reordering of equalities.

I am including the minimized test case from our run on SMT-LIB. I am
ommittin the test case from #3692 because I couldn't trigger it on
master (not surprising since the issue requires very specific
circumstances to actually occur). However, I was able to reproduce the
issue on the CVC4 version mentioned in the issue and confirmed that this
fix worked for that older version.


src/theory/bv/theory_bv_rewrite_rules_normalization.h
```diff
@@ -385,7 +385,7 @@ inline Node RewriteRule<PlusCombineLikeTerms>::apply(TNode node)
   std::map<Node, BitVector> factorToCoefficient;
 
   // combine like-terms
-  for (unsigned i = 0; i < node.getNumChildren(); ++i)
+  for (size_t i = 0; i < node.getNumChildren(); ++i)
```
aniemetz, line 388:
```suggestion
  for (size_t i = 0, n = node.getNumChildren(); i < n; ++i)
```

PR #4439, opened 2020-05-04
Do not traverse beneath quantifiers in unconstrained simplification pass

### LINKED ISSUE #4437: Soundness bug on using (set-logic QF_AUFBVLIA) on BV formula
Hi,
```
[570] % cvc4 small.smt2 
sat
[571] % cvc4 --strings-exp small.smt2 
unsat
[572] % cat small.smt2 
(set-logic QF_AUFBVLIA)
(declare-fun a () (_ BitVec 8))
(declare-fun b () (_ BitVec 8))
(assert (forall ((c (_ BitVec 8))) (= (bvashr c a) b)))
(check-sat)
[573] %
```
unsat is the right answer; strictly speaking, the formula is not QF_AUFVLIA because of the forall, but cvc4 doesn't complain and give an incorrect answer.

OS: Ubuntu 18.04
Commit: 8236d7f

ajreynol:
Fixes #4437.

The benchmark is malformed (but "unsat" if we ignore the set-logic command).  We incorrectly answer "sat" for it, due to unconstrained simplification replacing the body of the quantified formula with a fresh Boolean variable, it is being run since the logic is incorrectly specified.

Notice that unconstrained simplification is not run by default when quantifiers are not in the input logic.  However, there are a number of ways unconstrained simplification can be run on quantifiers due to combinations of options.

This makes the unconstrained simplification pass robust to quantifiers by not attempting to add substitutions for terms beneath quantifiers.


ajreynol:
Great catch, I'll fix and add this regression as well.

ajreynol:
I'm going to submit a simpler more conservative fix, closing this PR.

PR #4468, opened 2020-05-05
Update copyright year and AUTHORS/THANKS files.

aniemetz:


THANKS
```diff
@@ -21,9 +23,13 @@ Thanks to:
 - Thomas Hunger for some important patches to CVC4's SWIG interfaces in March
   2014.
 
+- Andrew V. Jones for several fixes in 2019 and 2020.
+
 - Makai Mann of Stanford University for updating the build scripts to support
   computing coverage in 2017.
```
HanielB, line 29: Didn't we agree that once somebody is in AUTHORS the person is out of THANKS?

aniemetz: Ah right, good catch. Fixed.

PR #4485, opened 2020-05-09
Add proof skolem cache

ajreynol:
This adds the general utility for maintaining mappings from Skolems to their witness form via attributes.

I am sending this as a PR now since it would be helpful to use this class for fixing some of the recent unconstrained-simp bugs.


src/expr/proof_skolem_cache.h
```diff
@@ -0,0 +1,123 @@
...
+
+#include <string>
+
+#include "expr/node.h"
+
+namespace CVC4 {
+
+/**
+ * A manager for skolems that can be used in proofs. This is designed to be
+ * trusted interface to NodeManager::mkSkolem, where one
```
HanielB, line 28: to be a

src/expr/proof_skolem_cache.h
```diff
@@ -0,0 +1,123 @@
...
+   * @param flags The flags for the Skolem (see NodeManager::mkSkolem)
+   * @return The skolem whose witness form is registered by this class.
+   */
+  static Node mkSkolem(Node v,
+                       Node pred,
+                       const std::string& prefix,
+                       const std::string& comment = "",
+                       int flags = NodeManager::SKOLEM_DEFAULT);
+  /**
+   * Same as above, but for special case for (witness ((x T)) (= x t))
```
HanielB, line 92: phrasing

src/expr/proof_skolem_cache.cpp
```diff
@@ -0,0 +1,183 @@
...
+                                int flags)
+{
+  Assert(v.getKind() == BOUND_VARIABLE);
+  // make the witness term
+  NodeManager* nm = NodeManager::currentNM();
+  Node bvl = nm->mkNode(BOUND_VAR_LIST, v);
+  // translate pred to witness form, since pred itself may contain skolem
+  Node predw = getWitnessForm(pred);
+  // make the witness term, which should not contain any skolem
+  Node w = nm->mkNode(CHOICE, bvl, predw);  // will change to WITNESS
```
HanielB, line 51: Hasn't master already been updated to use `WITNESS` as the kind?

src/expr/proof_skolem_cache.h
```diff
@@ -0,0 +1,123 @@
...
+   * where T is the type of t. This skolem is unique for each t.
+   */
+  static Node mkPurifySkolem(Node t,
```
HanielB, line 95: I think it'd be good to say here why there is a dedicated attribute for this case.

ajreynol:
Marking do not merge since I have at least one more feature I'd like to add to this class on stringsPf branch.

src/expr/proof_skolem_cache.cpp
ajreynol: Not yet.

ajreynol:
@HanielB  I addressed the review. Notice that I also added a few more features in my latest commit that are being used on the strings proof branch.

src/expr/proof_skolem_cache.h
HanielB: You didn't address this comment.

PR #4487, opened 2020-05-12
CegqiBv: Clean up after renaming options.

aniemetz:


src/options/quantifiers_options.toml
```diff
@@ -1763,7 +1763,7 @@ header = "options/quantifiers_options.h"
   long       = "quant-epr"
   type       = "bool"
   default    = "false"
-  help       = "infer whether in effectively propositional fragment, use for cbqi"
+  help       = "infer whether in effectively propositional fragment, use for cegqiBv"
```
ajreynol, line 1766: This should just be "cegqi".

ajreynol, line 1766:
```suggestion
  help       = "infer whether in effectively propositional fragment, use for cegqi"
```

aniemetz: Fixed.

PR #4493, opened 2020-05-14
Fix parametric datatype instantiation

ajreynol:
A bug was introduced when adding the Node-level datatype implementation in https://github.com/CVC4/CVC4/commit/d803e7fcf60f9bb847853fe6ccf7589b94b76922. The code did not probably get the arity of a sort constructor.  This adds TypeNode::getSortConstructorArity and uses it during parametric datatype type resolution.


src/expr/type_node.cpp
```diff
@@ -459,9 +459,16 @@ TypeNode TypeNode::instantiateParametricDatatype(
   return nm->mkTypeNode(kind::PARAMETRIC_DATATYPE, paramsNodes);
 }
 
+unsigned TypeNode::getSortConstructorArity() const
```
4tXJ7f, line 462:
```suggestion
uint64_t TypeNode::getSortConstructorArity() const
```

The `SortArityAttr` atrribute maps to `uint64_t`. We could also consider returning a `size_t` here, I have no strong opinions.

src/expr/type_node.h
```diff
@@ -674,6 +674,9 @@ class TypeNode {
   /** Is this a sort constructor kind */
   bool isSortConstructor() const;
 
+  /** Get sort constructor arity */
+  unsigned getSortConstructorArity() const;
```
4tXJ7f, line 678:
```suggestion
  uint64_t getSortConstructorArity() const;
```

src/expr/dtype_cons.cpp
```diff
@@ -595,9 +608,9 @@ TypeNode DTypeConstructor::doParametricSubstitution(
     children.push_back(
         doParametricSubstitution((*i), paramTypes, paramReplacements));
   }
-  for (size_t i = 0, psize = paramTypes.size(); i < psize; ++i)
+  for (unsigned i = 0; i < paramTypes.size(); ++i)
```
4tXJ7f, line 611: Why this change?

ajreynol: This was unintentional, ill revert.

PR #4494, opened 2020-05-14
Add a method for retrieving base of a constant array through API

makaimann:


src/api/cvc4cpp.h
```diff
@@ -883,6 +883,13 @@ class CVC4_PUBLIC Term
    */
   bool isConst() const;
 
+  /**
+   *  Return the base (element stored at all indices) of a constant array
+   *  throws an exception if the kind is not STORE_ALL
+   *  @return the base value
+   */
+  Term getStoreAllBase() const;
```
mpreiner, line 891: I'm not sure if we should call the API function like this. StoreAll is the internal name, but since it's a constant array we should probably use a name that includes constarray? Maybe something like `getConstArrayBase`?

makaimann: Yeah that's fair! I also like your name better, but I thought it would be good to match the `Kind` name. Because the user would do something like `if (t.getKind() == STORE_ALL)`. What do you think? It can also just be documented well so it's clear how to use it.

mpreiner: Oh, I missed that the exported kind is also STORE_ALL. In this case it's probably better to leave the name as it is now.

aniemetz: Maybe we should change the name of the exported kind then. It is kinda ugly anyways.

makaimann: Yeah that makes sense to me. Made the change here: 4aed6b547cd2866ce64f3f5c766c43bfbbe7a70a. We'll see if Travis passes.

PR #4496, opened 2020-05-15
Fix getKind for Python bindings

makaimann:
I noticed recently that getKind for Op and Term was not correct in the python bindings. I added data structures to keep track of the Kind objects and the Python names (which are different from the C++ Kind names). Now, in `getKind` it can just pass the integer representation when construction a `kind` object (the Cython wrapper for a Kind).

### DIFF
--- a/src/api/python/genkinds.py
+++ b/src/api/python/genkinds.py
@@ -69,11 +69,21 @@ from cvc4kinds cimport *
 import sys
 from types import ModuleType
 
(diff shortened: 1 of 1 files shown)


makaimann:
Yeah that's a good idea. Unfortunately there's not actually any specific tests for python yet. But it definitely makes sense to start a unit test suite. I usually use pytest. Does that work?

PR #4504, opened 2020-05-20
Fix missing check for cardinality one in unconstrained simplifier

### LINKED ISSUE #4482: (unconstrained-simp) Soundness bug on formula (assert (distinct a mkTuple))
Hi,
```
[506] % cvc4 small.smt2
unsat
[507] % cvc4 --unconstrained-simp small.smt2
sat
[508] % 
[508] % cat small.smt2
(set-logic ALL)
(declare-fun a () Tuple)
(assert (distinct a mkTuple))
(check-sat)
[509] %
```

OS: Ubuntu 18.04
Commit: 00badd3

ajreynol:
Fixes #4482.


src/preprocessing/passes/unconstrained_simplifier.cpp
```diff
@@ -236,6 +236,10 @@ void UnconstrainedSimplifier::processUnconstrained()
               break;
             }
           }
+          if (parent[0].getType().getCardinality().isOne())
+          {
+            break;
+          }
```
4tXJ7f, line 242: Do we still need the check of `dt.isRecursiveSingleton(tn.toType())` below or does this check subsume that?

ajreynol: We still need it.  This is a really tricky case: this handles infinite streams having uninterpreted sorts (which may or may not have cardinality one, which cause the stream to have either infinite cardinality or cardinality one).

PR #4511, opened 2020-05-21
(proof-new) Update proof checker.

ajreynol:
This adds new required features to proof checker and the base class of proof rule checker. 

This is required as the first dependency towards merging further infrastructure related to proofs.


src/expr/proof_checker.h
```diff
@@ -36,17 +39,67 @@ class ProofRuleChecker
...
+  /** Single arg version */
+  Node checkChildrenArg(PfRule id, const std::vector<Node>& children, Node arg);
+  /** No arg version */
+  Node checkChildren(PfRule id, const std::vector<Node>& children);
+  /** Single child only version */
+  Node checkChild(PfRule id, Node child);
+  /** Single argument only version */
+  Node checkArg(PfRule id, Node arg);
+
+  /** mk and node */
```
HanielB, line 66: capitalize "and" to avoid confusion

src/expr/proof_checker.h
```diff
@@ -36,17 +39,67 @@ class ProofRuleChecker
...
+  static Node mkAnd(const std::vector<Node>& a);
+  /** get an index from a node, return false if we fail */
+  static bool getIndex(TNode n, uint32_t& i);
```
HanielB, line 69: In retrospect I think it's better to not specialize this for "index". We could be using args for something else. So I think it's better to change to `getConstInt`.

src/expr/proof_checker.h
```diff
@@ -81,12 +134,43 @@ class ProofChecker
              const std::vector<std::shared_ptr<ProofNode>>& children,
              const std::vector<Node>& args,
              Node expected = Node::null());
+  /**
+   * Same as above, without conclusions instead of proof node children. This
+   * is used for debugging.
```
HanielB, line 139: Please clarify the difference and why this is more suited for debugging.

src/expr/proof_checker.cpp
```diff
@@ -14,8 +14,100 @@
...
+                             const std::vector<Node>& args)
+{
+  // convert to witness form
+  std::vector<Node> childrenw = children;
+  std::vector<Node> argsw = args;
+  ProofSkolemCache::convertToWitnessFormVec(childrenw);
+  ProofSkolemCache::convertToWitnessFormVec(argsw);
+  Node res = checkInternal(id, childrenw, argsw);
+  // res is in terms of witness form, convert back to Skolem form
+  return ProofSkolemCache::getSkolemForm(res);
```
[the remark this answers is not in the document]
ajreynol: Not easily.  I don't want the user of the class to worry about witness vs. skolem form. Thus witness form is always protected from the user.  We could think about optimizations later.

src/expr/proof_checker.h
ajreynol: I'll call it getUInt32

src/expr/proof_checker.cpp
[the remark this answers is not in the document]
HanielB: Sounds good

src/expr/proof_checker.h
HanielB: sounds good

PR #4513, opened 2020-05-22
(proof-new) Add rewrite corresponding to regular expression inclusion

ajreynol:
This introduces a rewrite based on regular expression inclusion (using calls to the RegExpEntail utility function).

This allows us to justify the regular expression inclusion inference as a rewrite.


src/theory/strings/sequences_rewriter.cpp
```diff
@@ -951,20 +952,48 @@ Node SequencesRewriter::rewriteAndOrRegExp(TNode node)
     else if (std::find(node_vec.begin(), node_vec.end(), ni) == node_vec.end())
     {
       node_vec.push_back(ni);
+      unsigned pindex = ni.getKind() == REGEXP_COMPLEMENT ? 1 : 0;
```
4tXJ7f, line 955:
```suggestion
      uint32_t pindex = ni.getKind() == REGEXP_COMPLEMENT ? 1 : 0;
```

PR #4514, opened 2020-05-22
Add support for SAT solver Kissat.

aniemetz:


src/prop/kissat.cpp
```diff
@@ -0,0 +1,179 @@
...
+}
+
+SatValue KissatSolver::solve(long unsigned int&)
+{
+  Unimplemented() << "Setting limits for Kissat not supported yet";
+};
+
+SatValue KissatSolver::solve(const std::vector<SatLiteral>& assumptions)
+{
+  Unimplemented() << "Incrementel solving with Kissat not supported yet";
```
4tXJ7f, line 130:
```suggestion
  Unimplemented() << "Incremental solving with Kissat not supported yet";
```

src/prop/kissat.h
```diff
@@ -0,0 +1,102 @@
...
+  void init();
+
+  kissat* d_solver;
+
+  unsigned d_nextVarIdx;
+  bool d_okay;
+  SatVariable d_true;
+  SatVariable d_false;
+
+  struct Statistics
```
4tXJ7f, line 84: [Nested classes/structs go before methods/data members](https://github.com/CVC4/CVC4/wiki/Code-Style-Guidelines#declaration-order).

PR #4515, opened 2020-05-22
Cryptominisat: Clean up initialization on creation.

aniemetz:


4tXJ7f:
Actually, since we are already moving the data members, maybe we could also move the `Statistics` class up (see comment on Kissat PR).

PR #4519, opened 2020-05-22
Refactor operator elimination in arithmetic

### LINKED ISSUE #4481: Soundness bug --unconstrained-simp
```
[667] % z3 small.smt2
unsat
[668] % cvc4 small.smt2
unsat
[669] % cvc4 --unconstrained-simp small.smt2
sat
[670] % 
[670] % cat small.smt2
(set-logic ALL)
(declare-fun a () Int)
(assert (= (div a 2) 2))
(assert (= (div a 3) 0))
(check-sat)
[671] %
```

OS: Ubuntu 18.04
Commit: 00badd3

### LINKED ISSUE #4484: (unconstrained-simp) Soundness bug on QF_NIRA formula
Hi,
```
[505] % z3 small.smt2
unsat
[506] % cvc4 small.smt2
unsat
[507] % cvc4 --unconstrained-simp small.smt2
sat
[508] % 
[508] % cat small.smt2
(set-logic QF_NIRA)
(declare-fun a () Real)
(assert (= (to_int a) 2))
(assert (= (to_int (/ a 2.0)) 2))
(check-sat)
[509] %
```

OS: Ubuntu 18.04
Commit: 00badd3

### LINKED ISSUE #4486: cvc4 soundness bug --unconstrained-simp with is_int
```
[505] % z3 small.smt2 
sat
[506] % cvc4 small.smt2 
sat
[507] % cvc4 --unconstrained-simp small.smt2 
unsat
[508] % 
[508] % cat small.smt2 
(set-logic ALL)
(declare-fun x () Real)
(assert (is_int x))
(assert (is_int (+ x 1)))
(check-sat)
[509] %
```

OS: Ubuntu 18.04
Commit: 00badd3

ajreynol:
This is a major refactor of how operators are eliminated in arithmetic.  Currently there were (at least) two things wrong:
(1) ppRewriteTerm sent lemmas on the output channel. This behavior is incompatible with how preprocessing works. In particular, this caused unconstrained simplification to be unaware of terms from such lemmas, leading to incorrect "sat" answers.
(2) Lemmas used to eliminate certain "div-like" terms were processed in a context-independent way. However, lemmas should be cached in a user-context-dependent way. This was leading to incorrect "sat" answers in incremental.

The solution to these issues is to eliminate operators via the construction of `witness` terms. No lemmas are sent out, and instead these lemmas are the consequence of term formula removal in the standard way.

As a result of the refactor, 2 quantifiers regressions time out due to infinite branch and bound issues (one only during --check-unsat-cores).  These appear to be random and I've changed the options to avoid these issues.  3 others now have check-model warnings, which I've added `--quiet` to.  Improving check-model will be addressed on a future PR.

This PR is not required for SMT COMP since we have workarounds that avoid both the incorrect behaviors in our scripts. 

Also notice that --rewrite-divk is effectively now enabled by default always. 

Fixes #4484, fixes #4486, fixes #4481.


src/theory/arith/theory_arith_private.cpp
```diff
@@ -1176,126 +1083,443 @@ Node TheoryArithPrivate::ppRewriteTerms(TNode n) {
...
+      break;
+    }
+
+    case kind::ABS:
+    {
+      return nm->mkNode(kind::ITE,
+                        nm->mkNode(kind::LT, node[0], nm->mkConst(Rational(0))),
+                        nm->mkNode(kind::UMINUS, node[0]),
+                        node[0]);
+      break;
```
4tXJ7f, line 1431: This could also be a rewrite, no?

ajreynol: Yes, let's address in later PR.

PR #4522, opened 2020-05-22
Use prenex normal form when using cegqi-nested-qe

ajreynol:
Previously, we used a specialized variant of prenex normal form that allowed top level disjunctions. However, the method to put quantifiers into this form led to variable shadowing on some benchmarks in SMT-LIB LRA.

This simplifies the code so that we use standard prenex normal form when cegqi-nested-qe is used and deletes the old variant (DISJ_NORMAL).


ajreynol:
Will run performance tests on this, marking work-in-progress for now.

4tXJ7f:
After talking to @ajreynol offline, we have decided not to include this change in the SMT-COMP submission. I have updated the milestone accordingly.

ajreynol:
Unfortunately this PR leads to worse performance overall on LIA/LRA/BV:
```
               CVC4-052220-fpre_cbqi_nqe    6595   2396.97    1117   1231.94      7712      8872
           CVC4-050220-a_cbqi_nqe_rename    6631   2305.54    1126   1286.89      7757      8872



For unsatisfiable : 
           CVC4-050220-a_cbqi_nqe_rename       X      45
               CVC4-052220-fpre_cbqi_nqe       9       X

For satisfiable : 
           CVC4-050220-a_cbqi_nqe_rename       X      15
               CVC4-052220-fpre_cbqi_nqe       6       X
```

ajreynol:
I think I'm fine with merging this PR for the release. I've opened https://github.com/CVC4/cvc4-wishues/issues/73 to say we should reimplement the option deleted here.

PR #4526, opened 2020-05-23
Add the Expr-level sequence datatype

ajreynol:
This adds the expr-level sequence datatypes.  This is required for further progress on sequences.  However, note that this class may be deleted in the further when the Expr level is replaced.

### DIFF
--- a/src/expr/expr_sequence.cpp
+++ b/src/expr/expr_sequence.cpp
@@ -91,8 +91,8 @@ std::ostream& operator<<(std::ostream& os, const ExprSequence& s)
 
 size_t ExprSequenceHashFunction::operator()(const ExprSequence& es) const
 {
-  return TypeHashFunction()(es.getType())
-         + SequenceHashFunction()(es.getSequence());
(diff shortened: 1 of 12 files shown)


src/expr/expr_sequence.cpp
```diff
@@ -0,0 +1,98 @@
...
+std::ostream& operator<<(std::ostream& os, const ExprSequence& s)
+{
+  return os << "__expr_sequence__(" << s.getType() << ", " << s.getSequence()
+            << ")";
+}
+
+size_t ExprSequenceHashFunction::operator()(const ExprSequence& es) const
+{
+  return TypeHashFunction()(es.getType())
+         + SequenceHashFunction()(es.getSequence());
```
4tXJ7f, line 95: [Use the pair hash function to get a better hash](https://github.com/CVC4/CVC4/blob/master/src/util/hash.h#L60-L66). Or we could change the hash function in `Sequence` to take the hash of the type into account and then simply return the sequence hash here.

PR #4533, opened 2020-05-26
(proof-new) Proof step buffer utilities

ajreynol:
This class is used for delaying proof node creation until it is necessary.


src/expr/proof_step_buffer.h
```diff
@@ -0,0 +1,91 @@
...
+  ProofStep(PfRule r,
+            const std::vector<Node>& children,
+            const std::vector<Node>& args);
+  /** The proof rule */
+  PfRule d_rule;
+  /** The proof children */
+  std::vector<Node> d_children;
+  /** The proof arguments */
+  std::vector<Node> d_args;
+};
```
HanielB, line 42: Why isn't the conclusion of the step part of the proof step? In the usage below it seems everything would just be simpler if this was the case.

ajreynol: I'd like to separate premises from conclusion, a proof step is everything that is necessary for deriving a conclusion.

ajreynol: In particular, one reason for this separation: I am often mapping Node (conclusions) to their corresponding ProofStep.  Having conclusion within the ProofStep in this case is redundant.

HanielB: OK, that sounds good. Could you just add this motivation for the doc of the step? Might be good to have this motivation said explicitly.

PR #4542, opened 2020-05-31
Do not cache operator eliminations in arith

### LINKED ISSUE #4525: (check-unsat-cores) Internal errror on unsat core with undefined fns
Hi, I'm getting an internal error with the following formula with two divide-by-zeros:
[bug.smt2.gz](https://github.com/CVC4/CVC4/files/4671671/bug.smt2.gz)
```
% cat bug.smt2                
(assert (> (cot 0.0) (/ 1 0)))
(check-sat)

% cvc4 --check-unsat-cores bug.smt2 -vv
cvc4 --check-models bug.smt2                            
...
solving...
SmtEngine::checkUnsatCore(): generating unsat core
SmtEngine::checkUnsatCore(): pushing core assertions (size == 1)
SmtEngine::checkUnsatCore(): pushing core member (> (cot (/ 0 1)) (/ 1 0))
SmtEngine: setting unsatCores
...
converting to CNF...
+ (not (>= (+ (* (- 1.0) (divByZero 1.0)) (divByZero 1.0)) 0.0))
+ true
solving...
...
SmtEngine::checkUnsatCore(): result is sat
Fatal failure within void CVC4::SmtEngine::checkUnsatCore() at /home/dylan/git/constant-swap/scripts/.solvers/cvc4/src/smt/smt_engine.cpp:2487
Internal error detectedSmtEngine::checkUnsatCore(): produced core was satisfiable.
[1]    2072598 abort (core dumped)  cvc4 --check-unsat-cores bug.smt2
```

**Command line arguments**: --check-unsat-cores
**CVC4 version/commit**: 257269e
**Operating system**: Kernel: 4.19.121-1-MANJARO 
**`configure.sh` options**: --assertions
**`configure.sh` output**: happy to provide upon request

4tXJ7f:
Fixes #4525. The actual problem in the issue is not that the unsat core
is satisfiable but that our unsat core check is not working as intended.
Our unsat core check uses the same `ExprManager` as the main `SmtEngine`
and thus also shares the same attributes for nodes. However, since we
have a different `SmtEngine` instance for the check, we also have
different instances of `TheoryArith`. This leads to the check failing
due to the following:

- Our only input assertion is `(> (cot 0.0) (/ 1 0)))`.
- `TheoryArith::expandDefinition()` gets called on `(> (cot 0.0) (/ 1
  0))` but nothing happens.
- `TheoryArith::expandDefinition()` gets called on `(/ 1 0)`, which gets
  expanded as expected but no attribute is set because it happens in a
  simple `TheoryArith::eliminateOperators()` call.
- `TheoryArith::expandDefinition()` on `(cot (/ 0 1))` first expands to
  `(/ 1 0)` (not cached) and then we expand it recursively to the
  expected term and cache `(/ 1 0) ---> (ite (= 0 0) (divByZero 1.0) (/
  1 0))`.

Up to this point, things are suboptimal but there are no correctness
issues. The problem starts when we do the same process in the unsat core
check:

- Our only input assertion is again `(> (cot 0.0) (/ 1 0)))`.
- `TheoryArith::expandDefinition()` gets called on `(> (cot 0.0) (/ 1
  0))` but nothing happens.
- `TheoryArith::expandDefinition()` gets called on `(/ 1 0)`, which gets
  expanded as expected but no attribute is set or checked because it
  happens in a simple `TheoryArith::eliminateOperators()` call. Note
  that the skolem introduced here for the division-by-zero case is
  different from the skolem introduced above because this is in a
  different `TheoryArith` instance that does not know the existing
  skolems.
- `TheoryArith::expandDefinition()` on `(cot (/ 0 1))` first expands to
  `(/ 1 0)` (not cached) and then we use the cache from our solving call
  to expand it `(/ 1 0) ---> (ite (= 0 0) (divByZero 1.0) (/ 1 0))`.
  Note that `divByZero` here is the skolem from the main solver.

As a result, the preprocessed assertions mix skolems from the main
`SmtEngine` with the `SmtEngine` of the unsat core check, making the
constraints satisfiable.

To solve this problem, this commit removes the caching-by-attribute
mechanism. The reason for removing it is that it is currently
ineffective (since most eliminations do not seem to be cached) and there
are caches at other levels, e.g. when expanding definitions. If we deem
the operator elimination to be a bottleneck, we can introduce a similar
mechanism at the level of `TheoryArith`.

### DIFF
--- a/src/theory/arith/theory_arith_private.cpp
+++ b/src/theory/arith/theory_arith_private.cpp
@@ -1112,14 +1112,8 @@ void TheoryArithPrivate::checkNonLinearLogic(Node term)
   }
 }
 
-struct ArithElimOpAttributeId
-{
-};
-typedef expr::Attribute<ArithElimOpAttributeId, Node> ArithElimOpAttribute;
-
 Node TheoryArithPrivate::eliminateOperatorsRec(Node n)
 {
-  ArithElimOpAttribute aeoa;
   Trace("arith-elim") << "Begin elim: " << n << std::endl;
   NodeManager* nm = NodeManager::currentNM();
   std::unordered_map<Node, Node, TNodeHashFunction> visited;
@@ -1138,18 +1132,11 @@ Node TheoryArithPrivate::eliminateOperatorsRec(Node n)
     }
     else if (it == visited.end())
     {
-      if (cur.hasAttribute(aeoa))
+      visited[cur] = Node::null();
+      visit.push_back(cur);
+      for (const Node& cn : cur)
       {
-        visited[cur] = cur.getAttribute(aeoa);
-      }
-      else
-      {
-        visited[cur] = Node::null();
-        visit.push_back(cur);
-        for (const Node& cn : cur)
-        {
-          visit.push_back(cn);
-        }
+        visit.push_back(cn);
       }
     }
     else if (it->second.isNull())
@@ -1180,7 +1167,6 @@ Node TheoryArithPrivate::eliminateOperatorsRec(Node n)
         // are defined in terms of other non-standard operators.
         ret = eliminateOperatorsRec(retElim);
       }
-      cur.setAttribute(aeoa, ret);
       visited[cur] = ret;
     }
   } while (!visit.empty());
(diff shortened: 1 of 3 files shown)


ajreynol:
Thanks a lot for looking into this.

I agree with the solution in the PR for now.

The more permanent fix will be to not cache the skolems in TheoryArith.  In the new system of generating skolems, these should be cached at the same level of attributes, which would also fix this issue.

PR #4543, opened 2020-05-31
Incorporate sequences into the word interface

ajreynol:
Also renames a function mkWord -> mkWordFlatten.

### DIFF
--- a/src/theory/strings/word.cpp
+++ b/src/theory/strings/word.cpp
@@ -32,7 +32,7 @@ Node Word::mkEmptyWord(TypeNode tn)
   else if (tn.isSequence())
   {
     std::vector<Expr> seq;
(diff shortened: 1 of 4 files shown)


src/theory/strings/word.cpp
```diff
@@ -28,23 +29,27 @@ Node Word::mkEmptyWord(TypeNode tn)
   {
     return mkEmptyWord(CONST_STRING);
   }
+  else if (tn.isSequence())
+  {
+    std::vector<Expr> seq;
+    return NodeManager::currentNM()->mkConst(ExprSequence(tn.toType(), seq));
```
4tXJ7f, line 35:
```suggestion
    return NodeManager::currentNM()->mkConst(ExprSequence(tn.getSequenceElementType().toType(), seq));
```

PR #4549, opened 2020-06-02
New C++ API: Keep reference to solver object in non-solver objects.

aniemetz:
This is in preparation for adding guards to ensure that sort and term
arguments belong to the same solver.

@makaimann

### DIFF
--- a/src/api/cvc4cpp.cpp
+++ b/src/api/cvc4cpp.cpp
@@ -2198,9 +2198,8 @@ DatatypeConstructor Datatype::getConstructorForName(
 Datatype::const_iterator::const_iterator(const Solver* slv,
                                          const CVC4::Datatype& dtype,
                                          bool begin)
+    : d_solver(slv), d_int_ctors(dtype.getConstructors())
 {
-  d_solver = slv;
-  d_int_ctors = dtype.getConstructors();
   const std::vector<CVC4::DatatypeConstructor>* cons =
       static_cast<const std::vector<CVC4::DatatypeConstructor>*>(d_int_ctors);
   for (const auto& c : *cons)
--- a/src/api/cvc4cpp.cpp
+++ b/src/api/cvc4cpp.cpp
@@ -702,6 +702,8 @@ class CVC4ApiExceptionStream
   CVC4_API_CHECK(isDefinedKind(kind)) \
       << "Invalid kind '" << kindToString(kind) << "'";
 
+#define CVC4_API_SORT_CHECK_SOLVER(sort)
+
 #define CVC4_API_KIND_CHECK_EXPECTED(cond, kind) \
   CVC4_PREDICT_TRUE(cond)                        \
   ? (void)0                                      \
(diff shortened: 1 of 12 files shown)


src/api/cvc4cpp.cpp
```diff
@@ -2169,28 +2192,34 @@ DatatypeConstructor Datatype::getConstructorForName(
...
-  return (*d_dtype)[index];
+  return DatatypeConstructor(d_solver, (*d_dtype)[index]);
 }
 
-Datatype::const_iterator::const_iterator(const CVC4::Datatype& dtype,
+Datatype::const_iterator::const_iterator(const Solver* slv,
+                                         const CVC4::Datatype& dtype,
                                          bool begin)
 {
+  d_solver = slv;
```
4tXJ7f, line 2202: Minor: Could be moved to initializer list.

src/parser/parser.cpp
```diff
@@ -166,7 +161,7 @@ api::Sort Parser::getSort(const std::string& name)
 {
   checkDeclaration(name, CHECK_DECLARED, SYM_SORT);
   assert(isDeclared(name, SYM_SORT));
-  api::Sort t = api::Sort(d_symtab->lookupType(name));
+  api::Sort t = api::Sort(d_solver, d_symtab->lookupType(name));
```
4tXJ7f, line 164: Just a minor note on these kinds of assignments, we can actually just write `api::Sort t(d_solver, d_symtab->lookupType(name));` here. Probably out of scope for this PR but just something I noticed.

src/parser/tptp/Tptp.g
```diff
@@ -1437,12 +1437,12 @@ tffLetTermBinding[std::vector<CVC4::api::Term> & bvlist,
                   CVC4::api::Term& lhs,
                   CVC4::api::Term& rhs]
   : term[lhs] EQUAL_TOK term[rhs]
-  {
-    PARSER_STATE->checkLetBinding(bvlist, lhs, rhs, false);
-    std::vector<api::Term> lchildren(++lhs.begin(), lhs.end());
-    rhs = MK_TERM(api::LAMBDA, MK_TERM(api::BOUND_VAR_LIST, lchildren), rhs);
-    lhs = api::Term(lhs.getExpr().getOperator());
-  }
+{
```
4tXJ7f, line 1440: This is very nit-picky but the indentation here is off: the curly brackets should align with the beginning of `term[lhs]` above.

aniemetz: Clang-format screwed with this one, did we ever decide what to do to resolve this problem? Switch it off for the .g files?

src/api/cvc4cpp.cpp
aniemetz: Yeah, didn't move it because the rest is also initialized in the body.

src/parser/tptp/Tptp.g
ajreynol: We don't have a good solution.  I think we decided to do our best effort to keep things close to clang-formatted everywhere, but don't require it if its a major effort.

PR #4550, opened 2020-06-02
Fix handling of Boolean term variables

### LINKED ISSUE #4446: Invalid model with QF_AUFLIA formula
```
[633] % cvc4 --check-models small.smt2
Fatal failure within void CVC4::TheoryEngine::checkTheoryAssertionsWithModel(bool) at theory/theory_engine.cpp:2283
Internal error detectedTHEORY_UF has an asserted fact that the model doesn't satisfy.
The fact: (a BOOLEAN_TERM_VARIABLE_275)
Model value: false
Aborted
[634] % cat small.smt2
(set-logic QF_AUFLIA)
(declare-fun a (Bool) Bool)
(declare-fun b () Bool)
(declare-fun c () Bool)
(assert (or (a b) (a c)))
(check-sat)
[635] %
```

OS: Ubuntu 18.04
Commit: 8236d7f

4tXJ7f:
Fixes #4446. This commit fixes two issues related to the handling of
Boolean term variables:

- Removing Boolean subterms and replacing them with a Boolean term
  variable introduces an equality of the form `(= v t)` where `v` is the
  Boolean term variable and `t` is the term. It is important that these
  equalities do not get removed. This commit changes
  `Theory::isLegalElimination()` to take this into account.
- The incorrect model reported in the issue was caused by some of the
  Boolean term variables not being assigned values by the SAT solver
  when we decided that the benchmark is satisfiable. Our justification
  heuristic (correctly) decided that it is enough to satisfy one of the
  disjuncts in the assertion to satisfy the whole assertion. However,
  the assignments that we received from the SAT solver restricted us to
  pick a specific value for one of the Boolean constants:

  ```
  Literal | Value | Expr
  ---------------------------------------------------------
  '7 ' 0 c
  '0 ' 1 true
  '1 ' 0 false
  '2 ' 0 (a BOOLEAN_TERM_VARIABLE_274)
  '5 ' _ b
  '3 ' 1 (a BOOLEAN_TERM_VARIABLE_277)
  '4 ' _ BOOLEAN_TERM_VARIABLE_274
  '6 ' 0 BOOLEAN_TERM_VARIABLE_277
  ```

  This meant that we had to pick `true` for `BOOLEAN_TERM_VARIABLE_274`
  but we picked `false` since we simply treated it as an unassigned
  variable. In general, the justification heuristic handles embedded
  skolems introduced by term removal first and asks the SAT solver to
  decide on Boolean term variables. However, for some logics, such as
  `QF_AUFLIA`, we use the justification heuristic not for decisions but
  only to decide when to stop, so those decisions were not done. This
  commit introduces a conservative fix that adds a check after
  satsifying all the assertions that makes sure that the equalities
  introduced for Boolean terms are satisfied as well. Due to the eager
  treatment of Boolean term variables when we use the justification
  heuristic also for decisions, it is likely that we don't really have
  the problem in that case but it doesn't hurt to enable the fix. It is
  also possible that this fix is only required with models but it is
  definitely safer to just always enable it (there could be tricky
  corner cases involving the theory combination between UF and Boolean
  terms). Additionally, this commit changes the ITE-specific naming in
  the justification heuristic to reflect more accurately that we are in
  general dealing with skolems introduced by term removals and not only
  due to ITE removal.


ajreynol:
Discussing offline, @4tXJ7f thinks that there is possibly an simpler and better long term fix for this issue.

PR #4551, opened 2020-06-02
Update CEGQI to use lemma status instead of forcing preprocess

ajreynol:
This PR removes a hack in counterexample-guided quantifier instantiation.

The issue is the CEGQI needs to know the form of certain lemmas, after theory preprocessing.  CEGQI needs to know this since it must construct solutions (e.g. solved form of equalities) for free variables in these lemmas, which includes fresh variables in the lemma after postprocess like ITE skolems.

Previously, CEGQI was hacked so that it applied the preprocessing eagerly so that it had full knowledge of the postprocessed lemma.  This PR updates CEGQI so that it uses the returned LemmaStatus as the way of getting the lemma after postprocess.  It also fixes the lemma status returned by TheoryEngine so that all lemmas not just the "main lemma" are returned as a conjunction.

This PR is in preparation for major refactoring to theory preprocessing for the sake of proofs.


src/theory/quantifiers/cegqi/ceg_instantiator.cpp
```diff
@@ -1592,36 +1594,51 @@ void CegInstantiator::registerCounterexampleLemma( std::vector< Node >& lems, st
...
+    }
+    if (std::find(d_vars.begin(), d_vars.end(), ces) != d_vars.end())
+    {
+      // already processed variable
+      continue;
+    }
+    if (ces.getType().isBoolean())
+    {
+      // Boolean variables, including the counterexample literal, don't matter
+      // since the are always assigned model value.
```
mpreiner, line 1632:
```suggestion
      // since they are always assigned a model value.
```

mpreiner:
Labeling as do-no-merge until suggestion is applied.

PR #4554, opened 2020-06-02
Printing FP values as binary or indexed BVs according to option

HanielB:
Changes the printer to print indexed BV values by default, as is the case with
BV values.


4tXJ7f:
Before I forget: It would be good to also update the `NEWS` file to reflect this change.

HanielB:
As discussed offline, it's not clear how this could be refactored. The (partial) duplication seems necessary and also occurs for other types.

PR #4556, opened 2020-06-03
(proof-new) Rename ProofSkolemCache to SkolemManager

ajreynol:
This PR changes the design of ProofSkolemCache so that it has facilities for tracking proofs.  This is required so that the term formula removal pass can be made proof-producing. 

This makes it so that ProofSkolemCache is renamed to SkolemManager, which lives in NodeManager.  Creating (most) skolems must be accompanied by a corresponding ProofGenerator that can provide the proof for the existential corresponding to their witness form.

Further updates will include refactoring the mkSkolemExists method of SkolemManager so that its contract wrt proofs is simpler.

Once all calls to mkSkolem go through the standard interface, then NodeManager::mkSkolem will be moved to SkolemManager::mkSkolemInternal.


src/expr/skolem_manager.h
```diff
@@ -97,21 +103,30 @@ class ProofSkolemCache
...
+                      int flags = NodeManager::SKOLEM_DEFAULT,
+                      ProofGenerator* pg = nullptr);
   /**
    * Same as above, but for special case of (witness ((x T)) (= x t))
    * where T is the type of t. This skolem is unique for each t, which we
    * implement via an attribute on t. This attribute is used to ensure to
    * associate a unique skolem for each t.
+   *
+   * Notice that a purification skolem is trivial to justify, and hence it
+   * does not require a proof generator.
```
HanielB, line 119: Is this worth the special casing?

ajreynol: What do you mean?

HanielB: My impression is that when using the skolem manager I'd have to check whether I have a proof generator or I have to compute myself (in this straightforward case). Wouldn't it be better to be uniform? This should not be saving much.

ajreynol: Then this class would be harder to use.  I would have to update everyone who wants to construct a purification variable to construct a dummy proof generator that does the trivial justification.

HanielB: I see. In that case the other special case is better :)

PR #4557, opened 2020-06-03
New C++ Api: First batch of API guards.

aniemetz:
This is the first batch of API guards, mainly extending existing guards
in the Solver object with checks that Ops, Terms, Sorts, and datatype objects
are associated to this solver object.

This further changes how `DatatypeConstructorDecl` objects are created. Previously,
they were not created via the Solver object (while `DatatypeDecl` was). Now, they are
created via `Solver::mkDatatypeConstructorDecl`, consistent with how `DatatypeDecl`
objects are created.

### DIFF
--- a/src/api/cvc4cpp.cpp
+++ b/src/api/cvc4cpp.cpp
@@ -702,8 +702,6 @@ class CVC4ApiExceptionStream
   CVC4_API_CHECK(isDefinedKind(kind)) \
       << "Invalid kind '" << kindToString(kind) << "'";
 
-#define CVC4_API_SORT_CHECK_SOLVER(sort)
-
 #define CVC4_API_KIND_CHECK_EXPECTED(cond, kind) \
   CVC4_PREDICT_TRUE(cond)                        \
   ? (void)0                                      \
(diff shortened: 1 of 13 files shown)


makaimann:
Also, FYI @kris-brown. This PR will change the datatypes interface.

src/api/cvc4cpp.cpp
```diff
@@ -2749,13 +2775,27 @@ std::vector<Sort> Solver::mkDatatypeSortsInternal(
...
+  for (size_t i = 0, ndts = dtypedecls.size(); i < ndts; ++i)
+  {
+    CVC4_API_ARG_AT_INDEX_CHECK_EXPECTED(this == dtypedecls[i].d_solver,
+                                         "datatype declaration",
+                                         dtypedecls[i],
+                                         i)
+        << "a datatype declaration associated to this solver object";
+    CVC4_API_CHECK(this == dtypedecls[i].d_solver)
+        << "Given datatype declaration at index " << i
+        << " is not associated with this solver";
```
4tXJ7f, line 2787: It looks like these two checks are the same? Am I missing something?

aniemetz: Indeed. The second was replaced by the first but I forgot to delete it.

aniemetz:
Don't merge yet, we have a segfault in the datatypes Python example.

PR #4559, opened 2020-06-03
(proof-new) Adding rules and proof checker for EUF

HanielB:


src/expr/node_manager.h
```diff
@@ -444,6 +444,9 @@ class NodeManager {
   /** Get a Kind from an operator expression */
   static inline Kind operatorToKind(TNode n);
 
+  /** Get corresponding application kindfor function */
```
ajreynol, line 447: kindfor --> kind for.
Also can you give some documentation for why this function is necessary?

src/expr/node_manager.h
```diff
@@ -1236,6 +1239,28 @@ inline Kind NodeManager::operatorToKind(TNode n) {
   return kind::operatorToKind(n.d_nv);
 }
 
+inline Kind NodeManager::getKindForFunction(TNode fun)
```
ajreynol, line 1242: Move this to cpp.

src/expr/node_manager.h
```diff
@@ -444,6 +444,9 @@ class NodeManager {
   /** Get a Kind from an operator expression */
   static inline Kind operatorToKind(TNode n);
 
+  /** Get corresponding application kindfor function */
+  static inline Kind getKindForFunction(TNode fun);
```
ajreynol, line 448: We don't use `inline` anymore. This is assumed to be managed by the compiler.

HanielB: Missed that memo. :) Will fix.

src/expr/node_manager.h
```diff
@@ -444,7 +444,18 @@ class NodeManager {
...
-  /** Get corresponding application kindfor function */
+  /** Get corresponding application kind for function
+   *
+   * Different functional nodes are applied differently. For example,
+   * uninterpreted functions are applied via APPLY_UF, while constructors via
+   * APPLY_CONSTRUCTOR. This method provides the correct application according
+   * to which kind of function fun is.
+   *
+   * @param fun The functional node
+   * @return the correct application kind for fun. If fun is not an
```
ajreynol, line 455: Actually this isn't quite accurate, it just depends on the type. E.g. fun doesn't have to be an uninterpreted function, it just has to have function type?

HanielB: You're right. But then the more fundamental question is: who has "function type"?

ajreynol: well the point is that this method is based on type, not term, so just describe it that way.  You actually might even consider having this method take a TypeNode not a Node.

HanielB: I changed the doc. I think it makes sense for this to live in node manager as it refers to "how do I apply this node". Just so happens that it's solely based on its type (which whoever is calling does not necessarily need to know).

PR #4561, opened 2020-06-03
Changing default language

HanielB:
Useful to avoid issues when a language is not set and it cannot be easily inferred (for example via the API). Since the language that covers most operators in CVC4 is the SMT one we use that as default now.


HanielB:
There is one failing regression due to printing of BV constants. Merging this after #4554 will take care of this, as that PR changes the default behavior of the SMT printer.

PR #4562, opened 2020-06-04
Add support for eqrange predicate

mpreiner:
This commit adds support for an `eqrange` predicate. `(eqrange a b i j)` is true if arrays `a` and `b` is equal on all indices within indices `i` and `j`, i.e., `\forall k . i <= k <= j --> a[k] = b[k]`. Requires option `--arrays-exp` to be enabled.


src/smt/set_defaults.cpp
```diff
@@ -267,6 +267,21 @@ void setDefaults(SmtEngine& smte, LogicInfo& logic)
...
+      logic = logic.getUnlockedCopy();
+      logic.enableQuantifiers();
+      logic.lock();
+    }
+    if (!options::fmfBound.wasSetByUser())
+    {
+      options::fmfBound.set(true);
+      Trace("smt") << "turning on fmf-bound, for arrays-exp" << std::endl;
+    }
+  }
```
HanielB, line 283: Document why this is necessary

test/regress/regress0/eqrange1.smt2
```diff
@@ -0,0 +1,15 @@
+(set-logic QF_AUFBV)
+(set-option :arrays-exp true)
+(set-option :quiet true) ; Suppress Warning
```
HanielB, line 3: what warning?

src/theory/arrays/theory_arrays_type_rules.h
```diff
@@ -224,6 +224,49 @@ struct ArrayPartialSelectTypeRule {
...
+        throw TypeCheckingExceptionPrivate(
+            n, "eqrange lower index type does not match array index type");
+      }
+      if (!indexRangeType2.isSubtypeOf(indexType))
+      {
+        throw TypeCheckingExceptionPrivate(
+            n, "eqrange upper index type does not match array index type");
+      }
+    }
+    return nodeManager->booleanType();
```
ajreynol, line 266: There should probably be an additional check here to ensure the type of `indexRangeType1` and `indexRangeType2` are handled (bv/fp/int/real). Either that or the semantics should incorporate types in a generic way (e.g. these arguments are ignored for all other types).

test/regress/regress0/eqrange1.smt2
ajreynol: This is to suppress a warning about check-model involving quantifiers I believe.

HanielB: I see. In this case it might be good to add a comment to the regression.

src/smt/set_defaults.cpp
mpreiner: It's not really necessary, but allows to answer `sat` in more cases (by default).

mpreiner:
> One more change: can you add smt2 printing support for eqrange?

Done

PR #4573, opened 2020-06-05
Use NlLemma utility for all lemmas in non-linear.

ajreynol:
This makes it much easier to set custom properties of lemmas (e.g. preprocess) and also will allow proof tracking and debugging in the future.

This also enables a fix on the IAND branch related to preprocessing lemmas.

### DIFF
--- a/src/theory/arith/nl/nl_lemma_utils.cpp
+++ b/src/theory/arith/nl/nl_lemma_utils.cpp
@@ -21,6 +21,12 @@ namespace theory {
 namespace arith {
 namespace nl {
 
+std::ostream& operator<<(std::ostream& out, NlLemma& n)
+{
+  out << n.d_lemma;
+  return out;
+}
+
 bool SortNlModel::operator()(Node i, Node j)
 {
   int cv = d_nlm->compare(i, j, d_isConcrete, d_isAbsolute);
--- a/src/theory/arith/nl/nl_lemma_utils.h
+++ b/src/theory/arith/nl/nl_lemma_utils.h
@@ -27,16 +27,24 @@ namespace nl {
 class NlModel;
 
 /**
- * A side effect of adding a lemma in the non-linear solver. This is used
- * to specify how the state of the non-linear solver should update. This
- * includes:
+ * The data structure for a single lemma to process by the non-linear solver,
+ * including the lemma itself and whether it should be preprocessed (see
+ * OutputChannel::lemma).
+ *
+ * This also includes data structures that encapsulate the side effect of adding
+ * this lemma in the non-linear solver. This is used to specify how the state of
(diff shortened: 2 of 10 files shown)


src/theory/arith/nl/nl_lemma_utils.cpp
```diff
@@ -21,6 +21,12 @@ namespace theory {
 namespace arith {
 namespace nl {
 
+std::ostream& operator<<(std::ostream& out, NlLemma& n)
+{
+  out << n.d_lemma;
+  return out;
```
yoni206, line 27: It is this both returned and moved to the parameter?

src/theory/arith/nl/nl_lemma_utils.h
```diff
@@ -27,16 +27,24 @@ namespace nl {
...
- * to specify how the state of the non-linear solver should update. This
- * includes:
+ * The data structure for a single lemma to process by the non-linear solver,
+ * including the lemma itself and whether it should be preprocessed (see
+ * OutputChannel::lemma).
+ *
+ * This also includes data structures that encapsulate the side effect of adding
+ * this lemma in the non-linear solver. This is used to specify how the state of
+ * the non-linear solver should update. This includes:
  * - A set of secant points to record (for transcendental secant plane
```
yoni206, line 37: Does it include anything else?

ahmed-irfan:
is it the case that after merging this PR, LemmaSideEffects will not exit? Instead NlLemma structure will contain not only the lemmas but also its side-effects?

ajreynol:
@ahmed-irfan yes,the NlLemma subsumes the previous NlLemmaSideEffect class.  It includes the lemma itself, instructions for sending out, and all side effects to process.

ajreynol:
@yoni the d_preprocess flag is set to true frequently in the transcendental benchmarks. Our coverage build says that a block of code that sets d_preprocess to true is called 126 times: https://cvc4.cs.stanford.edu/downloads/builds/coverage/cvc4-2020-05-29/index.src_theory_arith_transcendental_solver.cpp.html.

src/theory/arith/nl/nl_lemma_utils.cpp
ajreynol: This is the common way of writing to streams, for another example, see: https://github.com/CVC4/CVC4/blob/master/src/expr/proof_rule.cpp#L96.

src/theory/arith/nl/nl_lemma_utils.h
ajreynol: No, this is the only thing included, for now.  I tend to write the documentation in this way since it is robust to changes (say if we accidentally add a new field but don't add to this list).

PR #4574, opened 2020-06-05
(proof-new) Add abstract proof generator class

ajreynol:
Also adds a commonly used proof generator: the proof reference proof generator.

### DIFF
--- a/src/expr/proof_generator.h
+++ b/src/expr/proof_generator.h
@@ -1,0 +1,115 @@
+/*********************                                                        */
+/*! \file proof_generator.h
+ ** \verbatim
+ ** Top contributors (to current version):
+ **   Andrew Reynolds
+ ** This file is part of the CVC4 project.
+ ** Copyright (c) 2009-2019 by the authors listed in the file AUTHORS
+ ** in the top-level source directory) and their institutional affiliations.
+ ** All rights reserved.  See the file COPYING in the top-level source
+ ** directory for licensing information.\endverbatim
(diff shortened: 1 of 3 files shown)


src/expr/proof_generator.h
```diff
@@ -0,0 +1,115 @@
...
+
+#include "expr/node.h"
+#include "expr/proof.h"
+#include "expr/proof_node.h"
+
+namespace CVC4 {
+
+/**
+ * An abstract proof generator class, to be used in combination with
+ * ProofOutputChannel (see theory/proof_output_channel.h).
```
HanielB, line 28: It's more general than that now, no?

src/expr/proof_generator.h
```diff
@@ -0,0 +1,115 @@
...
+ * multiple instances of ProofGenerator objects, e.g. if it has more than one
+ * way of justifying lemmas or conflicts.
+ *
+ * A proof generator has two main interfaces for generating proofs:
+ * (1) getProofFor, and (2) addProofTo. The latter is optional. If no
+ * implementation is provided, then addProofTo(f, pf) calls getProofFor(f) and
+ * links the top node of the returned proof into pf. Note that addProofTo can be
+ * overridden by an instance of this class as an optimization to call
+ * CDProof::addStep instead of CDProof::addProof for the top-most step of the
+ * proof of f.
```
HanielB, line 41: Can you describe things here abstractly before operationally?

PR #4577, opened 2020-06-06
Add logic check for define-fun(s)-rec

4tXJ7f:
This commit adds a logic check for `define-fun-rec`/`define-funs-rec` at
the level of the new API that checks whether the logic is quantified and
includes UF. To make sure that the parser actually executes that check,
this commit converts the `DefineFunctionRecCommand` command to use the
new API instead of the old one. This temporarily breaks the `exportTo`
functionality for `DefineFunctionRecCommand` but this is not currently
used within the CVC4 code base (and it seems unlikely that users use
commands).


src/api/cvc4cpp.cpp
```diff
@@ -4263,34 +4264,43 @@ Term Solver::defineFun(Term fun,
...
 {
   CVC4_API_SOLVER_TRY_CATCH_BEGIN;
-  CVC4_API_ARG_CHECK_EXPECTED(fun.getSort().isFunction(), fun) << "function";
-  std::vector<Sort> domain_sorts = fun.getSort().getFunctionDomainSorts();
-  size_t size = bound_vars.size();
-  CVC4_API_ARG_SIZE_CHECK_EXPECTED(size == domain_sorts.size(), bound_vars)
-      << "'" << domain_sorts.size() << "'";
-  for (size_t i = 0; i < size; ++i)
+
+  if (fun.getSort().isFunction())
```
aniemetz, line 4268:
```suggestion
  CVC4_API_ARG_CHECK_EXPECTED(fun.getSort().isFunction() || bound_vars.size() == 0, fun)
        << "function or nullary symbol";
```
Can't this simplified like this? (The else case below would of course be deleted then.)

src/api/cvc4cpp.cpp
```diff
@@ -4350,34 +4368,50 @@ Term Solver::defineFunRec(Term fun,
...
-  for (size_t i = 0; i < size; ++i)
+
+  CVC4_API_CHECK(d_smtEngine->getUserLogicInfo().isQuantified())
+      << "recursive function definitions require a logic with quantifiers";
+  CVC4_API_CHECK(
+      d_smtEngine->getUserLogicInfo().isTheoryEnabled(theory::THEORY_UF))
+      << "recursive function definitions require a logic with uninterpreted "
+         "functions";
+
+  if (fun.getSort().isFunction())
```
aniemetz, line 4379: Ditto

src/api/cvc4cpp.cpp
```diff
@@ -4404,38 +4446,46 @@ void Solver::defineFunsRec(const std::vector<Term>& funs,
...
         << "function associated to this solver object";
-    CVC4_API_ARG_CHECK_EXPECTED(fun.getSort().isFunction(), fun) << "function";
     CVC4_API_SOLVER_CHECK_TERM(term);
 
-    std::vector<Sort> domain_sorts = fun.getSort().getFunctionDomainSorts();
-    size_t size = bvars.size();
-    CVC4_API_ARG_SIZE_CHECK_EXPECTED(size == domain_sorts.size(), bvars)
-        << "'" << domain_sorts.size() << "'";
-    for (size_t i = 0; i < size; ++i)
+    if (fun.getSort().isFunction())
```
aniemetz, line 4451: Ditto

src/smt/command.cpp
```diff
@@ -1427,6 +1440,9 @@ void DefineFunctionRecCommand::invoke(SmtEngine* smtEngine)
 Command* DefineFunctionRecCommand::exportTo(
     ExprManager* exprManager, ExprManagerMapCollection& variableMap)
 {
+  Unimplemented();
+
+#if 0
```
aniemetz, line 1445: Why not remove this? Do you plan to reenable it?

src/api/cvc4cpp.cpp
4tXJ7f: Actually, that's what I had at first but it didn't work out. The reason is that `fun.getSort().getFunctionDomainSorts()` throws an error when it is called on a non-function sort.

src/smt/command.cpp
4tXJ7f: Yes, so we will have to decide whether exporting commands is a useful feature and if yes, we will have to implement a feature to export terms from one `Solver` instance to another. 

Another way to temporarily solve this problem would be to add a second overload of `exportTo` that takes a `Solver` instance and an `ExprManagerMapCollection`. The default implementation would simply get the `ExprManager` and call the current variant. For the commands that use the new solver interface, we could then export the `Expr`s of the `Term`s of `d_solver`, wrap them in `Term` instances belonging to the solver that we export to. What do you think?

src/api/cvc4cpp.cpp
aniemetz: Oh, I see. Alright.

src/smt/command.cpp
aniemetz: If we don't need it, we can remove it. We can always add this later if we actually need it.

PR #4580, opened 2020-06-06
(proof-new) Theory proof step buffer utility

ajreynol:
This is a common class for adding steps for theory-specific proof rules into a ProofStepBuffer.

### DIFF
--- a/src/theory/theory_proof_step_buffer.h
+++ b/src/theory/theory_proof_step_buffer.h
@@ -1,0 +1,79 @@
+/*********************                                                        */
+/*! \file theory_proof_step_buffer.h
+ ** \verbatim
+ ** Top contributors (to current version):
+ **   Andrew Reynolds
+ ** This file is part of the CVC4 project.
+ ** Copyright (c) 2009-2019 by the authors listed in the file AUTHORS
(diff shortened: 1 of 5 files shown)


src/theory/theory_proof_step_buffer.h
```diff
@@ -0,0 +1,75 @@
...
+  bool applyPredIntro(Node tgt,
+                      const std::vector<Node>& exp,
+                      MethodId ids = MethodId::SB_DEFAULT,
+                      MethodId idr = MethodId::RW_REWRITE);
+  /**
+   * Apply predicate elimination. This method returns the result of applying
+   * the rule MACRO_SR_PRED_ELIM on src, exp. The returned formula is equivalent
+   * to src assuming exp.
+   */
+  Node applyPredElim(Node src,
```
HanielB, line 65: It's confusing that this method, as the above, tries to do something but this one produces a node and the above a boolean, and the same naming pattern is used for all the methods. Could you change this? You could make this one also return a bool and take another node as argument.

ajreynol: Predicate elimination never fails, thus Node is the right return value.  I will comment on this.

PR #4585, opened 2020-06-08
Cardinality-related inferences per type in theory of strings

ajreynol:
Towards theory of sequences.

This updates various inference steps in the theory of strings that are based on collecting all equivalence classes to be per string-like type.


src/theory/strings/solver_state.cpp
```diff
@@ -331,9 +339,13 @@ void SolverState::separateByLength(const std::vector<Node>& n,
...
   {
-    cols.push_back(std::vector<Node>());
-    cols.back().insert(cols.back().end(), p.second.begin(), p.second.end());
-    lts.push_back(leqc_to_eqc[p.first]);
+    Assert(!p.second.empty());
+    // get the type of the collection
+    TypeNode stn = p.second[0].getType();
+    cols[stn].push_back(std::vector<Node>());
+    cols[stn].back().insert(
+        cols[stn].back().end(), p.second.begin(), p.second.end());
```
4tXJ7f, line 347:
```suggestion
    cols[stn].emplace_back(p.second.begin(), p.second.end());
```

PR #4586, opened 2020-06-08
(proof-new) Refactor skolemization

ajreynol:
This PR refactors skolemization in SkolemManager so that we use a "curried" form, where witnesses for a variable `x` is based on the existential where the prefix up to `x` has already been skolemized.

This additionally adds more utility functions that will be used in the internal proof checker for quantifiers.


src/expr/skolem_manager.h
```diff
@@ -149,13 +169,35 @@ class SkolemManager
    * Mapping from witness terms to proof generators.
    */
   std::map<Node, ProofGenerator*> d_gens;
+  /** Map to canonical bound variables */
```
HanielB, line 172: Say why this is necessary.

src/expr/skolem_manager.h
```diff
@@ -149,13 +169,35 @@ class SkolemManager
...
+   * For example, calling this method on:
+   *   (exists ((x Int) (y Int)) (P x y))
+   * will return:
+   *   (witness ((x Int)) (exists ((y Int)) (P x y)))
+   * If q is not an existentially quantified formula, then null is
+   * returned and an assertion failure is thrown.
+   *
+   * This method additionally updates qskolem to be the skolemized form of q.
+   * In the above example, this is set to:
+   *   (exists ((y Int)) (P (witness ((x Int)) (exists ((y Int)) (P x y))) y))
```
HanielB, line 192: by using `y` here and `y'` in `mkSkolemize` above it becomes confusing whether you're renaming `y` or not. Just using `y'` would suffice.

src/expr/skolem_manager.cpp
```diff
@@ -20,6 +20,7 @@ using namespace CVC4::kind;
 
 namespace CVC4 {
 
+// note that these could be internal maps in SkolemManager
```
HanielB, line 23: Not clear. Please clarify.

src/expr/skolem_manager.cpp
```diff
@@ -50,43 +51,91 @@ Node SkolemManager::mkSkolem(Node v,
   Node predw = getWitnessForm(pred);
   // make the witness term, which should not contain any skolem
   Node w = nm->mkNode(WITNESS, bvl, predw);
-  // store the mapping to proof generator
-  d_gens[w] = pg;
+  // store the mapping to proof generator if it exists
+  if (pg != nullptr)
+  {
+    Node q = nm->mkNode(EXISTS, w[0], w[1]);
+    // notice this may overwrite an existing proof generator
```
HanielB, line 58: clarify why this is not an issue

src/expr/skolem_manager.cpp
```diff
@@ -50,43 +51,91 @@ Node SkolemManager::mkSkolem(Node v,
...
+  {
+    Assert(currQ.getKind() == EXISTS && av == currQ[0][0]);
+    Node sk = skolemize(currQ, currQ, prefix, comment, flags);
+    Trace("sk-manager-debug")
+        << "made skolem " << sk << " for " << av << std::endl;
+    skolems.push_back(sk);
+  }
+  if (pg != nullptr)
+  {
+    // notice this may overwrite an existing proof generator
```
HanielB, line 84: clarify why this is not an issue

src/expr/skolem_manager.cpp
```diff
@@ -50,43 +51,91 @@ Node SkolemManager::mkSkolem(Node v,
...
+                                int flags,
+                                ProofGenerator* pg)
+{
+  Trace("sk-manager-debug") << "mkSkolemize..." << std::endl;
+  Assert(q.getKind() == EXISTS);
+  Node currQ = q;
+  for (const Node& av : q[0])
+  {
+    Assert(currQ.getKind() == EXISTS && av == currQ[0][0]);
+    Node sk = skolemize(currQ, currQ, prefix, comment, flags);
```
HanielB, line 77: please add a comment to say that `currQ` will be update by `skolemize` and this will guarantee the right shape for the next step

src/expr/skolem_manager.h
```diff
@@ -149,13 +169,35 @@ class SkolemManager
...
+   */
+  Node skolemize(Node q,
+                 Node& qskolem,
+                 const std::string& prefix,
+                 const std::string& comment = "",
+                 int flags = NodeManager::SKOLEM_DEFAULT);
+  /** Get or make bound variable */
```
HanielB, line 199: say this uses the canonical mapping above to ensure uniqueness up to `t,s`. Also clarify what should `t` and `s` be.

src/expr/skolem_manager.cpp
```diff
@@ -50,43 +51,91 @@ Node SkolemManager::mkSkolem(Node v,
...
   {
-    if (av == v)
+    if (v.isNull())
     {
-      foundVar = true;
+      v = av;
       continue;
     }
+    // must make fresh variable to avoid shadowing, which is unique per
+    // variable av to ensure that this method is deterministic
```
HanielB, line 110: clarify why it's important that this method is deterministic

src/expr/skolem_manager.h
```diff
@@ -95,20 +95,34 @@ class SkolemManager
...
-   * returned and an assertion failure is thrown.
+   * Make skolemized form of existentially quantified formula q, and store its
+   * Skolems into the argument skolems.
+   *
+   * For example, calling this method on:
+   *   (exists ((x Int) (y Int)) (P x y))
+   * returns:
+   *   (P w1 w2)
+   * where w1 and w2 are skolems with witness forms:
+   *   (witness ((x Int)) (exists ((y' Int)) (P x y')))
```
HanielB, line 106: say you're renaming to avoid shadowing

src/expr/skolem_manager.cpp
```diff
@@ -50,43 +51,91 @@ Node SkolemManager::mkSkolem(Node v,
...
+    Node bvl = nm->mkNode(BOUND_VAR_LIST, ovarsW);
     pred = nm->mkNode(EXISTS, bvl, pred);
+    // skolem form keeps the old variables
+    bvl = nm->mkNode(BOUND_VAR_LIST, ovars);
+    qskolem = nm->mkNode(EXISTS, bvl, pred);
   }
-  return mkSkolem(v, pred, prefix, comment, flags, pg);
+  Trace("sk-manager-debug") << "call sub mkSkolem" << std::endl;
+  // don't use a proof generator, since this may be an intermediate, partially
+  // skolemized formula.
```
HanielB, line 129: Clarify that the proof generator for the whole skolemized formula is used is set in `mkSkolemize`

src/expr/skolem_manager.h
ajreynol: Yes, this should be `y'`.

PR #4588, opened 2020-06-08
(proof-new) Add trust node utility

ajreynol:
This is a core data structure for associating a formula with a proof generator.


src/theory/trust_node.h
```diff
@@ -0,0 +1,148 @@
...
+  LEMMA,
+  PROP_EXP,
+  INVALID
+};
+/**
+ * Converts a proof rule to a string. Note: This function is also used in
+ * `safe_print()`. Changing this function name or signature will result in
+ * `safe_print()` printing "<unsupported>" instead of the proper strings for
+ * the enum values.
+ *
```
alex-ozdemir, line 41:
```suggestion
 * Returns a string with static lifetime: it should not be freed.
```

src/theory/trust_node.h
```diff
@@ -0,0 +1,148 @@
...
+ * OutputChannel::lemma. That justification is intended to be given by the
+ * generator that is required to construct this object.
+ *
+ * They are intended to be constructed by ProofGenerator objects themselves (a
+ * proof generator wraps itself in TrustNode it returns) and passed
+ * to ProofOutputChannel by theory solvers.
+ *
+ * The static functions for constructing them check that the generator, if
+ * provided, is capable of proving the given conflict or lemma, or an assertion
+ * failure occurs. Otherwise an assertion error is given.
```
alex-ozdemir, line 71:
```suggestion
 * failure occurs. Otherwise an assertion error is given.
 *
 * While this is not enforced, a `TrustNode` generally encapsulates a **closed** proof
 * of the formula: one without free assumptions.
```

alex-ozdemir: If we don't `TrustNode`s for open proofs, I'd strengthen "generally" to "always".

ajreynol: Good point. And "generally" is fine for now I think.

PR #4589, opened 2020-06-08
 (proof-new) Add lazy proof utility

ajreynol:
Adds an extension of CDProof where facts can be mapped to (lazy) proof generators.


src/expr/lazy_proof.cpp
```diff
@@ -0,0 +1,176 @@
...
+  std::shared_ptr<ProofNode> opf = CDProof::mkProof(fact);
+  Assert(opf != nullptr);
+  if (!hasGenerators())
+  {
+    Trace("lazy-cdproof") << "...no generators, finished" << std::endl;
+    // optimization: no generators, we are done
+    return opf;
+  }
+  // otherwise, we traverse the proof opf and fill in the ASSUME leafs that
+  // have generators
```
HanielB, line 44: Please also say why you need to do a custom traversal and not use something like `getFreeAssumptions` and fill the generators of those guys.

src/expr/lazy_proof.h
```diff
@@ -0,0 +1,97 @@
...
+   */
+  std::shared_ptr<ProofNode> mkProof(Node fact) override;
+  /** Add step by generator
+   *
+   * This method stores that expected can be proven by proof generator pg if
+   * it is required to do so. This mapping is maintained in the remainder of
+   * the current context (according to the context c provided to this class).
+   *
+   * It is important to note that pg is asked to provide a proof for expected
+   * only when no other call for expected is provided via addStep. In
```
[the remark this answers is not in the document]
ajreynol: What is unclear?

src/expr/lazy_proof.cpp
ajreynol: There isn't anything custom here.  In theory this traversal could be refactored to use a common utility as getFreeAssumptions.  Although note that getFreeAssumptions will be later refactored as well (it should not live in ProofNode).  This is future work.

src/expr/lazy_proof.h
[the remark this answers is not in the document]
HanielB: "only when no other call for expected is provided via addStep". It's not clear what this means.

src/expr/lazy_proof.cpp
HanielB: I'd prefer having this clear by using `getFreeAssumptions` now. It'd just make this method clear now and later when getFreeAssumptions changes to some util it's easy to update this method.

ajreynol: `getFreeAssumptions` is not in master, not is it going to be merged to master. This will be refactored later to use a proof traversal utility, or be internal to ProofNodeManager.

ajreynol: Sorry, I meant to say "getFreeAssumptionsMap", which is what would be required here.

PR #4591, opened 2020-06-09
(proof-new) Remove arith-snorm option.

ajreynol:
This option only marginally helped and will be difficult to support with the new proof infrastructure.


ajreynol:
No worries, this was my code originally :)

PR #4593, opened 2020-06-09
(proof-new) Add quantifiers proof checker

ajreynol:
Adds core rules for quantifiers, some of which also will be used as a general way for introducing and skolemizing witness terms.

This PR also changes the order of other rules in proof_rule.h/cpp which was in an abnormal order from previous merges.


src/expr/proof_rule.h
```diff
@@ -378,101 +425,86 @@ enum class PfRule : uint32_t
...
+  // Conclusion: (exists ((x T)) F[x])
+  // where x is a BOUND_VARIABLE unique to the pair F,t.
+  EXISTS_INTRO,
+  // ======== Skolemize
+  // Children: (P:(exists ((x1 T1) ... (xn Tn)) F))
+  // Arguments: none
+  // ----------------------------------------
+  // Conclusion: (or (not (exists ((x1 T1) ... (xn Tn)) F)) F*sigma)
+  // sigma maps x1 ... xn to their representative skolems obtained by
+  // SkolemManager::mkSkolemExists.
```
HanielB, line 499: It'd be good to say how that creates correspondences with witnesses.

src/theory/quantifiers/proof_checker.cpp
```diff
@@ -0,0 +1,104 @@
...
+{
+  NodeManager* nm = NodeManager::currentNM();
+  // compute what was proven
+  if (id == PfRule::WITNESS_INTRO || id == PfRule::EXISTS_INTRO)
+  {
+    Assert(children.size() == 1);
+    Assert(args.size() == 1);
+    SkolemManager* sm = nm->getSkolemManager();
+    Node p = SkolemManager::getSkolemForm(children[0]);
+    Node t = SkolemManager::getSkolemForm(args[0]);
```
HanielB, line 46: Are these necessary or are you just being careful? I thought that these intro rules would be applied over skolem forms. Is it due to nested skolemizations?

src/theory/quantifiers/proof_checker.cpp
```diff
@@ -0,0 +1,104 @@
...
+    Assert(children.size() == 1);
+    Assert(args.empty());
+    if (children[0].getKind() != NOT || children[0][0].getKind() != FORALL)
+    {
+      return Node::null();
+    }
+    SkolemManager* sm = nm->getSkolemManager();
+    std::vector<Node> echildren;
+    echildren.insert(
+        echildren.begin(), children[0][0].begin(), children[0][0].end());
```
HanielB, line 68: nit: you could directly create the vector with the iterators.

src/theory/quantifiers/proof_checker.cpp
ajreynol: All external functions must call getSkolemForm.  For instance, this includes substitutions/rewrites/theory-specific proof rules.  Witness form is only used internally in the proof checker.

ajreynol: This policy has been updated on proof-new recently.  The proof checker now operates on skolem form. I've updated the PR to simplify these calls.

PR #4595, opened 2020-06-10
Move sygus datatype utility functions to their own file

ajreynol:
Separates them from the core datatype utilities.

Code move only.


src/theory/datatypes/sygus_datatype_utils.cpp
```diff
@@ -0,0 +1,758 @@
...
+  }
+  else if (tn.isTester())
+  {
+    return APPLY_TESTER;
+  }
+  else if (tn.isFunction())
+  {
+    return APPLY_UF;
+  }
+  return UNDEFINED_KIND;
```
HanielB, line 121: could use `NodeManager::getKindForFunction` rather than doing this yourself

PR #4603, opened 2020-06-12
(proof-new) Term conversion proof generator utility

ajreynol:
This utility is used for constructing any proof where a term is "converted" into another by small step rewrites.  This utility will be used to construct the skeleton of the proofs of rewrites, preprocessing steps, expand definitions, results of substitutions, and so on.


src/expr/term_conversion_proof_generator.cpp
```diff
@@ -0,0 +1,246 @@
...
+      visited[cur] = Node::null();
+      // did we rewrite the current node (possibly at pre-rewrite)?
+      Node rcur = getRewriteStep(cur);
+      if (!rcur.isNull())
+      {
+        // d_proof should have a proof of cur = rcur. Hence there is nothing
+        // to do here, as pf will reference prg to get the proof from d_proof.
+        rewritten[cur] = rcur;
+        visit.push_back(cur);
+        visit.push_back(rcur);
```
HanielB, line 128: Say why you're not directly assigning `cur` to `rcur` in `visited`

src/expr/term_conversion_proof_generator.cpp
```diff
@@ -0,0 +1,246 @@
...
+  std::unordered_map<TNode, Node, TNodeHashFunction>::iterator itr;
+  std::vector<TNode> visit;
+  TNode cur;
+  visit.push_back(t);
+  do
+  {
+    cur = visit.back();
+    visit.pop_back();
+    it = visited.find(cur);
+
```
HanielB, line 116: It'd be useful to say explicitly in the traversal code below the points that correspond to pre/post rewrites

src/expr/term_conversion_proof_generator.cpp
```diff
@@ -0,0 +1,246 @@
...
+        {
+          if (cur != ret)
+          {
+            visit.push_back(cur);
+          }
+          // d_proof should have a proof of ret = rret, hence nothing to do
+          // here, for the same reasons as above.
+          rewritten[ret] = rret;
+          visit.push_back(ret);
+          visit.push_back(rret);
```
HanielB, line 215: ditto

src/expr/term_conversion_proof_generator.h
```diff
@@ -0,0 +1,119 @@
...
+class TConvProofGenerator : public ProofGenerator
+{
+ public:
+  /** Constructor
+   *
+   * @param pnm The proof node manager for constructing ProofNode objects.
+   * @param c The context that this class depends on. If none is provided,
+   * this class is context-independent.
+   */
+  TConvProofGenerator(ProofNodeManager* pnm, context::Context* c = nullptr);
```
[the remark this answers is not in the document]
ajreynol: Context-free refers to the term context, not the temporal context.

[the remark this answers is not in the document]
HanielB: I see.

PR #4605, opened 2020-06-12
Add support for minimal unsat cores

4tXJ7f:
This commit adds support for computing minimal unsat cores. The
algorithm implemented in this commit is just a trivial deletion-based
algorithm that tries to remove each assertion in the unsat core
individually.


src/smt/smt_engine.cpp
```diff
@@ -2510,14 +2511,89 @@ UnsatCore SmtEngine::getUnsatCoreInternal()
...
 }
 
+UnsatCore SmtEngine::reduceUnsatCore(const UnsatCore& core)
+{
+  Assert(options::unsatCores())
+      << "cannot reduce unsat core if unsat cores are turned off";
+
+  Notice() << "SmtEngine::reduceUnsatCore(): reducing unsat core" << endl;
+  std::unordered_set<Expr, ExprHashFunction> removed;
+  for (UnsatCore::iterator it = core.begin(); it != core.end(); ++it)
```
HanielB, line 2535: Couldn't you be using a range-based `for` here?

src/options/smt_options.toml
```diff
@@ -220,6 +220,14 @@ header = "options/smt_options.h"
   notifies   = ["notifyBeforeSearch"]
   help       = "turn on unsat core generation"
 
+[[option]]
+  name       = "minimalUnsatCores"
```
HanielB, line 224: The unsat cores you are generating here are not guaranteed to be minimal, right? In that case I'd prefer `minimizeUnsatCores` rather than `minimalUnsatCores`.

ahmed-irfan: if it is not minimal, then I would prefer `smallUnsatCore` or `reducedUnsatCore`

4tXJ7f: They are guaranteed to be _minimal_ unsat cores (i.e. removing any of the assertions makes the constraints satisfiable) but not _minimum_ unsat cores (following the terminology e.g. [here](https://en.wikipedia.org/wiki/Unsatisfiable_core) and [here](https://arxiv.org/ftp/arxiv/papers/1401/1401.3878.pdf)). The only caveat is that we may not end up with minimal unsat cores when we get `unknown` answers. What do you think?

HanielB: I think that the Wikipedia article you linked is nice. :) But the terminology is confusing regardless. The key difference is between local vs global, so it seems a bit arbitrary to use "minimal" for the local case but "minimum" for the global one.  I'd go with `smallUnsatCore` just to be non-committal.

aniemetz: It is common to use `minimal` if it can't be further reduced but is not globally the minimum, but only locally. I personally don't like `smallUnsatCore` -- it doesn't imply what it actually is, minimal in the sense that if you remove any assertion it's going to be sat.

HanielB:
Once again I thought this was already in master. :) I guess the outstanding thing is just solving the conflicts, other than picking one of the naming options discussed above?

4tXJ7f:
I've updated the PR and applied @HanielB's suggestions about range-based for loops.

HanielB:
Should we just merge then?

PR #4607, opened 2020-06-12
 Inferences and model construction taking into account seq.unit

ajreynol:
Towards theory of sequences.

This makes the strings solver handle `seq.unit`, which requires two new inferences and updates to its model construction.

It also fixes a bug in the best content heuristic, which previously failed to update the best score.


src/theory/strings/base_solver.cpp
```diff
@@ -67,19 +68,75 @@ void BaseSolver::checkInit()
...
+          if (prev.isNull() || n.isConst())
+          {
+            d_eqcInfo[eqc].d_bestContent = n;
+            d_eqcInfo[eqc].d_bestScore = 0;
+            d_eqcInfo[eqc].d_base = n;
+            d_eqcInfo[eqc].d_exp = Node::null();
+          }
         }
-        else if (tn.isInteger())
+        if (tn.isInteger())
```
4tXJ7f, line 133: I am not quite sure, I understand why this changed to an `if` from an `else if`. What it the remaining code could still do something meaningful?

ajreynol: This is so that seq.unit terms are indexed (and hence we mark congruence for them). Previously it was assumed that constant (or now, constant-like) terms had 0 children.

PR #4612, opened 2020-06-15
Allow for better interaction of Integer/Rational with mpz_class/mpq_class.

nafur:
Right now, both `Integer` and `Rational` do not allow to be constructed from their internal representation (like `mpz_class`, `mpq_class` or their cln equivalents).
With the integration of libpoly, this is an unnecessary inefficiency, in particular as these constructors already exist and are used (privately). This PR makes them public.
Furthermore, this PR returns from `getValue()` by `const &` instead of by value. This also is just an inefficiency without any notable benefit. (If one really want to change the underlying member, we can also access it via `reinterpret_cast<mpz_class*>(&rational)` instead of `const_cast<mpz_class*>(&rational.getValue())`...)


nafur:
The inclusion order for cln is fixed now, for gmp it seemed fine.

aniemetz:
I meant the order of declarations in the class, there's multiple `public`, `private` and `protected` blocks, and in general it is not in the order imposed by our guidelines.

nafur:
I don't see any protected sections, but I moved the private ones to the bottom...

src/util/integer_cln_imp.h
```diff
@@ -510,53 +471,101 @@ class CVC4_PUBLIC Integer {
...
+    return (a >= b) ? a : b;
   }
 
-  friend class CVC4::Rational;
-};/* class Integer */
+ private:
+  /**
+   * Stores the value of the rational is stored in a C++ CLN integer class.
+   */
+  cln::cl_I d_value;
```
aniemetz, line 527: Data members come after member functions.

src/util/integer_gmp_imp.h
```diff
@@ -468,76 +463,97 @@ class CVC4_PUBLIC Integer {
...
   }
 
-  friend class CVC4::Rational;
-};/* class Integer */
+ private:
+  /**
+   * Stores the value of the rational is stored in a C++ GMP integer class.
+   * Using this instead of mpz_t allows for easier destruction.
+   */
+  mpz_class d_value;
```
aniemetz, line 537: Ditto

PR #4615, opened 2020-06-15
Support AND/OR definitions in lambda to array rewriting

HanielB:
This makes the conversion between lambdas and arrays, done in the theory builtin rewriter and used in higher-order model construction, robust to function definitions in terms of AND/OR.

This also improves tracing and makes a few parts of the code adhere to code guidelines.

### DIFF
--- a/src/theory/builtin/theory_builtin_rewriter.cpp
+++ b/src/theory/builtin/theory_builtin_rewriter.cpp
@@ -262,8 +262,12 @@ Node TheoryBuiltinRewriter::getArrayRepresentationForLambdaRec(TNode n,
       bool pol = curr[0].getKind() != kind::NOT;
       bool inverted = (pol && ck == kind::AND) || (!pol && ck == kind::OR);
       index_eq = pol ? curr[0] : curr[0][0];
+      // processed : the value that is determined by the first child of curr
+      // remainder : the remaining children of curr
       Node processed, remainder;
-      processed = nm->mkConst(pol && !inverted);
+      // the value is the polarity of the first child or its inverse if we are
+      // in the inverted case
+      processed = nm->mkConst(!inverted? pol : !pol);
       // build an OR/AND with the remaining components
       if (curr.getNumChildren() == 2)
       {
--- a/src/theory/builtin/theory_builtin_rewriter.cpp
+++ b/src/theory/builtin/theory_builtin_rewriter.cpp
@@ -414,8 +414,8 @@ Node TheoryBuiltinRewriter::getArrayRepresentationForLambdaRec(TNode n,
     }
     Trace("builtin-rewrite-debug2") << "  make array store all " << curr.getType() << " annotated : " << array_type << std::endl;
     Assert(curr.getType().isSubtypeOf(array_type.getArrayConstituentType()));
-    curr = nm->mkConst(ArrayStoreAll(
-        (static_cast<ArrayType>(array_type.toType())), curr.toExpr()));
+    curr = nm->mkConst(
+        ArrayStoreAll((ArrayType(array_type.toType())), curr.toExpr()));
     Trace("builtin-rewrite-debug2") << "  build array..." << std::endl;
     // can only build if default value is constant (since array store all must be constant)
     Trace("builtin-rewrite-debug2") << "  got constant base " << curr << std::endl;
--- a/src/theory/builtin/theory_builtin_rewriter.cpp
+++ b/src/theory/builtin/theory_builtin_rewriter.cpp
@@ -201,9 +201,12 @@ Node TheoryBuiltinRewriter::getArrayRepresentationForLambdaRec(TNode n,
 
   Node first_arg = n[0][0];
   Node rec_bvl;
-  if( n[0].getNumChildren()>1 ){
+  unsigned size = n[0].getNumChildren();
+  if (size > 1)
+  {
     std::vector< Node > args;
-    for( unsigned i=1; i<n[0].getNumChildren(); i++ ){
+    for (unsigned i = 1; i < size; i++)
+    {
       args.push_back( n[0][i] );
     }
     rec_bvl = nm->mkNode(kind::BOUND_VAR_LIST, args);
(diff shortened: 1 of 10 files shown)


src/theory/builtin/theory_builtin_rewriter.cpp
```diff
@@ -234,13 +237,67 @@ Node TheoryBuiltinRewriter::getArrayRepresentationForLambdaRec(TNode n,
...
+          << "  process base : " << curr << std::endl;
+      // curr = Rewriter::rewrite(curr);
+      // Trace("builtin-rewrite-debug2")
+      //     << "  rewriten base : " << curr << std::endl;
+      // Complex Boolean return cases, in which
+      //  (1) lambda x. (= x v1) v ... becomes
+      //      lambda x. (ite (= x v1) true [...])
+      //
+      //  (2) lambda x. (not (= x v1)) ^ ... becomes
+      //      lambda x. (ite (= x v1) true [...])
```
ajreynol, line 252: You mean this branch is false?

src/theory/builtin/theory_builtin_rewriter.cpp
```diff
@@ -234,13 +237,67 @@ Node TheoryBuiltinRewriter::getArrayRepresentationForLambdaRec(TNode n,
...
     else
     {
       Trace("builtin-rewrite-debug2")
           << "  process base : " << curr << std::endl;
-      // Boolean return case, e.g. lambda x. (= x v) becomes
-      // lambda x. (ite (= x v) true false)
-      bool pol = curr.getKind() != kind::NOT;
+      // Simple Boolean return cases, in which
+      //  (1) lambda x. (= x v) becomes lambda x. (ite (= x v) true false)
+      //  (2) lambda x. v becomes lambda x. (ite (= v true) true false)
```
ajreynol, line 298: Probably (= v true) should be (= x v) ?

src/theory/builtin/theory_builtin_rewriter.cpp
HanielB: Yes, will fix.

src/theory/builtin/theory_builtin_rewriter.cpp
HanielB: Yes, will fix.

src/theory/builtin/theory_builtin_rewriter.cpp
```diff
@@ -234,13 +237,67 @@ Node TheoryBuiltinRewriter::getArrayRepresentationForLambdaRec(TNode n,
...
+      // Trace("builtin-rewrite-debug2")
+      //     << "  rewriten base : " << curr << std::endl;
+      // Complex Boolean return cases, in which
+      //  (1) lambda x. (= x v1) v ... becomes
+      //      lambda x. (ite (= x v1) true [...])
+      //
+      //  (2) lambda x. (not (= x v1)) ^ ... becomes
+      //      lambda x. (ite (= x v1) true [...])
+      //
+      // Note the negateg cases of the lhs of the OR/AND operators above are
```
ajreynol, line 254: negateg -> negated

src/theory/builtin/theory_builtin_rewriter.cpp
```diff
@@ -330,28 +393,39 @@ Node TheoryBuiltinRewriter::getArrayRepresentationForLambdaRec(TNode n,
...
+    for (unsigned i = 0; i < size; i++)
+    {
+      unsigned index = (size - 1) - i;
+      array_type = nm->mkArrayType(n[0][index].getType(), array_type);
     }
     Trace("builtin-rewrite-debug2") << "  make array store all " << curr.getType() << " annotated : " << array_type << std::endl;
     Assert(curr.getType().isSubtypeOf(array_type.getArrayConstituentType()));
-    curr = NodeManager::currentNM()->mkConst(ArrayStoreAll(((ArrayType)array_type.toType()), curr.toExpr()));
+    curr = nm->mkConst(ArrayStoreAll(
+        (static_cast<ArrayType>(array_type.toType())), curr.toExpr()));
```
ajreynol, line 418: This should actually be `ArrayType(array_type.toType())` (it's a construction of an ArrayType, not a cast).

src/theory/builtin/theory_builtin_rewriter.cpp
```diff
@@ -234,13 +237,67 @@ Node TheoryBuiltinRewriter::getArrayRepresentationForLambdaRec(TNode n,
...
+      //  (1) lambda x. (= x v1) v ... becomes
+      //      lambda x. (ite (= x v1) true [...])
+      //
+      //  (2) lambda x. (not (= x v1)) ^ ... becomes
+      //      lambda x. (ite (= x v1) true [...])
+      //
+      // Note the negateg cases of the lhs of the OR/AND operators above are
+      // handled by pushing the recursion to the then-branch, with the
+      // else-branch being the constant value.
+      bool pol = curr[0].getKind() != kind::NOT;
```
ajreynol, line 257: I don't quite follow the inverted part, can you give an example?

src/theory/builtin/theory_builtin_rewriter.cpp
```diff
@@ -234,13 +237,75 @@ Node TheoryBuiltinRewriter::getArrayRepresentationForLambdaRec(TNode n,
...
+      // would be
+      //  (1') lambda x. (not (= x v1)) v ... becomes
+      //       lambda x. (ite (= x v1) [...] true)
+      // thus requiring the rest of the disjunction to be further processed in
+      // the then-branch as the current value.
+      bool pol = curr[0].getKind() != kind::NOT;
+      bool inverted = (pol && ck == kind::AND) || (!pol && ck == kind::OR);
+      index_eq = pol ? curr[0] : curr[0][0];
+      Node processed, remainder;
+      processed = nm->mkConst(pol && !inverted);
```
ajreynol, line 266: Still confused at this value, since it prefers a value of "false".  Are you sure this is correct?  I would have perhaps expected `pol != inverted` ?

HanielB: Yes, this is not correct. If we are in the inverted case then the polarity is flipped, otherwise it is as computed. I'll update and add more doc, good catch.

PR #4617, opened 2020-06-15
(proof-new) Updates to proof node manager

ajreynol:
Main feature added is the mkScope interface, which is agnostic to symmetry of (dis)equalities.

It also adds a check for cyclic proofs when using the interface `ProofNodeManager::updateNode`.

Note that an earlier version of this method was agnostics to predicates vs Boolean equality with constants.  This is no longer required.


src/expr/proof_node_manager.cpp
```diff
@@ -39,27 +42,164 @@ std::shared_ptr<ProofNode> ProofNodeManager::mkNode(
...
 
+std::shared_ptr<ProofNode> ProofNodeManager::mkNode(
+    PfRule id,
+    std::shared_ptr<ProofNode> child1,
+    const std::vector<Node>& args,
+    Node expected)
+{
+  std::vector<std::shared_ptr<ProofNode>> children;
+  children.push_back(child1);
+  return mkNode(id, children, args, expected);
```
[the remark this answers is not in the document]
ajreynol: On second thought, I'm going to delete this method, it's not necessary anymore.

src/expr/proof_node_manager.h
```diff
@@ -73,6 +73,50 @@ class ProofNodeManager
...
+   *   pf = TRANS( ASSUME( x=y ), ASSUME( y=z ) )
+   *   assumps = { y=x, y=z }
+   * This method will modify pf to be:
+   *   pf = TRANS( SYMM( ASSUME( y=x ) ), ASSUME( y=z ) )
+   * so that y=x matches the free assumption. The returned proof is:
+   *   SCOPE(TRANS( SYMM( ASSUME( y=x ) ), ASSUME( y=z ) ) :args { y=x, y=z })
+   *
+   * Additionally, if both ensureClosed and doMinimize are true, assumps is
+   * updated to contain exactly the free asumptions of pf. The minimized
+   * vector is passed as arguments to the SCOPE.
```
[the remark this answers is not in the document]
ajreynol: This comment is more relevant where we are using the proof equality engine, which isn't specific to this method.

src/expr/proof_node_manager.cpp
```diff
@@ -154,7 +135,7 @@ std::shared_ptr<ProofNode> ProofNodeManager::mkScope(
   }
   else if (ac.size() < assumps.size())
   {
-    // always must remove duplicates
+    // remove duplicates to avoid redundant literals in clauses
```
HanielB, line 138: I imagine this is an optimization then? Was it not done before?

src/expr/proof_node_manager.h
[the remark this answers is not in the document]
HanielB: Indeed. Could you say though that this method is also doing removal of duplicates?

src/expr/proof_node_manager.cpp
ajreynol: It is an optimization that has always been done, it used to be required, now the code is robust to  if it weren't done (I think).

HanielB: The SAT solver silently eliminates such duplicates. What is puzzling me is just that this is being done here rather than in the user of this class.

ajreynol: ah, so I think originally the ProofEqEngine was doing this itself. This code is actually based off a number of theories explain methods which were doing it themselves. I think it's a good idea to always do this here.

HanielB: If we are doing proofs, indeed. But if we are not? Since this is about the explanation assumptions itself I'd expect this to be made independent of proofs.

ajreynol: The reason this is done here when proofs are enabled is because it is convenient to do so here, since we've computed `ac`.  Thus, we shouldn't require the user to clean this up themselves, since then we'd actually be converting to/from vector/unordered_set twice.

If proof are disabled, then it is up to the user eliminate duplicates themselves.  That is a separate issue.

Regardless, it's for performance reasons that we don't insist that duplicates are eliminated before calling mkScope.

HanielB: OK. In that case then I'd like to have this explicitly controlled via a parameter, like `doMinimize`, which would be `true` by default. What do you think?

ajreynol: Is there a use case for this?  If so, I can add this. However, I'd prefer not adding this.

HanielB: Testing efficiency, which I assume is the same use case of `doMinimize`?

PR #4624, opened 2020-06-16
(proof-new) Add REWRITE trust node kind.

ajreynol:
This will be used for a number of purposes, including tracking proofs for rewriting and preprocessing.


src/theory/trust_node.cpp
```diff
@@ -75,7 +83,12 @@ TrustNodeKind TrustNode::getKind() const { return d_tnk; }
 
 Node TrustNode::getNode() const
 {
-  return d_tnk == TrustNodeKind::LEMMA ? d_proven : d_proven[0];
+  switch (d_tnk)
+  {
+    case TrustNodeKind::LEMMA: return d_proven;
+    case TrustNodeKind::REWRITE: return d_proven[1];
+    default: return d_proven[0];
```
HanielB, line 90: Say explicitly what are the cases here.

PR #4625, opened 2020-06-16
Add missing REQUIRES to new regressions.

aniemetz:


test/regress/regress0/bv/issue-4076.smt2
```diff
@@ -1,3 +1,4 @@
+; REQUIRES: no-competition
```
ajreynol, line 1: I dont think this REQUIRES is necessary.

PR #4629, opened 2020-06-17
Improve memory management in Java bindings

### LINKED ISSUE #2846: JVM crash in finalizer

4tXJ7f:
Fixes #2846. One of the challenges of the Java bindings is that the
garbage collector can delete unused objects at any time in any order.
This is an issue with CVC4's API because we require all `Expr`s to be
deleted before the corresponding `ExprManager`. In the past, we were
using `NewGlobalRef`/`DeleteGlobalRef` on the wrapper object of
`ExprManager`. The problem is that we can have multiple instances of the
wrapper that internally all refer to the same `ExprManager`. This commit
implements a different approach where the Java wrappers hold an explicit
reference to the `ExprManager`. The commit also removes some unused or
unimportant API bits from the bindings.


src/bindings/java/cvc4_std_vector.i
```diff
@@ -0,0 +1,206 @@
+/**
+ * The following file is based on
+ * https://github.com/swig/swig/blob/master/Lib/java/std_vector.i
+ *
+ * Note: The SWIG library is under a different license than SWIG itself. See
+ * https://github.com/swig/swig/blob/master/LICENSE for details.
+ *
+ * This file define the macro SWIG_STD_VECTOR_EM to wrap a C++ std::vector for
```
mpreiner, line 8:
```suggestion
 * This file defines the macro SWIG_STD_VECTOR_EM to wrap a C++ std::vector for
```

PR #4630, opened 2020-06-17
Do not traverse WITNESS for partial substitutions in extended rewriter

### LINKED ISSUE #4620: (--ext-rewrite-quant) Refutational soundness bug on NIA formula with exists and div operators
Hi,
For this formula, CVC4 with --ext-rewrite-quant gives an incorrect answer:
```
[516] % z3release small.smt2 
sat
[517] % cvc4 small.smt2 
sat
[518] % cvc4 --ext-rewrite-quant small.smt2 
unsat
[519] % cat small.smt2 
(set-logic NIA)
(assert (exists ((x Int)) (= (div 1 x x) x)))
(check-sat)
[520] %
```

OS: Ubuntu 18.04
Commit: e5f880a

ajreynol:
Fixes #4620.

The extended rewrite (and A B) ---> (and A (B * { A -> true } ) triggers an unsoundness when B contains witness terms.  More generally, contextual substitution *into* witness terms is unsound since the skolemization of witness terms is added globally whereas the substitution corresponds to a fact that holds locally.  This means that `A -> true` above may be added as a global constraint indirectly through witness skolemization.

A general example of this unsoundness:
```
(or (and (>= x 0) (P (witness ((z Int)) (=> (>= x 0) (= (* z z) x))))) F)
rewrites to
(or (and (>= x 0) (P (witness ((z Int)) (= (* z z) x)))) F)
preprocesses to
(and (or (and (>= x 0) (P k)) F) (= (* k k) x))
which now implies that (>= x 0) by the last conjunct, whereas this was not implied in the original formula
```

This PR limits the kinds that can be traversed when applying substitutions in the extended rewriter, including from the rewrite above. In particular, the fix ensures that the partialSubstitute method is used in the extended rewriter when applicable, which now explicitly disallows recursion on WITNESS.

Notice that this fixes contextual substitutions in the extended rewriter, but does not fix the more general issue. In particular, we still should be careful to check if contextual substitutions are applied by any other preprocessing passes.


src/theory/quantifiers/extended_rewrite.cpp
```diff
@@ -488,6 +488,8 @@ Node ExtendedRewriter::extendedRewriteIte(Kind itek, Node n, bool full)
     {
       // reverse substitution to opposite child
       // r{ x -> t } = s  implies  ite( x=t ^ C, s, r ) ---> r
+      // We can use ordinary substitute since the result of the substitution
+      // is not being returned.
```
HanielB, line 492: Can you say more on why this is not an issue? Why not returning `nn` is OK? It's not used in what is ultimately returned?

test/regress/regress0/fp/ext-rew-test.smt2
```diff
@@ -1,3 +1,4 @@
+; REQUIRES: symfpu
```
HanielB, line 1: This regression now requires symfpu to be solved?

src/theory/quantifiers/extended_rewrite.h
```diff
@@ -200,11 +200,17 @@ class ExtendedRewriter
   /** Partial substitute
    *
    * Applies the substitution specified by assign to n, recursing only beneath
-   * terms whose Kind appears in rec_kinds.
+   * terms whose Kind appears in rkinds (when rkinds is empty), and additional
```
HanielB, line 203: additionally

src/theory/quantifiers/extended_rewrite.h
```diff
@@ -200,11 +200,17 @@ class ExtendedRewriter
...
+   * always disallow recursing beneath WITNESS.
```
HanielB, line 204: disallows

src/theory/quantifiers/extended_rewrite.cpp
```diff
@@ -1607,6 +1598,20 @@ Node ExtendedRewriter::partialSubstitute(Node n,
...
 }
 
+Node ExtendedRewriter::partialSubstitute(Node n,
+                        std::vector<Node>& vars,
+                        std::vector<Node>& subs,
+                        std::map<Kind, bool>& rkinds)
+{
+  Assert (vars.size()==subs.size());
+  std::map<Node, Node> assign;
+  for (size_t i=0, nvars=vars.size(); i<nvars; i++)
```
HanielB, line 1608: format

test/regress/regress0/fp/ext-rew-test.smt2
ajreynol: It did in an intermediate commit.  Regardless, this should be enabled for this benchmark since its FP.

src/theory/quantifiers/extended_rewrite.h
```diff
@@ -200,11 +200,17 @@ class ExtendedRewriter
...
-   * terms whose Kind appears in rec_kinds.
+   * terms whose Kind appears in rkinds (when rkinds is empty), and additionally
+   * never recursing beneath WITNESS.
    */
   Node partialSubstitute(Node n,
                          std::map<Node, Node>& assign,
                          std::map<Kind, bool>& rkinds);
+  /** same as above, with vectors */
+  Node partialSubstitute(Node n,
+                         std::vector<Node>& vars,
```
nafur, line 211: As both `vars` and `subs` are not modified, I'd pass them as `const &`. The same is true for `assign` in the previous overload of `partialSubstitute`.

ajreynol: Agreed, I'll update.

PR #4632, opened 2020-06-18
Add Match utility function.

abdoo8080:


src/expr/node_algorithm.cpp
```diff
@@ -634,5 +634,87 @@ void getComponentTypes(
...
+      }
+      else
+      {
+        // the two subterms are not equal
+        return false;
+      }
+    }
+    else
+    {
+      // if the two term are not equal, at least make sure that they have the
```
HanielB, line 702: the two term -> the two terms
at least make sure -> make sure

src/expr/node_algorithm.cpp
```diff
@@ -634,5 +634,87 @@ void getComponentTypes(
...
+      // same number of children and that their operators are equal
+      if (curr.first.getNumChildren() != curr.second.getNumChildren()
+          || curr.first.getOperator() != curr.second.getOperator())
```
HanielB, line 705: motivate the need to compare the number of children on top of comparing tho operators. Also motivate why it does not suffice to compare just the kinds.

src/expr/node_algorithm.cpp
```diff
@@ -634,5 +634,87 @@ void getComponentTypes(
...
+      {
+        return false;
+      }
+      // recurse on children
+      for (size_t i = 0, n = curr.first.getNumChildren(); i < n; ++i)
+      {
+        stack.push_back({curr.first[i], curr.second[i]});
```
HanielB, line 712: why push_back here but emplace_back above?

src/expr/node_algorithm.h
```diff
@@ -185,6 +185,29 @@ Node substituteCaptureAvoiding(TNode n,
...
+ * then a call to unify should return `true` with subs = {(F, b)}
+ *
+ * @param x the term (containing free vars) to compare an instance term
+ * against
+ * @param y the instance term in question
+ * @param subs the mapping from free vars in `x` to terms in `y`
+ * @return whether or not `y` is an instance of `x`
+ */
+bool unify(Node x,
+           Node y,
```
HanielB, line 208: Since it's common to associate "x" and "y" to variables and the arguments here are arbitrary nodes I'd prefer to have these named n1 and n2. Doing this you could then change the F above standing for a free variable to x.

src/expr/node_algorithm.cpp
abdoo8080: Done!

src/expr/node_algorithm.cpp
abdoo8080: Nice catch!

src/expr/node_algorithm.h
abdoo8080: Done!

src/expr/node_algorithm.cpp
abdoo8080: Done!

HanielB:
Thanks for the changes, @abdoo8080.

@martin-cs: This code is for syntactic unification. The quantifiers handling in an SMT solver is generally doing it modulo the E-graph, so it makes sense for not necessarily having such a utility there. That said, this might indeed be redundant. Could you give a motivation for the PR, @abdoo8080?

Another thing: what this method does is actually matching, not unification (i.e. you're not looking to instantiate variables in `n2` to variables in `n1`). Could you update accordingly?

ajreynol:
I've been discussing with @abdoo8080 about adding this method.

@martin-cs : indeed as Haniel mentions, (perhaps surprisingly) my code in quantifiers typically has not needed the basic version of this method.  Or if it has, I have forgotten.  Regardless, the plan is for this utility method to be the standard and then potentially refactor old hidden occurrences of it.

The motivation for the PR is that coincidentally, two algorithms that are currently in development require it (proof reconstruction for rewrites and solution reconstruction for SyGuS).

ajreynol:
Also, I agree with @HanielB , the name of this method I guess should be named "match" not "unify"?

test/unit/expr/node_algorithm_black.h
```diff
@@ -130,4 +130,68 @@ class NodeAlgorithmBlack : public CxxTest::TestSuite
     TS_ASSERT(result[*d_boolTypeNode].find(d_nodeManager->operatorOf(EQUAL))
               != result[*d_boolTypeNode].end());
   }
+
+  void testUnify()
```
HanielB, line 134: Update to `match` as well

abdoo8080: Done!

PR #4638, opened 2020-06-20
Interpolation step 1

Ying1123:
This is the first step of adding Interpolation. The whole change will be adding the API for (get-interpol s B), which is aim for computes an I that A->I and I->B. Here A is the assertions in the stack.

The first step creates the API framework, while omits the implementation for getting interpolation.


src/smt/set_defaults.cpp
```diff
@@ -1024,9 +1027,10 @@ void setDefaults(SmtEngine& smte, LogicInfo& logic)
     // template inference for invariant synthesis, and single invocation
     // techniques.
     bool reqBasicSygus = false;
-    if (options::produceAbducts())
+    if (options::produceAbducts()
+        || options::produceInterpols() != options::ProduceInterpols::NONE)
     {
-      // if doing abduction, we should filter strong solutions
+      // if doing abduction or interpolation, we should filter strong solutions
```
yoni206, line 1033: Not so sure about that. @ajreynol and @HanielB , is this true for inteprolation?

src/smt/set_defaults.cpp
```diff
@@ -1363,11 +1367,14 @@ void setDefaults(SmtEngine& smte, LogicInfo& logic)
           "--sygus-expr-miner-check-timeout=N requires "
           "--sygus-expr-miner-check-use-export");
     }
-    if (options::sygusRewSynthInput() || options::produceAbducts())
+    if (options::sygusRewSynthInput() || options::produceAbducts()
```
yoni206, line 1370: ditto.

src/smt/smt_engine.h
```diff
@@ -618,6 +620,23 @@ class CVC4_PUBLIC SmtEngine
    */
   Expr doQuantifierElimination(const Expr& e, bool doFull, bool strict = true);
 
+  /**
+   * This method asks this SMT engine to find an interpolation with respect to
```
yoni206, line 624:
```suggestion
   * This method asks this SMT engine to find an interpolat with respect to
```

src/smt/smt_engine.h
```diff
@@ -618,6 +620,23 @@ class CVC4_PUBLIC SmtEngine
...
+   * the current assertion stack (call it A) and the conjecture (call it B). If
+   * this method returns true, then interpol is set to a formula I such that A ^
```
yoni206, line 626:
```suggestion
   * this method returns true, then interpolant is set to a formula I such that A ^
```

src/smt/smt_engine.h
ahmed-irfan: interpolant

src/smt/smt_engine.h
```diff
@@ -618,6 +620,23 @@ class CVC4_PUBLIC SmtEngine
...
+   * ~I and I ^ ~B are both unsatisfiable.
+   *
+   * The argument grammarType is a sygus datatype type that encodes the syntax
+   * restrictions on the shape of possible solutions.
```
HanielB, line 630: shape -> shapes

src/smt/smt_engine.h
```diff
@@ -932,6 +951,18 @@ class CVC4_PUBLIC SmtEngine
    * unsatisfiable. If not, then the found solutions are wrong.
    */
   void checkSynthSolution();
+
+  /**
+   * Check that a solution to an interpolation conjecture is indeed a solution.
```
HanielB, line 956: Use "interpolation problem" rather than "interpolation conjecture" since you are using "conjecture" to refer to the `B` in the `get-interpolant` call.

src/smt/command.h
```diff
@@ -1060,6 +1060,53 @@ class CVC4_PUBLIC GetSynthSolutionCommand : public Command
   SmtEngine* d_smtEngine;
 }; /* class GetSynthSolutionCommand */
 
+/** The command (get-interpol s B)
+ *
+ * This command asks for an interpolation from the current set of assertions and
```
HanielB, line 1065: interpolation -> interpolant

src/smt/command.h
```diff
@@ -1060,6 +1060,53 @@ class CVC4_PUBLIC GetSynthSolutionCommand : public Command
...
+ * conjecture (goal) given by the argument B.
```
HanielB, line 1066:
```suggestion
 * conjecture (goal) B.
```

src/smt/command.h
```diff
@@ -1060,6 +1060,53 @@ class CVC4_PUBLIC GetSynthSolutionCommand : public Command
...
+ * The symbol s is the name for the interpolation predicate. If we successfully
+ * find a predicate P, then the output response of this command is: (define-fun
+ * s () Bool P)
+ */
+class CVC4_PUBLIC GetInterpolCommand : public Command
+{
+ public:
+  GetInterpolCommand();
+  GetInterpolCommand(const std::string& name, Expr conj);
+  GetInterpolCommand(const std::string& name, Expr conj, const Type& gtype);
```
HanielB, line 1077: document this one to say what `gtype` is

src/smt/command.h
```diff
@@ -1060,6 +1060,53 @@ class CVC4_PUBLIC GetSynthSolutionCommand : public Command
...
+
+  /** Get the conjecture of the interpolation query */
+  Expr getConjecture() const;
+  /** Get the grammar type given for the interpolation query */
```
HanielB, line 1081:
```suggestion
  /** Get the grammar sygus datatype type given for the interpolation query */
```

src/options/smt_options.toml
```diff
@@ -651,6 +651,34 @@ header = "options/smt_options.h"
...
+  help = "don't compute interpolants"
+[[option.mode.DEFAULT]]
+  name = "default"
+  help = "use the default grammar for the theory or the user-defined grammar if given"
+[[option.mode.ASSUMPTIONS]]
+  name = "assumptions"
+  help = "use only operators that occur in the assumptions"
+[[option.mode.CONCLUSION]]
+  name = "conclusion"
+  help = "use only operators that occur in the conclusion"
```
HanielB, line 674: wouldn't it be better to use "conjecture" rather than "conclusion", as in the rest of the PR?

src/api/cvc4cpp.h
```diff
@@ -3010,6 +3010,15 @@ class CVC4_PUBLIC Solver
    */
   void pop(uint32_t nscopes = 1) const;
 
+
+  /**
+   * Get an interpolant
+   * SMT-LIB: ( get-interpol B )
+   * Requires to enable option 'produce-interpols'.
+   * @return a Term I such that A->I and I->B are valid, where A is the current set of assertions and B is given in the input.
```
HanielB, line 3018: clang-format

src/smt/set_defaults.cpp
HanielB: I don't know. I defer to the interpolation users. :)

src/smt/set_defaults.cpp
Ying1123: what do you mean?

src/api/cvc4cpp.cpp
```diff
@@ -4642,6 +4642,15 @@ void Solver::pop(uint32_t nscopes) const
   CVC4_API_SOLVER_TRY_CATCH_END;
 }
 
+bool Solver::getInterpolant(Term B, Term& output) const {
```
aniemetz, line 4645: This should be accompanied by a unit test in `test/unit/api/solver_black.h`.

src/api/cvc4cpp.cpp
```diff
@@ -4642,6 +4642,15 @@ void Solver::pop(uint32_t nscopes) const
...
+  Expr result;
```
aniemetz, line 4646:
```suggestion
  CVC4_API_SOLVER_TRY_CATCH_BEGIN;
  Expr result;
```

src/api/cvc4cpp.cpp
```diff
@@ -4642,6 +4642,15 @@ void Solver::pop(uint32_t nscopes) const
...
+  bool success = d_smtEngine->getInterpol(*B.d_expr, result);
+  if (success) {
+    output = Term(output.d_solver, result);
+  }
+  return success;
```
aniemetz, line 4651:
```suggestion
  return success;
  CVC4_API_SOLVER_TRY_CATCH_END;
```

src/api/cvc4cpp.h
```diff
@@ -3010,6 +3010,15 @@ class CVC4_PUBLIC Solver
    */
   void pop(uint32_t nscopes = 1) const;
 
+
+  /**
+   * Get an interpolant
+   * SMT-LIB: ( get-interpol B )
+   * Requires to enable option 'produce-interpols'.
```
aniemetz, line 3017: Describe the arguments with `@param`.

src/smt/command.cpp
```diff
@@ -2139,6 +2139,91 @@ std::string GetSynthSolutionCommand::getCommandName() const
...
+      d_sygus_grammar_type(gtype),
+      d_resultStatus(false)
+{
+}
+
+Expr GetInterpolCommand::getConjecture() const { return d_conj; }
+Type GetInterpolCommand::getGrammarType() const { return d_sygus_grammar_type; }
+Expr GetInterpolCommand::getResult() const { return d_result; }
+
+void GetInterpolCommand::invoke(SmtEngine* smtEngine)
```
aniemetz, line 2161: It would be good if this already used the new API instead of the deprecated old API (which will be removed shortly). For an example, see the `DefineFunctionRecCommand`: https://github.com/CVC4/CVC4/blob/master/src/smt/command.h#L509

src/smt/command.cpp
```diff
@@ -2139,6 +2139,91 @@ std::string GetSynthSolutionCommand::getCommandName() const
...
+          << std::endl;
+    }
+    else
+    {
+      out << "none" << std::endl;
+    }
+  }
+}
+
+Command* GetInterpolCommand::exportTo(ExprManager* exprManager,
```
aniemetz, line 2204: Ditto

src/api/cvc4cpp.cpp
yoni206: Definitely, but I think only in the next PR, this PR does not include implementations.

src/smt/set_defaults.cpp
Ying1123: Who are the interpolation users then?

src/api/cvc4cpp.h
Ying1123: @yoni206 could you take a look that if I added correctly?

src/smt/command.cpp
Ying1123: Good point. Could you give me more clue on how to do this? I tried to add `api::` in front of `Expr` or `SmtEngine` etc, but got errors of "has not been declared".

src/smt/set_defaults.cpp
ahmed-irfan: maybe, I could add here. But first, what is a strong solution?

ajreynol: This refers to logical strength.  For abduction, we always prefer solutions F1 instead of F2 if F2 => F1.

For interpolation, logical strength is not a clear heuristic for which interpolant is best.  I would not recommend enabling this option (you should not change the condition on line 1063).

HanielB: Ideally I think this should be controlled via a separate option so it can be properly evaluated.

ajreynol: Yes, which means that we should not hardcode setting this option by default.  (The extra option you talk about is the one that is being forced to true here).

src/api/cvc4cpp.cpp
aniemetz: Right, so either we can add a stub in the test as to not forget about it, or we need a way to keep track of it. I'd prefer the former if possible.

src/api/cvc4cpp.h
aniemetz: Yes, looks good. Just two minor things (one below). Can you maybe rename B to something more meaningful? Maybe use `term` or something similar? That's what we typically use in the API.

src/api/cvc4cpp.h
```diff
@@ -3010,6 +3010,17 @@ class CVC4_PUBLIC Solver
    */
   void pop(uint32_t nscopes = 1) const;
 
+  /**
+   * Get an interpolant
+   * SMT-LIB: ( get-interpol B )
+   * Requires to enable option 'produce-interpols'.
+   * @param B the conjecture term
+   * @param output a Term I such that A->I and I->B are valid, where A is the
+   * current set of assertions and B is given in the input.
```
aniemetz, line 3019:
```suggestion
   *        current set of assertions and B is given in the input.
```

src/smt/command.cpp
aniemetz: `Expr` is part of the old API, the new API (in `cvc4cpp.h`) wraps `Expr` in `Term`. So what you need is `api::Solver` instead of `SmtEngine`, and then you call `Solver::getInterpol` instead of `SmtEngine::getInterpol`. We can discuss this more offline if you have more questions.

src/smt/set_defaults.cpp
HanielB: Ah, right. :)

ahmed-irfan: yeah, I agree that we should not use this option. 

(Also note that the usefulness of a stronger Interpolant in Model Checking is not evident.)

src/api/cvc4cpp.h
Ying1123: renamed to term.

src/api/cvc4cpp.cpp
```diff
@@ -4693,6 +4693,30 @@ void Solver::pop(uint32_t nscopes) const
   CVC4_API_SOLVER_TRY_CATCH_END;
 }
 
+bool Solver::getInterpolant(Term B, Term& output) const {
```
aniemetz, line 4696: Rename `B` matching to the declaration in the header.

src/api/cvc4cpp.cpp
```diff
@@ -4693,6 +4693,30 @@ void Solver::pop(uint32_t nscopes) const
...
+  Expr result;
+  bool success = d_smtEngine->getInterpol(*B.d_expr, result);
+  if (success) {
+    output = Term(output.d_solver, result);
+  }
+  return success;
+  CVC4_API_SOLVER_TRY_CATCH_END;
+}
+
+bool Solver::getInterpolant(Term B, const Type& gtype, Term& output) const
```
aniemetz, line 4707: Ditto

src/smt/command.cpp
```diff
@@ -2133,6 +2133,90 @@ std::string GetSynthSolutionCommand::getCommandName() const
...
+      d_sygus_grammar_type(gtype),
+      d_resultStatus(false)
+{
+}
+
+api::Term GetInterpolCommand::getConjecture() const { return d_conj; }
+Type GetInterpolCommand::getGrammarType() const { return d_sygus_grammar_type; }
+api::Term GetInterpolCommand::getResult() const { return d_result; }
+
+void GetInterpolCommand::invoke(SmtEngine* smtEngine)
```
aniemetz, line 2158: Parameter `smtEngine` is unused.

src/smt/command.cpp
```diff
@@ -2133,6 +2133,90 @@ std::string GetSynthSolutionCommand::getCommandName() const
...
+          << std::endl;
+    }
+    else
+    {
+      out << "none" << std::endl;
+    }
+  }
+}
+
+Command* GetInterpolCommand::exportTo(ExprManager* exprManager,
```
aniemetz, line 2201: This should not have `ExprManager` as argument -- it will probably even need no argument since you have `api::Solver` as a member.

src/smt/command.cpp
Ying1123: But this is same as in the class `DeclareFunctionRecCommand`. They are inherit from `Command`, which has `invoke` with parameter `smtEngine`. If we want to change all these, it maybe better in another pull request?

src/smt/command.cpp
Ying1123: Same reason here. It is inherit from `Command`.

HanielB: For simplicity we can have this PR using the old way of doing commands and I can change all of them simultaneously afterwards, as discussed in the CVC4 meeting.

PR #4641, opened 2020-06-20
(proof-new) Add proof-new to options file

ajreynol:
Adds proof-new as an option.  This is required for adding code that is guarded by this option while we are in the process of merging work on the new proofs infrastructure.

Enabling the option currently throws an option exception.


src/options/smt_options.toml
```diff
@@ -164,6 +164,23 @@ header = "options/smt_options.h"
   read_only  = true
   help       = "output proofs after every UNSAT/VALID response"
 
+[[option]]
+  name       = "proofNew"
+  category   = "regular"
+  long       = "proof-new"
+  type       = "bool"
+  default    = "false"
+  help       = "do proof production in the new format"
```
HanielB, line 173: new infrastructure rather than new format

PR #4642, opened 2020-06-20
(proof-new) Updates to strings core solver

ajreynol:
This updates the core strings solver in preparation for proofs.

The main changes include:
- The addition of the strings PfRule enum values.
- The definition of a "getConclusion" static method, used by the core solver, and to be used in the future by the strings proof checker.  This includes several optional forms of lemmas, which are added as options in this PR.
- Major simplifications to our inference schemas for disequality handling (a STRING_DECOMPOSE inference rule).  Note this is the only significant intended behavioral change in this PR.
- Minor updates to the form of inferences send to inference manager, for instance to orient equalities in the expected way, and to reorder assumptions. These changes are done for uniformity and make the strings proof reconstruction from inference steps easier.


src/theory/strings/core_solver.cpp
```diff
@@ -550,8 +560,12 @@ void CoreSolver::checkNormalFormsEq()
       NormalForm& nfe_eq = getNormalForm(itn->second);
       // two equivalence classes have same normal form, merge
       std::vector<Node> nf_exp;
-      nf_exp.push_back(utils::mkAnd(nfe.d_exp));
-      nf_exp.push_back(eqc_to_exp[itn->second]);
+      nf_exp.insert(nf_exp.end(), nfe.d_exp.begin(), nfe.d_exp.end());
```
4tXJ7f, line 563: Nit: These two lines can be combined.

src/theory/strings/core_solver.cpp
```diff
@@ -917,16 +1079,17 @@ void CoreSolver::processNEqc(Node eqc,
...
+          std::vector<Node> exp;
           //conflict
           Trace("strings-solve") << "Normal form for " << n << " cannot be contained in constant " << c << std::endl;
           // conflict, explanation is:
           //  n = base ^ base = c ^ relevant porition of ( n = N[n] )
-          std::vector< Node > exp;
-          d_bsolver.explainConstantEqc(n,eqc,exp);
           // Notice although not implemented, this can be minimized based on
           // firstc/lastc, normal_forms_exp_depend.
           exp.insert(exp.end(), nfi.d_exp.begin(), nfi.d_exp.end());
```
4tXJ7f, line 1089: Nit: can be combined with the declaration of `exp`.

src/theory/strings/core_solver.cpp
```diff
@@ -1131,6 +1298,14 @@ void CoreSolver::processSimpleNEq(NormalForm& nfi,
 
     if (d_state.areEqual(xLenTerm, yLenTerm))
     {
+      std::vector<Node> ant;
+      NormalForm::getExplanationForPrefixEq(nfi, nfj, index, index, ant);
+      if (x.isConst() && y.isConst())
+      {
+        // if both are constant, its just a constant conflict
```
4tXJ7f, line 1305:
```suggestion
        // if both are constant, it's just a constant conflict
```

src/theory/strings/core_solver.cpp
```diff
@@ -1486,54 +1629,69 @@ void CoreSolver::processSimpleNEq(NormalForm& nfi,
             Trace("strings-entail")
                 << "  explanation was : " << et.second << std::endl;
             lentTestSuccess = e;
-            lentTestExp = et.second;
+            lenConstraint = entLit;  // et.second;
```
4tXJ7f, line 1632:
```suggestion
            lenConstraint = entLit;
```

src/theory/strings/core_solver.cpp
```diff
@@ -1954,47 +2106,38 @@ void CoreSolver::processDeq(Node ni, Node nj)
...
+        Trace("strings-solve")
+            << "Non-Simple Case 1 : add lemmas " << std::endl;
+        SkolemCache* skc = d_termReg.getSkolemCache();
+
+        for (unsigned r = 0; r < 2; r++)
         {
-          antecNewLits.push_back(ni.eqNode(nj).negate());
+          Node ux = r == 0 ? x : y;
+          Node uxLen = nm->mkNode(STRING_LENGTH, x);
+          Node uyLen = nm->mkNode(STRING_LENGTH, y);
```
4tXJ7f, line 2119: I believe that there is a bug here: If `r == 1`, then we decompose `y` using `len(y)`, instead of `len(x)`.

```suggestion
          Node ux = r == 0 ? x : y;
          Node uxLen = nm->mkNode(STRING_LENGTH, r == 0 ? x : y);
          Node uyLen = nm->mkNode(STRING_LENGTH, r == 0 ? y : x);
```

src/theory/strings/core_solver.h
```diff
@@ -219,6 +217,45 @@ class CoreSolver
...
+
+  /**
+   * This returns the conclusion of the proof rule corresponding to splitting
+   * on the arrangement of terms t and s appearing in an equation of the form
+   *   t ++ t' = s ++ s' or t' ++ t = s' ++ s
+   * where we are in the second case if isRev is true. This method is called
+   * both by the core solver and by the strings proof checker.
+   *
+   * @param t The first term
+   * @param s The second term
```
4tXJ7f, line 229: This is not really true for the `STRING_DECOMPOSE` rule. Maybe we should use different symbols in the equations above and make it clear that `t` and `s` are the characteristic terms for that equation + rule and not necessarily the terms in the equation themselves?

ajreynol: I'll just split to a new utility method, since indeed STRING_DECOMPOSE doesn't fit this pattern very well.

src/theory/strings/core_solver.cpp
ajreynol: Yes you're right, good catch.

src/theory/strings/core_solver.h
```diff
@@ -255,6 +255,25 @@ class CoreSolver
...
+   * length l if isRev is false (resp. true).
+   * 
+   * @param x The string term
+   * @param l The length term
+   * @param isRev Whether the equation is in a reverse direction
+   * @param skc The skolem cache (to allocate fresh variables if necessary)
+   * @param newSkolems The vector to add new variables to
+   * @return The conclusion of the inference.
+   */
+  static Node getDecomposeConclusion(Node x,
```
4tXJ7f, line 272: Apply ClangFormat.

PR #4645, opened 2020-06-22
When printing a separation logic model, ensure that the heap and nil are separate

andrewvaughanj:
## Issue

When formatting a model for a separation logic formula, CVC4 reports the `nil` element as part of the heap:

```
(heap
(sep (pto 1 3735928560) (pto 0 3735928559))
(= (as sep.nil Int) 4222467823)
)
```

which is confusing as the `nil` element *is not* part of the heap!

## Resolution

This PR changes it such that the `nil` term gets printed as its own (separate 😛) part of the model:

```
(heap
(sep (pto 1 3735928560) (pto 0 3735928559))
)
(nil
(= (as sep.nil Int) 4222467823)
)
```

### DIFF
--- a/src/printer/smt2/smt2_printer.cpp
+++ b/src/printer/smt2/smt2_printer.cpp
@@ -1358,13 +1358,16 @@ void Smt2Printer::toStream(std::ostream& out, const Model& m) const
   this->Printer::toStream(out, m);
   out << ")" << endl;
   //print the heap model, if it exists
-  Expr h, neq;
-  if( m.getHeapModel( h, neq ) ){
+  Expr h, nil;
+  if (m.getHeapModel(h, nil))
+  {
     // description of the heap+what nil is equal to fully describes model
     out << "(heap" << endl;
     out << h << endl;
-    out << neq << endl;
-    out << ")" << std::endl;
+    out << ")" << endl;
+    out << "(nil" << endl;
+    out << nil << endl;
+    out << ")" << endl;
   }
 }
 


ajreynol:
I'm not sure I'm in favor of this format, since it adds another (unguarded) keyword `nil` to our output and unfortunately isn't backwards compatible with the previous format.

At the moment, the motivation for putting the definition of nil since it's an additional piece of information pertaining to how the heap is interpreted, since sep.nil has a distinguished semantics.  

Is there a compelling reason why this is necessary?  As in, does this make things significantly easier for the user to parse?

Regardless, a change to the format of the output here could potentially be refactored further, say by using `define-fun` for sep.nil perhaps.

andrewvaughanj:
> Is there a compelling reason why this is necessary? As in, does this make things significantly easier for the user to parse?

In all honesty: no.

I was migrating some of my code from using the legacy CVC4 Python module to use `pycvc4`, which at the time (16 days ago), did not have `getSepHeapExpr` and `getSepNilExpr` (since fixed in #4663; 11 days ago).

As those calls didn't exist, I resorted to using `printModel()`. When I saw that `heap` and `nil` were the "same" term in the model, I just just thought "huh, that's weird, I don't think it is supposed to be like that", and raised this PR.

I wrote my first bit of separation logic code off of the back of FLoC 2018 (when we spoke in person) and this is the first time in two years I've ever looked at the result of `getModel`/`printModel`. If `getSepHeapExpr`/`getSepNilExpr` had already been exposed in `pycvc4`, I probably never would have realised this was an issue ... that's how "minor" this is to me!

All of that being said, I still feel the current presentation isn't correct, because it somehow suggests that `nil` is a `heap` term (which it isn't). Maybe the output should be changed to be something like:

```
(sep
    (heap
    ...
    )
    (nil
    ...
    )
)
```

to make it explicit? Leaving the `heap` as-is and having `nil` as a function would also work (because it makes `heap` and `nil` "distinct").

Just to be clear: I'm also fine for this to be closed as `WONTFIX` for backwards compatibility reasons. There is nothing that I am restricted from doing with the current output.

ajreynol:
Closing this, we may revisit this in the future, but yes, I'd like to keep backwards compatibility for now.

PR #4648, opened 2020-06-23
(proof-new) Update Theory interface for proof-new

ajreynol:
This includes 4 changes:
- Theory constructor takes a ProofNodeManager,
- Theory::explain returns a TrustNode (of kind PROP_EXP),
- Theory::expandDefinitions returns a TrustNode (of kind REWRITE),
- Theory::ppRewrite returns a TrustNode (of kind REWRITE).
These are all currently planned updates to the interface of Theory. 

This PR also connects some of the existing proof rule checkers into the proof checker, if one is provided to the constructor.  It updates TheoryEngine and other places to process TrustNode in trivial ways (converting them into Node). These calls will later be updated as needed for proof support.

This PR is also contingent on the performance tests for proof-new on SMT-LIB.


src/theory/uf/theory_uf.cpp
```diff
@@ -349,6 +357,8 @@ void TheoryUF::presolve() {
   Debug("uf") << "uf: begin presolve()" << endl;
   if(options::ufSymmetryBreaker()) {
     vector<Node> newClauses;
+    // would need to make this proof producing, which is not supposed to be
+    // something easy...
```
HanielB, line 361: Remove this comment. :)

src/smt/process_assertions.cpp
```diff
@@ -738,7 +738,8 @@ Node ProcessAssertions::expandDefinitions(
         theory::Theory* t = d_smt.d_theoryEngine->theoryOf(node);
 
         Assert(t != NULL);
-        node = t->expandDefinition(n);
+        TrustNode trn = t->expandDefinition(n);
+        node = trn.isNull() ? Node(n) : trn.getNode();
```
HanielB, line 742: shouldn't you be using `static_cast` here (and all other such occurrences in the PR)?

src/theory/theory.h
```diff
@@ -579,9 +599,13 @@ class Theory {
    * Given an atom of the theory coming from the input formula, this
    * method can be overridden in a theory implementation to rewrite
    * the atom into an equivalent form.  This is only called just
-   * before an input atom to the engine.
+   * before an input atom to the engine. The result of the method is
```
HanielB, line 602: the first sentence is a bit strange. Rephrase?

src/theory/strings/theory_strings.cpp
```diff
@@ -189,18 +190,20 @@ bool TheoryStrings::propagate(TNode literal) {
...
 }
 
-
-Node TheoryStrings::explain( TNode literal ){
+TrustNode TheoryStrings::explain(TNode literal)
+{
   Debug("strings-explain") << "explain called on " << literal << std::endl;
   std::vector< TNode > assumptions;
   d_im->explain(literal, assumptions);
+  Node ret;
```
HanielB, line 198: You could remove the block below and do:
```suggestion
  Node ret;
  ret = assumptions.empty()
            ? d_true
            : assumptions.size() == 1
                  ? assumptions[0]
                  : NodeManager::currentNM()->mkNode(kind::AND, assumptions);
```
If you don't want to then could you clang-format it, as you are already changing most of the method?

src/smt/process_assertions.cpp
ajreynol: No, this is not a cast, it's a constructor.

src/theory/strings/theory_strings.cpp
ajreynol: This code is deleted in proof-new, let's not worry about this.

src/smt/process_assertions.cpp
HanielB: I see. I have done this via casting, any idea which would be better?

ajreynol: Casting is more roundabout and is generally worse.

HanielB: OK, good to know.

4tXJ7f: I agree. I'd only use casts for pointers and other primitive types.

src/theory/bv/theory_bv.cpp
```diff
@@ -221,17 +224,18 @@ Node TheoryBV::expandDefinition(Node node)
...
   }
     break;
 
   default:
-    return node;
     break;
   }
-
-  Unreachable();
+  if (!ret.isNull())
```
4tXJ7f, line 234: To be more consistent with other theories, we could also require `node != ret` here but it's not strictly necessary.

src/theory/datatypes/theory_datatypes.cpp
```diff
@@ -583,6 +584,7 @@ Node TheoryDatatypes::expandDefinition(Node n)
       }
     }
   }
+  Node n_ret;
```
4tXJ7f, line 587:
```suggestion
  Node ret;
```

(naming conventions)

PR #4650, opened 2020-06-23
New C++ API: Remove examples for old API.

aniemetz:
This removes obsolete examples for the old API in preparation of making
the old API private. Examples for the new API are renamed from
*-new.cpp to *.cpp.


examples/api/bitvectors.cpp
```diff
@@ -1,8 +1,8 @@
 /*********************                                                        */
-/*! \file bitvectors.cpp
+/*! \file bitvectors-new.cpp
```
ajreynol, line 2: Fix name (same for other files).

aniemetz: Whoops!

PR #4655, opened 2020-06-24
Re-implement handling of --tlimit

nafur:
As a first step within this project, this PR provides a new implementation that backs `--tlimit`. It uses `setitimer` (as `timer_settime` is not available on MacOS) to make the OS send a signal after the given wall clock time has passed.
In more detail, this PR:
- removes the current handling of `--tlimit` (`TlimitListener` and its integration in the `NodeManager`)
- adds a new `TimeLimitListener` that lives in `src/main`
- uses `TimeLimitListener` directly in `runCvc4()`
- adds a signal handler for `SIGALRM` (that also uses the existing `timeout_handler`)


ajreynol:
There are some design conflicts with my branch that decouples options and NodeManager.  In particular, TLimitListener and RLimitListener are deleted on my branch.  We should discuss this in more detail.

nafur:
Okay, then let's wait with this PR until yours has been merged.

src/main/time_limit.h
```diff
@@ -0,0 +1,35 @@
...
+#include "options/options.h"
+
+namespace CVC4 {
+namespace main {
+
+/** Manages an overall wall-clock time limit for the solver binary.
+ * It works as a listener to an Options object.
+ * When notified, it retrieves the time limit and creates a POSIX timer (via
+ * setitimer()). This timer signals its expiration with an SIGALRM that is
+ * handled by timeout_handler() in util.cpp Note that this class expects a
```
mpreiner, line 16:
```suggestion
 * handled by timeout_handler() in util.cpp. Note that this class expects a
```

PR #4660, opened 2020-06-25
(proof-new) Proof rule checkers run on skolem forms

ajreynol:
Previously, the proof rule checkers were run on witness form for convenience.  However, it is more flexible for the sake of the internal calculus to run on Skolem forms.  The main reason is that the conversion to/from witness form does not preserve terms that contain `witness`.  Our preprocessing steps rely on using witness terms.  This design change additionally simplifies a lot of the code of the builtin proof rule checker.  

Instead, witness forms can be queried on an as-needed basis, e.g. in MACRO_SR_PRED_TRANSFORMATION.


src/theory/builtin/proof_checker.h
```diff
@@ -67,24 +67,30 @@ class BuiltinProofRuleChecker : public ProofRuleChecker
...
    * specifying a call to Rewriter::rewrite.
    * @return The rewritten form of n.
    */
   static Node applyRewrite(Node n, MethodId idr = MethodId::RW_REWRITE);
   /**
-   * Apply substitution on n (in witness form). This encapsulates the exact
-   * behavior of a SUBS step in a proof. Substitution is on the Skolem form of
-   * n.
+   * Get substitution. Updates vars/subs to the substitution specified by
+   * exp for the substitution method ids.
```
HanielB, line 81:
```suggestion
   * exp (e.g. as an equality) for the substitution method ids.
```

src/theory/builtin/proof_checker.h
```diff
@@ -67,24 +67,30 @@ class BuiltinProofRuleChecker : public ProofRuleChecker
...
+
+  /**
+   * Apply substitution on n in skolem form. This encapsulates the exact
+   * behavior of a SUBS step in a proof.
    *
-   * @param n The node (in witness form) to substitute,
-   * @param exp The (set of) equalities (in witness form) corresponding to the
-   * substitution
+   * @param n The node to substitute,
+   * @param exp The (set of) equalities corresponding to the substitution
```
HanielB, line 93: Not really equalities, right? Maybe "set of terms"? Or "set of equality literals or predicates"?

PR #4661, opened 2020-06-26
(proof-new) Improve rewriter for WITNESS

ajreynol:
Proof checking failures revealed that we are not rewriting witness for Boolean variables `(witness ((x Bool)) x) ---> true` and `(witness ((x Bool)) (not x)) ---> false`.

Also adds 2 assertions that are required for elimination `(witness ((x T)) (= x t)) ---> t`.  These assertions should always hold due to the witness terms we construct.


src/theory/builtin/theory_builtin_rewriter.cpp
```diff
@@ -100,10 +100,26 @@ RewriteResponse TheoryBuiltinRewriter::postRewrite(TNode node) {
       {
         if (node[1][i] == node[0][0])
```
HanielB, line 101: Please add the high-level abstraction of the rewrite, as is in the PR description

src/theory/builtin/theory_builtin_rewriter.cpp
```diff
@@ -100,10 +100,26 @@ RewriteResponse TheoryBuiltinRewriter::postRewrite(TNode node) {
...
+                                   << node[1][1 - i] << std::endl;
+          // based on the witness terms we construct, this should be a legal
+          // elimination
+          Assert(!expr::hasSubterm(node[1][1 - i], node[0][0]));
+          Assert(node[1][i].getType().isSubtypeOf(node[0][0].getType()));
           return RewriteResponse(REWRITE_DONE, node[1][1 - i]);
         }
       }
     }
+    else if (node[1] == node[0][0])
```
HanielB, line 113: ditto

src/theory/builtin/theory_builtin_rewriter.cpp
```diff
@@ -100,10 +100,26 @@ RewriteResponse TheoryBuiltinRewriter::postRewrite(TNode node) {
...
+    {
+      return RewriteResponse(REWRITE_DONE,
+                             NodeManager::currentNM()->mkConst(true));
+    }
+    else if (node[1].getKind() == kind::NOT && node[1][0] == node[0][0])
```
HanielB, line 118: ditto

PR #4663, opened 2020-06-26
Add API for retrieving separation heap/nil term

4tXJ7f:
This commit extends the API to support the retrieval of heap/nil term
when separation logic is used and changes the corresponding system test
accordingly. This commit is in preparation of making the constructor of
`ExprManager` private.


src/api/cvc4cpp.h
```diff
@@ -3003,6 +3003,18 @@ class CVC4_PUBLIC Solver
    */
   std::vector<Term> getValue(const std::vector<Term>& terms) const;
 
+  /**
+   * When using separation logic, obtain the term for the heap.
+   * @return The term for the heap
+   */
+  Term getSeparationHeapTerm() const;
```
ajreynol, line 3010: I think the better name for this is `getSeparationHeap`. The reason is that the heap in separation logic is a finite map (although it is expressed by a `Term`, it is not a term).

4tXJ7f: `getSeparationHeap` sounds good to me.

PR #4664, opened 2020-06-26
 (proof-new) Updates to skolem manager interface

ajreynol:
Adds a fix for mkPurifySkolem and introduces new interfaces in preparation for arithmetic operator elimination and term formula removal proofs.


src/expr/skolem_manager.cpp
```diff
@@ -55,12 +57,19 @@ Node SkolemManager::mkSkolem(Node v,
   // store the mapping to proof generator if it exists
   if (pg != nullptr)
   {
-    Node q = nm->mkNode(EXISTS, w[0], w[1]);
+    // we cache based on the original (Skolem) form
```
HanielB, line 60: say why

src/expr/skolem_manager.h
```diff
@@ -86,14 +86,21 @@ class SkolemManager
    * @param pg The proof generator for this skolem. If non-null, this proof
    * generator must respond to a call to getProofFor(exists v. pred) during
    * the lifetime of the current node manager.
+   * @param retWitness Whether we wish to return the witness term for the
+   * given Skolem, which notice is of the form (witness v. pred), where pred
+   * is in Skolem form. A typical use case of setting this flag to true
+   * is preprocessing passes that eliminate terms. Using a witness term
+   * instead of its corresponding Skolem indicates that the body of the witness
+   * term needs to be added as an assertion, e.g. by the term formula remover.
```
HanielB, line 94: I dislike that this is creating a bogus term. Isn't there a better way of tracking which assertion we should add?

ajreynol: It's not a bogus term though, its just one that is not purely in witness form or in skolem form.  The purpose of the witness terms is to track which assertion to add, in many cases you want to add an assertion containing skolems.

HanielB: This is to save converting `predw` to `pred` back again, right? Could you say that explicitly in the documentation, that this is a small optimization?

ajreynol: No, this is a much deeper issue.

Take arithmetic's operator elimination techniques for term `t`. Currently, e.g. arithmetic operator elimination is returning the witness form `witness x. P[w_s]` for the eliminated form of `t`, where the subterms that its witness form depends on `s` have been converted to witness form. This means that if `s` itself contains Skolem, we are actually digging into `t` and returning (a potentially large) witness term during ppRewrite.  This happens to work since term formula removal caches results, but it will have to recompute the skolem form `s` given input `w_s`, and moreover will have to do this repeatedly for terms with nested skolems.  This PR allows one to simply do `witness x. P[s]` as a returned term during ppRewrite, which avoids this behavior and is more accurate to the intention of the preprocessing pass (which only concerns one level).

HanielB: I see. Could you add this detailed explanation to the documentation?

ajreynol: I think this is more suited to be explained where it is actually performed.  This documentation should just concern what the method does, and not what it could be used for.

HanielB: Fair enough. But when that explanation is added I think it'd be good to link to it in this method so that somebody looking only at it doesn't thing "why create this funny term?".

PR #4665, opened 2020-06-26
Proof Rules and Checker for Arithmetic

alex-ozdemir:
This PR introduces proof rules for arithmetic and their checker.

The code is dead, since these rules are never used.

### DIFF
--- a/src/expr/proof_rule.h
+++ b/src/expr/proof_rule.h
@@ -507,6 +507,69 @@ enum class PfRule : uint32_t
   // sigma maps x1 ... xn to t1 ... tn.
   INSTANTIATE,
 
+  // Note: an ArithLiteral is a term of the form (>< poly const)
+  // where
+  //   >< is >=, >, ==, <, <=, or not(== ...).
+  //   poly is a polynomial
+  //   const is a rational constant
+
+  // Children: (P1:l1, ..., Pn:ln)
+  //           where each li is an ArithLiteral
+  //           not(= ...) is dis-allowed!
+  //
+  // Arguments: (k1, ..., kn), non-zero reals
+  // ---------------------
+  // Conclusion: (>< (* k t1) (* k t2))
+  //    where >< is the fusion of the combination of the ><i, (flipping each it
+  //    its ki is negative). >< is always one of <, <=
+  //    NB: this implies that lower bounds must have negative ki,
+  //                      and upper bounds must have positive ki.
+  //    t1 is the sum of the polynomials.
+  //    t2 is the sum of the constants.
+  SCALE_SUM_UPPER_BOUNDS,
+
+  // ======== Tightening Strict Integer Upper Bounds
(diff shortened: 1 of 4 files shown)


src/expr/proof_rule.h
```diff
@@ -507,6 +507,69 @@ enum class PfRule : uint32_t
   // sigma maps x1 ... xn to t1 ... tn.
   INSTANTIATE,
 
+  // Note: an ArithLiteral is a term of the form (>< poly const)
+  // where
+  //   >< is >=, >, ==, <, <=, or not(== ...).
+  //   poly is a polynomial
+  //   const is a rational constant
+
```
ajreynol, line 515: Can you make sure this is formatted the same as above (e.g. with `======`).

src/expr/proof_rule.h
```diff
@@ -507,6 +507,69 @@ enum class PfRule : uint32_t
...
+  // Children: (A B)
+  // Arguments: (C)
+  // ---------------------
+  // Conclusion: (C),
+  //                 where (not A) (not B) and C
+  //                   are (> x c) (< x c) and (= x c)
+  //                   in some order
+  //                 note that "not" here denotes arithmetic negation, flipping
+  //                 >= to <, etc.
+  TRICHOTOMY,
```
ajreynol, line 557: Wondering if we should rename to `ARITH_TRICHOTOMY`.

src/expr/proof_rule.h
```diff
@@ -507,6 +507,69 @@ enum class PfRule : uint32_t
...
+  // Arguments: (k1, ..., kn), non-zero reals
+  // ---------------------
+  // Conclusion: (>< (* k t1) (* k t2))
+  //    where >< is the fusion of the combination of the ><i, (flipping each it
+  //    its ki is negative). >< is always one of <, <=
+  //    NB: this implies that lower bounds must have negative ki,
+  //                      and upper bounds must have positive ki.
+  //    t1 is the sum of the polynomials.
+  //    t2 is the sum of the constants.
+  SCALE_SUM_UPPER_BOUNDS,
```
ajreynol, line 529: Might be best to keep the rule naming so they are prefixed by theories, so this would be `ARITH_SCALE_SUM_UPPER_BOUNDS`.  (The ones with `INT_` are fine already, since i guess these are specific to integers).

src/expr/proof_rule.h
alex-ozdemir: Sure. Done in  8c5438338.

src/expr/proof_rule.h
alex-ozdemir: Sure. Done in 15173a063.

src/expr/proof_rule.h
alex-ozdemir: Sure. Done in 15173a063.

PR #4668, opened 2020-06-28
Add internal support for integer and operator

ajreynol:
Towards merging iand branch to master. This adds internal support for an "integer AND" operator.


src/theory/arith/arith_rewriter.cpp
```diff
@@ -371,6 +372,50 @@ RewriteResponse ArithRewriter::postRewriteMult(TNode t){
...
+      continue;
+    }
+    if (t[i].getConst<Rational>().sgn() == 0)
+    {
+      // ((_ iand k) 0 y) ---> 0
+      return RewriteResponse(REWRITE_DONE, t[i]);
+    }
+    // TODO ((_ iand k) 2^k-1 y) ---> y
+
+    // constants c out of bounds can be normalized to (mod c 2^k)
```
yoni206, line 415: What happens if they don't?

src/theory/arith/kinds
```diff
@@ -133,4 +133,15 @@ nullaryoperator PI "pi"
 
 typerule PI ::CVC4::theory::arith::RealNullaryOperatorTypeRule
 
+
+constant IAND_OP \
+	::CVC4::IntAnd \
+	"::CVC4::UnsignedHashFunction< ::CVC4::IntAnd >" \
+	"util/iand.h" \
```
yoni206, line 140: Can you document somewhere the precise semantics, especially for cases where the arguments are negative or too big?

src/theory/arith/arith_rewriter.cpp
```diff
@@ -371,6 +372,50 @@ RewriteResponse ArithRewriter::postRewriteMult(TNode t){
...
+    if (!t[i].isConst())
+    {
+      continue;
+    }
+    if (t[i].getConst<Rational>().sgn() == 0)
+    {
+      // ((_ iand k) 0 y) ---> 0
+      return RewriteResponse(REWRITE_DONE, t[i]);
+    }
+    // TODO ((_ iand k) 2^k-1 y) ---> y
```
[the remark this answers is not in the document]
ajreynol: Let's worry about this on future PR.

src/theory/arith/arith_rewriter.cpp
ajreynol: This is a rewrite rule, it doesn't need to be applied, this is another TODO, I'll leave this for later.

src/theory/arith/kinds
ajreynol: Done.

yoni206: Thanks. I guess it is fixed now, modulo the fact that I did not find a specification of INT_TO_BITVECTOR (specifically for the case of negatives/out-of-bounds). If there is one, can you send a link? If there is none, maybe this can be added in a separate PR?

ajreynol: It's here: 
https://github.com/CVC4/CVC4/pull/4668/commits/a76e594560440f683092f09f2cfe0f92a5601127#diff-63290c50fbea42b02ed8329a2666682bR136
The semantics specified here covers all integers i1, i2.

yoni206: yes, but it relies on an understanding of `int2bv` operator, this is the one I was asking about.

PR #4672, opened 2020-06-29
fix: add missing cmake command to INSTALL.md steps

pnwamk:


mpreiner:
This line should not be necessary since configure already calls cmake here: https://github.com/CVC4/CVC4/blob/master/configure.sh#L481 . Did you encounter any issues with this setup?

pnwamk:
Ah, my mistake. During the configuration it indeed invoked cmake but there was an error due to a missing Python module (`toml`) and I didn't connect the two until just now. Apologies for the noise!

PR #4677, opened 2020-06-30
Interpolation Step 2

Ying1123:
This is the second step of adding Interpolation. The whole change will be adding the API for (get-interpol s B), which is aim for computes an I that A->I and I->B. Here A is the assertions in the stack.

The second step creates the component for computing interpolation, while omits the implementation.


src/theory/quantifiers/sygus/sygus_interpol.h
```diff
@@ -0,0 +1,179 @@
...
+  bool SolveInterpolation(const std::string& name,
+                          const std::vector<Node>& axioms,
+                          const Node& conj,
+                          const TypeNode& itpGType,
+                          Expr& interpol);
+
+ private:
+  /**
+   * Collects symbols from axioms (axioms) and conjecture (conj), which stored
+   * in d_syms. And computes the shared symbols between axioms and conjecture,
```
yoni206, line 78:
```suggestion
   * in d_syms, and computes the shared symbols between axioms and conjecture,
```

src/theory/quantifiers/sygus/sygus_interpol.h
```diff
@@ -0,0 +1,179 @@
...
+   * stored in d_symsShared.
+   */
+  void collectSymbols(const std::vector<Node>& axioms, const Node& conj);
+
+  /**
+   * Creates free variables and shared free variables from d_syms and
+   * d_symsShared, which stored in d_vars and d_varsShared. And also creates the
```
yoni206, line 85:
```suggestion
   * d_symsShared, which are stored in d_vars and d_varsShared. And also creates the
```

src/theory/quantifiers/sygus/sygus_interpol.h
```diff
@@ -0,0 +1,179 @@
...
+   * same set of variables for formal argument list, which stored in d_vlvs and
+   * d_vlvsShared. Extracts the types of shared variables, which stored in
+   * d_varTypesShared. Creates the formal argument list of the
+   * interpol-to-synthesis, stored in d_abvlShared.
+   *
+   * The argument needsShared denotes if we want to restrict the argument list
```
yoni206, line 91: Can you document when would `needsShared` be false?

src/theory/quantifiers/sygus/sygus_interpol.h
```diff
@@ -0,0 +1,179 @@
...
+   * of the interpol-to-synthesis to be over shared variables or all the
+   * variables.
+   */
+  void createVariables(bool needsShared);
+
+  /**
+   * Set up the grammar for the interpol-to-synthesis.
+   *
+   * The user-defined grammar will be encoded by itpGType. In DEFAULT option, it
```
yoni206, line 100: Can you mention where this option is set (something like `options::smt::interpolationMode` or something like that?

src/theory/quantifiers/sygus/sygus_interpol.h
```diff
@@ -0,0 +1,179 @@
...
+   *
+   * The output argument interpol is the solution to the synthesis conjector.
+   *
+   * The relationship between the free variables of asserts and the formal
+   * argument list of the interpol-to-synthesize are tracked by the attribute
+   * SygusVarToTermAttribute.
+   *
+   * In particular, solutions to the synthesis conjecture will be in the form
+   * of a closed term (lambda varlist. t). The intended solution, which is a
+   * term whose free variables are a subset of asserts, is the term
```
HanielB, line 66:
```suggestion
   * term whose free variables correspond to a subset of the free symbols in asserts, is the term
```

src/theory/quantifiers/sygus/sygus_interpol.h
```diff
@@ -0,0 +1,179 @@
...
+{
+ public:
+  SygusInterpol();
+
+  SygusInterpol(LogicInfo logic);
+
+  /**
+   * Returns the sygus conjecture corresponding to the interpolation problem for
+   * input problem (F above) given by axioms (Fa above), and conj (Fc above).
+   *  Note that axioms is expected to be a subset of asserts.
```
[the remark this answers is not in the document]
HanielB: You starting use `asserts` here but you do not define it.

src/theory/quantifiers/sygus/sygus_interpol.h
```diff
@@ -0,0 +1,179 @@
...
+   * The argument name is the name for the interpol-to-synthesize.
+   *
+   * The type itpGType (if non-null) is a sygus datatype type that encodes the
+   * grammar that should be used for solutions of the interpolation conjecture.
+   *
+   * The output argument interpol is the solution to the synthesis conjector.
+   *
+   * The relationship between the free variables of asserts and the formal
+   * argument list of the interpol-to-synthesize are tracked by the attribute
+   * SygusVarToTermAttribute.
```
HanielB, line 62: Better describe what is the relationship and moreover why it is necessary.

src/theory/quantifiers/sygus/sygus_interpol.h
```diff
@@ -0,0 +1,179 @@
...
+   * Collects symbols from axioms (axioms) and conjecture (conj), which stored
+   * in d_syms. And computes the shared symbols between axioms and conjecture,
+   * stored in d_symsShared.
+   */
+  void collectSymbols(const std::vector<Node>& axioms, const Node& conj);
+
+  /**
+   * Creates free variables and shared free variables from d_syms and
+   * d_symsShared, which stored in d_vars and d_varsShared. And also creates the
+   * same set of variables for formal argument list, which stored in d_vlvs and
```
HanielB, line 86:
```suggestion
   * corresponding set of variables for the formal argument list, which is stored in d_vlvs and
```

src/theory/quantifiers/sygus/sygus_interpol.h
```diff
@@ -0,0 +1,179 @@
...
+   *   (assert A)
+   *   (get-interpol s B)
+   * In the copy of the SMT engine where these commands are issued, we maintain
+   * A in the assertion stack. To solve the interpolation problem, instead of
+   * modifying the assertion stack to remove A and add the sygus conjecture
+   * (exists I. ...), we invoke a fresh copy of the SMT engine and leave the
+   * assertion stack unchaged. This copy of the SMT engine can be further
+   * queried for information regarding further solutions.
+   */
+  std::unique_ptr<SmtEngine> d_subsolver;
```
HanielB, line 149:
```suggestion
  std::unique_ptr<SmtEngine> d_subSolver;
```

src/theory/quantifiers/sygus/sygus_interpol.h
```diff
@@ -0,0 +1,179 @@
...
+   *
+   * This is a fresh copy of the SMT engine which is used for making calls
+   * especially for interpolation problem. In particular, consider the input:
+   *   (assert A)
+   *   (get-interpol s B)
+   * In the copy of the SMT engine where these commands are issued, we maintain
+   * A in the assertion stack. To solve the interpolation problem, instead of
+   * modifying the assertion stack to remove A and add the sygus conjecture
+   * (exists I. ...), we invoke a fresh copy of the SMT engine and leave the
+   * assertion stack unchaged. This copy of the SMT engine can be further
```
HanielB, line 146:
```suggestion
   * original assertion stack unchaged. This copy of the SMT engine can be further
```

src/theory/quantifiers/sygus/sygus_interpol.h
```diff
@@ -0,0 +1,179 @@
...
+   * queried for information regarding further solutions.
```
HanielB, line 147: Give more details about this.

src/theory/quantifiers/sygus/sygus_interpol.h
```diff
@@ -0,0 +1,179 @@
...
+   * Get the synthesis solution, stored in interpol.
+   *
+   * The argument itp is the interpolation predicate.
+   */
+  bool findInterpol(Expr& interpol, Node itp);
+
+  /** The SMT engine subsolver
+   *
+   * This is a fresh copy of the SMT engine which is used for making calls
+   * especially for interpolation problem. In particular, consider the input:
```
HanielB, line 139:
```suggestion
   * This is a fresh copy of the SMT engine which is used for solving
   * the interpolation problem. In particular, * consider the input:
```

src/theory/quantifiers/sygus/sygus_interpol.h
```diff
@@ -0,0 +1,179 @@
...
+   * In the copy of the SMT engine where these commands are issued, we maintain
+   * A in the assertion stack. To solve the interpolation problem, instead of
+   * modifying the assertion stack to remove A and add the sygus conjecture
+   * (exists I. ...), we invoke a fresh copy of the SMT engine and leave the
+   * assertion stack unchaged. This copy of the SMT engine can be further
+   * queried for information regarding further solutions.
+   */
+  std::unique_ptr<SmtEngine> d_subsolver;
+
+  // The logic for the local copy of SMT engine (d_subsolver).
```
HanielB, line 151: change comment format (here and below)

src/theory/quantifiers/sygus/sygus_interpol.h
```diff
@@ -0,0 +1,179 @@
...
+  // variables created from d_syms for formal argument list.
+  std::vector<Node> d_vlvs;
+  // free variables created from d_symsShared.
+  std::vector<Node> d_varsShared;
+  // variables created from d_symShared for formal argument list.
+  std::vector<Node> d_vlvsShared;
+  // types of shared variables between axioms and conjecture.
+  std::vector<TypeNode> d_varTypesShared;
+  // formal argument list of the interpol-to-synthesis.
+  Node d_abvlShared;
```
HanielB, line 169: why abvl?

src/theory/quantifiers/sygus/sygus_interpol.cpp
```diff
@@ -0,0 +1,95 @@
...
+#include "printer/sygus_print_callback.h"
+#include "theory/datatypes/sygus_datatype_utils.h"
+#include "theory/quantifiers/quantifiers_attributes.h"
+#include "theory/quantifiers/quantifiers_rewriter.h"
+#include "theory/quantifiers/sygus/sygus_grammar_cons.h"
+#include "theory/quantifiers/term_util.h"
+#include "theory/rewriter.h"
+
+using namespace CVC4::kind;
+using namespace std;
```
HanielB, line 32: We've been leaning more towards explicitly using the prefixes `std` and `kind` in the code, rather than assuming the namespace.

aniemetz: Yes, we strongly encourage to not `using namespace std`. We should probably make this a style guideline and actually not permit it, what do you think @4tXJ7f @mpreiner @ajreynol ?

Ying1123: I agree with you. This is been added accidentally.

src/theory/quantifiers/sygus/sygus_interpol.h
Ying1123: you mean why this name? I don't know, I copied from get-abduct

HanielB: Could you say why the name @ajreynol ? Not sure if the name just carries automatically to the interpolation case.

ajreynol: abduct bound variable list maybe?

Ying1123: Oh, so I should call it ibvl. Cool

src/theory/quantifiers/sygus/sygus_interpol.h
Ying1123: This is also copied from abduction, @ajreynol could you answer this question?

ajreynol: Are you using this attribute?

Ying1123: I tried to remove it, and it did passes the test cases. So I will remove it.

4tXJ7f:
When we merge this, it would be great if we could change the title, e.g. "Interpolation: Add interface for SyGuS interpolation module".

PR #4682, opened 2020-07-01
Transfer ownership of internal Options from NodeManager to SmtEngine

ajreynol:
This PR decouples Options from NodeManager. Instead, options now live in SmtEngine.

The changes that were required for this PR include:
- The main internal options object is now owned by SmtEngine instead of ExprManager.
- The ownership resource manager is moved from NodeManager to SmtEngine.
- Node manager listeners are deleted, timeouts and resource limits are set during SmtEngine::finishInit.
- A temporary hack was added to make the last constructed SmtEngine to be the one in scope. This ensures that options are in scope whenever an SmtEngine is created.
- The methods for invoking "subsolvers" (theory/smt_engine_subsolver.h,cpp) was simplified, as versions of these calls that change options do not have to clone a new copy of the ExprManager anymore.
- Resource manager was removed from the smt2 parser.
- Minor refactoring was done in SmtEngine to copy "original options" so that options are restored to their state after parsing command line options on reset.
- Updates to unit tests to ensure conformance to new options scoping.


src/expr/expr_manager_template.cpp
```diff
@@ -907,9 +886,6 @@ Type ExprManager::getType(Expr expr, bool check)
 }
 
 Expr ExprManager::mkVar(const std::string& name, Type type, uint32_t flags) {
-  Assert(NodeManager::currentNM() == NULL)
```
nafur, line 910: How is this related to this PR?

src/options/quantifiers_options.toml
```diff
@@ -1608,15 +1608,6 @@ header = "options/quantifiers_options.h"
   default    = "false"
   help       = "compute backwards filtering to compute whether previous solutions are filtered based on later ones"
 
-
-[[option]]
```
nafur, line 1612: Is this related to this PR?

src/parser/parser.cpp
```diff
@@ -727,19 +725,12 @@ Command* Parser::nextCommand()
     }
   }
   Debug("parser") << "nextCommand() => " << cmd << std::endl;
-  if (cmd != NULL && dynamic_cast<SetOptionCommand*>(cmd) == NULL &&
-      dynamic_cast<QuitCommand*>(cmd) == NULL) {
-    // don't count set-option commands as to not get stuck in an infinite
```
nafur, line 732: Are we sure that the problem mentioned in the comment is no longer an issue?

src/options/quantifiers_options.toml
ajreynol: Yes, this option was used for determining whether we would try to clone an ExprManager when calling subsolvers (to set optimal options on that subsolver).  Since cloning the ExprManager is no longer necessary, this option is no longer necessary.

src/smt/set_defaults.cpp
```diff
@@ -1401,24 +1401,6 @@ void setDefaults(SmtEngine& smte, LogicInfo& logic)
     options::bvLazyRewriteExtf.set(false);
   }
 
-  if (!options::sygusExprMinerCheckUseExport())
```
nafur, line 1404: How is this related to this PR?

src/parser/parser.cpp
ajreynol: The issue being referenced (I believe) is talking about a potential bad behavior of the resource manager itself. Thus, deleting this block of code makes the issue irrelevant.

src/smt/set_defaults.cpp
ajreynol: See comment explaining why this option is no longer necessary.

src/smt/smt_engine.cpp
```diff
@@ -942,7 +977,17 @@ LogicInfo SmtEngine::getUserLogicInfo() const
...
-void SmtEngine::setFilename(std::string filename) { d_filename = filename; }
+
+void SmtEngine::setFilename(std::string filename)
+{
+  d_filename = filename;
+
+  // Copy the original options. This is called prior to beginning parsing.
+  // Hence reset should revert to these options, which note is after reading
+  // the command line.
+  d_originalOptions.copyValues(d_options);
```
nafur, line 988: Why is this done in `setFilename`?

src/smt/smt_engine.cpp
```diff
@@ -2535,8 +2580,11 @@ void SmtEngine::checkUnsatCore() {
   Notice() << "SmtEngine::checkUnsatCore(): generating unsat core" << endl;
   UnsatCore core = getUnsatCore();
 
-  SmtEngine coreChecker(d_exprManager);
+  SmtEngine coreChecker(d_exprManager, &d_options);
+  coreChecker.setIsInternalSubsolver();
   coreChecker.setLogic(getLogicInfo());
+  coreChecker.setOption("check-unsat-cores", false);
```
nafur, line 2586: Has this change (of disabling unsat cores and proof) anything to do with this PR?

ajreynol: This was done before, but in a more roundabout way (see diff).

src/smt/smt_engine.cpp
ajreynol: I could change the name of this method to `notifyStartParsing`? Regardless it will probably be the case that this is refactored soon.

src/theory/quantifiers/expr_miner.cpp
```diff
@@ -69,32 +69,16 @@ Node ExprMiner::convertToSkolem(Node n)
...
-                                  ExprManagerMapCollection& varMap,
-                                  Node query,
-                                  bool& needExport)
+void ExprMiner::initializeChecker(std::unique_ptr<SmtEngine>& checker,
+                                  Node query)
 {
   // Convert bound variables to skolems. This ensures the satisfiability
   // check is ground.
   Node squery = convertToSkolem(query);
-  if (options::sygusExprMinerCheckUseExport())
```
nafur, line 81: Why is this option no longer checked? Has it been removed?

src/theory/quantifiers/sygus/sygus_repair_const.cpp
```diff
@@ -229,48 +228,17 @@ bool SygusRepairConst::repairSolution(Node sygusBody,
...
-  bool needExport = true;
-  std::unique_ptr<SmtEngine> simpleSmte;
-  std::unique_ptr<api::Solver> slv;
-  ExprManager* em = nullptr;
-  SmtEngine* repcChecker = nullptr;
-  ExprManagerMapCollection varMap;
-
-  if (options::sygusRepairConstTimeout.wasSetByUser())
-  {
-    // To support a separate timeout for the subsolver, we need to create
```
nafur, line 247: So we no longer want/need this possibility?

src/theory/quantifiers/expr_miner.cpp
ajreynol: Yes.

src/smt/smt_engine.cpp
nafur: I'd suggest renaming it.

src/expr/expr_manager_template.cpp
ajreynol: This is a rather bizarre assertion that checks whether there is a node manager in scope to determine if we are calling this method internally or externally.

After this PR, this distinction no longer holds, since NodeManager is always in scope whenever an SmtEngine has been created.

src/smt/smt_engine_scope.cpp
```diff
@@ -46,9 +46,16 @@ ProofManager* currentProofManager() {
...
+ResourceManager* currentResourceManager()
+{
+  return s_smtEngine_current->getResourceManager();
+}
+
 SmtScope::SmtScope(const SmtEngine* smt)
     : NodeManagerScope(smt->d_nodeManager),
-      d_oldSmtEngine(s_smtEngine_current) {
+      d_oldSmtEngine(s_smtEngine_current),
+      d_optionsScope(smt ? &const_cast<SmtEngine*>(smt)->getOptions() : nullptr)
```
4tXJ7f, line 57: We should not use `const_cast`. Why not make the argument to SmtScope non-const?

src/theory/smt_engine_subsolver.cpp
```diff
@@ -41,60 +41,40 @@ Result quickCheck(Node& query)
   return Result(Result::SAT_UNKNOWN, Result::REQUIRES_FULL_CHECK);
 }
 
-void initializeSubsolver(SmtEngine* smte)
+void initializeSubsolver(std::unique_ptr<SmtEngine>& smte,
```
4tXJ7f, line 44: I have a slight preference for actually just returning a `std::unique_ptr<SmtEngine>` but we can address that later.

src/smt/smt_engine.cpp
```diff
@@ -647,7 +631,7 @@ class SmtEnginePrivate : public NodeManagerListener {
 
 }/* namespace CVC4::smt */
 
-SmtEngine::SmtEngine(ExprManager* em)
+SmtEngine::SmtEngine(ExprManager* em, Options* optr)
```
4tXJ7f, line 634: Minor question: Why `optr` instead of just `opt`?

src/smt/smt_engine.cpp
4tXJ7f, line 2586: Nit: Why use the external way of setting options here instead of `coreChecker.getOptions()[options::checkUnsatCores]`?

src/smt/smt_engine.cpp
```diff
@@ -705,6 +709,35 @@ SmtEngine::SmtEngine(ExprManager* em)
...
+  }
+  if (options::cumulativeMillisecondLimit() != 0)
+  {
+    d_resourceManager->setTimeLimit(options::cumulativeMillisecondLimit(),
+                                    true);
+  }
+  if (options::cpuTime())
+  {
+    d_resourceManager->useCPUTime(true);
+  }
```
4tXJ7f, line 739: One thing that we should consider here: Should we do those in `SmtEngine::setOption()` since they are options that could change after the engine has finished initializing? I think that's why we may have had these callbacks.

test/regress/regress2/quantifiers/sygus-inst-ufbv-sdlx-fixpoint-5.smt2
```diff
@@ -1,4 +1,4 @@
-; COMMAND-LINE: --sygus-inst
+; COMMAND-LINE: --sygus-inst --no-check-unsat-cores
```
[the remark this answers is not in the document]
ajreynol: No, unfortunately this benchmark went to timing out with unsat cores incidentally on this PR, I'm not sure why.

src/theory/theory_engine.h
```diff
@@ -363,6 +363,7 @@ class TheoryEngine {
   /** Constructs a theory engine */
   TheoryEngine(context::Context* context,
                context::UserContext* userContext,
+               ResourceManager* rm,
```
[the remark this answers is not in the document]
ajreynol: I'm not sure about doing this, since it makes the conceptual dependencies circular (TheoryEngine <-> SmtEngine).  Let's think about this  more after this PR.

src/expr/node_manager.cpp
```diff
@@ -93,11 +92,8 @@ namespace attr {
 typedef expr::Attribute<attr::LambdaBoundVarListTag, Node> LambdaBoundVarListAttr;
 
 NodeManager::NodeManager(ExprManager* exprManager)
-    : d_options(new Options()),
-      d_statisticsRegistry(new StatisticsRegistry()),
-      d_resourceManager(new ResourceManager(*d_statisticsRegistry, *d_options)),
+    : d_statisticsRegistry(new StatisticsRegistry()),
```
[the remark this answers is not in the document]
ajreynol: Sounds good.

src/smt/smt_engine.cpp
ajreynol: `optr` is "option pointer" :)

src/smt/smt_engine.cpp
ajreynol: Sure I can change this.

src/smt/smt_engine.cpp
ajreynol: Yes, I have (several) branches that attempt to refactor this which do what you suggest. Let's address on future PR.

src/smt/smt_engine_scope.cpp
ajreynol: Because we have to support making SmtScope on constant SmtEngine methods (where `this` is const).  I had played around with several ways of refactoring this, but for now I am copying the same hack as 2 lines below (line 60) where const_cast is already used.

src/theory/smt_engine_subsolver.cpp
ajreynol: Agreed

src/smt/smt_engine.cpp
ajreynol: What exactly are you suggesting to change this to?
We could change to `coreChecker.getOptions().setOption("check-unsat-cores", "false");` I guess?

test/regress/regress2/quantifiers/sygus-inst-ufbv-sdlx-fixpoint-5.smt2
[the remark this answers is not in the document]
4tXJ7f: Hm, it would be good to get a sense of what the difference is. Were we previously using a subsolver with exported `Expr`s?

src/theory/theory_engine.h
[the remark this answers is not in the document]
4tXJ7f: I am not generally a fan of circular dependencies but I think that `SmtEngine` (or some similar object) will be one of the few exceptions in the future. If we want to get rid of all the singletons then we have to have some object that grants you access to all the "global" information.

src/smt/smt_engine_scope.cpp
4tXJ7f: I see. That's really ugly but I suppose there is not much to do and it will be resolved once we get rid of singletons.

src/smt/smt_engine.cpp
4tXJ7f: Sorry, I wrote something wrong above. I meant `coreChecker.getOptions().set(options::checkUnsatCores, false)`.

src/theory/theory_engine.h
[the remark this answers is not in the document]
aniemetz: I agree with @4tXJ7f .

src/smt/smt_engine.h
```diff
@@ -171,8 +171,12 @@ class CVC4_PUBLIC SmtEngine
     SMT_MODE_INTERPOL
   };
 
-  /** Construct an SmtEngine with the given expression manager.  */
-  SmtEngine(ExprManager* em);
+  /**
+   * Construct an SmtEngine with the given expression manager.
+   * If provided, optr is a set of options that should initialize the values
```
aniemetz, line 176:
```suggestion
   * If provided, optr is a pointer to a set of options that should initialize the values
```

src/api/cvc4cpp.cpp
```diff
@@ -5188,6 +5187,12 @@ ExprManager* Solver::getExprManager(void) const { return d_exprMgr.get(); }
  */
 SmtEngine* Solver::getSmtEngine(void) const { return d_smtEngine.get(); }
 
+/**
+ * !!! This is only temporarily available until the parser is fully migrated to
+ * the new API. !!!
+ */
+Options& Solver::getOptions(void) { return d_smtEngine->getOptions(); }
```
aniemetz, line 5194: What is the parser going to use in the future?

ajreynol: It's not clear. 

I think there is a fundamental strange issue that the parser is designed to be a user of the API, but the options need to be used both internally and by the parser.

If we think that options (from the SmtEngine) are needed permanently, I can update this comment.

src/smt/smt_engine.cpp
ajreynol: Updated this.

src/preprocessing/passes/sygus_inference.cpp
```diff
@@ -300,8 +300,9 @@ bool SygusInference::solveSygus(std::vector<Node>& assertions,
   Trace("sygus-infer") << "*** Return sygus inference : " << body << std::endl;
 
   // make a separate smt call
-  SmtEngine rrSygus(nm->toExprManager());
-  rrSygus.setLogic(smt::currentSmtEngine()->getLogicInfo());
+  SmtEngine* currSmt = smt::currentSmtEngine();
```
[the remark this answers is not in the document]
ajreynol: Will do this on future PR.

test/regress/regress2/quantifiers/sygus-inst-ufbv-sdlx-fixpoint-5.smt2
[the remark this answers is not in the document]
ajreynol: This benchmark no longer times out with unsat cores, I've fixed the regression.

src/api/cvc4cpp.cpp
aniemetz: Resolved in offline discussion.

test/regress/regress2/quantifiers/sygus-inst-ufbv-sdlx-fixpoint-5.smt2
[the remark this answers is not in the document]
aniemetz: Resolved, not an issue anymore.

src/theory/theory_engine.h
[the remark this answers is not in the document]
aniemetz: To be addressed in a later PR.

PR #4684, opened 2020-07-01
Add deps/install/lib to RPATH and warn user when using dynamic libs.

nafur:
Installing the cvc4 binary does not work right now if it links against a shared library obtained via one of the contrib scripts.
This PR thus adds `deps/install/lib` to the RPATH so that the installed binary works at all in this case.

This change however paves the way to more problems: If one install such a dynamically linked binary and then removes (or updates) one the shared libraries in `deps/install/lib`, the installed binary most probably stops working.
Hence, this PR checks whether this may happen (whether we link dynamically and link against a shared library from `deps/install/lib`) and, if this is the case, informs and warns the user about this issue.
If the user tries to install to the default install prefix (`/usr/local`) we disallow installation entirely.


CMakeLists.txt
```diff
@@ -557,6 +564,41 @@ endif()
...
+  # Check if we actually use such "dangerous" libraries
+  if(list_len GREATER 0)
+    # Print a generic warning
+    install(CODE "message(WARNING \"You are installing a dynamically linked binary of CVC4 which may be a problem if you are using any dynamically linked third-party library that you obtained through one of the contrib/get-xxx scripts. The binary uses the rpath mechanism to find these locally, hence executing such a contrib script removing the \\\"deps/install\\\" folder most probably breaks the installed binary! Consider installing the dynamically linked dependencies on your system manually or link cvc4 statically.\")")
+    # Print the libraries in question
+    foreach(lib ${LIBS_SHARED_FROM_DEPS})
+      install(CODE "message(WARNING \"The following library is used by the cvc4 binary: ${lib}\")")
+    endforeach()
+    # Check if we use a custom installation prefix
+    if(CMAKE_INSTALL_PREFIX EQUAL "/usr/local")
```
mpreiner, line 594: Use `if(CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT)` instead of checking `/usr/local`.

 https://cmake.org/cmake/help/latest/variable/CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT.html#variable:CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT

CMakeLists.txt
```diff
@@ -557,6 +564,41 @@ endif()
...
+    # Filter out those that are linked dynamically and come from deps/install
+    if(lib MATCHES ".*/deps/install/lib/.*\.so")
+      list(APPEND LIBS_SHARED_FROM_DEPS ${lib})
+    endif()
+  endforeach()
+  list(LENGTH LIBS_SHARED_FROM_DEPS list_len)
+  # Check if we actually use such "dangerous" libraries
+  if(list_len GREATER 0)
+    # Print a generic warning
+    install(CODE "message(WARNING \"You are installing a dynamically linked binary of CVC4 which may be a problem if you are using any dynamically linked third-party library that you obtained through one of the contrib/get-xxx scripts. The binary uses the rpath mechanism to find these locally, hence executing such a contrib script removing the \\\"deps/install\\\" folder most probably breaks the installed binary! Consider installing the dynamically linked dependencies on your system manually or link cvc4 statically.\")")
```
mpreiner, line 588: Split string into multiple lines (80 chars)?

CMakeLists.txt
```diff
@@ -557,6 +564,41 @@ endif()
...
+    # Print the libraries in question
+    foreach(lib ${LIBS_SHARED_FROM_DEPS})
+      install(CODE "message(WARNING \"The following library is used by the cvc4 binary: ${lib}\")")
+    endforeach()
+    # Check if we use a custom installation prefix
+    if(CMAKE_INSTALL_PREFIX EQUAL "/usr/local")
+      install(CODE "message(FATAL_ERROR \"To avoid installing a soon-to-be-broken binary, system-wide installation is disabled if the binary depends on locally-built shared libraries.\")")
```
mpreiner, line 595: Ditto.

CMakeLists.txt
```diff
@@ -557,6 +564,41 @@ endif()
...
+    else()
+      install(CODE "message(WARNING \"You have selected a custom install directory ${CMAKE_INSTALL_PREFIX}, so we expect you understood the previous warning and know what you are doing.\")")
```
mpreiner, line 597: Ditto.

CMakeLists.txt
nafur: This is an interesting issue I already pondered when implementing this. The issue is, that we want to check something related, but different to what `CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT` tells us.
`CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT` is true if there was no `-DCMAKE_INSTALL_PREFIX=...` passed to cmake. However, if it was passed, this does not mean everything is fine. A user (or another script) could pass `/usr/local` explicitly. A user could change the build type with `ccmake`. In both these cases, simply checking `CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT` is not sufficient, but the current check is fine.

I guess I don't see good solutions here. For example, we would not warn users installing to `/usr` or `/opt` or alike. But I still prefer the current one over `CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT` as it warns slightly more users that are at risk.

CMakeLists.txt
nafur: Done.

CMakeLists.txt
nafur: Done.

CMakeLists.txt
nafur: Done.

PR #4689, opened 2020-07-02
Improve and fix secant and tangent lemmas in the transcendental solver

### LINKED ISSUE #4686: Fatal failure at src/theory/arith/nl/ transcendental_solver.cpp:1014
```
% cvc4 small.smt2
Fatal failure within bool CVC4::theory::arith::nl::TranscendentalSolver::checkTfTangentPlanesFun(CVC4::Node, 
unsigned int, std::vector<CVC4::theory::arith::nl::NlLemma>&) at src/theory/arith/nl/
transcendental_solver.cpp:1014
Check failure
 d_model.computeAbstractModelValue(lem) == d_false
Aborted
[515] % 
[515] % cat small.smt2
(set-logic QF_NRAT)
(declare-fun a () Real)
(declare-fun b () Real)
(declare-fun c () Real)
(assert (distinct a 1 (- (exp (* (- a) b)) c)))
(check-sat)
[516] %
```
OS: Ubuntu 18.04
Commit:  5266e8e

ajreynol:
Fixes #4686. 

This benchmark failed an assertion that corresponded to the fact that a refinement lemma did not evaluate to false in the current model (and hence does not rule out the current model).

This was caused by applying the rewriter in a way that led to an incorrect approximation. This meant that some tangent and secant lemmas would be ineffective.  

The benchmark in that issue now times out, but makes progress in the refinement lemmas it generates.

This also simplifies and improves the use of approximated values (instead of model values) when constructing tangent and secant lemmas.


src/theory/arith/nl/transcendental_solver.cpp
```diff
@@ -1450,8 +1449,14 @@ std::pair<Node, Node> TranscendentalSolver::getTfModelBounds(Node tf,
...
+      // { x -> M(tf[0]) }
+      // Notice that we compute the model value of tfs first, so that
+      // the call to rewrite below does not modify the term, where notice that
+      // M_A( x*x { x -> M_A(t) } ) = M_A(t)*M_A(t)
+      // is not equal to
+      // M_A( x*x { x -> t } ) = M_A( t*t )
+      Node mtfs = d_model.computeAbstractModelValue(tfs);
+      pab = pab.substitute(tfv, mtfs);
       pab = Rewriter::rewrite(pab);
       Node v_pab = d_model.computeAbstractModelValue(pab);
```
ahmed-irfan, line 1461: why do we need to get model value for pab? isn't pab already without variables (after subsititution and rewriting)?

src/theory/arith/nl/transcendental_solver.cpp
```diff
@@ -776,6 +776,13 @@ bool TranscendentalSolver::checkTfTangentPlanesFun(Node tf,
   bool is_tangent = false;
   bool is_secant = false;
   std::pair<Node, Node> mvb = getTfModelBounds(tf, d);
+  // this is the approximated value of tf(c), which is a value such that:
+  //    M_A(tf(c)) >= poly_appox_c >= tf(c) or
```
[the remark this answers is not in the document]
ajreynol: The abstract model.

src/theory/arith/nl/transcendental_solver.cpp
ajreynol: Yes, I believe you're right.  I'll remove this line.

src/theory/arith/nl/transcendental_solver.cpp
```diff
@@ -1450,11 +1449,18 @@ std::pair<Node, Node> TranscendentalSolver::getTfModelBounds(Node tf,
     Node pab = pbounds[index];
     if (!pab.isNull())
     {
-      // { x -> tf[0] }
-      pab = pab.substitute(tfv, tfs);
+      // { x -> M(tf[0]) }
```
ahmed-irfan, line 1452: i guess `M` should be `M_A`

PR #4696, opened 2020-07-05
Always Update Git information when rebuilding

4tXJ7f:
Commit 61734b41b7b96e7e7cbf46021a357d840d64b42e changed the way some of
our source files are generated. However, the change meant that once
`git_versioninfo.cpp` was generated, it was never updated again: The
custom command for `git_versioninfo.cpp` has no dependencies, so CMake
does not rebuild it unless the output file is missing [0]. This commit
reverts the change to our `gen-gitinfo` target. I am not sure if there
is a better solution because we actually have to run `GitInfo.cmake` to
see if there have been any changes in the Git information. Introducing a
dependency on all source files is not sufficient because other files or
just the branch name may change. Note: The original solution only
updates the timestamp of `git_versioninfo.cpp` if its contents actually
change (`GitInfo.cmake` uses `configure_file()` to generate
`git_versioninfo.cpp`, which only updates the timestamp when the
contents changed [1]), so we don't do any unnecessary work.

[0] https://cmake.org/cmake/help/latest/command/add_custom_command.html
[1] https://cmake.org/cmake/help/latest/command/configure_file.html

cc @andrewvaughanj


andrewvaughanj:
Hmmm, yeah, sorry 😞 

My only thought would be this (for now):

- always configure (`git_versioninfo.cpp`), but call it something else (`git_versioninfo.cpp.curr`)

- if a previous file exists, [compare](https://stackoverflow.com/a/54354122) the new file against the previous one (i.e., compare `git_versioninfo.cpp` against `git_versioninfo.cpp.curr`)

- If `git_versioninfo.cpp` doesn't exist *or* `git_versioninfo.cpp.curr` has different content, overwrite `git_versioninfo.cpp` 

- Let `cmake` do the rest (i.e., if the `git` info has changed, `git_versioninfo.cpp` will get overwritten and then binaries will get rebuilt; if the git info hasn't changed, `git_versioninfo.cpp` won't be touched, and no work will be done)

I thought about adding a dependency on `.git/HEAD` or `.git/packed-refs`, but I use [worktrees](https://git-scm.com/docs/git-worktree) for my CVC4 branches, so wouldn't work.

This solution doesn't give `no-op` builds, but it *does* ensure the compiler only runs when the `git` info has changed.

I can implement this if you think the idea is reasonable?

andrewvaughanj:
See: https://github.com/4tXJ7f/CVC4/pull/21

andrewvaughanj:
I'm very sorry for the noise here, but neither this PR or my PR work.

There is some sequence issue in how CMake detects these dependencies. I'm looking at this.

Do this:

- build this branch

- checkout a "fresh" branch (so the `git` info changes)

- do a build

The build will _only_ regenerate `./src/base/git_versioninfo.cpp` but it _does not_ cause CMake to actually do a build.

If you do a _second_ build, only *then* is the updated `git_versioninfo.cpp` included (same behaviour in both PRs).

andrewvaughanj:
Should be fixed in https://github.com/4tXJ7f/CVC4/pull/21/commits/9fdd973fb252c63710e9fc228efaaafe44798945 (same PR as above)

4tXJ7f:
Awesome, thank you very much for looking into this!

> Hmmm, yeah, sorry 😞 

No worries, your contributions are greatly appreciated and we all make mistakes :)

> This solution doesn't give `no-op` builds, but it *does* ensure the compiler only runs when the `git` info has changed.

So this part is actually guaranteed by `configure_file()` because it does not update the file if the contents do not change.

> The build will _only_ regenerate `./src/base/git_versioninfo.cpp` but it _does not_ cause CMake to actually do a build.

I can confirm this issue with the original code and I can also confirm that your fix works. I have merged your proposal into this PR. Thanks again for looking into this.

4tXJ7f:
@mpreiner I am marking this PR as "do not merge" because I want to ensure that @andrewvaughanj is added as a co-author of the commit.

andrewvaughanj:
### Discussion on current changes

So I was surprised this works, but it does appear to:

```
# let's not change the branch
[avj@tempvm build]$ stat ./src/base/git_versioninfo.cpp |& grep "Modify" && make |& cat > /dev/null && stat ./src/base/git_versioninfo.cpp |& grep "Modify"
Modify: 2020-07-07 08:11:06.259479311 +0100
Modify: 2020-07-07 08:11:06.259479311 +0100

# let's change the branch
[avj@tempvm build]$ stat ./src/base/git_versioninfo.cpp |& grep "Modify" && git checkout -b another_branch && make |& cat > /dev/null && stat ./src/base/git_versioninfo.cpp |& grep "Modify"
Modify: 2020-07-07 08:11:06.259479311 +0100
Switched to a new branch 'another_branch'
Modify: 2020-07-07 08:14:41.754127299 +0100

# let's not change the branch
[avj@tempvm build]$ stat ./src/base/git_versioninfo.cpp |& grep "Modify" && make |& cat > /dev/null && stat ./src/base/git_versioninfo.cpp |& grep "Modify"
Modify: 2020-07-07 08:14:41.754127299 +0100
Modify: 2020-07-07 08:14:41.754127299 +0100
```

which seems to be related to this: https://gitlab.kitware.com/cmake/cmake/-/issues/16856

That is:

> Generated files are modified and their timestamp updated on subsequent cmake runs only if their content is changed

Based on this, it seems like we don't need my silly "file compare" magic 🎉 

### Discussion on `BYPRODUCTS`

You'll notice I changed to GNU Make for this; this was to ensure that `BYPRODUCTS` was doing what we want when using `make`. 

What version of `cmake` are you on @4tXJ7f? The documentation for `BYPRODUCTS` changed between:

- [3.2](https://cmake.org/cmake/help/v3.2/command/add_custom_target.html), which is the minimum version of CMake that CVC4 supports; and

- [3.16](https://cmake.org/cmake/help/v3.16/command/add_custom_target.html)

(it also changed in [3.17](https://cmake.org/cmake/help/v3.17/command/add_custom_target.html), but ignore this change)

Just for clarity, for 3.16, this line was ***removed***:

> The BYPRODUCTS option is ignored on non-Ninja generators except to mark byproducts GENERATED.

I can try an older CMake, but reading that, I do not know if (when on older CMake versions) `BYPRODUCTS` is going to work correctly. What we've seen is that *without* the `BYPRODUCTS` line, CMake doesn't know to rebuild `libcvc4.so`; my concern is that if that line is ignored for CMake < 3.16 *and* you're not using Ninja, then it is possible `libcvc4.so` won't be built.

If it is ignored, then even doing a pull of master could be confusing: CMake will re-generate `git_versioninfo.cpp`, compile all of the new files (but *not* compile `git_versioninfo.cpp`, because it doesn't know the unconditional call to `GitInfo.cmake` will change the file), and re-link `libcvc4.so`. This will have those new changes in but not the updated hash.

I'll build 3.2 to see what happens.

andrewvaughanj:
I'm happy this works with GNU Make + CMake 3.4 (see #4702 for why I used 3.4 and not 3.2):

```
[avj@tempvm build]$ cmake --version
cmake version 3.4.0

CMake suite maintained and supported by Kitware (kitware.com/cmake).
[avj@tempvm build]$ make --version
GNU Make 4.2.1
Built for x86_64-redhat-linux-gnu
Copyright (C) 1988-2016 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
[avj@tempvm build]$ make -j12 |& egrep "(git_versioninfo.cpp.o|libcvc4.so)"
[avj@tempvm build]$ make -j12 |& egrep "(git_versioninfo.cpp.o|libcvc4.so)"
[avj@tempvm build]$ git checkout -b $(($(date +%s%N)/1000000))
Switched to a new branch '1594109898388'
[avj@tempvm build]$ make -j12 |& egrep "(git_versioninfo.cpp.o|libcvc4.so)"
[  4%] Building CXX object src/CMakeFiles/cvc4.dir/base/git_versioninfo.cpp.o
[  4%] Linking CXX shared library libcvc4.so
[avj@tempvm build]$ make -j12 |& egrep "(git_versioninfo.cpp.o|libcvc4.so)"
[avj@tempvm build]$
```

4tXJ7f:
Sorry, I thought that I had already replied to this. Thanks a lot for trying it! I am ok with this solution. @mpreiner, what do you think?

PR #4704, opened 2020-07-07
Add getName() method to options.

mpreiner:
getName() returns the long option name if it exists and an empty string otherwise.


src/options/mkoptions.py
```diff
@@ -228,6 +229,14 @@
   return Options::current()->wasSetByUser(*this);
 }}"""
 
+TPL_IMPL_GET_NAME = \
+"""inline const std::string {name}__option_t::getName() const
```
4tXJ7f, line 233:
```suggestion
"""const std::string {name}__option_t::getName() const
```

mpreiner: `inline` is needed here since it's implemented in the header (outside of the struct)... We should probably change that when we refactor options.

4tXJ7f: Ah right, makes sense.

PR #4705, opened 2020-07-07
Fix component contains for splicing due to substring.

### LINKED ISSUE #4701: Fatal failure at src/theory/strings/strings_entail.cpp:311
Hi,
For this formula:
```
(set-logic QF_S)
(declare-fun a () String)
(declare-fun b () String)
(declare-fun c () String)
(declare-fun d () String)
(declare-fun e () String)
(declare-fun f () String)
(declare-fun g () String)
(declare-fun h () String)
(declare-fun i () String)
(declare-fun j () String)
(declare-fun k () String)
(declare-fun l () String)
(assert (= f j "abcdefg"))
(assert (= e (str.++ (str.substr b 0 (str.len k)) (str.substr a 0 (str.len d)))))
(assert (= l (str.++ (str.substr c 0 (str.len e)) f)))
(assert (= (str.substr b 0 (str.len k)) "aa"))
(assert (= (str.replace (str.replace c e (str.at c 0)) "a" (str.at c 0)) (str.++ i j)))
(assert (= i (str.++ h g) "aaaa"))
(assert (str.in_re l (re.++ (str.to_re "abcdefgh") (re.* re.allchar))))
(check-sat)
```
CVC4 --strings-exp  throws out a fatal failure:
```
Fatal failure within int CVC4::theory::strings::StringsEntail::componentContains(std::vector<CVC4::NodeTemplate<true> >&, std::vector<CVC4::NodeTemplate<true> >&, std::vector<CVC4::NodeTemplate<true> >&, std::vector<CVC4::NodeTemplate<true> >&, bool, int) at src/theory/strings/strings_entail.cpp:311
Check failure
 n1rb_last.isNull()
Aborted
```

OS: Ubuntu 18.04
Commit: 9678f58

ajreynol:
Fixes #4701.  That benchmark now times out.


nafur:
> LGTM, thanks for the fix. It would be great to add a unit test for this, though.

How about this regression?
```
(declare-fun a () String)
(declare-fun b () String)
(declare-fun c () String)
(declare-fun e () String)
(assert (= e (str.++ b (str.substr a 0 1))))
(assert (= a (str.substr c 0 (str.len e))))
(assert (= "a" b))
(assert (= (str.++ b a) (str.replace c e a)))
(check-sat)
```

ajreynol:
I would favor a regression instead of a unit test for this, since the interface/contract for the method in question is potentially subject to change.   I've added the one posted by @nafur above (thanks).

PR #4709, opened 2020-07-08
(proof-new) Add ONCE and FIXPOINT modes for term conversion proof generator

ajreynol:
We need multiple policies for generic proofs for term conversion, in particular, substitution has a "apply once" semantics which does not apply rewrite steps to a fixpoint like the rewriter.

This is required for eliminating SUBS steps in the final proof.

Also note that an internal method for getting proofs was generalized, which will be required for doing multiple variants of proofs in this utility in the future.


src/expr/term_conversion_proof_generator.cpp
```diff
@@ -18,27 +18,57 @@ using namespace CVC4::kind;
 
 namespace CVC4 {
 
+std::ostream& operator<<(std::ostream& out, TConvPolicy tcpol)
+{
+  switch (tcpol)
+  {
+    case TConvPolicy::FIXPOINT: out << "FIXPOINT"; break;
+    case TConvPolicy::ONCE: out << "ONCE"; break;
+    default: out << "TConvPolicy:unknown"; break;
```
HanielB, line 27:
```suggestion
    default: out << "TConvPolicy:unknown"; 
```

src/expr/term_conversion_proof_generator.cpp
```diff
@@ -18,27 +18,57 @@ using namespace CVC4::kind;
...
+  }
+  // should not rewrite term to two different things
+  if (hasRewriteStep(t))
+  {
+    Assert(getRewriteStep(t) == s);
+    return;
+  }
   Node eq = t.eqNode(s);
   d_proof.addLazyStep(eq, pg);
   d_rewriteMap[t] = s;
```
HanielB, line 56: You could encapsulate all but the step addition to `d_proof` in a `addRewriteStep` method that returns a boolean and is called by all three versions of `addRewriteStep`.

src/expr/term_conversion_proof_generator.cpp
```diff
@@ -72,29 +111,23 @@ std::shared_ptr<ProofNode> TConvProofGenerator::getProofFor(Node f)
...
     Trace("tconv-pf-gen") << "...failed, mismatch: returned proof concludes "
-                          << pf->getResult() << std::endl;
+                          << conc << ", expected " << f << std::endl;
     Assert(false);
     return nullptr;
   }
   Trace("tconv-pf-gen") << "... success" << std::endl;
-  return pf;
+  return lpf.getProofFor(f);
+  ;
```
HanielB, line 126:
```suggestion
```

src/expr/term_conversion_proof_generator.cpp
```diff
@@ -121,13 +154,21 @@ std::shared_ptr<ProofNode> TConvProofGenerator::getProofForRewriting(Node t)
       if (!rcur.isNull())
       {
         // d_proof should have a proof of cur = rcur. Hence there is nothing
```
HanielB, line 156: I'd replace "should have" by "must have" or "has"

src/expr/term_conversion_proof_generator.cpp
```diff
@@ -121,13 +154,21 @@ std::shared_ptr<ProofNode> TConvProofGenerator::getProofForRewriting(Node t)
...
+        if (d_policy == TConvPolicy::FIXPOINT)
+        {
+          // It may be the case that rcur also rewrites, thus we cannot assign
+          // the final rewritten form for cur yet. Instead we revisit cur after
+          // finishing visiting rcur.
+          rewritten[cur] = rcur;
+          visit.push_back(cur);
+          visit.push_back(rcur);
+        }
+        else
```
HanielB, line 167: Just for safety, check that policy is `ONCE`, in case we add more in the future and to guarantee this is not forgotten.

src/expr/term_conversion_proof_generator.cpp
ajreynol: This would be a compile error?

HanielB: That's strange. The `default` case does not need a break.

src/expr/term_conversion_proof_generator.cpp
```diff
@@ -101,6 +68,21 @@ bool TConvProofGenerator::hasRewriteStep(Node t) const
...
+  {
+    return Node::null();
+  }
+  // should not rewrite term to two different things
+  if (!getRewriteStep(t).isNull())
+  {
+    Assert(getRewriteStep(t) == s);
+    return Node::null();
+  }
+  return t.eqNode(s);
```
HanielB, line 83:
```suggestion
  d_rewriteMap[t] = s;
  return t.eqNode(s);
```

src/expr/term_conversion_proof_generator.cpp
```diff
@@ -40,36 +40,14 @@ TConvProofGenerator::~TConvProofGenerator() {}
...
-    return;
-  }
-  // should not rewrite term to two different things
-  if (hasRewriteStep(t))
-  {
-    Assert(getRewriteStep(t) == s);
-    return;
-  }
-  Node eq = t.eqNode(s);
+  Node eq = registerRewriteStep(t,s);
```
HanielB: If you have `registerRewriteStep(t,s);` return a null node and not check that will screw up the lines below, no?

ajreynol: ah yeah, whoops.

PR #4711, opened 2020-07-09
Remove cumulative time limits and cpu time limits

nafur:
This PR removes two things from the resource manager:
- cumulative time limits
- cpu time limits

Cumulative time limiting has been moved to the binary and is (as before) accessible via `--tlimit`. As per discussion among the devs, we no longer support time limits based on CPU time and thus everything related to that is removed as well.
Note that this includes the option `--cpu-time`, removes an argument from `SmtEngine::setTimeLimit()` and the method `SmtEngine::getTimeRemaining()` .


nafur:
As for formatting, I'd kindly ask to push this to #4732 which also touches these files.

aniemetz:
As clarified offline, clang-formatting is a requirement for merging back to master.

PR #4712, opened 2020-07-09
Associate all lemmas in non-linear arithmetic with an inference identifier

ajreynol:
This marks all lemmas in non-linear arithmetic with an identifier, which indicates informally the kind of justification that was used for them.  The main motivation for this is for debugging the behavior of the non-linear solver.

The number of inferences can then be seen with `--stats`:
```
nl::inferences, [(SPLIT_ZERO : 19), (SIGN : 4), (COMPARISON : 29)]
```

The same design was used in strings and has been quite helpful.

This also adds a few high level stats to the new statistics class for non-linear.

### DIFF
--- a/src/theory/arith/nl/nl_solver.cpp
+++ b/src/theory/arith/nl/nl_solver.cpp
@@ -177,8 +177,7 @@ std::vector<NlLemma> NlSolver::checkSplitZero()
       eq = Rewriter::rewrite(eq);
       Node literal = d_containing.getValuation().ensureLiteral(eq);
       d_containing.getOutputChannel().requirePhase(literal, true);
-      lemmas.push_back(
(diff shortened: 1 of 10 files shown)


src/theory/arith/nl/nl_solver.cpp
```diff
@@ -177,7 +177,8 @@ std::vector<NlLemma> NlSolver::checkSplitZero()
       eq = Rewriter::rewrite(eq);
       Node literal = d_containing.getValuation().ensureLiteral(eq);
       d_containing.getOutputChannel().requirePhase(literal, true);
-      lemmas.push_back(literal.orNode(literal.negate()));
+      lemmas.push_back(
+          NlLemma(literal.orNode(literal.negate()), Inference::SPLIT_ZERO));
```
nafur, line 181: Also use `emplace_back()` here.

PR #4716, opened 2020-07-10
Implement --tlimit for windows

nafur:
The new mechanism for `--tlimit` only works for POSIX compatible systems (that implement `setitimer`). This PR implements an alternative (though roughly equivalent) approach for windows, based on `SetWaitableTimer()`.

To make this work (without code duplication) we need to call the `timeout_handler` function from `time_limit.cpp` as the windows solution employs an asynchronous callback instead of signals. I used the opportunity to rename `util.cpp` to `signal_handlers.cpp` (as it really does not do anything else) and reformat the file.

As I do not have a windows system at hand, I have not been able to actually test this apart from making sure that it compiles with the mingw setup.


src/main/signal_handlers.h
```diff
@@ -0,0 +1,17 @@
+#ifndef CVC4__MAIN__SIGNAL_HANDLERS_H
```
4tXJ7f, line 1: Add our boilerplate comment.

src/main/time_limit.cpp
```diff
@@ -1,18 +1,51 @@
 #include "time_limit.h"
```
4tXJ7f, line 1: Add our boilerplate comment.

src/main/signal_handlers.h
```diff
@@ -0,0 +1,17 @@
...
+#define CVC4__MAIN__SIGNAL_HANDLERS_H
+
+namespace CVC4 {
+namespace main {
+namespace signal_handlers {
+
+void timeout_handler();
+
+void install();
+void cleanup();
```
4tXJ7f, line 11: Add comments documenting the methods.

src/main/signal_handlers.cpp
```diff
@@ -1,5 +1,5 @@
 /*********************                                                        */
-/*! \file util.cpp
+/*! \file signal_handlers.cpp
```
4tXJ7f, line 2: Very much in favor of this name change :)

src/main/signal_handlers.cpp
```diff
@@ -54,15 +55,27 @@ namespace main {
...
     pExecutor->safeFlushStatistics(STDERR_FILENO);
   }
 }
 
+void timeout_handler()
+{
+  safe_print(STDERR_FILENO, "CVC4 interrupted by timeout.\n");
+  print_statistics();
+  abort();
+}
```
4tXJ7f, line 77: It feels like it would be more consistent to move this function to src/main/time_limit.cpp (similar to `win_timeout_handler()`).

src/main/signal_handlers.cpp
```diff
@@ -71,11 +84,7 @@ void* cvc4StackBase;
 #endif /* HAVE_SIGALTSTACK */
 
 /** Handler for SIGXCPU and SIGALRM, i.e., timeout. */
-void timeout_handler(int sig, siginfo_t* info, void*) {
-  safe_print(STDERR_FILENO, "CVC4 interrupted by timeout.\n");
-  print_statistics();
-  abort();
-}
+void timeout_handler(int sig, siginfo_t* info, void*) { timeout_handler(); }
```
4tXJ7f, line 87: It feels like it would be more consistent to move this function to src/main/time_limit.cpp (similar to `win_timeout_handler()`).

src/main/signal_handlers.cpp
```diff
@@ -286,43 +324,49 @@ void cvc4_init()
...
 
   struct sigaction act2b;
   act2b.sa_sigaction = timeout_handler;
   act2b.sa_flags = SA_SIGINFO;
   sigemptyset(&act2b.sa_mask);
-  if(sigaction(SIGALRM, &act2b, NULL)) {
+  if (sigaction(SIGALRM, &act2b, NULL))
+  {
     throw Exception(string("sigaction(SIGALRM) failure: ") + strerror(errno));
   }
```
4tXJ7f, line 348: I'd suggest to move the installation of SIGXCPU/SIGALRM to `install_time_limit()` to mirror the Windows side.

src/main/time_limit.cpp
```diff
@@ -1,27 +1,47 @@
...
 #if defined(__MINGW32__) || defined(__MINGW64__)
 void CALLBACK win_timeout_handler(LPVOID lpArg,
                                   DWORD dwTimerLowValue,
                                   DWORD dwTimerHighValue)
-
 {
   signal_handlers::timeout_handler();
 }
+#else
+void posix_timeout_handler(int sig, siginfo_t* info, void*) {
```
4tXJ7f, line 42: Apply ClangFormat.

4tXJ7f:
I have sanity checked this on Windows (ran https://github.com/CVC4/CVC4/blob/df642ec7d4eef0e2f994751be53e66201f2b92f9/test/regress/regress3/strings/extf_d_perf.smt2 with different `--tlimits`) and it seems to work :)

PR #4719, opened 2020-07-10
Statistics on instantiations per quantified formula.

ajreynol:
This adds a new print format for instantiations --print-instantiations=num, which prints the total number of instantations of quantified formulas.  This count is user-context-dependent, which is in sync with the existing print-instantiation format (list).

It also simplifies and improves printing of Instantiation Tries.


src/theory/quantifiers/inst_match_trie.cpp
```diff
@@ -392,7 +386,6 @@ bool CDInstMatchTrie::recordInstLemma(Node q,
 void CDInstMatchTrie::print(std::ostream& out,
                             Node q,
                             std::vector<TNode>& terms,
-                            bool& firstTime,
```
HanielB, line 395: Comment on the PR description why you are removing this.

src/theory/quantifiers/instantiate.cpp
```diff
@@ -264,9 +264,8 @@ bool Instantiate::addInstantiation(
     return false;
   }
 
-  d_total_inst_debug[q]++;
+  d_total_inst_debug[q] = d_total_inst_debug[q] + 1;
```
HanielB, line 267: Why this change?

src/theory/quantifiers/quantifiers_attributes.cpp
```diff
@@ -50,8 +50,9 @@ void QuantAttributes::setUserAttribute( const std::string& attr, Node n, std::ve
     SygusAttribute ca;
     n.setAttribute( ca, true );
   }
-  else if (attr == "quant-name")
+  else if (attr == "qid")
   {
+    // using z3 syntax "qid"
```
HanielB, line 55: I was going to ask why use this rather than `:named` but I think this comment answers the question. :)

src/options/quantifiers_options.toml
```diff
@@ -1962,6 +1962,30 @@ header = "options/quantifiers_options.h"
...
+  type       = "PrintInstMode"
+  default    = "LIST"
+  help       = "print format for printing instantiations"
+  help_mode  = "Print format for printing instantiations."
+[[option.mode.LIST]]
+  name = "list"
+  help = "Print the list of instantiations per quantified formula."
+[[option.mode.NUM]]
+  name = "num"
+  help = "Print the total number of instantiations per quantified formula."
```
HanielB, line 1980: Maybe helpful to say that only prints if more than 0 instantiations.

src/theory/quantifiers/instantiate.cpp
ajreynol: because its context-dependent now, which doesnt support `++`.

PR #4723, opened 2020-07-11
Supporting seq.nth

yoni206:
This PR adds support for `seq.nth` operator by eliminating it during `expandDefinitions`, based on sub-sequences.
Tests that use this operator are also included.


ajreynol:
Can you add this example?
```
(declare-fun a () (Seq Int))
(declare-fun b () (Seq Int))
(declare-fun c () (Seq Int))
(declare-fun x () Int)
(declare-fun y () Int)
(assert (or (= a b) (= a c)))
(assert (not (= (seq.nth a x) (seq.nth b x))))
(assert (not (= (seq.nth a y) (seq.nth c y))))
(check-sat)
```

ajreynol:
Your example with "sat" should be able to pass `--check-models`.  This will require adding a basic rewriter for seq.nth in src/theory/strings/sequences_rewriter

src/theory/strings/theory_strings.cpp
```diff
@@ -604,6 +604,28 @@ TrustNode TheoryStrings::expandDefinition(Node node)
     return TrustNode::mkTrustRewrite(node, ret, nullptr);
   }
 
+  if (node.getKind() == SEQ_NTH)
+  {
+    // str.nth(s,i) --->
+    //   witness k. ite(0 <= i < len(s), unit(k) = seq.at(s,i), true)
```
4tXJ7f, line 610: Isn't this just:
```
witness k. 0 <= i < len(s) => unit(k) = seq.at(s,i)
```?

Comment (4tXJ7f) [reply src/theory/strings/theory_strings.cpp:610]: Actually, I think that this needs to be:
```
witness k. ite(0 <= i < len(s), unit(k) = seq.at(s,i), USeq(s,i))
```

where `USeq(s,i)` is an uninterpreted function.

Comment (4tXJ7f) [reply src/theory/strings/theory_strings.cpp:610]: Ah, I guess Andy's test case would catch this as well :)

Comment (ajreynol) [reply src/theory/strings/theory_strings.cpp:610]: Yes, that was the point of my test :)  And yes, we need the uninterpreted function.

**Comment by ajreynol** [comment]
@yoni206 missing a file from your last commit I think?

**Comment by yoni206** [comment]
Actually, not quite ready yet. I will let you know once I am done with correctly addressing all the comments.

File: src/theory/strings/theory_strings.cpp [hunk 2]
```diff
@@ -594,6 +594,39 @@ TrustNode TheoryStrings::expandDefinition(Node node)
     return TrustNode::mkTrustRewrite(node, ret, nullptr);
   }
 
+  if (node.getKind() == SEQ_NTH)
+  {
+    // str.nth(s,i) --->
+    //   witness k. ite(0 <= i < len(s), unit(k) = seq.at(s,i), k = uf(s,i))
```
4tXJ7f, line 600: The comment needs to be updated.

src/theory/strings/skolem_cache.h
```diff
@@ -138,6 +138,8 @@ class SkolemCache
     //   where b is a regular expression, n is the number of occurrences of b
     //   in a, and k(0)=0.
     SK_OCCUR_LEN,
+    // exists a. 0 <= i < len(s) -> unit(a) = seq.at(s,i)
```
4tXJ7f, line 141: Update comment to make clear that `a` is a function (analogous to the comments above).

yoni206: right! updated. Please let me know if any other changes are needed.

PR #4724, opened 2020-07-11
Remove ExprSequence

4tXJ7f:
Now that we can break the old API, we don't need an `ExprSequence`
anymore. This commit removes `ExprSequence` and replaces all of its uses
with `Sequence`. Note that I had to temporarily make `sequence.h` public
because we currently include it in a "public" header because we still
generate the code for `Expr::mkConst<Sequence>()`.


src/expr/sequence.h
```diff
@@ -131,33 +137,31 @@ class Sequence
...
-
  private:
   /**
    * Returns a negative number if *this < y, 0 if *this and y are equal and a
    * positive number if *this > y.
    */
   int cmp(const Sequence& y) const;
   /** The element type of the sequence */
-  TypeNode d_type;
+  std::unique_ptr<TypeNode> d_type;
```
ajreynol, line 162: Why the change to unique ptr here?  Is this for destruction issues?

src/expr/sequence.h
```diff
@@ -131,33 +137,31 @@ class Sequence
...
   /** The data of the sequence */
-  std::vector<Node> d_seq;
+  std::unique_ptr<std::vector<Node>> d_seq;
```
ajreynol, line 164: Ditto.

src/expr/sequence.h
4tXJ7f: No, the issue is that `TypeNode` is only forward declared here (there's a circular dependency). Right now, we don't have an easy way around that. Using an `std::unique_ptr` is what we do for other such classes as well, e.g. https://github.com/CVC4/CVC4/blob/master/src/expr/array_store_all.h#L59-L60.

src/expr/sequence.h
4tXJ7f: Yeah, I have been thinking about this one. I don't think we actually need it. I will try and report back.

src/expr/sequence.h
ajreynol: Right, I figured it was something like that.  Looks good to me.

PR #4725, opened 2020-07-12
Add support for string/sequence update

ajreynol:
This adds basic support for string/sequence updating, which has a semantics that is length preserving.


src/util/string.cpp
```diff
@@ -401,10 +401,32 @@ bool String::hasSuffix(const String& y) const
   return true;
 }
 
+String String::update(std::size_t i, const String& t) const
+{
+  if (i < size())
+  {
+    std::vector<unsigned> vec;
+    vec.insert(vec.begin(), d_str.begin(), d_str.begin() + i);
```
4tXJ7f, line 409:
```suggestion
    std::vector<unsigned> vec(d_str.begin(), d_str.begin() + i);
```

src/expr/sequence.cpp
```diff
@@ -252,6 +252,29 @@ bool Sequence::hasSuffix(const Sequence& y) const
   return true;
 }
 
+Sequence Sequence::update(size_t i, const Sequence& t) const
+{
+  Assert(d_type == t.d_type);
+  if (i < size())
+  {
+    std::vector<Node> vec;
+    vec.insert(vec.begin(), d_seq.begin(), d_seq.begin() + i);
```
4tXJ7f, line 261:
```suggestion
    std::vector<Node> vec(d_seq.begin(), d_seq.begin() + i);
```

src/theory/strings/sequences_rewriter.cpp
```diff
@@ -1800,6 +1806,51 @@ Node SequencesRewriter::rewriteSubstr(Node node)
...
+      {
+        // start beyond the maximum size of strings
+        // thus, it must be beyond the end point of this string
+        return returnRewrite(node, s, Rewrite::UPD_CONST_INDEX_MAX_OOB);
+      }
+      else if (node[1].getConst<Rational>().sgn() < 0)
+      {
+        // start before the beginning of the string
+        Node ret = Word::mkEmptyWord(node.getType());
+        return returnRewrite(node, ret, Rewrite::UPD_CONST_INDEX_NEG);
```
4tXJ7f, line 1833:
```suggestion
        // start before the beginning of the string
        return returnRewrite(node, s, Rewrite::UPD_CONST_INDEX_NEG);
```

ajreynol: This isn't correct, s is not the empty string here.

PR #4727, opened 2020-07-13
Remove sygus print callback

ajreynol:
This removes sygus print callbacks, since they are no longer necessary to support the sygus v1 parser.

This involved generalizing the scope of the datatype utility sygusToBuiltin.  Now, printing "as sygus" simply is accomplished by doing sygusToBuiltin conversion and then calling the printer on the builtin term.

This is required for further work towards eliminating the Expr layer.

FYI @4tXJ7f


src/theory/datatypes/sygus_datatype_utils.cpp
```diff
@@ -343,6 +351,9 @@ Node sygusToBuiltin(Node n, bool isExternal)
     it = visited.find(cur);
     if (it == visited.end())
     {
+      // notice this condition succeeds in roughly 99% of the executions of this
```
HanielB, line 354: Could you give more details on this number? For typical benchmarks, regressions...? Circa 20XX?

src/theory/datatypes/sygus_datatype_utils.cpp
```diff
@@ -359,6 +370,27 @@ Node sygusToBuiltin(Node n, bool isExternal)
...
+          // use the previously constructed variable for it
+          visited[cur] = cur.getAttribute(SygusToBuiltinVarAttribute());
+        }
+        else
+        {
+          std::stringstream ss;
+          ss << cur;
+          const DType& dt = cur.getType().getDType();
+          // make a fresh variable
+          NodeManager * nm = NodeManager::currentNM();
```
HanielB, line 387:
```suggestion
          NodeManager* nm = NodeManager::currentNM();
```
Moreover move this to outside the loop.

src/theory/datatypes/sygus_datatype_utils.cpp
ajreynol: Coverage tests :)  They are helpful.

HanielB: Ah, nice. :)

src/theory/datatypes/sygus_datatype_utils.cpp
ajreynol: Its intentionally here for performance reasons (this is partly the point of the comment mentioned above).

PR #4736, opened 2020-07-13
Fix caching in TheoryEngine::getExplanation()

4tXJ7f:
Commit 64a7db86206aa0993f75446a3e7f77f3c0c023c6 introduced a caching
mechanism in `TheoryEngine::getExplanation()`. However, there seem to be
issues related to the timestamps of the explanations. This commit fixes
the issue by making the timestamp part of the cache. The change ensures
that explanations for a certain node never rely on other explanations
for that node with a later timestamp.

Note: With those changes, `regress0/nl/sin-cos-346-b-chunk-0169.smt2` was
timing out, so I added `--nl-ext-tplanes` to its arguments.


test/regress/CMakeLists.txt
```diff
@@ -977,6 +977,8 @@ set(regress_0_tests
   regress0/strings/issue4376.smt2
   regress0/strings/issue4662-consume-nterm.smt2
   regress0/strings/issue4674-recomp-nf.smt2
+  regress0/strings/issue4735.smt2
```
ajreynol, line 980: These take a little bit more time I think, can you move to regress1?

PR #4739, opened 2020-07-14
Debug instantiations output

ajreynol:
This prints # instantiations per round (during solving) for each quantified formula with `--debug-inst`, giving output like this:
```
(num-instantiations myQuant1 1)
(num-instantiations myQuant2 1)
unsat
```

It also changes the default value of print-inst-full to match the behavior of unsat-cores vs unsat-cores-full (by default, nameless quantifiers are ignored). 

It fixes an issue with qid, where mkVar was accidentally used instead of mkConst, leading to assertion failures in debug.  Marking major since this fixes debug regress1.


src/options/quantifiers_options.toml
```diff
@@ -1992,9 +1992,17 @@ header = "options/quantifiers_options.h"
   category   = "regular"
   long       = "print-inst-full"
   type       = "bool"
-  default    = "true"
+  default    = "false"
   help       = "print instantiations for formulas that do not have given identifiers"
 
+[[option]]
+  name       = "debugInst"
+  category   = "regular"
```
4tXJ7f, line 2000: This should arguably be an `expert` option.

src/theory/quantifiers/instantiate.h
```diff
@@ -202,8 +202,13 @@ class Instantiate : public QuantifiersUtil
   Node getTermForType(TypeNode tn);
   //--------------------------------------end general utilities
 
-  /** debug print, called once per instantiation round. */
-  void debugPrint();
+  /**
+   * Debug print, called once per instantiation round. This prints
+   * instantiations added this round to trace inst-per-quant-round, if
+   * applicable, and prints to out of options if the option debug-inst is
+   * enabled.
```
4tXJ7f, line 209: This sentence doesn't quite parse for me.

src/options/quantifiers_options.toml
ajreynol: This is actually a feature request from a user :)

4tXJ7f: Ok, SG :)

PR #4745, opened 2020-07-14
Add option manager and simpler option listener

ajreynol:
This adds the "OptionManager" class, which will live in SmtEngine.  This is the required infrastructure for implementing all "reactive" options, i.e. those that must take effect immediately. 

This PR does not enable this class yet, it simply adds the definitions.

After this PR, we can connect it to SmtEngine and delete the old options listener infrastructure.


src/options/options_template.cpp
```diff
@@ -224,10 +225,11 @@ void runBoolPredicates(T, std::string option, bool b, options::OptionsHandler* h
 }
 
 
-Options::Options()
+Options::Options(OptionsListener* ol)
     : d_holder(new options::OptionsHolder())
     , d_handler(new options::OptionsHandler(this))
-    , d_beforeSearchListeners()
+    , d_beforeSearchListeners(),
```
nafur, line 231: I'd move the comma to the next line.

src/options/options.h
```diff
@@ -415,6 +419,13 @@ class CVC4_PUBLIC Options {
   void flushOut();
 
  private:
+  /** Pointer to the options listener, if one exists */
+  OptionsListener* d_olisten;
```
nafur, line 423: I'm somewhat confused here. You added a generic `OptionsListener` class.
- Is this supposed to replace all the old listeners that we have right now? (`d_beforeSearchListeners` et al)
- As this is generic, I'd suggest using (the equivalent of) `ListenerCollection` here to allow for multiple listeners. If we know there will only ever by a single one, why not use a `OptionsManager*` directly and avoid this layer of (conceptual) indirection?

src/smt/options_manager.cpp
```diff
@@ -0,0 +1,148 @@
...
+    Warning.getStream() << Command::printsuccess(value);
+    *options::out() << Command::printsuccess(value);
+  }
+  // otherwise, no action is necessary
+}
+
+void OptionsManager::finishInit(LogicInfo& logic, bool isInternalSubsolver)
+{
+  // set up the timeout
+  d_resourceManager->setHardLimit(options::hardLimit());
```
nafur, line 119: This conflicts with #4732 which removes the hard limit.

src/smt/options_manager.cpp
```diff
@@ -0,0 +1,148 @@
...
+  if ((*d_options)[options::cumulativeResourceLimit] != 0)
+  {
+    d_resourceManager->setResourceLimit(options::cumulativeResourceLimit(),
+                                        true);
+  }
+  if ((*d_options)[options::perCallMillisecondLimit] != 0)
+  {
+    d_resourceManager->setTimeLimit(options::perCallMillisecondLimit(), false);
+  }
+  if ((*d_options)[options::cumulativeMillisecondLimit] != 0)
```
nafur, line 133: This conflicts with #4711 which removes cumulative time limits (from the resource manager).

src/smt/options_manager.cpp
```diff
@@ -0,0 +1,148 @@
...
+  {
+    d_resourceManager->setTimeLimit(options::cumulativeMillisecondLimit(),
+                                    true);
+  }
+  if ((*d_options)[options::cpuTime])
```
nafur, line 138: This option is removed in #4711 as well.

src/options/options_template.cpp
ajreynol: Ah sorry, will clang-format.

src/options/options.h
ajreynol: Yes, this replaces all listeners. 

Concerning ListenerCollection, I am also planning to delete this class entirely. I believe that this is overdesigned: if the user of this class wants to register multiple listeners, then this is equivalent to implementing multiple things in their (single) listener.  The only reason for having a listener collection is if multiple independent entities want to listen to the same options object, which is not the case. 

The level of indirection here is to avoid dependencies.  The code for "implementing" the options (setting output, resources) should not live in the options code itself.  This separates the two and allows all code to live in `src/smt/`. Indeed its true that we may only have a single listener, but that was the purpose of this.

ajreynol:
Regarding the conflicts, I am in favor of merging your changes to resource manager first, I can update this PR when that is done.

src/options/options.h
nafur: I see. Note that the `ListenerCollection` is also used in the `ResourceManager`. Similarly, the `SmtEngine` seems to be the only entity registering a listener there right now. We planned to allow registering listeners (for resource out, I think) via the API...
We may thus need (something like) the `ListenerCollection`after all. But I agree that the current implementation seems overdesigned...

ajreynol: Yes, I managed to clean up resource manager in this regard as well. For reference, here's my branch: 

https://github.com/CVC4/CVC4/compare/master...ajreynol:refactorOptListen

I would prefer designing the new listener stuff from scratch if it is needed for the API.  I would still argue that this doesn't have to impact the listener in options.  That is, I would prefer registering multiple user-provided callbacks in the derived OptionsListener to separate the complexity of these things.  This way options just has to make one single callback.

nafur: Okay.

PR #4747, opened 2020-07-15
(proof-new) Adding API for converting EqProof into ProofNode

HanielB:
Also puts EqProof into its own module. Next will come the implementation of the API.

### DIFF
--- a/src/theory/uf/eq_proof.h
+++ b/src/theory/uf/eq_proof.h
@@ -121,7 +121,7 @@ class EqProof
    * There are two cases. In the simplest one the general shape of the EqProof
    * is
    *  (= t1 t2) (= t3 t2) (= (t1 t3) false)
-   *  ------------------------------------- TRANS
+   *  ------------------------------------- EQP::TR
    *             false = true
    *
    * which is expanded into
@@ -136,12 +136,13 @@ class EqProof
    *             false = true
    *
    * by explicitly adding the transitivity step for inferring (= t1 t3) and its
-   * predicate equality.
+   * predicate equality. Note that we differentiate (from now on) the EqProof
+   * rules ids from those of ProofRule by adding the prefix EQP:: to the former.
    *
    * In the other case, the general shape of the EqProof is
    *
    *  (= (= t1 t2) false) (= t1 x1) ... (= xn t3) (= t2 y1) ... (= ym t4)
-   * -------------------------------------------------------------------- TRANS
+   * ------------------------------------------------------------------- EQP::TR
    *         (= (= t4 t3) false)
    *
    * which is converted into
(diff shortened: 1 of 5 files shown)


src/theory/uf/eq_proof.h
```diff
@@ -0,0 +1,338 @@
...
+  /** Reduce the a congruence EqProof into a transitivity matrix
+   *
+   * Given a congruence EqProof of (= (f a0 ... an-1) (f b0 ... bn-1)), reduce
+   * its justification into a matrix
+   *
+   *   [0]   -> p_{0,0} ... p_{m_0,0}
+   *   ...
+   *   [n-1] -> p_{0,n} ... p_{m_n-1,n-1}
+   *
+   * where f has arity n and each p_{0,i} ... p_{m_i, i} contains a transitivity
```
ajreynol, line 240: Not "contains", I think this should say "for each i, .... is a transitivity chain".  I was confused why there were seemingly multiple proofs of the same thing for each range.

src/theory/uf/eq_proof.h
```diff
@@ -0,0 +1,338 @@
...
+   *   [0] -> (= a0 b0)
+   *   [1] -> (= a1 b1)
+   *   [2] -> (= a2 b2)
+   *
+   * The more complex case of congruence proofs has transitivity steps as the
+   * first child of CONG steps. For example
+   *                    P0
+   *  ------- REFL  ----------
+   *  (= f f)       (= a0 c)            P'
+   *  ----------------------- CONG   ---------
```
ajreynol, line 274: Ditto for EqProof::.

src/theory/uf/eq_proof.h
```diff
@@ -0,0 +1,338 @@
...
+   *   (= t1 x1) ... (= xn t3)      (= t2 y1) ... (= ym t4)
+   *  ------------------------ TR  ------------------------ TR
+   *   (= t1 t3)                    (= t2 t4)
+   *  ----------- SYMM             ----------- SYMM
+   *   (= t3 t1)                    (= t4 t2)
+   *  ---------------------------------------- CONG
+   *   (= (= t3 t4) (= t1 t2))                         (= (= t1 t2) false)
+   *  --------------------------------------------------------------------- TR
+   *           (= (= t3 t4) false)
+   *          --------------------- MACRO_SR_PRED_TRANSFORM
```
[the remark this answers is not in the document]
HanielB: Is this what you mean?
```
 (= t1 x1) ... (= xn t3)      (= t2 y1) ... (= ym t4)
------------------------ TR  ------------------------ TR
 (= t1 t3)                    (= t2 t4)
----------- SYMM             ----------- SYMM                             
 (= t3 t1)                    (= t4 t2)                                         
---------------------------------------- CONG                             -------------------PRED_INTRO  --------------- R    
 (= (= t3 t4) (= t1 t2))                         (= (= t1 t2) false)       (= (= t3 t4) (= t4 t3))       (= false false)      
--------------------------------------------------------------------- TR   -------------------------------------------- CONG  
         (= (= t3 t4) false)                                               (= (= (= t3 t4) false) (= (= t4 t3) false)         
        ------------------------------------------------------------------------------------------------------------- EQ_RES 
                                           (= (= t4 t3) false)                      
```
The annoying thing is that we don't have `(= t3 t4)` to just do a SYMM step, so we'd still depend on a macro.

src/theory/uf/eq_proof.h
HanielB: I was trying to say that the chain could be minimized, i.e. not all equalities may be required for justifying the premise. I'll change to what you suggested to avoid confusion.

src/theory/uf/eq_proof.h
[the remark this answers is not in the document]
ajreynol: Ah yeah, its more complicated, you'd need SCOPE/EQ_RESOLVE too, I agree its not worthwhile here.

PR #4751, opened 2020-07-15
Ensure that :named terms are made global in the symbol table.

### LINKED ISSUE #4320: Fatal failure at src/expr/symbol_table.cpp:430
```
[566] % z3 small.smt2
sat
((b true))
[567] % 
[567] % cvc4 -q small.smt2
sat
Fatal failure within CVC4::Expr CVC4::SymbolTable::Implementation::lookup(const string&) const at src/expr/symbol_table.cpp:430
Check failure
 isBound(name)
Aborted
[568] % 
[568] % cat small.smt2
(assert (let ((a (! true :named b))) a))
(check-sat)
(get-value (b))
[569] %
```

OS: Ubuntu 18.04
Commit: f0e6c10

nafur:
In #4320, it was found that we throw an exception when calling `(get-value x)` where `x` is a named term that was named within a scope (for example a let expression).
According to section 4.1.6 of SMTLIB v2.6, named terms should be made available globally, as if they were defined by `define-fun` directly before the command containing the `:named x`.
While some care seemed to be taken, `x` would still vanish from the symbol table. This PR fixes this issue by setting `ExprManager::VAR_FLAG_GLOBAL` on `x`.

While it fixes the example from #4320, I'd really like someone with more experience to check it.
In particular: does this flag do the right thing in combination with `(push)` and `(pop)`? In my understanding of the SMTLIB standard, the following should be possible (allowing to redeclare the same named term within different scopes):
> (push)
> (assert (let ((a (! true :named b))) a))
> (check-sat)
> (pop)
> (push)
> (assert (let ((a (! false :named b))) a))
> (check-sat)
> (pop)

This, however, fails with the current PR as `b` remains in the symbol after `(pop)`, thus giving:

> (error "Parse Error: test.smt2:6.34: Symbol 'b' previously declared as a variable
>
>  (assert (let ((a (! false :named b))) a))
>                                   ^
> ")

I mark this as do not merge until this has been clarified.

Fixed #4320.

### DIFF
--- a/src/parser/smt2/smt2.cpp
+++ b/src/parser/smt2/smt2.cpp
@@ -1362,7 +1362,7 @@ api::Term Smt2::setNamedAttribute(api::Term& expr, const SExpr& sexpr)
   // check that sexpr is a fresh function symbol, and reserve it
   reserveSymbolAtAssertionLevel(name);
   // define it
-  api::Term func = bindVar(name, expr.getSort(), ExprManager::VAR_FLAG_DEFINED || ExprManager::VAR_FLAG_GLOBAL);
+  api::Term func = bindVar(name, expr.getSort(), ExprManager::VAR_FLAG_DEFINED | ExprManager::VAR_FLAG_GLOBAL);
   // remember the last term to have been given a :named attribute
   setLastNamedTerm(expr, name);
   return func;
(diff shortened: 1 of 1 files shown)


nafur:
Okay, this is way beyond my current understanding of how scopes are handled in CVC4...

ajreynol:
This may be impacted by the new design of the symbol manager.  The ExprManager flags like VAR_FLAG_GLOBAL will be migrated to the new symbol manager object.

We can probably close this PR.

ajreynol:
@nafur what do you think? Close this PR?  I'm positive this will be addressed differently in the new infrastructure.

PR #4753, opened 2020-07-15
Make ExtTheory a utility and not a member of Theory

ajreynol:
Previously, we assumed that ExtTheory, the module for doing context-dependent simplification, was one-to-one with Theory.  This design is not necessary.  This makes this class a utility, which can be used as needed.  This makes e.g. the initialization of TheoryStrings much easier, since the ExtTheory object can be created first.


src/theory/bv/theory_bv.h
```diff
@@ -180,6 +181,9 @@ class TheoryBV : public Theory {
   /** Index of the next literal to propagate */
   context::CDO<unsigned> d_literalsToPropagateIndex;
 
+  /** Extended theory module, for context-dependent simplification. */
+  std::unique_ptr<ExtTheory> d_extTheory;
```
4tXJ7f, line 185: This is minor but since we are always creating it together with `TheoryBv`, does it have to be a `unique_ptr` or could it just be a regular data member?

src/theory/bv/bv_subtheory_core.h
```diff
@@ -91,6 +92,9 @@ class CoreSolver : public SubtheorySolver {
   bool d_preregisterCalled;
   bool d_checkCalled;
   
+  /** Pointer to the extended theory module. */
+  ExtTheory * d_extTheory;
```
4tXJ7f, line 96: Apply ClangFormat.

src/theory/bv/theory_bv.h
ajreynol: I made it a pointer here since I think it *should* be configurable whether we create it or not :)

4tXJ7f: Ok, that's fair :)

PR #4754, opened 2020-07-15
Split abstract values utility from SmtEngine

ajreynol:
Towards refactoring SmtEngine.

### DIFF
--- a/src/smt/abstract_values.h
+++ b/src/smt/abstract_values.h
@@ -29,7 +29,7 @@ namespace smt {
 /**
  * This utility is responsible for constructing and maintaining abstract
  * values, which are used in some responses to e.g. get-value / get-model
- * commands.
+ * commands. See SMT-LIB standard 2.6 page 65 for details.
  */
 class AbstractValues
 {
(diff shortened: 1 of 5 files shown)


src/smt/abstract_values.h
```diff
@@ -0,0 +1,78 @@
...
+ * commands.
+ */
+class AbstractValues
+{
+  typedef std::unordered_map<Node, Node, NodeHashFunction> NodeToNodeHashMap;
+ public:
+  AbstractValues(NodeManager* nm);
+  ~AbstractValues();
+  /**
+   * Substitute away all AbstractValues in a node.
```
yoni206, line 41: What does that mean? example?

src/smt/abstract_values.h
```diff
@@ -0,0 +1,78 @@
...
+  Node mkAbstractValue(TNode n);
+
+ private:
+  /** Pointer to the used node manager */
+  NodeManager* d_nm;
+  /**
+   * A context that never pushes/pops, for use by CD structures (like
+   * SubstitutionMaps) that should be "global".
+   */
+  context::Context d_fakeContext;
```
yoni206, line 58: Would it make sense / worth while to create a sub-class of `Context`called `FakeContext` that overrides `push` and `pop` with an exception?

ajreynol: This may be overkill, I think its fine to just enforce this by encapsulation.

PR #4755, opened 2020-07-15
Use lemma property enum for OutputChannel::lemma

ajreynol:
There are 3 Boolean flags for OutputChannel::lemma, and plans to add another for relevance.

This makes them into a enum.


ajreynol:
Note that im a bit unclear about the best practice for the `operator&` methods, open for suggestions if there is a better way.

src/theory/output_channel.h
```diff
@@ -32,6 +32,36 @@
...
+  SEND_ATOMS = 4
+};
+/** Define operator lhs | rhs */
+LemmaProperty operator|(LemmaProperty lhs, LemmaProperty rhs);
+/** Define operator lhs |= rhs */
+LemmaProperty& operator|=(LemmaProperty& lhs, LemmaProperty rhs);
+/** Define operator lhs &= rhs */
+LemmaProperty& operator&=(LemmaProperty& lhs, LemmaProperty rhs);
+/** check set lhs & rhs */
+bool operator&(LemmaProperty lhs, LemmaProperty rhs);
```
nafur, line 54: I understand what you do and why you do it, but nevertheless see this as an implementation bug waiting to happen. As it stands, `|` and `&` do significantly different things. For example, I'd expect `lp & (REMOVABLE | SEND_ATOMS)` to return a `LemmaProperty`, not a `bool`. Also `a &= b` does something else than `a = a & b` and `|` and `&` really should be bitwise operators.
I would suggest to have `LemmaProperty operator&(LemmaProperty lhs, LemmaProperty rhs);` and have a separate method `bool abc(LemmaProperty bitmask, LemmaProperty flag)` called `test` or `isSet` or alike.

I guess this is what some people dislike about being able to overload operators in C++ ;-)

```suggestion
LemmaProperty operator&(LemmaProperty lhs, LemmaProperty rhs);
bool test(LemmaProperty lhs, LemmaProperty rhs);
```

src/theory/output_channel.h
```diff
@@ -32,6 +32,36 @@
 namespace CVC4 {
 namespace theory {
 
+/** Properties of lemmas */
+enum class LemmaProperty : uint32_t
```
nafur, line 36: Is there a particular reason why we explicitly specify the underlying datatype? (Why would we want to enforce a `uint32_t`?) I'd suggest to simply drop it.

```suggestion
enum class LemmaProperty
```

src/theory/output_channel.cpp
```diff
@@ -0,0 +1,108 @@
...
+  lhs = static_cast<LemmaProperty>(static_cast<uint32_t>(lhs)
+                                   & static_cast<uint32_t>(rhs));
+  return lhs;
+}
+bool operator&(LemmaProperty lhs, LemmaProperty rhs)
+{
+  LemmaProperty p = static_cast<LemmaProperty>(static_cast<uint32_t>(lhs)
+                                               & static_cast<uint32_t>(rhs));
+  return p != LemmaProperty::NONE;
+}
```
nafur, line 74: See comment in the header.
```suggestion
LemmaProperty operator&(LemmaProperty lhs, LemmaProperty rhs)
{
  return static_cast<LemmaProperty>(static_cast<uint32_t>(lhs)
                                    & static_cast<uint32_t>(rhs));
}
LemmaProperty& operator&=(LemmaProperty& lhs, LemmaProperty rhs)
{
  lhs = lhs & rhs;
  return lhs;
}
bool test(LemmaProperty lhs, LemmaProperty rhs)
{
  LemmaProperty p = static_cast<LemmaProperty>(static_cast<uint32_t>(lhs)
                                               & static_cast<uint32_t>(rhs));
  return p != LemmaProperty::NONE;
}
```

src/theory/arith/nl/nonlinear_extension.h
```diff
@@ -247,14 +248,10 @@ class NonlinearExtension
    * ensureLiteral respectively.
    */
   bool checkModel(const std::vector<Node>& assertions,
-                  const std::vector<Node>& false_asserts,
```
nafur, line 250: Do these changes belong in this PR?

src/theory/output_channel.h
ajreynol: Yes I totally agree.  And yes, I think I'm convinced now that we should just hardcode methods for testing each bit, e.g.:
```
bool isRemovable(LemmaProperty p);
bool isPreprocess(LemmaProperty p);
bool isSendAtoms(LemmaProperty p);
```

src/theory/output_channel.h
ajreynol: I am actually using this in the explicit casting for `operator&`. Dropping this would not ensure the proper cast.
Alternatively, I noticed its possible to get the underlying type automatically via type traits using C++14? https://en.cppreference.com/w/cpp/header/type_traits
I think I'd rather just hard code this for now.

nafur: According to https://en.cppreference.com/w/cpp/language/enum, the underlying type defaults to `int`. Unless you need way more bits, you could remove the explicit type specification and cast to `int`...

src/theory/arith/nl/nonlinear_extension.h
ajreynol: No ill revert.

src/theory/engine_output_channel.cpp
```diff
@@ -82,7 +82,7 @@ theory::LemmaStatus EngineOutputChannel::lemma(TNode lemma,
   d_engine->d_outputChannelUsed = true;
 
   PROOF({
-    bool preprocess = p & LemmaProperty::PREPROCESS;
+    bool preprocess = isPreprocess(p);
```
nafur, line 85: Shouldn't it be `isLemmaPropertyPreprocess()` here?

src/theory/output_channel.cpp
```diff
@@ -17,13 +17,39 @@
...
+{
+  return static_cast<LemmaProperty>(static_cast<uint32_t>(lhs)
+                                    | static_cast<uint32_t>(rhs));
+}
+LemmaProperty& operator|=(LemmaProperty& lhs, LemmaProperty rhs)
+{
+  lhs = lhs | rhs;
+  return lhs;
+}
+LemmaProperty operator|(LemmaProperty lhs, LemmaProperty rhs)
```
nafur, line 30:
```suggestion
LemmaProperty operator&(LemmaProperty lhs, LemmaProperty rhs)
```

src/theory/output_channel.cpp
```diff
@@ -17,13 +17,39 @@
...
+{
+  return static_cast<LemmaProperty>(static_cast<uint32_t>(lhs)
+                                    & static_cast<uint32_t>(rhs));
+}
+LemmaProperty& operator|=(LemmaProperty& lhs, LemmaProperty rhs)
```
nafur, line 35:
```suggestion
LemmaProperty& operator&=(LemmaProperty& lhs, LemmaProperty rhs)
```

src/theory/output_channel.h
ajreynol: I'm not sure what our coding guidelines say for this.  @aniemetz @4tXJ7f  ?

src/theory/engine_output_channel.cpp
ajreynol: Yes, I'm using CI to check compiler errors for me unfortunately :)

src/theory/output_channel.h
4tXJ7f: I think that since we use bit-wise operators (`|`) on the values of the enum, we should explicitly specify an unsigned type. Yes, it doesn't really matter since the values are all within the range of positive integer values but applying bit-wise operators to signed integers still feels wrong.

PR #4761, opened 2020-07-17
Fix Clang warnings

4tXJ7f:
This commit addresses warnings that we are getting with Clang 10.0.0.
Most of the warnings are related to unused private fields and mismatched
types when using range-based for loops (resulting in unnecessary
copies).


src/expr/dtype.cpp
```diff
@@ -167,7 +167,7 @@ bool DType::resolve(const std::map<std::string, TypeNode>& resolutions,
 
   d_involvesExt = false;
   d_involvesUt = false;
-  for (const std::shared_ptr<DTypeConstructor> ctor : d_constructors)
+  for (const std::shared_ptr<DTypeConstructor>& ctor : d_constructors)
```
nafur, line 170: How about adding a reference in the loop before as well?
Also: why not `const auto&`?
```suggestion
  for (const auto& ctor : d_constructors)
```

src/smt/model_core_builder.cpp
```diff
@@ -27,7 +27,7 @@ bool ModelCoreBuilder::setModelCore(const std::vector<Expr>& assertions,
   if (Trace.isOn("model-core"))
   {
     Trace("model-core") << "Compute model core, assertions:" << std::endl;
-    for (const Node& a : assertions)
+    for (const Expr& a : assertions)
```
nafur, line 30:
```suggestion
    for (const auto& a : assertions)
```

src/preprocessing/passes/ho_elim.cpp
```diff
@@ -319,7 +319,7 @@ PreprocessingPassResult HoElim::applyInternal(
   {
     std::map<Node, Node> lproc = newLambda;
     newLambda.clear();
-    for (const std::pair<Node, Node>& l : lproc)
+    for (const std::pair<const Node, Node>& l : lproc)
```
nafur, line 322:
```suggestion
    for (const auto& l : lproc)
```

src/printer/dagification_visitor.cpp
```diff
@@ -59,7 +59,7 @@ bool DagificationVisitor::alreadyVisited(TNode current, TNode parent) {
     // search for variables that start with the let prefix
     std::unordered_set<TNode, TNodeHashFunction> vs;
     expr::getVariables(current, vs);
-    for (const TNode v : vs)
+    for (const TNode& v : vs)
```
nafur, line 62:
```suggestion
    for (const auto& v : vs)
```

src/theory/arith/nl/nl_model.cpp
```diff
@@ -320,7 +320,7 @@ bool NlModel::checkModel(const std::vector<Node>& assertions,
     Node mg = nm->mkSkolem("model", nm->booleanType());
     gs.push_back(mg);
     // assert the constructed model as assertions
-    for (const std::pair<const Node, std::pair<Node, Node>> cb :
+    for (const std::pair<const Node, std::pair<Node, Node>>& cb :
```
nafur, line 323:
```suggestion
    for (const auto& cb :
```

src/theory/bv/theory_bv_utils.h
```diff
@@ -142,7 +142,10 @@ Node mkAnd(const std::vector<NodeTemplate<ref_count>>& conjunctions)
   if (all.size() == 1) { return conjunctions[0]; }
 
   NodeBuilder<> conjunction(kind::AND);
-  for (const Node& n : all) { conjunction << n; }
+  for (const TNode& n : all)
```
nafur, line 145:
```suggestion
  for (const auto& n : all)
```

src/theory/bv/theory_bv_utils.h
```diff
@@ -161,7 +164,10 @@ Node mkOr(const std::vector<NodeTemplate<ref_count>>& nodes)
   if (all.size() == 1) { return nodes[0]; }
 
   NodeBuilder<> disjunction(kind::OR);
-  for (const Node& n : all) { disjunction << n; }
+  for (const TNode& n : all)
```
nafur, line 167:
```suggestion
  for (const auto& n : all)
```

src/theory/quantifiers/sygus_inst.cpp
```diff
@@ -152,7 +152,7 @@ void SygusInst::registerQuantifier(Node q)
   /* Collect extra symbols in 'q' to be used in the grammar. */
   std::unordered_set<Node, NodeHashFunction> syms;
   expr::getSymbols(q, syms);
-  for (const TNode& var : syms)
+  for (const Node& var : syms)
```
nafur, line 155:
```suggestion
  for (const auto& var : syms)
```

src/theory/quantifiers/sygus_sampler.cpp
```diff
@@ -126,7 +126,7 @@ void SygusSampler::initializeSygus(TermDbSygus* tds,
   {
     TypeNode svt = sv.getType();
     // is it equivalent to a previous variable?
-    for (const std::pair<Node, unsigned>& v : var_to_type_id)
+    for (const std::pair<const Node, unsigned>& v : var_to_type_id)
```
nafur, line 129:
```suggestion
    for (const auto& v : var_to_type_id)
```

src/theory/quantifiers/term_database.cpp
```diff
@@ -273,14 +273,14 @@ void TermDb::computeUfEqcTerms( TNode f ) {
     ops.insert(ops.end(), d_ho_op_slaves[f].begin(), d_ho_op_slaves[f].end());
   }
   eq::EqualityEngine* ee = d_quantEngine->getActiveEqualityEngine();
-  for (const Node& ff : ops)
+  for (const TNode& ff : ops)
```
nafur, line 276:
```suggestion
  for (const auto& ff : ops)
```

src/theory/quantifiers/term_database.cpp
```diff
@@ -309,7 +309,7 @@ void TermDb::computeUfTerms( TNode f ) {
   unsigned relevantCount = 0;
   eq::EqualityEngine* ee = d_quantEngine->getActiveEqualityEngine();
   NodeManager* nm = NodeManager::currentNM();
-  for (const Node& ff : ops)
+  for (const TNode& ff : ops)
```
nafur, line 312:
```suggestion
  for (const auto& ff : ops)
```

src/theory/sets/cardinality_extension.cpp
```diff
@@ -168,7 +168,7 @@ void CardinalityExtension::checkCardinalityExtended(TypeNode& t)
       const std::map<Node, Node>& negativeMembers =
           d_state.getNegativeMembers(representative);
 
-      for (const std::pair<Node, Node>& negativeMember : negativeMembers)
+      for (const std::pair<const Node, Node>& negativeMember : negativeMembers)
```
nafur, line 171:
```suggestion
      for (const auto& negativeMember : negativeMembers)
```

src/theory/sets/theory_sets_private.cpp
```diff
@@ -1341,7 +1341,8 @@ bool TheorySetsPrivate::collectModelInfo(TheoryModel* m)
   {
     const std::map<TypeNode, std::vector<TNode> >& slackElements =
         d_cardSolver->getFiniteTypeSlackElements();
-    for (const std::pair<TypeNode, std::vector<TNode> >& pair : slackElements)
+    for (const std::pair<const TypeNode, std::vector<TNode> >& pair :
```
nafur, line 1344:
```suggestion
    for (const auto& pair :
```

src/theory/strings/inference_manager.cpp
```diff
@@ -532,7 +532,7 @@ void InferenceManager::explain(TNode literal,
   {
     ee->explainPredicate(atom, polarity, tassumptions);
   }
-  for (const TNode a : tassumptions)
+  for (const TNode& a : tassumptions)
```
nafur, line 535:
```suggestion
  for (const auto& a : tassumptions)
```

src/theory/strings/inference_manager.cpp
```diff
@@ -544,7 +544,7 @@ void InferenceManager::explain(TNode literal,
   {
     Debug("strings-explain-debug")
         << "Explanation for " << literal << " was " << std::endl;
-    for (const TNode a : tassumptions)
+    for (const TNode& a : tassumptions)
```
nafur, line 547:
```suggestion
    for (const auto& a : tassumptions)
```

src/theory/uf/symmetry_breaker.cpp
```diff
@@ -629,7 +629,7 @@ bool SymmetryBreaker::invariantByPermutations(const Permutation& p) {
     subs.clear();
     repls.clear();
     bool first = true;
-    for (const Node& nn : p)
+    for (const TNode& nn : p)
```
nafur, line 632:
```suggestion
    for (const auto& nn : p)
```

src/expr/dtype.cpp
nafur: While I see your concern, I'd still push for using `auto`, at least if the container we are iterating has been declared within a few lines of the loop. (This is not the case here, but for almost all the remaining comments).
I argue that the inconvenience of having long text (that don't really further the understanding of what's going on) and the danger of getting the type wrong (maybe also forgetting to change it in a refactoring, giving rise to superfluous copying) outweighs the benefits of seeing the type right there instead of having to look for the declaration of the container, at least if said declaration is only a few lines away.

aniemetz: I agree with @nafur and we explicitly allow the use of `auto` in our guidelines for exactly that reason. @ajreynol if you prefer to not use it , that's fine, but it is certainly allowed in general and even encouraged in some cases: https://github.com/CVC4/CVC4/wiki/Code-Style-Guidelines#auto

4tXJ7f:
I've updated the branch and resolved merge conflicts. As discussed during last week's meeting, I have a slight preference to keep the explicit types for now but agree that we should allow both in certain cases.

src/expr/dtype.cpp
```diff
@@ -171,7 +171,7 @@ bool DType::resolve(const std::map<std::string, TypeNode>& resolutions,
...
```
nafur, line 174: Add the `const&` in the loop before as well.

src/expr/lazy_proof_chain.cpp
```diff
@@ -40,7 +40,7 @@ const std::map<Node, std::shared_ptr<ProofNode>> LazyCDProofChain::getLinks()
     const
 {
   std::map<Node, std::shared_ptr<ProofNode>> links;
-  for (const std::pair<const Node, ProofGenerator*>& link : d_gens)
+  for (const std::pair<const Node, ProofGenerator* const>& link : d_gens)
```
nafur, line 43: Though we have no explicit recommendation, we seem to put `const` before the type everywhere else.

src/printer/dagification_visitor.cpp
nafur, line 62: So actually, after having thought about this some more (and written [this](https://github.com/CVC4/CVC4/wiki/Developer-Guide#dealing-with-expressions-nodes-and-tnodes)) I might even argue for only having `const TNode` here... :smile:

nafur (changes requested):
Almost good...

src/expr/dtype.cpp
4tXJ7f: I don't think we can, since we are actually changing members of `ctor`, e.g. `ctor->d_constructor.setAttribute(DTypeIndexAttr(), index)`.

src/expr/lazy_proof_chain.cpp
4tXJ7f: `const ProofGenerator*` and `ProofGenerator* const` are different. The latter means that the pointer itself cannot be changed (e.g. to point at a different object) while the former means that the object itself cannot be changed and it is what we seem to be getting from `d_gens` according to the compiler.

src/printer/dagification_visitor.cpp
4tXJ7f: I see your point but I think that we anyway make an exception for range-based for loops with `const Node&`, so `const TNode&` IMHO makes sense for the sake of uniformity (and to shut up the compiler :) ).

src/expr/lazy_proof_chain.cpp
nafur: Ah, indeed. I always get confused about whether `const T*` is equivalent to `T const *` or `T* const`... everything is fine then.

src/printer/dagification_visitor.cpp
nafur: I guess I agree. The only argument for `const TNode` would be improved cache locality, and I'd hope that this is not a big issue for a loop...

src/expr/dtype.cpp
nafur: The `const&` would only be about the `std::shared_ptr` itself, not what it points to. (It's not `std::shared_ptr<const DTypeConstructor>`...)

aniemetz:
Closed due to too many conflicts. Some are already fixed on master, the remainder should be fixed while checking current state of master with `-Wunused-private-field`. Will be less work than fixing all the conflicts.

PR #4763, opened 2020-07-17
Improve accuracy of resource limitation

nafur:
The main goal of the resource limitation mechanism (via `--rlimit`) is to have a deterministic limitation on runtime. The ultimate goal would be that the actual runtime grows linearly with the spent resources.
To achieve this, this PR does the following:
- introduce new resources spent in places that are not yet covered
- find resource weights that best approximate the runtime

It provides a contrib script `learn_resource_weights.py` that uses statistics from a given benchmark run and uses linear regression to find good resource weights. The script also evaluates this estimate and identifies outliers, benchmarks for which this approximation is particularly bad. This probably indicates that on such a benchmark, some part of the code takes a significant amount of time but is not yet represented by a resource.

Eventually, we should use the resulting resource weights as defaults for the options like `--rewrite-step`, `--theory-check-step`, etc.


src/theory/rewriter.cpp
```diff
@@ -205,11 +205,9 @@ Node Rewriter::rewriteTo(theory::TheoryId theoryId,
   }
   // Rewrite until the stack is empty
   for (;;){
-
-    if (hasSmtEngine &&
-		d_iterationCount % ResourceManager::getFrequencyCount() == 0) {
+    if (hasSmtEngine)
```
4tXJ7f, line 208: I think that we are able to get rid of `d_iterationCount` altogether.

src/theory/arith/theory_arith_private.cpp
```diff
@@ -3478,6 +3478,23 @@ void TheoryArithPrivate::check(Theory::Effort effortLevel){
...
+    }
+  }
+  else
+  {
+    for (std::size_t i = 0; i < d_dualSimplex.getPivots(); ++i)
+    {
+      d_containing.d_out->spendResource(
+          ResourceManager::Resource::ArithPivotStep);
+    }
+  }
```
4tXJ7f, line 3496: How about:

```suggestion
  size_t nPivots = options::useFC() ? d_fcSimplex.getPivots() : d_dualSimplex.getPivots();
  for (std::size_t i = 0; i < d_fcSimplex.getPivots(); ++i)
  {
    d_containing.d_out->spendResource(
        ResourceManager::Resource::ArithPivotStep);
  }
```

I actually started thinking about it because the loops don't quite conform to our style guidelines for [loop conditions](https://github.com/CVC4/CVC4/wiki/Code-Style-Guidelines#loop-conditions).

src/prop/cnf_stream.cpp
```diff
@@ -709,11 +708,7 @@ void TseitinCnfStream::convertAndAssert(TNode node, bool negated) {
   Debug("cnf") << "convertAndAssert(" << node
                << ", negated = " << (negated ? "true" : "false") << ")" << endl;
 
-  if (d_convertAndAssertCounter % ResourceManager::getFrequencyCount() == 0) {
-    d_resourceManager->spendResource(ResourceManager::Resource::CnfStep);
-    d_convertAndAssertCounter = 0;
-  }
-  ++d_convertAndAssertCounter;
+  d_resourceManager->spendResource(ResourceManager::Resource::CnfStep);
```
4tXJ7f, line 711: I believe that we can get rid of `d_convertAndAssertCounter` altogether.

contrib/learn_resource_weights.py
```diff
@@ -0,0 +1,167 @@
...
+
+import json
+import logging
+
+logging.basicConfig(format='[%(levelname)s] %(message)s')
+
+
+def parse_commandline():
+    """Parse commandline arguments"""
+    import argparse
```
4tXJ7f, line 11: I prefer to have all imports at the top of the file unless there is a good reason not to do so. Imports at the top of the file is also consistent with other Python scripts [[1](https://github.com/CVC4/CVC4/blob/c9e23f66383a4d490aca6d082d40117fe799ee4b/src/options/mkoptions.py), [2](https://github.com/CVC4/CVC4/blob/c9e23f66383a4d490aca6d082d40117fe799ee4b/test/regress/run_regression.py), [3](https://github.com/CVC4/CVC4/blob/c9e23f66383a4d490aca6d082d40117fe799ee4b/test/signatures/run_test.py)].

contrib/learn_resource_weights.py
```diff
@@ -0,0 +1,167 @@
...
+    parser = argparse.ArgumentParser(description='export and analyze resources from statistics',
+                                     formatter_class=argparse.ArgumentDefaultsHelpFormatter)
+    parser.add_argument('command', choices=[
+                        'parse', 'analyze'], help='path of benchmark results')
```
4tXJ7f, line 15: This seems to copy the help text from the option `basedir`. How about:

```suggestion
                        'parse', 'analyze'], help='task to perform')
```

contrib/learn_resource_weights.py
```diff
@@ -0,0 +1,167 @@
...
+    with gzip.GzipFile(args.json, 'w') as fout:
+        fout.write(json.dumps(data).encode('utf-8'))
+
+
+def get_sorted_values(data):
+    """Transform [['name', value], ...] to [value, ...] sorted by names"""
+    return [d[1] for d in sorted(data)]
+
+
+if args.command == 'parse':
```
4tXJ7f, line 57: Having the "main" code interspersed with function definitions makes it a bit hard to read (IMHO). Could we define a `main()` function?

contrib/learn_resource_weights.py
```diff
@@ -0,0 +1,167 @@
...
+    if args.basedir == None:
+        raise Exception('Specify basedir for parsing!')
+    import glob
+    import re
+    filename_re = re.compile(
+        '.*_incremental_([A-Z_]+/[0-9a-zA-Z_=\\./-]+\\.smt2)/output\\.log')
```
4tXJ7f, line 63: Why not simply:
```suggestion
        '.*\\.smt2)/output\\.log')
```
to make sure that it also works on custom benchmark sets.

contrib/learn_resource_weights.py
```diff
@@ -0,0 +1,167 @@
...
+        predict = float(r.predict([vals])) / args.mult
+        outliers['over-estimated'].append([predict /
+                                           actual, predict, actual, filename])
+        outliers['under-estimated'].append([actual /
+                                            predict, predict, actual, filename])
+
+    for out in outliers:
+        logging.info('Showing outliers for {}'.format(out))
+        filtered = outliers[out]
+        #filtered = filter(lambda o: o[2] > 1 and o[2] < 30, filtered)
```
4tXJ7f, line 158:
```suggestion
```

contrib/learn_resource_weights.py
```diff
@@ -0,0 +1,167 @@
...
+    logging.info('Extracting resources')
+    resources = set()
+    for f in data:
+        spend_something = False
+        for r in data[f]['resources']:
+            if r[0].startswith('Bv') and r[1] > 0:
+                spend_something = True
+            resources.add(r[0])
+        if spend_something and f.startswith('QF_BV/') and data[f]['time'] < 0.004:
+            print(f)
```
4tXJ7f, line 105: Am I missing something or does this look like a leftover of a debugging session? :)

4tXJ7f (changes requested):
Cool! Some minor comments.

src/theory/rewriter.cpp
nafur: I think I've done that... `grep -Inri d_iterationCount src/` doesn't find anything...

src/prop/cnf_stream.cpp
nafur: Ditto.

contrib/learn_resource_weights.py
```diff
@@ -0,0 +1,161 @@
+#!/usr/bin/env python3
+
```
ajreynol, line 2: This should be documented with at least some of the text from PR description.

ajreynol (changes requested):

contrib/learn_resource_weights.py
nafur: I've added some documentation (run the script with `--help`). The PR description already has a full paragraph about this script.

ajreynol (approved):
LGTM

src/smt/smt_engine.cpp
```diff
@@ -263,15 +263,15 @@ void SmtEngine::finishInit()
   // dumping the command twice.
```
4tXJ7f, line 263: It seems like this file didn't change at all (or very little?) except for formatting?

4tXJ7f (approved):
LGTM, thanks for the updates. Before merging, PTAL at src/smt/smt_engine.cpp.

src/smt/smt_engine.cpp
nafur: Indeed, undid all the reformatting for now.

PR #4764, opened 2020-07-17
Replace options listener infrastructure

ajreynol:
This replaces the old options listener infrastructure with the OptionsManager introduced in https://github.com/CVC4/CVC4/commit/cb8d041d3820a46721f689f188839184003e0e7c.

It eliminates a "beforeSearchListener", which was a custom way of *some* options throwing a modal exception if they were set after initialization. Now all options are consistent: no option can be set after initialization.

It also moves managed ostream objects to the OptionsManager. 

@mpreiner The next step will be to remove the "notifies" field from the Options build system and then proceed with cleaning `src/options/`.


src/util/resource_manager.h
```diff
@@ -200,8 +197,8 @@ class CVC4_PUBLIC ResourceManager
   /** Counter indicating how often to check resource manager in loops */
   static const uint64_t s_resourceCount;
 
-  /** Receives a notification on reaching a limit. */
-  ListenerCollection d_listeners;
+  /** Receives a notification on reaching a resource out. */
```
nafur, line 200: It may still be a (per-call) timeout. The only thing that was removed was the cumulative time limit.

src/util/resource_manager.h
```diff
@@ -163,11 +163,8 @@ class CVC4_PUBLIC ResourceManager
 
   /**
    * Registers a listener that is notified on a resource out.
-   *
-   * This Registration must be destroyed by the user before this
-   * ResourceManager.
    */
-  ListenerCollection::Registration* registerListener(Listener* listener);
+  void registerResourceOutListener(Listener* listener);
```
nafur, line 167: See below.

PR #4765, opened 2020-07-17
Fix regular expression delta for complement

### LINKED ISSUE #4759: (default)  Fatal failure at  /src/./expr/node_builder.h:655
```
[561] % cvc4 small.smt2
Fatal failure within CVC4::NodeBuilder<N>& CVC4::NodeBuilder<N>::append(CVC4::TNode) 
[with unsigned int nchild_thresh = 10; CVC4::TNode = CVC4::NodeTemplate<false>] /src/./expr/node_builder.h:655
Check failure
 !n.isNull()
Cannot use NULL Node as a child of a Node
Aborted
[562] % 
[562] % cat small.smt2
(set-logic QF_S)
(declare-fun a () String)
(assert (str.in_re "" (re.++ (str.to_re a) (re.comp re.all))))
(check-sat)
[563] %
```
OS: Ubuntu 18.04
Commit: 70353e7

ajreynol:
Fixes #4759 . 

Also refactors this method.


ajreynol:
REGEXP_EMPTY means "contains no string", not "contains the empty string".

nafur:
> REGEXP_EMPTY means "contains no string", not "contains the empty string".

Ah I see.

src/theory/strings/regexp_operation.cpp
```diff
@@ -117,161 +117,146 @@ RegExpConstType RegExpOpr::getRegExpConstType(Node r)
 
 // 0-unknown, 1-yes, 2-no
 int RegExpOpr::delta( Node r, Node &exp ) {
-  Trace("regexp-delta") << "RegExp-Delta starts with /" << mkString( r ) << "/" << std::endl;
+  std::map<Node, std::pair<int, Node> >::iterator itd = d_delta_cache.find(r);
```
nafur, line 120:
```suggestion
  auto itd = d_delta_cache.find(r);
```
And if you really don't want `auto`, I'd suggest to at least use a `const_iterator`.

PR #4775, opened 2020-07-20
Migrating from using the 'glpk-cut-log' repo to using an official GPLK release + a set of patches

andrewvaughanj:
This PR attempts to migrate from @timothy-king's repo for [glpk-cut-log](https://github.com/timothy-king/glpk-cut-log) to using a set of patches against the official release that 'glpk-cut-log' was based off of (4.52).

This is in preparation of trying to rework these patches to be against the latest GPLK release (4.65). If we can do this, then it makes it easier for CVC4's dependancy on GPLK to be able to follow upstream (rather than being stuck on a release that is 6.5 years old!).

I have added GPLK as an option for CI, but I do not know if we actually want this. My concern with adding this to CI is that we're then not testing non-GPL builds, which doesn't seem ideal.

However, before starting to rework the patches to be against 4.65, I want to make sure that things are actually working/stable against 4.52; so having at least one CI target that *does* use GPLK would be great.


andrewvaughanj:
I think we still need to decide if we want to add a GPL configuration to CI or not.

mpreiner:
In general, we want to keep the number of necessary CI builds low (due to the high volume on PRs). I'd rather push this to the nightlies since it's a non-critical build.

andrewvaughanj:
Okay, so shall I back-out the Github Action changes? Will this "obviously fail" (in a way you can tell me) if the nightlies start to break?

mpreiner:
Yes, please remove the GH Action changes, I'll add a gpl nightly build soon.

andrewvaughanj:
@mpreiner *poof* CI changes are gone.

PR #4776, opened 2020-07-20
Consider negation's proof before triggering arith pfs.

### LINKED ISSUE #4714: (--check-unsat-cores) Fatal failure at  src/proof/arith_proof_recorder.cpp:45
```
[587] % cvc4 small.smt2
unknown
[588] % cvc4 --check-unsat-cores small.smt2
Fatal failure within void CVC4::proof::ArithProofRecorder::saveFarkasCoefficients(CVC4::Node,  
CVC4::theory::arith::RationalVectorCP) at src/proof/arith_proof_recorder.cpp:45 
Check failure 
 nonNegativeChild.getType().isBoolean() && nonNegativeChild[0].getType().isReal() 
Aborted
[589] % 
[589] % cat small.smt2
(set-logic UFNIRA)
(declare-fun c (Int) Int)
(define-fun d ((k Int)) Int (- (c k) 10))
(define-fun j ((k Int) (e Int)) Bool (<= 0 e (d k)))
(define-fun f ((k Int) (a Int) (b Int)) Int (ite (= b 1) a (mod a b)))
(define-fun g ((k Int) (a Int) (b Int)) Int (f k (* a (c b)) (c k)))
(define-fun h ((k Int) (a Int) (b Int)) Int (f k (+ a b) (c k)))
(assert (= (c 0) 1))
(assert (exists ((l Int) (k Int)) (and (j k l) (distinct (g k (h k l (g k l l)) l) (g k (h k l 0) l)))))
(check-sat)
[590] %
```
OS: Ubuntu 18.04
Commit:  ba7cda7

alex-ozdemir:
The current arith proof machinery can prove conflicts which are
explained by assumptions and tightened assumptions.

Previously we verified that the conflict constraint was explained in
terms of assumptions and tightened assumptions, before trying to
save/produce a proof.

We did not verify that the negated constraint was an assumption or
tightened assumption.

This caused us to attempt to prove conflicts around constraints whose
negations where proven by general means (in the case of #4714, by the
equality engine), which the proof machinery was not prepared to handle.

This PR also checks that the negate constraint is an assumption or
tightened assumption, before saving the proof.

Thanks, Gereon, for doing the initial investigation into this!

fixes #4714

Co-authored-by: Gereon Kremer <gereon.kremer@cs.rwth-aachen.de>


src/theory/arith/constraint.cpp
```diff
@@ -536,6 +526,16 @@ bool Constraint::hasSimpleFarkasProof() const
...
 }
 
+bool Constraint::isPossiblyTightenedAssumption() const
+{
+  // ... that antecdent must be an assumption ...
+  return (isAssumption()
+          || (hasIntTightenProof()
+              && getConstraintRule().d_antecedentEnd != AntecedentIdSentinel
+              && d_database->getAntecedent(getConstraintRule().d_antecedentEnd)
+                     ->isAssumption()));
```
nafur, line 536: I would suggest to unroll the logic into code...
```suggestion
  
  if (isAssumption()) return true;
  if (!hasIntTightenProof()) return false;
  if (getConstraintRule().d_antecedentEnd == AntecedentIdSentinel) return false;
  return d_database->getAntecedent(getConstraintRule().d_antecedentEnd)
                     ->isAssumption();
```

src/theory/arith/theory_arith_private.cpp
```diff
@@ -1933,7 +1933,7 @@ void TheoryArithPrivate::outputConflicts(){
         }
 
         Assert(conflict.getNumChildren() == pf.d_farkasCoefficients->size());
-        if (confConstraint->hasSimpleFarkasProof())
+        if (confConstraint->hasSimpleFarkasProof() && confConstraint->getNegation()->isPossiblyTightenedAssumption())
```
nafur, line 1936: Is this formatted?

alex-ozdemir: Ah! It was not. Good catch. Fixed in be32fcb48.

alex-ozdemir:
Thanks for the review, Gereon (@nafur). I think that I've addressed your comments. Why don't you take a look?

PR #4777, opened 2020-07-20
Support uninterpreted constants in the evaluator

ajreynol:


test/regress/regress1/sygus/eval-uc.sy
```diff
@@ -0,0 +1,16 @@
+; EXPECT: unsat
+; COMMAND-LINE: --lang=sygus2 --sygus-out=status
```
yoni206, line 2:
```suggestion
; COMMAND-LINE: --lang=sygus2 --sygus-out=status --uf-ho
```

test/regress/regress1/sygus/eval-uc.sy
```diff
@@ -0,0 +1,16 @@
...
+(set-logic ALL)
+(declare-sort S 0)
+(declare-var f Bool)
+(declare-var u (-> S Bool))
+(declare-var new_f Bool)
+(declare-var new_u (-> S Bool))
+(define-fun init ((f Bool) (u (-> S Bool))) Bool (and f (forall ((x S)) (not (u x)))))
+(define-fun trans ((f Bool) (u (-> S Bool)) (new_f Bool) (new_u (-> S Bool))) Bool (and (not new_f) (exists ((x S)) (forall ((y S)) (= (new_u y) (or (u y) (= y x)))))))
```
yoni206, line 10: This line should be deleted.

test/regress/regress1/sygus/eval-uc.sy
```diff
@@ -0,0 +1,16 @@
...
+;(define-fun trans ((f Bool) (u (-> S Bool)) (new_f Bool) (new_u (-> S Bool))) Bool (and new_f (exists ((x S)) (forall ((y S)) (= (new_u y) (or (u y) (= y x)))))))
```
yoni206, line 11: This line should be uncommented

ajreynol: This makes the benchmarks become "unknown" (infeasible).  I'd like to keep this example as unsat. While the original bug does not occur with the above line, I'd like to keep it like this for reference.

yoni206: Maybe try the original one now? I think it should be feasible, and then maybe it makes sense to add it to the regressions.
```
(set-logic ALL)

; type S
; mutable relation f
; mutable relation u(S)
; init f & forall X. !u(x)
; transition trans1(x: S)
;   # insert element into u, changing f from true to false
;   modifies f, u
;   f & !new(f) & forall Y. new(u(Y)) <-> (u(Y) | Y = x)
; transition trans2(x: S)
;   # remove element from u, setting f to true
;   modifies f, u
;   new(f) &  forall Y. new(u(Y)) <-> (u(Y) & Y != x)
; invariant [safety] forall X,Y. u(X) & u(Y) -> X = Y
; invariant forall X. u(X) -> !f

(declare-sort S 0)
(declare-var f Bool)
(declare-var u (-> S Bool))
(declare-var new_f Bool)
(declare-var new_u (-> S Bool))

(define-fun init ((f Bool) (u (-> S Bool))) Bool (and f (forall ((x S)) (not (u x)))))

(define-fun trans1 ((f Bool) (u (-> S Bool)) (new_f Bool) (new_u (-> S Bool))) Bool (and f (not new_f) (exists ((x S)) (forall ((y S)) (= (new_u y) (or (u y) (= y x)))))))

(define-fun trans2 ((f Bool) (u (-> S Bool)) (new_f Bool) (new_u (-> S Bool))) Bool (and new_f (exists ((x S)) (forall ((y S)) (= (new_u y) (or (u y) (= y x)))))))

(define-fun safe ((f Bool) (u (-> S Bool))) Bool (forall ((x S) (y S)) (=> (and (u x) (u y)) (= x y))))



(synth-fun inv_matrix ((f Bool) (u (-> S Bool)) (x S)) Bool 
   ((Start Bool))
   ((Start Bool ( 
     (or (not (u x)) (not f))
   ))))
;     f
;     (u StartS)
;     (and Start Start)
;     (or Start Start)
;     (not Start)
;   ))
;   (StartS S ( x ))
;)
;)



(define-fun inv ((f Bool) (u (-> S Bool))) Bool (and (forall ((x S)) (inv_matrix f u x)) (safe f u)))


(constraint (=> (init f u) (inv f u)))
(constraint (=> (and (inv f u) (trans1 f u new_f new_u)) (inv new_f new_u)))
(constraint (=> (and (inv f u) (trans2 f u new_f new_u)) (inv new_f new_u)))


(check-synth)
```

yoni206: When I do the suggested change I indeed get `unknown` on production but an error on debug:
```
Fatal failure within bool CVC4::theory::quantifiers::Instantiate::addInstantiation(CVC4::Node, std::vector<CVC4::NodeTemplate<true> >&, bool, bool, bool) at /home/yoniz/git/CVC4_1/src/theory/quantifiers/
instantiate.cpp:176
Check failure

 false

Aborted (core dumped)

```

ajreynol: This is another unrelated issue, pertaining to the use of nested quantifiers + uninterpreted sorts in sygus conjectures. I would suggest addressing this on another PR, since it is more complicated to resolve.

PR #4779, opened 2020-07-21
GH Actions: Cancel builds on push, remove redundant mac OS build.

aniemetz:


.github/workflows/ci.yml
```diff
@@ -14,6 +21,10 @@ jobs:
           debug-cln
         ]
 
+        exclude:
+          - name: production-clang
+            os: macos-latest
+
```
4tXJ7f, line 27:
```suggestion
            - name: debug
              os: macos-latest

```

As discussed offline, I think that it would be good to actually remove two macOS builds to speed up our CI because we only have [5 macOS workers](https://docs.github.com/en/actions/getting-started-with-github-actions/about-github-actions#usage-limits).

.github/workflows/ci.yml
```diff
@@ -2,8 +2,15 @@ on: [push, pull_request]
 name: CI
 
 jobs:
-  build:
+  test:
```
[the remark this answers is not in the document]
aniemetz: Ah, right. Whoops :D

.github/workflows/ci.yml
aniemetz: Removed all but one production build.

PR #4781, opened 2020-07-22
Replace Expr with Node in Term/Op

4tXJ7f:
This commit changes `Term` and `Op` to use `Node`s internally instead of
`Expr`s. This is a step towards removing the `Expr`-layer. This commit also
adds some missing checks regarding the number of arguments for a given
kind that were previously missing, which caused issues in unit tests when
using `Node` instead of `Expr`.


src/api/cvc4cpp.cpp
```diff
@@ -642,7 +643,16 @@ uint32_t minArity(Kind k)
 {
   Assert(isDefinedKind(k));
   Assert(isDefinedIntKind(extToIntKind(k)));
-  return CVC4::ExprManager::minArity(extToIntKind(k));
+  uint32_t min = CVC4::ExprManager::minArity(extToIntKind(k));
+
+  // special cases for API level
+  // higher-order logic perspective at API
```
aniemetz, line 649: This sentence is a bit weird.

PR #4783, opened 2020-07-22
(proof-new) Arithmetic operator elim proof producing

ajreynol:
This updates the interface for arithmetic operator elimination for the new proof format.

The actual proof production of the operator elimination class (providing proofs for introduced witness terms) will be done in a separate PR.

This also changes the witness terms introduced by this class so their body is in Skolem form, which simplifies term formula removal.

### DIFF
--- a/src/theory/arith/operator_elim.h
+++ b/src/theory/arith/operator_elim.h
@@ -30,7 +30,13 @@ class OperatorElim : public EagerProofGenerator
  public:
   OperatorElim(ProofNodeManager* pnm, const LogicInfo& info);
   ~OperatorElim() {}
-  /** eliminate */
+  /** Eliminate operators in this term.
+    *
+    * Eliminate operators in term n. If n has top symbol that is not a core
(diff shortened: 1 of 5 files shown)


src/theory/arith/operator_elim.h
```diff
@@ -17,43 +17,26 @@
...
 namespace arith {
 
-class OperatorElim
+class OperatorElim : public EagerProofGenerator
 {
  public:
-  OperatorElim(const LogicInfo& info);
+  OperatorElim(ProofNodeManager* pnm, const LogicInfo& info);
   ~OperatorElim() {}
+  /** eliminate */
```
alex-ozdemir, line 33: I think it would be good to have some more documentation for this, since this is now the public interface.

```suggestion
  /** Eliminate operators in this term.
    *
    * Eliminate operators in term n. If n has top symbol that is not a core
    * one (including division, int division, mod, to_int, is_int, syntactic sugar
    * transcendental functions), then we replace it by a form that eliminates
    * that operator. This may involve the introduction of witness terms.
    */
```

PR #4797, opened 2020-07-27
Generalize interface for candidate rewrite database

ajreynol:
This class will be used as a utility in a new algorithm for solution reconstruction and requires a generalized interface.

FYI @abdoo8080


src/theory/quantifiers/candidate_rewrite_database.h
```diff
@@ -93,14 +110,18 @@ class CandidateRewriteDatabase : public ExprMiner
   ExtendedRewriter* d_ext_rewrite;
   /** the function-to-synthesize we are testing (if sygus) */
   Node d_candidate;
+  /** whether we are checking equivalence using subsolver */
+  bool d_doCheck;
+  /** if true, we use acceleration for symmetry breaking rewrites */
```
HanielB, line 115: could add a reference to where acceleration is defined

PR #4799, opened 2020-07-27
Delete solver pointer in Cython __dealloc__

makaimann:
This PR fixes a memory leak in the python bindings. The Cython `Solver` object keeps a pointer to a solver, which was never deleted.

To get this to work, we need to allow Python to do reference counting. Otherwise, it might delete the underlying solver object when there are still other objects that depend on it at the C++ level (e.g. `Terms`/`Sorts`). To this end, this PR adds a reference to the (Cython-level) `Solver` object in all classes that depend on the solver (i.e. keep a pointer to the solver at the C++ level)


src/api/python/cvc4.pxi
```diff
@@ -53,6 +53,22 @@ def expand_list_arg(num_req_args=0):
...
+# https://groups.google.com/forum/#!topic/cython-users/fuKd-nQLpBs
+# To get around that you can either have a nullary constructor and assignment
+# or, use a pointer (which is what we chose)
+# An additional complication of this is that to free up resources, you must
+# know when to delete the object
+# Python will not follow the same scoping rules as in C++, so it must be
+# able to reference count. To do this correctly, the solver must be a
+# reference in the Python class for any class that keeps a pointer to
+# the solver in C++ (to ensure the solver is not deleted before something
+# that depends on it).
```
aniemetz, line 70: Add `.` after sentences.

makaimann: Cool, addressed with 0f7fc58550187e2df979d2109a50bbe10c8f0c8e.

makaimann:
Looks like there are some issues related to changes in master (the grammar interface). I'll fix that and push soon.

PR #4800, opened 2020-07-27
Fixes for degenerate case of sygus decision tree learning

### LINKED ISSUE #4790: Sygus ite performance issue
**Describe the bug**
Using sygus mode, the ite operator times out with a nonterminal as a parameter, even is that nonterminal is a single terminal symbol.

This sygus query times out
```
(set-logic NIA)
(synth-fun patternGen ((i Int)) Int
((ITE Int) (CONST Int) (I Int)  (B Bool))
( (ITE Int (
    (ite (<= i CONST) I I)
    
  ))
  (CONST Int (0))
  (I Int (i 1
    (+ I I) (- I I) (* I I)
  ))
  (B Bool (
    (<= I I)
  ))
)
)
(declare-var i Int)
(constraint (= (patternGen 0) 1))
(constraint (= (patternGen 1) 1))
(constraint (= (patternGen 2) 1))
(check-synth)
```

But when the CONST in the ite is replaced with the nontermial 0 directly (i.e ```(ite (<= i 0) I I)```), it finishes immediately and returns  ```(define-fun patternGen ((i Int)) Int (ite (<= i 0) 1 1))```. The same thing happens if CONST is a list of lots of nonterminal ints - it times out with CONST, but terminates if I just define ITE be ```(ite (<= i 0) I I) (ite (<= i 1) I I) (ite (<= i 2) I I) ...```.

**CVC4 version/commit**:  1.7
**Operating system**: Ubuntu Linux

ajreynol:
Fixes #4790. 

Fixes two bugs in the decision tree learning solver for sygus, one involving evaluation of "templated" enumerators, and one involving ITE strategies where child types are different than the root.


src/theory/quantifiers/sygus/sygus_unif_io.cpp
```diff
@@ -571,43 +576,22 @@ void SygusUnifIo::notifyEnumeration(Node e, Node v, std::vector<Node>& lemmas)
...
+      TNode tbv = bv;
+      Node stempl = templ.substitute(targ, tbv);
       std::vector<Node> sresults;
-      for (const Node& res : base_results)
-      {
-        TNode tres = res;
-        Node sres;
-        // It may not be constant, e.g. if we involve a partial operator
-        // like datatype selectors. In this case, we avoid using the evaluator,
-        // which expects a constant substitution.
```
HanielB, line 593: Why not do this anymore?

src/theory/quantifiers/sygus/sygus_unif_io.cpp
```diff
@@ -1442,12 +1429,30 @@ Node SygusUnifIo::constructSol(
...
           else
           {
-            // TODO (#1250) : degenerate case where children have different
-            // types?
-            indent("sygus-sui-dt", ind);
-            Trace("sygus-sui-dt") << "return PBE: failed ITE strategy, "
-                                      "cannot find a distinguishable condition"
-                                  << std::endl;
+            // if the child types are different, it could still make a
+            // difference to recurse
```
HanielB, line 1433: give more details, possibly an example

src/theory/quantifiers/sygus/sygus_unif_io.cpp
```diff
@@ -49,6 +49,11 @@ bool UnifContextIo::updateContext(SygusUnifIo* sui,
   Node poln = pol ? d_true : d_false;
   for (unsigned i = 0; i < vals.size(); i++)
```
HanielB, line 50: since you're modifying below, update the loop above to cache the size

src/theory/quantifiers/sygus/sygus_unif_io.cpp
```diff
@@ -49,6 +49,11 @@ bool UnifContextIo::updateContext(SygusUnifIo* sui,
...
   {
+    if (vals[i].isNull())
+    {
+      // nothing can be inferred if the evaluation is unknown
```
HanielB, line 54: Give an example of what would lead to not having a value for an example

src/theory/quantifiers/sygus/sygus_unif_io.cpp
ajreynol: This is old code that was doing things suboptimally.  The evaluator class is now robust to this.

src/theory/quantifiers/sygus/sygus_unif_io.cpp
```diff
@@ -47,13 +47,20 @@ bool UnifContextIo::updateContext(SygusUnifIo* sui,
...
+      // partial functions.
+      continue;
+    }
+    if (v != poln)
     {
-      if (d_vals[i] == d_true)
+      if (v == d_true)
       {
-        d_vals[i] = d_false;
+        vals[i] = d_false;
```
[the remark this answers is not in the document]
ajreynol: Oh yikes, this was a typo. I'm surprised it still passes.

PR #4801, opened 2020-07-27
Add option to only build library

makaimann:
This PR would add an option to only build the CVC4 library and not the parser or executable. This can be used for projects that only intend to use CVC4 through the API.

It seems to be working now, but it's not necessarily the cleanest solution. In particular, if you'd like the polarity to be different I'm happy to change that. Polarity meaning something like `"${WITH_BINARY}" STREQUAL "YES"` instead of `NOT "${LIB_ONLY} STREQUAL "YES"` which is admittedly a little strange.


configure.sh
```diff
@@ -334,6 +339,8 @@ do
     --symfpu-dir) die "missing argument to $1 (try -h)" ;;
     --symfpu-dir=*) symfpu_dir=${1##*=} ;;
 
+    --lib-only) lib_only=YES ;;
```
mpreiner, line 342: Use ON/OFF similar to the other options.

src/CMakeLists.txt
```diff
@@ -843,7 +843,10 @@ set(KINDS_FILES
 add_subdirectory(base)
 add_subdirectory(expr)
 add_subdirectory(options)
-add_subdirectory(parser)
+if (NOT "${LIB_ONLY}" STREQUAL "YES")
+  message("Adding parser")
```
mpreiner, line 847: I don't think that we need this message.

src/CMakeLists.txt
```diff
@@ -843,7 +843,10 @@ set(KINDS_FILES
 add_subdirectory(base)
 add_subdirectory(expr)
 add_subdirectory(options)
-add_subdirectory(parser)
+if (NOT "${LIB_ONLY}" STREQUAL "YES")
```
mpreiner, line 846:
```suggestion
if (NOT LIB_ONLY)
```

Should also work.

src/CMakeLists.txt
```diff
@@ -935,7 +938,10 @@ target_link_libraries(cvc4 ${RT_LIBRARIES})
 # target_link_libraries(...) with object libraries for cmake versions <= 3.12.
 # Thus, we can only visit main as soon as all dependencies for cvc4 are set up.
 
-add_subdirectory(main)
+if (NOT "${LIB_ONLY}" STREQUAL "YES")
```
mpreiner, line 941:
```suggestion
if (NOT LIB_ONLY)
```

src/CMakeLists.txt
```diff
@@ -935,7 +938,10 @@ target_link_libraries(cvc4 ${RT_LIBRARIES})
...
+  message("Adding main")
```
mpreiner, line 942: Ditto

test/CMakeLists.txt
```diff
@@ -21,7 +21,9 @@ add_custom_target(check
 #-----------------------------------------------------------------------------#
 # Add subdirectories
 
-add_subdirectory(regress)
+if (NOT "${LIB_ONLY}" STREQUAL "YES")
```
mpreiner, line 24:
```suggestion
if (NOT LIB_ONLY)
```

makaimann:
Thanks for the review! Addressed your comments in the last few commits.

configure.sh
```diff
@@ -80,6 +80,9 @@ Optional Path to Optional Packages:
   --poly-dir=PATH          path to top level of LibPoly source tree
   --symfpu-dir=PATH        path to top level of SymFPU source tree
 
+Build limitations:
+  --lib-only               only build the library, but not the executable or the parser (default: off)
```
mpreiner, line 84: Make sure this line is 80 chars.

src/CMakeLists.txt
mpreiner: This message can be removed.

configure.sh
makaimann: Thanks -- done in 085bf094d918953c803d8b76848996ae4178445e

makaimann:
> `BUILD_LIB_ONLY` should be initialized in CMakeLists.txt similar to `USE_PYTHON3`.

I made this change in 8d59ae83bb24d5e886f487d8a8355269c22af8ba. I also added it in the printed configuration, but don't feel strongly about it being there or not.

PR #4807, opened 2020-07-28
(proof-new) Theory preprocessor proof producing

ajreynol:
This makes the theory preprocessor proof producing in the new infrastructure, which involves updating its interfaces to TrustNode in a few places.


ajreynol:
Marking work in progress since there are other PRs that will need to be merged first.

src/theory/theory_preprocessor.cpp
```diff
@@ -24,31 +26,76 @@ namespace CVC4 {
...
+                                       RemoveTermFormulas& tfr,
+                                       ProofNodeManager* pnm)
     : d_engine(engine),
       d_logicInfo(engine.getLogicInfo()),
       d_ppCache(),
-      d_tfr(tfr)
+      d_tfr(tfr),
+      d_pfContext(),
+      d_tpg(pnm ? new TConvProofGenerator(
+                      pnm, &d_pfContext, TConvPolicy::FIXPOINT)
```
HanielB, line 37: FIXPOINT is the default there

src/theory/theory_preprocessor.cpp
```diff
@@ -24,31 +26,76 @@ namespace CVC4 {
...
+                : nullptr),
+      d_lp(pnm ? new LazyCDProof(pnm, nullptr, &d_pfContext) : nullptr)
 {
+  if (isProofEnabled())
```
HanielB, line 41: This test looks at `d_tpg`, which is set above based on `pnm`. So better to just test on pnm? Moreover, the test below of pnm being not null should always suceed?

src/theory/theory_preprocessor.cpp
```diff
@@ -24,31 +26,76 @@ namespace CVC4 {
...
   // Remove the ITEs
   Trace("te-tform-rm") << "Remove term formulas from " << ppNode << std::endl;
   TrustNode ttfr = d_tfr.run(ppNode, newLemmas, newSkolems, false);
   Trace("te-tform-rm") << "..done " << ttfr.getNode() << std::endl;
   Node retNode = ttfr.getNode();
+  if (d_tpg != nullptr)
+  {
+    if (retNode != ppNode)
+    {
+      if (ttfr.getGenerator() != nullptr)
```
HanielB, line 92: I guess no chance of putting these three nested `if` in one? :P

src/theory/theory_preprocessor.cpp
```diff
@@ -235,18 +294,70 @@ Node TheoryPreprocessor::ppTheoryRewrite(TNode term)
...
+    {
+      if (d_tpg->hasRewriteStep(term))
+      {
+        // should match
+        Assert(d_tpg->getRewriteStep(term) == termr);
+      }
+      else
+      {
+        d_tpg->addRewriteStep(term, termr, PfRule::REWRITE, {}, {term});
+      }
```
HanielB, line 323: Please invert this so that the assertion is in the `else` case

src/theory/theory_preprocessor.cpp
```diff
@@ -235,18 +294,70 @@ Node TheoryPreprocessor::ppTheoryRewrite(TNode term)
...
+    }
+  }
+  return termr;
+}
+
+Node TheoryPreprocessor::preprocessWithProof(Node term)
+{
+  // important that it is in rewritten form, to ensure that the rewrite steps
+  // recorded in d_tpg are functional.
```
HanielB, line 332: clarify

PR #4809, opened 2020-07-28
When linking Editline, use 'pkg-config' to correctly find the link-time dependencies

andrewvaughanj:
## Issue

Currently, it is not possible to compile CVC4 statically *and* with libedit, due to linking issues

## Resolution

This PR uses CMake's built-in module around `pkg-config` to find the correct link flags.

## Question

- [ ] Is it acceptable to need `pkgconfig` on macOS? Interestingly, the build instructions on macOS are very sparse at it is ...

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -85,6 +85,7 @@ jobs:
           cxxtest \
           cln \
           gmp \
+          pkgconfig \
           swig
         python3 -m pip install toml
         python3 -m pip install setuptools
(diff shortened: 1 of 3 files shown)


andrewvaughanj:
Converted to a draft -- the use of `lbsd` needs refinement.

It is needed on Ubuntu (to link statically) but isn't installed by default on CentOS (if you install `libedit-devel`).

mpreiner:
@andrewvaughanj can you please squash and sign-off your commits?

andrewvaughanj:
> @andrewvaughanj can you please squash and sign-off your commits?

Done!

I forgot I did these on the alternative VM, so it didn't sign-off.

Issue #4808, opened 2020-07-28
Build failures on older platforms due to C++17 code


andrewvaughanj:
## Issue

I was just looking at some static build issues under Ubuntu 16.04, but it seems that I cannot compile CVC4 at all:

```
../src/theory/quantifiers/sygus/sygus_interpol.cpp: In member function 'void CVC4::theory::quantifiers::SygusInterpol::getIncludeCons(const std::vector<CVC4
::NodeTemplate<true> >&, const Node&, std::map<CVC4::TypeNode, std::unordered_set<CVC4::NodeTemplate<true>, CVC4::NodeHashFunction> >&)':
../src/theory/quantifiers/sygus/sygus_interpol.cpp:130:16: error: expected unqualified-id before '[' token
     for (auto& [tn, axiomsOps] : include_cons_axioms)
                ^
../src/theory/quantifiers/sygus/sygus_interpol.cpp:130:16: error: expected ';' before '[' token
../src/theory/quantifiers/sygus/sygus_interpol.cpp:130:17: error: 'tn' was not declared in this scope
     for (auto& [tn, axiomsOps] : include_cons_axioms)
                 ^
../src/theory/quantifiers/sygus/sygus_interpol.cpp:130:21: error: 'axiomsOps' was not declared in this scope
     for (auto& [tn, axiomsOps] : include_cons_axioms)
                     ^
../src/theory/quantifiers/sygus/sygus_interpol.cpp: In lambda function:
../src/theory/quantifiers/sygus/sygus_interpol.cpp:130:32: error: expected '{' before ':' token
     for (auto& [tn, axiomsOps] : include_cons_axioms)
                                ^
../src/theory/quantifiers/sygus/sygus_interpol.cpp: In member function 'void CVC4::theory::quantifiers::SygusInterpol::getIncludeCons(const std::vector<CVC4
::NodeTemplate<true> >&, const Node&, std::map<CVC4::TypeNode, std::unordered_set<CVC4::NodeTemplate<true>, CVC4::NodeHashFunction> >&)':
../src/theory/quantifiers/sygus/sygus_interpol.cpp:130:32: error: expected ';' before ':' token
../src/theory/quantifiers/sygus/sygus_interpol.cpp:130:32: error: expected primary-expression before ':' token
../src/theory/quantifiers/sygus/sygus_interpol.cpp:130:32: error: expected ')' before ':' token
[...]

PR #4810, opened 2020-07-28
fixing issue #4808.

Ying1123:
fixing issue #4808. Remove structural binding, which only supported by c++17.


andrewvaughanj:
I've just built this on Ubuntu 16.04 (the platform that CVC4 uses for certain builds), and I can confirm that this PR builds on that platform.

PR #4811, opened 2020-07-29
Interpolation: Add implementation for SyGuS interpolation module (step4)

Ying1123:
This is the 4th step of adding interface for SyGuS Interpolation module. The whole change will be adding the API for (get-interpol s B), which is aim for computes an I that A->I and I->B. Here A is the assertions in the stack.

The 4th step finished the implementation of the interpolation module.


Ying1123:
> Looks good. Is this the right point to add the regression tests? or is that planned for a future pr?

There is another pr for link the module with smt_engine, which will be the final step.

src/theory/quantifiers/sygus/sygus_interpol.cpp
```diff
@@ -204,18 +204,122 @@ TypeNode SygusInterpol::setSynthGrammar(const TypeNode& itpGType,
...
+                                << std::endl;
+  // Fa( x ) => A( x )
+  Node firstImplication = nm->mkNode(kind::IMPLIES, Fa, itpApp);
+  Trace("sygus-interpol-debug")
+      << "first implication: " << firstImplication << std::endl
+      << std::endl;
+  // A( x ) => Fc( x )
+  Node Fc = conj;
+  Fc =
+      Fc.substitute(d_syms.begin(), d_syms.end(), d_vars.begin(), d_vars.end());
```
HanielB, line 266: Rather than two substitutions you could build the whole interpolation conjecture first and only then substitute, no? Seems more natural.

src/theory/quantifiers/sygus/sygus_interpol.cpp
```diff
@@ -204,18 +204,122 @@ TypeNode SygusInterpol::setSynthGrammar(const TypeNode& itpGType,
...
+      interpoln_reduced = interpoln[1];
+    }
+    else
+    {
+      interpoln_reduced = interpoln;
+    }
+    if (interpoln.getNumChildren() != 0 && interpoln[0].getNumChildren() != 0)
+    {
+      std::vector<Node> formals;
+      for (Node n : interpoln[0])
```
HanielB, line 307:
```suggestion
      for (const Node& n : interpoln[0])
```

src/theory/quantifiers/sygus/sygus_interpol.cpp
```diff
@@ -204,18 +204,122 @@ TypeNode SygusInterpol::setSynthGrammar(const TypeNode& itpGType,
...
 }
 
 bool SygusInterpol::findInterpol(Expr& interpol, Node itp)
 {
+  // get the synthesis solution
+  std::map<Expr, Expr> sols;
+  d_subSolver->getSynthSolutions(sols);
+  Assert(sols.size() == 1);
+  std::map<Expr, Expr>::iterator its = sols.find(itp.toExpr());
+  if (its != sols.end())
```
HanielB, line 290: You could invert this test to throw the exception here and return. Avoids the extra indentation for the large block of code.

src/theory/quantifiers/sygus/sygus_interpol.cpp
```diff
@@ -204,18 +204,122 @@ TypeNode SygusInterpol::setSynthGrammar(const TypeNode& itpGType,
...
+  {
+    Trace("sygus-interpol")
+        << "SmtEngine::getInterpol: solution is " << its->second << std::endl;
+    Node interpoln = Node::fromExpr(its->second);
```
HanielB, line 294: Add documentation to describe what is going on below.

PR #4812, opened 2020-07-30
sygus python api

yoni206:
This PR extends the python api with support for sygus functionality. 
This required the addition of a nullary constructor for `Grammar` in the c++ api.
A unit test is also included, and is a translation of the corresponding c++ api unit test.

Examples are not added yet, but are ready and planned for a next PR (in order to keep this one shorter).

### DIFF
--- a/src/api/cvc4cpp.h
+++ b/src/api/cvc4cpp.h
@@ -1955,7 +1955,7 @@ class CVC4_PUBLIC Grammar
   void addRules(Term ntSymbol, std::vector<Term> rules);
 
   /**
-   * Nullary constructor. Needed for the python api.
+   * Nullary constructor. Needed for the Cython API.
    */
   Grammar();
 
--- a/src/api/cvc4cpp.h
+++ b/src/api/cvc4cpp.h
@@ -1954,6 +1954,11 @@ class CVC4_PUBLIC Grammar
    */
   void addRules(Term ntSymbol, std::vector<Term> rules);
 
+  /**
+   * Nullary constructor. Needed for the python api.
+   */
+  Grammar();
+
  private:
   /**
    * Constructor.
--- a/src/api/python/cvc4.pxi
+++ b/src/api/python/cvc4.pxi
@@ -804,7 +804,7 @@ cdef class Solver:
         grammar.cgrammar = self.csolver.mkSygusGrammar(<const vector[c_Term]&> bvc, <const vector[c_Term]&> ntc)
         return grammar
 
-    def mkSygusVar(self, Sort sort, symbol=""):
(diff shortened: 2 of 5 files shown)


src/api/cvc4cpp.h
```diff
@@ -1954,6 +1954,11 @@ class CVC4_PUBLIC Grammar
    */
   void addRules(Term ntSymbol, std::vector<Term> rules);
 
+  /**
+   * Nullary constructor. Needed for the python api.
```
makaimann, line 1958: Super minor, but lets call this the Cython API for clarity. Technically the old SWIG python bindings haven't been deprecated yet.

src/api/python/cvc4.pxi
```diff
@@ -773,6 +793,63 @@ cdef class Solver:
...
+        cdef vector[c_Term] bvc
+        cdef vector[c_Term] ntc
+        for bv in boundVars:
+            bvc.push_back((<Term?> bv).cterm)
+        for nt in ntSymbols:
+            ntc.push_back((<Term?> nt).cterm)
+        grammar.cgrammar = self.csolver.mkSygusGrammar(<const vector[c_Term]&> bvc, <const vector[c_Term]&> ntc)
+        return grammar
+
+    def mkSygusVar(self, Sort sort, symbol=""):
```
makaimann, line 807: Also minor, but you can type the symbol as a `str`, right?

src/api/python/cvc4.pxi
```diff
@@ -773,6 +793,63 @@ cdef class Solver:
...
+        term.cterm = self.csolver.mkSygusVar(sort.csort, symbol.encode())
+        return term
+
+    def addSygusConstraint(self, Term t):
+        self.csolver.addSygusConstraint(t.cterm)
+
+    def addSygusInvConstraint(self, Term inv_f, Term pre_f, Term trans_f, Term post_f):
+        self.csolver.addSygusInvConstraint(inv_f.cterm, pre_f.cterm, trans_f.cterm, post_f.cterm)
+
+    def synthFun(self, symbol, bound_vars, Sort sort, Grammar grammar=None):
```
makaimann, line 818: Could type the symbol as a `str` if you'd like. Also, I didn't realize that you could use `None` as a default argument even with a type declaration -- very cool!

src/api/cvc4cpp.h
4tXJ7f: The old SWIG bindings are actually already removed on `master` :) I don't have a preference w.r.t. Cython vs Python API though.

makaimann: Oh great, didn't realize that! Either is fine then.

PR #4815, opened 2020-07-30
Fix expected output of a regression

### LINKED ISSUE #1649: Fix datatype-dump regression test
The following regression test is currently not passing (and disabled): https://github.com/CVC4/CVC4/blob/c6b2e085d4eb2c232a528a96e13fc7b65fd98fea/test/regress/regress0/datatypes/datatype-dump.cvc. Either the expected output needs to be fixed or the output of CVC4.

nafur:
The regression `datatypes/datatype-dump.cvc` has been disabled for two years (see #1649). However, it really seems that it works, but the output has changed a bit.
This slighty rearranges the expected output and enabled the regression.
Fixed #1649.


ajreynol:
Does the dumped regression reparse?  The output looks wrong to me, the datatype is declared after the assertion.

nafur:
As it turns out, the issue actually persists (and this PR is incorrect).

PR #4817, opened 2020-07-31
Add API interface for specialized constructor term

ajreynol:
This is an interface to get the instantiated parametric datatype constructor operator to apply when the type of this operator cannot be inferred without a cast.

This is required for eliminating the Expr-level datatype and calls to ExprManager from the parsers.


src/api/cvc4cpp.cpp
```diff
@@ -2136,6 +2136,25 @@ Term DatatypeConstructor::getConstructorTerm() const
...
+  NodeManager* nm = d_solver->getNodeManager();
+  // apply type ascription to the operator
+  Term sctor =
+      api::Term(d_solver,
+                nm->mkNode(kind::APPLY_TYPE_ASCRIPTION,
+                           nm->mkConst(AscriptionType(
+                               d_ctor
+                                   ->getSpecializedConstructorType(
+                                       retSort.getType()))),
+                           d_ctor->getConstructor()));
```
aniemetz, line 2152: Don't you have to kick off type checking for this term?

ajreynol: Yes, good catch, I've added this.

PR #4821, opened 2020-08-01
Ensure strict length constraint for decompose rule

### LINKED ISSUE #4820: (default) Performance issue on QF_S formula
~~~~
(set-logic QF_S)
(declare-fun _substvar_270_ () String)
(declare-fun _substvar_273_ () Bool)
(declare-fun _substvar_275_ () Bool)
(declare-fun _substvar_293_ () Bool)
(declare-const v0 Bool)
(declare-const v1 Bool)
(declare-const v2 Bool)
(declare-const v3 Bool)
(declare-const Str3 String)
(declare-const Str7 String)
(declare-const Str15 String)
(assert (or (xor true true (xor v0 v3 false v1 false v2 v2 v1 false v2 v1) true true (=> (or v1 (distinct Str15 _substvar_270_ Str3 (str.++ Str7 Str7)) v2) false) false true) _substvar_275_))
(check-sat)
~~~~

cvc4 0f040cb
~~~~
time z3 xx.smt2
sat

real    0m0.016s
user    0m0.012s
sys     0m0.004s

cvc4 --tlimit 30000 xx.smt2
CVC4 interrupted by timeout.
Aborted
~~~~

4tXJ7f:
Fixes #4820. The performance issue was caused by
0988217562006d3f59e01dc261f39121df6d75f5. That commit introduced an
option (`--strings-len-conc`) that optionally moves the length
constraint for skolems to the conclusion of an inference instead of
making it part of the term registration. However, for
`DEQ_DISL_STRINGS_SPLIT`, we were only asserting that `LENGTH_GEQ_ONE`
in the case where this option was not enabled, instead of asserting that
the length of the skolem is equal to the component on the other side of
the disequality. This lead to an infinite loop of inferences because we
effectively were just splitting a component into two skolems and the
only restriction was that the first one was non-empty. We guessed the
second skolem to be empty, so the first skolem was equal to the
component, the skolem was marked congruent, and we ended up with the
same normal form as before.

The commit fixes the issue by adding an argument to
`getDecomposeConclusion()` that specifies whether to add the length
constraint or not. This option is used to always add the length
constraint in the case of `DEQ_DISL_STRINGS_SPLIT`.


src/theory/strings/core_solver.cpp
```diff
@@ -2121,7 +2125,8 @@ void CoreSolver::processDeq(Node ni, Node nj)
           Node uxLen = nm->mkNode(STRING_LENGTH, ux);
           Node uyLen = nm->mkNode(STRING_LENGTH, uy);
           std::vector<Node> newSkolems;
-          Node conc = getDecomposeConclusion(ux, uyLen, false, skc, newSkolems);
+          Node conc =
+              getDecomposeConclusion(ux, uyLen, false, true, skc, newSkolems);
```
ajreynol, line 2129: I believe this should be `false, options::stringLenConc()` instead of `false, true`, right?

src/theory/strings/core_solver.cpp
```diff
@@ -2120,17 +2124,22 @@ void CoreSolver::processDeq(Node ni, Node nj)
...
+          // After step 3, `k1` is marked congruent because `x` is the older
+          // variable. So we get `x` in the normal form again.
           std::vector<Node> newSkolems;
-          Node conc = getDecomposeConclusion(ux, uyLen, false, skc, newSkolems);
+          Node conc =
+              getDecomposeConclusion(ux, uyLen, false, true, skc, newSkolems);
           Assert(newSkolems.size() == 2);
-          if (options::stringLenConc())
-          {
             d_termReg.registerTermAtomic(newSkolems[0], LENGTH_IGNORE);
```
ajreynol, line 2142: Identation is off?

src/theory/strings/core_solver.cpp
4tXJ7f: As discussed offline, I've added a clarifying comment about this.

src/theory/strings/core_solver.cpp
```diff
@@ -2120,19 +2124,24 @@ void CoreSolver::processDeq(Node ni, Node nj)
...
           Assert(newSkolems.size() == 2);
-          if (options::stringLenConc())
-          {
-            d_termReg.registerTermAtomic(newSkolems[0], LENGTH_IGNORE);
-          }
-          else
-          {
-            d_termReg.registerTermAtomic(newSkolems[0], LENGTH_GEQ_ONE);
-          }
+          d_termReg.registerTermAtomic(newSkolems[1], LENGTH_GEQ_ONE);
```
ajreynol, line 2142: I believe this rule is totally symmetric newSkolems[0] / newSkolems[1]. I think we should have the same policy for both skolems, so add `d_termReg.registerTermAtomic(newSkolems[0], LENGTH_GEQ_ONE);`.

ajreynol: Never mind, missed the fact this is within the for loop.

PR #4823, opened 2020-08-01
Add methods for constructing datatype types from NodeManager

ajreynol:
This is work towards eliminating the Expr-level datatype.

This PR implements the required methods for constructing datatype types from NodeManager.

In particular, this PR copies the "mkMutualDatatypeTypes" methods and converts them to Node-level.

The next PRs will be in preparation for using these methods instead of the Expr-level ones.

It also adds a flag `d_isRecord` to DType, which is required for supporting record printing in the cvc printer, which will be updated in another PR.

It also eliminates an interface for constructing constructor types via Expr-level DatatypeConstructor objects, which was unused.


src/expr/dtype.h
```diff
@@ -216,6 +216,9 @@ class DType
   /** set that this datatype is a tuple */
   void setTuple();
 
+  /** set that this datatype is a recird */
```
4tXJ7f, line 219:
```suggestion
  /** set that this datatype is a record */
```

src/smt/listeners.cpp
```diff
@@ -59,11 +59,15 @@ void SmtNodeManagerListener::nmNotifyNewSortConstructor(TypeNode tn,
...
-    const std::vector<DatatypeType>& dtts, uint32_t flags)
+    const std::vector<TypeNode>& dtts, uint32_t flags)
 {
   if ((flags & ExprManager::DATATYPE_FLAG_PLACEHOLDER) == 0)
   {
-    std::vector<Type> types(dtts.begin(), dtts.end());
+    std::vector<Type> types;
+    for (const TypeNode& dt : dtts)
+    {
+      types.push_back(dt.toType());
```
4tXJ7f, line 69: It would maybe be good to assert here that `dt` is a datatype type.

src/expr/node_manager.cpp
```diff
@@ -462,28 +464,132 @@ TypeNode NodeManager::mkSequenceType(TypeNode elementType)
...
+{
+  std::set<TypeNode> unresolvedTypes;
+  return mkMutualDatatypeTypes(datatypes, unresolvedTypes, flags);
+}
 
-    // should be guaranteed here already, but just in case
-    Assert(!sort.isFunctionLike());
+std::vector<TypeNode> NodeManager::mkMutualDatatypeTypes(
+    std::vector<DType>& datatypes,
+    std::set<TypeNode>& unresolvedTypes,
```
4tXJ7f, line 487:
```suggestion
    const std::vector<DType>& datatypes,
    const std::set<TypeNode>& unresolvedTypes,
```

src/expr/node_manager.h
```diff
@@ -878,8 +884,61 @@ class NodeManager {
...
+   *   DATATYPE
+   *     list = cons(car:list, cdr:list) | nil;
+   *   END;
+   *
+   * then no complicated Type needs to be created, and the above,
+   * simpler form of mkMutualDatatypeTypes() is enough.
+   */
+  std::vector<TypeNode> mkMutualDatatypeTypes(
+      std::vector<DType>& datatypes,
+      std::set<TypeNode>& unresolvedTypes,
```
4tXJ7f, line 939:
```suggestion
      const std::vector<DType>& datatypes,
      const std::set<TypeNode>& unresolvedTypes,
```

src/expr/node_manager.h
```diff
@@ -878,8 +884,61 @@ class NodeManager {
...
+
+  /** Make a type representing the given datatype. */
+  TypeNode mkDatatypeType(DType& datatype, uint32_t flags = DATATYPE_FLAG_NONE);
+
+  /**
+   * Make a set of types representing the given datatypes, which may be
+   * mutually recursive.
+   */
+  std::vector<TypeNode> mkMutualDatatypeTypes(
+      std::vector<DType>& datatypes, uint32_t flags = DATATYPE_FLAG_NONE);
```
4tXJ7f, line 906:
```suggestion
      const std::vector<DType>& datatypes, uint32_t flags = DATATYPE_FLAG_NONE);
```

PR #4824, opened 2020-08-01
Split dump manager from SmtEngine

ajreynol:
Towards splitting SmtEngine.

This moves utilities related to managing information for dumping to its own utility, DumpManager.

Its current responsibilities are to track information about how to print a model, and the implementation of some dumping traces, although its responsibilities should be extended further so that SmtEngine is not responsible for any command dumping. This is future work.


src/smt/dump_manager.h
```diff
@@ -0,0 +1,104 @@
...
+ private:
+  /** Fully inited */
+  bool d_fullyInited;
+
+  /**
+   * A list of commands that should be in the Model globally (i.e.,
+   * regardless of push/pop).  Only maintained if produce-models option
+   * is on.
+   */
+  std::vector<Command*> d_modelGlobalCommands;
```
nafur, line 78:
```suggestion
  std::vector<std::unique_ptr<Command>> d_modelGlobalCommands;
```

nafur, line 78:
```suggestion
  std::vector<std::unique_ptr<Command>> d_modelGlobalCommands;
```

src/smt/dump_manager.h
```diff
@@ -0,0 +1,104 @@
...
+   * A list of commands that should be in the Model locally (i.e.,
+   * it is context-dependent on push/pop).  Only maintained if
+   * produce-models option is on.
+   */
+  CommandList d_modelCommands;
+  /** 
+   * A list of model commands allocated to d_modelCommands at any time. This
+   * is maintained for memory management purposes.
+   */
+  std::vector<Command*> d_modelCommandsAlloc;
```
nafur, line 90:
```suggestion
  std::vector<std::unique_ptr<Command>> d_modelCommandsAlloc;
```

src/smt/dump_manager.h
```diff
@@ -0,0 +1,104 @@
...
+
+  /**
+   * A vector of declaration commands waiting to be dumped out.
+   * Once the SmtEngine is fully initialized, we'll dump them.
+   * This ensures the declarations come after the set-logic and
+   * any necessary set-option commands are dumped.
+   */
+  std::vector<Command*> d_dumpCommands;
```
nafur, line 98:
```suggestion
  std::vector<std::unique_ptr<Command>> d_dumpCommands;
```

ajreynol:
Yes, you're right about the memory management.

I'm not really sure why this was done in a custom way before, I assumed it wasn't possible, but I think your suggestion works.

PR #4825, opened 2020-08-02
Updates to dtype constructor in preparation for eliminating Expr-level datatype

ajreynol:


src/expr/dtype_cons.h
```diff
@@ -195,6 +206,12 @@ class DTypeConstructor
    *   stoa(T,C,index)
    */
   int getSelectorIndexInternal(Node sel) const;
+  /** get selector index for name
+   *
+   * Returns the index of selector with the given name, or -1 if it
+   * does not exist.
+   */
+  int getSelectorIndexForName(const std::string& name) const;
```
4tXJ7f, line 214: Hmm, maybe we should make this an `ssize_t`?

ajreynol: I would prefer `int` right now since `getSelectorIndexInternal` (an analogous method) returns `int`.  Both these could be refactored in the future but I prefer they are synchronized.

PR #4837, opened 2020-08-03
Set isClosed argument to false in ProofEqEngine::assertFact 

alex-ozdemir:
It's supposed to take open proofs.

Question: Why not have `ProofEqEngine::assertFact` **always** pass `isClosed=false` to `addLazyStep`? Doesn't the interface of `ProofEqEngine::assertFact` require that the proof is open?


ajreynol:
Discussed offline, should simplify based on @alex-ozdemir 's suggestion above.

PR #4838, opened 2020-08-03
Updates not related to creation for eliminating Expr-level datatype

ajreynol:
This updates remaining uses of the Expr-level Datatype that are not related to their creation / memory management in ExprManager.

This required updating a unit test from Expr -> Node.


src/expr/symbol_table.cpp
```diff
@@ -212,11 +213,10 @@ Expr OverloadedTypeTrie::getOverloadedFunctionForTypes(
           Trace("parser-overloading")
               << "Parametric overloaded datatype selector " << name << " "
               << tna << std::endl;
-          DatatypeType tnd = static_cast<DatatypeType>(argTypes[i]);
-          const Datatype& dt = tnd.getDatatype();
+          const DType& dt = TypeNode::fromType(argTypes[i]).getDType();
           // tng is the "generalized" version of the instantiated parametric
           // type tna
-          Type tng = dt.getDatatypeType();
+          Type tng = dt.getTypeNode().toType();
```
4tXJ7f, line 219: I think that we will ultimately want to convert `SymbolTable` to be a term-level class (and move it to src/parser) but this is fine for now.

test/unit/util/datatype_black.h
```diff
@@ -46,172 +46,188 @@ class DatatypeBlack : public CxxTest::TestSuite {
...
-    leaf.addArg("leaf", integerType);
+    std::shared_ptr<DTypeConstructor> leaf =
+        std::make_shared<DTypeConstructor>("leaf");
+    leaf->addArg("leaf", integerType);
     tree.addConstructor(leaf);
 
     Debug("datatypes") << tree << std::endl;
-    DatatypeType treeType = d_em->mkDatatypeType(tree);
+    TypeNode treeType = d_nm->mkDatatypeType(tree);
     Debug("datatypes") << treeType << std::endl;
```
4tXJ7f, line 134: What about the tests for `getConstructor()`? Don't we have something similar at the Node level?

src/expr/symbol_table.cpp
ajreynol: Yes that's right.  This PR concerns eliminating Expr-level Datatype and not Expr-level Type, and thus is out of scope in this PR.

test/unit/util/datatype_black.h
ajreynol: We don't have any equivalent tests for constructor.  We have a DTypeConstructor, although its tests are grouped with DType in this unit test.

PR #4842, opened 2020-08-04
(proof-new) Callbacks for term-context-sensitive terms

ajreynol:
Callbacks for term-context-sensitive terms.

It gives two examples of such callbacks, "remove term formula (rtf) context" to be used in the remove term formulas pass and "polarity context" which is a highly common pattern in preprocessing.

These utilities will be critical for maintain proofs for term-context-sensitive rewrites.  An example user of these utilities is TermFormulaRemoval, which rewrites terms depending on whether they occur beneath quantifiers/terms.  This utility currently has bugs in its proof generation since its proof cache does not consider the term context.  This PR updates that class to use this utility.  Its proof generator will be similarly extended in `proof-new` to synchronize its cache, relative to term context identifiers.  Concretely, a TermContext callback will be passed to TConvProofGenerator, which will put together proof skeletons in a term-context-sensitive manner.


HanielB:
What is the plan in using this in terms of CDProof? Or is it not gonna be relevant for it? I ask because we could maybe avoid caching modulo bound variables and do tree traversal under binders.

ajreynol:
CDProof doesn't have a notion of traversing terms, so this won't be relevant for it.  On the proof side, the main purpose of these utilities are to generalize TConvProofGenerator.

src/expr/term_context.h
```diff
@@ -0,0 +1,143 @@
...
+ * Computes whether we are inside a term (as opposed to being part of Boolean
+ * skeleton) and whether we are inside a quantifier. For example, for:
+ *   (and (= a b) (forall ((x Int)) (P x)))
+ * we have the following mappings (term -> inTerm,inQuant)
+ *   (= a b) @ [0] -> false, false
+ *   a @ [0,1] -> true, false
+ *   (P x) @ [1,1] -> false, true
+ *    x @ [1,1,0] -> true, true
+ * Notice that the hash of a child can be computed from the parent's hash only,
+ * and hence this can be implemented without considering paths.
```
HanielB, line 65: It's a bit odd to have this here when it's already said in the abstract class. Makes you wonder whether the abstract class' claim that contexts are always locals may not hold. So I'd rather remove this comment.

src/expr/term_context.h
```diff
@@ -0,0 +1,143 @@
...
+ */
+class RtfTermContext : public TermContext
+{
+ public:
+  RtfTermContext() {}
+  /** The initial value: not in a term context or beneath a quantifier. */
+  uint32_t initialValue() const override;
+  /** Compute the value */
```
HanielB, line 73: Say abstractly what is the pattern of the values

src/expr/term_context.h
```diff
@@ -0,0 +1,143 @@
...
+  uint32_t computeValue(TNode t, uint32_t tval, size_t index) const override;
+  /** get hash value from the flags */
+  static uint32_t getValue(bool inQuant, bool inTerm);
+  /** get flags from the hash value */
+  static void getFlags(uint32_t val, bool& inQuant, bool& inTerm);
+
+ private:
+  /**
+   * Returns true if the children of t should be considered in a "term" context.
```
HanielB, line 82: Given the hard-coded way in which this is computed it'd be helpful to clarify the policy here for the case that somebody ever needs to change this.

src/expr/term_context.h
```diff
@@ -0,0 +1,143 @@
...
+ * We have that the P at path [0] has polarity true and the P at path [1,0] has
+ * polarity false.
+ *
+ * Finally, notice that polarity does not correspond to an entailed value. Thus,
+ * for the formula:
+ *   (or P Q)
+ * we have that
+ *   P @ [0] -> true
+ *   Q @ [1] -> true
+ * although neither is entailed.
```
HanielB, line 117: Not sure what this is supposed to convey. What would be an "entailed value"?

src/expr/term_context.h
```diff
@@ -0,0 +1,143 @@
...
+ *
+ * Notice that the hash of a child can be computed from the parent's hash only,
+ * and hence this can be implemented without considering paths.
```
HanielB, line 120: ditto with the comment on Rtf context

src/expr/term_context.h
```diff
@@ -0,0 +1,143 @@
...
+ */
+class PolarityTermContext : public TermContext
+{
+ public:
+  PolarityTermContext() {}
+  /** The initial value: true polarity. */
+  uint32_t initialValue() const override;
+  /** Compute the value */
```
HanielB, line 128: ditto with the computeValue comment on RtF context

src/expr/term_context.h
ajreynol: A value that the formula entails.

src/expr/term_context.h
ajreynol: What do you mean?

src/expr/term_context.h
ajreynol: I've updated the comment.

src/expr/term_context.h
HanielB: The value is computed in a given way, so that for a context like this you get a value like that. I think it'd be interesting to have an idea of that from this point.

src/expr/term_context.h
HanielB: I see. I think this more confuses than clarifies things if one does not expect polarity to correspond to entailed values.

ajreynol: This is an important distinction, there will be another variant of this class called EntailedPolarityTermContext. This comment is in preparation for that class, where this distinction will be important.

src/expr/term_context.h
ajreynol: That's what the comment at the top of the class is for.  The user should look there instead.  

Notice that more generally any method marked `override` should not need significant documentation.

src/expr/term_context.h
HanielB: I see.

src/expr/term_context.h
HanielB: Sorry, I don't mean how the polarity is determined but what is the structure of the hash.

PR #4843, opened 2020-08-04
GH Actions: Remove cancel action.

aniemetz:
The previously introduced action to cancel running builds is not able to
cancel builds that don't run on the same fork. As a consequence,
when pushing to a branch for which a PR has been submitted, builds on the
main repository are not cancelled.

This removes the cancel build. If we want behavior similar to how it was
on Travis, we need a workaround / more sophisticated solution since GH
Actions doesn't really allow / support this (due to permission issues).


aniemetz:
I guess, yes.

PR #4848, opened 2020-08-05
Improve rewriting of str.<=

4tXJ7f:
This commit improves our rewriting of `str.<=` slightly. If we have
constant prefixes that are different, we can always rewrite the term to
a constant. Previously, we were only doing so when the result was
`false`.


ajreynol:
Is there a regression for this?

4tXJ7f:
Added a regression and a unit test for the rewrite.

ajreynol (approved):

PR #4854, opened 2020-08-05
Split preprocessor from SmtEngine

ajreynol:
This splits a collection of utilities from SmtEngine that work in cooperation to preprocess assertions (Boolean circuit propagator, preprocessing context, process assertions, term formula removal).

It updates various interfaces in SmtEngine from Expr -> Node and simplifies SmtEngine to use this utility.

### DIFF
--- a/src/smt/preprocessor.h
+++ b/src/smt/preprocessor.h
@@ -82,7 +82,7 @@ class Preprocessor
   Node simplify(const Node& e, bool removeItes = false);
   /**
    * Expand the definitions in a term or formula e.  No other
-   * simplification or normalization is done. 
+   * simplification or normalization is done.
    *
    * @param e The node to expand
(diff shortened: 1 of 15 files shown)


src/smt/preprocessor.h
```diff
@@ -0,0 +1,138 @@
...
+   *
+   * @param e The node to expand
+   * @param expandOnly if true, then the expandDefinitions function of
+   * TheoryEngine is not called on subterms of e.
+   * @return The expanded term.
+   */
+  Node expandDefinitions(const Node& e, bool expandOnly = false);
+  /** Same as above, with a cache of previous results. */
+  Node expandDefinitions(
+      const Node& e,
```
nafur, line 95:
```suggestion
      const Node& n,
```

PR #4855, opened 2020-08-06
Split SmtEngineState from SmtEngine

ajreynol:
This splits a utility for tracking the "basic" state of the SmtEngine.  This class tracks its high-level state, including the "SMT mode", last/expected status and manages the contexts.  It is *not* responsible more detailed state information (e.g. the assertions).

Marking as complex since this is a fairly critical change.


src/smt/smt_engine_state.h
```diff
@@ -0,0 +1,240 @@
...
+  /**
+   * Cleanup, which pops all contexts to level zero.
+   */
+  void cleanup();
+  /**
+   * Do all pending pops, which ensures that the context levels are up-to-date.
+   * This method should be called by the SmtEngine before using any of its
+   * members that rely on the context (e.g. PropEngine or TheoryEngine).
+   */
+  void doPendingPops();
```
4tXJ7f, line 111: This should be moved down to `context management`, I think.

src/smt/smt_engine_state.h
```diff
@@ -0,0 +1,240 @@
...
+   */
+  bool isQueryMade() const;
+  /** Return the user context level.  */
+  size_t getNumUserLevels() const;
+  /** Get the status of the last check-sat */
+  Result getStatus() const;
+  /** Get the SMT mode we are in */
+  SmtMode getMode() const;
+  /** return the input name (if any) */
+  std::string getFilename() const;
```
4tXJ7f, line 160: To avoid copying the string when someone asks for it:

```suggestion
  const std::string& getFilename() const;
```

src/smt/smt_engine_state.h
```diff
@@ -0,0 +1,240 @@
...
+   * Do all pending pops, which ensures that the context levels are up-to-date.
+   * This method should be called by the SmtEngine before using any of its
+   * members that rely on the context (e.g. PropEngine or TheoryEngine).
+   */
+  void doPendingPops();
+  /**
+   * Set that the file name of the current instance is the given string. This
+   * is used for various purposes (e.g. get-info, SZS status).
+   */
+  void setFilename(std::string filename);
```
4tXJ7f, line 116:
```suggestion
  void setFilename(const std::string& filename);
```

src/smt/smt_engine_state.h
```diff
@@ -0,0 +1,240 @@
...
+class SmtEngineState
+{
+ public:
+  SmtEngineState(SmtEngine& smt);
+  ~SmtEngineState() {}
+  /**
+   * Notify that the expected status of the next check-sat is given by the
+   * string status, which should be one of "sat", "unsat" or "unknown".
+   */
+  void notifyExpectedStatus(std::string status);
```
4tXJ7f, line 56:
```suggestion
  void notifyExpectedStatus(const std::string& status);
```

src/smt/smt_mode.h
```diff
@@ -0,0 +1,57 @@
...
+ ** This file is part of the CVC4 project.
+ ** Copyright (c) 2009-2020 by the authors listed in the file AUTHORS
+ ** in the top-level source directory) and their institutional affiliations.
+ ** All rights reserved.  See the file COPYING in the top-level source
+ ** directory for licensing information.\endverbatim
+ **
+ ** \brief Enumeration type for the mode of an SmtEngine.
+ **/
+
+#include "cvc4_public.h"
```
4tXJ7f, line 15: Can we make this private already or does this break things?

src/smt/smt_engine.cpp
```diff
@@ -919,18 +923,17 @@ bool SmtEngine::isDefinedFunction( Expr func ){
...
 
 Result SmtEngine::check() {
-  Assert(d_fullyInited);
-  Assert(d_pendingPops == 0);
+  Assert(d_state->isFullyReady());
 
   Trace("smt") << "SmtEngine::check()" << endl;
 
-
+  std::string filename = d_state->getFilename();
```
4tXJ7f, line 930:
```suggestion
  const std::string& filename = d_state->getFilename();
```

src/smt/smt_engine.cpp
```diff
@@ -950,14 +953,15 @@ Result SmtEngine::check() {
...
 
-  return Result(result, d_filename);
+  return Result(result, filename);
 }
 
 Result SmtEngine::quickCheck() {
-  Assert(d_fullyInited);
+  Assert(d_state->isFullyInited());
   Trace("smt") << "SMT quickCheck()" << endl;
+  std::string filename = d_state->getFilename();
```
4tXJ7f, line 962:
```suggestion
  const std::string& filename = d_state->getFilename();
```

src/smt/smt_engine.cpp
```diff
@@ -1168,37 +1172,8 @@ Result SmtEngine::checkSatisfiability(const vector<Node>& assumptions,
...
-    }
-    else if (d_status.asSatisfiabilityResult().isSat() == Result::SAT)
-    {
-      d_smtMode = SMT_MODE_SAT;
-    }
-    else
-    {
-      d_smtMode = SMT_MODE_SAT_UNKNOWN;
-    }
+    // update our state the based on the result of the check-sat
```
4tXJ7f, line 1175: The sentence doesn't parse for me. How about:

```suggestion
    // notify our state of the check-sat result
```

src/smt/smt_engine.cpp
```diff
@@ -1230,7 +1205,7 @@ Result SmtEngine::checkSatisfiability(const vector<Node>& assumptions,
     Result::UnknownExplanation why = d_resourceManager->outOfResources()
                                          ? Result::RESOURCEOUT
                                          : Result::TIMEOUT;
-    return Result(Result::SAT_UNKNOWN, why, d_filename);
+    return Result(Result::SAT_UNKNOWN, why, d_state->getFilename());
```
4tXJ7f, line 1208: Don't we also need to inform our state about this result?

src/smt/smt_engine.cpp
```diff
@@ -767,7 +771,7 @@ void SmtEngine::defineFunction(Expr func,
 {
   SmtScope smts(this);
   finalOptionsAreSet();
```
4tXJ7f, line 773: Could the state handle calling `finalOptionsAreSet()`? I think that it could be argued that the state determines when the final options must be set.

src/smt/smt_engine_state.h
```diff
@@ -0,0 +1,240 @@
...
+  std::string getFilename() const;
+  //---------------------------- end queries
+
+ private:
+  /** Pushes the user and SAT contexts */
+  void push();
+  /** Pops the user and SAT contexts */
+  void pop();
+  /** Pops the user and SAT contexts to the given level */
+  void popto(int toLevel);
```
4tXJ7f, line 169: How about:

```suggestion
  void popTo(size_t toLevel);
```

(to mirror the return type of `getNumUserLevels()`)

src/smt/smt_engine_state.h
```diff
@@ -0,0 +1,240 @@
...
+  /**
+   * The expected status of the next satisfiability check.
+   */
+  Result d_expectedStatus;
+
+  /** The SMT mode, for details see enumeration above. */
+  SmtMode d_smtMode;
+
+  /**
+   * The input file name (if any) or the name set through setInfo (if any)
```
4tXJ7f, line 232:
```suggestion
   * The input file name (if any)
```

src/smt/smt_engine_state.h
```diff
@@ -0,0 +1,240 @@
...
+   * @param hasAssumptions Whether the prior call to check-sat had assumptions.
+   * If so, we pop a context.
+   * @param r The result of the check-sat call.
+   */
+  void notifyCheckSatResult(bool hasAssumptions, Result r);
+  /**
+   * Notify that we finished an abduction query, where success is whether the
+   * command was successful. This is managed independently of the above
+   * calls for notifying check-sat. In other words, if a get-abduct command
+   * is issued to an SmtEngine, it may use a satsisfiability call (if desired)
```
4tXJ7f, line 87:
```suggestion
   * is issued to an SmtEngine, it may use a satisfiability call (if desired)
```

src/smt/smt_engine_state.h
ajreynol: This is a wrapper method to Context::popto which takes `int` unfortunately, I believe we should keep it like this until that method is change.

src/smt/smt_engine_state.h
ajreynol: setInfo still sets this field actually.

src/smt/smt_mode.h
ajreynol: I tried :) 

Unfortunately it breaks things (spent an hour trying to fix it but couldn't).  I think we really should make SmtEngine private sooner than later, which would make it easy to make this private.

src/smt/smt_engine.cpp
ajreynol: Yeah I think this is a good suggestion, I'll make a few more changes.

src/smt/smt_engine.cpp
ajreynol: No. SmtEngineState::notifyCheckSatResult() is for cases where the result was computed and the solver should continue as normal.  I would recommend a separate callback for this case, e.g. SmtEngineState::notifyInterupt(), which however would do nothing at the moment (which is why I haven't added it).  I'll add a comment about this.

Regardless this is making it much more clear to me what we would need to do to implement safe interupts.

src/smt/smt_engine.cpp
```diff
@@ -2926,10 +2805,7 @@ void SmtEngine::setSygusConjectureStale()
     return;
   }
   d_private->d_sygusConjectureStale = true;
-  if (options::incrementalSolving())
-  {
-    internalPop();
-  }
+  // TODO (project #7): if incremental, we should pop a context
```
nafur, line 2808: Ditto.

src/smt/smt_engine.cpp
```diff
@@ -2541,8 +2507,8 @@ void SmtEngine::getInstantiationTermVectors( Expr q, std::vector< std::vector< E
 
 vector<Expr> SmtEngine::getAssertions() {
```
nafur, line 2508:
```suggestion
std::vector<Expr> SmtEngine::getAssertions() {
```

src/smt/smt_engine.cpp
```diff
@@ -700,9 +704,9 @@ CVC4::SExpr SmtEngine::getInfo(const std::string& key) const
   }
   if (key == "assertion-stack-levels")
   {
-    AlwaysAssert(d_userLevels.size()
-                 <= std::numeric_limits<unsigned long int>::max());
-    return SExpr(static_cast<unsigned long int>(d_userLevels.size()));
+    size_t ulevel = d_state->getNumUserLevels();
```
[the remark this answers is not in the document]
ajreynol: While I do prefer having `std::` on most things, I don't follow this convention, I don't believe its necessary in general to put std:: on basic types.

src/smt/smt_engine.cpp
```diff
@@ -1489,17 +1467,8 @@ Result SmtEngine::checkSynth()
...
-      // later
-      internalPush();
-      assertFormula(body, true);
-    }
-    else
-    {
-      query = body.toExpr();
-    }
+    // TODO (project #7): if incremental, we should push a context and assert
+    query = body.toExpr();
```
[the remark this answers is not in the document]
ajreynol: It was implemented at a placeholder but not tested (and now abandoned since we hit other snags when trying to do SyGuS+incremental).  The implementation was not ideal since it was making a low level call to `internalPush` which is now a private method of SmtEngineState.  If we implement this in the future, we should figure out how to do this cleanly.

src/smt/smt_engine.cpp
ajreynol: Responded on other comment.

src/smt/smt_engine_state.cpp
```diff
@@ -0,0 +1,302 @@
...
+      d_needPostsolve(false),
+      d_status(),
+      d_expectedStatus(),
+      d_smtMode(SmtMode::START)
+{
+}
+
+void SmtEngineState::notifyExpectedStatus(const std::string& status)
+{
+  Assert(status == "sat" || status == "unsat" || status == "unknown")
```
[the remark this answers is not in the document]
ajreynol: Yes I think so.  Notice this is an `Assert` so it only happens in debug mode.  Essentially this level of debugging makes the SmtEngineState defensive as a standalone utility, in case SmtEngine or other classes have further uses of it.

src/smt/smt_engine_state.cpp
```diff
@@ -0,0 +1,302 @@
...
+  // push if there are assumptions
+  if (hasAssumptions)
+  {
+    internalPush();
+  }
+}
+
+void SmtEngineState::notifyCheckSatResult(bool hasAssumptions, Result r)
+{
+  d_needPostsolve = true;
```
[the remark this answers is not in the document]
ajreynol: That comment should be updated, will do.

The SmtEngine is responsible for making the actual calls to TheoryEngine.  `presolve` is now independent of the concern of the state since SmtEngine does it alone.

src/smt/smt_engine_state.h
```diff
@@ -0,0 +1,245 @@
...
+   * @param r The result of the check-sat call.
+   */
+  void notifyCheckSatResult(bool hasAssumptions, Result r);
+  /**
+   * Notify that we finished an abduction query, where success is whether the
+   * command was successful. This is managed independently of the above
+   * calls for notifying check-sat. In other words, if a get-abduct command
+   * is issued to an SmtEngine, it may use a satisfiability call (if desired)
+   * to solve the abduction query. This method is called *in addition* to
+   * the above calls to notifyCheckSat / notifyCheckSatResult in this case.
```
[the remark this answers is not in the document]
ajreynol: Yes, I'll add.

src/smt/smt_engine_state.h
```diff
@@ -0,0 +1,245 @@
...
+  void finishInit();
+  /**
+   * Prepare for a shutdown of the SmtEngine, which does pending pops and
+   * pops the user context to zero.
+   */
+  void shutdown();
+  /**
+   * Cleanup, which pops all contexts to level zero.
+   */
+  void cleanup();
```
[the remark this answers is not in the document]
ajreynol: We may be able to simplify this, but I'm being paranoid for now to ensure the same behavior as before.

We should generally revisit the entire shutdown procedure in SmtEngine to see if things can be refactored once SmtEngine is more stable.

src/smt/smt_engine_state.h
```diff
@@ -0,0 +1,245 @@
...
+   * pending pops.
+   */
+  bool isFullyReady() const;
+  /**
+   * Return true if a notifyCheckSat call has been made, e.g. a query has been
+   * issued to the SmtEngine.
+   */
+  bool isQueryMade() const;
+  /** Return the user context level.  */
+  size_t getNumUserLevels() const;
```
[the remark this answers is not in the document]
ajreynol: Ditto about not using std:: for basic types.

src/smt/smt_mode.h
```diff
@@ -0,0 +1,57 @@
...
+#include <iosfwd>
+
+namespace CVC4 {
+
+/**
+ * The mode of the solver, which is an extension of Figure 4.1 on
+ * page 52 of the SMT-LIB version 2.6 standard
+ * http://smtlib.cs.uiowa.edu/papers/smt-lib-reference-v2.6-r2017-07-18.pdf
+ */
+enum class SmtMode : uint32_t
```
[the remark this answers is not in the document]
ajreynol: @4tXJ7f may know?

src/smt/smt_engine.cpp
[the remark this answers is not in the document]
nafur: I usually avoid `using namespace std;` altogether, which implies using `std::` everywhere. But I see that our guidelines only discourage `using namespace` in header files...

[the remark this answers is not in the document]
ajreynol: Right, I agree with you about not using `std`. It's just that my impression is that `size_t` is low-level enough that it doesn't warrant having std:: on it. (I actually thought that `size_t` can be used without being in the `std` namespace).

src/smt/smt_mode.h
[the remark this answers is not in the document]
4tXJ7f: I guess as long as we are only using it as a pure `enum` and are not converting it to an integer, it is fine to leave out the type.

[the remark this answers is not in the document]
ajreynol: I'll remove.

src/smt/smt_engine.cpp
```diff
@@ -2670,7 +2669,7 @@ void SmtEngine::setUserAttribute(const std::string& attr,
   SmtScope smts(this);
   finishInit();
   std::vector<Node> node_values;
-  for( unsigned i=0; i<expr_values.size(); i++ ){
+  for( std::size_t i=0, n = expr_values.size(); i<n; i++ ){
```
4tXJ7f, line 2672: Apply ClangFormat.

src/smt/smt_engine_state.h
4tXJ7f: Right, but the comment suggests that `setInfo` is a method in this class which it is not as far as I can tell. Maybe write `or the name set through (set-info :filename ...)`?

src/smt/smt_engine.cpp
4tXJ7f: Hmm, I think that this breaks doing `(get-info :reason-unknown)` after a timeout, no? Maybe this is a bit out of scope for this PR though. We probably need a separate state `INVALID` until our interrupts are safe.

ajreynol: Sure, note this PR didnt change this behavior.

PR #4856, opened 2020-08-06
Refactoring the rewriter of sets

ajreynol:
Changes it so that we don't flatten unions if at least one child is non-constant, since this may lead to children that are non-constant by mixing constant/non-constant elements and is generally expensive for large unions of singleton elements.  

The previous rewriting policy was causing an incorrect model in a separation logic benchmark reported by Andrew Jones, due to unions of constant elements that were unsorted (and hence not considered constants).  We now have the invariant that all subterms that are unions of constant elements are set constants.

Note this PR changes the normal form of set constants to be (union (singleton c1) ... (union (singleton cn-1) (singleton cn) ... ) not (union ... (union (singleton c1) (singleton c2)) ... (singleton cn)). 

It changes a unit test which was impacted by this change which was failing due to hardcoding the enumeration order in the unit test. The test is now agnostic to the order of elements.


src/theory/sets/normal_form.h
```diff
@@ -56,76 +71,78 @@ class NormalForm {
...
+        }
         Debug("sets-checknormal")
-            << "[sets-checknormal]              element = " << n[1][0] << " "
-            << n[1][0].getId() << std::endl;
-        if (n[1][0] >= prvs) return false;
-        prvs = n[1][0];
-        n = n[0];
+            << "[sets-checknormal]              element = " << n[0][0] << " "
+            << n[0][0].getId() << std::endl;
+        if (!prvs.isNull() && n[0][0] >= prvs)
```
mudathirmahgoub, line 91: According to the documentation above, I guess this line should be 
```
if (!prvs.isNull() && n[0][0] < prvs)```

File: src/theory/sets/normal_form.h [hunk 2]
```diff
@@ -56,76 +71,78 @@ class NormalForm {
...
           << n[0].getId() << std::endl;
-      if (n[0] >= prvs) return false;
-
-      // we made it
-      return true;
-
-    } else {
-      return false;
+      // compare last ID
+      if (n[0] < prvs)
```
mudathirmahgoub, line 113: Similarly, this line should be 
```
if (n[0] >= prvs)
```

src/theory/sets/theory_sets_private.cpp
```diff
@@ -1292,6 +1292,7 @@ bool TheorySetsPrivate::collectModelInfo(TheoryModel* m)
     else
     {
       std::vector<Node> els;
+      std::set<Node> elsSet;
```
mudathirmahgoub, line 1295: Why is this local variable declared here? It is unused.

src/theory/sets/theory_sets_rewriter.cpp
```diff
@@ -38,12 +38,11 @@ bool checkConstantMembership(TNode elementTerm, TNode setTerm)
...
   Assert(setTerm.getKind() == kind::UNION
-         && setTerm[1].getKind() == kind::SINGLETON)
+         && setTerm[0].getKind() == kind::SINGLETON)
       << "kind was " << setTerm.getKind() << ", term: " << setTerm;
 
-  return
-    elementTerm == setTerm[1][0] ||
-    checkConstantMembership(elementTerm, setTerm[0]);
+  return elementTerm == setTerm[0][0]
+         || checkConstantMembership(elementTerm, setTerm[1]);
```
mudathirmahgoub, line 45: It would be nice if `checkConstantMembership` is added to the header file as a private member.

src/theory/sets/normal_form.h
ajreynol: Yes, you're right, I've changed the doc.

src/theory/sets/normal_form.h
ajreynol: Same.

src/theory/sets/theory_sets_private.cpp
ajreynol: Will delete.

src/theory/sets/theory_sets_rewriter.cpp
ajreynol: Done

src/theory/sets/theory_sets_rewriter.cpp
```diff
@@ -163,23 +164,13 @@ RewriteResponse TheorySetsRewriter::postRewrite(TNode node) {
...
-      std::sort( els.begin(), els.end() );
-      Node rew = NormalForm::mkBop( kind::INTERSECTION, els, node.getType() );
-      if( rew!=node ){
-        Trace("sets-rewrite") << "Sets::rewrite " << node << " -> " << rew << std::endl;
-      }
-      return RewriteResponse(REWRITE_DONE, rew);
     }
-    /*
-    } else if (node[0] > node[1]) {
+    else if (node[0] > node[1])
```
mudathirmahgoub, line 168: Following the new documentation, shouldn't this be `node[0] < node[1]`.

src/theory/sets/theory_sets_rewriter.cpp
```diff
@@ -200,19 +191,16 @@ RewriteResponse TheorySetsRewriter::postRewrite(TNode node) {
...
-      NormalForm::getElementsFromBop( kind::UNION, node, els );
-      std::sort( els.begin(), els.end() );
-      Node rew = NormalForm::mkBop( kind::UNION, els, node.getType() );
-      if( rew!=node ){
-        Trace("sets-rewrite") << "Sets::rewrite " << node << " -> " << rew << std::endl;
-      }
-      Trace("sets-rewrite") << "...no rewrite." << std::endl;
-      return RewriteResponse(REWRITE_DONE, rew);
     }
+    else if (node[0] > node[1])
```
mudathirmahgoub, line 198: same here

src/theory/sets/theory_sets_rewriter.cpp
ajreynol: This is unrelated to the new documentation.  This concerns non-constant nodes, whereas the documentation concerns constant nodes.

test/unit/theory/theory_sets_type_enumerator_white.h
```diff
@@ -49,32 +49,35 @@ class SetEnumeratorWhite : public CxxTest::TestSuite
...
     SetEnumerator setEnumerator(d_nm->mkSetType(boolType));
     TS_ASSERT(!setEnumerator.isFinished());
 
+    std::vector<Node> elems;
+
     Node actual0 = *setEnumerator;
-    Node expected0 =
-        d_nm->mkConst(EmptySet(d_nm->mkSetType(boolType)));
-    TS_ASSERT_EQUALS(expected0, actual0);
+    addAndCheckUnique(actual0, elems);
```
mudathirmahgoub, line 68: I guess this is way simpler test than the one I wrote before.

ajreynol: Yes, in general, it can be a bit annoying to have unit tests that are hardcoded to the current behavior for something that can change in benign ways like the type enumerator.

mudathirmahgoub (approved):
Thanks for fixing the issue and removing `getElementsFromBop`.

PR #4859, opened 2020-08-07
Move datatype index constant to its own file

ajreynol:
Towards removing the Expr-level datatype.

Moves DatatypeIndex to its own file, which is the only thing that is necessary remaining from `expr/datatype.h`. 

Also updates the datatype kinds file in preparation for the removal.


src/expr/datatype_index.h
```diff
@@ -0,0 +1,71 @@
...
+ **
+ ** \brief A class representing an index to a datatype living in NodeManager.
+ **/
+
+#include "cvc4_public.h"
+
+#ifndef CVC4__DATATYPE_INDEX_H
+#define CVC4__DATATYPE_INDEX_H
+
+#include <iostream>
```
4tXJ7f, line 20:
```suggestion
#include <iosfwd>
```

PR #4863, opened 2020-08-09
Final preparations for changing API to use the Node-level datatype

ajreynol:
This includes all fixes encountered while fixing unit tests with the Term -> Node version of Datatypes in the API.

After all pending PRs are merged, the next step will be to convert the new API to use e.g. CVC4::DType instead of CVC4::Datatype everywhere.


src/expr/dtype.h
```diff
@@ -191,6 +190,31 @@ class DType
    * be unique; they are for convenience and pretty-printing only.
    */
   void addConstructor(std::shared_ptr<DTypeConstructor> c);
+  /** add sygus constructor
+   *
+   * This adds a sygus constructor to this datatype, where
+   * this datatype should be currently unresolved. Note this method is
+   * syntax sugar for adding a normal constructor and setting it to be a
```
4tXJ7f, line 197:
```suggestion
   * syntactic sugar for adding a normal constructor and setting it to be a
```

PR #4864, opened 2020-08-09
Make valuation class more robust to null underlying TheoryEngine.

ajreynol:
In some use cases (unit tests, old proofs infrastructure), we use a Theory with no associated TheoryEngine. This PR makes the Valuation class more robust to this case.

This includes making the "unevaluated kinds" a no-op in this case (this is necessary for Theory::finishInit with no TheoryEngine) and adding some assertions to cases that the Theory should never call without TheoryEngine.

This is required for a new policy for dynamically configuring equality engine infrastructure in Theory.


src/theory/valuation.cpp
```diff
@@ -102,36 +106,69 @@ bool Valuation::hasSatValue(TNode n, bool& value) const {
...
 
 Node Valuation::getModelValue(TNode var) {
+  Assert(d_engine != nullptr);
   return d_engine->getModelValue(var);
 }
 
 TheoryModel* Valuation::getModel() {
+  if (d_engine == nullptr)
+  {
+    // no theory engine, this we don't have a model object
```
4tXJ7f, line 121:
```suggestion
    // no theory engine, thus we don't have a model object
```

PR #4867, opened 2020-08-10
Add the distributed equality engine manager

ajreynol:
This is the first step towards making the approach for theory combination in CVC4 configurable.

This PR introduces the "distributed equality engine manager", which encapsulates the current policy that TheoryEngine uses regarding equality engines: each Theory creates a separate copy of an equality engine.  The eventual plan is that the official equality engine of Theory is *not* necessarily unique to the theory, if equality engines are shared.  A variant of this class could implement that policy.

This PR does not impact the code, it simply adds the definition of the class.  A forthcoming PR will integrate this class into TheoryEngine, which will use dynamic allocation for equality engine objects.

FYI @barrettcw


src/theory/ee_manager_distributed.h
```diff
@@ -0,0 +1,147 @@
...
+ public:
+  virtual ~EqEngineManager() {}
+  /**
+   * Get the equality engine theory information for theory with the given id.
+   */
+  const EeTheoryInfo* getEeTheoryInfo(TheoryId tid) const;
+
+ protected:
+  /** Information related to the equality engine, per theory. */
+  std::map<TheoryId, EeTheoryInfo> d_einfo;
```
4tXJ7f, line 60: I think we generally use an array to map `TheoryId` to some object (e.g. [here](https://github.com/CVC4/CVC4/blob/master/src/theory/theory_engine.h#L134)). Maybe we should do the same here?

src/theory/ee_manager_distributed.h
```diff
@@ -0,0 +1,147 @@
...
+};
+
+/**
+ * The (distributed) equality engine manager. This encapsulates the current
+ * architecture for managing equalities in the TheoryEngine, in which all
+ * theories maintain their own copy of an equality engine.
```
4tXJ7f, line 66: This comment will be outdated in the future. Maybe say something like:

> The (distributed) equality engine manager. This encapsulates an architecture, in which all theories maintain their own copy of an equality engine.

src/theory/ee_manager_distributed.h
```diff
@@ -0,0 +1,147 @@
...
+  eq::EqualityEngine* getMasterEqualityEngine();
+
+ private:
+  /** Allocate equality engine that is context-dependent on c with info esi */
+  eq::EqualityEngine* allocateEqualityEngine(EeSetupInfo& esi,
+                                             context::Context* c);
+  /** Reference to the theory engine */
+  TheoryEngine& d_te;
+  /** notify class for master equality engine */
+  class MasterNotifyClass : public theory::eq::EqualityEngineNotify
```
4tXJ7f, line 103: Move inner class before data members and methods.

src/theory/ee_manager_distributed.cpp
```diff
@@ -0,0 +1,126 @@
...
+
+eq::EqualityEngine* EqEngineManagerDistributed::allocateEqualityEngine(
+    EeSetupInfo& esi, context::Context* c)
+{
+  if (esi.d_notify != nullptr)
+  {
+    return new eq::EqualityEngine(
+        *esi.d_notify, c, esi.d_name, esi.d_constantsAreTriggers);
+  }
+  // the theory don't care about explicit notifications
```
4tXJ7f, line 121:
```suggestion
  // the theory doesn't care about explicit notifications
```

src/theory/ee_manager_distributed.h
ajreynol: I think that's only for cases where efficiency is highly important, I would prefer a more typical map here.

PR #4869, opened 2020-08-10
Refactor functions that print commands (Part 1)

abdoo8080:
This PR is a step towards migrating commands to the Term/Sort level. The functions for printing `synth-fun` command and its grammar were modified to remove their dependency on command objects and use nodes instead of exprs and types. Similar changes to other functions that print commands will follow.


src/smt/smt_engine.cpp
```diff
@@ -1344,9 +1344,29 @@ void SmtEngine::declareSynthFun(const std::string& id,
     setUserAttribute("sygus-synth-grammar", func, attr_value, "");
   }
   Trace("smt") << "SmtEngine::declareSynthFun: " << func << "\n";
-  // dumping SynthFunCommand from smt-engine is currently broken (please take at
-  // CVC4/cvc4-projects#211)
-  // Dump("raw-benchmark") << SynthFunCommand(id, func, sygusType, isInv, vars);
+
```
ajreynol, line 1347: Can you add a guard around this block of code? 
```
if (Dump.isOn("raw-benchmark"))
{
  ...
}
```
Also maybe add a comment regarding why this is a special case, i.e. "!!! TEMPORARY: We cannot construct a SynthFunCommand since we cannot construct a Term-level Grammar from a Node-level sygus TypeNode. Thus we must print the command using the Node-level utility method for now."

src/printer/printer.cpp
```diff
@@ -117,5 +117,19 @@ Printer* Printer::getPrinter(OutputLanguage lang)
   return d_printers[lang].get();
 }
 
+void printUnknown(std::ostream& out, const std::string& name)
```
ajreynol, line 120: I would call this `printUnknownCommand`.

src/smt/smt_engine.cpp
abdoo8080: Done!

src/printer/printer.cpp
abdoo8080: Done!

PR #4871, opened 2020-08-10
(proof-new) Improving proof-production in Equality Engine

HanielB:
This commit improves functionalities of the equality engine so that it is easier to produce proofs for its reasoning. They are:
- avoiding assertion of already entailed predicates/equalities.
- better EqProof of disequalities with constants
- correct EqProof involving n-ary congruence kinds


src/theory/uf/equality_engine.h
```diff
@@ -415,6 +415,22 @@ class EqualityEngine : public context::ContextNotifyObj {
...
+   *
+   * The equality is only built if the nodes the ids correspond to are not
+   * internal nodes in the equality engine, i.e., they correspond to full
+   * applications of the respective kinds. Since the equality engine also
+   * applies congruence over n-ary kinds, internal nodes, i.e., partial
+   * applications, may still correspond to "full applications" in the
+   * first-order sense. Therefore this method also checks, in the case of n-ary
+   * congruence kinds, if an equality between "full applications" can be built.
+   */
+  void buildEqConclusion(EqualityNodeId id1,
```
ajreynol, line 430: I would prefer if this was not a method of this class, since this is purely for the sake of proofs.  I would either inline this into where its used or add it in an external helper class.

ajreynol: Actually on second thought, let's just make it clear that this is for proofs only, either as the first comment above this method, or using a block `// -------------------------- for proof-new` / `//---------------------- end for proof-new`.

HanielB: OK. FWIW, inlining would be an issue because there are two occurrences and an external method would be an issue because it uses a few internal datastructures of the equality engine.

Issue #4866, opened 2020-08-10
Error after three (reset) command


sy6sy2:
**Describe the bug**
After three `(reset)` command CVC4 (i) ignore `--bv-print-consts-as-indexed-symbols` option and (ii) trigger an error on `(reset)` command.

Reproduce the bug with this simple formula and the command given under the formula:

```SMT
(set-option :produce-models true)
(set-logic QF_ABV)

(declare-fun a () (_ BitVec 32))
(assert (= a (_ bv1 32)))

(check-sat)
(get-value (a))

(reset)

(set-option :produce-models true)
(set-logic QF_ABV)

(declare-fun a () (_ BitVec 32))
(assert (= a (_ bv1 32)))

(check-sat)
(get-value (a))

(reset)

(set-option :produce-models true)
(set-logic QF_ABV)

(declare-fun a () (_ BitVec 32))
(assert (= a (_ bv1 32)))

(check-sat)
(get-value (a))

(reset)

(set-option :produce-models true)
(set-logic QF_ABV)

(declare-fun a () (_ BitVec 32))
(assert (= a (_ bv1 32)))

(check-sat)
(get-value (a))
```

Result:

```bash
sat
((a (_ bv1 32)))
sat
((a (_ bv1 32)))
sat
((a #b00000000000000000000000000000001))
(error "Parse Error: /tmp/test.smt2:32.6: expected SMT-LIBv2 command, got `reset'.

  (reset)
   ^
")
```

*Note*: If the bug is triggered with an SMT2 input file please minimize it with [ddsmt](https://github.com/aniemetz/ddSMT).

**Command line arguments**: `cvc4 --bv-print-consts-as-indexed-symbols /tmp/test.smt2`
**CVC4 version/commit**: This is CVC4 version 1.9-prerelease [git master 33b96c65]
compiled with GCC version 4.2.1 Compatible Apple LLVM 11.0.3 (clang-1103.0.32.62)
on Aug 10 2020 11:27:09
**Operating system**: macOS 10.15.6

**`configure.sh` options**
Paste `configure.sh` here.

**`configure.sh` output**
Paste the output of `configure.sh` here.

nafur:
Indeed, this error is triggered by
```
(reset)
(reset)
(reset)
```

nafur:
After many detours, I found the following:
[...]

PR #4873, opened 2020-08-11
Fix memory issue in AntlrInput::parseError

### LINKED ISSUE #4069: constraints with no options trigger heap-buffer-overflow
Hi,
For this formula:
[outputbof.txt](https://github.com/CVC4/CVC4/files/4328373/outputbof.txt)
The address sanitizer detects a heap-buffer-overflow vulnerability:
```
=================================================================
==3147995==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x62e00000b419 at pc 0x7fa48a172a17 bp 0x7ffff3da3340 sp 0x7ffff3da3330
READ of size 1 at 0x62e00000b419 thread T0
    #0 0x7fa48a172a16 in CVC4::parser::parseErrorHelper(char const*, int, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) /home/legend/CVC4asan/src/parser/antlr_input.cpp:419
    #1 0x7fa48a173bc0 in CVC4::parser::AntlrInput::parseError(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, bool) /home/legend/CVC4asan/src/parser/antlr_input.cpp:507
    #2 0x7fa48a177a3a in CVC4::parser::AntlrInput::reportError(ANTLR3_BASE_RECOGNIZER_struct*) /home/legend/CVC4asan/src/parser/antlr_input_imports.cpp:267
    #3 0x7fa48a45b139 in termNonVariable /home/legend/CVC4asan/build/src/parser/smt2/Smt2Parser.c:11607
    #4 0x7fa48a4696f6 in term /home/legend/CVC4asan/build/src/parser/smt2/Smt2Parser.c:10272
    #5 0x7fa48a46a0c1 in termList /home/legend/CVC4asan/build/src/parser/smt2/Smt2Parser.c:12986
    #6 0x7fa48a45c154 in termNonVariable /home/legend/CVC4asan/build/src/parser/smt2/Smt2Parser.c:10628
    #7 0x7fa48a4696f6 in term /home/legend/CVC4asan/build/src/parser/smt2/Smt2Parser.c:10272
    #8 0x7fa48a46a0c1 in termList /home/legend/CVC4asan/build/src/parser/smt2/Smt2Parser.c:12986
    #9 0x7fa48a45c154 in termNonVariable /home/legend/CVC4asan/build/src/parser/smt2/Smt2Parser.c:10628
    #10 0x7fa48a4696f6 in term /home/legend/CVC4asan/build/src/parser/smt2/Smt2Parser.c:10272
    #11 0x7fa48a4a9391 in command /home/legend/CVC4asan/build/src/parser/smt2/Smt2Parser.c:3466
    #12 0x7fa48a4b4369 in parseCommand /home/legend/CVC4asan/build/src/parser/smt2/Smt2Parser.c:2221
    #13 0x7fa48a1886f9 in CVC4::parser::Parser::nextCommand() /home/legend/CVC4asan/src/parser/parser.cpp:706
    #14 0x55c83a4c42bb in runCvc4(int, char**, CVC4::Options&) /home/legend/CVC4asan/src/main/driver_unified.cpp:469
    #15 0x55c83a4bc6a8 in main /home/legend/CVC4asan/src/main/main.cpp:55
    #16 0x7fa4856dd022 in __libc_start_main (/usr/lib/libc.so.6+0x27022)
    #17 0x55c83a4bcb2d in _start (/home/legend/CVC4asan/build/bin/cvc4+0x15b2d)

0x62e00000b419 is located 0 bytes to the right of 45081-byte region [0x62e000000400,0x62e00000b419)
allocated by thread T0 here:
    #0 0x7fa48a880b3a in __interceptor_malloc /build/gcc/src/gcc/libsanitizer/asan/asan_malloc_linux.cc:144
    #1 0x7fa48a4cf6a6 in antlr3read8Bit (/home/legend/CVC4asan/build/src/parser/libcvc4parser.so.6+0x45e6a6)
    #2 0x7fa48a4cf624 in antlr3CreateFileStream (/home/legend/CVC4asan/build/src/parser/libcvc4parser.so.6+0x45e624)
    #3 0x7fa48a4cf1a4 in antlr3FileStreamNew (/home/legend/CVC4asan/build/src/parser/libcvc4parser.so.6+0x45e1a4)
    #4 0x7fa48a169bbf in CVC4::parser::newAntlr3FileStream(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) /home/legend/CVC4asan/src/parser/antlr_input.cpp:83
    #5 0x7fa48a16bc84 in CVC4::parser::AntlrInputStream::newFileInputStream(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, bool) /home/legend/CVC4asan/src/parser/antlr_input.cpp:150
    #6 0x7fa48a182bd7 in CVC4::parser::Input::newFileInput(CVC4::language::input::Language, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, bool) /home/legend/CVC4asan/src/parser/input.cpp:62
    #7 0x7fa48a1aefec in CVC4::parser::ParserBuilder::build() /home/legend/CVC4asan/src/parser/parser_builder.cpp:72
    #8 0x55c83a4c41ae in runCvc4(int, char**, CVC4::Options&) /home/legend/CVC4asan/src/main/driver_unified.cpp:455
    #9 0x55c83a4bc6a8 in main /home/legend/CVC4asan/src/main/main.cpp:55
    #10 0x7fa4856dd022 in __libc_start_main (/usr/lib/libc.so.6+0x27022)

SUMMARY: AddressSanitizer: heap-buffer-overflow /home/legend/CVC4asan/src/parser/antlr_input.cpp:419 in CVC4::parser::parseErrorHelper(char const*, int, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)
Shadow bytes around the buggy address:
  0x0c5c7fff9630: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x0c5c7fff9640: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x0c5c7fff9650: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x0c5c7fff9660: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x0c5c7fff9670: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
=>0x0c5c7fff9680: 00 00 00[01]fa fa fa fa fa fa fa fa fa fa fa fa
  0x0c5c7fff9690: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x0c5c7fff96a0: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x0c5c7fff96b0: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x0c5c7fff96c0: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x0c5c7fff96d0: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
Shadow byte legend (one shadow byte represents 8 application bytes):
  Addressable:           00
  Partially addressable: 01 02 03 04 05 06 07 
  Heap left redzone:       fa
  Freed heap region:       fd
  Stack left redzone:      f1
  Stack mid redzone:       f2
  Stack right redzone:     f3
  Stack after return:      f5
  Stack use after scope:   f8
  Global redzone:          f9
  Global init order:       f6
  Poisoned by user:        f7
  Container overflow:      fc
  Array cookie:            ac
  Intra object redzone:    bb
  ASan internal:           fe
  Left alloca redzone:     ca
  Right alloca redzone:    cb
  Shadow gap:       
```

nafur:
The `parseErrorHelper` message can run into memory errors when the it prints the last line and this last line is not terminated by a newline. I suspect other conditions may trigger it as well (like only using `\r` as line terminator).
This PR computes the size of the buffer from the start of the current line to the end of the buffer and makes sure that we never go beyond this buffer. The calculation looks a bit awkward, I've not been able to obtain this information from the ANTLR input stream in a nicer way...
(I found this issue when looking at #4866).
Fixes #4069 .


4tXJ7f:
If I am not mistaken, this may also fix #4069. I remember looking into the issue at some point and I had a similar fix but I seem to remember that there was another, similar, issue. I'll try to remember what it was.

nafur:
Indeed, #4069 is also fixed by this. #4069 is essentially an input that "just ends" in the middle of some expression, hence the error happens on the last line which is not terminated by a newline.
It goes from an asan error to
```
(error "Parse Error: test.smt2:323.646: Unexpected token: '<<EOF>>'.")
```

src/parser/antlr_input.cpp
```diff
@@ -501,9 +507,14 @@ std::string parseErrorHelper(const char* lineStart, int charPositionInLine, cons
 
 void AntlrInput::parseError(const std::string& message, bool eofException)
 {
-  string updatedMessage = parseErrorHelper((const char*)d_antlr3InputStream->getLineBuf(d_antlr3InputStream),
-                                           d_lexer->getCharPositionInLine(d_lexer),
-                                           message);
+  auto lineLength = d_antlr3InputStream->size(d_antlr3InputStream)
+                    - (static_cast<char*>(d_antlr3InputStream->currentLine)
+                       - static_cast<char*>(d_antlr3InputStream->data));
```
4tXJ7f, line 512: Hm, so this may actually the issue that I've run into previously when trying to fix the issue: doesn't this assume that all the lines are allocated linearly in the memory?

nafur: You mean ANTLR could use the buffer in a cyclic fashion so that the current line is followed by something else than the succeeding line, right?
I don't think this would be a problem, but may highlight a naming issue.
Note that printing still stops when we hit a `\n` and `lineLength` is only used when there is no `\n` between `lineStart` (or rather `getLineBuf()` and the end of the buffer. A name like `remainingBufferSize` (instead of `lineLength`) may be more appropriate here...

4tXJ7f: Hm, the problem is that the individual lines can be at completely arbitrary places in memory, when using the buffered input. It may be the case that your solution works when reading from a file but fails when `cat`ing it through `stdin`. Could you try that?

For buffered inputs, the lines are stored [here](https://github.com/CVC4/CVC4/blob/8374a8dd4bf740bf26748c1dbe1616ad798cf624/src/parser/line_buffer.h#L67) and allocated [here](https://github.com/CVC4/CVC4/blob/8374a8dd4bf740bf26748c1dbe1616ad798cf624/src/parser/line_buffer.cpp#L82-L86). I believe that the `LineBuffer` is only used when reading from `stdin` (see [this](https://github.com/CVC4/CVC4/blob/8374a8dd4bf740bf26748c1dbe1616ad798cf624/src/main/driver_unified.cpp#L425)).

nafur: Ah, good point. It seems we can use `sizeBuf` instead of `size()` though:

> $ cat test_all.sh 
> #!/bin/bash
> bin/cvc4 test_4866.smt2
> bin/cvc4 --lang=smt2 < test_4866.smt2
> bin/cvc4 --incremental test_4069.smt2
> bin/cvc4 --incremental --lang=smt2 < test_4069.smt2

> $ ./test_all.sh 
> (error "Parse Error: test_4866.smt2:3.6: expected SMT-LIBv2 command, got `reset'.
> 
>   (reset)
>    ^
> ")
> (error "Parse Error: <stdin>:2.6: expected SMT-LIBv2 command, got `reset'.
> 
>   (reset)
>    ^
> ")
> (error "Parse Error: test_4069.smt2:323.646: Unexpected token: '<<EOF>>'.")
> (error "Parse Error: <stdin>:324.0: Unexpected token: '<<EOF>>'.")

(`bin/cvc4` being compiled with `Debug --asan --ubsan`)

4tXJ7f: Hmm, I am actually mostly worried about `static_cast<char*>(d_antlr3InputStream->currentLine) - static_cast<char*>(d_antlr3InputStream->data)` because I am worried that that is not accurate when we allocate the lines in a way that is not adjacent. This may only happen when you do it truly interactively, i.e. you only feed a command to CVC4 after you get an answer for a previous one (if you pass the whole file to `stdin` it may be the case that we allocate all the lines at once and you don't have the problem, I'd need to check).

nafur: I agree that this solution may not be completely accurate if the lines are not in order (within the buffer). However, this also requires that the line we are looking at is not properly terminated, and it is directly followed (i.e. not separated by a newline character) by another line.
But even in this case, this PR improves the current situation: instead of memory errors we may only print another line after the one we are actually interested in.
I'd be glad if you can come up with an even better solution, but this `ANTLR3_INPUT_STREAM_struct` really does not give you a lot to work with...

src/parser/antlr_input.cpp
```diff
@@ -501,9 +507,14 @@ std::string parseErrorHelper(const char* lineStart, int charPositionInLine, cons
 
 void AntlrInput::parseError(const std::string& message, bool eofException)
 {
-  string updatedMessage = parseErrorHelper((const char*)d_antlr3InputStream->getLineBuf(d_antlr3InputStream),
-                                           d_lexer->getCharPositionInLine(d_lexer),
-                                           message);
+  auto lineLength = d_antlr3InputStream->sizeBuf
+                    - (static_cast<char*>(d_antlr3InputStream->currentLine)
+                       - static_cast<char*>(d_antlr3InputStream->data));
+  string updatedMessage = parseErrorHelper(
```
aniemetz, line 513:
```suggestion
  std::string updatedMessage = parseErrorHelper(
```

PR #4875, opened 2020-08-12
Replace Expr-level datatype with Node-level DType

ajreynol:
This PR makes two simultaneous changes:
- The new API uses Node-level DType instead of Expr-level Datatype.  This required converting 2 of the last remaining calls in the parser that used Expr to use the new API.
- Internally constructed datatypes (e.g. for sygus) use Node-level DType instead of Expr-level Datatype.  Note this required moving a block of code for handling a corner case of sygus from Datatype -> DType.

This PR removes : 
- ExprManger::mkDatatypeType.
- The Expr-level datatype itself.  This PR removes all references to its include file.

It also updates one remaining unit test from Expr -> Node.

This PR will enable further removal of other datatype-specific things in the Expr layer.


test/unit/theory/type_enumerator_white.h
```diff
@@ -145,51 +146,68 @@ class TypeEnumeratorWhite : public CxxTest::TestSuite {
...
+  {
+    DType colors("Colors");
+    colors.addConstructor(std::make_shared<DTypeConstructor>("red"));
+    colors.addConstructor(std::make_shared<DTypeConstructor>("orange"));
+    colors.addConstructor(std::make_shared<DTypeConstructor>("yellow"));
+    colors.addConstructor(std::make_shared<DTypeConstructor>("green"));
+    colors.addConstructor(std::make_shared<DTypeConstructor>("blue"));
+    colors.addConstructor(std::make_shared<DTypeConstructor>("violet"));
+    TypeNode colorsType = d_nm->mkDatatypeType(colors);
+    // TODO
```
4tXJ7f, line 189: What's with those `TODO`s?

src/expr/dtype.cpp
```diff
@@ -245,6 +245,45 @@ void DType::addSygusConstructor(Node op,
...
+  // Notice we only want to do this for sygus datatypes that are user-provided.
+  // At the moment, the condition !allow_all implies the grammar is
+  // user-provided and hence may require a default constant.
+  // TODO (https://github.com/CVC4/cvc4-projects/issues/38):
+  // In an API for SyGuS, it probably makes more sense for the user to
+  // explicitly add the "any constant" constructor with a call instead of
+  // passing a flag. This would make the block of code unnecessary.
+  if (allowConst && !allowAll)
+  {
+    // if i don't already have a constant (0-ary constructor)
```
4tXJ7f, line 262:
```suggestion
    // if I don't already have a constant (0-ary constructor)
```

src/api/cvc4cpp.cpp
```diff
@@ -3062,19 +3155,27 @@ std::vector<Sort> Solver::mkDatatypeSortsInternal(
                                          dtypedecls[i],
                                          i)
         << "a datatype declaration with at least one constructor";
+    CVC4_API_SOLVER_TRY_CATCH_BEGIN;
     datatypes.push_back(dtypedecls[i].getDatatype());
+    CVC4_API_SOLVER_TRY_CATCH_END;
```
4tXJ7f, line 3160: Maybe better to put `CVC4_API_SOLVER_TRY_CATCH_BEGIN`/`CVC4_API_SOLVER_TRY_CATCH_END` at the beginning and end of the function block (like the other functions in this file).

test/unit/theory/type_enumerator_white.h
ajreynol: Whoops, I'll remove.

PR #4878, opened 2020-08-12
[proof-new] Generalize handling of MERGED_THROUGH_CONSTANST in EqProof conversion

HanielB:
Now that we are using the constant folding in equality engine for more than equality it is necessary to generalize the previously-hard-coded handling of MERGED_THROUGH_CONSTANTS.

### DIFF
--- a/src/theory/uf/eq_proof.cpp
+++ b/src/theory/uf/eq_proof.cpp
@@ -480,7 +480,7 @@ bool EqProof::expandTransitivityForDisequalities(
     p->addStep(congConclusion,
                PfRule::CONG,
                substPremises,
-               {nm->operatorOf(kind::EQUAL)},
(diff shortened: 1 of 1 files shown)


src/theory/uf/eq_proof.cpp
```diff
@@ -480,7 +480,7 @@ bool EqProof::expandTransitivityForDisequalities(
     p->addStep(congConclusion,
                PfRule::CONG,
                substPremises,
-               {nm->operatorOf(kind::EQUAL)},
+               {ProofRuleChecker::mkKindNode(kind::EQUAL)},
```
ajreynol, line 483: This breaks the version of CONG on master, but since this is not active code yet, maybe its ok.  I will submit the update to CONG in the meantime.

PR #4880, opened 2020-08-12
Split SmtSolver from SmtEngine

ajreynol:
This class is responsible for maintaining TheoryEngine and PropEngine and implementing the `check-sat` command.  It also has an interface for processing/pushing the current assertions into the PropEngine.

This class will be used directly by other extension solvers (for abduction, interpolation, qe, sygus, etc.).


src/smt/smt_solver.cpp
```diff
@@ -0,0 +1,251 @@
...
+
+TheoryEngine* SmtSolver::getTheoryEngine() const
+{
+  return d_theoryEngine.get();
+}
+
+prop::PropEngine* SmtSolver::getPropEngine() const
+{
+  return d_propEngine.get();
+}
```
nafur, line 248: ditto.

src/smt/smt_engine.cpp
```diff
@@ -253,7 +252,13 @@ SmtEngine::SmtEngine(ExprManager* em, Options* optr)
   d_resourceManager->registerListener(d_routListener.get());
   // make statistics
   d_stats.reset(new SmtEngineStatistics());
-  
+  // make the SMT solver
+  d_smtSolver.reset(new SmtSolver(*this,
+                                  *d_state.get(),
+                                  d_resourceManager.get(),
+                                  *d_pp.get(),
+                                  *d_stats.get()));
```
nafur, line 260:
```suggestion
  d_smtSolver.reset(new SmtSolver(*this,
                                  *d_state,
                                  d_resourceManager.get(),
                                  *d_pp,
                                  *d_stats));
```

src/smt/smt_engine.cpp
```diff
@@ -280,11 +285,24 @@ Result SmtEngine::getStatusOfLastCommand() const
...
+
+TheoryEngine* SmtEngine::getTheoryEngine() const
+{
+  return d_smtSolver->getTheoryEngine();
+}
+
+prop::PropEngine* SmtEngine::getPropEngine() const
+{
+  return d_smtSolver->getPropEngine();
+}
```
nafur, line 305: ditto.

src/smt/smt_engine.cpp
```diff
@@ -1007,71 +957,47 @@ theory::TheoryModel* SmtEngine::getAvailableModel(const char* c) const
...
-void SmtEngine::notifyPushPre() { processAssertionsInternal(); }
+void SmtEngine::notifyPushPre() { d_smtSolver->processAssertions(*d_asserts); }
+
 void SmtEngine::notifyPushPost()
 {
   TimerStat::CodeTimer pushPopTimer(d_stats->d_pushPopTime);
-  d_propEngine->push();
+  PropEngine* pe = getPropEngine();
+  Assert(pe != nullptr);
+  pe->push();
```
nafur, line 967: This is a pattern I've seen a few times now.
I'd suggest the following instead, as it seems more natural to me:
```suggestion
  Assert(getPropEngine() != nullptr);
  getPropEngine()->push();
```

src/smt/smt_engine.cpp
```diff
@@ -2465,45 +2350,43 @@ bool SmtEngine::getAbduct(const Node& conj, Node& abd)
...
-    for( unsigned i=0; i<qs_n.size(); i++ ){
-      qs.push_back( qs_n[i].toExpr() );
-    }
-  }else{
-    Assert(false);
+  TheoryEngine* te = getTheoryEngine();
+  Assert(te != nullptr);
+  std::vector<Node> qs_n;
+  te->getInstantiatedQuantifiedFormulas(qs_n);
+  for (unsigned i = 0; i < qs_n.size(); i++)
```
nafur, line 2357:
```suggestion
  for (std::size_t i = 0, n = qs_n.size(); i < n; i++)
```

src/smt/smt_engine.cpp
```diff
@@ -2465,45 +2350,43 @@ bool SmtEngine::getAbduct(const Node& conj, Node& abd)
...
-    for( unsigned i=0; i<insts_n.size(); i++ ){
-      insts.push_back( insts_n[i].toExpr() );
-    }
-  }else{
-    Assert(false);
+  TheoryEngine* te = getTheoryEngine();
+  Assert(te != nullptr);
+  std::vector<Node> insts_n;
+  te->getInstantiations(Node::fromExpr(q), insts_n);
+  for (unsigned i = 0; i < insts_n.size(); i++)
```
nafur, line 2369:
```suggestion
  for (std::size_t i = 0, n = insts_n.size(); i < n; i++)
```

src/smt/smt_engine.cpp
```diff
@@ -2465,45 +2350,43 @@ bool SmtEngine::getAbduct(const Node& conj, Node& abd)
...
-      }
-      tvecs.push_back( tvec );
+  TheoryEngine* te = getTheoryEngine();
+  Assert(te != nullptr);
+  std::vector<std::vector<Node>> tvecs_n;
+  te->getInstantiationTermVectors(Node::fromExpr(q), tvecs_n);
+  for (unsigned i = 0; i < tvecs_n.size(); i++)
+  {
+    std::vector<Expr> tvec;
+    for (unsigned j = 0; j < tvecs_n[i].size(); j++)
```
nafur, line 2385:
```suggestion
  for (std::size_t i = 0, n = tvecs_n.size(); i < n; i++)
  {
    std::vector<Expr> tvec;
    for (std::size_t j = 0, m = tvecs_n[i].size(); j < m; j++)
```

src/smt/smt_engine.cpp
ajreynol: Is this about the const?  This is actually required to make things compile in several places.

src/smt/smt_solver.cpp
```diff
@@ -0,0 +1,251 @@
...
+   * are unregistered by the obsolete PropEngine object before registered
+   * again by the new PropEngine object */
+  d_propEngine.reset(nullptr);
+  d_propEngine.reset(new PropEngine(
+      d_theoryEngine.get(), d_smt.getContext(), d_smt.getUserContext(), d_rm));
+
+  Trace("smt-debug") << "Setting up theory engine..." << std::endl;
+  d_theoryEngine->setPropEngine(getPropEngine());
+  Trace("smt-debug") << "Finishing init for theory engine..." << std::endl;
+  d_theoryEngine->finishInit();
```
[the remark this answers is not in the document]
ajreynol: The theory engine is not reset upon reset assertions, only the prop engine.

Notice that the pointer from TheoryEngine to PropEngine is such that`TheoryEngine::finishInit` does not depend on knowing the associated PropEngine. I'll clarify this.

src/smt/smt_solver.cpp
```diff
@@ -0,0 +1,251 @@
...
+  Trace("limit") << "SmtSolver::check(): cumulative millis "
+                 << d_rm->getTimeUsage() << ", resources "
+                 << d_rm->getResourceUsage() << endl;
+
+  Result r = Result(result, filename);
+
+  if ((options::solveRealAsInt() || options::solveIntAsBV() > 0)
+      && r.asSatisfiabilityResult().isSat() == Result::UNSAT)
+  {
+    r = Result(Result::SAT_UNKNOWN, Result::UNKNOWN_REASON);
```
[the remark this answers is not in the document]
ajreynol: Interesting, I was preserving behavior but I think you're right, I'll correct this.

src/smt/smt_solver.cpp
```diff
@@ -0,0 +1,251 @@
...
+      r = Result(Result::SAT);
+    }
+    else if (r.asSatisfiabilityResult().isSat() == Result::SAT)
+    {
+      // only if satisfaction complete
+      LogicInfo logic = d_smt.getLogicInfo();
+      if (logic.isPure(theory::THEORY_ARITH) || logic.isPure(theory::THEORY_BV))
+      {
+        r = Result(Result::UNSAT);
+      }
```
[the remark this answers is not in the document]
ajreynol: This holds for linear arithmetic, as well as any theory with no partial operators.  I think we should check for linear (apparently this was wrong before).

src/smt/smt_solver.cpp
```diff
@@ -0,0 +1,251 @@
...
+
+  // then, initialize the assertions
+  as.initializeCheckSat(assumptions, inUnsatCore, isEntailmentCheck);
+
+  // make the check
+  Assert(d_smt.isFullyInited());
+
+  Trace("smt") << "SmtSolver::check()" << endl;
+
+  const std::string& filename = d_state.getFilename();
```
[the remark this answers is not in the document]
ajreynol: I have a slight preference for creating the local variable here.

PR #4883, opened 2020-08-12
(proof-new) SMT proof postprocess callback

ajreynol:
This is the callback class for processing the final proof, which connects proofs of preprocessing and expands unwanted macro steps based on proof granularity mode.

The next step will be to add the ProofNodeUpdater that uses this callback and runs final sanity checks for checking proofs.


src/smt/proof_post_processor.cpp
```diff
@@ -0,0 +1,572 @@
+/*********************                                                        */
+/*! \file proof_post_processor.cpp
+ ** \verbatim
+ ** Top contributors (to current version):
+ **   Morgan Deters, Andrew Reynolds, Tim King
+ ** This file is part of the CVC4 project.
+ ** Copyright (c) 2009-2019 by the authors listed in the file AUTHORS
```
HanielB, line 7:
```suggestion
 ** Copyright (c) 2009-2020 by the authors listed in the file AUTHORS
```

src/smt/proof_post_processor.h
```diff
@@ -0,0 +1,106 @@
+/*********************                                                        */
+/*! \file proof_post_processor.h
+ ** \verbatim
+ ** Top contributors (to current version):
+ **   Andrew Reynolds
+ ** This file is part of the CVC4 project.
+ ** Copyright (c) 2009-2019 by the authors listed in the file AUTHORS
```
HanielB, line 7:
```suggestion
 ** Copyright (c) 2009-2020 by the authors listed in the file AUTHORS
```

src/smt/proof_post_processor.h
```diff
@@ -0,0 +1,106 @@
...
+  ProofPostprocessCallback(ProofNodeManager* pnm,
+                           SmtEngine* smte,
+                           ProofGenerator* pppg);
+  ~ProofPostprocessCallback() {}
+  /**
+   * Initialize, called once for each new ProofNode to process. This initializes
+   * static information to be used by successive calls to update.
+   */
+  void initializeUpdate();
+  /** set eliminate rule */
```
HanielB, line 48: say this is populating `d_elimRules`

src/smt/proof_post_processor.h
```diff
@@ -0,0 +1,106 @@
...
+  /** The witness form assumptions used in the proof */
+  std::vector<Node> d_wfAssumptions;
+  /** Kinds of proof rules we are eliminating */
+  std::unordered_set<PfRule, PfRuleHashFunction> d_elimRules;
+  //---------------------------------reset at the begining of each update
+  /** Mapping assumptions to their proof from preprocessing */
+  std::map<Node, std::shared_ptr<ProofNode> > d_assumpToProof;
+  //---------------------------------end reset at the begining of each update
+  /**
+   * Expand macros in the given application, add the expanded proof to cdp.
```
HanielB, line 79: Say what are the macros (MACRO_* rules in proof_rule.h). Give an example of an expansion. Also say that this method calls `update` to perform possible post-processing in the rules it introduces as a result of the expansion.

src/smt/proof_post_processor.h
```diff
@@ -0,0 +1,106 @@
...
+   * Add proof for witness form. This returns the equality t = toWitness(t)
+   * and ensures that the proof of this equality has been added to cdp.
+   * Notice the proof of this fact may have open assumptions of the form:
+   *   k = toWitness(k)
+   * where k is a skolem. Furthermore, note that all open assumptions of this
+   * form are available via d_wfpm.getWitnessFormEqs() in the remainder of
+   * the lifetime of this class.
+   */
+  Node addProofForWitnessForm(Node t, CDProof* cdp);
+  /** Apply transivity if necessary for the arguments */
```
HanielB, line 95: say that `tchildren` are already ordered as a transitivity chain. Say that it returns `null` if `tchildren` is empty and the only child if singleton..

src/smt/proof_post_processor.cpp
```diff
@@ -0,0 +1,572 @@
...
+  size_t tsize = tchildren.size();
+  if (tsize > 1)
+  {
+    Node lhs = tchildren[0][0];
+    Node rhs = tchildren[tsize - 1][1];
+    Node eq = lhs.eqNode(rhs);
+    cdp->addStep(eq, PfRule::TRANS, tchildren, {});
+    return eq;
+  }
+  else if (tchildren.size() == 1)
```
HanielB, line 546:
```suggestion
  else if (tsize == 1)
```

src/smt/proof_post_processor.h
```diff
@@ -0,0 +1,106 @@
...
+  /** Kinds of proof rules we are eliminating */
+  std::unordered_set<PfRule, PfRuleHashFunction> d_elimRules;
+  //---------------------------------reset at the begining of each update
+  /** Mapping assumptions to their proof from preprocessing */
+  std::map<Node, std::shared_ptr<ProofNode> > d_assumpToProof;
+  //---------------------------------end reset at the begining of each update
+  /**
+   * Expand macros in the given application, add the expanded proof to cdp.
+   */
+  Node expandMacros(PfRule id,
```
HanielB, line 81: describe the parameters and the return

src/smt/proof_post_processor.h
```diff
@@ -0,0 +1,129 @@
...
+  /** Kinds of proof rules we are eliminating */
+  std::unordered_set<PfRule, PfRuleHashFunction> d_elimRules;
+  //---------------------------------reset at the begining of each update
+  /** Mapping assumptions to their proof from preprocessing */
+  std::map<Node, std::shared_ptr<ProofNode> > d_assumpToProof;
+  //---------------------------------end reset at the begining of each update
+  /**
+   * Expand rules in the given application, add the expanded proof to cdp.
+   * The set of rules we expand is configured by calls to setEliminateRule
+   * above. This method calls update to perform possible post-processing in the
```
HanielB, line 84: I think this is a bit misleading. We will expand if the rule has been set via `setEliminateRule`, but there is a hidden assumption that a rule will only be eliminated if it was added via `setEliminateRule` *and* it is MACRO_*, SUBS or REWRITE.

ajreynol: I'll add this to the doc of setEliminateRule.

Issue #4778, opened 2020-07-21
Error in option parsing: Error in --lang: Error in option parsing: unknown input language `sygus'


peleghila:
Upgrading to version 1.8, `--lang sygus` returns the error 

```>cvc4-1.8-win64-opt.exe --lang sygus
(error "Error in option parsing: Error in --lang: Error in option parsing: unknown input language `sygus'
Try --language help")
```

despite the fact that `sygus` is listed under `--lang help` and appears to be supported in src/options/language.cpp. (As an aside, `--language help`, the recommended option, is invalid.)

**CVC4 version/commit**: cvc4-1.8-win64-opt.exe downloaded from site (calling `--version` returns `1.9`)
**Operating system**: Windows 10

nafur:
Thanks for the report. About a month ago, support for sygus **version 1** has been removed from CVC4 (see #4651 for details), which was available via `--lang sygus1` at this point.
Now, we only support sygus2 via `--lang sygus2`. I think we currently do not support `--lang sygus` to avoid confusion concerning these different versions.

I hope sygus2 works for you, please reopen this ticket if it does not.

peleghila:
Actually `--lang sygus1` looks like it's working just fine. Anyway, it would be nice to remove this from the `--help` list just to avoid the confusion.

nafur:
Ah, I see, my bad. Indeed, these changes came after the release. Though `sygus1` still works in the current release, I'd thus still suggest to move to `sygus2` at some point.

peleghila:
It looks from the doc that `sygus` is supposed to indicate v2 (that's what I tried to use it for).

```
> cvc4-1.8-win64-opt.exe --lang help
Languages currently supported as arguments to the -L / --lang option:
  auto                           attempt to automatically determine language
  cvc4 | presentation | pl       CVC4 presentation language
  smt | smtlib | smt2 |
  smt2.0 | smtlib2 | smtlib2.0   SMT-LIB format 2.0
  smt2.5 | smtlib2.5             SMT-LIB format 2.5
  smt2.6 | smtlib2.6             SMT-LIB format 2.6 with support for the strings standard
  tptp                           TPTP format (cnf, fof and tff)
  sygus1                         SyGuS version 1.0
  sygus | sygus2                 SyGuS version 2.0
```

nafur:
This issue was fixed with #4884. Please come back to use if you have any further problems.

PR #4884, opened 2020-08-13
fixes a typo in help output and removes unsupported option

polgreen:
This PR fixes a typo in an error message and also removes the option `--lang sygus` from the help output. 

Currently `--lang sygus` throws a parsing error and CVC4 doesn't support sygus 1.0 anymore, so I took a guess that this should be removed from the help file?

~~~
$ cvc4 --lang sygus
(error "Error in option parsing: Error in --lang: Error in option parsing: unknown input language `sygus'
Try --language help")

~~~

### DIFF
--- a/src/options/language.cpp
+++ b/src/options/language.cpp
@@ -181,9 +181,8 @@ InputLanguage toInputLanguage(std::string language) {
     return input::LANG_SMTLIB_V2_6;
   } else if(language == "tptp" || language == "LANG_TPTP") {
     return input::LANG_TPTP;
-  }
-  else if (language == "sygus2" || language == "LANG_SYGUS_V2")
-  {
+  } else if(language == "sygus" || language == "sygus2" ||
+          language == "LANG_SYGUS" || language == "LANG_SYGUS_V2") {
     return input::LANG_SYGUS_V2;
   }
   else if (language == "auto" || language == "LANG_AUTO")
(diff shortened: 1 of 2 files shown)


nafur:
This in some sense also fixes #4778.
Open for discussion: would we rather remove `sygus` or have `sygus` as synonym for `sygus2`?

HanielB:
To keep in line with what we do with SMT-LIB I think we should have both. So rather than removing `--lang sygus` we should support it and remove `--lang sygus1`.

nafur:
@polgreen Would you add support for `--lang sygus`? Essentially, you just need to add it at https://github.com/CVC4/CVC4/blob/master/src/options/language.cpp#L185 just like it is checked at https://github.com/CVC4/CVC4/blob/master/src/options/language.cpp#L149

polgreen:
> To keep in line with what we do with SMT-LIB I think we should have both. So rather than removing `--lang sygus` we should support it as a synonym for `sygus2` and remove `--lang sygus1`.

lang sygus1 is already removed. Ok, so sygus should just map to sygus 2, I can do that

HanielB:
Thanks for taking care of this, @polgreen !

mpreiner:
@polgreen Can you please squash your commits and sign them off: https://github.com/CVC4/CVC4/blob/master/CONTRIBUTING.md

Thanks!

polgreen:
> @polgreen Can you please squash your commits and sign them off: https://github.com/CVC4/CVC4/blob/master/CONTRIBUTING.md
> 
> Thanks!

done

PR #4885, opened 2020-08-13
(proof-new) Prepare the theory of strings for proof reconstruction

ajreynol:
This updates the internal data structure for strings inferences "InferInfo" such that it is amenable to proof reconstruction.

Currently, the explanation for a conclusion is in two parts:
(1) d_ant, the antecedents that can be explained by the current equality engine,
(2) d_antn, the antecedents that cannot.

For proof reconstruction, the order of some antecedents matters. This is difficult with the above data structure since elements in these two vectors are not given an ordering relative to each other.  After this PR, we store:
(1) d_ant, *all* antecedants, which are ordered in a way that is amenable to proofs (to be introduced on following PRs),
(2) d_noExplain, the *subset* of d_ant that cannot be explained by the current equality engine.

This PR modifies calls to InferenceManager in preparation for extending it with a proof reconstructor `InferProofCons` which will convert strings::InferInfo into instructions for building ProofNode.


src/theory/strings/core_solver.cpp
```diff
@@ -1311,8 +1311,8 @@ void CoreSolver::processSimpleNEq(NormalForm& nfi,
       NormalForm::getExplanationForPrefixEq(nfi, nfj, index, index, ant);
       if (x.isConst() && y.isConst())
       {
-        // if both are constant, it's just a constant conflict
-        d_im.sendInference(ant, d_false, Inference::N_CONST, true);
+        // if both are constant, its just a constant conflict
```
4tXJ7f, line 1314:
```suggestion
        // if both are constant, it's just a constant conflict
```

src/theory/strings/inference_manager.h
```diff
@@ -109,52 +109,61 @@ class InferenceManager
...
-   * This function should be called when ( exp ^ exp_n ) => eq. The set exp
+   * This function should be called when exp => eq. The set exp
    * contains literals that are explainable, i.e. those that hold in the
    * equality engine of the theory of strings. On the other hand, the set
-   * exp_n ("explanations new") contain nodes that are not explainable by the
-   * theory of strings. This method may call sendLemma or otherwise add a
-   * InferInfo to d_pending, indicating a fact should be asserted to the
-   * equality engine. Overall, the result of this method is one of the
-   * following:
+   * noExplain contain nodes that are not explainable by the theory of strings.
```
4tXJ7f, line 118:
```suggestion
   * noExplain contains nodes that are not explainable by the theory of strings.
```

src/theory/strings/regexp_solver.cpp
```diff
@@ -480,19 +485,21 @@ bool RegExpSolver::checkEqcIntersect(const std::vector<Node>& mems)
...
+      d_im.sendInference(
+          vec_nodes, conc, Inference::RE_INTER_CONF, false, true);
       // conflict, return
       return false;
     }
     // rewrite to ensure the equality checks below are precise
-    Node mres = Rewriter::rewrite(nm->mkNode(STRING_IN_REGEXP, mi[0], resR));
-    if (mres == mi)
+    Node mres = nm->mkNode(STRING_IN_REGEXP, mi[0], resR);
+    Node mresr = Rewriter::rewrite(nm->mkNode(STRING_IN_REGEXP, mi[0], resR));
```
4tXJ7f, line 495:
```suggestion
    Node mresr = Rewriter::rewrite(mres);
```

src/theory/strings/core_solver.cpp
```diff
@@ -1638,7 +1646,7 @@ void CoreSolver::processSimpleNEq(NormalForm& nfi,
             lentTestSuccess = e;
             lenConstraint = entLit;
             // its not explained by the equality engine of this class
-            iinfo.d_antn.push_back(lenConstraint);
+            iinfo.d_noExplain.push_back(lenConstraint);
```
4tXJ7f, line 1649: Don't we also need to push to `iinfo.d_ant`?

ajreynol: It is being pushed into `d_ant`.  The length constraint is being accumulated into `lcVec`, which is added to `d_ant` on line 1711.  I'll add some further comments to make this clearer.

4tXJ7f: Oh, I see. Thanks for the comment.

PR #4888, opened 2020-08-13
Increase regress level to 2 for production build.

mpreiner:


ajreynol:
So I think I'm in favor of enabling regress2 as well.  However, can you move all regress0-2 benchmarks that currently take more than 20 seconds into regress3?  I believe there are at least two like this (a strings one and a sygus-inst one).

ajreynol:
This looks good, should the title be "Increase regress level to 2..." ?  Also I think the regression list may need to be updated?

mpreiner:
> This looks good, should the title be "Increase regress level to 2..." ? Also I think the regression list may need to be updated?

git add -u is your friend ... fixed :-)

PR #4890, opened 2020-08-13
Dynamic allocation of equality engine in Theory

ajreynol:
This PR updates Theory so that equality engines are allocated dynamically. The plan is to make this configurable based on the theory combination method.

The fundamental changes include:
- Add `d_equalityEngine` (raw) pointer to Theory, which is the "official" equality engine of the theory.
- Standardize methods for initializing Theory. This is now made more explicit in the documentation in theory.h, and includes a method `finishInitStandalone` for users of Theory that don't have an associated TheoryEngine.
- Refactor TheoryEngine::finishInit, including how Theory is initialized to incorporate the new policy.
- Things related to master equality engine are now specific to EqEngineManagerDistributed and hence can be removed from TheoryEngine. This will be further refactored in forthcoming PRs.

Note that the majority of changes are due to changing `d_equalityEngine.` to `d_equalityEngine->` throughout. 

FYI @barrettcw .


ajreynol:
@barrettcw , yes.  Arith and bv (in lazy mode) are still using equality engines in the same way as before, the only difference is now they are being assigned pointers to them from the equality engine manager.

src/theory/theory.h
```diff
@@ -264,17 +294,33 @@ class Theory {
...
+  /** Called to set the quantifiers engine. */
+  void setQuantifiersEngine(QuantifiersEngine* qe);
+  /** Called to set the decision manager. */
+  void setDecisionManager(DecisionManager* dm);
+  /**
+   * Finish theory initialization.  At this point, options and the logic
+   * setting are final, the master equality engine and quantifiers
+   * engine (if any) are initialized, and the official equality engine of this
+   * theory has been assigned.  This base class implementation
+   * does nothing. This should be done by TheoryEngine only.
```
4tXJ7f, line 312:
```suggestion
   * does nothing. This should be called by TheoryEngine only.
```

src/theory/theory.h
```diff
@@ -222,7 +244,15 @@ class Theory {
...
-
+  /**
+   * Pointer to the official equality engine of this theory, which is owned by
+   * the equality engine manager of TheoryEngine.
+   */
+  eq::EqualityEngine* d_equalityEngine;
+  /**
+   * The official equality engine, if we allocated it.
+   */
+  std::unique_ptr<eq::EqualityEngine> d_alocEqualityEngine;
```
4tXJ7f, line 255:
```suggestion
  std::unique_ptr<eq::EqualityEngine> d_allocEqualityEngine;
```

src/theory/theory.h
```diff
@@ -77,11 +77,34 @@ namespace eq {
...
+ * is either:
+ *
+ * (A) Using Theory as a standalone object, not associated with a TheoryEngine.
+ * In this case, simply call the public initialization method
+ * (Theory::finishInitStandalone).
+ *
+ * (B) TheoryEngine, which determines how the Theory acts in accordance with
+ * its theory combination policy. We require the following steps in order:
+ * (B.1) Get information about which equality engine notifications the Theory
+ * would like to be notified of (Theory::needsEqualityEngine).
```
4tXJ7f, line 91: Wouldn't it be more accurate to say something like:

> (B.1) Get information about whether this Theory requires an equality engine for congruence-based reasoning (Theory::needsEqualityEngine).

ajreynol: It will be both, whether we want to maintain one, and which notifications we want.  I'll update the comment.

PR #4891, opened 2020-08-13
Split SygusSolver from SmtEngine

ajreynol:
This is the solver for standard SyGuS queries.  Notice it now depends only on SmtSolver and not SmtEngine.

This PR updates Expr -> Node for the sygus interface in SmtEngine.

SmtEnginePrivate is no longer needed and is deleted with this PR.


src/smt/sygus_solver.cpp
```diff
@@ -0,0 +1,403 @@
...
+                         "unknown.";
+    }
+    else if (r.asSatisfiabilityResult().isSat())
+    {
+      InternalError()
+          << "SygusSolver::checkSynthSolution(): produced solution leads to "
+             "satisfiable negated conjecture.";
+    }
+  }
+}
```
nafur, line 389: It seems you forgot to remove the corresponding code from `smt_engine.cpp`...

src/smt/sygus_solver.h
```diff
@@ -0,0 +1,182 @@
...
+   * Declared SyGuS variables may be used in SyGuS constraints, in which they
+   * are assumed to be universally quantified.
+   *
+   * In SyGuS semantics, declared functions are treated in the same manner as
+   * declared variables, i.e. as universally quantified (function) variables
+   * which can occur in the SyGuS constraints that compose the conjecture to
+   * which a function is being synthesized. Thus declared functions should use
+   * this method as well.
+   */
+  void declareSygusVar(const std::string& id, Node var, TypeNode type);
```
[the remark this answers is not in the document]
ajreynol: Node is a pointer, and this is actually the default way of passing Node around.  It was actually a bit strange to me to have `const Expr&` (likewise `const Node&`) in some places in the SmtEngine.

src/smt/sygus_solver.cpp
ajreynol: Ah right, good catch.

src/smt/sygus_solver.h
```diff
@@ -0,0 +1,182 @@
...
+   * (2) there is a sygus conjecture that has been added as an assertion
+   * internally to this SMT engine, and there have been further calls such that
+   * the asserted conjecture is no longer up-to-date.
+   *
+   * This method should be called when new sygus constraints are asserted and
+   * when functions-to-synthesize are declared. This function pops a user
+   * context if we are in incremental mode and the sygus conjecture was
+   * previously not stale.
+   */
+  bool setSygusConjectureStale();
```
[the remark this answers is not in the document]
ajreynol: Fixed, this happened due to not updating this branch after a merge.

src/smt/sygus_solver.cpp
nafur: `getSynthSolutions` still exists in `smt_engine.cpp`...

PR #4893, opened 2020-08-14
(cad solver) Use the current model as initial assignment

nafur:
This PR implements a first naive way to employ the linear model (obtained from the nonlinear extension) to guide the initial sampling within the cad solver.


src/theory/arith/nl/cad/cdcac.h
```diff
@@ -57,6 +57,12 @@ class CDCAC
   /** Collect variables from the constraints and compute a variable ordering. */
   void computeVariableOrdering();
 
+  /**
+   * Extract an initial assignment from the given model.
```
ajreynol, line 61: Can you make the doc mention ran_variable?  This method is computing the model for that variable?  And storing it internally?

src/theory/arith/nl/cad/cdcac.h
```diff
@@ -81,6 +87,14 @@ class CDCAC
    */
   std::vector<CACInterval> getUnsatIntervals(std::size_t cur_variable) const;
 
+  /**
+   * Sample outside of the set of intervals.
+   * Uses a given initial value from mInitialAssignment if possible.
```
ajreynol, line 92: Document the return value.

PR #4898, opened 2020-08-14
Minor cleanup related to notifications

ajreynol:
This includes eliminating TheoryBV's call to eqNotifyNewEqClass and fixing an issue with string's eqNotifyNewEqClass method, which was registering constant integers. 

It also removes some unnecessary methods in Theory.

### DIFF
--- a/src/theory/bv/bv_subtheory_core.cpp
+++ b/src/theory/bv/bv_subtheory_core.cpp
@@ -102,8 +102,11 @@ void CoreSolver::preRegister(TNode node) {
       }
   } else {
     d_equalityEngine.addTerm(node);
-    // register with the extended theory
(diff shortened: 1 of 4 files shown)


src/theory/bv/bv_subtheory_core.cpp
```diff
@@ -102,6 +102,11 @@ void CoreSolver::preRegister(TNode node) {
       }
   } else {
     d_equalityEngine.addTerm(node);
+    // Register with the extended theory, for context-dependent simplification.
+    // Notice we do this for registered terms not internally generated
```
4tXJ7f, line 106:
```suggestion
    // Notice we do this for registered terms but not internally generated
```

PR #4901, opened 2020-08-16
Merge initialization steps in TheoryModelBuilder

ajreynol:
Currently when constructing models, we traverse the equality engine of the model 3 times during initialization.  This PR merges these 3 traversals.

This refactoring is necessary to update model building for the "centralized" approach for equality reasoning, where it will be important to traverse the equality engine of the model in a careful way (to skip irrelevant terms).

The PR also makes a few minor optimizations for e.g. reducing map lookups, and adds more documentation.


src/theory/theory_model_builder.cpp
```diff
@@ -389,124 +388,197 @@ bool TheoryEngineModelBuilder::buildModel(TheoryModel* tm)
...
-    typeConstSet.setTypeEnumeratorProperties(&tep);
-  }
-
-  // AJR: build ordered list of types that ensures that base types are
-  // enumerated first.
-  // (I think) this is only strictly necessary for finite model finding +
-  // parametric types instantiated with uninterpreted sorts, but is probably
-  // a good idea to do in general since it leads to models with smaller term
-  // sizes.
+  // An order list of types, such that T1 comes before T2 if T1 is a "component
```
yoni206, line 411:
```suggestion
  // An ordered list of types, such that T1 comes before T2 if T1 is a "component
```

src/theory/theory_model_builder.cpp
```diff
@@ -389,124 +388,197 @@ bool TheoryEngineModelBuilder::buildModel(TheoryModel* tm)
...
-      // Record as rep if this node was specified as a representative
-      if (tm->d_reps.find(n) != tm->d_reps.end())
+
+      // For each term n in this equivalence class:
+      // (1) Register its assignable subterms,
+      // (2) Compute whether it is a constant or assigned representative,
+      // (3) If we don't have a constant representative, compute information
+      // regarding how we will assign values.
+
+      // (1) Add assignable subterms, which ensures that e.g. models for
```
yoni206, line 489: Is this (1) the same as the (1) above? If so, I would try to avoid writing it twice in different words. Maybe move all the text to above, and here just write (1)? Same for the other numbers before.

src/theory/theory_model_builder.cpp
```diff
@@ -389,124 +388,197 @@ bool TheoryEngineModelBuilder::buildModel(TheoryModel* tm)
...
+        continue;
+      }
+      assignable = true;
+      if (!computeAssigners)
+      {
+        // we don't compute assigners, skip
+        continue;
+      }
+      // process the assignment exclusion set for term n
+      // was it processed as a slave of a group?
```
yoni206, line 539: what does this line mean? where is "slave" defined?

yoni206 (changes requested):

src/theory/theory_model_builder.cpp
ajreynol: Good point, I'll clean this up.

yoni206 (approved):
Sorry to keep this hanging. All was addressed, just forgot to approve.

PR #4903, opened 2020-08-17
(proof-new) Add the strings proof constructor

ajreynol:
This is the method for converting strings InferInfo into instructions for building ProofNode.  Notice that this is done as a standalone module, and thus the theory of strings uses Inferences only, not PfRule. 

The next step will be to integrate this utility into InferenceManager.


ajreynol:
This doesnt compile until the proof equality engine is merged (https://github.com/CVC4/CVC4/pull/4881).

ajreynol:
Additionally depends on https://github.com/CVC4/CVC4/pull/4962.

src/theory/strings/infer_proof_cons.cpp
```diff
@@ -0,0 +1,1043 @@
...
+  if (d_lazyFactMap.find(fact) != d_lazyFactMap.end())
+  {
+    Trace("strings-ipc-debug") << "...duplicate!" << std::endl;
+    return;
+  }
+  Node symFact = CDProof::getSymmFact(fact);
+  if (!symFact.isNull())
+  {
+    if (d_lazyFactMap.find(symFact) != d_lazyFactMap.end())
+    {
```
4tXJ7f, line 53: Combine `if` statements?

src/theory/strings/infer_proof_cons.cpp
```diff
@@ -0,0 +1,1043 @@
...
+                             TheoryProofStepBuffer& psb,
+                             bool& useBuffer)
+{
+  // by default, don't use the buffer
+  useBuffer = false;
+  // Must flatten children with respect to AND to be ready to explain.
+  // We store the index where each flattened vector begins, since some
+  // explanations are grouped together using AND.
+  size_t expIndex = 0;
+  std::map<size_t, size_t> startExpIndex;
```
4tXJ7f, line 76: Why isn't this a `vector<size_t>`?

src/theory/strings/infer_proof_cons.cpp
```diff
@@ -0,0 +1,1043 @@
...
+    case Inference::EXTF_N:
+    case Inference::EXTF_D:
+    case Inference::EXTF_D_N:
+    case Inference::I_CONST_CONFLICT:
+    case Inference::UNIT_CONST_CONFLICT:
+    {
+      if (!ps.d_children.empty())
+      {
+        std::vector<Node> exps;
+        exps.insert(exps.end(), ps.d_children.begin(), ps.d_children.end() - 1);
```
4tXJ7f, line 125:
```suggestion
        std::vector<Node> exps(ps.d_children.begin(), ps.d_children.end() - 1);
```

src/theory/strings/infer_proof_cons.cpp
```diff
@@ -0,0 +1,1043 @@
...
+    case Inference::EXTF_EQ_REW:
+    case Inference::INFER_EMP:
+    {
+      // the last child is the predicate we are operating on, move to front
+      Node src = ps.d_children[ps.d_children.size() - 1];
+      std::vector<Node> expe;
+      if (ps.d_children.size() > 1)
+      {
+        expe.insert(expe.end(), ps.d_children.begin(), ps.d_children.end() - 1);
+      }
```
4tXJ7f, line 150:
```suggestion
      std::vector<Node> expe(ps.d_children.begin(), ps.d_children.end() - 1);
```

src/theory/strings/infer_proof_cons.cpp
```diff
@@ -0,0 +1,1043 @@
...
+    case Inference::SSPLIT_VAR_PROP:
+    case Inference::SSPLIT_CST:
+    case Inference::SSPLIT_VAR:
+    {
+      Trace("strings-ipc-core") << "Generate core rule for " << infer
+                                << " (rev=" << isRev << ")" << std::endl;
+      // All of the above inferences have the form:
+      //   <explanation for why t and s have the same prefix/suffix> ^
+      //   t = s ^
+      //  <length constraint>?
```
4tXJ7f, line 193:
```suggestion
      //   <length constraint>?
```

src/theory/strings/infer_proof_cons.cpp
```diff
@@ -0,0 +1,1043 @@
...
+            lenConstraint.insert(lenConstraint.end(),
+                                 ps.d_children.begin() + mainEqIndex + 1,
+                                 ps.d_children.end());
+          }
+        }
+      }
+      else
+      {
+        if (nchild >= 1)
+        {
```
4tXJ7f, line 227: Combine into `else if`?

src/theory/strings/infer_proof_cons.cpp
```diff
@@ -0,0 +1,1043 @@
...
+      childrenSRew.insert(childrenSRew.end(),
+                          ps.d_children.begin(),
+                          ps.d_children.begin() + mainEqIndex);
+      Node mainEqSRew =
+          psb.tryStep(PfRule::MACRO_SR_PRED_ELIM, childrenSRew, {});
+      if (CDProof::isSame(mainEqSRew, mainEq))
+      {
+        Trace("strings-ipc-core") << "...undo step" << std::endl;
+        // the rule added above was not necessary
+        psb.popStep();
```
4tXJ7f, line 258: Couldn't that just be automatically done in `TheoryProofStepBuffer`?

src/theory/strings/infer_proof_cons.cpp
```diff
@@ -0,0 +1,1043 @@
...
+                                   MethodId::RW_REWRITE_EQ_EXT))
+        {
+          Trace("strings-ipc-core") << "Transformed to " << conc
+                                    << " via pred transform" << std::endl;
+          // success
+          useBuffer = true;
+          Trace("strings-ipc-core") << "...success!" << std::endl;
+        }
+        // Otherwise, note that EMP rules conclude ti = "" where
+        // t1 ++ ... ++ tn == "". However, these are very rare applied, let
```
4tXJ7f, line 312:
```suggestion
        // t1 ++ ... ++ tn == "". However, these are very rarely applied, let
```

src/theory/strings/infer_proof_cons.cpp
```diff
@@ -0,0 +1,1043 @@
...
+      if (conc.getKind() != OR)
+      {
+        Assert(false) << "Expected OR conclusion for " << infer;
+      }
+      else
+      {
+        ps.d_rule = PfRule::SPLIT;
+        Assert(ps.d_children.empty());
+        ps.d_args.push_back(conc[0]);
+      }
```
4tXJ7f, line 544:
```suggestion
      Assert(conc.getKind() == OR) << "Expected OR conclusion for " << infer;
      ps.d_rule = PfRule::SPLIT;
      Assert(ps.d_children.empty());
      ps.d_args.push_back(conc[0]);
```

src/theory/strings/infer_proof_cons.cpp
```diff
@@ -0,0 +1,1043 @@
...
+      ps.d_rule = PfRule::STRING_CODE_INJ;
+      Assert(conc.getKind() == OR && conc.getNumChildren() == 3
+             && conc[2].getKind() == EQUAL);
+      ps.d_args.push_back(conc[2][0]);
+      ps.d_args.push_back(conc[2][1]);
+    }
+    break;
+    // ========================== unit injectivity
+    case Inference::UNIT_INJ: { ps.d_rule = PfRule::STRING_SEQ_UNIT_INJ;
+    }
```
4tXJ7f, line 664:
```suggestion
    case Inference::UNIT_INJ: { ps.d_rule = PfRule::STRING_SEQ_UNIT_INJ;
    }
```

Formatting seems off here.

src/theory/strings/infer_proof_cons.cpp
```diff
@@ -0,0 +1,1043 @@
...
+    break;
+    // ========================== prefix conflict
+    case Inference::PREFIX_CONFLICT:
+    {
+      Trace("strings-ipc-prefix") << "Prefix conflict..." << std::endl;
+      std::vector<Node> eqs;
+      for (const Node e : ps.d_children)
```
4tXJ7f, line 671:
```suggestion
      for (const Node& e : ps.d_children)
```

src/theory/strings/infer_proof_cons.cpp
```diff
@@ -0,0 +1,1043 @@
...
+        {
+          // not sure how to use this assumption
+          Assert(false) << "Unexpected premise " << e << " for " << infer;
+        }
+      }
+      if (eqs.empty())
+      {
+        break;
+      }
+      // connect via transitivity?
```
4tXJ7f, line 727:
```suggestion
      // connect via transitivity
```

Or why is there a question mark?

src/theory/strings/infer_proof_cons.cpp
```diff
@@ -0,0 +1,1043 @@
...
+        Node rcurr =
+            polarity ? catom[1] : nm->mkNode(REGEXP_COMPLEMENT, catom[1]);
+        reis.push_back(rcurr);
+        Node mem = nm->mkNode(STRING_IN_REGEXP, catom[0], rcurr);
+        reiChildren.push_back(mem);
+        reiChildrenOrig.push_back(c);
+      }
+      // go back and justify each premise
+      bool successChildren = true;
+      for (unsigned i = 0, nchild = reiChildren.size(); i < nchild; i++)
```
4tXJ7f, line 795:
```suggestion
      for (size_t i = 0, nchild = reiChildren.size(); i < nchild; i++)
```

src/theory/strings/infer_proof_cons.cpp
```diff
@@ -0,0 +1,1043 @@
...
+          break;
+        }
+      }
+      if (!successChildren)
+      {
+        break;
+      }
+      Node mem = psb.tryStep(PfRule::RE_INTER, reiChildren, {});
+      // Node rei = reis.size() == 1 ? reis[0] : nm->mkNode(REGEXP_INTER, reis);
+      // Node mem = nm->mkNode(STRING_IN_REGEXP, x, rei);
```
4tXJ7f, line 812:
```suggestion
```

src/theory/strings/infer_proof_cons.cpp
```diff
@@ -0,0 +1,1043 @@
...
+
+Node InferProofCons::convertTrans(Node eqa,
+                                  Node eqb,
+                                  TheoryProofStepBuffer& psb)
+{
+  if (eqa.getKind() != EQUAL || eqb.getKind() != EQUAL)
+  {
+    return Node::null();
+  }
+  for (unsigned i = 0; i < 2; i++)
```
4tXJ7f, line 976:
```suggestion
  for (uint32_t i = 0; i < 2; i++)
```

src/theory/strings/infer_proof_cons.cpp
```diff
@@ -0,0 +1,1043 @@
...
+  {
+    Node eqaSym = i == 0 ? eqa[1].eqNode(eqa[0]) : eqa;
+    for (unsigned j = 0; j < 2; j++)
```
4tXJ7f, line 979:
```suggestion
    for (uint32_t j = 0; j < 2; j++)
```

4tXJ7f (changes requested):
Just some minor comments.

src/theory/strings/infer_proof_cons.cpp
ajreynol: It can be, this must have been leftover from a previous design.

src/theory/strings/infer_proof_cons.cpp
ajreynol: This would lead to proof checking failures (due to wrongly applied SPLIT) instead of a debug assertion failure + an application of STRING_TRUST.  This behavior is intentional here.

src/theory/strings/infer_proof_cons.cpp
ajreynol: TheoryProofStepBuffer is not necessarily tied to CDProof.  Essentially, this code is here because the strings proof constructor is using a CDProof, which handles some inference steps automatically (those that return true for CDProof::isSame), but other uses of TheoryProofStepBuffer could be different.

src/theory/strings/infer_proof_cons.cpp
4tXJ7f: I think that it would be good to define a special function for cases like this that will fail depending on how pedantic proof checking is.

ajreynol: This could go in a place like TheoryProofStepBuffer::tryStep. Note that it would be against the philosophy of our infrastructure to allow a bogus ProofNode to be generated (analogous to why we disallow non-well-typed Node).

src/theory/strings/infer_proof_cons.cpp
ajreynol: Strangely, this is clang-formatted.

src/theory/strings/infer_proof_cons.cpp
ajreynol: Actually, looking at this again ,this is a fairly specific case of proof reconstruction where we have a target conclusion (conc) and are trying to find its proof.  I think adding generic utilities for this would be overkill at this point, since this is the only place where a kind of "proof inference" is done.

4tXJ7f (approved):
LGTM, thanks for the updates. Before merging, could you add a comment regarding https://github.com/CVC4/CVC4/pull/4903#discussion_r498191738 that explains why it is ok to ignore the assertion in production?

PR #4904, opened 2020-08-17
(cad solver) Add a partial check method.

nafur:
This PR extends the CAD-based solver to enable partial checks. Essentially, we only collect the first interval that is excluded for the first variable and return that one as a lemma. This does not leave a lot of choice on "how partial" the check should be, but it is fairly easy to implement and does not add additional overhead.
It also fixes some confusion in `excluding_interval_to_lemma`...


src/theory/arith/nl/cad/cdcac.h
```diff
@@ -132,7 +132,8 @@ class CDCAC
    * UNSAT and an infeasible subset can be extracted from the returned covering.
    * Implements Algorithm 2.
    */
-  std::vector<CACInterval> getUnsatCover(std::size_t cur_variable = 0);
+  std::vector<CACInterval> getUnsatCover(std::size_t cur_variable = 0,
```
ajreynol, line 135: Can you document the new argument?

nafur:
> I assume this is not connected to NonLinearExtension yet?

No, not yet.

PR #4905, opened 2020-08-17
Refactor functions that print commands (Part 2)

abdoo8080:
This PR is a step towards migrating commands to the Term/Sort level. It replaces the dynamic casts for printing commands with direct calls to corresponding functions. Those functions now take node level arguments instead of commands to make them available for internal code.


src/printer/tptp/tptp_printer.cpp
```diff
@@ -91,6 +82,210 @@ void TptpPrinter::toStream(std::ostream& out, const UnsatCore& core) const
...
+                                       const std::string& flag,
+                                       SExpr sexpr) const
+{
+  printUnknownCommand(out, "set-option");
+}
+
+void TptpPrinter::toStreamCmdGetOption(std::ostream& out,
+                                       const std::string& flag) const
+{
+  printUnknownCommand(out, "get-option");
```
ajreynol, line 246: Can you omit any overrides that just print unknown?

abdoo8080: Done!

src/printer/cvc/cvc_printer.cpp
```diff
@@ -139,7 +139,7 @@ void CvcPrinter::toStream(
   if(n.getMetaKind() == kind::metakind::CONSTANT) {
     switch(n.getKind()) {
     case kind::BITVECTOR_TYPE:
-      out << "BITVECTOR(" << n.getConst<BitVectorSize>().d_size << ")";
+      out << "BITVECTOR(" << n.getConst<BitVectorSize>().d_size << ')';
```
ajreynol, line 142: Why this change?

src/smt/command.cpp
```diff
@@ -430,6 +492,23 @@ Command* CheckSatCommand::clone() const
 
 std::string CheckSatCommand::getCommandName() const { return "check-sat"; }
 
+void CheckSatCommand::toStream(std::ostream& out,
+                               int toDepth,
+                               bool types,
+                               size_t dag,
+                               OutputLanguage language) const
+{
+  if (d_expr.isNull())
```
ajreynol, line 501: I think I'd prefer to make `Printer::toStreamCmdCheckSat` robust to whether the expression was null and avoid the case split here.

src/printer/cvc/cvc_printer.cpp
abdoo8080: I have a personal preference for characters. Though, it looks like the change affected code unrelated to my PR. Reverted.

src/smt/command.cpp
```diff
@@ -105,6 +105,32 @@ std::ostream& operator<<(std::ostream& out, BenchmarkStatus status)
   }
 }
 
+std::vector<Node> exprVectorToNodes(const std::vector<Expr>& exprs)
```
ajreynol, line 108: Can you mark this with `// !!!!! temporary until commands are migrated` ?

PR #4907, opened 2020-08-18
Changes assertion (about maximum set cardinality) to an exception.

### LINKED ISSUE #4374: Fatal failure at theory/sets/cardinality_extension.cpp:991 (no option)
Hi,
For the following formula:
~~~~
(declare-fun st5 () (Set Int))
(assert (> (card st5) (* 79 (* 564 79 564 564) 564 (* 564 79 564 564))))
(check-sat)
~~~~

CVC4 throws out a fatal failure:
~~~~
Fatal failure within void CVC4::theory::sets::CardinalityExtension::mkModelValueElementsFor(CVC4::theory::Valuation&, CVC4::Node, std::vector<CVC4::NodeTemplate<true> >&, const std::map<CVC4::NodeTemplate<true>, CVC4::NodeTemplate<true> >&, CVC4::theory::TheoryModel*) at /home/peisen/test/tofuzz/CVC4/src/theory/sets/cardinality_extension.cpp:991
Check failure

 v.getConst<Rational>() <= 0x7fffffffffffffffL
Exceeded LONG_MAX in sets model
Aborted (core dumped
~~~~

OS: Ubuntu 16.04
Commit: ad7907adf

nafur:
Changes the assertion that checks for the maximum cardinality of set models to an exception, following #4374.
Also cleans up the code around it: previously, the `Rational` was checked against `LONG_MAX`, converted to `std::uint32_t` and then stored into an `unsigned`. Now we use `std::uint32_t` all the way.
Fixes #4374.


src/theory/sets/cardinality_extension.cpp
```diff
@@ -997,9 +997,15 @@ void CardinalityExtension::mkModelValueElementsFor(
...
-      Assert(v.getConst<Rational>() <= LONG_MAX)
-          << "Exceeded LONG_MAX in sets model";
-      unsigned vu = v.getConst<Rational>().getNumerator().toUnsignedInt();
+      if (v.getConst<Rational>() > UINT32_MAX)
+      {
+        std::stringstream ss;
+        ss << "The model for " << eqc << " was computed to have cardinality "
+           << v << ". We only allow sets up to cardinality " << UINT32_MAX
+           << ".";
+        throw Exception(ss.str());
```
ajreynol, line 1006: I believe we should use `LogicException` here.  There is an analogous piece of code in strings that should do the same thing: https://github.com/CVC4/CVC4/blob/master/src/theory/strings/theory_strings.cpp#L346, would be good to coordinate these two.

PR #4909, opened 2020-08-18
Standardize collectModelInfo and theory-specific collectRelevantTerms

ajreynol:
This is work towards a configurable approach to equality engine management.  This PR does not change any behavior, it only reorganizes the code.

This PR introduces the standard template for `collectModelInfo`, which isolates the usage of equality engine in relation to the model.  In the future, theories will be encouraged to use the standard template for `collectModelInfo` and override `collectRelevantTerms`/`collectModelValues` only.  This is to allow custom theory-independent modifications to building models (e.g. don't assert equality engine to model if we are using a centralized approach).

This PR standardizes TheoryArrays and TheoryDatatypes custom implementation of `collectRelevantTerms` as work towards using the standard template for `collectModelInfo`.  Notice this required separating two portions of a loop in TheoryArrays::collectModelInfo which was doing two different things (collecting arrays and relevant terms).

### DIFF
--- a/src/theory/datatypes/theory_datatypes.cpp
+++ b/src/theory/datatypes/theory_datatypes.cpp
@@ -1522,9 +1522,9 @@ bool TheoryDatatypes::collectModelInfo(TheoryModel* m)
   Trace("dt-model") << std::endl;
   printModelDebug( "dt-model" );
   Trace("dt-model") << std::endl;
-  
+
   std::set<Node> termSet;
-  
+
   // Compute terms appearing in assertions and shared terms, and in inferred equalities
   computeRelevantTerms(termSet);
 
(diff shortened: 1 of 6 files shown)


src/theory/datatypes/theory_datatypes.cpp
```diff
@@ -2250,12 +2250,13 @@ Node TheoryDatatypes::mkAnd( std::vector< TNode >& assumptions ) {
   }
 }
 
-void TheoryDatatypes::getRelevantTerms( std::set<Node>& termSet ) {
+void TheoryDatatypes::computeRelevantTerms( std::set<Node>& termSet,
```
4tXJ7f, line 2253: Apply ClangFormat.

src/theory/theory.h
```diff
@@ -203,11 +197,30 @@ class Theory {
...
+                                    bool includeShared = true) const;
+  /**
+   * Helper function for computeRelevantTerms
+   */
+  void collectTerms(TNode n,
+                    std::set<Kind>& irrKinds,
+                    std::set<Node>& termSet) const;
+  /**
+   * Same as above, but with empty irrKinds. This version can be overridden
+   * by the theory, e.g. by restricting or extended the set of terms returned
```
4tXJ7f, line 211:
```suggestion
   * by the theory, e.g. by restricting or extending the set of terms returned
```

PR #4913, opened 2020-08-18
(proof-new) Add the SMT proof post processor

ajreynol:
This PR adds the proof post-processor, which is the final authority on putting together the overall proof.  It additionally implements lazy pedantic failures and statistics.

This PR also corrects a subtle bug in the elimination of SUBS which requires a TRUE_INTRO/FALSE_INTRO in certain cases.


src/smt/proof_post_processor.cpp
```diff
@@ -568,5 +583,101 @@ bool ProofPostprocessCallback::addToTransChildren(Node eq,
...
+}
+
+bool ProofPostprocessFinalCallback::shouldUpdate(ProofNode* pn)
+{
+  PfRule r = pn->getRule();
+  if (r == PfRule::THEORY_REWRITE)
+  {
+    Trace("final-pf-hole") << "hole: " << r << " : " << pn->getResult()
+                           << std::endl;
+  }
```
HanielB, line 616: wrap this on a `Trace.isOn`

src/smt/proof_post_processor.h
```diff
@@ -119,12 +119,82 @@ class ProofPostprocessCallback : public ProofNodeUpdaterCallback
    * @return The conclusion of the TRANS step.
    */
   Node addProofForTrans(const std::vector<Node>& tchildren, CDProof* cdp);
+  /**
+   * Add proof for substitution step. Some substitutions are derived based
+   * on viewing a formula as a Boolean assigned (see MethodId::SB_LITERAL for
```
HanielB, line 124:
```suggestion
   * on viewing a formula as a Boolean assignment (see MethodId::SB_LITERAL for
```

PR #4914, opened 2020-08-18
[Regressions] Do not test `--check-proofs` anymore

4tXJ7f:
In preparation of removing the old proof module, this commit changes the
regression runner to not add the flag `--check-proofs` anymore when
running the regressions.


ajreynol:
I'm wondering if we also need to change https://github.com/CVC4/CVC4/pull/4914/files#diff-0be9eff83fa81a0e2eee83d4f5a03de3L330 ?  This appears to check whether the regression runs --check-proofs.  Perhaps we should also remove `--check-proofs` and `--no-check-proofs` from all regressions?

PR #4918, opened 2020-08-19
Make sets and strings solver states inherit from TheoryState

ajreynol:
This is towards the new standard for theory solvers.

This PR makes the custom states of sets and strings inherit from the standard base class TheoryState.  It also makes a minor change to InferenceManager/SolverState to make sets more in line with the plan for a standard base class InferenceManager.

Followup PRs will establish the official TheoryState classes for all other theories (which in most cases will be an instance of the base class).


src/theory/sets/inference_manager.h
```diff
@@ -109,6 +109,9 @@ class InferenceManager
   /** Have we sent lem as a lemma in the current user context? */
   bool hasLemmaCached(Node lem) const;
 
+  /** Set conf is a conflict node to be sent on the output channel.  */
```
4tXJ7f, line 112: It sounds a bit like two sentences got mixed up here. How about:

```suggestion
  /** 
   * Send conflict.
   * @param conf The conflict node to be sent on the output channel
   */
```

PR #4919, opened 2020-08-19
Split QuantElimSolver from SmtEngine

ajreynol:
Towards refactoring SmtEngine / converting Expr -> Node.


src/smt/quant_elim_solver.cpp
```diff
@@ -0,0 +1,106 @@
...
+  children.push_back(q.getKind() == EXISTS ? q[1] : q[1].negate());
+  children.push_back(n_attr);
+  Node ne = nm->mkNode(EXISTS, children);
+  Trace("smt-qe-debug") << "Query for quantifier elimination : " << ne
+                        << std::endl;
+  Assert(ne.getNumChildren() == 3);
+  // We consider this to be an entailment check, which also avoids incorrect
+  // status reporting (see SmtEngineState::d_expectedStatus).
+  Result r =
+      d_smtSolver.checkSatisfiability(as, std::vector<Node>{ne}, false, true);
```
nafur, line 64: Here, you changed the `inUnsatCore` flag to `false`. Just to make sure that you intended to do that...

nafur (approved):
Just one minor thing, otherwise LGTM.

src/smt/quant_elim_solver.cpp
ajreynol: Yes, it made no sense why inUnsatCore would be true.  I've also commented on why the other flag is true (which was also strange to me).

PR #4920, opened 2020-08-19
Add TheoryState objects to each Theory

ajreynol:
This initializes all theories with a TheoryState object (apart from `bool` and `builtin` which do not require one).

Two additional theories are known to require special state objects: TheoryArith, which has a custom way of detecting when in conflict, and TheoryQuantifiers, which can leverage a special state object for the purposes of refactoring and splitting apart QuantifiersEngine further.  All other theories use default TheoryState objects.

Notice this PR does not update the theories to use these states yet, it simply adds the objects.


src/theory/quantifiers/theory_quantifiers.h
```diff
@@ -19,14 +19,11 @@
...
 #define CVC4__THEORY__QUANTIFIERS__THEORY_QUANTIFIERS_H
 
-#include "context/context.h"
 #include "expr/node.h"
-#include "theory/output_channel.h"
 #include "theory/quantifiers/proof_checker.h"
 #include "theory/quantifiers/quantifiers_rewriter.h"
+#include "theory/quantifiers/quantifiers_state.h"
 #include "theory/theory.h"
-#include "theory/valuation.h"
```
4tXJ7f, line 28: It would be better to keep the includes for `context.h`, `output_channel.h`, and `valuation.h` because those are arguments to the constructor of `TheoryQuantifiers` for example.

src/theory/quantifiers/quantifiers_state.h
```diff
@@ -0,0 +1,40 @@
...
+namespace theory {
+namespace quantifiers {
+
+/**
+ * The quantifiers state.
+ */
+class QuantifiersState : public TheoryState
+{
+ public:
+  QuantifiersState(context::Context* c, context::UserContext* u, Valuation val);
```
4tXJ7f, line 32: This line looks a bit longer than 80 characters. ClangFormat?

src/theory/arith/theory_arith_private.h
```diff
@@ -324,17 +324,20 @@ class TheoryArithPrivate {
   /** This is a conflict that is magically known to hold. */
   void raiseBlackBoxConflict(Node bb);
 
-private:
+  /**
+   * Returns true iff a conflict has been raised. This method is public since
+   * it is needed by the ArithState class to know whether we are in conflict.
+   */
+  inline bool anyConflict() const
```
4tXJ7f, line 331:
```suggestion
  bool anyConflict() const
```

src/theory/quantifiers/quantifiers_state.h
ajreynol: It's actually exactly 80 characters :)

PR #4924, opened 2020-08-20
Simplify and fix care graph for ufHo

ajreynol:
We now separate APPLY_UF and HO_APPLY.  We do not generate care pairs based on comparing APPLY_UF terms with HO_APPLY terms, which led to type errors previously.  
Fixes #4990.


nafur (approved):
LGTM.

nafur (approved):
Even better :-)

PR #4934, opened 2020-08-22
Add the distributed model manager

ajreynol:
This class is responsible for model building when using a distributed approach for equality engines.

This PR defines the class but does not yet activate it in TheoryEngine.

This includes some model-specific things from TheoryEngine which will be migrated to this class on the next PR.


src/theory/model_manager_distributed.h
```diff
@@ -0,0 +1,60 @@
+/*********************                                                        */
+/*! \file ee_manager_distributed.h
```
4tXJ7f, line 2:
```suggestion
/*! \file model_manager_distributed.h
```

src/theory/model_manager.h
```diff
@@ -0,0 +1,114 @@
...
+ **
+ ** \brief Abstract management of models for TheoryEngine.
+ **/
+
+#include "cvc4_private.h"
+
+#ifndef CVC4__THEORY__MODEL_MANAGER__H
+#define CVC4__THEORY__MODEL_MANAGER__H
+
+#include <map>
```
4tXJ7f, line 20:
```suggestion
```

src/theory/model_manager_distributed.h
```diff
@@ -0,0 +1,60 @@
...
+ **
+ ** \brief Management of a distributed approach for model generation.
+ **/
+
+#include "cvc4_private.h"
+
+#ifndef CVC4__THEORY__MODEL_MANAGER_DISTRIBUTED__H
+#define CVC4__THEORY__MODEL_MANAGER_DISTRIBUTED__H
+
+#include <map>
```
4tXJ7f, line 20:
```suggestion
```

src/theory/model_manager_distributed.cpp
```diff
@@ -0,0 +1,76 @@
...
+  meec->push();
+
+  // Collect model info from the theories
+  Trace("model-builder") << "ModelManagerDistributed: Collect model info..."
+                         << std::endl;
+  // Consult each active theory to get all relevant information concerning the
+  // model, which includes both dump their equality information and assigning
+  // values.
+  for (TheoryId theoryId = theory::THEORY_FIRST; theoryId < theory::THEORY_LAST;
+       ++theoryId)
```
4tXJ7f, line 47: What about the order of the theories, e.g. arith before strings?

src/theory/model_manager.cpp
```diff
@@ -0,0 +1,177 @@
...
+ ** Top contributors (to current version):
+ **   Andrew Reynolds
+ ** This file is part of the CVC4 project.
+ ** Copyright (c) 2009-2020 by the authors listed in the file AUTHORS
+ ** in the top-level source directory) and their institutional affiliations.
+ ** All rights reserved.  See the file COPYING in the top-level source
+ ** directory for licensing information.\endverbatim
+ **
+ ** \brief Management of a distributed approach for equality engines over
+ ** all theories.
```
4tXJ7f, line 13:
```suggestion
 ** \brief Abstract management of models for TheoryEngine.
```

src/theory/model_manager.cpp
```diff
@@ -0,0 +1,177 @@
...
+
+  // prepare the model, which is specific to the manager
+  if (!prepareModel())
+  {
+    Trace("model-builder") << "ModelManager: fail prepare model" << std::endl;
+    return false;
+  }
+
+  // now, finish building the model
+  d_modelBuiltSuccess = finishBuildModel();
```
4tXJ7f, line 94: The method `finishBuildModel()` seems to essentially be a one-liner. Maybe inline that function here?

src/theory/model_manager.h
```diff
@@ -0,0 +1,114 @@
...
+namespace CVC4 {
+
+class TheoryEngine;
+
+namespace theory {
+
+/**
+ * A base class for managing models. Its main feature is to implement a
+ * buildModel command, which can be specific e.g. to the kind of equality
+ * engine management mode we are using.
```
4tXJ7f, line 35: I am not quite following that comment. `buildModel` is (currently?) not actually specific to the kind of equality engine management.

src/theory/model_manager.h
```diff
@@ -0,0 +1,114 @@
...
+  /**
+   * Post process model, which is used as a way of each theory adding additional
+   * information to the model after successfully building a model.
+   */
+  void postProcessModel(bool incomplete);
+  /** Get a pointer to model object maintained by this class. */
+  theory::TheoryModel* getModel();
+  //------------------------ finer grained control over model building
+  /**
+   * Prepare model, which the manager-specific method for setting up the
```
4tXJ7f, line 68:
```suggestion
   * Prepare model, which is the manager-specific method for setting up the
```

src/theory/model_manager.cpp
ajreynol: It's actually plan to be called manually by the model-based combination method.

src/theory/model_manager.h
ajreynol: buildModel depends on prepareModel, which is overriden by ModelManagerDistributed.  I'll clarify this comment.

src/theory/model_manager_distributed.cpp
ajreynol: This behavior is preserved in this commit.  This relies on the same thing we've relied on before (regarding the list of theories). I'll add a comment here to mention this.

PR #4937, opened 2020-08-24
Add a few basic extensions for equality engine

ajreynol:
This includes a standard method for safe explanations and the option to disable all trigger terms.


src/theory/uf/equality_engine.h
```diff
@@ -80,31 +80,41 @@ class EqualityEngine : public context::ContextNotifyObj {
...
   EqualityEngine* d_masterEqualityEngine;
 
-public:
-
+ public:
   /**
    * Initialize the equality engine, given the notification class.
+   *
+   * @param constantTriggers Whether we treat constants as trigger terms
+   * @param anyTermTriggers Whether we use any terms as triggers
```
4tXJ7f, line 88: Why not have an `enum TriggerMode` with something like `NONE`, `EXPLICIT_ONLY`, `CONSTANTS_AND_EXPLICIT`?

src/theory/uf/equality_engine.h
```diff
@@ -767,6 +782,22 @@ class EqualityEngine : public context::ContextNotifyObj {
   void explainPredicate(TNode p, bool polarity, std::vector<TNode>& assertions,
                         EqProof* eqp = nullptr) const;
 
+  //--------------------------- standard safe explanation methods
+  /**
+   * Explain literal, add its explanation to assumptions. This method does not
+   * add duplicates to assumptions. It additionally contains assertions for
+   * whether literal does hold in this class. If lit is a disequality, it
```
4tXJ7f, line 789:
```suggestion
   * add duplicates to assumptions. It requires that the literal
   * holds in this class. If lit is a disequality, it
```

(I was confused about "assertions" referring to C++ assertions here.

src/theory/uf/equality_engine.h
ajreynol: I don't think this is worthwhile yet, and will cause a lot of conflicts in upstream branches unfortunately.  Perhaps after centralEe and proof-new are merged we could reconsider this?

4tXJ7f: Yeah, I am fine with deferring this to later.

PR #4938, opened 2020-08-24
Do not use relevance during non-linear check model

ajreynol:
This led to a model soundness issue in rare cases where a relevant literal was dropped due to an entailment check by an irrelevant literal.


nafur (approved):
LGTM.

PR #4939, opened 2020-08-24
Add the combination engine

ajreynol:
This adds the combination engine, which is the module of TheoryEngine which implements the `combineTheories` method and owns the various components of theory combination, which includes equality engine manager, model manager, and the "shared solver" (to come later).  It will have two variants, `CombinationCareGraph` and `CombinationModelBased`, the former is added with this PR.

FYI @barrettcw 

The next PR will connect this module to TheoryEngine and remove a few existing methods from TheoryEngine, as they are implemented in the modules of this class.


src/theory/combination_care_graph.h
```diff
@@ -0,0 +1,53 @@
...
+ **
+ ** \brief Management of a care graph based approach for theory combination.
+ **/
+
+#include "cvc4_private.h"
+
+#ifndef CVC4__THEORY__COMBINATION_CARE_GRAPH__H
+#define CVC4__THEORY__COMBINATION_CARE_GRAPH__H
+
+#include <map>
```
4tXJ7f, line 20:
```suggestion
#include <vector>
```

src/theory/combination_care_graph.h
```diff
@@ -0,0 +1,53 @@
...
+#include <memory>
```
4tXJ7f, line 21:
```suggestion
```

src/theory/combination_care_graph.h
```diff
@@ -0,0 +1,53 @@
...
+
+namespace CVC4 {
+
+class TheoryEngine;
+
+namespace theory {
+
+/**
+ * Manager for doing theory combination using care graphs. This is typically
+ * done via a distributed equalty engine architecture.
```
4tXJ7f, line 33:
```suggestion
 * done via a distributed equality engine architecture.
```

src/theory/combination_care_graph.h
```diff
@@ -0,0 +1,53 @@
...
+ */
+class CombinationCareGraph : public CombinationEngine
+{
+ public:
+  CombinationCareGraph(TheoryEngine& te,
+                       const std::vector<Theory*>& paraTheories);
+  ~CombinationCareGraph();
+
+  /** Build model */
```
4tXJ7f, line 42:
```suggestion
```

src/theory/combination_engine.h
```diff
@@ -0,0 +1,122 @@
...
+ **
+ ** \brief Abstract interface for theory combination.
+ **/
+
+#include "cvc4_private.h"
+
+#ifndef CVC4__THEORY__COMBINATION_ENGINE__H
+#define CVC4__THEORY__COMBINATION_ENGINE__H
+
+#include <map>
```
4tXJ7f, line 20:
```suggestion
#include <vector>
```

src/theory/combination_care_graph.cpp
```diff
@@ -0,0 +1,93 @@
...
+  }
+
+  Trace("combineTheories")
+      << "TheoryEngine::combineTheories(): care graph size = "
+      << careGraph.size() << std::endl;
+
+  // Now add splitters for the ones we are interested in
+  CareGraph::const_iterator care_it = careGraph.begin();
+  CareGraph::const_iterator care_it_end = careGraph.end();
+
```
4tXJ7f, line 52:
```suggestion
```

src/theory/combination_care_graph.cpp
```diff
@@ -0,0 +1,93 @@
...
+  prop::PropEngine* propEngine = d_te.getPropEngine();
+  for (; care_it != care_it_end; ++care_it)
+  {
+    const CarePair& carePair = *care_it;
+
```
4tXJ7f, line 57:
```suggestion
  for (const CarePair& carePair : careGraph)
  {
```

src/theory/combination_care_graph.cpp
ajreynol: This is necessary in the for loop below.

PR #4940, opened 2020-08-25
Connect combination engine to theory engine

ajreynol:
This connects the implementation of `CombinationEngine` into `TheoryEngine`. By default, the combination engine of theory engine is `CombinationCareGraph`.

This PR also consolidates and simplifies how models are built, note that:
- The `TheoryModel` object no longer tracks whether it is built, instead that is the job of `ModelManager`,
- The `TheoryModelBuilder` object is no longer responsible for calling TheoryEngine's collect model info method.
- Quantifiers engine uses a simpler interface for ensuring that TheoryEngine's model is built.

This PR also makes some minor simplifications to TheoryEngine from the centralEe branch.

Note that no significant behavior changes are intended in this PR.


src/theory/theory_engine.cpp
```diff
@@ -131,27 +131,32 @@ std::string getTheoryString(theory::TheoryId id)
...
+  else
   {
-    d_curr_model_builder = d_quantEngine->getModelBuilder();
-    d_curr_model = d_quantEngine->getModel();
-  } else {
-    d_curr_model = new theory::TheoryModel(
-        d_userContext, "DefaultModel", options::assignFunctionValues());
-    d_aloc_curr_model = true;
+    AlwaysAssert(false) << "TheoryEngine::finishInit: theory combination mode "
+                        << options::tcMode() << " not supported";
```
4tXJ7f, line 159:
```suggestion
    Unimplemented() << "TheoryEngine::finishInit: theory combination mode "
                    << options::tcMode() << " not supported";
```

src/theory/theory_model_builder.h
```diff
@@ -46,19 +46,22 @@ class TheoryEngineModelBuilder
...
+   * for building a model. This class is responsible simply for ensuring
+   * that all equivalence classes of the equality engine of m are assigned
+   * constants.
    *
    * This constructs the model m, via the following steps:
-   * (1) call TheoryEngine::collectModelInfo,
-   * (2) builder-specified pre-processing,
-   * (3) find the equivalence classes of m's
+   * (1) builder-specified pre-processing,
+   * (2) find the equivalence classes of m's
```
4tXJ7f, line 59:
```suggestion
   * (2) find the equivalence classes of m
```

src/theory/theory_engine.cpp
```diff
@@ -311,34 +305,45 @@ void TheoryEngine::preRegister(TNode preprocessed) {
...
       // Pre-register the terms in the atom
-      Theory::Set theories = NodeVisitor<PreRegisterVisitor>::run(d_preRegistrationVisitor, preprocessed);
+      Theory::Set theories = NodeVisitor<PreRegisterVisitor>::run(
+          d_preRegistrationVisitor, preprocessed);
       theories = Theory::setRemove(THEORY_BOOL, theories);
-      // Remove the top theory, if any more that means multiple theories were involved
+      // Remove the top theory, if any more that means multiple theories were
+      // involved
       bool multipleTheories = Theory::setRemove(Theory::theoryOf(preprocessed), theories);
+#ifdef CVC4_ASSERTIONS
```
4tXJ7f, line 316: I prefer to use `if (Configuration::isAssertionBuild()) { ... }` instead of using `#ifdefs` because it is more idiomatic.

src/theory/theory_engine.cpp
```diff
@@ -572,27 +580,9 @@ void TheoryEngine::check(Theory::Effort effort) {
...
-        {
-          // Do post-processing of model from the theories (used for THEORY_SEP
-          // to construct heap model)
-          postProcessModel(d_curr_model);
-          // also call the model builder's post-process model
-          d_curr_model_builder->postProcessModel(d_incomplete.get(),
-                                                 d_curr_model);
-        }
-      }
+      // Do post-processing of model from the theories (used for THEORY_SEP
```
4tXJ7f, line 583:
```suggestion
      // Do post-processing of model from the theories (e.g. used for THEORY_SEP
```

src/theory/theory_model_builder.h
ajreynol: actually this line is broken in two, this says "m's equality engine".

PR #4941, opened 2020-08-25
Eliminating spurious replay of define funs expansion when checking unsat cores

HanielB:
Doing it via commands being added to the `coreChecker` SMT engine is not necessary since we can directly added assertions after expansion from the original SMT engine.


HanielB:
FYI @yoni206

PR #4943, opened 2020-08-26
[API] Fix Python Examples

4tXJ7f:
When testing the API examples, Python examples were not included. This
commit changes that and fixes multiple minor issues that came up once
the tests were enabled:

- It adds `Solver::supportsFloatingPoint()` as an API method that
  returns whether CVC4 is configured to support floating-point numbers
  or not (this is useful for failing gracefully when floating-point
  support is not available, e.g. in the case of our floating-point
  example).
- It updates the `expections.py` example to use the new API.
- It fixes the `sygus-fun.py` example. The example was passing a _set_
  of non-terminals to `Solver::mkSygusGrammar()` but the order in which
  the non-terminals are passed in matters because the first non-terminal
  is considered to be the starting terminal. The commit also updates the
  documentation of that function.
- It fixes the Python API for datatypes. The `__getitem__` function had
  a typo and the `datatypes.py` example was crashing as a result.

Additionally, the commit:
- Renames `Solver::getRoundingmodeSort()` to 
  `Solver::getRoundingModeSort()` for consistency.
- Adds support for `RoundTowardNegative` in the Python API.


src/api/cvc4cpp.cpp
```diff
@@ -3793,6 +3801,8 @@ Term Solver::mkNegZero(uint32_t exp, uint32_t sig) const
 Term Solver::mkRoundingMode(RoundingMode rm) const
 {
   CVC4_API_SOLVER_TRY_CATCH_BEGIN;
+  CVC4_API_CHECK(Configuration::isBuiltWithSymFPU())
+      << "Expected CVC4 to be compiled with SymFPU support";
```
aniemetz, line 3805: Can you add this check also to:
https://github.com/CVC4/CVC4/blob/master/src/api/cvc4cpp.cpp#L3313
https://github.com/CVC4/CVC4/blob/master/src/api/cvc4cpp.cpp#L3278?
I noticed that it's missing there, too.

PR #4944, opened 2020-08-26
(new theory) Update TheoryUF to new interface

ajreynol:
This updates TheoryUF to use the 4 check callbacks instead of implementing check, and uses the official TheoryState object instead of its `context::CDO<bool> d_conflict` field.

It also makes a minor change to collectModelValues for const and to preNotifyFact to include an isInternal flag.

### DIFF
--- a/src/theory/uf/theory_uf.cpp
+++ b/src/theory/uf/theory_uf.cpp
@@ -67,8 +67,6 @@ TheoryUF::TheoryUF(context::Context* c,
   }
   // indicate we are using the default theory state object
   d_theoryState = &d_state;
-
-  d_needsSharedTermEqFacts = false;
 }
 
 TheoryUF::~TheoryUF() {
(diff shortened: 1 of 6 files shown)


src/theory/uf/theory_uf.cpp
```diff
@@ -126,79 +125,90 @@ static Node mkAnd(const std::vector<TNode>& conjunctions) {
...
-    } else {
-      d_equalityEngine->assertPredicate(atom, polarity, fact);
+    }
+    // don't need to assert if not producing models
+    if (!options::produceModels())
+    {
+      return true;
     }
   }
+  return false;
```
4tXJ7f, line 194:
```suggestion
  }
  // don't need to assert if not producing models
  return !options::produceModels();
```

ajreynol: Actually this "don't assert if produceModels is false" only applies to cardinality constraints, so the return statement would actually be inside the if block.

PR #4945, opened 2020-08-26
Add irrelevant kinds infrastructure to TheoryModel

ajreynol:
Currently, Theory is responsible for implementing a `computeRelevantTerms` method collects the union of:
(1) The terms appearing in its assertions and its shared terms,
(2) The set of additional terms the theory believes are relevant.

Currently, (1) is computed by an internal Theory method `computeRelevantTermsInternal`, and the overall `computeRelevantTerms` is overridden by the theory (for datatypes and arrays only) for also including (2).

The new plan is that Theory will only need to compute (2).  Computing (1) will be the job of `ModelManager::collectAssertedTerms` and the model manager will call `Theory::computeRelevantTerms` as an additional step prior to its calls to collect model info.  

This will make certain optimizations possible in theory combination (e.g. tracking asserted terms incrementally).

This PR adds the `ModelManager::collectAssertedTerms` method and also adds infrastructure for irrelevant kinds in the model object (which have an analogous interface as when "unevaluated" kinds are marked during initialization).  It does not connect the implementation yet.

FYI @barrettcw

### DIFF
--- a/src/theory/model_manager.cpp
+++ b/src/theory/model_manager.cpp
@@ -162,7 +162,6 @@ bool ModelManager::collectModelBooleanVariables()
   return true;
 }
 
-
 void ModelManager::collectAssertedTerms(TheoryId tid,
                                         std::set<Node>& termSet,
                                         bool includeShared) const
--- a/src/theory/model_manager.cpp
+++ b/src/theory/model_manager.cpp
@@ -162,5 +162,64 @@ bool ModelManager::collectModelBooleanVariables()
   return true;
 }
 
+
+void ModelManager::collectAssertedTerms(TheoryId tid,
+                                        std::set<Node>& termSet,
+                                        bool includeShared) const
+{
+  Theory* t = d_te.theoryOf(tid);
+  // Collect all terms appearing in assertions
+  context::CDList<Assertion>::const_iterator assert_it = t->facts_begin(),
+                                             assert_it_end = t->facts_end();
+  for (; assert_it != assert_it_end; ++assert_it)
+  {
+    collectTerms(tid, *assert_it, termSet);
+  }
+
+  if (includeShared)
+  {
+    // Add terms that are shared terms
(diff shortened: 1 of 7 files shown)


src/theory/model_manager.h
```diff
@@ -91,6 +91,22 @@ class ModelManager
    * @return true if we are in conflict.
    */
   bool collectModelBooleanVariables();
+  /**
+   * Collect asserted terms for theory with the given identifier, add to
+   * termSet.
+   *
+   * @param tid The theory whose assertions we are collecting
+   * @param termSet The set to add terms
```
4tXJ7f, line 99:
```suggestion
   * @param termSet The set to add terms to
```

src/theory/model_manager.cpp
```diff
@@ -162,5 +162,63 @@ bool ModelManager::collectModelBooleanVariables()
...
+    cur = visit.back();
+    visit.pop_back();
+    if (termSet.find(cur) != termSet.end())
+    {
+      // already visited
+      continue;
+    }
+    Kind k = cur.getKind();
+    // only add to term set if a relevant kind
+    if (irrKinds.find(k) == irrKinds.end())
```
4tXJ7f, line 210: Shouldn't we also make sure that `Theory::theoryOf(cur) == tid`? Also, aren't we traversing irrelevant kinds multiple times?

ajreynol: The former should always hold, I can add it as an assertion.

We are traversing terms with irrelevant kinds multiple times.  This isn't worth caching since all terms with irrelevant kinds are predicate symbols and cant be nested.  It's fairly unlikely that this will change in the future (I can't currently imagine any non-predicate symbols that are irrelevant).

4tXJ7f: After discussing this offline, this is ok because the children of an irrelevant kind are cached if they are visited multiple times.

PR #4948, opened 2020-08-26
Add the theory inference manager

ajreynol:
This class is a wrapper around OutputChannel and EqualityEngine.  It is preferred that the Theory use this interface when asserting "internal facts" to the equality engine, and for sending lemmas, conflicts and propagations on the output channel.

This class will be useful when trying new methods for theory combination, where all theories behavior can be modified in a standard way based on modifications to the base inference manager class.

### DIFF
--- a/src/theory/theory_inference_manager.cpp
+++ b/src/theory/theory_inference_manager.cpp
@@ -12,7 +12,7 @@
  ** \brief An inference manager for Theory
  **/
 
-#include "theory/inference_manager.h"
+#include "theory/theory_inference_manager.h"
 
 #include "theory/theory.h"
 #include "theory/uf/equality_engine.h"
--- a/src/theory/theory_inference_manager.cpp
+++ b/src/theory/theory_inference_manager.cpp
@@ -52,6 +52,7 @@ void TheoryInferenceManager::conflict(TNode conf)
 {
   if (!d_theoryState.isInConflict())
   {
+    d_theoryState.notifyInConflict();
     d_out.conflict(conf);
   }
 }
--- a/src/theory/theory_inference_manager.cpp
+++ b/src/theory/theory_inference_manager.cpp
@@ -1,0 +1,145 @@
+/*********************                                                        */
+/*! \file inference_manager.cpp
+ ** \verbatim
+ ** Top contributors (to current version):
+ **   Andrew Reynolds
+ ** This file is part of the CVC4 project.
+ ** Copyright (c) 2009-2020 by the authors listed in the file AUTHORS
+ ** in the top-level source directory) and their institutional affiliations.
+ ** All rights reserved.  See the file COPYING in the top-level source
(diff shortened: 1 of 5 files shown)


src/theory/theory_inference_manager.h
```diff
@@ -0,0 +1,148 @@
+/*********************                                                        */
+/*! \file inference_manager.h
```
4tXJ7f, line 2:
```suggestion
/*! \file theory_inference_manager.h
```

src/theory/theory_inference_manager.cpp
```diff
@@ -0,0 +1,147 @@
+/*********************                                                        */
+/*! \file inference_manager.cpp
```
4tXJ7f, line 2:
```suggestion
/*! \file theory_inference_manager.cpp
```

src/theory/theory_inference_manager.h
```diff
@@ -0,0 +1,148 @@
...
+ ** All rights reserved.  See the file COPYING in the top-level source
+ ** directory for licensing information.\endverbatim
+ **
+ ** \brief An inference manager for Theory
+ **/
+
+#include "cvc4_private.h"
+
+#ifndef CVC4__THEORY__INFERENCE_MANAGER_H
+#define CVC4__THEORY__INFERENCE_MANAGER_H
```
4tXJ7f, line 18:
```suggestion
#ifndef CVC4__THEORY__THEORY_INFERENCE_MANAGER_H
#define CVC4__THEORY__THEORY_INFERENCE_MANAGER_H
```

src/theory/theory_inference_manager.h
```diff
@@ -0,0 +1,148 @@
...
+   * explanations are SAT-context-dependent, this set is also
+   * SAT-context-dependent.
+   */
+  NodeSet d_keep;
+};
+
+}  // namespace theory
+}  // namespace CVC4
+
+#endif /* CVC4__THEORY__INFERENCE_MANAGER_H */
```
4tXJ7f, line 148:
```suggestion
#endif /* CVC4__THEORY__THEORY_INFERENCE_MANAGER_H */
```

src/theory/theory_inference_manager.h
```diff
@@ -0,0 +1,148 @@
...
+   * T-propagate literal lit, possibly encountered by equality engine,
+   * returns false if we are in conflict.
+   *
+   * Note that this is the preferred method to call on
+   * EqualityEngineNotify::eqNotifyTriggerPredicate and
+   * EqualityEngineNotify::eqNotifyTriggerTermEquality.
+   */
+  bool propagateLit(TNode lit);
+  /**
+   * Return an explanation for the literal represented by parameter n
```
4tXJ7f, line 82:
```suggestion
   * Return an explanation for the literal represented by parameter lit
```

PR #4950, opened 2020-08-27
(new theory) Update TheoryQuantifiers to the new interface

ajreynol:
TheoryQuantifiers is a theory that passes quantified formulas to QuantifiersEngine. This updates it to the new check template  (see https://github.com/CVC4/CVC4/pull/4929).

Also does some minor cleanup in the cpp file.

### DIFF
--- a/src/theory/quantifiers/theory_quantifiers.cpp
+++ b/src/theory/quantifiers/theory_quantifiers.cpp
@@ -141,10 +141,8 @@ void TheoryQuantifiers::postCheck(Effort level)
   getQuantifiersEngine()->check(level);
 }
 
-bool TheoryQuantifiers::preNotifyFact(TNode atom,
-                                      bool polarity,
-                                      TNode fact,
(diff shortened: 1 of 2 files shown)


src/theory/quantifiers/theory_quantifiers.cpp
```diff
@@ -29,12 +29,12 @@
 #include "theory/quantifiers_engine.h"
 #include "theory/valuation.h"
 
-using namespace std;
-using namespace CVC4;
 using namespace CVC4::kind;
```
HanielB, line 32: Weren't we trying to keep this explicit as well?

ajreynol: No, I like using it for files with heavy use of kind since it shrinks line sizes.

HanielB: OK, sounds good.

PR #4953, opened 2020-08-27
(new theory) Update TheoryFP to the new interface

ajreynol:
This updates the theory of floating points to the new interface (see #4929).

Notice that TheoryFP was not adding trigger terms to its equality engine (which should be done during `notifySharedTerm`), and thus was not propagating equalities between shared terms in combined theories.  This PR updates its notifySharedTerm method to the default one.

FYI @martin-cs

### DIFF
--- a/src/theory/fp/theory_fp.cpp
+++ b/src/theory/fp/theory_fp.cpp
@@ -1021,7 +1021,7 @@ Node TheoryFp::getModelValue(TNode var) {
 bool TheoryFp::collectModelInfo(TheoryModel* m)
 {
   std::set<Node> relevantTerms;
-  // Work out which variables are needed	
+  // Work out which variables are needed
   computeRelevantTerms(relevantTerms);
   // this override behavior to not assert equality engine
   return collectModelValues(m, relevantTerms);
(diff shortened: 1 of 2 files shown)


martin-cs (approved):
Thanks for this (and the notify).  Not notifying the central equality engine is definitely a fuck-up on my part -- thanks for catching it.

If I understand the architecture changes correctly, the loop of processing assigned atoms and the handling of conflicts has moved inside the main SMT engine, making the theory a bit more functional and less of a separate decision procedure.  As the (current) theory solver doesn't actually do much (apart from invoke it's equality solver) on normal checks, I don't see this as being an issue.

ajreynol:
Yes... btw the "main SMT engine" as you call it here is the "TheoryEngine", not to be confused with "SmtEngine" which manages the entire system.  And yes, the new changes basically give less responsibility to the developers of individual theory solvers. Instead of implementing a standalone check method, they now implement smaller callbacks.  Notice however the behavior of the system is still the same (for now), it's just that more behavior is standardized.  I am working on a branch to modify the internal architecture, will keep you posted if TheoryFP gives me any troubles (I don't expect it to).

martin-cs:
Apologies for the confusion.  Yes, TheoryFP should be fairly well behaved and do tell me if not!

If / as / when you are refactoring the TheoryEngine could you also produce a diagram version of how and when the various calls are made.  There was one in the old wiki and I found it invaluable when learning and explaining how the code works.

ajreynol:
Yep, I am planning to update the documentation at some point, new diagrams indeed would be helpful.

PR #4954, opened 2020-08-27
Add the buffered inference manager

ajreynol:
This class implements a highly common pattern of buffering facts and lemmas to send on the output channel.  It is planned that the inference managers of strings, sets, datatypes, (non-linear) arithmetic, sep, quantifiers will inherit from this class.

This PR adds basic calls to add lemmas on the output channel from InferenceManager.

It introduces a `Lemma` virtual class which arith::nl::NlLemma and strings::InferInfo will inherit from.

This is required to begin refactoring a flexible configurable strategy for non-linear arithmetic, and will make it easier to further develop towards a configurable approach for theory combination.

### DIFF
--- a/src/theory/inference_manager_buffered.h
+++ b/src/theory/inference_manager_buffered.h
@@ -113,7 +113,10 @@ class InferenceManagerBuffered : public TheoryInferenceManager
    *
    * This method is called to indicate this class should send a phase
    * requirement request to the output channel for literal lit to be
-   * decided with polarity pol.
+   * decided with polarity pol. The literal lit should be a SAT literal
+   * by the time that doPendingPhaseRequirements is called. Typically,
+   * lit is a literal that is a subformula of a pending lemma that is processed
+   * prior to sending the phase requirement.
    */
   void addPendingPhaseRequirement(Node lit, bool pol);
   /** Do pending facts
@@ -139,7 +142,10 @@ class InferenceManagerBuffered : public TheoryInferenceManager
    * cleared after this call.
    */
   void doPendingLemmas();
-  /** Do pending phase requirements */
+  /**
+   * Do pending phase requirements. Calls the output channel for all pending
+   * phase requirements and clears d_pendingReqPhase.
+   */
   void doPendingPhaseRequirements();
 
  protected:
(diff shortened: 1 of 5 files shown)


src/theory/inference_manager_buffered.h
```diff
@@ -0,0 +1,157 @@
...
+};
+
+/**
+ * The buffered inference manager.  This class implements standard methods
+ * for buffering facts, lemmas and phase requirements.
+ */
+class InferenceManagerBuffered : public TheoryInferenceManager
+{
+  typedef context::CDHashSet<Node, NodeHashFunction> NodeSet;
+
```
nafur, line 67:
```suggestion
```
I don't think we need this guy.

src/theory/theory_inference_manager.h
```diff
@@ -105,6 +105,11 @@ class TheoryInferenceManager
    * been provided in a custom way.
    */
   void trustedConflict(TrustNode tconf);
+  /** Send lemma lem with property p on the output channel. */
+  LemmaStatus lemma(TNode lem, LemmaProperty p = LemmaProperty::NONE);
+  /** Send (trusted) lemma lem with property p on the output channel. */
+  LemmaStatus trustedLemma(TrustNode tlem,
+                           LemmaProperty p = LemmaProperty::NONE);
```
nafur, line 112: I'd suggest to pass stuff by `const&`...
```suggestion
  LemmaStatus trustedLemma(const TrustNode& tlem,
                           LemmaProperty p = LemmaProperty::NONE);
```

src/theory/theory_inference_manager.cpp
```diff
@@ -111,6 +111,17 @@ TrustNode TheoryInferenceManager::explainConflictEqConstantMerge(TNode a,
                   << " mkTrustedConflictEqConstantMerge";
 }
 
+LemmaStatus TheoryInferenceManager::lemma(TNode lem, LemmaProperty p)
+{
+  return d_out.lemma(lem, p);
+}
+
+LemmaStatus TheoryInferenceManager::trustedLemma(TrustNode tlem,
+                                                 LemmaProperty p)
```
nafur, line 120:
```suggestion
LemmaStatus TheoryInferenceManager::trustedLemma(const TrustNode& tlem,
                                                 LemmaProperty p)
```
ditto

nafur (approved):
Just two minor issues, otherwise LGTM.

PR #4955, opened 2020-08-27
(proof-new) Make CombinationEngine proof producing

ajreynol:
Makes combination engine proof producing (for Boolean splits). Followup PRs will start to add proof production in TheoryEngine.

### DIFF
--- a/src/theory/combination_care_graph.cpp
+++ b/src/theory/combination_care_graph.cpp
@@ -22,8 +22,10 @@ namespace CVC4 {
 namespace theory {
 
 CombinationCareGraph::CombinationCareGraph(
-    TheoryEngine& te, const std::vector<Theory*>& paraTheories)
-    : CombinationEngine(te, paraTheories)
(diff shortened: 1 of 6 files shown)


src/theory/combination_care_graph.cpp
```diff
@@ -62,7 +64,19 @@ void CombinationCareGraph::combineTheories()
...
 
     Node split = equality.orNode(equality.notNode());
-    sendLemma(split, carePair.d_theory);
+    TrustNode tsplit;
+    if (isProofEnabled())
+    {
+      // make proof of splitting lemma
+      std::vector<Node> pfArgs;
+      pfArgs.push_back(equality);
+      tsplit = d_cmbsPg->mkTrustNode(split, PfRule::SPLIT, pfArgs);
```
HanielB, line 73:
```suggestion
      tsplit = d_cmbsPg->mkTrustNode(split, PfRule::SPLIT, {equality});
```

PR #4958, opened 2020-08-27
(cad-solver) Fixed excluding lemma generation.

nafur:
The lemma generation for partial cad checks had a number of issues that have been fixed in this PR.
The previous version had both soundness issues and a very naive approach to handling algebraic numbers.
This new version is sound (fingers crossed) and allows to construct more precise, but also more complex lemmas.
To avoid constructing very large lemmas, a (somewhat arbitrary) limit based on the size of coefficients was added.

### DIFF
--- a/src/theory/arith/nl/poly_conversion.h
+++ b/src/theory/arith/nl/poly_conversion.h
@@ -119,6 +119,13 @@ RealAlgebraicNumber node_to_ran(const Node& n, const Node& ran_variable);
  */
 poly::Value node_to_value(const Node& n, const Node& ran_variable);
 
+/**
+ * Give a rough estimate of the bitsize of the representation of `v`.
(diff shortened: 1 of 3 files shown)


src/theory/arith/nl/poly_conversion.h
```diff
@@ -98,7 +98,8 @@ Node value_to_node(const poly::Value& v, const Node& ran_variable);
  * )
  */
 Node excluding_interval_to_lemma(const Node& variable,
-                                 const poly::Interval& interval);
+                                 const poly::Interval& interval,
+                                 bool allowNonlinearLemma);
```
ajreynol, line 102: Can you document the new argument here?

nafur: See new commit.

PR #4960, opened 2020-08-28
Add ArithLemma and arith::InferenceManager

nafur:
This PR adds a new `ArithLemma` that is essentiall `NlLemma`, but derived from the new `theory::Lemma` and meant to be used all across the arithmetic theory.
Also, based on `theory::InferenceManagerBuffered` this PR adds `arith::InferenceManager` that shall become the sole interface between the arithmetic theory and the `OutputChannel`.


src/theory/arith/inference_manager.h
```diff
@@ -0,0 +1,94 @@
...
+ * buffered inference manager.
+ */
+class InferenceManager : public InferenceManagerBuffered
+{
+  using NodeSet = context::CDHashSet<Node, NodeHashFunction>;
+
+ public:
+  InferenceManager(TheoryArith& ta, ArithState& astate, ProofNodeManager* pnm);
+
+  /** Add a lemma (as pending lemma) to the underlying inference manager. */
```
ajreynol, line 50: I wouldn't call it the underlying inference manager, I'd call it "this inference manager".

src/theory/inference_manager_buffered.h
```diff
@@ -146,6 +146,11 @@ class InferenceManagerBuffered : public TheoryInferenceManager
    */
   void doPendingPhaseRequirements();
 
+  /** Returns the number of pending lemmas. */
+  std::size_t countPendingLemmas() const;
```
ajreynol, line 150: What about the name `numPendingLemmas` instead?

src/theory/arith/arith_lemma.h
```diff
@@ -0,0 +1,76 @@
...
+#include "theory/arith/nl/inference.h"
+#include "theory/inference_manager_buffered.h"
+#include "theory/output_channel.h"
+
+namespace CVC4 {
+namespace theory {
+namespace arith {
+
+/**
+ * The data structure for a single lemma to process by the arithmetic theory,
```
ajreynol, line 31: I'm a bit confused why this class is necessary? I was actually expecting that the existing NlLemma class would derive from Lemma.  That way these members (which are specific to non-linear) continues to live in nl/.  I think the current design allows this, right?

src/theory/arith/inference_manager.h
```diff
@@ -0,0 +1,94 @@
...
+  void addLemma(const Node& lemma, nl::Inference inftype);
+
+  /** Add a lemma (as waiting lemma). */
+  void addWaitingLemma(std::shared_ptr<ArithLemma> lemma);
+  /** Add a lemma (as waiting lemma). */
+  void addWaitingLemma(const ArithLemma& lemma);
+  /** Add a lemma (as waiting lemma). */
+  void addWaitingLemma(const Node& lemma, nl::Inference inftype);
+
+  /** Flush all waiting lemmas to the underlying inference manager (as pending lemmas). */
```
ajreynol, line 64: Does this move the waiting lemmas as pending lemmas? Or send them? 

Would it be easier to just have this method send all waiting lemmas?  It seems roundabout to do multiple steps of moving the lemmas around to me.

src/theory/arith/arith_lemma.h
nafur: As discussed offline: We now have `NlLemma` derive from `ArithLemma`.

src/theory/arith/inference_manager.h
nafur: As of now, this behaviour seems more in line with how I expect the new strategy for the nonlinear extension to work. I added a comment to make clearer what this method does.

src/theory/arith/arith_lemma.h
```diff
@@ -0,0 +1,76 @@
...
+  /** secant points to add
+   *
+   * A member (tf, d, c) in this vector indicates that point c should be added
+   * to the list of secant points for an application of a transcendental
+   * function tf for Taylor degree d. This is used for incremental linearization
+   * for underapproximation (resp. overapproximations) of convex (resp.
+   * concave) regions of transcendental functions. For details, see
+   * Cimatti et al., CADE 2017.
+   */
+  std::vector<std::tuple<Node, unsigned, Node> > d_secantPoint;
```
ajreynol, line 65: Remove this.

src/theory/arith/inference_manager.h
```diff
@@ -0,0 +1,127 @@
...
+ */
+class InferenceManager : public InferenceManagerBuffered
+{
+  using NodeSet = context::CDHashSet<Node, NodeHashFunction>;
+
+ public:
+  InferenceManager(TheoryArith& ta, ArithState& astate, ProofNodeManager* pnm);
+
+  /** Add a lemma (as pending lemma) to the this inference manager. */
+  void addLemma(std::shared_ptr<ArithLemma> lemma);
```
ajreynol, line 54: I prefer this name `addPendingArithLemma` here.

src/theory/arith/inference_manager.h
```diff
@@ -0,0 +1,127 @@
...
+  /** Add a lemma (as pending lemma) to the this inference manager. */
+  void addLemma(const ArithLemma& lemma);
+  /** Add a lemma (as pending lemma) to the this inference manager. */
+  void addLemma(const Node& lemma, nl::Inference inftype);
+
+  /** Add a lemma (as waiting lemma). */
+  void addWaitingLemma(std::shared_ptr<ArithLemma> lemma);
```
ajreynol, line 61: I think it'd be better if we had a `isWaiting` flag for the above methods.

src/theory/arith/inference_manager.h
```diff
@@ -0,0 +1,127 @@
...
+  void flushWaitingLemmas();
+
+  /** Add a conflict to the this inference manager. */
+  void addConflict(const Node& conf, nl::Inference inftype);
+
+  /** Returns the number of pending lemmas. */
+  std::size_t numWaitingLemmas() const;
+
+  /** Checks whether the given lemma is already present in the cache. */
+  virtual bool hasCachedLemma(TNode lem, LemmaProperty p) override
```
ajreynol, line 80: Move the implementation to cpp.

src/theory/arith/inference_manager.h
```diff
@@ -0,0 +1,127 @@
...
+    }
+    return TheoryInferenceManager::hasCachedLemma(lem, p);
+  }
+
+ protected:
+  /**
+   * Adds the given lemma to the cache. Returns true if it has not been in the
+   * cache yet.
+   */
+  virtual bool cacheLemma(TNode lem, LemmaProperty p) override
```
ajreynol, line 94: Move implementation to cpp.

ajreynol:
Also, I'm wondering the order of merges, can we merge https://github.com/CVC4/CVC4/pull/4990 before this one?

nafur:
> LGTM, thanks for the updates. Btw, I guess the plan is to move arith::nl::Inference to arith::Inference?

Yes, this will eventually happen. Maybe in the next PR :smile:

PR #4962, opened 2020-08-28
(proof-new) Updates to builtin proof checker

ajreynol:
This includes support for some trusted rules (whose use is to come). It also modifies THEORY_REWRITE so that it is a "trusted" rule, that is, it is not re-checked.  The reason for this is that TheoryRewriter is *not* deterministic. An example of non-determinism in TheoryRewriter are rules that introduce bound variables (e.g. quantifiers rewriter) since these bound variables are fresh each time it is invoked.  Non-deterministic theory rewriters cannot be rechecked, which was leading to failures on `proof-new` at proof check time.  The other way to fix this would be to cache TheoryRewriter responses, but then the checker would only be verifying that the caching was being done properly. This is not worthwhile.


src/theory/builtin/proof_checker.h
```diff
@@ -144,13 +140,21 @@ class BuiltinProofRuleChecker : public ProofRuleChecker
    */
   static void addMethodIds(std::vector<Node>& args, MethodId ids, MethodId idr);
 
+  /** get a TheoryId from a node, return false if we fail */
+  static bool getTheoryId(TNode n, TheoryId& tid);
```
4tXJ7f, line 144: Could consider returning a pair instead.

ajreynol: This is intentional actually, it makes the control flow simpler since you can have a condition that calls this method and then be ensured that under that condition tid is valid. This is a style we've been following for other utility methods.

PR #4963, opened 2020-08-28
(new theory) Update TheoryStrings to new standard

ajreynol:
This updates theory of strings to the new standard.

This makes strings use the standard template for check and collectModelInfo.  It also updates its inference manager to the standard and makes use of `assertFactInternal` for processing internal facts. This now enables `preNotifyFact` and `notifyFact` to be defined in TheoryStrings instead of inside inference manager, which is clearer and eliminates some dependencies within inference manager.

Note that the inference manager of strings for now inherits from `TheoryInferenceManager`. Further standardization will make it inherit from the new `InferenceManagerBuffered` class.

This design will be merged into `proof-new`, which also has significant changes to strings inference manager.


src/theory/strings/theory_strings.h
```diff
@@ -97,27 +97,35 @@ class TheoryStrings : public Theory {
...
   void preRegisterTerm(TNode n) override;
   /** Expand definition */
   TrustNode expandDefinition(Node n) override;
-  /** Check at effort e */
-  void check(Effort e) override;
-  /** needs check last effort */
+  //--------------------------------- standard check
+  /** Do we need a check call at last call effort? */
   bool needsCheckLastEffort() override;
+  /** Notify fact */
```
4tXJ7f, line 107:
```suggestion
```

src/theory/strings/theory_strings.h
```diff
@@ -97,27 +97,35 @@ class TheoryStrings : public Theory {
...
+  bool preNotifyFact(TNode atom,
+                     bool pol,
+                     TNode fact,
+                     bool isPrereg,
+                     bool isInternal) override;
+  /** Notify fact */
```
4tXJ7f, line 113:
```suggestion
```

src/theory/strings/inference_manager.cpp
```diff
@@ -308,27 +298,15 @@ void InferenceManager::doPendingFacts()
...
-    // external assertions, which enter as facts through sendAssumption.
-    d_keep.insert(facts);
-    d_keep.insert(exp);
     i++;
   }
   d_pending.clear();
 }
 
 void InferenceManager::doPendingLemmas()
 {
```
4tXJ7f: Discussed offline: This is an unrelated change that should be reverted.

src/theory/strings/theory_strings.cpp
```diff
@@ -654,25 +625,68 @@ TrustNode TheoryStrings::expandDefinition(Node node)
...
-
-    Trace("strings-assertion") << "get assertion: " << fact << endl;
-    d_im.sendAssumption(fact);
+    if (polarity && atom[1].getKind() == REGEXP_CONCAT)
+    {
+      Node eqc = d_equalityEngine->getRepresentative(atom[0]);
+      d_state.addEndpointsToEqcInfo(atom, atom[1], eqc);
+    }
   }
+  // process the conflict
```
4tXJ7f, line 662:
```suggestion
  // process pending conflicts due to reasoning about endpoints
```

PR #4964, opened 2020-08-28
Removes old proof code

HanielB:
This deletes much of the old proof code. Basically everything but the minimal necessary infra-structure for producing unsat cores. That includes dependency tracking in preprocessing, the prop engine proof and the unsat core computation code in the old proof manager.

It also cleans interfaces that were using old-proof-code specific constructs (such as `LemmaProofRecipe`). When possible or when it made sense standalone local proof production code was kept, active or not (such is in the equality engine and in the arithmetic solver).


src/preprocessing/passes/quantifier_macros.cpp
```diff
@@ -432,9 +439,10 @@ Node QuantifierMacros::simplify( Node n ){
           for( unsigned i=0; i<children.size(); i++ ){
             Node etc = TypeNode::getEnsureTypeCondition( children[i], tno[i] );
             if( etc.isNull() ){
-              //if this does fail, we are incomplete, since we are eliminating quantified formula corresponding to op, 
+              // if this does fail, we are incomplete, since we are eliminating
+              // quantified formula corresponding to op,
               //  and not ensuring it applies to n when its types are correct.
-              //Assert( false );
+              // Assert( false );
```
ajreynol, line 445: Delete?

src/proof/clause_id.h
```diff
@@ -28,6 +28,12 @@ namespace CVC4 {
  */
 typedef unsigned ClauseId;
 
+const ClauseId ClauseIdEmpty(-1);
```
ajreynol, line 31: What are these? Can you add a some quick (basic) documentation here? E.g. "Temporary, to do delete when X" or whatever.

test/unit/prop/cnf_stream_white.h
```diff
@@ -240,18 +241,17 @@ class CnfStreamWhite : public CxxTest::TestSuite {
   //              kind::ITE,
   //              d_nodeManager->mkVar(d_nodeManager->booleanType()),
   //              d_nodeManager->mkVar(d_nodeManager->integerType()),
-  //              d_nodeManager->mkVar(d_nodeManager->integerType())
-  //          ),
-  //          d_nodeManager->mkVar(d_nodeManager->integerType())
-  //                            ), false, false, RULE_INVALID, Node::null());
-  //
-  //}
+  //              d_nodeManager->mkVar(d_nodeManager->integerType())),
```
ajreynol, line 244: Delete this block

test/regress/regress1/quantifiers/dump-inst-proof.smt2
```diff
@@ -1,5 +1,5 @@
 ; REQUIRES: proof
-; COMMAND-LINE: --dump-instantiations --proof --print-inst-full
+; COMMAND-LINE: --dump-instantiations --produce-unsat-cores --print-inst-full
```
ajreynol, line 2: I like this change.  It didn't make much sense to use `--proof` for this feature.

src/theory/uf/equality_engine_types.h
```diff
@@ -63,20 +63,24 @@ static const EqualityEdgeId null_edge = (EqualityEdgeId)(-1);
...
-  /** Equality was merged to false, due to both sides of equality being a constant */
+  /** Terms were merged due to theory reasoning */
   MERGED_THROUGH_CONSTANTS,
-  /** (for proofs only) Equality was merged due to transitivity */
+  /** Terms were merged due to transitivity */
   MERGED_THROUGH_TRANS,
-
-  /** Reason types beyond this constant are theory specific reasons */
-  NUMBER_OF_MERGE_REASONS
+  /** Terms were merged due to arrays read-over-write */
```
ajreynol, line 78: These shouldn't have to be added here.  Can you try to delete these as well?  Or perhaps at least mark "//!!!!!!!! temporary".  I believe I can delete them if you need.

src/theory/theory.cpp
```diff
@@ -23,6 +23,7 @@
 
 #include "base/check.h"
 #include "expr/node_algorithm.h"
+#include "options/smt_options.h"
```
ajreynol, line 26: Why is this needed?

src/theory/arith/theory_arith_private.cpp
```diff
@@ -4716,7 +4719,8 @@ bool TheoryArithPrivate::rowImplicationCanBeApplied(RowIndex ridx, bool rowUp, C
       }
       Node implication = implied->externalImplication(explain);
       Node clause = flattenImplication(implication);
-      PROOF(if (d_containing.d_proofRecorder
+#if 0
```
ajreynol, line 4722: Delete?

src/theory/arith/theory_arith_private.cpp
```diff
@@ -4700,8 +4700,11 @@ bool TheoryArithPrivate::rowImplicationCanBeApplied(RowIndex ridx, bool rowUp, C
   if( !assertedToTheTheory && canBePropagated && !hasProof ){
     ConstraintCPVec explain;
 
-    PROOF(d_farkasBuffer.clear());
-    RationalVectorP coeffs = NULLPROOF(&d_farkasBuffer);
+#if 0
```
ajreynol, line 4703: Delete?

src/theory/arith/theory_arith_private.cpp
```diff
@@ -1923,7 +1924,8 @@ void TheoryArithPrivate::outputConflicts(){
       ++conflicts;
       Debug("arith::conflict") << "d_conflicts[" << i << "] " << conflict
                                << " has proof: " << hasProof << endl;
-      PROOF(if (d_containing.d_proofRecorder && confConstraint->hasFarkasProof()
+#if 0
```
ajreynol, line 1927: Delete?  @alex-ozdemir  may know best about this.  Wondering if this is required on proof-new.

src/theory/arith/constraint.h
```diff
@@ -877,9 +872,12 @@ class Constraint {
   }
 
   inline RationalVectorCP getFarkasCoefficients() const {
-    return NULLPROOF(getConstraintRule().d_farkasCoefficients);
+#if 0
```
ajreynol, line 875: Ditto.

src/theory/arith/constraint.cpp
```diff
@@ -1374,12 +1418,11 @@ void Constraint::impliedByFarkas(const ConstraintCPVec& a, RationalVectorCP coef
   AntecedentId antecedentEnd = d_database->d_antecedents.size() - 1;
 
   RationalVectorCP coeffsCopy;
-  if(PROOF_ON()){
-    Assert(coeffs != RationalVectorCPSentinel);
-    coeffsCopy = new RationalVector(*coeffs);
-  } else {
-    coeffsCopy = RationalVectorCPSentinel;
-  }
+#if 0
```
ajreynol, line 1421: Ditto.

src/theory/arith/constraint.cpp
```diff
@@ -1359,10 +1402,11 @@ void Constraint::impliedByFarkas(const ConstraintCPVec& a, RationalVectorCP coef
   Assert(negationHasProof() == nowInConflict);
   Assert(allHaveProof(a));
 
-  Assert(PROOF_ON() == (coeffs != RationalVectorCPSentinel));
-  // !PROOF_ON() => coeffs == RationalVectorCPSentinel
-  //  PROOF_ON() => coeffs->size() == a.size() + 1
-  Assert(!PROOF_ON() || coeffs->size() == a.size() + 1);
+#if 0
```
ajreynol, line 1405: Ditto.

src/theory/arith/constraint.cpp
```diff
@@ -1213,7 +1257,7 @@ void Constraint::impliedByUnate(ConstraintCP imp, bool nowInConflict){
   AntecedentId antecedentEnd = d_database->d_antecedents.size() - 1;
 
   RationalVectorP coeffs;
-  if(PROOF_ON()){
+#if 0
```
ajreynol, line 1260: Ditto.

src/theory/arith/constraint.cpp
```diff
@@ -701,109 +702,153 @@ bool Constraint::wellFormedFarkasProof() const {
   if(antecedent  == NullConstraint) { return false; }
 
 #if IS_PROOFS_BUILD
-  if(!PROOF_ON()){ return cr.d_farkasCoefficients == RationalVectorCPSentinel; }
-  Assert(PROOF_ON());
+  return cr.d_farkasCoefficients == RationalVectorCPSentinel;
+#if 0
```
ajreynol, line 706: Ditto.

src/proof/clause_id.h
HanielB: These are used to control registration of clauses in the old sat proof (which is used for unsat cores, so we have to keep it for now). I'll add a comment. As a rule of thumb, everything that is in `/proof` currently will be deleted in its current form.

src/theory/theory.cpp
HanielB: Because there is a call to `options::produceModels()` in the file. This was not necessary before because the inclusion-chain to `theory.h` had `smt_options.h` before (not sure where) but does not any longer.

HanielB:
Answering the general concerns:

- (1) : There was a lot of code in the arithmetic solver to make it proof producing that I was afraid of deleting. The one who would know better how it relates to the plan for `proof-new` would be @alex-ozdemir, so to be on the safe side I kept it while not executing. But we can go ahead and delete.

- (2) : I think this is not as problematic. The original formatting is strange and any modification leads to clang-format changing things significantly. I think it makes sense to just apply them. I can handle the merge as they will be with the code I've modified in proof-new.

src/theory/uf/equality_engine_types.h
HanielB: The array solver sends these reasons when it's asserting equalities. Previously it would dynamically generate these three, so there was some dedicated code for allowing for that. By making them static the handling of reasons is simpler.

ajreynol: just write temporary then, I'll remove on proof-new.

HanielB: OK, done.

src/theory/theory.cpp
ajreynol: Makes sense.

ajreynol:
Regarding the merge: I think we may need to collaborate.  I'm estimating there will be several non-trivial conflicts in code I've modified on proof-new.

alex-ozdemir:
An initial comment: the `ArithProofRecorder` class (and its use-sites) should be removed. The old machinery used them, but the new machinery does not (since the new machinery builds proofs at lemma/conflict time, it doesn't need a separate persistent storage for this information: this information ends up in an `EagerProofGenerator`.

HanielB:
Good to know, @alex-ozdemir! I've update the PR to remove the ArithProofRecorder and moreover the dead code I had kept.

src/smt/term_formula_removal.cpp
```diff
@@ -402,13 +407,14 @@ Node RemoveTermFormulas::replace(TNode node, bool inQuant, bool inTerm) const {
...
 
-// returns true if the children of node should be considered nested terms 
+// returns true if the children of node should be considered nested terms
 bool RemoveTermFormulas::hasNestedTermChildren( TNode node ) {
-  return theory::kindToTheoryId(node.getKind())!=theory::THEORY_BOOL && 
-         node.getKind()!=kind::EQUAL && node.getKind()!=kind::SEP_STAR && 
-         node.getKind()!=kind::SEP_WAND && node.getKind()!=kind::SEP_LABEL && 
-         node.getKind()!=kind::BITVECTOR_EAGER_ATOM;
-         // dont' worry about FORALL or EXISTS (handled separately)
+  return theory::kindToTheoryId(node.getKind()) != theory::THEORY_BOOL
```
ajreynol, line 412: Note that this block of code has been moved on proof-new.  Thus you will get a merge conflict unless you revert this formatting.

HanielB:
I would like to have the approval of @alex-ozdemir as well. All the deletions were somewhat straightforward except for the arithmetic part.

src/theory/arith/callbacks.cpp
```diff
@@ -87,37 +87,25 @@ void FarkasConflictBuilder::reset(){
   d_consequent = NullConstraint;
   d_constraints.clear();
   d_consequentSet = false;
-  PROOF(d_farkas.clear());
```
alex-ozdemir, line 90: Ah, I see now that I should have been more clear earlier about how arith records farkas coefficients, and what we need to keep/remove.

Prior to this PR, arith maintained farkas coefficients in two locations:
a) In data structures in the `src/theory/arith` directory (e.g. `FarkasConflictBuilder` and `ConstraintRule`).
b) In data structures in the `src/proof` directory (e.g. `ArithProofRecorder`).

The former may be context-dependent. The latter were persistent.

We can remove the latter. They're replace by eager proof generators, etc., in the new proof system. **We cannot remove the former**: the new system uses them.

Generally, in the core of arithmetic, this means that we should treat proofs as "always on" and not trim code that is guarded by compile-time proof flags. Thus, this command should be kept, just `d_farkas.clear()`.

On the other hand, we can remove all of `src/proof/arith_*`, and direct references to such machinery in `src/theory/arith/`.

src/theory/arith/callbacks.cpp
```diff
@@ -87,37 +87,25 @@ void FarkasConflictBuilder::reset(){
...
 }
 
 /* Adds a constraint to the constraint under construction. */
 void FarkasConflictBuilder::addConstraint(ConstraintCP c, const Rational& fc){
-  Assert(
-      !PROOF_ON()
-      || (!underConstruction() && d_constraints.empty() && d_farkas.empty())
-      || (underConstruction() && d_constraints.size() + 1 == d_farkas.size()));
-  Assert(PROOF_ON() || d_farkas.empty());
+  Assert(d_farkas.empty());
```
alex-ozdemir, line 95:
```suggestion
  Assert((!underConstruction() && d_constraints.empty() && d_farkas.empty())
      || (underConstruction() && d_constraints.size() + 1 == d_farkas.size()));
```

Same as above, we should treat proofs as "always on", but otherwise not modify the core of how arith computes and stores farkas coefficients.

src/theory/arith/callbacks.cpp
```diff
@@ -87,37 +87,25 @@ void FarkasConflictBuilder::reset(){
...
 
   if(d_consequent == NullConstraint){
     d_consequent = c;
   } else {
     d_constraints.push_back(c);
   }
-  PROOF(d_farkas.push_back(fc););
-  Assert(!PROOF_ON() || d_constraints.size() + 1 == d_farkas.size());
-  Assert(PROOF_ON() || d_farkas.empty());
+  Assert(d_farkas.empty());
```
alex-ozdemir, line 103:
```suggestion
  d_farkas.push_back(fc);
  Assert(d_constraints.size() + 1 == d_farkas.size());
```
ditto

src/theory/arith/callbacks.cpp
```diff
@@ -87,37 +87,25 @@ void FarkasConflictBuilder::reset(){
...
 
 void FarkasConflictBuilder::addConstraint(ConstraintCP c, const Rational& fc, const Rational& mult){
   Assert(!mult.isZero());
-  if(PROOF_ON() && !mult.isOne()){
-    Rational prod = fc * mult;
-    addConstraint(c, prod);
-  }else{
-    addConstraint(c, fc);
-  }
+  addConstraint(c, fc);
```
alex-ozdemir, line 108:
```suggestion
  if(!mult.isOne()){
    Rational prod = fc * mult;
    addConstraint(c, prod);
  }else{
    addConstraint(c, fc);
  }
```
ditto

src/theory/arith/callbacks.cpp
```diff
@@ -132,7 +120,6 @@ void FarkasConflictBuilder::makeLastConsequent(){
     ConstraintCP last = d_constraints.back();
     d_constraints.back() = d_consequent;
     d_consequent = last;
```
alex-ozdemir, line 122:
```suggestion
    d_consequent = last;
    std::swap( d_farkas.front(), d_farkas.back() );
```

src/theory/arith/callbacks.cpp
```diff
@@ -144,15 +131,11 @@ void FarkasConflictBuilder::makeLastConsequent(){
 ConstraintCP FarkasConflictBuilder::commitConflict(){
   Assert(underConstruction());
   Assert(!d_constraints.empty());
-  Assert(
-      !PROOF_ON()
-      || (!underConstruction() && d_constraints.empty() && d_farkas.empty())
-      || (underConstruction() && d_constraints.size() + 1 == d_farkas.size()));
-  Assert(PROOF_ON() || d_farkas.empty());
+  Assert(d_farkas.empty());
```
alex-ozdemir, line 134:
```suggestion
  Assert((!underConstruction() && d_constraints.empty() && d_farkas.empty())
      || (underConstruction() && d_constraints.size() + 1 == d_farkas.size()));
```
ditto

src/theory/arith/callbacks.cpp
```diff
@@ -144,15 +131,11 @@ void FarkasConflictBuilder::makeLastConsequent(){
...
   Assert(d_consequentSet);
 
   ConstraintP not_c = d_consequent->getNegation();
-  RationalVectorCP coeffs = NULLPROOF(&d_farkas);
+  RationalVectorCP coeffs = nullptr;
```
alex-ozdemir, line 138:
```suggestion
  RationalVectorCP coeffs = &d_farkas;
```

src/theory/arith/constraint.cpp
```diff
@@ -648,8 +647,7 @@ ConstraintCP ConstraintDatabase::getAntecedent (AntecedentId p) const {
 
 
 void ConstraintRule::print(std::ostream& out) const {
-  
-  RationalVectorCP coeffs = NULLPROOF(d_farkasCoefficients);
+  RationalVectorCP coeffs = nullptr;
```
alex-ozdemir, line 650:
```suggestion
  RationalVectorCP coeffs = d_farkasCoefficients;
```

src/theory/arith/constraint.cpp
```diff
@@ -701,109 +699,9 @@ bool Constraint::wellFormedFarkasProof() const {
...
-  Debug("constraints::wffp") << "final sum: " << lhs << " <= " << rhs << endl;
-  // 0 = lhs <= rhs < 0
-  return (lhs.isNull() || (Constant::isMember(lhs) && Constant(lhs).isZero()))
-         && rhs.sgn() < 0;
-
-#else  /* IS_PROOFS_BUILD */
-  return true;
+  return cr.d_farkasCoefficients == RationalVectorCPSentinel;
 #endif /* IS_PROOFS_BUILD */
+  return true;
```
alex-ozdemir, line 704: Ditto: assume that `IS_PROOFS_BUILD` is true here.

src/theory/arith/constraint.cpp
```diff
@@ -1213,18 +1110,7 @@ void Constraint::impliedByUnate(ConstraintCP imp, bool nowInConflict){
...
-    Rational first(sgns.first);
-    Rational second(sgns.second);
-
-    coeffs = new RationalVector();
-    coeffs->push_back(first);
-    coeffs->push_back(second);
-  } else {
-    coeffs = RationalVectorPSentinel;
-  }
+  coeffs = RationalVectorPSentinel;
```
alex-ozdemir, line 1113:
```suggestion
  std::pair<int, int> sgns = unateFarkasSigns(getNegation(), imp);

  Rational first(sgns.first);
  Rational second(sgns.second);

  coeffs = new RationalVector();
  coeffs->push_back(first);
  coeffs->push_back(second);
```

src/theory/arith/constraint.cpp
```diff
@@ -1359,10 +1245,7 @@ void Constraint::impliedByFarkas(const ConstraintCPVec& a, RationalVectorCP coef
   Assert(negationHasProof() == nowInConflict);
   Assert(allHaveProof(a));
 
-  Assert(PROOF_ON() == (coeffs != RationalVectorCPSentinel));
-  // !PROOF_ON() => coeffs == RationalVectorCPSentinel
-  //  PROOF_ON() => coeffs->size() == a.size() + 1
-  Assert(!PROOF_ON() || coeffs->size() == a.size() + 1);
+  Assert(coeffs == RationalVectorCPSentinel);
```
alex-ozdemir, line 1248:
```suggestion
  Assert(coeffs != RationalVectorCPSentinel);
  Assert(coeffs->size() == a.size() + 1);
```

src/theory/arith/constraint.h
```diff
@@ -339,7 +338,7 @@ struct ConstraintRule {
     , d_proofType(pt)
     , d_antecedentEnd(antecedentEnd)
   {
-    Assert(PROOF_ON() || coeffs == RationalVectorCPSentinel);
+    Assert(coeffs == RationalVectorCPSentinel);
```
alex-ozdemir, line 341:
```suggestion
```

src/theory/arith/constraint.h
```diff
@@ -789,10 +788,6 @@ class Constraint {
       ConstraintP constraint = crp->d_constraint;
       Assert(constraint->d_crid != ConstraintRuleIdSentinel);
       constraint->d_crid = ConstraintRuleIdSentinel;
-
-      PROOF(if (crp->d_farkasCoefficients != RationalVectorCPSentinel) {
-        delete crp->d_farkasCoefficients;
-      });
```
alex-ozdemir, line 795: Ditto: we should keep this branch.

src/theory/arith/constraint.h
```diff
@@ -876,10 +871,8 @@ class Constraint {
     return getConstraintRule().d_antecedentEnd;
   }
 
-  inline RationalVectorCP getFarkasCoefficients() const {
-    return NULLPROOF(getConstraintRule().d_farkasCoefficients);
-  }
-  
+  inline RationalVectorCP getFarkasCoefficients() const { return nullptr; }
```
alex-ozdemir, line 874:
```suggestion
  inline RationalVectorCP getFarkasCoefficients() const { return getConstraintRule().d_farkasCoefficients; }
```

src/theory/arith/linear_equality.cpp
```diff
@@ -510,11 +510,11 @@ void LinearEqualityModule::propagateBasicFromRow(ConstraintP c){
   RowIndex ridx = d_tableau.basicToRowIndex(basic);
 
   ConstraintCPVec bounds;
-  RationalVectorP coeffs = NULLPROOF(new RationalVector());
+  RationalVectorP coeffs = nullptr;
```
alex-ozdemir, line 513:
```suggestion
  RationalVectorP coeffs = new RationalVector();
```

src/theory/arith/theory_arith.cpp
```diff
@@ -41,7 +41,6 @@ TheoryArith::TheoryArith(context::Context* c,
       d_internal(
           new TheoryArithPrivate(*this, c, u, out, valuation, logicInfo, pnm)),
       d_ppRewriteTimer("theory::arith::ppRewriteTimer"),
-      d_proofRecorder(nullptr),
```
alex-ozdemir, line 44: Great! This is an example of a reference to the `src/proof` tree, and, thus, it should be removed :+1: .

src/theory/arith/theory_arith_private.cpp
```diff
@@ -1923,35 +1924,6 @@ void TheoryArithPrivate::outputConflicts(){
...
-        }
-
-        Assert(conflict.getNumChildren() == pf.d_farkasCoefficients->size());
-        if (confConstraint->hasSimpleFarkasProof()
-            && confConstraint->getNegation()->isPossiblyTightenedAssumption())
-        {
-          d_containing.d_proofRecorder->saveFarkasCoefficients(
-              conflictInFarkasCoefficientOrder, pf.d_farkasCoefficients);
-        }
-      })
```
alex-ozdemir, line 1954: Great! An example of code that's just feeding information into `ArithProofRecorder`. This should be removed :+1: .

src/theory/arith/theory_arith_private.cpp
```diff
@@ -4699,9 +4669,7 @@ bool TheoryArithPrivate::rowImplicationCanBeApplied(RowIndex ridx, bool rowUp, C
 
   if( !assertedToTheTheory && canBePropagated && !hasProof ){
     ConstraintCPVec explain;
-
-    PROOF(d_farkasBuffer.clear());
-    RationalVectorP coeffs = NULLPROOF(&d_farkasBuffer);
+    RationalVectorP coeffs = nullptr;
```
alex-ozdemir, line 4672:
```suggestion
    d_farkasBuffer.clear();
    RationalVectorP coeffs = &d_farkasBuffer;
```

src/theory/arith/theory_arith_private.cpp
```diff
@@ -4716,38 +4684,6 @@ bool TheoryArithPrivate::rowImplicationCanBeApplied(RowIndex ridx, bool rowUp, C
...
-        Assert(conflictInFarkasCoefficientOrder.getNumChildren()
-               == coeffs->size());
-        if (std::all_of(explain.begin(), explain.end(), [](ConstraintCP c) {
-              return c->isAssumption() || c->hasIntTightenProof();
-            }))
-        {
-          d_containing.d_proofRecorder->saveFarkasCoefficients(
-              conflictInFarkasCoefficientOrder, coeffs);
-        }
-      })
```
alex-ozdemir, line 4750: Great! This should also be removed :+1:

src/theory/arith/callbacks.cpp
HanielB: I see, thanks! At first I had put most of this code under `#if 0 ... #endif` statements because it looked important. :) I have applied the changes you suggested, PTAL.

src/theory/arith/constraint.cpp
HanielB: In the code block executed when `IS_PROOFS_BUILD` holds, the statement 
```
  return cr.d_farkasCoefficients == RationalVectorCPSentinel;
```
would be executed only if proofs are off, which we should assume is never the case, if I understand correctly. In that case I assume that the code that should survive is the one after, in the old code, `Assert(PROOF_ON());`?

HanielB:
@alex-ozdemir I noticed after the changes that unless everything involving d_farkasCoefficients is guarded with the compile flag we will have issues. So in the interest of simplicity I left its declaration unguarded, which will have the rest of the code as a no-op in a non-proofs-build, if I'm not mistaken. What do you think?

src/theory/arith/constraint.cpp
```diff
@@ -551,46 +550,48 @@ bool Constraint::hasTrichotomyProof() const {
...
+      if (antecdent == NullConstraint)
+      {
+        break;
+      }
+      antecdent->printProofTree(out, depth + 1);
     }
-    antecdent->printProofTree(out, depth + 1);
   }
-#else  /* IS_PROOFS_BUILD */
   out << "Cannot print proof. This is not a proof build." << endl;
```
alex-ozdemir, line 594:
```suggestion
```

Should this be removed?

HanielB: So that we keep things synced with proof-new I've added a return statement in the end of the `if` block and kept the prenting message.

PR #4966, opened 2020-08-28
Simplify interface for computing relevant terms.

ajreynol:
This is a followup to https://github.com/CVC4/CVC4/pull/4945 which simplifies the contract for computeRelevantTerms.


src/theory/theory.h
```diff
@@ -688,13 +672,29 @@ class Theory {
...
-   * Same as above, but with empty irrKinds. This version can be overridden
-   * by the theory, e.g. by restricting or extending the set of terms returned
-   * by computeRelevantTermsInternal, which is called by default with no
-   * irrKinds.
+   * Scans the current set of assertions and shared terms top-down
+   * until a theory-leaf is reached, and adds all terms found to
+   * termSet.  This is used by collectModelInfo to delimit the set of
+   * terms that should be used when constructing a model.
+   *
+   * irrKinds: The kinds of terms that appear in assertions that should *not*
```
4tXJ7f, line 680:
```suggestion
   * @param irrKinds The kinds of terms that appear in assertions that should *not*
```

src/theory/theory.h
```diff
@@ -688,13 +672,29 @@ class Theory {
...
+   * be included in termSet. Note that the kinds EQUAL and NOT are always
+   * treated as irrelevant kinds.
+   *
+   * includeShared: Whether to include shared terms in termSet. Notice that
```
4tXJ7f, line 684:
```suggestion
   * @param includeShared Whether to include shared terms in termSet. Notice that
```

PR #4967, opened 2020-08-28
Incremental support for bv_to_int

yoni206:
This PR adds support for incremental solving in bv_to_int.
This amounts to:
1. using context dependent data structures
2. adding a test

In addition, we check for parametrized operations in a more robust way (using `kind::metakind::PARAMETERIZED`) and rename some tests for convenience.


src/preprocessing/passes/bv_to_int.cpp
```diff
@@ -643,7 +643,7 @@ Node BVToInt::bvToInt(Node n)
               Assert(d_bvToIntCache.find(a) != d_bvToIntCache.end());
               Assert(i >= j);
               Node div = d_nm->mkNode(
-                  kind::INTS_DIVISION_TOTAL, d_bvToIntCache[a], pow2(j));
+                  kind::INTS_DIVISION_TOTAL, d_bvToIntCache[a].get(), pow2(j));
```
ajreynol, line 646: You don't need these .get() calls (applies to many of the changes in the PR).

PR #4968, opened 2020-08-28
Add the inference manager for datatypes

ajreynol:
This is in preparation for converting datatypes to the new standard. It adds a specialized version of inference manager buffered that datatypes uses.  This required adding several utility methods to its base classes.

A follow up PR will connect this to TheoryDatatypes.


src/theory/datatypes/inference_manager.h
```diff
@@ -0,0 +1,108 @@
...
+
+/**
+ * The datatypes inference manager. The main unique features of this inference
+ * manager are:
+ * (1) Explicit caching of lemmas,
+ * (2) A custom process() method with relies on a policy determining which
+ * facts must be sent as lemmas (mustCommunicateFact).
+ * (3) Methods for tracking when lemmas and facts have been processed.
+ */
+class InferenceManager : public InferenceManagerBuffered
```
4tXJ7f, line 36: I find it a bit confusing the also call this `InferenceManager` in the new architecture, can we maybe call it `DatatypesInferenceManager`?

src/theory/datatypes/inference_manager.h
```diff
@@ -0,0 +1,108 @@
...
+   * to reset? 
+   */
+  bool hasAddedFact() const;
+  /** 
+   * Have we sent a lemma (via process) since the most recent call to reset? 
+   */
+  bool hasAddedLemma() const;
+  /**
+   * If not cached, send lemma on lem the output channel and cache. Returns
+   * true if a lemma was sent.
```
4tXJ7f, line 65:
```suggestion
   * If not cached, send lemma lem on the output channel and cache. Returns
   * true if a lemma was sent.
```

src/theory/datatypes/inference_manager.h
```diff
@@ -0,0 +1,108 @@
...
+   */
+  bool doSendLemma(Node lem, LemmaProperty p = LemmaProperty::NONE, bool cached=true);
+  /** Multi-version of the above, returns true if any lemma was sent. */
```
4tXJ7f, line 68:
```suggestion
  /** Multi-lemma version of the above, returns true if any lemma was sent. */
```

src/theory/datatypes/inference_manager.cpp
```diff
@@ -0,0 +1,190 @@
...
+  d_true = NodeManager::currentNM()->mkConst(true);
+}
+
+void InferenceManager::reset()
+{
+  d_addedLemma = false;
+  d_addedFact = false;
+}
+
+bool InferenceManager::mustCommunicateFact(Node n, Node exp) const
```
4tXJ7f, line 44: I am wondering if this should be a general method in the `InferenceManager` class that the individual theories override?

src/theory/datatypes/inference_manager.cpp
```diff
@@ -0,0 +1,190 @@
...
+  return d_addedFact;
+}
+bool InferenceManager::hasAddedLemma() const
+{
+  return d_addedLemma;
+}
+bool InferenceManager::doSendLemma(Node lem, LemmaProperty p, bool cached)
+{
+  // don't cache lemmas with non-standard properties
+  Assert (!cached || p==LemmaProperty::NONE);
```
4tXJ7f, line 151: Apply ClangFormat.

src/theory/datatypes/inference_manager.cpp
```diff
@@ -0,0 +1,190 @@
...
+  bool doSend = false;
+  if (!cached)
+  {
+    // always send
+    doSend = true;
+  }
+  else if (d_lemmasSent.find(lem) == d_lemmasSent.end())
```
4tXJ7f, line 158: I am wondering if we should move the caching this to the buffered inference manager (together with a boolean that enables/disables caching of lemmas).

src/theory/datatypes/inference_manager.h
ajreynol: @nafur and I discussed this exact issue, we were leaning towards plain "InferenceManager".  Notice that the generic base class is intentionally called "TheoryInferenceManager".

src/theory/datatypes/inference_manager.cpp
ajreynol: It's possible as a next step.  The even more pressing candidate is actually the d_hasAddedFact / d_hasAddLemma / reset functionality.  That is used in many theories.

src/theory/datatypes/inference_manager.cpp
ajreynol: I had considered this, however, that caching varies per theory.

The issue is that you must also be careful about LemmaProperty.  This leads to two cases:
(1) Theories don't use LemmaProperty and thus can cache uniformly `Node` only (for performance),
(2) Theories use LemmaProperty and must cache based on `std::pair<Node, LemmaProperty>`. 

Moreover, theories may know more efficient ways of caching e.g. non-linear arithmetic maintains two lemma caches for LemmaProperty={none, preprocess}. 

In short, I think caching isn't clear enough of a common feature since each theory has its own justifiably preferred way of doing it (in my estimation).  We could consider standardizing this later as the code develops.

nafur: I support @4tXJ7f. Implementing this cache in a class derived from buffered inference manager is incredibly difficult and error-prone as it really needs to know what exactly its base class is doing. The cache should contain the lemmas that were sent to the output channel, and this is virtually impossible to get right if there is an abstraction layer in between that caches lemmas and may even drop lemmas.

As for the different caches for different lemma properties: I'm not to fond of this solution anyway, as it paves the way to adding the same lemma multiple times.

src/theory/datatypes/inference_manager.h
nafur: The main argument was that one should never have two `InferenceManager` classes in scope anyway, as this would mix different theories.

src/theory/datatypes/inference_manager.h
```diff
@@ -0,0 +1,111 @@
...
+
+/**
+ * The datatypes inference manager. The main unique features of this inference
+ * manager are:
+ * (1) Explicit caching of lemmas,
+ * (2) A custom process() method with relies on a policy determining which
+ * facts must be sent as lemmas (mustCommunicateFact).
+ * (3) Methods for tracking when lemmas and facts have been processed.
+ */
+class InferenceManager : public InferenceManagerBuffered
```
nafur, line 36: I really don't like how this class uses `InferenceManagerBuffered`. The pure existence of `doSendLemma()` suggests to me that one is not supposed to call `addPendingLemma()`, though this is technically allowed (by `public` inheritance) and also expected (`process()` expects some lemmas to be pending as it calls `doPendingLemmas()`). This means that certain lemmas are cached, and some are not, without documentation, though.

src/theory/datatypes/inference_manager.h
```diff
@@ -0,0 +1,111 @@
...
+  bool mustCommunicateFact(Node n, Node exp) const;
+  /** A cache of all lemmas sent */
+  NodeSet d_lemmasSent;
+  /** Common nodes */
+  Node d_true;
+  /**
+   * This flag is set to true during a full effort check if this theory
+   * called d_out->lemma(...).
+   */
+  bool d_addedLemma;
```
nafur, line 99: IMHO, this flag does not belong here, but rather in `TheoryInferenceManager`. This class has no reasonable way to make sure that this flag does what it documentation claims (as long as we publicly inherit from `InferenceManagerBuffered`).

src/theory/datatypes/inference_manager.h
```diff
@@ -0,0 +1,111 @@
...
+  bool hasAddedLemma() const;
+  /**
+   * If not cached, send lemma lem on the output channel and cache. Returns
+   * true if a lemma was sent.
+   */
+  bool doSendLemma(Node lem,
+                   LemmaProperty p = LemmaProperty::NONE,
+                   bool cached = true);
+  /** Multi-lemma version of the above, returns true if any lemma was sent. */
+  bool doSendLemmas(const std::vector<Node>& lem);
```
nafur, line 71: Are these supposed to be `public`? Their doc should at least say something about their relation to `addPendingLemma()` and `lemma()`, when one should call which, etc.

src/theory/datatypes/inference_manager.cpp
```diff
@@ -0,0 +1,184 @@
...
+    return true;
+  }
+  Trace("dt-lemma-debug") << "Do not need to communicate " << n << std::endl;
+  return false;
+}
+
+void InferenceManager::process()
+{
+  // process pending lemmas, used infrequently, only for definitional lemmas
+  doPendingLemmas();
```
nafur, line 81: If `numPendingLemmas() > 0`, we should set `d_addedLemmas = true` here.

src/theory/datatypes/inference_manager.cpp
```diff
@@ -0,0 +1,184 @@
...
+  // now process the pending facts
+  size_t i = 0;
+  NodeManager* nm = NodeManager::currentNM();
+  while (!d_theoryState.isInConflict() && i < d_pendingFact.size())
+  {
+    std::pair<Node, Node>& pfact = d_pendingFact[i];
+    Node fact = pfact.first;
+    Node exp = pfact.second;
```
nafur, line 89:
```suggestion
    const std::pair<Node, Node>& pfact = d_pendingFact[i];
    const Node& fact = pfact.first;
    const Node& exp = pfact.second;
```

src/theory/datatypes/inference_manager.cpp
```diff
@@ -0,0 +1,184 @@
...
+        Trace("dt-lemma-debug") << "Trivial explanation." << std::endl;
+      }
+      else
+      {
+        Trace("dt-lemma-debug") << "Get explanation..." << std::endl;
+        std::vector<TNode> assumptions;
+        explain(exp, assumptions);
+        if (assumptions.empty())
+        {
+          lem = fact;
```
nafur, line 108: We have `lem == fact` at this point anyway.
```suggestion
```

nafur (changes requested):
As already indicated, I have some concerns on which class in the hierarchy should be responsible for certain, things, in particular the caching...

src/theory/datatypes/inference_manager.cpp
ajreynol: Ok, given that it is critical to non-linear arithmetic at the moment, I'm on board with standardizing this now. Perhaps we should meet this week to discuss.

@nafur Regarding adding the same lemma with multiple properties several times, this is actually a question of model soundness, since it may be incorrect to drop a lemma with a different property (as a simple example say a lemma is sent with NONE and then afterwards with NEEDS_JUSTIFY). Its required in some cases to ensure this is done correctly.

src/theory/datatypes/inference_manager.h
ajreynol: Discussed offline, I am standardizing both the caching of lemmas and the tracking of whether lemmas/facts have been sent, thus eliminating the need for doSendLemma.

src/theory/theory_inference_manager.h
```diff
@@ -114,11 +114,31 @@ class TheoryInferenceManager
...
-  LemmaStatus lemma(TNode lem, LemmaProperty p = LemmaProperty::NONE);
-  /** Send (trusted) lemma lem with property p on the output channel. */
-  LemmaStatus trustedLemma(const TrustNode& tlem,
-                           LemmaProperty p = LemmaProperty::NONE);
+  /**
+   * Send (trusted) lemma lem with property p on the output channel.
+   *
+   * @param tlem The trust node containing the lemma and its proof generator.
+   * @param p The property of the lemma
+   * @param doCache If true, we
```
nafur, line 134: Add a proper comment here... the doc of `doCache` end prematurely...

src/theory/datatypes/inference_manager.cpp
ajreynol: Or should this be TNode?  I forget?

src/theory/datatypes/inference_manager.cpp
ajreynol: This is now standardized at a higher level.

src/theory/datatypes/inference_manager.h
nafur: I like the new solution way better.

nafur (changes requested):
Only a few minor issues now.

src/theory/datatypes/inference_manager.cpp
nafur: I think you use `fact` and `exp` as `Node` later on when calling `mustCommunicateFact()` (which should take them by `const&` as well, in my opinion...

nafur (approved):
LGTM now.

PR #4970, opened 2020-08-28
Changing the handled operators in bv2int preprocessing pass

yoni206:
Some of the bit-vector operators are directly translated to integers, while others are eliminated before the translation.
This PR changes the set of operators that we eliminate (and as a consequence, also the set of operators that we handle directly):
1. The only bit-wise operator that is translated is `bvand`. The rest are now eliminated.
2. `bvneg` is now eliminated.
3. The various `division` operators are still eliminated, but using different rewrite rules.
4. zero-extend and sign-extend are now handled directly.
5. shifting is changed to favor ITEs over non-linear multiplication.


src/preprocessing/passes/bv_to_int.cpp
```diff
@@ -190,24 +180,27 @@ Node BVToInt::eliminationPass(Node n)
...
                                   RewriteRule<SltEliminate>,
                                   RewriteRule<SgtEliminate>,
-                                  RewriteRule<SgeEliminate>,
-                                  RewriteRule<ShlByConst>,
-                                  RewriteRule<LshrByConst> >::apply(current);
+                                  RewriteRule<SgeEliminate>>::apply(current);
       // save in the cache
       d_eliminationCache[current] = currentEliminated;
-      // also assign the eliminated now to itself to avoid revisiting.
+      // also assign the eliminated now to itself to avoid revisiting.
```
ajreynol, line 203: Is there a format issue here with new space?

test/regress/regress2/bv_to_int_shifts.smt2
```diff
@@ -1,17 +1,18 @@
 ; COMMAND-LINE: --solve-bv-as-int=sum --bvand-integer-granularity=1 --no-check-models  --no-check-unsat-cores
 ; EXPECT: sat
 (set-logic QF_BV)
-(declare-fun s () (_ BitVec 64))
-(declare-fun t () (_ BitVec 64))
-(declare-fun splust () (_ BitVec 64))
-(declare-fun shift1 () (_ BitVec 64))
-(declare-fun shift2 () (_ BitVec 64))
-(declare-fun negshift1 () (_ BitVec 64))
+(declare-fun s () (_ BitVec 4))
```
ajreynol, line 4: Why this change?

ajreynol (changes requested):
Mostly looks good, have a few minor questions.

src/preprocessing/passes/bv_to_int.cpp
yoni206: fixed [fc48ecc55625bb34937d7ec1eab62f8732857d7b](https://github.com/CVC4/CVC4/pull/4970/commits/fc48ecc55625bb34937d7ec1eab62f8732857d7b)

test/regress/regress2/bv_to_int_shifts.smt2
yoni206: just to make the test run faster. 4 bits is interesting enough for these operators in terms of correctness / debug assertions.

ajreynol (approved):

PR #4975, opened 2020-08-29
Basic proof support in inference manager

ajreynol:
This adds basic support for asserting internal facts with proofs in the inference manager class.

The purpose of this PR is twofold:
(1) From the point of view of proofs, this PR standardizes the management of proof equality engine within inference manager. Theories no longer have to manually construct proof equality engines, and instead are recommended to create inference managers.
(2) From the point of view of the new approach to theory combination, this PR ensures standard theory callbacks (preNotifyFact / notifyFact) are used for internal facts, regardless of whether proofs are enabled.

This will simplify several of the current (unmerged) changes for proof production in theory solvers on `proof-new`.

Notice this PR adds the utility method NodeManager::mkAnd, which is arguably long overdue.

Also notice this code is not yet active, but will be used on proof-new after this PR is merged.

### DIFF
--- a/src/expr/node_manager.h
+++ b/src/expr/node_manager.h
@@ -1381,7 +1381,7 @@ Node NodeManager::mkAnd(const std::vector<NodeTemplate<ref_count> >& children)
   {
     return children[0];
   }
-  return mkNode(AND, a);
+  return mkNode(kind::AND, children);
 }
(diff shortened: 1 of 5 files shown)


nafur (approved):
LGTM

PR #4979, opened 2020-08-31
Fix LIBRARY_INSTALL_DIR

FabianWolff:
On Debian (for instance), libraries aren't installed into `/usr/lib/`, but into something like `/usr/lib/x86_64-linux-gnu/`. In particular, this means that setting the `LIBRARY_INSTALL_DIR` to `lib` in the top-level `CMakeLists.txt` file is wrong.

Luckily, there is a simple solution: CMake provides [`CMAKE_INSTALL_LIBDIR`](https://cmake.org/cmake/help/v3.0/module/GNUInstallDirs.html) for this very purpose, which has sensible defaults and can be set by the user. In particular, since `CMAKE_INSTALL_LIBDIR` is a standardized variable, tools like the ones used for building Debian packages can set it to what they want it to be, whereas using a custom variable like `LIBRARY_INSTALL_DIR` wouldn't work in this setting.

You might also want to consider using `CMAKE_INSTALL_{BINDIR,INCLUDEDIR}` instead of setting your own defaults; I think this would be a "cleaner" solution, but it's not as important as the library installation directory.


nafur:
Unless there is a reason not to: why not replace `LIBRARY_INSTALL_DIR` by `CMAKE_INSTALL_LIBDIR` everywhere? (and the same for the other variables).
We should check how this behaves on windows, though (see the code at https://gitlab.kitware.com/cmake/cmake/-/issues/20250 for example), and I think one needs to include `GNUInstallDirs` before using these variables.

FabianWolff:
Sounds good to me. And you're right, those variables need an `include(GNUInstallDirs)`. I replaced the variables now, and it seems to work; I can't test it on Windows, though.

4tXJ7f (approved):
This looks great, thank you. I've tested our cross-compilation of Windows binaries with this PR and it seemed to work fine. We don't support building on Windows directly, so we don't have to worry about that.

PR #4980, opened 2020-08-31
Remove build path from generated files to improve reproducibility

FabianWolff:
The [Reproducible Builds](https://reproducible-builds.org/) project aims to make software builds _reproducible_ (and thus _verifiable_, i.e. anybody can check whether a given binary actually was generated from the given sources).

CVC4 cannot yet be built reproducibly. One issue are [build paths](https://reproducible-builds.org/docs/build-path/): The generated header files include absolute paths, so if CVC4 is rebuilt in a different directory, the result will be different.

This pull request fixes this specific issue by making the paths in the generated files relative. This doesn't make the CVC4 build reproducible yet, but it's one step. One remaining issue, for instance, is this function in `src/base/configuration.cpp`:
```
std::string Configuration::getCompiledDateTime() {
  return __DATE__ " " __TIME__;
}
```
Obviously, including the current date and time also makes the build non-reproducible, so perhaps you could get rid of this function at some point (for the Debian package, I'm currently patching it to use the last entry of the package changelog as the date and time).


nafur:
Unfortunately, this change obstructs the error reporting in the generated files. For gcc, we used to get something like
```
/home/gereon/cvc4_master/src/expr/expr_manager_template.cpp: In member function ‘CVC4::Expr CVC4::ExprManager::mkExpr(CVC4::Kind, CVC4::Expr)’:
/home/gereon/cvc4_master/src/expr/expr_manager_template.cpp:39:7: error: ‘stringstreamasd’ was not declared in this scope
   39 |       stringstream statName; \
      |       ^~~~~~~~~~~~~~~
/home/gereon/cvc4_master/src/expr/expr_manager_template.cpp:39:7: note: in definition of macro ‘INC_STAT’
   39 |       stringstream statName; \
      |       ^~~~~~~~~~~~~~~
/home/gereon/cvc4_master/src/expr/expr_manager_template.cpp:40:7: error: ‘statName’ was not declared in this scope
   40 |       statName << "expr::ExprManager::" << kind; \
      |       ^~~~~~~~
/home/gereon/cvc4_master/src/expr/expr_manager_template.cpp:40:7: note: in definition of macro ‘INC_STAT’
   40 |       statName << "expr::ExprManager::" << kind; \
      |       ^~~~~~~~
```
which now becomes
```<CVC4>/src/expr/expr_manager_template.cpp: In member function ‘CVC4::Expr CVC4::ExprManager::mkExpr(CVC4::Kind, CVC4::Expr)’:
<CVC4>/src/expr/expr_manager_template.cpp:39:7: error: ‘stringstreamasd’ was not declared in this scope
<CVC4>/src/expr/expr_manager_template.cpp:39:7: note: in definition of macro ‘INC_STAT’
<CVC4>/src/expr/expr_manager_template.cpp:40:7: error: ‘statName’ was not declared in this scope
<CVC4>/src/expr/expr_manager_template.cpp:40:7: note: in definition of macro ‘INC_STAT’
```
A possibility would be to make the path relative to the generated file. Using `../../src/expr/expr_manager_template.cpp` yields:
```
../../src/expr/expr_manager_template.cpp: In member function ‘CVC4::Expr CVC4::ExprManager::mkExpr(CVC4::Kind, CVC4::Expr)’:
../../src/expr/expr_manager_template.cpp:39:7: error: ‘stringstreamasd’ was not declared in this scope
   39 |       stringstream statName; \
      |       ^~~~~~~~~~~~~~~
../../src/expr/expr_manager_template.cpp:39:7: note: in definition of macro ‘INC_STAT’
   39 |       stringstream statName; \
      |       ^~~~~~~~~~~~~~~
../../src/expr/expr_manager_template.cpp:40:7: error: ‘statName’ was not declared in this scope
   40 |       statName << "expr::ExprManager::" << kind; \
      |       ^~~~~~~~
../../src/expr/expr_manager_template.cpp:40:7: note: in definition of macro ‘INC_STAT’
   40 |       statName << "expr::ExprManager::" << kind; \
      |       ^~~~~~~~
```
Now, I'd consider the filename slightly misleading, but showing the code works...

FabianWolff:
> Unfortunately, this change obstructs the error reporting in the generated files.

I see; I hadn't thought of the fact that the compiler does actually use the paths for loading the snippets. As you say, your solution using actual, working relative paths isn't really satisfactory either; but I still believe reproducibility to be a worthy goal. Perhaps you could post-process the installed header files with `sed` or something to replace the build directory with something else?

In fact, if you think about it, once the header files are installed, the build path can't be assumed to exist anymore, and so reporting an error in a file under the build path is actually wrong. So maybe you should drop the `#line` directives altogether in the installed headers (i.e. in a post-processing step, similar to `fix-install-headers.sh`)?

PR #4982, opened 2020-08-31
Add the shared solver

ajreynol:
This PR adds the definition for the "shared solver", it does not connect it to TheoryEngine/CombinationEngine yet. This is an object that behaves like a "glue" theory solver and has a special role in TheoryEngine.

In the current architecture, the "SharedTermsDatabase" is the "shared solver", although in the central architecture, the shared solver will be required to behave differently.  This PR adds the abstract definition of shared solver, where notice SharedSolverDistributed is a thin layer on top of SharedTermsDatabase.

In a followup PR, CombinationEngine will maintain a shared solver and ~5 blocks of code in TheoryEngine will be callbacks to the SharedSolver.  Also, eventually the code for SharedTermsDatabase should be split: the parts involving equality reason/propagation/explanation should be migrated into SharedSolverDistributed, and the parts related to registration will remain in SharedTermsDatabase (to also be used in the planned SharedSolverCentral).

FYI @barrettcw


src/theory/shared_solver.h
```diff
@@ -0,0 +1,135 @@
...
+   * Get the equality status of a and b, which first asks if the shared
+   * terms database as an equality status, and otherwise asks the appropriate
+   * Theory.
+   *
+   * This method is used by theories via Valuation mostly for determining their
+   * care graph.
+   */
+  virtual EqualityStatus getEqualityStatus(TNode a, TNode b);
+  /**
+   * Explain literal, which returns a conjunction of literals that that entail
```
4tXJ7f, line 88:
```suggestion
   * Explain literal, which returns a conjunction of literals that entail
```

src/theory/shared_solver.cpp
```diff
@@ -0,0 +1,116 @@
...
+}
+
+bool SharedSolver::needsEqualityEngine(theory::EeSetupInfo& esi)
+{
+  return false;
+}
+
+void SharedSolver::preRegisterShared(TNode t, bool multipleTheories)
+{
+  // register it with the equality engine manager if shared is enabled
```
4tXJ7f, line 44:
```suggestion
  // register it with the equality engine manager if sharing is enabled
```

src/theory/shared_solver.cpp
```diff
@@ -0,0 +1,116 @@
...
+  if (d_logicInfo.isSharingEnabled())
+  {
+    preRegisterSharedInternal(t);
+  }
+  // if multiple theories are present in t
+  if (multipleTheories)
+  {
+    // Collect the shared terms if there are multiple theories
+    // This calls Theory::addSharedTerm, possibly multiple times
+    NodeVisitor<SharedTermsVisitor>::run(d_sharedTermsVisitor, t);
```
4tXJ7f, line 54: Eventually, we should consider moving the `SharedTermsVisitor` to this source file since it is (presumably) not going to be used anywhere else.

src/theory/shared_solver.cpp
```diff
@@ -0,0 +1,116 @@
...
+  }
+}
+
+void SharedSolver::preNotifySharedFact(TNode atom)
+{
+  if (d_sharedTerms.hasSharedTerms(atom))
+  {
+    // Always notify the theories the shared terms, which is independent of
```
4tXJ7f, line 62:
```suggestion
    // Always notify the theories of the shared terms, which is independent of
```

src/theory/shared_solver.h
```diff
@@ -0,0 +1,135 @@
...
+   * fact is asserted in TheoryEngine, just before it is dispatched to the
+   * appropriate theory.
+   *
+   * This calls Theory::notifySharedTerm for the shared terms of the atom.
+   */
+  void preNotifySharedFact(TNode atom);
+  /**
+   * Get the equality status of a and b, which first asks if the shared
+   * terms database as an equality status, and otherwise asks the appropriate
+   * Theory.
```
4tXJ7f, line 81: The base class doesn't actually do that, only the `SharedSolverDistributed`. Maybe just say "Get the equality status of a and b"?

4tXJ7f (approved):
LGTM, just some minor typos.

src/theory/shared_solver.cpp
ajreynol: Right, this also the case for parts of SharedTermsDatabase.

src/theory/shared_solver.h
ajreynol: Ah, right, I'll update.

PR #4984, opened 2020-08-31
(proof-new) Improve debugging infrastructure for open proofs

ajreynol:
This includes more versions of checking whether a proof node is closed and standardizing output.


src/expr/proof_generator.cpp
```diff
@@ -150,11 +175,53 @@ void pfgEnsureClosedWrt(Node proven,
...
+{
+  Assert(!proven.isNull());
+  // proof generator may be null
+  ensureClosedWrtInternal(proven, pg, nullptr, assumps, c, ctx, reqGen);
+}
+
+void pfnEnsureClosed(ProofNode* pn, const char* c, const char* ctx)
+{
+  std::vector<Node> assumps;
+  ensureClosedWrtInternal(Node::null(), nullptr, pn, assumps, c, ctx, false);
```
HanielB, line 216:
```suggestion
  ensureClosedWrtInternal(Node::null(), nullptr, pn, {}, c, ctx, false);
```

HanielB (approved):
LGTM, just a minor suggestion

PR #4986, opened 2020-09-01
 (new theory) Update TheoryDatatypes to the new standard

ajreynol:
Updates it to use the new inference manager as well as updating its check to the standard callbacks.


nafur (approved):
LGTM.

PR #4989, opened 2020-09-01
Change the unavailable mercury repository for ABC to the GitHub code base

mmuesly:
The get-abc script failed as the `https://bitbucket.org/alanmi/abc` repository is no longer available.

It seems like the code moved to GitHub: https://github.com/berkeley-abc/abc

After a quick search in the git history, it seems like the commit https://github.com/berkeley-abc/abc/commit/3d7034bf619aada30b0ba1e8afcb151d304ab556 is the counterpart to the previously used changeset `53f39c11b58d`.  I adapt the install script to retrieve the code from GitHub instead of bitbucket.

In addition, the `cp --parents` command is not available on OSX by default. Therefore, I recommend to replace this with a `rsync` command on OSX.


contrib/get-abc
```diff
@@ -6,8 +6,9 @@ ABC_DIR="$DEPS_DIR/abc"
 commit=53f39c11b58d
```
nafur, line 6:
```suggestion
commit=3d7034bf619aada30b0ba1e8afcb151d304ab556
```

contrib/get-abc
```diff
@@ -6,8 +6,9 @@ ABC_DIR="$DEPS_DIR/abc"
...
 
 check_dep_dir "$ABC_DIR"
-setup_dep "https://bitbucket.org/alanmi/abc/get/$commit.tar.gz" "$ABC_DIR"
+git clone https://github.com/berkeley-abc/abc.git "$ABC_DIR"
 cd "$ABC_DIR"
+git checkout 3d7034bf
```
nafur, line 11:
```suggestion
setup_dep "https://github.com/berkeley-abc/abc/archive/$commit.tar.gz" "$ABC_DIR"
cd "$ABC_DIR"
```

nafur (changes requested):
Thanks for the input!
I think we prefer not using git but download tarballs.

mmuesly:
@nafur thank you for reviewing my changes. I adopted it according to your suggestion.

nafur (approved):
This should be good. Could you please sign off your commit? (see https://github.com/CVC4/CVC4/wiki/Developer-Guide#pull-requests, in particular the "External contributor" part...)
Thanks!

mmuesly:
I added the signed-off line to the commit message.

PR #4990, opened 2020-09-01
Add TheoryInference base class

ajreynol:
This introduces a TheoryInference base class, which is a generalization and cleaner version of the former `Lemma` object.  This changes the name of Lemma -> SimpleTheoryLemma, and makes the callback responsible for calling the inference manager.

This PR also updates the datatypes inference manager to use the new style.  This required adding some additional interfaces to TheoryInferenceManager.


src/theory/theory_inference.h
```diff
@@ -0,0 +1,102 @@
...
+class SimpleTheoryLemma : public TheoryInference
+{
+ public:
+  SimpleTheoryLemma(Node n, LemmaProperty p, ProofGenerator* pg);
+  virtual ~SimpleTheoryLemma() {}
+  /**
+   * Send the lemma using inference manager im, return true if the lemma was
+   * sent.
+   */
+  bool process(TheoryInferenceManager* im) override;
```
nafur, line 63: If we are to derive from this class, this should be `virtual`.
```suggestion
  virtual bool process(TheoryInferenceManager* im) override;
```

src/theory/theory_inference.h
```diff
@@ -0,0 +1,102 @@
...
+class SimpleTheoryInternalFact : public TheoryInference
+{
+ public:
+  SimpleTheoryInternalFact(Node conc, Node exp, ProofGenerator* pg);
+  virtual ~SimpleTheoryInternalFact() {}
+  /**
+   * Send the lemma using inference manager im, return true if the lemma was
+   * sent.
+   */
+  bool process(TheoryInferenceManager* im) override;
```
nafur, line 90: ditto.
```suggestion
  virtual bool process(TheoryInferenceManager* im) override;
```

nafur (commented):
A quick comment.

src/theory/theory_inference.h
```diff
@@ -0,0 +1,102 @@
...
+ **
+ ** \brief The theory inference utility
+ **/
+
+#include "cvc4_private.h"
+
+#ifndef CVC4__THEORY__THEORY_INFERENCE_H
+#define CVC4__THEORY__THEORY_INFERENCE_H
+
+#include "context/cdhashmap.h"
```
nafur, line 20: I don't think we need that include...
```suggestion
```

src/theory/theory_inference.h
```diff
@@ -0,0 +1,102 @@
...
+ * call to TheoryInferenceManager for lemmas or facts.
+ */
+class TheoryInference
+{
+ public:
+  virtual ~TheoryInference() {}
+  /**
+   * Called by the provided inference manager to process this inference. This
+   * method should make a single call to inference manager to process this
+   * inference (e.g. a call to trustedLemma, assertFactInternal etc.).
```
nafur, line 42: I have two issues with this comment:
1. Why only a single call? We explicitly want to do multiple calls here: one to push this lemma to the output channel and one to do side effects, for example.
2. While I agree with the overall decision, I expect the fact that the lemma needs to push itself to the output channel in a callback called by the inference manager to be somewhat counter intuitive to the average user. I'd thus try to stress that this callback needs to do this.

src/theory/theory_inference.h
```diff
@@ -0,0 +1,102 @@
...
+   *
+   * @return true if the inference was successfully processed by the inference
+   * manager.
```
nafur, line 45: When would we return `false`? Is this an error condition? Can it mean that it was already cached?

src/theory/theory_inference_manager.h
```diff
@@ -145,6 +155,48 @@ class TheoryInferenceManager
...
+   * @param exp The set of (all) literals that imply conc
+   * @param noExplain The subset of exp that should not be explained by the
+   * equality engine
+   * @param pg The proof generator who can provide a proof of conc.
+   * @return true if the lemma was sent on the output channel.
+   */
+  bool lemmaExp(Node conc,
+                const std::vector<Node>& exp,
+                const std::vector<Node>& noExplain,
+                ProofGenerator* pg = nullptr);
```
nafur, line 199: If `pg` needs to provide a proof, why can it be `nullptr`?

src/theory/theory_inference_manager.cpp
```diff
@@ -84,6 +84,27 @@ void TheoryInferenceManager::trustedConflict(TrustNode tconf)
   }
 }
 
+void TheoryInferenceManager::conflictExp(PfRule id,
+                                         const std::vector<Node>& exp,
+                                         const std::vector<Node>& args)
+{
+  if (!d_theoryState.isInConflict())
```
nafur, line 91: Do we actually want to simply drop a conflict? I'd guess we should at least communicate this fact (by a `bool` return value)...

src/theory/inference_manager_buffered.h
```diff
@@ -154,9 +126,9 @@ class InferenceManagerBuffered : public TheoryInferenceManager
 
  protected:
   /** A set of pending lemmas */
-  std::vector<std::shared_ptr<Lemma>> d_pendingLem;
+  std::vector<std::shared_ptr<TheoryInference>> d_pendingLem;
   /** A set of pending facts, paired with their explanations */
```
nafur, line 130: Update the comment.

src/theory/datatypes/inference_manager.h
```diff
@@ -25,6 +25,38 @@ namespace CVC4 {
 namespace theory {
 namespace datatypes {
 
+class DatatypesInference : public TheoryInference
```
nafur, line 28: Why not inherit from `SimpleTheoryInternalFact`? It has exactly the same members (+ the proof generator). Don't we need the proof generator for `DatatypesInference` as well?

nafur (changes requested):
Some comments...

src/theory/datatypes/inference_manager.h
ajreynol: Sure, this is a good idea (proof generators will be added later).

src/theory/theory_inference.h
ajreynol: For 1, I'll say this is the typical (but not required) use case.

src/theory/theory_inference.h
ajreynol: If its been cached, although more generally we may also return false if we process as a fact and the fact was redundant (still debating about whether we need to do this).  I'll add a comment here.

src/theory/theory_inference_manager.cpp
ajreynol: This is the same behavior as normal conflict/trustedConflict above.  I'm open to changing the return value later (maybe a separate PR?). I actually was tempted to change this to `Node` so that the theory is informed of what the conflict was (for debugging).

src/theory/theory_inference_manager.h
ajreynol: Its optional, if none is provided, we don't justify the step. I'll update the documentation.

nafur (approved):
LGTM

PR #4992, opened 2020-09-01
Add arithmetic-specific, runtime, proof-macros.

alex-ozdemir:
We'll use this to gate farkas-coefficient machinery after we remove the
old proof-macros.

I've changed the macros slightly from the `proof-new` branch: I removed the dependence on `options::proof()` (no longer wanted) and `options::unsatCores()` (I had copied this from the original proof macros, but it's not needed either, since we're in a theory).

### DIFF
--- a/src/theory/arith/proof_macros.h
+++ b/src/theory/arith/proof_macros.h
@@ -4,7 +4,7 @@
  ** Top contributors (to current version):
  **   Alex Ozdemir
  ** This file is part of the CVC4 project.
- ** Copyright (c) 2009-2019 by the authors listed in the file AUTHORS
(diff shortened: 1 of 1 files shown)


src/theory/arith/proof_macros.h
```diff
@@ -0,0 +1,34 @@
+/*********************                                                        */
+/*! \file proof_macros.h
+ ** \verbatim
+ ** Top contributors (to current version):
+ **   Alex Ozdemir
+ ** This file is part of the CVC4 project.
+ ** Copyright (c) 2009-2019 by the authors listed in the file AUTHORS
```
HanielB, line 7:
```suggestion
 ** Copyright (c) 2009-2020 by the authors listed in the file AUTHORS
```

PR #4993, opened 2020-09-01
Added a new rule to simplify (bvugt (bvurem T x) x)

### LINKED ISSUE #4936: Potential performance issue on QF_BV formula
Hi, for the following formula 
~~~~
(set-logic QF_BV)
(declare-const bv_40-0 (_ BitVec 40))
(assert (bvugt (bvurem ((_ rotate_right 6) bv_40-0) bv_40-0) bv_40-0))
(check-sat)
~~~~

CVC4 nightly https://cvc4.cs.stanford.edu/downloads/builds/x86_64-linux-opt/unstable/cvc4-2020-08-18-x86_64-linux-opt
~~~~
time ./yices_smt2 yy.smt2
unsat
real    0m0.258s
user    0m0.253s
sys     0m0.005s


./cvc4 --tlimit=30000 yy.smt2                                                                                                                        
CVC4 interrupted by timeout.
~~~~

nafur:
Motivated by #4936, this PR adds a new BV rewrite rule:
`(bvugt (bvurem T x) x)  ==>  (ite (= x 0) (bvugt T 0) false)`
Fixes #4936.


src/theory/bv/theory_bv_rewrite_rules_simplification.h
```diff
@@ -1557,6 +1557,40 @@ Node RewriteRule<ShiftZero>::apply(TNode node) {
...
+}
+
+template<> inline
+Node RewriteRule<UgtUrem>::apply(TNode node) {
+  Debug("bv-rewrite") << "RewriteRule<UgtUrem>(" << node << ")" << std::endl;
+  const Node& T = node[0][0];
+  const Node& x = node[1];
+  NodeManager* nm = NodeManager::currentNM();
+  return nm->mkNode(kind::AND,
+    nm->mkNode(kind::EQUAL, x, utils::mkConst(utils::getSize(x), 0)),
```
mpreiner, line 1587: How about you create a separate `Node zero = utils::mkConst(utils::getSize(x), 0)` since you need it twice?

src/theory/bv/theory_bv_rewrite_rules_simplification.h
```diff
@@ -1557,6 +1557,40 @@ Node RewriteRule<ShiftZero>::apply(TNode node) {
...
+ *   ==>  (and (=> (= x 0_k) (bvugt T x)) (=> (not (= x 0_k)) false))
+ *   ==>  (and (=> (= x 0_k) (bvugt T x)) (= x 0_k))
+ *   ==>  (and (bvugt T x) (= x 0_k))
+ *   ==>  (and (bvugt T 0_k) (= x 0_k))
+ */
+
+template <>
+inline bool RewriteRule<UgtUrem>::applies(TNode node)
+{
+  return (node.getKind() == kind::BITVECTOR_UGT
```
4tXJ7f, line 1574: I believe that this rule is only valid if `bv-div-zero-const` is `false` because otherwise a modulus of zero invalidates the rule:

```
(declare-const x (_ BitVec 4))                                               
(declare-const y (_ BitVec 4))                                               
(assert (not (= (bvugt (bvurem y x) x) (ite (= x #x0) (bvugt y #x0) false))))
(check-sat)                                                                  
```

returns `sat` with `--no-bv-div-zero-const`. The rule should only apply if that option is set to `true`.

4tXJ7f (changes requested):

src/theory/bv/theory_bv_rewrite_rules_simplification.h
nafur: Done.

src/theory/bv/theory_bv_rewrite_rules_simplification.h
nafur: That's right. I added this check, as well as two regressions.

mpreiner (approved):

4tXJ7f (approved):
LGTM, thanks for the update.

PR #4996, opened 2020-09-02
Make nonlinear extension (more) deterministic

nafur:
This PR tries to make the nonlinear extension more deterministic by keeping the order of input assertion (instead of taking them from a hash set). This makes the ordering somewhat more robust against varying node ids, which proved to be a problem for debugging...
Also adds a few logging messages at interesting places.


src/theory/arith/nl/nonlinear_extension.cpp
```diff
@@ -388,6 +390,19 @@ void NonlinearExtension::getAssertions(std::vector<Node>& assertions)
...
+       ++it)
+  {
+    Node lit = (*it).d_assertion;
+    auto iait = init_assertions.find(lit);
+    if (iait != init_assertions.end())
+    {
+      assertions.push_back(lit);
+      init_assertions.erase(iait);
+    }
+  }
```
ajreynol, line 405: Is the block of code for adding assertions directly from init_assertions below necessary anymore?

nafur: Unfortunately yes. The block above the new code may add new assertions to `init_assertions`...

ajreynol: I see, so below we are adding the leftovers?  Can you perhaps add a comment above the block of code below that mentions this?

PR #4997, opened 2020-09-02
Make CVC/API BV div/mod semantics match SMT-LIB

4tXJ7f:
This commit changes the semantics of the CVC language and with the
default semantics of the API for `BVUDIV`, `BVUREM`, `BVSDIV`, `BVSREM`,
and `BVSMOD` to match the semantics of SMT-LIB >=2.6. Relatedly, the
commit also adds comments to our API documentation for the different
semantics enabled by the `bv-div-zero-const` option.


src/smt/set_defaults.cpp
```diff
@@ -89,8 +89,7 @@ void setDefaults(LogicInfo& logic, bool isInternalSubsolver)
     // set this option if the input format is SMT LIB 2.6. We also set this
     // option if we are sygus, since we assume SMT LIB 2.6 semantics for sygus.
     options::bitvectorDivByZeroConst.set(
-        language::isInputLang_smt2_6(options::inputLanguage())
-        || language::isInputLangSygus(options::inputLanguage()));
+        !language::isInputLang_smt2_5(options::inputLanguage(), true));
```
aniemetz, line 92: Since 'true' is the default for this option, shouldn't this be **disabled** for smt-lib 2.5?

aniemetz (changes requested):

src/smt/set_defaults.cpp
4tXJ7f: Maybe I am missing something but the intention is to set the option to `false` when the input language is SMT-LIB 2.5 and `true` if it is any other language.

aniemetz (approved):

PR #4999, opened 2020-09-02
Basic integration of arith::InferenceManager

nafur:
This PR adds a first basic integration of the `arith::InferenceManager` into the arithmetic theory and the nonlinear extension in particular.
While the lemma collection mechanism (in the nonlinear extension) remains unchanged, the lemmas are ultimately not directly pushed to the output channel but instead added to the inference manager. Additionally, we no longer use the cache within the nonlinear extension but instead rely on the inference manager.
This PR is based on #4960.


nafur:
Benchmarks are in now, performance did not change.

src/theory/arith/inference.h
```diff
@@ -9,13 +9,13 @@
...
- ** \brief Inference enumeration for non-linear arithmetic.
+ ** \brief Inference enumeration for arithmetic.
  **/
 
 #include "cvc4_private.h"
 
-#ifndef CVC4__THEORY__ARITH__NL__INFER_INFO_H
-#define CVC4__THEORY__ARITH__NL__INFER_INFO_H
+#ifndef CVC4__THEORY__ARITH__INFERENCE_H
+#define CVC4__THEORY__ARITH__INFERENCE_H
```
ajreynol, line 18: Sorry this may be a bit of a pain, but I think we should update the IDs to have `NL_` to indicate they are coming from the non-linear solver. Also `T_` should become `NLT_`.

ajreynol (changes requested):
I think a (slightly tedious) change is needed for Inference, see comment.

src/theory/arith/inference.h
nafur: Done.

src/theory/arith/theory_arith.h
```diff
@@ -96,9 +97,18 @@ class TheoryArith : public Theory {
 
   std::pair<bool, Node> entailmentCheck(TNode lit) override;
 
+  /** Return a reference to the arith::InferenceManager. */
+  InferenceManager& getInferenceManager() {
```
ajreynol, line 101: Clang-format

ajreynol (approved):

PR #5001, opened 2020-09-02
(proof-new) Support proofs of quantifier instantiation

ajreynol:
This adds basic support for proofs of quantifier instantiation, which is the main method for sending lemmas from TheoryQuantifiers. Quantifier instantiation is also heavily used for solving extended string functions.


src/theory/quantifiers/instantiate.cpp
```diff
@@ -231,24 +235,95 @@ bool Instantiate::addInstantiation(
...
+  // Set up a proof if proofs are enabled. This proof stores a proof of
+  // the instantiation body with q as a free assumption.
+  std::shared_ptr<LazyCDProof> pfTmp;
+  if (isProofEnabled())
+  {
+    pfTmp.reset(new LazyCDProof(
+        d_pnm, nullptr, nullptr, "Instantiate::LazyCDProof::tmp"));
+  }
+
+  // construct the instan::tiation
```
HanielB, line 247: fix

src/theory/quantifiers/instantiate.cpp
```diff
@@ -231,24 +235,95 @@ bool Instantiate::addInstantiation(
...
   Trace("inst-assert") << "(assert " << body << ")" << std::endl;
 
-  Node lem = NodeManager::currentNM()->mkNode(kind::OR, q.negate(), body);
-  lem = Rewriter::rewrite(lem);
+  // construct the instantiation, and rewrite the lemma
+  Node lem = NodeManager::currentNM()->mkNode(kind::IMPLIES, q, body);
 
-  // check for lemma duplication
-  if (!d_qe->addLemma(lem, true, false))
+  // If proofs are enabled, attempt to construct the proof
```
HanielB, line 287: describe the proof being built

src/theory/quantifiers/instantiate.cpp
```diff
@@ -231,24 +235,95 @@ bool Instantiate::addInstantiation(
...
+      Node proven = tpBody.getProven();
+      // add the transformation proof, or THEORY_PREPROCESS if none provided
+      pfTmp->addLazyStep(proven,
+                         tpBody.getGenerator(),
+                         true,
+                         "Instantiate::getInstantiation:qpreprocess",
+                         false,
+                         PfRule::THEORY_PREPROCESS);
+      pfTmp->addStep(
+          body, PfRule::MACRO_SR_PRED_TRANSFORM, {orig_body, proven}, {body});
```
HanielB, line 276: Why don't you do EQ_RESOLVE?

ajreynol: Good catch, I had fixed that in another place but not here, this is indeed better.

HanielB (approved):

PR #5004, opened 2020-09-02
Minor updates to theory inference manager

ajreynol:
These updates are inspired by the current inference manager for sets.


nafur (approved):
LGTM.

PR #5006, opened 2020-09-02
Fix CryptoMiniSat build, regression

4tXJ7f:
This commit fixes builds that include CryptoMiniSat after commit
8ad308b23c705e73507a42d2425289e999d47f86 removed them. It also fixes one
of the regressions that requires unsat cores but was run when the build
was configured without them.


4tXJ7f:
I tested the build with CryptoMiniSat, CaDiCaL, and Kissat.

PR #5007, opened 2020-09-02
Update sets inference manager to inherit from InferenceManagerBuffered

ajreynol:
This updates the inference manager of sets to inherit from the standard base class InferenceManagerBuffered.

It matches that base class almost exactly, with cosmetic changes.  

Notice that sets currently has a design issue: the state object sends lemmas and hence currently requires a pointer to the theory.  This will be untangled in a future PR.

This PR makes a few very minor changes to the behavior, including not to use the "preprocess" property for subtype lemmas.


src/theory/inference_manager_buffered.cpp
```diff
@@ -29,10 +29,9 @@ InferenceManagerBuffered::InferenceManagerBuffered(Theory& t,
 {
 }
 
-bool InferenceManagerBuffered::hasProcessed() const
+bool InferenceManagerBuffered::hasPending() const
 {
-  return d_theoryState.isInConflict() || !d_pendingLem.empty()
-         || !d_pendingFact.empty();
+  return !d_pendingLem.empty() || !d_pendingFact.empty();
```
mudathirmahgoub, line 34: We can use these 2 functions:
`return hasPendingFact() || hasPendingLemma();`

src/theory/sets/inference_manager.cpp
```diff
@@ -96,26 +81,30 @@ bool InferenceManager::assertFactRec(Node fact, Node exp, int inferType)
   }
   bool polarity = fact.getKind() != NOT;
   TNode atom = polarity ? fact : fact[0];
+  if (d_state.isEntailed(atom, polarity))
```
mudathirmahgoub, line 84: How about moving this entailment checking to the top of the function. 
There is another entailment checking in line 40, which can be refactored with this one in line 84

ajreynol: The issue here is performance.  Note there are two different entailment checks here: either the entire formula `fact` is ready to be sent out as a formula, in which case we run an entailment check (once) on the full formula.  

Otherwise, we are traversing to each position in the formula (e.g. based on AND), and checking entailment when we get to a literal.  If we were to check at the beginning of the method, we would potentially be traversing on the order of `n^2` terms (checking entailment for each subformula) where `n` is the size of fact instead of just `n` (checking each leaf).

src/theory/sets/theory_sets_private.cpp
```diff
@@ -922,12 +894,6 @@ void TheorySetsPrivate::postCheck(Theory::Effort level)
   Trace("sets-check") << "Sets finish Check effort " << level << std::endl;
 }
 
-bool TheorySetsPrivate::preNotifyFact(TNode atom, bool polarity, TNode fact)
```
mudathirmahgoub, line 925: `preNotifyFact` is still declared in `theory_sets_private.h`. It should be removed if no longer needed.

ajreynol: Good catch, will remove.

mudathirmahgoub (approved):
The inference managers for sets is a lot cleaner now.

PR #5008, opened 2020-09-02
(proof-new) Make proofs in term formula removal term context sensitive

ajreynol:
Previously term formula removal proofs didnt not take the term context into account. This updates it to make use of the term context sensitive support in TConvProofGenerator.  More generally, it uses the term context computation as the standard way of caching the results of rewrites in this class (regardless of proofs).


HanielB (approved):
This looks great!

PR #5009, opened 2020-09-02
Split lazy bit-vector solver from TheoryBV 

mpreiner:
This commit separates the lazy bit-vector solver from TheoryBV, which is now a thin wrapper around a bit-vector solver `d_internal `. This will allow us to easily swap out the bit-vector solver in the future.

### DIFF
--- a/src/theory/bv/bv_solver.h
+++ b/src/theory/bv/bv_solver.h
@@ -19,7 +19,7 @@
 #ifndef CVC4__THEORY__BV__BV_SOLVER_H
 #define CVC4__THEORY__BV__BV_SOLVER_H
 
-#include "theory/bv/theory_bv.h"
+#include "theory/theory.h"
 
 namespace CVC4 {
 namespace theory {
@@ -28,7 +28,8 @@ namespace bv {
 class BVSolver
 {
  public:
-  BVSolver(TheoryBV& bv) : d_state(bv.d_state), d_inferManager(bv.d_infer){};
+  BVSolver(TheoryState& state, TheoryInferenceManager& inferMgr)
+      : d_state(state), d_inferManager(inferMgr){};
 
   virtual ~BVSolver(){};
 
--- a/src/theory/bv/bv_solver.h
+++ b/src/theory/bv/bv_solver.h
@@ -20,7 +20,6 @@
 #define CVC4__THEORY__BV__BV_SOLVER_H
 
 #include "theory/bv/theory_bv.h"
-#include "theory/theory.h"
 
 namespace CVC4 {
 namespace theory {
--- a/src/theory/bv/bv_solver.h
+++ b/src/theory/bv/bv_solver.h
@@ -1,5 +1,5 @@
 /*********************                                                        */
-/*! \file theory_bv_solver.h
+/*! \file bv_solver.h
  ** \verbatim
  ** Top contributors (to current version):
  **   Mathias Preiner
@@ -16,8 +16,8 @@
 
 #include "cvc4_private.h"
 
-#ifndef CVC4__THEORY__BV__THEORY_BV_SOLVER_H
-#define CVC4__THEORY__BV__THEORY_BV_SOLVER_H
+#ifndef CVC4__THEORY__BV__BV_SOLVER_H
+#define CVC4__THEORY__BV__BV_SOLVER_H
 
 #include "theory/bv/theory_bv.h"
 #include "theory/theory.h"
@@ -26,13 +26,12 @@ namespace CVC4 {
 namespace theory {
 namespace bv {
 
-class TheoryBVSolver
+class BVSolver
 {
  public:
-  TheoryBVSolver(TheoryBV& bv)
-      : d_bv(bv), d_state(bv.d_state), d_inferManager(bv.d_infer){};
+  BVSolver(TheoryBV& bv) : d_state(bv.d_state), d_inferManager(bv.d_infer){};
 
-  virtual ~TheoryBVSolver(){};
+  virtual ~BVSolver(){};
 
   /**
    * Returns true if we need an equality engine. If so, we initialize the
@@ -75,23 +74,18 @@ class TheoryBVSolver
 
   virtual TrustNode explain(TNode n)
   {
-    Unimplemented()
-        << "TheoryBVSolver propagated a node but doesn't implement the "
-           "TheoryBVSolver::explain() interface!";
+    Unimplemented() << "BVSolver propagated a node but doesn't implement the "
+                       "BVSolver::explain() interface!";
     return TrustNode::null();
   };
 
+  /**
+   * This is temporary only and will be deprecated soon in favor of
+   * Theory::collectModelValues.
+   */
   virtual bool collectModelInfo(TheoryModel* m) = 0;
 
   virtual std::string identify() const = 0;
-
-  virtual bool getCurrentSubstitution(int effort,
-                                      std::vector<Node>& vars,
-                                      std::vector<Node>& subs,
-                                      std::map<Node, std::vector<Node>>& exp)
-  {
-    return false;
-  };
 
   virtual Theory::PPAssertStatus ppAssert(
       TNode in, SubstitutionMap& outSubstitutions) = 0;
@@ -106,7 +100,7 @@ class TheoryBVSolver
 
   virtual EqualityStatus getEqualityStatus(TNode a, TNode b)
   {
-    return d_bv.getEqualityStatus(a, b);
+    return EqualityStatus::EQUALITY_UNKNOWN;
   }
 
   /** Called by abstraction preprocessing pass. */
@@ -114,7 +108,6 @@ class TheoryBVSolver
                                 std::vector<Node>& new_assertions) = 0;
 
  protected:
-  TheoryBV& d_bv;
   TheoryState& d_state;
   TheoryInferenceManager& d_inferManager;
 };
@@ -123,4 +116,4 @@ class TheoryBVSolver
 }  // namespace theory
 }  // namespace CVC4
 
-#endif /* CVC4__THEORY__BV__THEORY_BV_SOLVER_H */
+#endif /* CVC4__THEORY__BV__BV_SOLVER_H */
--- a/src/theory/bv/bv_subtheory_core.cpp
+++ b/src/theory/bv/bv_subtheory_core.cpp
@@ -19,7 +19,7 @@
 #include "options/bv_options.h"
 #include "options/smt_options.h"
 #include "smt/smt_statistics_registry.h"
-#include "theory/bv/theory_bv_lazy.h"
+#include "theory/bv/bv_solver_lazy.h"
 #include "theory/bv/theory_bv_utils.h"
 #include "theory/ext_theory.h"
 #include "theory/theory_model.h"
@@ -89,7 +89,7 @@ bool CoreSolverExtTheoryCallback::getReduction(int effort,
   return false;
 }
 
-CoreSolver::CoreSolver(context::Context* c, TheoryBVLazy* bv)
+CoreSolver::CoreSolver(context::Context* c, BVSolverLazy* bv)
     : SubtheorySolver(c, bv),
       d_notify(*this),
       d_isComplete(c, true),
--- a/src/theory/bv/bv_subtheory_core.cpp
+++ b/src/theory/bv/bv_subtheory_core.cpp
@@ -188,7 +188,8 @@ void CoreSolver::explain(TNode literal, std::vector<TNode>& assumptions) {
 bool CoreSolver::check(Theory::Effort e) {
   Trace("bitvector::core") << "CoreSolver::check \n";
 
-  d_bv->spendResource(ResourceManager::Resource::TheoryCheckStep);
+  d_bv->d_inferManager.spendResource(
+      ResourceManager::Resource::TheoryCheckStep);
 
   d_checkCalled = true;
   Assert(!d_bv->inConflict());
--- a/src/theory/bv/bv_subtheory_core.cpp
+++ b/src/theory/bv/bv_subtheory_core.cpp
@@ -19,7 +19,7 @@
 #include "options/bv_options.h"
 #include "options/smt_options.h"
 #include "smt/smt_statistics_registry.h"
-#include "theory/bv/theory_bv.h"
+#include "theory/bv/theory_bv_lazy.h"
 #include "theory/bv/theory_bv_utils.h"
 #include "theory/ext_theory.h"
 #include "theory/theory_model.h"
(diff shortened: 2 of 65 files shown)


src/theory/bv/theory_bv.h
```diff
@@ -115,177 +99,141 @@ class TheoryBV : public Theory {
...
+  /** A (default) theory state object */
+  TheoryState d_state;
 
-  // Are we in conflict?
-  context::CDO<bool> d_conflict;
+}; /* class TheoryBV */
 
-  // Invalidate the model cache if check was called
-  context::CDO<bool> d_invalidateModelCache;
+class TheoryBVSolver
```
ajreynol, line 138: I think this interface class might be overkill.  I think the end goal is to replace and delete the old solver, in which case we won't have multiple instances of this class.

Concretely my suggestion would be to have TheoryBv maintain a TheoryBvLazy class that is used temporarily while we migrate to the new solver.  Then this would allow the new implementation either to live directly within TheoryBV, or  otherwise the new solver would not have to be restricted to follow this interface (which is the job of TheoryBV anyways).

ajreynol (changes requested):
Some minor formatting requests and then a possible suggestion for the design.

src/theory/bv/theory_bv_solver.h
```diff
@@ -0,0 +1,130 @@
...
+  virtual void presolve() = 0;
+
+  virtual void notifySharedTerm(TNode t) = 0;
+
+  virtual EqualityStatus getEqualityStatus(TNode a, TNode b)
+  {
+    return EqualityStatus::EQUALITY_UNKNOWN;
+  }
+
+  virtual Node getModelValue(TNode var) = 0;
```
ajreynol, line 86: This is a deprecated interface for Theory, I do not expect this interface to be used by new solvers.

src/theory/bv/theory_bv_solver.h
```diff
@@ -0,0 +1,130 @@
...
+
+  // BV-specific methods
+  virtual bool applyAbstraction(const std::vector<Node>& assertions,
+                                std::vector<Node>& new_assertions) = 0;
+
+ protected:
+  TheoryBV& d_bv;
```
ajreynol, line 93: I'd add a reference to a TheoryState, and to a TheoryInferenceManager here.

src/theory/bv/theory_bv_solver.h
```diff
@@ -0,0 +1,130 @@
...
+
+  // Methods forwarded from TheoryBV
+
+  void computeAssertedTerms(std::set<Node>& termSet,
```
ajreynol, line 97: This method will be deleted from Theory soon.  I would make this specific to TheoryBVLazy for now.

src/theory/bv/theory_bv_solver.h
```diff
@@ -0,0 +1,130 @@
...
+  }
+
+  bool isLeaf(TNode node) { return d_bv.isLeaf(node); }
+
+  bool isLegalElimination(TNode x, TNode val)
+  {
+    return d_bv.isLegalElimination(x, val);
+  }
+
+  size_t numAssertions() { return d_bv.numAssertions(); }
```
ajreynol, line 111: This should not be used by new solvers, I would make this specific to TheoryBVLazy.

src/theory/bv/theory_bv_solver.h
```diff
@@ -0,0 +1,130 @@
...
+
+  theory::Assertion get() { return d_bv.get(); }
```
ajreynol, line 113: Ditto, this should not be used by new solvers, I would make this specific to TheoryBVLazy.

src/theory/bv/theory_bv_solver.h
```diff
@@ -0,0 +1,130 @@
...
+
+  bool done() { return d_bv.done(); }
```
ajreynol, line 115: Ditto, I would make this specific to TheoryBVLazy.

src/theory/bv/theory_bv_solver.h
```diff
@@ -0,0 +1,130 @@
...
+   * information regarding how it should be setup. For details, see the
+   * documentation in Theory::needsEqualityEngine.
+   */
+  virtual bool needsEqualityEngine(EeSetupInfo& esi) { return false; }
+
+  virtual void finishInit(){};
+
+  virtual void preRegisterTerm(TNode n) = 0;
+
+  virtual void check(Theory::Effort e) = 0;
```
ajreynol, line 47: This method should not be overridden by new solvers.  Instead, use preCheck / postCheck / preNotifyFact / notifyFact (these should probably be defined as virtual methods here).

src/theory/bv/theory_bv_solver.h
```diff
@@ -0,0 +1,130 @@
...
+  virtual TrustNode explain(TNode n)
+  {
+    Unimplemented() << "TheoryBVSolver::explain interface not implemented";
+  };
+
+  virtual bool collectModelInfo(TheoryModel* m) = 0;
+
+  virtual std::string identify() const = 0;
+
+  virtual bool getCurrentSubstitution(int effort,
```
ajreynol, line 62: This method is deprecated, it should be specific to TheoryBVLazy.

src/theory/bv/theory_bv_solver.h
```diff
@@ -0,0 +1,130 @@
...
+
+  virtual TrustNode ppRewrite(TNode t) { return TrustNode::null(); };
+
+  virtual void ppStaticLearn(TNode in, NodeBuilder<>& learned){};
+
+  virtual void presolve() = 0;
+
+  virtual void notifySharedTerm(TNode t) = 0;
+
+  virtual EqualityStatus getEqualityStatus(TNode a, TNode b)
```
ajreynol, line 81: This method should call d_bv.getEqualityStatus(a,b) as default I believe.

src/theory/bv/theory_bv_solver.h
```diff
@@ -0,0 +1,130 @@
...
+    return d_bv.isLegalElimination(x, val);
+  }
+
+  size_t numAssertions() { return d_bv.numAssertions(); }
+
+  theory::Assertion get() { return d_bv.get(); }
+
+  bool done() { return d_bv.done(); }
+
+  OutputChannel* getOutputChannel() { return d_bv.d_out; }
```
ajreynol, line 117: Output channel should not be used directly in new solvers, if necessary I'd make this specific to TheoryBVLazy.

src/theory/bv/theory_bv_solver.h
```diff
@@ -0,0 +1,130 @@
...
+
+  Valuation& getValuation() { return d_bv.d_valuation; }
```
ajreynol, line 119: Ditto, valuation should generally not be used directly, make specific to TheoryBVLazy (this is a part of TheoryState).

src/theory/bv/theory_bv_solver.h
```diff
@@ -0,0 +1,130 @@
...
+
+  eq::EqualityEngine* getEqualityEngine() { return d_bv.getEqualityEngine(); }
+
+  ExtTheory* getExtTheory() { return d_bv.d_extTheory.get(); }
```
ajreynol, line 123: This should only exist in TheoryBVLazy.

src/theory/bv/theory_bv_solver.h
```diff
@@ -0,0 +1,130 @@
...
+
+  theory::Assertion get() { return d_bv.get(); }
+
+  bool done() { return d_bv.done(); }
+
+  OutputChannel* getOutputChannel() { return d_bv.d_out; }
+
+  Valuation& getValuation() { return d_bv.d_valuation; }
+
+  eq::EqualityEngine* getEqualityEngine() { return d_bv.getEqualityEngine(); }
```
ajreynol, line 121: This should not be defined here, make this specific to TheoryBVLazy.

ajreynol (changes requested):
A number of interfaces in TheoryBVSolver are not necessary and should be confined if possible to the old solver. I've noted which of the interfaces should definitely be deleted, although I'm not sure how easily some of these things can currently be untangled.

ajreynol:
One open issue is the TheoryState and TheoryInferenceManager objects to be used by the new BV solver(s).  It is possible that one could want a custom state / inference manager for each new BV solver.

src/theory/bv/theory_bv_solver.h
ajreynol: Also notice that TheoryBVLazy can implement its current check as a preCheck, returning true.  (The new standard is strictly more flexible wrt the old one).

src/theory/bv/theory_bv_solver.h
ajreynol: I believe you can delete this virtual interface now without any issues.

src/theory/bv/theory_bv_solver.h
```diff
@@ -0,0 +1,126 @@
...
+
+  virtual TrustNode explain(TNode n)
+  {
+    Unimplemented()
+        << "TheoryBVSolver propagated a node but doesn't implement the "
+           "TheoryBVSolver::explain() interface!";
+    return TrustNode::null();
+  };
+
+  virtual bool collectModelInfo(TheoryModel* m) = 0;
```
ajreynol, line 84: Notice that Theory is not recommended to implement collectModelInfo anymore. This method will be replaced by "collectModelValues" (see https://github.com/ajreynol/CVC4/blob/master/src/theory/theory.h#L699), which has a simpler contract.  However, this will require some changes to the old lazy BV solver to use this interface, which I have on my branch.

Thus, for now, I am fine with having this virtual method to make this PR go through, but I'd recommend marking this as temporary with a comment.  I can replace this method with the new one (collectModelValues) on a subsequent PR.

src/theory/bv/theory_bv_solver.h
```diff
@@ -0,0 +1,126 @@
...
+  {
+    return d_bv.getEqualityStatus(a, b);
+  }
+
+  /** Called by abstraction preprocessing pass. */
+  virtual bool applyAbstraction(const std::vector<Node>& assertions,
+                                std::vector<Node>& new_assertions) = 0;
+
+ protected:
+  TheoryBV& d_bv;
```
ajreynol, line 117: It's a bit ugly to have a backwards reference to TheoryBV here.

Actually, can we move this to LazyBVSolver? (On second thought, getEqualityStatus should perhaps just return UNKNOWN here and not rely on d_bv, as now this is the only use of d_bv in this base class). This would make it so that TheoryBV uses this solver but not the other way around (with the exception of the old solver).

src/theory/theory_inference_manager.h
```diff
@@ -278,6 +278,17 @@ class TheoryInferenceManager
    * decided with polarity pol, for details see OutputChannel::requirePhase.
    */
   void requirePhase(TNode n, bool pol);
+
+  /**
+   * Forward to OutputChannel::spendResource() to spend resources..
```
ajreynol, line 283:
```suggestion
   * Forward to OutputChannel::spendResource() to spend resources.
```

src/theory/theory_inference_manager.h
```diff
@@ -278,6 +278,17 @@ class TheoryInferenceManager
...
+   */
+  void spendResource(ResourceManager::Resource r);
+
+  /**
+   * Forward to OutputChannel::safePoint() to spend resources..
```
ajreynol, line 288:
```suggestion
   * Forward to OutputChannel::safePoint() to spend resources.
```

src/theory/theory_state.h
```diff
@@ -72,6 +72,8 @@ class TheoryState
   /** Are we currently in conflict? */
   virtual bool isInConflict() const;
 
+  virtual bool isSatLiteral(TNode lit) const;
```
ajreynol, line 75:
```suggestion
  /** Returns true if lit is a SAT literal */
  virtual bool isSatLiteral(TNode lit) const;
```

ajreynol (changes requested):

src/theory/bv/theory_bv_solver.h
```diff
@@ -0,0 +1,126 @@
...
+#define CVC4__THEORY__BV__THEORY_BV_SOLVER_H
+
+#include "theory/bv/theory_bv.h"
+#include "theory/theory.h"
+
+namespace CVC4 {
+namespace theory {
+namespace bv {
+
+class TheoryBVSolver
```
ajreynol, line 29: Also, another comment here, how about the name `BVSolver`?  This makes it clear that this isn't a Theory object.

src/theory/bv/theory_bv_solver.h
mpreiner: Yep, that's possible. I moved it to TheoryBVLazy.

src/theory/bv/theory_bv_solver.h
mpreiner: That's a good point, I'll do more renaming to make this clearer.

src/theory/bv/bv_solver.h
```diff
@@ -0,0 +1,119 @@
...
+ **
+ ** Describes the interface for the internal bit-vector solver of TheoryBV.
+ **/
+
+#include "cvc4_private.h"
+
+#ifndef CVC4__THEORY__BV__BV_SOLVER_H
+#define CVC4__THEORY__BV__BV_SOLVER_H
+
+#include "theory/bv/theory_bv.h"
```
ajreynol, line 22: Perhaps we can remove this after the suggestion about the constructor below.

src/theory/bv/bv_solver.h
```diff
@@ -0,0 +1,119 @@
...
+#include "theory/theory.h"
+
+namespace CVC4 {
+namespace theory {
+namespace bv {
+
+class BVSolver
+{
+ public:
+  BVSolver(TheoryBV& bv) : d_state(bv.d_state), d_inferManager(bv.d_infer){};
```
ajreynol, line 32: I think we should just pass reference to a state and inference manager here.  The constructor for BVSolverLazy can take all 3 if it needs (or just TheoryBV from which these can be extracted), but we can avoid TheoryBV here.

src/theory/bv/theory_bv.h
```diff
@@ -22,40 +22,29 @@
...
-class CoreSolver;
-class InequalitySolver;
-class AlgebraicSolver;
-class BitblastSolver;
+class BVSolver;
 
-class EagerBitblastSolver;
+class TheoryBV : public Theory
+{
+  friend class BVSolver;
```
ajreynol, line 43: Might be able to delete this too.

ajreynol (changes requested):
A few minor things, getting close.

src/theory/bv/bv_subtheory_core.cpp
```diff
@@ -566,7 +567,7 @@ bool CoreSolver::doExtfInferences(std::vector<Node>& terms)
                               nm->mkNode(kind::LT, n, max));
         Trace("bv-extf-lemma")
             << "BV extf lemma (range) : " << lem << std::endl;
-        d_bv->getOutputChannel().lemma(lem);
+        d_bv->d_inferManager.lemma(lem);
```
ajreynol, line 570: You should use `d_inferManager` from the base class here (there are several other similar places like this).  I would not reference TheoryBV's inference manager like this since where this lives is potentially subject to change.

ajreynol (approved):
One last change, else LGTM.

PR #5010, opened 2020-09-02
Make ExtTheory independent of Theory

ajreynol:
This makes it so that ExtTheory uses a generic callback instead of relying on Theory.

The primary purpose of this PR is to eliminate the connection of TheoryBV and ExtTheory.  This PR moves all things related to ExtTheory in BV into CoreSolver.

It also refactors the use of ExtTheory in strings and arithmetic.


mpreiner (approved):
LGTM

PR #5012, opened 2020-09-03
(proof-new) Handle mismatched assumptions in proof equality engine during mkScope

ajreynol:
This modifies the fix for Boolean equalities with constants so that is addressed lazily during ProofNodeManager mkScope instead of when asserting assumptions to the proof equality engine.  This ensures that the default method for asserting facts to the equality engine for external assertions does not have any special cases.

A previously abandoned solution to this issue had made this a part of CDProof. Instead, this PR modifies the mkScope method only. The fix makes mkScope robust to *any* assumption mismatches in mkScope that can be fixed by rewriting, not just Boolean equality with constants.  It is intended to be used infrequently as a last resort when mkScope has mismatched assumptions.  The handling of mismatches due to symmetry remains in this method.


src/expr/proof_node_manager.cpp
```diff
@@ -83,30 +91,61 @@ std::shared_ptr<ProofNode> ProofNodeManager::mkScope(
...
+      std::unordered_map<Node, Node, NodeHashFunction>::iterator itr =
+          acr.find(ar);
+      if (itr != acr.end())
+      {
+        aMatch = itr->second;
+      }
+    }
+
+    // if we found a match either by symmetry or rewriting, then we connect
+    // the assumption here.
```
HanielB, line 129: Do I understand correctly that `MACRO_SR_PRED_TRANSFORM` steps are being added for symmetry as well? I'd rather have it special cased.

ajreynol: No, we are applying symmetry whenever possible, and only resort to rewriting when this fails.

HanielB: But the update to the proof node, when the assumption changed due to symmetry, is also represented as a `MACRO_SR_PRED_TRANSFORM` step, no?

ajreynol: Ah, so yeah, I am preserving the old behavior here, which indeed used MACRO_SR_PRED_TRANSFORM.  You have to be a bit careful here since SYMM has special status in CDProof.

I think we should do an optimization pass later where we eliminate these when possible, perhaps a separate PR.

HanielB: Probably we could defer this to "global optimizations of proof nodes". Perhaps to be done before we expand the coarse grained steps.

HanielB (approved):

PR #5013, opened 2020-09-03
(proof-new) Use deep copy for base lemma steps in proof equality engine

ajreynol:
Previously, strings on proof-new used a ProofStepBuffer for deriving the final conclusion of lemmas in proof equality engine (assertLemma).  This was recently changed to use the more standard ProofGenerator interface of assertLemma.  However, there was a bug in this method: the steps from this proof need to be deep copied into the proof we are generating or otherwise the explanation of these literals are not connected in the CDProof.  This updates the proof equality engine so that it instructs the proof generator to do a deep copy in this case.  Notice this is very rarely more than a single step.


HanielB (approved):
LGTM

PR #5015, opened 2020-09-03
Use arith::InferenceManager for CAD lemmas

nafur:
This makes the CAD solver use the new `arith::InferenceManager` instead of the previously used lemma collection scheme.


src/theory/arith/nl/nonlinear_extension.cpp
```diff
@@ -557,12 +557,15 @@ int NonlinearExtension::checkLastCall(const std::vector<Node>& assertions,
   }
   if (options::nlCad())
   {
-    lemmas = d_cadSlv.checkFull();
-    if (lemmas.empty())
+    d_cadSlv.checkFull();
+    if (!d_im.hasProcessed())
```
ajreynol, line 561: Should be careful here about hasPending vs. hasProcessed, this is being changed in https://github.com/CVC4/CVC4/pull/5007/files.  Can you say what you're wanting to check here?  E.g. conflict or any pending lemma (or already sent a lemma)?

nafur: I want to check (and I guess this is the check we want to have all over the place in the nonlinear extension) whether any conflict, lemma or fact has been added. Essentially, whether the previous call has done anything to advance the solving process. Right now, this would probably be `hasSentFact() || hasSentLemma() || hasPendingFact() || hasPendingLemma() || isInConflict()`...

ajreynol: Sets often uses `hasSentFact() || hasSentLemma() || isInConflict()` = `TheoryInferenceManager::hasProcessed()`.

Strings often uses `hasPendingFact() || hasPendingLemma() || isInConflict()` = `InferenceManagerBuffered::hasPending() || TheoryState::isInConflict()`.

So i think you'd want to do `TheoryInferenceManager::hasProcessed() || InferenceManagerBuffered::hasPending()` for this after the above PR is merged.  We could perhaps make this a helper method in `InferenceManagerBuffered`?

nafur: Yes, this probably makes sense. And we should have a good naming scheme for all these functions...

src/theory/arith/nl/nonlinear_extension.cpp
```diff
@@ -557,12 +557,15 @@ int NonlinearExtension::checkLastCall(const std::vector<Node>& assertions,
...
-    if (lemmas.empty())
+    d_cadSlv.checkFull();
+    if (!d_im.hasMadeProgress())
     {
       Trace("nl-cad") << "nl-cad found SAT!" << std::endl;
     }
-    filterLemmas(lemmas, wlems);
+    else
+    {
+      return 1;
```
ajreynol, line 567: This doesnt fit with the return value of this method, I assume this will change later? This entire method is probably being heavily refactored and moved somewhere else I'm guessing? Maybe add a comment here.

ajreynol (approved):
LGTM, one minor question/request.

ajreynol:
Also would suggest another name change `hasProcessed` / `hasMadeProgress` -> `hasSent` / `hasProcessed`

PR #5016, opened 2020-09-03
Add interfaces for making trust nodes in TheoryInferenceManager.

ajreynol:
This gives theories a finer grained control over explained lemmas and conflicts.

A theory may now use an inference manager to construct "explained" lemmas/conflicts e.g. via mkLemmaExp, subsequently do any theory-specific debugging or modification to that lemma before sending it via trustedLemma.

This is required for the new strings inference manager on proof-new.

This also adds a missing variant of conflicts for the proof equality engine.  It also does a minor simplification of a previous variant for constructing conflicts from proof equality engine based on a proof step buffer.


src/theory/theory_inference_manager.cpp
```diff
@@ -90,21 +92,52 @@ void TheoryInferenceManager::conflictExp(PfRule id,
...
+  }
+  // version without proofs
+  Node conf = mkExplainPartial(exp, {});
+  return TrustNode::mkTrustConflict(conf, nullptr);
+}
+
+void TheoryInferenceManager::conflictExp(const std::vector<Node>& exp,
+                                         ProofGenerator* pg)
+{
+  if (!d_theoryState.isInConflict())
```
nafur, line 119: It feels wrong to me to silently drop conflicts, just because we already have a conflict. Could we redirect this conflict as a lemma in this case?

nafur (approved):
Just one comment...

src/theory/theory_inference_manager.cpp
ajreynol: Note this isn't specific to this PR (the other variants have this), but I agree we could reconsider this.  If anything, I believe we would just remove the guard and send another conflict.

PR #5017, opened 2020-09-03
ICP-based solver for nonlinear arithmetic

nafur:
This PR adds a new icp-based solver to be integrated into the nonlinear extension.
It is not meant to be used as a stand-alone ICP solver. It does not implement splits (only propagations) and implements a rather aggressive budget mechanism that aims to quickly stop propagation to allow other solvers to take over. Additionally, it enforces a maximum bit size to avoid divergence.


ajreynol:
A few high-level things: can you post some results as a comment here? Also, indicate how to use the new solver in the PR description (which command line option).  I assume this solver is for QF_NRA?  What about other logics with non-linear real or integer?

nafur:
The new solver is extends the current nonlinear solver by a new way to generate lemmas and conflicts. Right now, it is enabled via `--nl-icp` and run first (before all other subsolvers). We should look at whether this is a good idea and how to properly integrate it when we refactor the nl strategy altogether.

It starts with the "full assignment" (all variables assigned to `(-infty, infty)`) and then uses constraints to contract these (solve constraint for one variable, substitute current assignment into right hand side, intersect current assignment with the resulting interval). The result can either be a new (or stricter) bound on a variable that is learnt as a lemma, or a conflict if an interval becomes the empty interval.

It is currently targeted at NRA, but should be sound for NIA as well. Better/additional propagations should be possible for integer variables, but I have not implemented this yet.

This solver slightly improves upon the currently best configuration (`QF_NRA`, `--theoryof-mode=type --nl-rlv=interleave --no-nl-ext --nl-cad` solves 10170 and `--theoryof-mode=type --nl-rlv=interleave --no-nl-ext --nl-icp --nl-cad` solves 10178).

There is still a bunch of possible improvements: the heuristics in the solver are very simple (which contraction candidates should we try? when should we stop? which lemmas should we learn?), integer variables are not exploited (simply contract intervals to the next integer bounds), integration into the nonlinear extension is extremely naive, the current model is not checked (we could stop as soon as we have excluded the current linear model).

src/theory/arith/nl/icp/icp_solver.cpp
```diff
@@ -0,0 +1,364 @@
...
+
+        d_im.addConflict(d_state.d_conflict, InferenceId::NL_ICP_CONFLICT);
+        did_progress = true;
+        progress = false;
+        break;
+    }
+  } while (progress);
+  if (did_progress)
+  {
+    for (const auto& l : generateLemmas())
```
ajreynol, line 351: This loop is really crazy to me :)

I guess this is assuming a fixed call to generateLemmas() and then iterates over what is returned?  Regardless can you cache a reference and iterate?

src/theory/arith/nl/icp/intersection.cpp
```diff
@@ -0,0 +1,233 @@
...
+    Trace("nl-icp") << "cur covers res but meet at upper" << std::endl;
+    cur = poly::Interval(get_lower(res),
+                         get_lower_open(res),
+                         get_upper(res),
+                         get_upper_open(cur) || get_upper_open(res));
+    if (get_upper_open(cur) && !get_upper_open(res))
+    {
+      return PropagationResult::CONTRACTED;
+    }
+    else
```
ajreynol, line 163: Nit: this else is not needed.

src/theory/arith/nl/poly_conversion.h
```diff
@@ -65,6 +65,9 @@ poly::UPolynomial as_poly_upolynomial(const CVC4::Node& n,
  * denominator (except for its sign, which is always positive, though).
  */
 poly::Polynomial as_poly_polynomial(const CVC4::Node& n, VariableMapper& vm);
+poly::Polynomial as_poly_polynomial(const CVC4::Node& n,
```
ajreynol, line 68: Document the additional argument.

src/theory/arith/normal_form.h
```diff
@@ -1377,6 +1389,9 @@ class Comparison : public NodeWrapper {
   Polynomial normalizedVariablePart() const;
   DeltaRational normalizedDeltaRational() const;
 
+  std::tuple<Polynomial, Kind, Constant> decompose(
```
ajreynol, line 1392: Document.

src/theory/theory_inference_manager.h
```diff
@@ -132,6 +132,10 @@ class TheoryInferenceManager
    * only be called if there is not yet proof support in the given theory.
    */
   void conflict(TNode conf);
+  /** The number of conflicts we have added since the last call to reset */
```
ajreynol, line 135: I don't think I like these methods: TheoryState currently says whether we are in conflict, this makes some confusion about what is the right place to ask.

ajreynol (changes requested):
Some requests.

src/theory/arith/nl/icp/icp_solver.cpp
nafur: Tbh, I really consider this to be better / more readable than the alternative:
```suggestion
    std::vector<Node> lemmas = generateLemmas();
    for (const auto& l : lemmas)
```
If you really prefer it, I'll change :-)

src/theory/arith/nl/icp/intersection.cpp
nafur: Agree.

src/theory/theory_inference_manager.h
nafur: Hm, I agree. I've removed them and check whether this has any consequences...

nafur: Still works.

src/theory/arith/nl/icp/icp_solver.cpp
ajreynol: Yes, I really prefer it :)

src/theory/arith/nl/icp/contraction_origins.cpp
```diff
@@ -0,0 +1,123 @@
...
+  {
+    print(os, indent + "\t", o);
+  }
+}
+
+inline std::ostream& operator<<(std::ostream& os,
+                                const ContractionOriginManager& com)
+{
+  os << "ContractionOrigins:" << std::endl;
+  for (const auto& vars : com.currentOrigins())
```
ajreynol, line 111: Ditto about caching here.

src/theory/arith/nl/icp/icp_solver.cpp
```diff
@@ -0,0 +1,364 @@
...
+        case Kind::DISTINCT: rel = poly::SignCondition::NE; break;
+        case Kind::GT: rel = poly::SignCondition::GT; break;
+        case Kind::GEQ: rel = poly::SignCondition::GE; break;
+        default: Assert(false) << "Unexpected kind: " << k;
+      }
+      poly::Rational rhsmult;
+      poly::Polynomial rhs = as_poly_polynomial(val, d_mapper, rhsmult);
+      rhsmult = poly::Rational(1) / rhsmult;
+      // only correct up to a constant (denominator is thrown away!)
+      // std::cout << "rhs = " << rhs << std::endl;
```
ajreynol, line 120: Nit: delete this.

ajreynol (approved):
LGTM, just a few minor requests.

PR #5020, opened 2020-09-04
bv2int: improvement in lazy failures

yoni206:
Previously, in case we encountered a term we cannot translate to integers (e.g. a bit-vector array), we would fail.
This PR improves that behavior by keeping the original term, and potentially wrapping it BV_TO_NAT and INT_TO_BV operators.

A test which includes a bit-vector array is included, which was not supported before. 

In addition, the treatment of uninterpreted functions is refactored and documented better.


src/preprocessing/passes/bv_to_int.cpp
```diff
@@ -688,11 +688,11 @@ Node BVToInt::bvToInt(Node n)
               /*
                * We replace all BV-sorts of the domain with INT
                * If the range is a BV sort, we replace it with INT
-               * We cache both the term itself (e.g., f(a)) and the function
-               * symbol f.
                */
 
-              //Construct the function itself
+              /*
```
ajreynol, line 693: A minor style comment: I prefer not using `/* */` in comments in cpp (it makes it harder to use block comments for disabling parts of the code for debugging).

src/preprocessing/passes/bv_to_int.cpp
```diff
@@ -725,57 +726,146 @@ Node BVToInt::bvToInt(Node n)
...
+              {
+                builder << child;
               }
-              else {
-                d_bvToIntCache[current] =
-                    d_nm->mkNode(oldKind, translated_children);
+              Node translation = builder.constructNode();
+              /**
+               * if the adjusted term is of type bit-vector
+               * we use BITVECTOR_TO_NAR.
```
ajreynol, line 862:
```suggestion
               * we use BITVECTOR_TO_NAT.
```

ajreynol (changes requested):
A few style comments.

Also, I'm not really following what the code is doing since this piece of code is really big.  Can you perhaps make a helper method like "ensureType" that ensures a term is casted to a particular type?

src/preprocessing/passes/bv_to_int.cpp
```diff
@@ -725,57 +724,92 @@ Node BVToInt::bvToInt(Node n)
...
+                achildren.push_back(intUF);
+                int i = 0;
+                vector<Expr> args;
+                for (TypeNode d : bvDomain)
+                {
+                  // Each bit-vector argument is casted to a natural number
+                  // using BITVECTOR_TO_NAT operator.
+                  // Other arguments are left intact.
+
+                  Node fresh_bound_var = d_nm->mkBoundVar(d);
```
ajreynol, line 743: Can't this use the helper method?

ajreynol (changes requested):

src/preprocessing/passes/bv_to_int.cpp
```diff
@@ -787,19 +786,118 @@ Node BVToInt::bvToInt(Node n)
...
+    Assert(tn.isInteger());
+    return d_nm->mkNode(kind::BITVECTOR_TO_NAT, n);
+  }
+}
+
+Node BVToInt::reconstructNode(Node node)
+{
+  // first, we adjust the children of the node as needed.
+  vector<Node> adjusted_children;
+  for (Node child : node)
```
ajreynol, line 880:
```suggestion
  for (const Node& child : node)
```

src/preprocessing/passes/bv_to_int.cpp
```diff
@@ -787,19 +786,118 @@ Node BVToInt::bvToInt(Node n)
...
+  // add the function definition to the smt engine.
+  smt::currentSmtEngine()->defineFunction(
+      bvUF.toExpr(), args, intApplication.toExpr(), true);
+}
+
+bool BVToInt::childrenTypesChanged(Node n)
+{
   bool result = false;
-  for (Node child : n) {
+  for (Node child : n)
```
ajreynol, line 833:
```suggestion
  for (const Node& child : n)
```

src/preprocessing/passes/bv_to_int.cpp
```diff
@@ -787,19 +786,118 @@ Node BVToInt::bvToInt(Node n)
...
+  }
+  else
+  {
+    Assert(n.getType().isBitVector());
+    Assert(tn.isInteger());
+    return d_nm->mkNode(kind::BITVECTOR_TO_NAT, n);
+  }
+}
+
+Node BVToInt::reconstructNode(Node node)
```
ajreynol, line 876: Can you have this method take the type to reconstruct to instead of always casting to an integer below?

src/preprocessing/passes/bv_to_int.cpp
```diff
@@ -787,19 +786,118 @@ Node BVToInt::bvToInt(Node n)
...
+    if (!newType.isSubtypeOf(originalType))
+    {
       result = true;
       break;
     }
   }
   return result;
 }
 
+Node BVToInt::castIfNeeded(Node n, TypeNode tn)
```
ajreynol, line 846: I think it'd be clearer if this method was merged with castToType (e.g. castToType always did these checks).

src/preprocessing/passes/bv_to_int.cpp
```diff
@@ -787,19 +786,118 @@ Node BVToInt::bvToInt(Node n)
...
+{
+  if ((n.getType().isBitVector() && tn.isInteger())
+      || (n.getType().isInteger() && tn.isBitVector()))
+  {
+    return castToType(n, tn);
+  }
+  else
```
ajreynol, line 853: We tend to not use `else` after a `return` statement for compactness.

ajreynol (changes requested):
A few more minor requests.

src/preprocessing/passes/bv_to_int.cpp
```diff
@@ -787,19 +787,107 @@ Node BVToInt::bvToInt(Node n)
...
+  if (node.getMetaKind() == kind::metakind::PARAMETERIZED)
+  {
+    builder << node.getOperator();
+  }
+  for (Node child : adjusted_children)
+  {
+    builder << child;
+  }
+  Node reconstruction = builder.constructNode();
+  // cast back to integers if the result is a bit-vector.
```
ajreynol, line 887: Update comment.

src/preprocessing/passes/bv_to_int.cpp
```diff
@@ -787,19 +787,107 @@ Node BVToInt::bvToInt(Node n)
...
       result = true;
       break;
     }
   }
   return result;
 }
 
+Node BVToInt::castToType(Node n, TypeNode tn)
+{
+  if (!((n.getType().isBitVector() && tn.isInteger())
```
ajreynol, line 849: Can you just say `n.getType().isSubtypeOf(tn)` here?

ajreynol (changes requested):

ajreynol (approved):
LGTM, thanks for the improvements.

PR #5021, opened 2020-09-04
Remove unused postsolve infrastructure

4tXJ7f:
The theory property `postsolve` was not used by any of the theories.
This commit removes the property and associated infrastructure.

### DIFF
--- a/src/theory/theory_engine.cpp
+++ b/src/theory/theory_engine.cpp
@@ -749,34 +749,6 @@ bool TheoryEngine::presolve() {
   return false;
 }/* TheoryEngine::presolve() */
 
-void TheoryEngine::postsolve() {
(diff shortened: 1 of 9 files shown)


src/theory/theory_engine.cpp
```diff
@@ -749,34 +749,6 @@ bool TheoryEngine::presolve() {
   return false;
 }/* TheoryEngine::presolve() */
 
-void TheoryEngine::postsolve() {
-  // no longer in SAT mode
-  d_inSatMode = false;
```
ajreynol, line 754: So the one change in this PR is that these flags are not reset, right?

Are these reset somewhere else?

ajreynol (changes requested):
One question.

PR #5022, opened 2020-09-04
Use new inference manager in transcendental solver

nafur:
This refactors the transcendental solver to add lemmas to the new inference manager instead of using the old lemma collection scheme.


ajreynol (approved):
LGTM.

Marking "do not merge" until we see the performance impact on the QF_NRAT benchmarks.

PR #5023, opened 2020-09-04
Improve OSX support by adding os detection and adapting calls for OSX during glpk install.

mmuesly:
On OSX there is another libtool already available.
The gnu libtoolize version installed with MacPorts or brew is called glibtoolize.
This change makes it easier to run the file on OSX.

Signed-off-by: Malte Mues (mail.mues@gmail.com)


4tXJ7f (approved):
Thank you for your contribution and sorry about the delay. This looks good to me.

PR #5025, opened 2020-09-04
(new theory) Update TheoryBV to new standard for collectModelValues

ajreynol:
This makes collectModelValues the main model interface in BV instead of collectModelInfo.  BV is no longer responsible for asserting its equality engine or computing relevant/asserted terms.

This involved updating the interface on many subsolvers of BvSolverLazy.  This includes moving the responsibility of addSharedTerm (regarding trigger terms) from the subsolvers to TheoryBV, this eventually will be automatically handled in Theory once all theories are migrated to the new standard.

This ensures that TheoryBV is updated to the new standard (check was already migrated on https://github.com/CVC4/CVC4/commit/c9e23f66383a4d490aca6d082d40117fe799ee4b).


mpreiner (approved):
LGTM. I think we can also eliminate the fullModel argument since I believe it's always `true`.

mpreiner (approved):

PR #5026, opened 2020-09-04
Use Result::Sat instead of BenchmarkStatus in printers.

abdoo8080:
This PR modifies the printers to use `Result::Sat`, "internal" version of `BenchmarkStatus`, for printing benchmark status commands.

### DIFF
--- a/src/smt/command.cpp
+++ b/src/smt/command.cpp
@@ -2981,10 +2981,8 @@ void SetBenchmarkStatusCommand::toStream(std::ostream& out,
   switch (d_status)
   {
     case BenchmarkStatus::SMT_SATISFIABLE: status = Result::SAT; break;
-    case BenchmarkStatus::SMT_UNSATISFIABLE:
-      status = Result::SAT_UNKNOWN;
-      break;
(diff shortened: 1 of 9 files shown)


src/smt/command.cpp
```diff
@@ -2977,7 +2977,17 @@ void SetBenchmarkStatusCommand::toStream(std::ostream& out,
                                          size_t dag,
                                          OutputLanguage language) const
 {
-  Printer::getPrinter(language)->toStreamCmdSetBenchmarkStatus(out, d_status);
+  Result::Sat status;
+  switch (d_status)
+  {
+    case BenchmarkStatus::SMT_SATISFIABLE: status = Result::SAT; break;
+    case BenchmarkStatus::SMT_UNSATISFIABLE:
+      status = Result::SAT_UNKNOWN;
```
ajreynol, line 2985: I think this was swapped?

ajreynol (changes requested):
One typo I think.

src/smt/command.cpp
abdoo8080: Fixed!

ajreynol (approved):
LGTM

PR #5027, opened 2020-09-04
Refactor collectModelInfo in TheoryArith

ajreynol:
This is work towards updating the arithmetic solver to the new standard, and in particular isolating TheoryArithPrivate as the "linear solver", and TheoryArith as the overall approach for arithmetic.

This transfers ownership of the non-linear extension from TheoryArithPrivate to TheoryArith.  The former still has a pointer to the non-linear extension, which will be removed with further refactoring.

This PR additionally moves the code that handles the interplay of the non-linear extension in TheoryArithPrivate::collectModelInfo to TheoryArith, and simplifies the model interface for TheoryArithPrivate.


src/theory/arith/theory_arith.cpp
```diff
@@ -123,7 +132,50 @@ void TheoryArith::propagate(Effort e) {
...
+    // an equality status that was agreed upon by both (linear) arithmetic
+    // and another theory. In this case, we must add a lemma, or otherwise
+    // we would terminate with an invalid model. Thus, we add a splitting
+    // lemma of the form ( x = v V x != v ) where v is the model value
+    // assigned by the non-linear solver to x.
+    if (d_nonlinearExtension != nullptr)
+    {
+      Node eq = p.first.eqNode(p.second);
+      Node lem = NodeManager::currentNM()->mkNode(kind::OR, eq, eq.negate());
+      d_out->lemma(lem);
```
nafur, line 174: This becomes interesting when CAD constructs a non-rational model... we should expect this to explode at some point :-)
Maybe add an assertion already that `!eq.isNull()` or something like that?

src/theory/arith/theory_arith.cpp
```diff
@@ -123,7 +132,50 @@ void TheoryArith::propagate(Effort e) {
...
+  d_internal->collectModelValues(termSet, arithModel);
+  // if non-linear is enabled, intercept the model, which may repair its values
+  if (d_nonlinearExtension != nullptr)
+  {
+    // Non-linear may repair values to satisfy non-linear constraints (see
+    // documentation for NonlinearExtension::interceptModel).
+    d_nonlinearExtension->interceptModel(arithModel);
+  }
+  // We are now ready to assert the model.
+  for (std::pair<const Node, Node>& p : arithModel)
```
nafur, line 157: I think this can be `const`:
```suggestion
  for (const std::pair<const Node, Node>& p : arithModel)
```

nafur (changes requested):
Just two comments.

src/theory/arith/theory_arith.cpp
ajreynol: Do you mean to check that p.first / p.second are not null? (eq is obvious not null)

nafur: I'm not entirely sure how all this works together, but CAD models may come as witness terms and thus `p.first` may be an arithmetic variable while `p.second` may be a Boolean formula. Unless witness terms are handled differently... but I don't really see that anywhere.

ajreynol: We should always ensure p.first and p.second have the same type, through witness.

This map is passed through to NlModel, updated possibly with witness terms here: https://github.com/CVC4/CVC4/blob/master/src/theory/arith/nl/nl_model.cpp#L1260.

nafur (approved):
LGTM.

PR #5030, opened 2020-09-04
Add asLemma flag to theory inference process

ajreynol:
This is required for strings, which uses the same data structure, InferInfo, for both lemmas and facts.  This ensures the process method of theory inference knows where we are a pending lemma or a pending fact.

It also makes a few changes necessary for the proof-new branch, including disabling the proof node manager in the inference manager for datatypes.


4tXJ7f (approved):
LGTM, just a typo.

PR #5031, opened 2020-09-04
Use lemma caching in theory of strings

ajreynol:
This updates theory of strings to do lemma caching, which is primarily required for the LEN_NORM rule.

This updates the strategy loop using the new theory inference manager features to achieve this, by checking `hasSentLemma` instead of `hasPendingLemma`.

NOTE: this should be performance tested.


ajreynol:
Closing this for now, I believe we determined this doesnt improve performance.

PR #5033, opened 2020-09-07
Add is_singleton operator to the theory of sets

mudathirmahgoub:
This pull request adds a new operator `is_singleton` for sets to determine whether a set is a singleton. 
Before this operator, the user either uses an existential quantifier or the cardinality operator in smtlib. The former affects the performance especially when polarity is negative. The latter requires the cardinality extension in sets theory. 
The implementation of  `is_singleton` only uses an internal existential quantifier with the hope of optimizing it later. 
New rewriting rules:
 - `(is_singleton (singleton x))` is rewritten as `true`.
 - `(is_singleton A)` is rewritten as `(= A (singleton x))` where `x` is a fresh skolem of the same type of elements of A.
 - `(choose (singleton x))` is rewritten as `x`.


src/theory/sets/theory_sets_private.cpp
```diff
@@ -1391,35 +1391,80 @@ TrustNode TheorySetsPrivate::expandDefinition(Node node)
...
-    return TrustNode::mkTrustRewrite(node, witness, nullptr);
+    case kind::CHOOSE: return expandChooseOperator(node);
+    case kind::IS_SINGLETON: return expandIsSingletonOperator(node);
+    default: return TrustNode::null();
+  }
+}
+
+TrustNode TheorySetsPrivate::expandChooseOperator(const Node& node)
+{
+  if (node[0].getKind() == SINGLETON)
```
ajreynol, line 1404: This should probably be part of the rewriter. Is there a reason to have it here?

src/theory/sets/theory_sets_private.cpp
```diff
@@ -1391,35 +1391,80 @@ TrustNode TheorySetsPrivate::expandDefinition(Node node)
...
+  Node memberAndEqual = member.andNode(equal);
+  Node ite = nm->mkNode(ITE, isEmpty, equal, memberAndEqual);
+  Node witnessVariables = nm->mkNode(BOUND_VAR_LIST, witnessVariable);
+  Node witness = nm->mkNode(WITNESS, witnessVariables, ite);
+  return TrustNode::mkTrustRewrite(node, witness, nullptr);
+}
+
+TrustNode TheorySetsPrivate::expandIsSingletonOperator(const Node& node)
+{
+  if (node[0].getKind() == SINGLETON)
```
ajreynol, line 1439: Ditto, this should be part of the rewriter.

src/theory/sets/theory_sets_type_rules.h
```diff
@@ -243,6 +243,31 @@ struct ChooseTypeRule
...
+        throw TypeCheckingExceptionPrivate(
+            n, "IS_SINGLETON operator expects a set, a non-set is found");
+      }
+    }
+    return nodeManager->booleanType();
+  }
+  inline static bool computeIsConst(NodeManager* nodeManager, TNode n)
+  {
+    Assert(n.getKind() == kind::IS_SINGLETON);
+    return false; // n[0].isConst();
```
ajreynol, line 267:
```suggestion
    return false;
```

ajreynol (changes requested):

src/theory/sets/theory_sets_private.cpp
mudathirmahgoub: I guess I had to add `Assert(node.getKind() == CHOOSE` here. The `choose` operator is expanded here, so the rewriter does not see any node with `CHOOSE` operator.

src/theory/sets/theory_sets_type_rules.h
mudathirmahgoub: I why does not `n[0].const()` work here? `is_singleton (singleton 1)` is a constant. An assertion error was thrown when I used `n[0].const()`.

ajreynol: Actually you should just delete this entire method.  Applications of is_singleton are never constant.

This method returns true if the node n is a canonical representation of its type. This is clearly not the case for is_singleton: it is not the canonical representation of a Boolean.

src/theory/sets/theory_sets_private.cpp
ajreynol: This method should only be called when node has kind CHOOSE.  Moreover you probably should only call this on rewritten nodes.  You should ensure these invariants and move this rewrite to rewriter, which will restrict the cases in which you need to call this method.

mudathirmahgoub: Added Assertion for `CHOOSE` operator.

src/theory/sets/theory_sets_private.cpp
```diff
@@ -1399,15 +1399,16 @@ TrustNode TheorySetsPrivate::expandDefinition(Node node)
...
-TrustNode TheorySetsPrivate::expandChooseOperator(const Node& node)
+TrustNode TheorySetsPrivate::expandChooseOperator(Node node)
 {
   Assert(node.getKind() == CHOOSE);
 
-  if (node[0].getKind() == SINGLETON)
+  // we call the rewriter here to handle the pattern (choose (singleton x))
+  // because the rewriter is called after expansion
+  node = Rewriter::rewrite(node);
+  if(node.getKind() != CHOOSE)
```
ajreynol, line 1409: Make this `Node rewritten = Rewriter::rewrite(node);` for the below.

src/theory/sets/theory_sets_private.cpp
```diff
@@ -1436,31 +1437,32 @@ TrustNode TheorySetsPrivate::expandChooseOperator(const Node& node)
...
-  if (node[0].getKind() == SINGLETON)
+  // we call the rewriter here to handle the pattern
+  // (is_singleton (singleton x)) because the rewriter is called after expansion
+  node = Rewriter::rewrite(node);
+  if(node.getKind() != IS_SINGLETON)
   {
-    //(= (is_singleton (singleton x)) is a tautology
-    // we return true for (is_singleton (singleton x))
-    return TrustNode::mkTrustRewrite(node, d_true, nullptr);
+    return TrustNode::null();
```
ajreynol, line 1449: Same here.

test/regress/CMakeLists.txt
```diff
@@ -1761,6 +1761,7 @@ set(regress_1_tests
   regress1/sets/infinite-type/sets-card-int-1.smt2
   regress1/sets/infinite-type/sets-card-int-2.smt2
   regress1/sets/insert_invariant_37_2.smt2
+  regress1/sets/is_signleton1.smt2
```
ajreynol, line 1764: Benchmark is misspelled.

src/theory/sets/kinds
```diff
@@ -103,6 +107,7 @@ construle INSERT        ::CVC4::theory::sets::InsertTypeRule
 construle CARD          ::CVC4::theory::sets::CardTypeRule
 construle COMPLEMENT    ::CVC4::theory::sets::ComplementTypeRule
 construle CHOOSE        ::CVC4::theory::sets::ChooseTypeRule
+construle IS_SINGLETON  ::CVC4::theory::sets::IsSingletonTypeRule
```
ajreynol, line 110: Delete this.

src/theory/sets/theory_sets_type_rules.h
```diff
@@ -243,6 +243,31 @@ struct ChooseTypeRule
...
+    {
+      if (!setType.isSet())
+      {
+        throw TypeCheckingExceptionPrivate(
+            n, "IS_SINGLETON operator expects a set, a non-set is found");
+      }
+    }
+    return nodeManager->booleanType();
+  }
+  inline static bool computeIsConst(NodeManager* nodeManager, TNode n)
```
ajreynol, line 264: I believe you can delete this method after you delete the construle entry from the kinds file.  You can (and should) also delete all other computeIsConst that simply return false.

ajreynol (changes requested):
A few more minor requests, otherwise this looks good.

ajreynol (approved):
LGTM, thanks for the fixes.

PR #5037, opened 2020-09-07
Split term registry from theory state in sets

ajreynol:
Currently, the theory state object SolverState in sets sends lemmas and has a reference to the parent theory.  This PR is work towards eliminating these dependencies.

This adds a TermRegistry object which is responsible for some of the tasks currently done by SolverState, including all those involving lemmas. 

Notice this also makes a bug fix in getUnivSetEqClass where the universe set was being returned instead of its representative.

A followup PR will make SolverState maintain SAT-context-dependent membership lists which is also required for breaking the dependence on the parent theory.


src/theory/sets/solver_state.cpp
```diff
@@ -98,8 +99,6 @@ void SolverState::registerTerm(Node r, TypeNode tnn, Node n)
   {
     if (nk == SINGLETON)
     {
-      // singleton lemma
-      getProxy(n);
```
mudathirmahgoub, line 102: This call here to getProxy was generating a proxy for kinds: EMPTYSET, INTERSECTION, SETMINUS, UNION, UNIVERSE_SET, and SINGLETON. 
Now proxy are generated only for SINGLETON in TheorySetsPrivate::fullEffortCheck() line [364](https://github.com/CVC4/CVC4/pull/5037/files#diff-aaa774b6c62e6fbcec9f9c49f7d60a3eR346). Does this mean the other kinds do not need proxy variables?

ajreynol: This is only applied for SINGLETON (see guard directly above on line 99).  The behavior should be preserved, I believe.

src/theory/sets/theory_sets_private.cpp
```diff
@@ -332,17 +337,23 @@ void TheorySetsPrivate::fullEffortCheck()
             tnct = n;
           }
         }
-        // register it with the state
+        // register it with the term registry
```
mudathirmahgoub, line 340: `d_state.registerTerm(eqc, tnn, n);` below does not register any node with the term registry. 
The `StateSovler` has no knowledge of `TermRegistry`.

src/theory/sets/theory_sets_private.cpp
```diff
@@ -1441,7 +1449,11 @@ Node TheorySetsPrivate::getChooseFunction(const TypeNode& setType)
   return chooseSkolem;
 }
 
-void TheorySetsPrivate::presolve() { d_state.reset(); }
+void TheorySetsPrivate::presolve()
+{
+  d_state.reset();
+  d_state.reset();
```
mudathirmahgoub, line 1455: Duplicate line

src/theory/sets/term_registry.cpp
```diff
@@ -0,0 +1,155 @@
...
+using namespace CVC4::kind;
+
+namespace CVC4 {
+namespace theory {
+namespace sets {
+
+TermRegistry::TermRegistry(SolverState& state,
+                           InferenceManager& im,
+                           SkolemCache& skc)
+    : d_state(state),
```
mudathirmahgoub, line 29: The field `d_state` is never used. Unless `SolverState` would be needed in the future, it can be deleted.

src/theory/sets/solver_state.cpp
mudathirmahgoub: Lemmas are only generated for SINGLETON. But proxy variables are created for other kinds. 
See line https://github.com/CVC4/CVC4/blob/master/src/theory/sets/solver_state.cpp#L385

ajreynol: This PR preserves the previous behavior for only calling `getProxy` for SINGLETON kinds during registerTerm (line 343 of the modified theory_sets_private.cpp).

src/theory/sets/term_registry.cpp
ajreynol: Will do.

src/theory/sets/theory_sets_private.cpp
ajreynol: This is a typo, I'll revert this update to the comment.

mudathirmahgoub (approved):

PR #5038, opened 2020-09-07
Add buffered inference manager to TheorySep

ajreynol:
This makes TheorySep use a standard buffered inference manager.  Notice the behavior in TheorySep::doPending was simplified to assert both facts and lemmas. Previously, this was doing something strange: if there were any lemmas, then both facts and lemmas would be processed as lemmas, otherwise the facts would be processed as facts. Notice that TheorySep currently does not use facts by default.  This design can be addressed in the future.

Note this PR eliminates the need for a custom explain method in TheorySep.


src/theory/sep/theory_sep.cpp
```diff
@@ -943,12 +922,7 @@ bool TheorySep::needsCheckLastEffort() {
 
 void TheorySep::conflict(TNode a, TNode b) {
   Trace("sep-conflict") << "Sep::conflict : " << a << " " << b << std::endl;
-  Node eq = a.eqNode(b);
-  std::vector<TNode> assumptions;
-  explain(eq, assumptions);
-  Node conflictNode = mkAnd(assumptions);
-  d_state.notifyInConflict();
-  d_out->conflict( conflictNode );
+  d_im.conflictEqConstantMerge(a,b);
```
4tXJ7f, line 925: Apply ClangFormat.

4tXJ7f (approved):
LGTM

PR #5039, opened 2020-09-08
Eliminate a custom use of TheorySep in quantifiers engine for EPR

ajreynol:
This moves the initialization of the connection between separation logic and EPR to the separation logic theory, which is a more logical place for this.  This eliminates a backwards reference to theory engine from quantifiers engine.

### DIFF
--- a/src/theory/sep/theory_sep.cpp
+++ b/src/theory/sep/theory_sep.cpp
@@ -1002,8 +1002,8 @@ void TheorySep::ppNotifyAssertions(const std::vector<Node>& assertions) {
     }
   }
   // initialize the EPR utility
-  QuantifiersEngine * qe = getQuantifiersEngine();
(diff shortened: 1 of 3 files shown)


src/theory/sep/theory_sep.cpp
```diff
@@ -1001,6 +1001,22 @@ void TheorySep::ppNotifyAssertions(const std::vector<Node>& assertions) {
       d_loc_to_data_type[d_type_ref] = d_type_data;
     }
   }
+  // initialize the EPR utility
+  QuantifiersEngine * qe = getQuantifiersEngine();
+  if (qe!=nullptr)
```
mpreiner, line 1006: Looks like this is not clang-formatted.

mpreiner (approved):
LGTM, please clang-format changes.

PR #5042, opened 2020-09-08
Standard equality engine notify class for Theory

ajreynol:
This makes a standard equality engine notify class that forwards the 3 mandatory callbacks to the provided inference manager (the other 3 callbacks may be specific to the theory). It updates TheoryUF to use this class, other theories will be updated to this style as more inference manager are added.

Notice that we could also forward the other 3 callbacks in this class to `Theory`, making eqNotifyNewClass, eqNotifyMerge, and eqNotifyDisequal virtual methods in Theory, which can be done on a future PR if needed.


src/theory/uf/theory_uf.h
```diff
@@ -40,43 +41,20 @@ class HoExtension;
 class TheoryUF : public Theory {
 
 public:
-
-  class NotifyClass : public eq::EqualityEngineNotify {
-    TheoryUF& d_uf;
+ class NotifyClass : public TheoryEqNotifyClass
+ {
+   TheoryUF& d_uf;
```
mpreiner, line 46: This should be in the private section below the public section.

mpreiner (approved):

PR #5044, opened 2020-09-09
Fixes for regular expressions + sygus

ajreynol:
Includes not constructing a default value for non-first class types (e.g. RegLan) and a missing printer case for re.diff.


HanielB (approved):
LGTM

PR #5045, opened 2020-09-09
Move sets member propagation to SolverState

ajreynol:
This eliminates the parent relationship from solver state to theory sets.


src/theory/sets/solver_state.cpp
```diff
@@ -469,6 +467,126 @@ const vector<Node> SolverState::getSetsEqClasses(const TypeNode& t) const
...
+void SolverState::addMember(TNode r, TNode atom)
+{
+  NodeIntMap::iterator mem_i = d_members.find(r);
+  size_t n_members = 0;
+  if (mem_i != d_members.end())
+  {
+    n_members = (*mem_i).second;
+  }
+  d_members[r] = n_members + 1;
+  if (n_members < d_members_data[r].size())
```
mudathirmahgoub, line 501: When do we need this if statement? Isn't `n_numbers >= d_members_data[r].size` and invariant?

src/theory/sets/solver_state.cpp
```diff
@@ -469,6 +467,126 @@ const vector<Node> SolverState::getSetsEqClasses(const TypeNode& t) const
...
+        {
+          // conflict
+          Assert(facts.empty());
+          Trace("sets-prop")
+              << "Propagate eq-mem conflict : " << exp << std::endl;
+          facts.push_back(exp);
+          return false;
+        }
+      }
+      if (n_members < d_members_data[t1].size())
```
mudathirmahgoub, line 575: same here

src/theory/sets/theory_sets_private.cpp
```diff
@@ -126,71 +125,27 @@ void TheorySetsPrivate::eqNotifyMerge(TNode t1, TNode t2)
     }
     // merge membership list
     Trace("sets-prop-debug") << "Copying membership list..." << std::endl;
-    NodeIntMap::iterator mem_i2 = d_members.find(t2);
-    if (mem_i2 != d_members.end())
+    // if t1 has a singleton and s2 does not, we may have new inferences to
```
mudathirmahgoub, line 128: typo `if s1`

src/theory/sets/solver_state.cpp
```diff
@@ -469,6 +467,126 @@ const vector<Node> SolverState::getSetsEqClasses(const TypeNode& t) const
...
+      }
+    }
+    if (add)
+    {
+      // if there is a concrete set in t1, propagate new facts or conflicts
+      if (!cset.isNull())
+      {
+        NodeManager* nm = NodeManager::currentNM();
+        Assert(m2[1].getType().isComparableTo(cset.getType()));
+        Assert(areEqual(m2[1], cset));
```
mudathirmahgoub, line 552: It seems to me this assertion is stronger than the assertion in the previous line which can be removed.

mudathirmahgoub (approved):
Thanks for the removing setParent function. Please address the comments before merge.

src/theory/sets/solver_state.cpp
ajreynol: No, n_members is context dependent and d_members_data is not, this means that n_members can backtrack to a smaller value.

PR #5046, opened 2020-09-09
(proof-new) Generalize single step helper in eager proof generator

ajreynol:
This allows single step proofs to have premises, closed by a SCOPE.  This will be useful for array lemmas.


src/theory/eager_proof_generator.cpp
```diff
@@ -93,13 +94,24 @@ TrustNode EagerProofGenerator::mkTrustNode(Node n,
...
+  // if no children, its easy
+  if (exp.empty())
+  {
+    std::shared_ptr<ProofNode> pf = d_pnm->mkNode(id, {}, args, conc);
+    return mkTrustNode(conc, pf, isConflict);
+  }
+  // otherwise, we use CDProof + SCOPE
+  CDProof cdp(d_pnm);
+  cdp.addStep(conc, id, exp, args);
+  std::shared_ptr<ProofNode> pf = cdp.getProofFor(conc);
```
HanielB, line 112: why not do a `ProofNodeManager::mkNode` directly?

src/theory/eager_proof_generator.cpp
```diff
@@ -93,13 +94,24 @@ TrustNode EagerProofGenerator::mkTrustNode(Node n,
...
+  std::shared_ptr<ProofNode> pfs = d_pnm->mkNode(PfRule::SCOPE, {pf}, exp);
```
HanielB, line 113: why not use the safer `ProofNodeManager::mkScope`?

src/theory/eager_proof_generator.cpp
ajreynol: Because mkNode requires ProofNode, not Node. CDProof converts Node to ASSUME proof nodes.

src/theory/eager_proof_generator.cpp
ajreynol: For efficiency, this is a fixed use case that cannot go wrong and hence requires no checking.

src/theory/eager_proof_generator.cpp
HanielB: Ah, right, sorry.

src/theory/eager_proof_generator.cpp
HanielB: I see. Please add some documentation here to say why it can't go wrong.

HanielB (approved):

PR #5047, opened 2020-09-09
(proof-new) Add the arrays proof checker

ajreynol:
Includes adding the standard method for constructing the extensionality skolem using the skolem manager.


src/theory/arrays/skolem_cache.cpp
```diff
@@ -0,0 +1,84 @@
...
+  if (deq.hasAttribute(eiva))
+  {
+    return deq.getAttribute(eiva);
+  }
+  TypeNode atn = a.getType();
+  Assert(atn.isArray());
+  Assert(atn == b.getType());
+  TypeNode atnIndex = atn.getArrayIndexType();
+  Node v = NodeManager::currentNM()->mkBoundVar(atnIndex);
+  deq.setAttribute(eiva, v);
```
mpreiner, line 78: Is it guaranteed that `deq` is alive when we set the attribute? How is the node kept alive?

ajreynol: This is a good point.  So the eventual idea here is that a ProofNode will be generated whose premise is `deq` and stores a conclusion that involves the skolem that relies on this variable.  This ensures that `deq` is alive for as long as we care about this variable (if `deq` is destroyed we can just create a new variable).

At the moment, I believe the code will guarantee this but it can probably be done in a way that makes this clearer, I'll update the code.

mpreiner (approved):

PR #5050, opened 2020-09-10
Move finite model minimization to UF last call effort

### LINKED ISSUE #4850: (default) Fatal failure at smt/smt_engine.cpp:2602
```
[505] % cvc4 small.smt2
sat
[506] % cvc4 --check-models small.smt2
Fatal failure within void CVC4::SmtEngine::checkModel(bool) at smt/smt_engine.cpp:2602
assertion:   (not (fmf.card b 1))
simplifies to: false
expected `true'.
Aborted
[507] % cat small.smt2
(set-logic UFC)
(declare-sort a 0)
(declare-fun b () a)
(assert (not (fmf.card b 1)))
(check-sat)
[508] %
```

OS: Ubuntu
Commit: c324bb21

ajreynol:
This moves model minimization happen in TheoryUF's last call effort check instead of being a custom call in quantifiers finite model finding. This is both a better design and avoids bugs when quantifiers are not enabled (for `QF_UF`+cardinality constraints).

Fixes #4850.

### DIFF
--- a/src/theory/quantifiers/fmf/model_engine.cpp
+++ b/src/theory/quantifiers/fmf/model_engine.cpp
@@ -87,12 +87,13 @@ void ModelEngine::check(Theory::Effort e, QEffort quant_e)
     }
     Trace("model-engine-debug") << "Check model..." << std::endl;
     d_incomplete_check = false;
-    //print debug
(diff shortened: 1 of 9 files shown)


HanielB (approved):
LGTM

Issue #4889, opened 2020-08-13
Stack overflow in RemoveTermFormulas::run.


aniemetz:
The input below triggers a stack overflow in `RemoveTermFormulas::run`.

```
#include <cvc4/api/cvc4cpp.h>                                                   
                                                                                   
using namespace CVC4::api;                                                      
                                                                                   
int main()                                                                      
{                                                                               
  Solver slv;                                                                   
  Sort sort_int = slv.getIntegerSort();                                         
  Sort sort_array = slv.mkArraySort(sort_int, sort_int);                        
  Sort sort_fp128 = slv.mkFloatingPointSort(15, 113);                           
  Sort sort_fp32 = slv.mkFloatingPointSort(8, 24);                              
  Sort sort_bool = slv.getBooleanSort();                                        
  Term const0 = slv.mkConst(sort_fp32, "_c0");                                  
  Term const1 = slv.mkConst(sort_fp32, "_c2");                                  
  Term const2 = slv.mkConst(sort_bool, "_c4");                                                                                                                                                                           
  Term ite = slv.mkTerm(ITE, const2, const1, const0);                           
  Term rem = slv.mkTerm(FLOATINGPOINT_REM, ite, const1);                        
  Term isnan = slv.mkTerm(FLOATINGPOINT_ISNAN, rem);                            
  slv.checkSatAssuming(isnan);                                                  
}
```

CVC4 master 3f77b4ac0d4ff8ab69
`./configure debug --asan --symfpu`

Backtrace:
```
#0  0x00007ffff3547dba in CVC4::RemoveTermFormulas::run (this=<error reading variable: Cannot access memory at address 0x7fffff7fec80>, node=<error reading variable: Cannot access memory at address 0x7fffff7fec78>, 
    output=<error reading variable: Cannot access memory at address 0x7fffff7fec70>, newSkolems=<error reading variable: Cannot access memory at address 0x7fffff7fec68>, inQuant=false, inTerm=true)
    at /cvc4/src/smt/term_formula_removal.cpp:71
#1  0x00007ffff354cb64 in CVC4::RemoveTermFormulas::run (this=0x61900000ad70, node=..., output=std::vector of length 0, capacity 0, newSkolems=std::vector of length 0, capacity 0, inQuant=false, inTerm=true)
    at /cvc4/src/smt/term_formula_removal.cpp:334
[...]

PR #5052, opened 2020-09-10
Reduce recursion in term formula removal

ajreynol:
This reduces the use of recursion in term formula removal module.  The recursion depth is now limited to the number of term positions on a path where a formula must be removed, instead of being limited to the overall formula depth.  This PR also fixes some documentation.

Notice for its main cache `d_tfCache`, this class uses a `CDInsertHashMap` (instead of a `CDHashMap`) which does not allow inserting more than once, hence an auxliary "processedChildren" vector had to be introduced to track whether we have processed children. It's not clear to me whether we should just use the more standard CDHashMap, which would simplify this.

One non-linear arithmetic regression went unsat -> unknown, I added `--nl-ext-tplanes` to fix this.

This should fix https://github.com/CVC4/CVC4/issues/4889.  It is possible to further reduce the recursion in this pass, which can be done on a followup PR if needed.


src/smt/term_formula_removal.cpp
```diff
@@ -372,29 +461,8 @@ Node RemoveTermFormulas::run(TCtxStack& ctx,
...
-
-  // If changes, we rewrite
-  if(somethingChanged) {
-    Node cached = nodeManager->mkNode(node.getKind(), newChildren);
-    d_tfCache.insert(curr, cached);
-    return cached;
-  }
-  d_tfCache.insert(curr, Node::null());
-  return node;
+  // will recurse
```
aniemetz, line 464: As discussed offline, this comment is confusing.

aniemetz (changes requested):
As discussed offline, can you add the triggering example from #4889 to `test/api`?

aniemetz (approved):

PR #5054, opened 2020-09-10
(proof-new) Add SMT proof manager

ajreynol:
This module is responsible for constructing the proof checker, proof node manager, and implementing dumping and checking of the final proof.

This PR includes setting up some connections into the various modules for proof production.


src/options/smt_options.toml
```diff
@@ -173,6 +173,34 @@ header = "options/smt_options.h"
...
+  help = "Do not improve the granularity of proofs."
+[[option.mode.REWRITE]]
+  name = "rewrite"
+  help = "Expand macros, but do not expand rewrite or substitution steps."
+[[option.mode.THEORY_REWRITE]]
+  name = "theory-rewrite"
+  help = "Expand macros, rewrite and substitution steps, but do not expand theory rewrite steps."
+[[option.mode.DSL_REWRITE]]
+  name = "dsl-rewrite"
+  help = "Expand all steps to the level of DSL rewrites and evaluation steps."
```
HanielB, line 195: This is a bit misleading naming scheme, since `rewrite` is the mode where we *do not* expand rewrites. Wouldn't it be better to have something like
- off
- macros
- rewrites-and-subs
- theory-rewrites

src/smt/proof_manager.h
```diff
@@ -0,0 +1,103 @@
...
+   * from pg, postprocesses it, and stores it in d_finalProof.
+   */
+  void setFinalProof(ProofGenerator* pg, context::CDList<Node>* al);
+  /** The false node */
+  Node d_false;
+  /** For the new proofs module */
+  std::unique_ptr<ProofChecker> d_pchecker;
+  /** A proof node manager based on the above checker */
+  std::unique_ptr<ProofNodeManager> d_pnm;
+  /** The preprocess proof manager. */
```
HanielB, line 88:
```suggestion
  /** The preprocess proof generator. */
```

src/options/smt_options.toml
ajreynol: Then I think the option should be called `proof-elim=`, since granularity refers to the smallest unit that is kept.  I think I prefer granularity still, perhaps the option description should be updated to make this clearer?

HanielB: I see, I got confused indeed. Improving the doc might clarify.

HanielB (approved):

PR #5055, opened 2020-09-10
bv2int: simpler translation for plus and times

yoni206:
This PR makes the translation of plus and times by:
1. using plain `mod` rather than introducing a skolem. If this proves to be a bottleneck, we can try to tackle it where `mod` is treated in the solver in the future. This will make it easier to introduce support for quantifiers, as we don't need to introduce new variables under quantification.
2. Making sure everything is binarized in more places of the translation.


ahmed-irfan (approved):
LGTM

PR #5056, opened 2020-09-10
(proof-new) Update TheoryEngine lemma and conflict to TrustNode

ajreynol:
This updates the theory engine interfaces for conflicts and lemmas to be in terms of TrustNode not Node. 

This also updates the return value of getExplanation methods in TheoryEngine to TrustNode, but it does not yet add the proof generation code to that method yet, which will come in a separate PR.


HanielB (approved):
LGTM

PR #5059, opened 2020-09-12
(proof-new) Make proofs mandatory in proof equality engine

ajreynol:
All uses of proof equality engine are now guarded such that the ordinary equality engine is used when proofs are not enabled. Thus, we can make proofs mandatory in proof equality engine. This eliminates the need for some guards.

Some indentation changed, but there are no additions in this PR.


HanielB (approved):
LGTM

PR #5060, opened 2020-09-12
[proof-new] Adds Lazy CDProof chain data-structure

HanielB:
A proof generator to facilitate connection of locally independent but globally dependent proofs. In particular this will be used to model the resolution chains done in Minisat.


src/expr/lazy_proof_chain.cpp
```diff
@@ -0,0 +1,274 @@
...
+  Assert(expandedToConnect.find(cur) != expandedToConnect.end());
+  return expandedToConnect[cur];
+}
+
+void LazyCDProofChain::addLazyStep(Node expected, ProofGenerator* pg)
+{
+  Assert(pg != nullptr);
+  Trace("lazy-cdproofchain") << "LazyCDProofChain::addLazyStep: " << expected
+                             << " set to generator " << pg->identify() << "\n";
+  // note this will rewrite the generator for expected, if any
```
ajreynol, line 193:
```suggestion
  // note this will replace the generator for expected, if any
```

src/expr/lazy_proof_chain.h
```diff
@@ -0,0 +1,125 @@
...
+  /** Constructor
+   *
+   * @param pnm The proof node manager for constructing ProofNode objects.
+   * @param c The context that this class depends on. If none is provided,
+   * this class is context-independent.
+   */
+  LazyCDProofChain(ProofNodeManager* pnm, context::Context* c = nullptr);
+  ~LazyCDProofChain();
+  /**
+   * Get lazDyet have a concrete proof and that are registered in the chains
```
ajreynol, line 50: Typo?

src/expr/lazy_proof_chain.h
```diff
@@ -0,0 +1,125 @@
...
+   */
+  void addLazyStep(Node expected,
+                   ProofGenerator* pg,
+                   const std::vector<Node>& assumptions,
+                   const char* ctx = "LazyCDProofChain::addLazyStep");
+
+  /** As above but does not do the closedness check. */
+  void addLazyStep(Node expected, ProofGenerator* pg);
+
+  /** Does the given fact have an explicitly provided generator? */
```
ajreynol, line 101:
```suggestion
  /** Does the given fact have an explicitly provided generator, modulo symmetry of (dis)equality? */
```

src/expr/lazy_proof_chain.cpp
```diff
@@ -0,0 +1,274 @@
...
+  {
+    cur = visit.back();
+    visit.pop_back();
+    const auto itVisited = visited.find(cur);
+    // pre-traversal
+    if (itVisited == visited.end())
+    {
+      Trace("lazy-cdproofchain")
+          << "LazyCDProofChain::getProofFor: check " << cur << "\n";
+      if (!hasGenerator(cur))
```
ajreynol, line 52: Let's avoid 2 map lookups by calling `getGeneratorFor` and having this case check if `nullptr`.

src/expr/lazy_proof_chain.cpp
```diff
@@ -0,0 +1,274 @@
...
+      // enqueue free assumptions to process
+      for (const std::pair<const Node, std::vector<ProofNode*>>& fap : famap)
+      {
+        // check cycles, which are cases in which we have facts in
+        // expandedToConnect but not already expanded
+        auto it = expandedToConnect.find(fap.first);
+        if (it != expandedToConnect.end() && !expanded.count(fap.first))
+        {
+          // mark as a not to expand assumption. We do this rather than just
+          // assining visited[fap.first] to a non-null value so that we properly
```
ajreynol, line 104:
```suggestion
          // assigning visited[fap.first] to a non-null value so that we properly
```

src/expr/lazy_proof_chain.cpp
```diff
@@ -0,0 +1,274 @@
...
+      visit.push_back(cur);
+      // enqueue free assumptions to process
+      for (const std::pair<const Node, std::vector<ProofNode*>>& fap : famap)
+      {
+        // check cycles, which are cases in which we have facts in
+        // expandedToConnect but not already expanded
+        auto it = expandedToConnect.find(fap.first);
+        if (it != expandedToConnect.end() && !expanded.count(fap.first))
+        {
+          // mark as a not to expand assumption. We do this rather than just
```
ajreynol, line 103: "mark as a not"?

src/expr/lazy_proof_chain.cpp
```diff
@@ -0,0 +1,274 @@
...
+namespace CVC4 {
+
+LazyCDProofChain::LazyCDProofChain(ProofNodeManager* pnm, context::Context* c)
+    : d_manager(pnm), d_context(), d_gens(c ? c : &d_context)
+{
+}
+
+LazyCDProofChain::~LazyCDProofChain() {}
+
+std::shared_ptr<ProofNode> LazyCDProofChain::getProofFor(Node fact)
```
ajreynol, line 30: I don't see why we need a post-order traversal in this method.  This seems unnecessary and suboptimal.  Shouldn't we just expand and get its free assumptions?  The future calls should do the right thing.

ajreynol (changes requested):
Generally the design of the class looks good to me, I think I agree it should be robust to symmetry of (dis)equality, although this is in theory optional. Another thing to note is that this class could have been another proof generation policy for LazyCDProof, since the two are both context-dependent maps from formulas to proof generators (minus the fact that LazyCDProof inherits from CDProof), although I think I'm fine with having them separate.

My main concern is with `getProofFor`, which I don't understand why it needs a postorder traversal.

src/expr/lazy_proof_chain.h
```diff
@@ -0,0 +1,135 @@
...
+   */
+  ProofGenerator* getGeneratorFor(Node fact);
+
+  /** identify */
+  std::string identify() const override;
+
+ private:
+  /** The proof manager, used for allocating new ProofNode objects */
+  ProofNodeManager* d_manager;
+  /** Whether this istance is robust to cycles in the chain. */
```
ajreynol, line 125:
```suggestion
  /** Whether this instance is robust to cycles in the chain. */
```

ajreynol (approved):
One minor change.

PR #5062, opened 2020-09-14
Fix type for Windows build

4tXJ7f:
The `BVToInt` preprocessing pass was using `uint`, which appears to be
undefined when we cross-compile for Windows. This commit fixes the issue
by using `size_t`.


nafur (approved):
LGTM.

PR #5065, opened 2020-09-14
Add simple BV solver

mpreiner:
This PR adds a simple BV solver that sends bit-blasting lemmas to the internal MiniSat.


src/smt/set_defaults.cpp
```diff
@@ -125,6 +125,14 @@ void setDefaults(LogicInfo& logic, bool isInternalSubsolver)
     }
   }
 
+  /* BVSolver::SIMPLE does not natively support int2bv and nat2bv and need to
```
aniemetz, line 128:
```suggestion
  /* BVSolver::SIMPLE does not natively support int2bv and nat2bv, they need to
```

src/theory/bv/bv_solver_simple.cpp
```diff
@@ -0,0 +1,283 @@
...
+  void bbAtom(TNode node) override;
+  Node getBBAtom(TNode atom) const override;
+  void storeBBAtom(TNode atom, Node atom_bb) override;
+  void storeBBTerm(TNode node, const Bits& bits) override;
+  bool hasBBAtom(TNode atom) const override;
+  Node getStoredBBAtom(TNode node);
+
+  void makeVariable(TNode var, Bits& bits) override;
+
+  bool collectModelValues(TheoryModel* m, const std::set<Node>& relevantTerms);
```
aniemetz, line 46: Document.

src/theory/bv/bv_solver_simple.cpp
```diff
@@ -0,0 +1,283 @@
...
+
+  prop::SatSolver* getSatSolver() override { return nullptr; }
+
+ private:
+  Node getModelFromSatSolver(TNode a, bool fullModel) override;
+
+  TNodeSet d_variables;
```
aniemetz, line 53: Document members.

src/theory/bv/theory_bv_rewriter.h
```diff
@@ -50,8 +50,9 @@ class TheoryBVRewriter : public TheoryRewriter
 
  private:
   static RewriteResponse IdentityRewrite(TNode node, bool prerewrite = false);
-  static RewriteResponse UndefinedRewrite(TNode node, bool prerewrite = false); 
-  
+  static RewriteResponse UndefinedRewrite(TNode node, bool prerewrite = false);
```
aniemetz, line 53: What's this for? What is it used for?

src/theory/bv/bv_solver_simple.cpp
```diff
@@ -0,0 +1,283 @@
...
+#include "theory/bv/bitblast/lazy_bitblaster.h"
+#include "theory/bv/theory_bv.h"
+#include "theory/bv/theory_bv_utils.h"
+#include "theory/theory_model.h"
+
+namespace CVC4 {
+namespace theory {
+namespace bv {
+
+class BBSimple : public TBitblaster<Node>
```
ajreynol, line 28: Can you add some documentation here above this class.

src/theory/bv/bv_solver_simple.h
```diff
@@ -0,0 +1,70 @@
...
+#include "theory/bv/bv_solver.h"
+
+namespace CVC4 {
+
+namespace theory {
+namespace bv {
+
+class BBSimple;
+
+class BVSolverSimple : public BVSolver
```
ajreynol, line 33: Add documentation (one paragraph or so) describing the main features of the solver.

src/theory/bv/bv_solver_simple.h
```diff
@@ -0,0 +1,70 @@
...
+                                  SubstitutionMap& outSubstitutions) override
+  {
+    return Theory::PPAssertStatus::PP_ASSERT_STATUS_UNSOLVED;
+  }
+
+  bool collectModelValues(TheoryModel* m,
+                          const std::set<Node>& termSet) override;
+
+ private:
+  std::unique_ptr<BBSimple> d_bitblaster;
```
ajreynol, line 59: Document.

src/theory/bv/bv_solver_simple.h
```diff
@@ -0,0 +1,70 @@
...
+  /** Caches lemmas added for BV atoms. */
+  std::unordered_set<Node, NodeHashFunction> d_lemmasCache;
+
+  void addBBLemma(TNode fact);
```
ajreynol, line 63: Document.

ajreynol (changes requested):
Some documentation requests

src/theory/bv/theory_bv_rewriter.h
mpreiner: That's unused and was probably used to initialize `d_rewriteTable` to indicate that a rewrite for a kind was not implemented. I can remove it if you want.

src/theory/bv/bv_solver_simple.cpp
```diff
@@ -0,0 +1,299 @@
...
+{
+}
+
+void BVSolverSimple::addBBLemma(TNode fact)
+{
+  if (!d_bitblaster->hasBBAtom(fact))
+  {
+    d_bitblaster->bbAtom(fact);
+  }
+  if (d_lemmasCache.find(fact) == d_lemmasCache.end())
```
ajreynol, line 251: This lemmas cache appears to be unused?  Regardless I don't think you want to use this: lemmas should be cached based on the user context, moreover lemmas are actually cached by default in the call to TheoryInferenceManager::lemma below.

ajreynol (changes requested):
One minor request, you can remove the `d_lemmaCache` field I believe, otherwise this looks good to me.

src/theory/bv/bv_solver_simple.cpp
mpreiner: done

ajreynol (approved):

PR #5067, opened 2020-09-14
Update copyright header script to support CMake and Python files

mpreiner:
This PR updates the update-copyright.pl script to also update/add copyright headers to CMake specific files. It further fixes a small typo in the header.


4tXJ7f (changes requested):
As discussed offline, it would be great if we could include Python files as well for consistency.

aniemetz (approved):

4tXJ7f (approved):
LGTM

PR #5069, opened 2020-09-16
[proof-new] A simple proof generator for buffered proof steps

HanielB:
This commit also modifies proof equality engine to use this new proof generator rather than the FactProofGenerator, on which this new one is based.


ajreynol (approved):
LGTM, let's merge https://github.com/CVC4/CVC4/pull/5059 before updating proof equality engine to use this.

PR #5073, opened 2020-09-16
Logic exception for arrays indexed by arrays

### LINKED ISSUE #4237: Soundness bugs in arrays + reals/ints (no option)
Hi,
For this formula:
[137.txt](https://github.com/CVC4/CVC4/files/4454419/137.txt)

CVC4 throws out a fatal failure:
~~~~
Fatal failure within void CVC4::theory::arrays::ArrayInfo::addIndex(CVC4::Node, CVC4::TNode) at /home/CVC4/src/theory/arrays/array_info.cpp:137
Check failure

 !i.getType().isArray()
~~~~

OS: Ubuntu 16.04
Commit: 2f8caabd

ajreynol:
This throws a logic exception when a term of array type whose index is also an array is registered to the theory of arrays.

It refactors the preRegisterTermInternal method of arrays so that all non-equality terms are added to the equality engine in the same block of code, which also checks for the type.

Fixes #4237.

FYI @barrettcw


nafur (approved):
LGTM.

PR #5075, opened 2020-09-16
Only rewrite replace_re(_all) if regexp is const

### LINKED ISSUE #5074: Assertion failure with Regexp.
CVC4 fails with
```
Fatal failure within static bool CVC4::theory::strings::RegExpEntail::testConstStringInRegExp(CVC4::String&, un
signed int, CVC4::TNode) at cvc4/src/theory/strings/regexp_entail.cpp:364
Check failure

 false
RegExp contains variables
Aborted (core dumped)

```

for

```
  #include <assert.h>                                                             
  #include "cvc4/api/cvc4cpp.h"                                                   
                                                                                  
  using namespace CVC4::api;                                                      
                                                                                  
  int main()                                                                      
  {                                                                               
    Solver slv;                                                                   
    Term c1 = slv.mkConst(slv.getIntegerSort());                                  
    Term t6 = slv.mkTerm(Kind::STRING_FROM_CODE, c1);                             
    Term t12 = slv.mkTerm(Kind::STRING_TO_REGEXP, t6);                            
    Term t14 = slv.mkTerm(Kind::STRING_REPLACE_RE, {t6, t12, t6});                                                                                                                                                             
    Term t16 = slv.mkTerm(Kind::STRING_CONTAINS, {t14, t14});                     
    slv.checkEntailed(t16);                                                       
                                                                                  
    return 0;                                                                     
  } 
```

CVC4 master 9e2a36f53007b932412a98c8e7ff1556a53f37c5
debug build

4tXJ7f:
Fixes #5074. This commit fixes an issue in the rewriter of
`str.replace_re` and `str.replace_re_all`. For both operators, we do two
kinds of rewrites: (1) If the first argument is a constant, then we
check whether the regular expression appears in that constant and (2) we
check whether the regular expression matches the empty string. Both of
those checks were assuming a constant regular expression but there was
no guard in place for it.  This commit adds the missing guard.

### DIFF
--- a/src/theory/strings/sequences_rewriter.cpp
+++ b/src/theory/strings/sequences_rewriter.cpp
@@ -3021,30 +3021,33 @@ Node SequencesRewriter::rewriteReplaceRe(Node node)
   Node y = node[1];
   Node z = node[2];
 
-  if (x.isConst())
+  if (RegExpEntail::isConstRegExp(y))
   {
-    // str.replace_re("ZABCZ", re.++("A", _*, "C"), y) ---> "Z" ++ y ++ "Z"
-    std::pair<size_t, size_t> match = firstMatch(x, y);
-    if (match.first != string::npos)
+    if (x.isConst())
     {
-      String s = x.getConst<String>();
-      Node ret = nm->mkNode(STRING_CONCAT,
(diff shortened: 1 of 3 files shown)


ajreynol (approved):
LGTM

PR #5076, opened 2020-09-16
Further standardization of datatypes

ajreynol:
We now have no custom calls to equality engine explain, and only 2 manual calls to equality engine (in its entailment check).  This also updates the notify class to the standard one.

This PR makes datatypes ready to start work on proofs.


4tXJ7f (approved):
LGTM

PR #5077, opened 2020-09-16
Dumping internal define-funs with no arguments

yoni206:
Currently, `--dump=assertions:...` fails if a `define-fun` command is executed internally (an example for this can be found [here](https://github.com/CVC4/CVC4/blob/bcaebfa163bb27e1cf14c0f763afb47b185a5f99/src/preprocessing/passes/real_to_int.cpp#L177).

The failure occurs because the printer expects a range type whenever a `define-fun` command is executed. However, when this command is used for 0-ary functions (variables), the getter for the range fails.

This PR fixes this by asking for a range only if the type is indeed a function. Otherwise, the original type is printed.

A test that currently fails but passes with this PR is included.

### DIFF
--- a/src/smt/node_command.cpp
+++ b/src/smt/node_command.cpp
@@ -166,7 +166,7 @@ void DefineFunctionNodeCommand::toStream(std::ostream& out,
 {
   TypeNode tn = d_fun.getType();
   bool hasRange =
-      ((tn.isFunction()) || (tn.isConstructor()) || (tn.isSelector()));
+      (tn.isFunction() || tn.isConstructor() || tn.isSelector());
   Printer::getPrinter(language)->toStreamCmdDefineFunction(
       out,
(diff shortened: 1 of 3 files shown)


src/smt/node_command.cpp
```diff
@@ -164,11 +164,14 @@ void DefineFunctionNodeCommand::toStream(std::ostream& out,
                                          size_t dag,
                                          OutputLanguage language) const
 {
+  TypeNode tn = d_fun.getType();
+  bool hasRange =
+      ((tn.isFunction()) || (tn.isConstructor()) || (tn.isSelector()));
```
ajreynol, line 169:
```suggestion
      (tn.isFunction() || tn.isConstructor() || tn.isSelector());
```

ajreynol (approved):
LGTM, one minor suggestion.

PR #5080, opened 2020-09-17
bv2int: quantifiers support

yoni206:
This PR adds quantifiers support for bv2int preprocessing pass, and adds regressions that use quantifiers.


ajreynol (changes requested):
One request for simplification.

src/preprocessing/passes/bv_to_int.cpp
```diff
@@ -645,6 +645,21 @@ Node BVToInt::translateWithChildren(Node original,
...
+    {
+      returnNode = d_nm->mkNode(oldKind, translated_children);
+      break;
+    }
+    case kind::FORALL:
+    {
+      returnNode = translateQuantifiedFormula(original);
+      break;
+    }
+    case kind::EXISTS:
```
ajreynol, line 658: You can assume that EXISTS is eliminated at this point (it is eliminated in the rewriter).

ajreynol (approved):
LGTM

PR #5083, opened 2020-09-17
Add method to get Python object from constant value term in Python API

makaimann:
This PR addresses issue https://github.com/CVC4/CVC4/issues/5014. It simply interprets the SMT-LIB string representation and produces a Python object. It currently supports booleans, ints, reals, bit-vectors, and arrays. The method (`toPythonObj`) is only valid to call if `isConst` returns true.


src/api/python/cvc4.pxi
```diff
@@ -1503,6 +1504,98 @@ cdef class Term:
...
+            is_neg = False
+            updated_string_repr = string_repr
+            if string_repr[:1] == '(-':
+                is_neg = True
+                updated_string_repr = string_repr[2:-1]
+
+            try:
+                # expecting format (/ a b)
+                splits = updated_string_repr.split()
+                assert len(splits) == 3
```
mpreiner, line 1557: If you expect above format you could also use this: `s.strip('()/').split()` which directly gets you`['a', 'b']` - would also save the den[-1] check below.

src/api/python/cvc4.pxi
```diff
@@ -1503,6 +1504,98 @@ cdef class Term:
...
+        res = None
+        if sort.isBoolean():
+            if string_repr == "true":
+                res = True
+            else:
+                assert string_repr == "false"
+                res = False
+        elif sort.isInteger():
+            is_neg = False
+            updated_string_repr = string_repr
```
mpreiner, line 1536: I think this could be simplified: `updated_string_repr = string_repr.strip('()') .replace(' ', '')`and `res = int(updated_string_repr)`. You shouldn't need the additional sign check.

src/api/python/cvc4.pxi
```diff
@@ -1503,6 +1504,98 @@ cdef class Term:
...
+        elif sort.isBitVector():
+            # expecting format #b<bits>
+            assert string_repr[:2] == "#b"
+            python_bin_repr = "0" + string_repr[1:]
+            try:
+                res = int(python_bin_repr, 2)
+            except:
+                raise ValueError("Failed to convert bitvector {} to an int".format(string_repr))
+        elif sort.isArray():
+            res = {}
```
mpreiner, line 1578: You could maybe use a defaultdict here so that the dict returns by default the base case instead of the `*` entry.

src/api/python/cvc4.pxi
makaimann: Good suggestion, thanks! Addressed in 3ba57ed7b00f65e1c1263934a07a90f62e67e645

src/api/python/cvc4.pxi
makaimann: Good point -- addressed in 3ba57ed7b00f65e1c1263934a07a90f62e67e645

src/api/python/cvc4.pxi
makaimann: That's very clever, I like it! Added in 3ba57ed7b00f65e1c1263934a07a90f62e67e645.

mpreiner (approved):
LGTM, please make sure that you have 80 char lines.

makaimann:
Thanks for the reminder! Just fixed that with the latest commit.

PR #5084, opened 2020-09-17
Fix assertion list for globally defined recursive functions

ajreynol:
This was uncovered by a (spurious) proof checking failure on proof-new.


4tXJ7f (approved):
LGTM

PR #5085, opened 2020-09-17
(cad-solver) Fix square-free-basis computation

nafur:
This PR fixes a subtle issue when making the polynomials of two subsequent (overlapping) intervals relative square-free. We now make sure that the resulting polynomials are ignored (if constant) or pushed to the lower dimension (if lower main variable). Also, we now appropriately update the main polynomials of the corresponding intervals.


ajreynol (approved):
LGTM

PR #5086, opened 2020-09-17
[proof-new] Have mkScope agnostic to True assumptions

HanielB:


src/expr/proof_node_manager.cpp
```diff
@@ -22,7 +22,10 @@ using namespace CVC4::kind;
 
 namespace CVC4 {
 
-ProofNodeManager::ProofNodeManager(ProofChecker* pc) : d_checker(pc) {}
+ProofNodeManager::ProofNodeManager(ProofChecker* pc)
+    : d_checker(pc), d_true(NodeManager::currentNM()->mkConst<bool>(true))
```
ajreynol, line 26:
```suggestion
    : d_checker(pc), d_true(NodeManager::currentNM()->mkConst(true))
```

ajreynol: Also I usually put this in the body of the constructor.

ajreynol (approved):
One minor request.

src/expr/proof_node_manager.cpp
HanielB: Any reason to put it in the body?

ajreynol: looks weird calling NodeManager::currentNM() in an initializer list.

HanielB: I see. I've changed it.

PR #5087, opened 2020-09-17
(proof-new) Fixes for theory engine proof generator

ajreynol:
Fixes an issue where conflicts stored in the theory engine proof generator where expecting to be (=> F false) instead of (not F).


HanielB (approved):
LGTM

PR #5088, opened 2020-09-17
(proof-new) Updates to proof node updater algorithm

ajreynol:
This updates the proof node updater so that we apply updates to a fixed point, stopping if the proof node is not updated or if the callback explicitly tells us not to continue.  This also fixes an issue where proof nodes would be updated to SCOPE and incorrectly traversed after updating.

This additionally adds debugging feature to proof node updater to track the moment at which an unexpected free assumption is introduced by an update.


HanielB (approved):
LGTM. The only thing is that `continueUpdate` blocks not unly recursively updating but also updating the node itself until a fix-point. I think it'd be good to clarify this in the different places in which `continueUpdate` is explained.

PR #5093, opened 2020-09-18
Fix muzzled builds

4tXJ7f:
Commit 2c2f05c96e021006275a2bc70b9ede70b280616d moved some function
definitions from `dump.h` to `dump.cpp`, which is good. However, the
corresponding definitions for muzzled builds weren't moved, so muzzled builds
defined the operator `<<` multiple times. This made our nightly
competition build fail. This commit fixes the issue by moving the
alternative definitions to the source file as well.


src/smt/dump.cpp
```diff
@@ -42,6 +44,16 @@ CVC4dumpstream& CVC4dumpstream::operator<<(const NodeCommand& nc)
...
 
+#else
+
+CVC4dumpstream& CVC4dumpstream::operator<<(const Command& c) { return *this; }
+CVC4dumpstream& CVC4dumpstream::operator<<(const NodeCommand& nc)
+{
+  return *this;
+}
+
+#endif
```
abdoo8080, line 55: Nit: Since preprocessing directives are not indented, it gets hard to track which `#endif` belongs to which `#if`. So, I find it helpful to mention the relevant condition next to `#endif`.
```suggestion
#endif /* CVC4_DUMPING && !CVC4_MUZZLE */
```

abdoo8080 (approved):
Thanks for catching my error! This is my very first review. So, please bear with my ignorance of the guidelines and the steps that I need to take 🙂.

4tXJ7f:
All good :) Thanks for the suggestion. I've committed it.

PR #5096, opened 2020-09-18
bv2int: new options for bvand translation

yoni206:
Currently, `(bvand x y)` is translated into a sum of ITEs. 
This PR introduces two more options for the translation of `(bvand x y)`:
1. bv: cast the integer translations of `x` and `y` back to bit-vectors, do a `bvand`, and cast the result to integers.
2. iand: use the builtin `iand` operator.

These options are added to many of the tests, and some new tests are added.
In addition, some tests are cleaned up (e.g., removing --no-check-unsat-cores for satisfiable benchmarks).
Finally, some tests are moved from regress0 to regress2 because they take several seconds to complete (2 -- 10 seconds).


test/regress/regress3/bv_to_int_check_ne_bvlshr0_4bit.smt2
```diff
@@ -0,0 +1,19 @@
+; COMMAND-LINE:  --solve-bv-as-int=bv 
+(set-logic BV)
```
ajreynol, line 2: Seems to be missing a status.

ajreynol: Note that regress3 is not checked by CI, so you should maybe ensure `make regress3` works locally.

ajreynol (changes requested):

test/regress/regress3/bv_to_int_check_ne_bvlshr0_4bit.smt2
yoni206: this file was actually redundant and did not appear in CMakeLists. fixed now. `ctest -R bv_to_int` passes locally.

ajreynol (approved):
LGTM

PR #5097, opened 2020-09-18
(proof-new) Add the term conversion sequence generator

ajreynol:
This class will be used in the theory preprocessor and is planned to be used in other preprocessing passes that can be understood as sequences of rewrite systems.


src/expr/proof_node_manager.h
```diff
@@ -80,6 +80,14 @@ class ProofNodeManager
    * @return The ASSUME proof of fact.
    */
   std::shared_ptr<ProofNode> mkAssume(Node fact);
+  /**
+   * Make transitivity proof, where children contains one or more proofs of
+   * equalities that form a chain.
```
HanielB, line 85: Say that chain is correctly ordered

src/expr/tconv_seq_proof_generator.cpp
```diff
@@ -0,0 +1,163 @@
+/*********************                                                        */
+/*! \file tconv_seq_proof_generator.h
+ ** \verbatim
+ ** Top contributors (to current version):
+ **   Andrew Reynolds
+ ** This file is part of the CVC4 project.
+ ** Copyright (c) 2009-2020 by the authors listed in the file AUTHORS
+ ** in the top-level source directory) and their institutional affiliations.
```
HanielB, line 8: I'm not sure what happened but I have noticed a lot of this open parenthesis in headers. Delete it.

src/expr/tconv_seq_proof_generator.h
```diff
@@ -0,0 +1,109 @@
...
+   * was called. Otherwise t_{i+1} = t_i.
+   * In other words, t_n is obtained by converting t_0, in order, based on the
+   * calls to registerConvertedTerm.
+   *
+   * @param f The equality fact to get the proof for.
+   * @return The proof for f.
+   */
+  std::shared_ptr<ProofNode> getProofFor(Node f) override;
+  /**
+   * Get subsequence proof, with given start and end steps (inclusive).
```
HanielB, line 71: Clarify who is f and what subsequence proof is this supposed to be.

src/expr/tconv_seq_proof_generator.h
```diff
@@ -0,0 +1,109 @@
...
+
+  /**
+   * Make trust node from a sequence of converted terms. The number of
+   * terms in cterms should be 1 + the number of component proof generators
+   * maintained by this class. This selects a proof generator that is capable
+   * of proving cterms[0] = cterms[cterms.size()-1], which is either this
+   * generator, or one of the component proof generators, if only one step
+   * rewrote. In the former case, all steps are registered to this class.
+   * Using a component generator is an optimization that saves having to
+   * save the conversion steps or use this class.
```
HanielB, line 87: Not easy to follow. Maybe give an example to clarify.

src/expr/tconv_seq_proof_generator.cpp
```diff
@@ -0,0 +1,163 @@
...
+  std::vector<std::shared_ptr<ProofNode>> transChildren;
+  std::pair<Node, size_t> currKey;
+  NodeIndexNodeMap::iterator itc;
+  // convert the term in sequence
+  for (size_t i = start; i <= end; i++)
+  {
+    currKey = std::pair<Node, size_t>(curr, i);
+    itc = d_converted.find(currKey);
+    // if we provided a conversion at this index
+    if (itc != d_converted.end())
```
HanielB, line 78: Say when it could be the case that we did not provide such a conversion

src/expr/tconv_seq_proof_generator.cpp
```diff
@@ -0,0 +1,163 @@
...
+  ProofGenerator* pg = nullptr;
+  for (size_t i = 0, nconvs = d_tconvs.size(); i < nconvs; i++)
+  {
+    if (cterms[i] == cterms[i + 1])
+    {
+      continue;
+    }
+    else if (pg == nullptr)
+    {
+      // maybe the i^th generator can explain it alone
```
HanielB, line 138: what if it can't?

src/expr/tconv_seq_proof_generator.cpp
```diff
@@ -0,0 +1,163 @@
...
+    }
+  }
+  if (useThis)
+  {
+    pg = this;
+    // if more than two steps, we must register each conversion step
+    for (size_t i = 0, nconvs = d_tconvs.size(); i < nconvs; i++)
+    {
+      registerConvertedTerm(cterms[i], cterms[i + 1], i);
+    }
```
HanielB, line 154: this can introduce empty points in the sequence?

src/expr/tconv_seq_proof_generator.cpp
ajreynol: This is leftover from a bug in the old copyright header.

src/expr/tconv_seq_proof_generator.cpp
ajreynol: yes but this is guarded in registerConvertedTerm

src/expr/tconv_seq_proof_generator.cpp
```diff
@@ -135,11 +135,13 @@ theory::TrustNode TConvSeqProofGenerator::mkTrustRewriteSequence(
     }
     else if (pg == nullptr)
     {
-      // maybe the i^th generator can explain it alone
+      // maybe the i^th generator can explain it alone. We may overwrite pg
+      // with this class if another step is encountered in this loop.
```
HanielB, line 139: I think it'd be good to say explicitly that unless there is another generator in the loop than this i^th generator *must* be able to explain it alone. This is what was a bit confusing to me.

HanielB (approved):
Thanks for the changes. Just had that one final comment.

PR #5098, opened 2020-09-18
Standardize equality engine notifications in sets

ajreynol:


src/theory/sets/theory_sets.cpp
```diff
@@ -201,37 +201,6 @@ bool TheorySets::isEntailed( Node n, bool pol ) {
...
-bool TheorySets::NotifyClass::eqNotifyTriggerTermEquality(TheoryId tag,
-                                                          TNode t1,
-                                                          TNode t2,
-                                                          bool value)
-{
-  Debug("sets-eq") << "[sets-eq] eqNotifyTriggerTermEquality: tag = " << tag
-                   << " t1 = " << t1 << "  t2 = " << t2 << "  value = " << value
-                   << std::endl;
-  d_theory.propagate(value ? t1.eqNode(t2) : t1.eqNode(t2).negate());
-  return true;
```
mudathirmahgoub, line 225: Just to satisfy my heart, it seems this function always returns true, however, the new one `TheoryEqNotifyClass::eqNotifyTriggerTermEquality` returns false when there is a conflict. 
Was it the case that sets solver never reaches this line in a state of a conflict? or the code was not correct, and you actually fixed a hidden bug :)

src/theory/sets/theory_sets_private.cpp
```diff
@@ -1234,44 +1234,8 @@ Node mkAnd(const std::vector<TNode>& conjunctions)
...
-OutputChannel* TheorySetsPrivate::getOutputChannel()
-{
-  return d_external.d_out;
-}
-
 Valuation& TheorySetsPrivate::getValuation() { return d_external.d_valuation; }
 
-void TheorySetsPrivate::conflict(TNode a, TNode b)
-{
-  Node conf = explain(a.eqNode(b));
```
mudathirmahgoub, line 1268: The new code uses `TheoryInferenceManager::explainConflictEqConstantMerge` which depends on `TheoryInferenceManager::explainConflictEqConstantMerge`. On the other hand, `TheorySetsPrivate::explain` depends on `EqualityEngine::explainEquality`. Although both `explain` the conflict, their code is different. Are they the same?

src/theory/sets/theory_sets.cpp
ajreynol: Yes, I'm surprised the code was initially like this.

I believe this does fix a hidden bug, albeit probably a minor one.  The return value is simply for telling some of the internals in the sets solver (e.g. in its equality engine) that it should terminate because we are going to backtrack.

src/theory/sets/theory_sets_private.cpp
ajreynol: Yes, in the end sets is making custom calls to explain equalities from equality engine, but the code should be the same.  The variants of TheorySetsPrivate::explain that do not call to theory inference manager will be deleted soon.

mudathirmahgoub (approved):
Thanks for refactoring redundant functions in TheorySets::NotifyClass.

PR #5103, opened 2020-09-20
Connect the shared solver to theory engine

ajreynol:
This makes SharedSolver the main communication point for TheoryEngine during solving for combination-related solving tasks.  This class is a generalization of `SharedTermsDatabase`, and in the distributed architecture is a wrapper around shared terms database.

It has 5 callbacks in theory engine: for preregistration, preNotifyFact (which calls addSharedTerms on theories), assertSharedEquality, explain, getEqualityStatus.

This PR has no intended behavior changes.

FYI @barrettcw


src/theory/ee_manager.h
```diff
@@ -60,6 +62,9 @@ class EqEngineManager
    *
    * This method is context-independent, and is applied once during
    * the lifetime of TheoryEngine (during finishInit).
+   *
+   * @param sharedSolver The shared solver that is being used in combination
```
4tXJ7f, line 66: This part of the comment should be moved up to the constructor.

4tXJ7f (approved):
Cool, just one minor comment about a comment :)

PR #5105, opened 2020-09-21
Refactor Commands to use the Public API.

abdoo8080:
This is work towards eliminating the Expr layer. This PR does the following:
- Replace Expr/Type with Term/Sort in the API for commands.
- Remove the command export functionality which is not supported.

Since many commands now call the `Solver` functions instead of the `SmtEngine` ones, their behavior may be a slightly different. For example, `(get-unsat-assumptions)` now only works in incremental mode. In some cases, `CVC4` will not recover from non-fatal errors.


src/main/command_executor.cpp
```diff
@@ -197,17 +197,21 @@ bool CommandExecutor::doCommandSingleton(Command* cmd)
...
+  }
+  else
+  {
+    cmd->invoke(solver, *out);
   }
   // ignore the error if the command-verbosity is 0 for this command
   std::string commandName =
       std::string("command-verbosity:") + cmd->getCommandName();
-  if(smt->getOption(commandName).getIntegerValue() == 0) {
+  if (solver->getSmtEngine()->getOption(commandName).getIntegerValue() == 0)
```
ajreynol, line 213:
```suggestion
  if (solver->getOption(commandName)=="0")
```

src/main/interactive_shell.cpp
```diff
@@ -321,10 +321,10 @@ Command* InteractiveShell::readCommand()
           s_declarations.insert(dynamic_cast<DeclareFunctionCommand*>(cmd)->getSymbol());
         } else if(dynamic_cast<DefineFunctionCommand*>(cmd) != NULL) {
           s_declarations.insert(dynamic_cast<DefineFunctionCommand*>(cmd)->getSymbol());
-        } else if(dynamic_cast<DeclareTypeCommand*>(cmd) != NULL) {
-          s_declarations.insert(dynamic_cast<DeclareTypeCommand*>(cmd)->getSymbol());
-        } else if(dynamic_cast<DefineTypeCommand*>(cmd) != NULL) {
-          s_declarations.insert(dynamic_cast<DefineTypeCommand*>(cmd)->getSymbol());
+        } else if(dynamic_cast<DeclareSortCommand*>(cmd) != NULL) {
+          s_declarations.insert(dynamic_cast<DeclareSortCommand*>(cmd)->getSymbol());
+        } else if(dynamic_cast<DefineSortCommand*>(cmd) != NULL) {
```
ajreynol, line 326: Remember to clang format.

src/smt/command.cpp
```diff
@@ -260,30 +241,25 @@ void EmptyCommand::toStream(std::ostream& out,
...
-void EchoCommand::invoke(SmtEngine* smtEngine, std::ostream& out)
+void EchoCommand::invoke(api::Solver* solver, std::ostream& out)
 {
   out << d_output << std::endl;
   Trace("dtview::command") << "* ~COMMAND: echo |" << d_output << "|~"
                            << std::endl;
   d_commandStatus = CommandSuccess::instance();
   printResult(out,
-              smtEngine->getOption("command-verbosity:" + getCommandName())
+              solver->getSmtEngine()
```
ajreynol, line 257: This might be able to be simplified using Solver::getOption, which returns a string that you can then just convert to an unsigned integer.

src/smt/command.cpp
```diff
@@ -1871,44 +1635,48 @@ void ExpandDefinitionsCommand::toStream(std::ostream& out,
...
+const std::vector<api::Term>& GetValueCommand::getTerms() const
+{
+  return d_terms;
+}
+void GetValueCommand::invoke(api::Solver* solver)
 {
   try
   {
-    ExprManager* em = smtEngine->getExprManager();
+    ExprManager* em = solver->getExprManager();
```
ajreynol, line 1666: I think we can delete this line.

ajreynol (changes requested):
Very happy to see this, just some minor requests.

src/main/interactive_shell.cpp
abdoo8080: Formatting 3 lines only does not look great, so I reformatted the try-catch block.

ajreynol (approved):
LGTM

PR #5110, opened 2020-09-21
bv2int: apply substitutions on translated formula.

### LINKED ISSUE #5095: (solve-bv-as-int) Fatal failure at preprocessing/passes/non_clausal_simp.cpp:128
Hi, for the following formula
~~~~
(set-logic BV)
(declare-fun _substvar_27_ () Bool)
(declare-const bv_40-3 (_ BitVec 40))
(assert (exists ((q110 (_ BitVec 40)) (q111 (_ BitVec 7))) (= (_ bv0 40) bv_40-3 (_ bv0 40) (_ bv0 40))))
(push 1)
(assert _substvar_27_)
(check-sat)
~~~~

cvc4 f92c4476e3
~~~~
/CVC4/build/bin/cvc4 -i --solve-bv-as-int=sum. xx.smt2  
Fatal failure within virtual CVC4::preprocessing::PreprocessingPassResult CVC4::preprocessing::passes::NonClausalSimp::applyInternal(CVC4::preprocessing::AssertionPipeline*) at /CVC4/src/preprocessing/passes/non_clausal_simp.cpp:128
Check failure

 top_level_substs.apply(learnedLiteral) == learnedLiteral

Aborted
~~~~

yoni206:
Closes #5095 .
The range constraints that this preprocessing pass introduces were not being passed through that top level substitution map, which caused an assertion failure.
This is fixed by applying the substitution map in the preprocessing pass before adding the constriants.
The test from #5095 is included.


ajreynol:
I believe to fix this you should simply apply bv-to-int before apply-subst: https://github.com/CVC4/CVC4/blob/master/src/smt/process_assertions.cpp#L210.

yoni206:
Replaced by #5208 .

PR #5111, opened 2020-09-21
Allow for theory combination of strings with nonlinear real arithmetic.

### LINKED ISSUE #5109: Allow for combination of strings and nonlinear real arithmetic
Within the logic setup in `set_defaults.cpp`, `QF_SNRA` triggers an unexpected result: the strings theory requires `LIA` and enforces `LIA` if arithmetic is disabled, we only have difference logic or we only have reals. For NRA, this triggers as well and enforces linear arithmetic, effectively downgrading to `QF_SLIRA`.

How this looks in action:
```
(set-logic QF_SNRA)
(declare-fun r2 () Real)
(assert (= 0 (* r2 r2)))
(check-sat)

$ bin/cvc4 delta.out.smt2 
(error "A non-linear fact was asserted to arithmetic in a linear logic.
The fact in question: (= (* r2 r2) 0.0)
")
```

nafur:
This PR makes sure that enabling strings and nonlinear real arithmetic at the same time works fine.
Right now, the configuration for strings enforces linear arithmetic if no integers are enabled, in particular for `NRA`.

Fixed #5109.


src/smt/set_defaults.cpp
```diff
@@ -553,6 +552,12 @@ void setDefaults(LogicInfo& logic, bool isInternalSubsolver)
       log.enableIntegers();
       log.arithOnlyLinear();
     }
+    if (!logic.areIntegersUsed())
```
ajreynol, line 555: Alternatively, we could just delete line 554 which sets that arithmetic should be only linear, right?  I guess this would by default make QF_S -> QF_SNIA?  If this is the case, I believe I agree with this PR.

nafur: I thought about that. If you are fine with switching from `QF_SLIA` to `QF_SNIA`, this would probably be better. Not only is it cleaner, it also avoids problem when we enable transcendentals (which are also disabled by `arithOnlyLinear`...)

ajreynol (approved):
LGTM

PR #5112, opened 2020-09-21
(proof-new) Make arrays proof producing

ajreynol:
This includes adding a basic inference manager to arrays that ensures that the correct applications of PfRule are given to the inference manager.

Note that some calls to lemma are yet to be converted. Also note that some minor simplifications were made for unnecessary parts of the code.


src/theory/arrays/theory_arrays.cpp
```diff
@@ -1179,38 +1180,28 @@ void TheoryArrays::presolve()
...
+  // the call to SkolemCache::getExtIndexSkolem should be deterministic, but use
+  // cache anyways for now
   Node skolem;
   std::unordered_map<Node, Node, NodeHashFunction>::iterator it = d_skolemCache.find(ref);
   if (it == d_skolemCache.end()) {
-    NodeManager* nm = NodeManager::currentNM();
-    skolem = nm->mkSkolem(name, type, comment);
+    Assert(ref.getKind() == kind::NOT && ref[0].getKind() == kind::EQUAL);
+    TNode a = ref[0][0];
+    TNode b = ref[0][1];
```
mpreiner, line 1192: `a` and `b` seem to be unused?

src/theory/arrays/theory_arrays.cpp
```diff
@@ -1347,26 +1339,22 @@ bool TheoryArrays::preNotifyFact(
 void TheoryArrays::notifyFact(TNode atom, bool pol, TNode fact, bool isInternal)
 {
   // if a disequality
-  if (atom.getKind() == kind::EQUAL && !pol)
+  if (atom.getKind() == kind::EQUAL && !pol && !isInternal)
   {
+    // Notice that we this should be an external assertion, since we do not
```
mpreiner, line 1344:
```suggestion
    // Notice that this should be an external assertion, since we do not
```

src/theory/arrays/theory_arrays.cpp
```diff
@@ -2053,14 +2047,13 @@ bool TheoryArrays::dischargeLemmas()
...
 void TheoryArrays::conflict(TNode a, TNode b) {
   Debug("pf::array") << "TheoryArrays::Conflict called" << std::endl;
-
-  explain(a.eqNode(b), d_conflictNode);
-
-  if (!d_inCheckModel) {
-    d_out->conflict(d_conflictNode);
+  if (d_inCheckModel)
+  {
+    // if in check model, don't send the conflict?
```
mpreiner, line 2052: why `?` ? :-)

src/theory/arrays/theory_arrays.cpp
ajreynol: I'll delete.

src/theory/arrays/theory_arrays.cpp
ajreynol: Haha, I still don't know why, but I guess the comment shouldn't have "?" :)

mpreiner (approved):

PR #5115, opened 2020-09-22
Modify lemma vs fact policy for datatype equalities

ajreynol:
This changes the lemma vs fact policy for datatype equalities.  Previously, datatype equalities were sent as lemmas unless they were over datatypes that were composed of datatypes only. This is now changed so that equalities that do not involve direct subterms with finite non-datatype types are kept internal.

The primary type of equality that this targets are "Instantiate" equalities, e.g. the conclusion of:
`(is-cons x) =>  x = (cons (head x) (tail x))`
These equalities have been observed to generate large amounts of new terms for many benchmarks.

With this PR, the the challenging Facebook benchmark goes from 2 min 45 sec -> 29 sec. If the instantiate rule is disabled altogether, it still correctly solves, and is faster (~14 seconds), which however is not correct in general.

This change triggered two other issues:
(1) A relations benchmark involving transitive closure now times out.  This has been a common issue for the relations solver and should be revisited.
(2) A potential issue with doPendingLemmas in InferenceManagerBuffer was uncovered. In rare cases, we can be re-entrant into this method since `OutputChannel::lemma` may trigger further preregistration of terms, which can trigger a recursive call to doPendingLemmas in the case of datatypes, which causes a segfault due to corrupting an iterator. This PR adds a simple guard for this method.

This PR also fixes some existing issues in computing cardinality for parametric datatypes.


ajreynol:
More generally, here are the results from the most recent set:
| benchmark                    | old    | new   |
| ---------------------------- | ------ | ----- |
| AccountLimits.smt2           | 18.34  | 10.87 |
| Authenticator.smt2           | 0.22   | 0.18  |
| Compare.smt2                 | 0.07   | 0.06  |
| CoreAddresses.smt2           | 1.34   | 1.15  |
| DesignatedDealer.smt2        | 154.54 | 33.43 |
| DualAttestationLimit.smt2    | 0.75   | 0.55  |
| Empty.smt2                   | 0.04   | 0.03  |
| Event.smt2                   | 2.71   | 1.21  |
| FixedPoint32.smt2            | 0.24   | 0.19  |
| LibraBlock.smt2              | 35.75  | 9.49  |
| LibraTimestamp.smt2          | 7.54   | 2.79  |
| LibraTransactionTimeout.smt2 | 1.07   | 0.69  |
| LibraVersion.smt2            | 59.73  | 12.91 |
| LibraVMConfig.smt2           | 65.62  | 23.06 |
| Offer.smt2                   | 0.33   | 0.35  |
| Option.smt2                  | 2.09   | 1.41  |
| RegisteredCurrencies.smt2    | 151.97 | 72.31 |
| Roles.smt2                   | 17.8   | 9.97  |
| Signature.smt2               | 0.04   | 0.11  |
| Signer.smt2                  | 0.04   | 0.1   |
| Testnet.smt2                 | 3.57   | 3.13  |
| Unhosted.smt2                | 1.73   | 1.26  |
| ValidatorConfig.smt2         | 13.51  | 10.82 |
| Vector.smt2                  | 0.07   | 0.04  |

barrettcw (commented):
What happens if you have x = C(S1(x),S2(y)) and S1(x) is a finite type but S2(y) is an infinite type?  Looks like in this case you don't send out the lemma, but is this correct?

ajreynol:
Yes I don't send out the lemma in this case.  The case is trickier, but I (think) it should be safe to keep equalities internal.  Let me play around with seeing if I can break things (with e.g. lists of Boolean).

barrettcw:
> Yes I don't send out the lemma in this case. The case is trickier, but I (think) it should be safe to keep equalities internal. Let me play around with seeing if I can break things (with e.g. lists of Boolean).

Here's a concrete example.  Suppose S1 : D -> BV[1] (where D is the datatype).  And let's say we have x,y,z:D.  Then we have the equations S1(x) /= S1(y), S1(x) /= S1(z), and S1(y) /= S1(z).  This is unsat, but all the terms here are datatypes so they won't get shared.

ajreynol:
Yes, I think there is an issue, here is a similar example I tried which gets "sat" incorrectly:
```
(set-logic ALL)
(set-info :status unsat)
(declare-datatypes ((list 0)) (
((cons (head Bool) (tail list)) (nil))
))
(declare-fun x1 () list)
(declare-fun x2 () list)
(declare-fun x3 () list)
(assert (= (tail x1) nil))
(assert (= (tail x2) nil))
(assert (= (tail x3) nil))
(assert (distinct x1 x2 x3 nil))
(check-sat)
```
What about: send the lemma if any `s(x)` in the instantiated term has finite external type?  I will try this now.

barrettcw:
Actually this is pretty interesting.  I think you can be a bit smarter - I think instead of sending lemmas, you can just label the selectors as shared terms, but only if the type is finite.  This actually raises an interesting theoretical question because in essence this means you are doing polite combination on some of the sorts and Nelson-Oppen on others!  I will ask Yoni and Ying to think about this!

ajreynol:
Btw, one thing to note is that we don't have an interface for forcing a term to be considered shared.  In fact, the relations solver has a hack for this (which is via a dummy lemma): https://github.com/CVC4/CVC4/blob/master/src/theory/sets/theory_sets_rels.cpp#L1186
Would be interested in if there was a better way of doing this. 

Regardless, I'll try sending the lemma for now.

barrettcw:
Hmm - yeah - we should have a better way to do that.

-Clark

src/expr/dtype_cons.h
```diff
@@ -286,6 +304,13 @@ class DTypeConstructor
                          std::vector<TypeNode>& processing,
                          std::map<TypeNode, Node>& gt,
                          bool isValue) const;
+  /**
+   * Compute cardinality info, returns a pair where its first component is
+   * an identifier indicating the cardinality type of this constructor for
+   * type t, and a Boolean indicating whether the constructor has any arguments
```
barrettcw, line 310: Should "t" be "tn" here?

barrettcw (approved):
LGTM

src/expr/dtype_cons.h
ajreynol: Yes, I'll fix.

PR #5117, opened 2020-09-22
Allow E-matching by default when strings are enabled

ajreynol:
This does not disable E-matching when strings are enabled with `--strings-exp`, due to potential applications where strings are combined with user-provided quantified formulas.

Notice that by default, we do not want E-matching applied to quantified formulas corresponding to reductions for extended string functions.  To correct this, all quantified formulas generated by strings are marked with an "internal quantified formula" attribute, which causes E-matching to ignore them.  This feature can in theory be generalized later for other internal sources of quantified formulas.


4tXJ7f (approved):
LGTM. Minor typo.

PR #5118, opened 2020-09-22
Fix compilation without LibPoly

4tXJ7f:
Commit e969318f12d4e8ee01b12933e9e60fafafd96963 introduced the ICP-based
solver for nonlinear arithmetic. That code, however, depends on LibPoly.
When configuring CVC4 without LibPoly, the code doesn't compile because
the class `ICPSolver` is missing. This commit adds a dummy version if
`ICPSolver` to remedy the issue.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,7 +23,7 @@ jobs:
 
         include:
           - name: production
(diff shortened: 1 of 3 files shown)


nafur (approved):
LGTM. Thanks for fixing this.

PR #5119, opened 2020-09-22
FP: Use Assert instead of PRECONDITION macro in converter.

aniemetz:
The FP converter (= word blaster) uses a PRECONDITION macro that is
defined in symFPU (if included) and calls traits::precondition. This is
a problem for two reasons. First, traits::precondition calls
AlwaysAssert, and PRECONDITION is called for every word-blasted node, so
this can be expensive. Second, in case of an assertion failure, we get
very generic failure messages, that don't allow to distinguish between
failures and are absolutely non-descriptive:
```
Check failure

b
```
This fixes both issues.

@martin-cs


ajreynol (approved):
LGTM

martin-cs (approved):
Yes; now this code is part of CVC4 this is the right way of doing this.

PR #5120, opened 2020-09-22
[Python API] Conversion to/from Unicode strings

### LINKED ISSUE #5024: Python API: Strings and Escape Characters
```
x = slv.mkString("\"test\"")
```

I would expect x to be ""x"", when I print out the term or solve for it, but instead I get 
```
In [24]: x
Out[24]: """test"""
```
Solving for x also results in value being `"""test"""`

I'm not sure if this is a bug or if there is another prescribed mechanism to apply escaped characters when generating strings.

4tXJ7f:
Fixes #5024. This commit adds a conversion from constant string terms to
native Python Unicode strings in `Term.toPythonObj()` and improves
`Solver.mkString()` to accept strings containing characters outside of
the printable range.


test/unit/api/python/test_to_python_obj.py
```diff
@@ -64,3 +62,15 @@ def testGetSymbol():
...
+
+
+def testGetString():
+    solver = pycvc4.Solver()
+
+    s1 = '"test\n"😃\\u{a}'
+    t1 = solver.mkString(s1)
+    assert s1 == t1.toPythonObj()
+
+    s2 = '❤️CVC4❤️'
```
ajreynol, line 74: Nice :)

ajreynol (approved):
LGTM

PR #5121, opened 2020-09-23
FP: Use Assert instead of AlwaysAssert in traits::(pre|post)condition.

aniemetz:
For the same reason as in #5119.
@martin-cs


ajreynol (approved):
LGTM

PR #5122, opened 2020-09-23
New C++ API: Catch and throw recoverable exception.

aniemetz:
This allows to distinguish recoverable from non-recoverable exceptions,
and recover if possible.


abdoo8080 (approved):
Thanks a lot for the PR!

PR #5126, opened 2020-09-23
Make IAND solver use inference manager.

nafur:
This PR moves the iand solver (within the nonlinear extension) to use the new inference manager to send lemmas.

@yoni206 I understand that this solver is mostly used by you. Can you check whether these changes entail behavioral changes? (They should not, "do not merge" until confirmed)


ajreynol (approved):
LGTM, although Yoni should have a quick look.

yoni206 (approved):
LGTM

PR #5127, opened 2020-09-23
Fix issue 5116

mudathirmahgoub:
This PR fixes the issue of rewriting equality between empty sets of different types as false. 
Now  `(= (as emptyset (Set Int)) (as emptyset (Set Real)))` is rewritten  as `true`.


src/theory/sets/theory_sets_rewriter.cpp
```diff
@@ -482,7 +482,16 @@ RewriteResponse TheorySetsRewriter::preRewrite(TNode node) {
...
   {
-    if(node[0] == node[1]) {
+    if (node[0] == node[1])
+    {
+      return RewriteResponse(REWRITE_DONE, nm->mkConst(true));
+    }
+    // check equality between empty sets (possibly of different types)
+    // e.g.  (= (as emptyset (Set Int)) (as emptyset (Set Real)))
+    if (node[0].getKind() == EMPTYSET && node[1].getKind() == EMPTYSET
+        && (node[0].getType().isSubtypeOf(node[1].getType())
```
ajreynol, line 492: I believe the 3rd condition is redundant: it should hold for all well-typed equalities.

ajreynol (changes requested):
Nice catch, one suggestion for improvement.

I'm wondering if there are more issues related to subtyping sets beyond this fix, or whether this happens to fix the current one.  Regardless, this is a correct and good rewrite to perform and suffices to close the current issue.

mudathirmahgoub:
This PR generated an assertion error for `--debug-check-models` and after an offline discussion, we decided to abandon sub-typing for sets altogether in future PR.

PR #5129, opened 2020-09-24
Use original quantified formula for single invocation reconstruction

ajreynol:
This fixes an issue with single invocation solution reconstruction where the preprocessed quantified formula contains internal skolems, thus prohibiting reconstruction.  The solution is to use the original quantified formula when doing solution reconstruction.  This adds a regression demonstrating the issue.


HanielB (approved):
LGTM

PR #5131, opened 2020-09-24
Add interface for getting relevant assertions

ajreynol:
This adds an interface to TheoryEngine for getting the current set of relevant assertions if it is available.

An interface to this can further be added to the API in a future PR.


4tXJ7f (approved):
LGTM

PR #5132, opened 2020-09-24
Implement bags rewriter

mudathirmahgoub:
This PR implements rewrite rules for bags. This PR focuses on rewrite rules for non constant nodes. 
Rewriting nodes with constant children is delegated to bags::NormalForm class (future PR).


src/theory/bags/bags_rewriter.cpp
```diff
@@ -0,0 +1,427 @@
...
+RewriteResponse BagsRewriter::rewriteEqual(const TNode& n) const
+{
+  Assert(n.getKind() == EQUAL);
+  if (n[0] == n[1])
+  {
+    // (= A A) = true where A is a bag
+    NodeManager* nm = NodeManager::currentNM();
+    return RewriteResponse(REWRITE_DONE, nm->mkConst(true));
+  }
+  if (n[0].getKind() == MK_BAG && n[1].getKind() == MK_BAG)
```
ajreynol, line 76: We don't want this rewrite.  Equalities (= t s) should either be rewritten to true, false, (= t s), or (= s t) due to the way theory combination works.  This violates this rule.

Also, we generally never introduce ITE in the rewriter, since this makes things harder to reason about.

src/theory/bags/bags_rewriter.cpp
```diff
@@ -0,0 +1,427 @@
...
+    Node emptybag = nm->mkConst(EmptyBag(n.getType()));
+    return RewriteResponse(REWRITE_DONE, emptybag);
+  }
+  return RewriteResponse(REWRITE_DONE, n);
+}
+
+RewriteResponse BagsRewriter::rewriteBagCount(const TNode& n) const
+{
+  Assert(n.getKind() == BAG_COUNT);
+  if (n[1].isConst() && n[1].getKind() == EMPTYBAG)
```
ajreynol, line 125: This should be generalized to handle all cases of constant arguments.  You should perhaps make a helper function `computeCount( Node x, Node b)` where x and b are constants.

src/theory/bags/bags_rewriter.cpp
```diff
@@ -0,0 +1,427 @@
...
+    return RewriteResponse(REWRITE_DONE, n[0]);
+  }
+  if (n[0].getKind() == EMPTYBAG)
+  {
+    // (union_max emptybag A) = A
+    return RewriteResponse(REWRITE_DONE, n[1]);
+  }
+  if (n[1].getKind() == UNION_MAX && (n[0] == n[1][0] || n[0] == n[1][1]))
+  {
+    // (union_max A (union_max A B) = (union_max A B)
```
ajreynol, line 155:
```suggestion
    // (union_max A (union_max A B)) = (union_max A B)
```

src/theory/bags/bags_rewriter.cpp
```diff
@@ -0,0 +1,427 @@
...
+    // (union_max A (union_max B A) = (union_max B A)
```
ajreynol, line 156:
```suggestion
    // (union_max A (union_max B A)) = (union_max B A)
```

src/theory/bags/bags_rewriter.cpp
```diff
@@ -0,0 +1,427 @@
...
+    return RewriteResponse(REWRITE_DONE, n[1]);
+  }
+  if (n[0].getKind() == UNION_MAX && (n[0][0] == n[1] || n[0][1] == n[1]))
+  {
+    // (union_max (union_max A B) A) = (union_max A B)
+    // (union_max (union_max B A) A) = (union_max B A)
+    return RewriteResponse(REWRITE_DONE, n[0]);
+  }
+
+  if (n[1].getKind() == UNION_DISJOINT && (n[0] == n[1][0] || n[0] == n[1][1]))
```
ajreynol, line 166: These could perhaps be combined with the cases above, this would make things easier to read.

src/theory/bags/bags_rewriter.cpp
```diff
@@ -0,0 +1,427 @@
...
+  {
+    // (union_disjoint A emptybag) = A
+    return RewriteResponse(REWRITE_DONE, n[0]);
+  }
+  if (n[0].getKind() == EMPTYBAG)
+  {
+    // (union_disjoint emptybag A) = A
+    return RewriteResponse(REWRITE_DONE, n[1]);
+  }
+  if ((n[0].getKind() == UNION_MAX && n[1].getKind() == INTERSECTION_MIN)
```
ajreynol, line 194: This rewrite seems fairly specific.  I would hold off on writing anything too specific like this until we have a better idea of whats needed.

src/theory/bags/bags_rewriter.cpp
```diff
@@ -0,0 +1,427 @@
...
+    // (bag.choose (mkBag x c)) = x where c is a constant > 0
+    return RewriteResponse(REWRITE_DONE, n[0][0]);
+  }
+  return RewriteResponse(REWRITE_DONE, n);
+}
+
+RewriteResponse BagsRewriter::rewriteCard(const TNode& n) const
+{
+  Assert(n.getKind() == BAG_CARD);
+  if (n[0].getKind() == EMPTYBAG)
```
ajreynol, line 380: Ditto, this should be generalized to when the argument is constant.

src/theory/bags/bags_rewriter.cpp
```diff
@@ -0,0 +1,427 @@
...
+  Assert(n.getKind() == BAG_IS_SINGLETON);
+  NodeManager* nm = NodeManager::currentNM();
+  if (n[0].getKind() == EMPTYBAG)
+  {
+    // (bag.is_singleton emptybag) = false
+    return RewriteResponse(REWRITE_DONE, nm->mkConst(false));
+  }
+  if (n[0].getKind() == MK_BAG)
+  {
+    // (bag.is_singleton (mkBag x c) = (c == 1)
```
ajreynol, line 417:
```suggestion
    // (bag.is_singleton (mkBag x c)) = (c == 1)
```

ajreynol (changes requested):
The most important thing is to add constant evaluation rules, which I'm not sure are complete here.

Also, I would strongly suggest having a helper method for each rewrite so that we can have uniform debugging information for all rewrites.  Concretely, we should have a returnRewrite method similar to strings: https://github.com/CVC4/CVC4/blob/master/src/theory/strings/sequences_rewriter.h#L125
This also means that we should have identifiers for each return statement in the rewriter where a node is rewritten.

src/theory/bags/bags_rewriter.cpp
mudathirmahgoub: Done by removing the ITE rule. Probably we would need this [ite commit](https://github.com/mudathirmahgoub/CVC4/commit/4c74f3357710cc11994446f735a59387777d52b4) later in the solver.

src/theory/bags/bags_rewriter.cpp
mudathirmahgoub: In this place I am assuming `x` is not a constant. I plan to implement `NormalForm::getNormalFrom` in another pull request to make this one easier.

src/theory/bags/bags_rewriter.cpp
mudathirmahgoub: Done

src/theory/bags/bags_rewriter.cpp
mudathirmahgoub: done

src/theory/bags/bags_rewriter.cpp
mudathirmahgoub: I agree it is very specific, but I like it :) Do you suggest to remove it?

src/theory/bags/bags_rewriter.cpp
mudathirmahgoub: I agree this should be moved to `NormalForm::getNormalForm` function. For now it is removed from here

src/theory/bags/bags_rewriter.cpp
mudathirmahgoub: Done.  Also removed the rule ` (bag.is_singleton emptybag)  = false` which should be in `NormalForm::getNormalForm`

src/theory/bags/bags_rewriter.cpp
mudathirmahgoub: Done.

mudathirmahgoub:
> 
> 
> The most important thing is to add constant evaluation rules, which I'm not sure are complete here.
> 
> Also, I would strongly suggest having a helper method for each rewrite so that we can have uniform debugging information for all rewrites. Concretely, we should have a returnRewrite method similar to strings: https://github.com/CVC4/CVC4/blob/master/src/theory/strings/sequences_rewriter.h#L125
> This also means that we should have identifiers for each return statement in the rewriter where a node is rewritten.

The functionality is implemented using class [`BagsRewriteResponse`](https://github.com/CVC4/CVC4/pull/5132/files#diff-9c87b3391a16b961fb878ce6a8b900dbR27) and enum [`Rewrite`](https://github.com/CVC4/CVC4/pull/5132/files#diff-b8787daa3713d5e91683690f3a21c312R30).

src/theory/bags/bags_rewriter.h
```diff
@@ -0,0 +1,193 @@
...
+
+#include "theory/rewriter.h"
+
+namespace CVC4 {
+namespace theory {
+namespace bags {
+
+struct BagsRewriteResponse
+{
+  RewriteStatus d_status;
```
ajreynol, line 29: Give one line documentation for these members.

src/theory/bags/bags_rewriter.h
```diff
@@ -0,0 +1,193 @@
...
+#ifndef CVC4__THEORY__BAGS__THEORY_BAGS_REWRITER_H
+#define CVC4__THEORY__BAGS__THEORY_BAGS_REWRITER_H
+
+#include "theory/rewriter.h"
+
+namespace CVC4 {
+namespace theory {
+namespace bags {
+
+struct BagsRewriteResponse
```
ajreynol, line 27: Give a brief description of the purpose of this class.

src/theory/bags/bags_rewriter.h
```diff
@@ -0,0 +1,193 @@
...
+
+  /**
+   * patterns for n
+   * - (bag.is_included A B) = ((difference_subtract A B) == emptybag)
+   */
+  BagsRewriteResponse rewriteIsIncluded(const TNode& n) const;
+
+  /**
+   * patterns for n:
+   * - (mkBag x 0) = (emptybag T) where T is the type of x
```
ajreynol, line 71: I would maybe keep this documentation elsewhere.

The problems is that rewrites may evolve rapidly and it is hard to keep documentation like this up to date.

src/theory/bags/bags_rewriter.cpp
```diff
@@ -0,0 +1,456 @@
...
+{
+  BagsRewriteResponse response;
+  if (n.isConst())
+  {
+    // no need to rewrite n if it is already in a normal form
+    response = BagsRewriteResponse(REWRITE_DONE, n, Rewrite::NO_REWRITE);
+  }
+  else if (NormalForm::AreChildrenConstants(n))
+  {
+    Node normal = NormalForm::getNormalForm(n);
```
ajreynol, line 65: I think the better name for this method is `NormalForm::evaluate`, whereas `getNormalForm` suggests something more general.  I believe we need a method whose sole purpose is to evaluate bag operators applied to constant arguments (which FYI will later need to be incorporated into https://github.com/CVC4/CVC4/blob/master/src/theory/evaluator.h when doing fast SyGuS involving bags).

ajreynol (changes requested):

src/theory/bags/bags_rewriter.cpp
ajreynol: Its fine, lets keep it :)

src/theory/bags/bags_rewriter.h
mudathirmahgoub: Done. Perhaps this can be refactored with `RewriteResponse` class in the future.

src/theory/bags/bags_rewriter.h
mudathirmahgoub: Done

src/theory/bags/bags_rewriter.h
mudathirmahgoub: I would argue the mental effort to understand the rule without documentation later surpasses the painful effort of updating the documentation :)
Also writing the documentation makes it harder to make errors like `n[0][0]` when you meant `n[0][1]`.

src/theory/bags/bags_rewriter.cpp
mudathirmahgoub: Done.

src/theory/bags/bags_rewriter.h
```diff
@@ -0,0 +1,194 @@
...
+#include "theory/rewriter.h"
+
+namespace CVC4 {
+namespace theory {
+namespace bags {
+
+/** a class represents the result of rewriting bag nodes */
+struct BagsRewriteResponse
+{
+  /** the rewritten node */
```
ajreynol, line 30: Data members should come after methods according to our style guidelines.

src/theory/bags/bags_rewriter.h
ajreynol: Ok, I would make the documentation less specific e.g. "patterns for rewrites include...".

src/theory/bags/bags_rewriter.h
```diff
@@ -0,0 +1,194 @@
...
+  BagsRewriteResponse rewriteEqual(const TNode& n) const;
+
+  /**
+   * patterns for n
+   * - (bag.is_included A B) = ((difference_subtract A B) == emptybag)
+   */
+  BagsRewriteResponse rewriteIsIncluded(const TNode& n) const;
+
+  /**
+   * patterns for n:
```
ajreynol, line 71:
```suggestion
   * rewrites for n include:
```

src/theory/bags/normal_form.h
```diff
@@ -36,6 +37,14 @@ class NormalForm
    * Also handles the corner cases of empty bag and singleton bag.
    */
   static bool checkNormalConstant(TNode n);
+  /**
+   * check whether all children of the given node are in normal form
+   */
+  static bool AreChildrenConstants(TNode n);
+  /**
+   * get the normal form of n
```
ajreynol, line 45: Update.

ajreynol (changes requested):
A few more minor requests.

src/theory/bags/bags_rewriter.h
mudathirmahgoub: Done

src/theory/bags/bags_rewriter.h
mudathirmahgoub: done for all comments

src/theory/bags/normal_form.h
mudathirmahgoub: done

ajreynol (approved):
LGTM, thanks for the updates.

PR #5134, opened 2020-09-25
Cleaning and documenting cnf stream

HanielB:
Moreover changes assertClause to return a bool, which is gonna be used by the proof cnf stream.


src/prop/cnf_stream.h
```diff
@@ -54,7 +59,122 @@ class CnfStream {
...
+   * Constructs a CnfStream that performs equisatisfiable CNF transformations
+   * and sends the generated clauses and to the given SAT solver. This does not
+   * take ownership of satSolver, registrar, or context.
+   *
+   * @param satSolver the sat solver to use.
+   * @param registrar the entity that takes care of preregistration of Nodes.
+   * @param context the context that the CNF should respect.
+   * @param outMgr Reference to the output manager of the smt engine. Assertions
+   * will not be dumped if outMgr == nullptr.
+   * @param fullLitToNodeMap maintain a full SAT-literal-to-Node mapping.
```
ajreynol, line 72: Resource manager is missing from doc.

src/prop/prop_engine.cpp
```diff
@@ -91,7 +91,7 @@ PropEngine::PropEngine(TheoryEngine* te,
   d_satSolver = SatSolverFactory::createDPLLMinisat(smtStatisticsRegistry());
 
   d_registrar = new theory::TheoryRegistrar(d_theoryEngine);
-  d_cnfStream = new CVC4::prop::TseitinCnfStream(
+  d_cnfStream = new CVC4::prop::CnfStream(
```
ajreynol, line 94: Why this change (and the ones below?).  Was the derived class redundant?

src/prop/cnf_stream.h
ajreynol: Also, I assume you are adding the resource manager because it will be used?

ajreynol (changes requested):
Looks good, just some minor questions and comments.

src/prop/prop_engine.cpp
HanielB: The CnfStream was only ever used as a TseitingCnfStream extension. So I collapsed the latter into the former.

src/prop/cnf_stream.h
HanielB: Will add. This was because it was used in the extension.

ajreynol (approved):

PR #5137, opened 2020-09-25
[proof-new] Adds a proof-producing CNF converter

HanielB:
A proof generator that wraps the original CNF stream, to be used when the prop engine is proof producing. It tracks in a lazy cdproof both the concrete clausification steps and the proof generators of the formulas being clausified (in particular, theory lemmas).


src/prop/proof_cnf_stream.cpp
```diff
@@ -0,0 +1,1265 @@
...
+    default:
+    {
+      // negate
+      Node nnode = negated ? node.negate() : static_cast<Node>(node);
+      // Atoms
+      SatLiteral lit = toCNF(node, negated);
+      bool added = d_cnfStream.assertClause(nnode, lit);
+      if (negated && added && nnode != node.notNode())
+      {
+        d_proof.addStep(
```
ajreynol, line 126: Can you explain whats going on here?

src/prop/proof_cnf_stream.cpp
```diff
@@ -0,0 +1,1265 @@
...
+      {
+        disjuncts.push_back(node[i].notNode());
+      }
+      Node clauseNode = NodeManager::currentNM()->mkNode(kind::OR, disjuncts);
+      d_proof.addStep(clauseNode, PfRule::NOT_AND, {node.notNode()}, {});
+      Trace("cnf") << "ProofCnfStream::convertAndAssertAnd: NOT_AND added "
+                   << clauseNode << "\n";
+      // justify normalized clause as well, since that's what will be saved in
+      // the SAT solver and registered with prop engine
+      Node normClauseNode = d_psb.factorReorderElimDoubleNeg(clauseNode);
```
ajreynol, line 185: It seems that there is an invariant that this method should be called whenever you registerSatAssumptions below.

Would it make sense to make a utility function that does this?  It seems this pattern is repeated several times below.

src/prop/proof_cnf_stream.cpp
```diff
@@ -0,0 +1,1265 @@
...
+                   << clauseNode << "\n";
+      Node normClauseNode = d_psb.factorReorderElimDoubleNeg(clauseNode);
+      if (Trace.isOn("cnf") && normClauseNode != clauseNode)
+      {
+        Trace("cnf")
+            << "ProofCnfStream::convertAndAssertIff: steps to normalized "
+            << normClauseNode << "\n";
+      }
+      MinisatSatSolver* minisat =
+          static_cast<MinisatSatSolver*>(d_cnfStream.d_satSolver);
```
ajreynol, line 381: Should you just do a single static_cast in the constructor?  Or better, have the constructor of this class take `MinisatSatSolver *`?

ajreynol (changes requested):
A few initial comments.

src/prop/proof_cnf_stream.cpp
HanielB: It's tracking double negation elimination. I'll add a comment.

src/prop/proof_cnf_stream.cpp
HanielB: I will actually make this class take the SatProofManager, that's the real dependency here.

src/prop/proof_cnf_stream.cpp
HanielB: Indeed. I've updated the class to do this.

ajreynol (approved):
LGTM

PR #5138, opened 2020-09-25
Restrict bvxnor to only allow two operands (was n-ary).

aniemetz:
Bit-vector operator bvxnor was previously mistakenly marked as
left-assoicative in SMT-LIB. This has recently been corrected in
SMT-LIB. We  now restrict bvxnor to only allow two operands in order to
avoid confusion about the semantics, since the behavior of n-ary
operands to bvxnor is now undefined in SMT-LIB.


4tXJ7f (approved):
LGTM

PR #5140, opened 2020-09-26
[proof-new] Adds a proof manager for the SAT solver

HanielB:
Tracks the refutation proof built by Minisat. See the header for extensive explanations.

This commit also adds a few dependencies for the SAT proof manager to work (making it a friend of the SAT solver, getting the cnf stream from theory proxy, having lazy cdproof chain give all the links).


src/prop/sat_proof_manager.h
```diff
@@ -0,0 +1,559 @@
...
+   */
+  LazyCDProofChain d_resChains;
+
+  /** The proof generator for resolution chains */
+  BufferedProofGenerator d_resChainPg;
+
+  /** The false node */
+  Node d_false;
+
+  /** All clauses added to the SAT solver, kept in a context-dependend manner.
```
ajreynol, line 534:
```suggestion
  /** All clauses added to the SAT solver, kept in a context-dependent manner.
```

src/prop/sat_proof_manager.cpp
```diff
@@ -0,0 +1,729 @@
...
+  for (unsigned i = 0, size = clause.size(); i < size; ++i)
+  {
+    SatLiteral satLit = MinisatSatSolver::toSatLiteral(clause[i]);
+    Trace("sat-proof") << satLit << " ";
+  }
+}
+
+Node SatProofManager::getClauseNode(SatLiteral satLit)
+{
+  Assert(d_proxy->getCnfStream()->getNodeCache().find(satLit)
```
ajreynol, line 53: Can you cache the CNF stream of d_proxy as a data member of this class?  It seems it is used very heavily.

ajreynol: Actually, I'm curious if you can just pass the CNF stream to the constructor of this class instead of d_proxy.

src/prop/sat_proof_manager.h
```diff
@@ -0,0 +1,559 @@
...
+{
+ public:
+  SatProofManager(Minisat::Solver* solver,
+                  TheoryProxy* proxy,
+                  context::UserContext* userContext,
+                  ProofNodeManager* pnm);
+
+  /** Marks the start of a resolution chain.
+   *
+   * The given clause, at the node level, is registered in d_resLinks with a
```
ajreynol, line 263: I assume this is connected to endResChain?  Can you say the expectation about the timeline of when these are called?  E.g. it is expected that there is a call of endResChain that follows startResChain after X.

src/prop/sat_proof_manager.h
```diff
@@ -0,0 +1,559 @@
...
+   * null pivot, since this is the first link in the chain.
+   *
+   * @param start a SAT solver clause (list of literals) that will be the first
+   * clause in a resolution chain.
+   */
+  void startResChain(const Minisat::Clause& start);
+  /** Adds a resolution step with a clause
+   *
+   * The resolution step is added to d_resLinks with the clause, at the node
```
ajreynol, line 272: Do indicate if these methods have some dependency on the other methods (occurring in between startResChain/endResChain).  Its not clear from this whether there is a dependency.

src/prop/sat_proof_manager.h
```diff
@@ -0,0 +1,559 @@
...
+   */
+  void finalizeProof(Minisat::Lit inConflict, bool adding = false);
+  /** As above, but uses the unit conflict clause saved in d_conflictLit. */
+  void finalizeProof();
+  /** Set the unit conflict clause d_conflictLit. */
+  void storeUnitConflict(Minisat::Lit inConflict);
+
+  /** Retrive the refutation proof
+   *
+   * If there is no chain for false in d_resChains, an assumption proof node is
```
ajreynol, line 322: I assume there are some dependencies for when this method should be called (after a call to finalizeProof maybe?).

It would be helpful to have some basic usage details on the order of the calls to these methods.  In other words, it would be nice to have something like a reg exp that describes the intended use of this class, for each check-sat call, like roughly:

`registerSatAssumptions; ( beginResChain (addResolutionStep )* endResChain)*; finalizeProof; getProof`

ajreynol (changes requested):
Thanks for the documentation, this is nice to see.

I have a few general comments to improve things.

src/prop/sat_proof_manager.h
```diff
@@ -0,0 +1,559 @@
...
+}
+
+namespace CVC4 {
+namespace prop {
+
+/**
+ * This class is responsible for managing the proof production of the SAT
+ * solver. It tracks resolutions produced during solving and stores them,
+ * independently, with the connection of the resolutions only made when the
+ * empty clause is produced and the refutation is complete. These resolutions
```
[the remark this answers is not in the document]
HanielB: The clausified preprocessed assertions and lemmas. I'll add.

src/prop/sat_proof_manager.h
HanielB: This is nice indeed, I'll add and clarify the dependencies in the methods doc as well.

ajreynol (approved):
LGTM, thanks for all the documentation.

PR #5148, opened 2020-09-28
Add new arithmetic BoundInference class

nafur:
This PR adds a new utility class that extracts bounds on single variables from theory atoms like `x > 0` or `y <= 17`. It does not perform further reasoning (like recognizing `x > y` with `y > 3` as a bound).
Note that the functionality was introduced in `arith/nl/icp/variable_bounds.h`, but is now available using `Node` only.
This PR also replaces the `VariableBounds` by `BoundInference` in the icp solver which allowed for checking the code.


ajreynol (approved):
LGTM

PR #5153, opened 2020-09-28
Reset assertions on resetAssertions

### LINKED ISSUE #5144: Refutation soundness issue on QF_LIA formula
Hi, for the following formula
~~~~
(set-logic QF_LIA)
(assert (= 0 1))
(reset-assertions)
(check-sat)
~~~~
https://cvc4.cs.stanford.edu/downloads/builds/x86_64-linux-opt/unstable/cvc4-2020-09-26-x86_64-linux-opt yields `unsat`,
but the formula should be `sat`.

ajreynol:
The assertions object is currently not cleared on resetAssertions, only the prop engine is reset. This means that if a user added assertion, did not use them in a check-sat, and then called reset-assertions, they would not be removed from the assertions pipeline (storing the pending assertions before they are sent to the prop engine).

This fixes #5144.

### DIFF
--- a/src/smt/smt_engine.cpp
+++ b/src/smt/smt_engine.cpp
@@ -2028,6 +2028,7 @@ void SmtEngine::resetAssertions()
         getOutputManager().getDumpOut());
   }
 
+  d_asserts->clearCurrent();
   d_state->notifyResetAssertions();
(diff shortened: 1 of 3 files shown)


4tXJ7f (approved):
LGTM

PR #5156, opened 2020-09-28
[proof-new] Updates to proof node updater

HanielB:


src/expr/proof_node_updater.cpp
```diff
@@ -73,6 +78,9 @@ void ProofNodeUpdater::processInternal(std::shared_ptr<ProofNode> pf,
   std::shared_ptr<ProofNode> cur;
   visit.push_back(pf);
   std::map<Node, std::shared_ptr<ProofNode>>::iterator itc;
+  // NOTE: temporary, debugging
```
ajreynol, line 81: This shouldn't go to master.

src/expr/proof_node_updater.cpp
```diff
@@ -82,19 +90,26 @@ void ProofNodeUpdater::processInternal(std::shared_ptr<ProofNode> pf,
   {
     cur = visit.back();
     visit.pop_back();
+    // don't continue visiting proof node
```
ajreynol, line 93: I don't think this is the right place for this, since we haven't determined yet if cur is a new proof node to be processed.  Moreover I think the shouldContinue should be a flag on shouldUpdate (see other comment).

src/expr/proof_node_updater.cpp
```diff
@@ -22,6 +22,11 @@ namespace CVC4 {
 ProofNodeUpdaterCallback::ProofNodeUpdaterCallback() {}
 ProofNodeUpdaterCallback::~ProofNodeUpdaterCallback() {}
 
+bool ProofNodeUpdaterCallback::shouldContinue(std::shared_ptr<ProofNode> pn)
```
ajreynol, line 25: Can we make this a flag on shouldUpdate so that it is parallel to how the shouldContinue flag works on update?

ajreynol (changes requested):

src/expr/proof_node_updater.h
```diff
@@ -36,7 +36,8 @@ class ProofNodeUpdaterCallback
   ProofNodeUpdaterCallback();
   virtual ~ProofNodeUpdaterCallback();
   /** Should proof pn be updated? */
-  virtual bool shouldUpdate(ProofNode* pn) = 0;
+  virtual bool shouldUpdate(std::shared_ptr<ProofNode> pn,
```
ajreynol, line 39: Document the continueUpdate flag.

HanielB: done

ajreynol (approved):

PR #5158, opened 2020-09-29
Add GH action to automatically update copyright headers.

mpreiner:
This PR adds a new CI build for pull requests that automatically executes the update-copyright.pl script on all files added/modified in the PR.

### DIFF
--- a/.github/workflows/update.yml
+++ b/.github/workflows/update.yml
@@ -1,0 +1,25 @@
+name: Update Copyright Headers
+
+on: [pull_request, pull_request_target]
+
+jobs:
+  update:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+          ref: ${{ github.head_ref }}
+
+      - id: files
+        uses: jitterbit/get-changed-files@v1
+
+      - name: Update headers
+        run: |
+          contrib/update-copyright.pl ${{ steps.files.outputs.added_modified }}
+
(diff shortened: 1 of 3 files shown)


4tXJ7f (approved):
🤘

nafur:
I **really** like this idea. A few questions:
- Will we have "Update copyright headers" commits all over the place in our PRs?
- If so: will they trigger new CI jobs?
- While this is not really a problem for our own PRs as we squash them anyway, what about PRs from external contributors?
- Can we do the same with formatting?

mpreiner:
>     * Will we have "Update copyright headers" commits all over the place in our PRs?
> 
After every push to your branch/PR the action is triggered and adds a commit.

>     * If so: will they trigger new CI jobs?
> 

It should not trigger a new CI job since the trigger is only on pull_request and pull_request_target

>     * While this is not really a problem for our own PRs as we squash them anyway, what about PRs from external contributors?
> 

External contributors have to squash and sign their commits so this shouldn't be an issue (they only have to squash+sign again if a header was updated).

>     * Can we do the same with formatting?

We can, but we decided yesterday in the meeting that we'll not automatically apply clang-format. Instead we'll add a check that fails if the code is not properly formatted. This avoids nasty merge conflicts when pulling from the branch since formatting potentially modifies a lot of places in the code.

mpreiner:
Unfortunately, it seems that updating headers does not work on PRs from forks (it does only work on the same repo). The changes need to be pushed to the original branch, but the PR on the main repository can not push commits to the fork.

PR #5159, opened 2020-09-29
(proof-new) Add proof managers and eager gens to arith

alex-ozdemir:
This commit threads ProofNodeManager, EagerProofGenerator, etc
throughout the arith theory into the appropriate locations. These
objects are currently unused, but will be used in future commits.

Crediting Andy, since he set up some of this.

Co-authored-by: Andrew Reynolds <andrew.j.reynolds@gmail.com>

### DIFF
--- a/src/theory/arith/theory_arith_private.h
+++ b/src/theory/arith/theory_arith_private.h
@@ -31,6 +31,7 @@
 #include "expr/metakind.h"
 #include "expr/node.h"
 #include "expr/node_builder.h"
+#include "expr/proof_generator.h"
 #include "options/arith_options.h"
 #include "smt/logic_exception.h"
 #include "smt_util/boolean_simplification.h"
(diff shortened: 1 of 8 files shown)


src/theory/arith/theory_arith_private.h
```diff
@@ -306,8 +317,9 @@ class TheoryArithPrivate {
 
   /** This is only used by simplex at the moment. */
   context::CDO<Node> d_blackBoxConflict;
-public:
+  context::CDO<std::shared_ptr<ProofNode>> d_blackBoxConflictPf;
```
ajreynol, line 320: Document this.

ajreynol (changes requested):
One documentation request.

src/theory/arith/theory_arith_private.h
alex-ozdemir: Good catch. Done in 23ed0f2.

ajreynol (approved):

PR #5161, opened 2020-09-29
[proof-new] Adds a proof post processor for the Prop Engine

HanielB:
The post processor connects the two proofs in the prop engine: the refutation proof in the SAT solver and the CNF transformation proof in the CNF stream. The proof generators from theory engine in the latter are also expanded during the connection.


ajreynol (approved):
LGTM

PR #5164, opened 2020-09-29
(proof-new) Preprocessing passes use proper interfaces to assertions pipeline

ajreynol:
This PR eliminates all uses of assertions pipeline that are not proper, which two-fold:
(1) The assertion list should never be modified in a custom way (without going through `replace` / `push_back`),
(2) Places where an assertion is "conjoined" to an existing spot in the vector should use `conjoin` instead of `replace`. 
This is required for proper proof generation.

This fixes https://github.com/CVC4/cvc4-projects/issues/75.


src/preprocessing/passes/bv_gauss.cpp
```diff
@@ -759,8 +758,10 @@ PreprocessingPassResult BVGauss::applyInternal(
       NodeManager *nm = NodeManager::currentNM();
       if (ret == BVGauss::Result::NONE)
       {
-        atpp.clear();
-        atpp.push_back(nm->mkConst<bool>(false));
+        assertionsToPreprocess->clear();
+        Node n = NodeManager::currentNM()->mkConst<bool>(false);
```
HanielB, line 762: used cached node manager

src/preprocessing/passes/bv_gauss.cpp
```diff
@@ -759,8 +758,10 @@ PreprocessingPassResult BVGauss::applyInternal(
       NodeManager *nm = NodeManager::currentNM();
```
HanielB, line 758: move this outside the loop

src/preprocessing/passes/global_negate.cpp
```diff
@@ -27,7 +27,8 @@ namespace CVC4 {
 namespace preprocessing {
 namespace passes {
 
-Node GlobalNegate::simplify(std::vector<Node>& assertions, NodeManager* nm)
+Node GlobalNegate::simplify(const std::vector<Node>& assertions,
+                            NodeManager* nm)
```
HanielB, line 31: maybe we wanna drop having the node manager as a parameter? What comes first to mind is not that this is being done to save a call currentNM().

src/preprocessing/passes/quantifier_macros.h
```diff
@@ -64,7 +64,7 @@ class QuantifierMacros : public PreprocessingPass
                        bool reqComplete);
   void addMacro(Node op, Node n, std::vector<Node>& opc);
   void debugMacroDefinition(Node oo, Node n);
-  bool simplify(std::vector<Node>& assertions, bool doRewrite = false);
+  bool simplify(AssertionPipeline* ap, bool doRewrite = false);
```
HanielB, line 67: maybe document? :)

src/preprocessing/passes/unconstrained_simplifier.cpp
```diff
@@ -855,9 +855,15 @@ PreprocessingPassResult UnconstrainedSimplifier::applyInternal(
...
+    for (size_t i = 0, asize = assertions.size(); i < asize; ++i)
     {
-      assertion = Rewriter::rewrite(d_substitutions.apply(assertion));
+      Node a = assertions[i];
+      Node as = Rewriter::rewrite(d_substitutions.apply(a));
+      if (a != as)
+      {
+        // replace the assertion
+        assertionsToPreprocess->replace(i, as);
+      }
```
HanielB, line 866: This test is already done inside the replace method, so no need to be done outside. The same goes to other occurrences throughout the PR.

HanielB (approved):
LGTM, just some minor requests

src/preprocessing/passes/global_negate.cpp
ajreynol: Its also potentially to make it possible to eliminate calls to currentNM() in case we ever decide to make NodeManager a non-singleton.  I'd prefer keeping this for now.

src/preprocessing/passes/unconstrained_simplifier.cpp
ajreynol: I'll change, although some are due to debugging traces.

src/preprocessing/passes/global_negate.cpp
HanielB: I see. OK.

PR #5167, opened 2020-09-29
Add additional ground terms to SyGuS instantiation grammar

mpreiner:
This PR adds options to add additional ground terms to the SyGuS instantiation grammars.


src/options/quantifiers_options.toml
```diff
@@ -1995,3 +1995,39 @@ header = "options/quantifiers_options.h"
...
+  name       = "sygusInstScope"
+  category   = "regular"
+  long       = "sygus-inst-scope=MODE"
+  type       = "SygusInstScope"
+  default    = "INSIDE"
+  help       = "select scope of ground terms"
+  help_mode  = "scope for collecting ground terms for the grammar."
+[[option.mode.INSIDE]]
+  name = "inside"
+  help = "use ground terms inside quantifier bodies only."
```
ajreynol, line 2009: I assume this means that grammar uses ground terms inside the *given* quantifier? Clarify this here.

src/options/quantifiers_options.toml
```diff
@@ -1995,3 +1995,39 @@ header = "options/quantifiers_options.h"
...
+[[option.mode.OUTSIDE]]
+  name = "outside"
+  help = "use ground terms outside of quantifier bodies only."
+[[option.mode.BOTH]]
+  name = "both"
+  help = "combines inside and outside."
+
+[[option]]
+  name       = "sygusInstGranularity"
```
ajreynol, line 2018: This option name is a bit too general sounding to me, I think `sygusInstTermSel` may be better (note this has some similarities with this option: https://github.com/CVC4/CVC4/blob/master/src/options/quantifiers_options.toml#L368).

src/theory/quantifiers/sygus_inst.cpp
```diff
@@ -138,6 +138,100 @@ bool SygusInst::checkCompleteFor(Node q)
...
+    cur = visit.back();
+    visit.pop_back();
+
+    if (cache.find(cur) != cache.end())
+    {
+      continue;
+    }
+    cache.insert(cur);
+
+    if (expr::hasBoundVar(cur) || cur.getType().isBoolean())
```
ajreynol, line 175: Why is this check for Boolean type necessary?  Its possible that we could be doing QBF and have a large ground term of type Boolean.

ajreynol (changes requested):

src/options/quantifiers_options.toml
mpreiner: done

src/options/quantifiers_options.toml
mpreiner: done

src/theory/quantifiers/sygus_inst.cpp
mpreiner: refactored as discussed off-line.

src/theory/quantifiers/sygus_inst.cpp
```diff
@@ -190,6 +381,13 @@ void SygusInst::preRegisterQuantifier(Node q)
   addCeLemma(q);
 }
 
+void SygusInst::ppNotifyAssertions(const std::vector<Node>& assertions)
+{
+  d_notified_assertions.clear();
```
ajreynol, line 386: This actually needs to be a user-context dependent list, where we don't clear and only add here.  This will ensure sygus-inst does the right thing in incremental mode.

ajreynol (changes requested):

src/theory/quantifiers/sygus_inst.cpp
ajreynol: (Which also assumes sygus will eventually be available in incremental mode...) :)

mpreiner: done, it's not a CDHashSet.

src/theory/quantifiers/sygus_inst.h
```diff
@@ -124,6 +127,15 @@ class SygusInst : public QuantifiersModule
   /* Indicates whether a counterexample lemma was added for a quantified
    * formula in the current context. */
   context::CDHashSet<Node, NodeHashFunction> d_ce_lemma_added;
+
+  /* Set of global ground terms in assertions (outside of quantifiers). */
+  std::unordered_map<TypeNode,
```
ajreynol, line 132: This also needs to be user-context dependent.

mpreiner: done

ajreynol (approved):

PR #5168, opened 2020-09-30
Remove too strict assertion to allow for approximate models

### LINKED ISSUE #5113: Fatal failure at src/theory/arith/theory_arith.cpp:161
Hi, for the following formula
~~~~
(set-logic ALL_SUPPORTED)
(declare-const i5 Int)
(declare-const r31 Real)
(assert (! (= (exp 0.4351869) r31) :named IP_8))
(check-sat-assuming ((! (or (>= (to_int 29963.48807) (to_int (to_real i5)))) :named IP_69) (! (or (>= (to_int 29963.48807) (to_int (to_real i5))) false) :named IP_74)))
~~~~
cvc4 71ab2d154b2f
~~~~
cvc4 xx.smt2 --check-models
Fatal failure within virtual bool CVC4::theory::arith::TheoryArith::collectModelValues(CVC4::theory::TheoryModel*, const std::set<CVC4::NodeTemplate<true> >&) at /CVC4/src/theory/arith/theory_arith.cpp:161
Check failure

 p.second.isConst()

Aborted
~~~~

nafur:
This PR simply removes an assertion that would trigger whenever the arithmetic theory asserts a model that contains something else than a constant. This can be the case with witness terms.
In offline discussion we concluded that this discussion was overly strict. Note that these examples may still hit an error during model validation (`Cannot run check-model on a model with approximate values.`).
This PR also fixes a debug output I found during debugging.

Fixes #5113.

### DIFF
--- a/src/smt/smt_engine.cpp
+++ b/src/smt/smt_engine.cpp
@@ -1610,7 +1610,8 @@ void SmtEngine::checkModel(bool hardFailure) {
   for(size_t k = 0; k < m->getNumCommands(); ++k) {
     const DeclareFunctionNodeCommand* c =
         dynamic_cast<const DeclareFunctionNodeCommand*>(m->getCommand(k));
-    Notice() << "SmtEngine::checkModel(): model command " << k << " : " << m->getCommand(k)->toString() << endl;
+    Notice() << "SmtEngine::checkModel(): model command " << k << " : "
+             << m->getCommand(k)->toString() << endl;
     if(c == NULL) {
       // we don't care about DECLARE-DATATYPES, DECLARE-SORT, ...
       Notice() << "SmtEngine::checkModel(): skipping..." << endl;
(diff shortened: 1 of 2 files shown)


src/smt/smt_engine.cpp
```diff
@@ -1610,7 +1610,7 @@ void SmtEngine::checkModel(bool hardFailure) {
   for(size_t k = 0; k < m->getNumCommands(); ++k) {
     const DeclareFunctionNodeCommand* c =
         dynamic_cast<const DeclareFunctionNodeCommand*>(m->getCommand(k));
-    Notice() << "SmtEngine::checkModel(): model command " << k << " : " << m->getCommand(k) << endl;
+    Notice() << "SmtEngine::checkModel(): model command " << k << " : " << m->getCommand(k)->toString() << endl;
```
ajreynol, line 1613: Format.

ajreynol (approved):
Remember to format.

PR #5172, opened 2020-09-30
Make SygusSolver use sygus attributes directly

ajreynol:
Previously, the SmtEngine was using an interface through TheoryEngine to set user attributes to indicate that e.g. a synth-fun is associated with a given grammar.  Since SmtEngine and its members are internal now, this can simply be done directly via attributes.  This makes it so that SygusSolver does not depend on TheoryEngine nor does it require the TheoryEngine to be initialized at the time a synth-fun is created.

### DIFF
--- a/src/smt/sygus_solver.cpp
+++ b/src/smt/sygus_solver.cpp
@@ -63,8 +63,6 @@ void SygusSolver::declareSynthFun(const std::string& id,
 {
   Trace("smt") << "SygusSolver::declareSynthFun: " << id << "\n";
   NodeManager* nm = NodeManager::currentNM();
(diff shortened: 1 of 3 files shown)


src/smt/sygus_solver.cpp
```diff
@@ -68,17 +69,17 @@ void SygusSolver::declareSynthFun(const std::string& id,
   if (!vars.empty())
```
abdoo8080, line 69: asdf

abdoo8080, line 69: Could you please remove the assertion in line 67

abdoo8080 (approved):
The changes fix the problem I am facing with synth-fun. LGTM!

PR #5177, opened 2020-10-01
Allow to use the initial assignment for CAD

nafur:
While the CAD subsolver already provided for a way to use the linear model to seed the model search, it was not actually used yet.
This PR now does use it, though it is disabled by a Boolean flag.


nafur:
Benchmarks show that it tends to improve on "not so good" configurations, but slightly worsens performance on the best configurations. This is why it is disabled right now, I plan to tweak the mechanism in the hope to improve here...

src/theory/arith/nl/cad/cdcac.h
```diff
@@ -183,6 +183,12 @@ class CDCAC
 
   /** The linear assignment used as an initial guess. */
   std::vector<poly::Value> d_initialAssignment;
+
+  /**
+   * Decides whether we use the model from the linear solver as an initial
+   * assignment to guide the sampling process.
+   */
+  static constexpr bool use_initial_assignment = false;
```
ajreynol, line 191: Should we just make this an option?

nafur: Done.

ajreynol (approved):

Issue #4631, opened 2020-06-18
(simplification=none) Refutational soundness bug on set logic formula (sets subtyping)


wintered:
Hi, 
the following trace exhibits a refutational soundness bug in CVC4.

```
[625] % cvc4 small.smt2
sat
[626] % cvc4 --simplification=none small.smt2
unsat
[627] % 
[627] % cat small.smt2
(set-logic ALL)
(declare-fun a () (Set Real))
(declare-fun b () (Set Int))
(assert (distinct b (intersection a (singleton 0.5))))
(assert (= b (as emptyset (Set Real))))
(check-sat)
[628] %
```
OS: Ubuntu 18.04
Commit: c7522585

ajreynol:
This involves set subtyping (constraints that compare (Set Int) with (Set Real)).  I believe we should eliminate this feature, which would make this benchmark ill-typed.  I will look into this more.

Issue #4507, opened 2020-05-21
(check-models) Error Satifsfying Assertion with formula containing member (sets subtypes)


wolffdy:
Hi, I'm getting an internal error with the following formula:
[bug.smt2.gz](https://github.com/CVC4/CVC4/files/4660826/bug.smt2.gz)
```
% cat bug.smt2                
(declare-const a Real)
(declare-const b Bool)
(declare-const c Bool)
(declare-fun d () (Set Int))
(declare-fun e () (Set Real))
(declare-fun f () (Set Real))
(assert (member a f))
(assert (= b (= d e)))
(assert (= c (= d f)))
(assert (and  (= c true)))
(check-sat)
% cvc4 --check-models bug.smt2
cvc4 --check-models bug.smt2                            
...
Fatal failure within void CVC4::SmtEngine::checkModel(bool) at /home/dylan/git/constant-swap/scripts/.solvers/cvc4/src/smt/smt_engine.cpp:2702
Internal error detectedSmtEngine::checkModel(): ERRORS SATISFYING ASSERTIONS WITH MODEL:
assertion:     (= b (= (as d (Set Real)) e))
simplifies to: false
expected `true'.
Run with `--check-models -v' for additional diagnostics.
[1]    1197635 abort (core dumped)  cvc4 --check-models bug.smt2
```

**Command line arguments**: --check-models
**CVC4 version/commit**: 3b50dfe
**Operating system**: Kernel: 4.19.121-1-MANJARO 
**`configure.sh` options**: --assertions
**`configure.sh` output**: happy to provide upon request

mudathirmahgoub:
Fixed by removing subtyping for sets ([5205](https://github.com/CVC4/CVC4/pull/5205)).

Issue #4502, opened 2020-05-19
(check-models) Error Satisfying Assertions (subtypes with parametric datatypes)


wolffdy:
Hi, I'm getting an internal error with the following formula:
[bug.smt2.gz](https://github.com/CVC4/CVC4/files/4649735/bug.smt2.gz)
```
% cat bug.smt2                
(declare-const a Int)
(declare-const b Real)
(declare-datatypes ( (c 2) ) ( (par ( d h ) ( (i (e d) (f h)))) ))
(declare-fun g () (c Int Real))
(assert (= (= g (i a b)) true))
(check-sat)

% cvc4 --check-models bug.smt2
bug.smt2:1.13: No set-logic command was given before this point.
bug.smt2:1.13: CVC4 will make all theories available.
bug.smt2:1.13: Consider setting a stricter logic for (likely) better performance.
bug.smt2:1.13: To suppress this warning in the future use (set-logic ALL).
Fatal failure within void CVC4::SmtEngine::checkModel(bool) at /home/dylan/git/constant-swap/scripts/.solvers/cvc4/src/smt/smt_engine.cpp:2702
Internal error detectedSmtEngine::checkModel(): ERRORS SATISFYING ASSERTIONS WITH MODEL:
assertion:     (= (= g (i a b)) true)
simplifies to: false
expected `true'.
Run with `--check-models -v' for additional diagnostics.
[1]    55491 abort (core dumped)  cvc4 --check-models bug.smt2
```

**Command line arguments**: --check-models
**CVC4 version/commit**: 3b50dfe
**Operating system**: Kernel: 4.19.121-1-MANJARO 
**`configure.sh` options**: --assertions
**`configure.sh` output**: happy to provide upon request

PR #5179, opened 2020-10-01
Remove subtyping for sets theory

mudathirmahgoub:
This PR removes subtyping for sets theory due to issues like #4502, #4507 and #4631.
Changes:
- Add SingletonOp for singletons to specify the type of the single element in the set.
- Add mkSingleton to the solver to enable the user to pass the sort of the single element.
- Update smt and cvc parsers to use mkSingleton when the kind is SINGLETON


src/parser/cvc/Cvc.g
```diff
@@ -598,6 +598,8 @@ using namespace CVC4::parser;
 #define SOLVER PARSER_STATE->getSolver()
 #undef MK_TERM
 #define MK_TERM SOLVER->mkTerm
+#undef MK_SINGLETON
+#define MK_SINGLETON SOLVER->mkSingleton
```
ajreynol, line 602: Since this is only used twice, I'd prefer if it were inlined.

src/theory/sep/theory_sep.cpp
```diff
@@ -1341,7 +1342,8 @@ Node TheorySep::mkUnion( TypeNode tn, std::vector< Node >& locs ) {
     for( unsigned i=0; i<locs.size(); i++ ){
       Node s = locs[i];
       Assert(!s.isNull());
-      s = NodeManager::currentNM()->mkNode( kind::SINGLETON, s );
+      Node op = NodeManager::currentNM()->mkConst(SingletonOp(s.getType()));
```
ajreynol, line 1345: We should add a mkSingleton utility to NodeManager for doing this.

ajreynol (changes requested):
This looks great, thanks for doing this.  I have a few suggestions.

src/parser/cvc/Cvc.g
mudathirmahgoub: Done

src/theory/sep/theory_sep.cpp
mudathirmahgoub: done

src/theory/quantifiers/sygus/sygus_grammar_cons.cpp
```diff
@@ -957,7 +957,13 @@ void CegGrammarConstructor::mkSygusDefaultGrammar(
       Trace("sygus-grammar-def") << "...add for singleton" << std::endl;
       std::vector<TypeNode> cargsSingleton;
       cargsSingleton.push_back(unresElemType);
-      sdts[i].addConstructor(SINGLETON, cargsSingleton);
+
+      // lambda x . (singleton (singleton_op T) x) where T = x.getType()
+      Node x = nm->mkBoundVar(etype);
+      Node vars = nm->mkNode(BOUND_VAR_LIST, x);
+      Node singleton = nm->mkSingleton(x.getType(), x);
```
ajreynol, line 964:
```suggestion
      Node singleton = nm->mkSingleton(etype, x);
```

src/theory/sep/theory_sep.cpp
```diff
@@ -412,7 +412,7 @@ void TheorySep::reduceFact(TNode atom, bool polarity, TNode fact)
     }
     else if (satom.getKind() == SEP_PTO)
     {
-      Node ss = nm->mkNode(SINGLETON, satom[0]);
+      Node ss = nm->mkSingleton(satom[0].getType(), satom[0]);
```
ajreynol, line 415: Note these uses of type are not safe, e.g. if satom[0] is intended to be a Real but is 1.0.  I will address this after this PR, as this may require more refactoring to TheorySep to know which type to use.

src/theory/sets/normal_form.h
```diff
@@ -45,10 +45,11 @@ class NormalForm {
     else
     {
       ElementsIterator it = elements.begin();
-      Node cur = nm->mkNode(kind::SINGLETON, *it);
+      Node cur = nm->mkSingleton(it->getType(), *it);
```
ajreynol, line 48: I think you should pass the type to use as an argument of this function.  It is definitely not safe to use getType here, e.g. think of set that mixes 1.0 and 1.5.

src/theory/sets/theory_sets_private.cpp
```diff
@@ -1145,7 +1145,7 @@ bool TheorySetsPrivate::collectModelValues(TheoryModel* m,
         {
           for (const std::pair<const Node, Node>& itmm : emems)
           {
-            Node t = nm->mkNode(kind::SINGLETON, itmm.first);
+            Node t = nm->mkSingleton(itmm.first.getType(), itmm.first);
```
ajreynol, line 1148: You should use the element type of the type of eqc here. Otherwise this may break if a value is 1.0.

src/theory/sets/theory_sets_rewriter.cpp
```diff
@@ -488,12 +488,12 @@ RewriteResponse TheorySetsRewriter::preRewrite(TNode node) {
   }
   else if (k == kind::INSERT)
   {
-    Node insertedElements = nm->mkNode(kind::SINGLETON, node[0]);
+    Node insertedElements = nm->mkSingleton(node[0].getType(), node[0]);
```
ajreynol, line 491: This is (barely) safe I think, although it is fairly sketchy, think `(insert 1.0 1.5)`.  I believe this code works technically due to the fact this is at preRewrite before 1.0 from the parser (which is `(/ 1.0 1.0)` or `(to_real 1.0)`) is rewritten to 1.

You should compute the type you are going to use upfront (node[0].getType()) and then use assertions for each `node[i]` below that their types are subtypes of node[0].getType().

src/theory/sets/theory_sets_type_enumerator.cpp
```diff
@@ -89,7 +89,7 @@ SetEnumerator& SetEnumerator::operator++()
     // get a new element and return it as a singleton set
     Node element = *d_elementEnumerator;
     d_elementsSoFar.push_back(element);
-    d_currentSet = d_nodeManager->mkNode(kind::SINGLETON, element);
+    d_currentSet = d_nodeManager->mkSingleton(element.getType(), element);
```
ajreynol, line 92: Can you find the proper sort to pass here? It should be available in the enumerator.

ajreynol (changes requested):
The structure of the PR looks great.

However, there are many places where the construction of singleton is not safe.  As a rule, you should always try to find a more accurate source of the intended type of singleton to ensure it has the correct type instead of taking the type of its argument (which is prone to subtyping issues).  I've given some comments about specific instances below.

src/theory/quantifiers/sygus/sygus_grammar_cons.cpp
mudathirmahgoub: Done. In this particular case, x is declared with this type. `etype` is shorter

src/theory/sets/normal_form.h
mudathirmahgoub: Done. The function has `TypeNode setType` which we can use to retrieve the element type.

src/theory/sets/theory_sets_private.cpp
mudathirmahgoub: Done. It is nice here we don't have to sets of different types in the same class.

src/theory/sets/theory_sets_type_enumerator.cpp
mudathirmahgoub: Done.

src/theory/sep/theory_sep.cpp
mudathirmahgoub: Added issue https://github.com/CVC4/cvc4-projects/issues/230

src/theory/sets/theory_sets_rewriter.cpp
mudathirmahgoub: Done. Funny I wrote `n[0].getType()` inside the loop, but then I looked at the type rule for insert, which uses type equality instead of subtyping. So I thought it is safe to use `n[i].getType()` here. 
Should we enable subtyping for the insert operator?

src/theory/sets/theory_sets_private.cpp
```diff
@@ -1145,7 +1145,7 @@ bool TheorySetsPrivate::collectModelValues(TheoryModel* m,
         {
           for (const std::pair<const Node, Node>& itmm : emems)
           {
-            Node t = nm->mkSingleton(itmm.first.getType(), itmm.first);
+            Node t = nm->mkSingleton(eqc.getType(), itmm.first);
```
ajreynol, line 1148: Shouldnt this be the element type of eqc.getType()?

ajreynol (changes requested):
One question about a type left.

Also, can you try the failing benchmark again?  Maybe it works now after the fixes?

src/theory/sets/theory_sets_private.cpp
mudathirmahgoub: Right. This is embarrassing !!

mudathirmahgoub: Fixed

src/theory/sep/theory_sep.cpp
```diff
@@ -1341,7 +1342,7 @@ Node TheorySep::mkUnion( TypeNode tn, std::vector< Node >& locs ) {
     for( unsigned i=0; i<locs.size(); i++ ){
       Node s = locs[i];
       Assert(!s.isNull());
-      s = NodeManager::currentNM()->mkNode( kind::SINGLETON, s );
+      s = NodeManager::currentNM()->mkSingleton(tn, s);
```
mudathirmahgoub, line 1345: Here I used tn which is an argument to `mkUnion` and the type of the elements. This function returns either an empty set, or a union of sets starting from a singleton. Perhaps it is much simpler to start from an empty set (which stores the type).

src/theory/sets/cardinality_extension.cpp
```diff
@@ -1028,7 +1028,7 @@ void CardinalityExtension::mkModelValueElementsFor(
           // the current members of this finite type.
 
           Node slack = nm->mkSkolem("slack", elementType);
-          Node singleton = nm->mkNode(SINGLETON, slack);
+          Node singleton = nm->mkSingleton(elementType, slack);
```
mudathirmahgoub, line 1031: This is safe because `elementType` is extracted from the set eqc which contains the constructed slack element.

src/theory/sets/cardinality_extension.cpp
```diff
@@ -1037,7 +1037,7 @@ void CardinalityExtension::mkModelValueElementsFor(
         else
         {
           els.push_back(
-              nm->mkNode(SINGLETON, nm->mkSkolem("msde", elementType)));
+              nm->mkSingleton(elementType, nm->mkSkolem("msde", elementType)));
```
mudathirmahgoub, line 1040: ditto

src/theory/sets/normal_form.h
```diff
@@ -44,11 +44,13 @@ class NormalForm {
     }
     else
     {
+      TypeNode elementType = setType.getSetElementType();
       ElementsIterator it = elements.begin();
-      Node cur = nm->mkNode(kind::SINGLETON, *it);
+      Node cur = nm->mkSingleton(elementType, *it);
```
mudathirmahgoub, line 49: This is safe here because `elementType` is extracted from `setType` that is passed as an argument to `elementsToSet` and is used to construct a set the are no elements.

src/theory/sets/normal_form.h
```diff
@@ -44,11 +44,13 @@ class NormalForm {
...
       while (++it != elements.end())
       {
-        cur = nm->mkNode(kind::UNION, nm->mkNode(kind::SINGLETON, *it), cur);
+        Node singleton = nm->mkSingleton(elementType, *it);
```
mudathirmahgoub, line 52: ditto

src/theory/sets/theory_sets_private.cpp
```diff
@@ -1143,9 +1143,10 @@ bool TheorySetsPrivate::collectModelValues(TheoryModel* m,
         const std::map<Node, Node>& emems = d_state.getMembers(eqc);
         if (!emems.empty())
         {
+          TypeNode elementType = eqc.getType().getSetElementType();
           for (const std::pair<const Node, Node>& itmm : emems)
           {
-            Node t = nm->mkNode(kind::SINGLETON, itmm.first);
+            Node t = nm->mkSingleton(elementType, itmm.first);
```
mudathirmahgoub, line 1149: This is safe because we are extracted the element type from the set eqc which contains the element itmm.first.

src/theory/sets/theory_sets_private.cpp
```diff
@@ -1358,7 +1359,7 @@ TrustNode TheorySetsPrivate::expandIsSingletonOperator(const Node& node)
 
   TypeNode setType = set.getType();
   Node boundVar = nm->mkBoundVar(setType.getSetElementType());
-  Node singleton = nm->mkNode(kind::SINGLETON, boundVar);
+  Node singleton = nm->mkSingleton(setType.getSetElementType(), boundVar);
```
mudathirmahgoub, line 1362: This is safe because we are extracting the element type from the set that is compared with the constructed singleton. `boundVar.getType()` is safe too.

src/theory/sets/theory_sets_rels.cpp
```diff
@@ -247,12 +247,13 @@ void TheorySetsRels::check(Theory::Effort level)
...
           for (unsigned i = 0, tlen = erType.getTupleLength(); i < tlen; i++)
           {
-            Node element = RelsUtils::nthElementOfTuple( eqc_node, i );
-
-            if( !element.isConst() ) {
-              makeSharedTerm( element );
+            Node element = RelsUtils::nthElementOfTuple(eqc_node, i);
+            if (!element.isConst())
+            {
+              makeSharedTerm(element, tupleTypes[i]);
```
mudathirmahgoub, line 256: I guess this is safe if `erType.getTupleTypes();` in line 250 above is safe.

src/theory/sets/theory_sets_rels.cpp
```diff
@@ -863,7 +864,7 @@ void TheorySetsRels::check(Theory::Effort level)
     sendInfer(fact, reason, "JOIN-Split-1");
     fact = NodeManager::currentNM()->mkNode(kind::MEMBER, mem2, join_rel[1]);
     sendInfer(fact, reason, "JOIN-Split-2");
-    makeSharedTerm( shared_x );
+    makeSharedTerm(shared_x, shared_type);
```
mudathirmahgoub, line 867: This is safe because `shared_type` (which is the type of `shared_x`) is extracted from a set type in line [824](https://github.com/mudathirmahgoub/CVC4/blob/removeSubtyping/src/theory/sets/theory_sets_rels.cpp#L824).

src/theory/sets/theory_sets_rels.cpp
```diff
@@ -1153,8 +1154,9 @@ void TheorySetsRels::check(Theory::Effort level)
       }
       return equal;
     } else if(!a.getType().isBoolean()){
-      makeSharedTerm(a);
-      makeSharedTerm(b);
+      // TODO(project##230): Find a safe type for the singleton operator
```
mudathirmahgoub, line 1157: Here I am not certain using `a.getType()` and `b.getType()`  below is safe. It seems this requiring refactoring `TheorySetsRels::areEqual( Node a, Node b )` to have the types for a and b.

src/theory/sets/theory_sets_rels.cpp
```diff
@@ -1183,14 +1185,15 @@ void TheorySetsRels::check(Theory::Effort level)
...
+  void TheorySetsRels::makeSharedTerm(Node n, TypeNode t)
+  {
     if (d_shared_terms.find(n) != d_shared_terms.end())
     {
       return;
     }
     Trace("rels-share") << " [sets-rels] making shared term " << n << std::endl;
     // force a proxy lemma to be sent for the singleton containing n
-    Node ss = NodeManager::currentNM()->mkNode(SINGLETON, n);
+    Node ss = NodeManager::currentNM()->mkSingleton(t, n);
```
mudathirmahgoub, line 1196: It is not possible to figure out if this singleton is safe without knowing the type. So the function is updated to receive the type as an argument. See the justification for the types at the locations where `makeSharedTerm(Node n, TypeNode t)` is called.

src/theory/sets/theory_sets_rels.cpp
```diff
@@ -1216,9 +1219,11 @@ void TheorySetsRels::check(Theory::Effort level)
       Trace("rels-debug") << "[Theory::Rels] Reduce tuple var: " << n[0] << " to a concrete one " << " node = " << n << std::endl;
       std::vector<Node> tuple_elements;
       tuple_elements.push_back((n[0].getType().getDType())[0].getConstructor());
-      for(unsigned int i = 0; i < n[0].getType().getTupleLength(); i++) {
+      std::vector<TypeNode> tupleTypes = n[0].getType().getTupleTypes();
+      for (unsigned int i = 0; i < n[0].getType().getTupleLength(); i++)
+      {
         Node element = RelsUtils::nthElementOfTuple(n[0], i);
-        makeSharedTerm(element);
+        makeSharedTerm(element, tupleTypes[i]);
```
mudathirmahgoub, line 1226: This is safe if `n[0].getType().getTupleTypes()` in line 1222 above is safe.

src/theory/sets/theory_sets_rewriter.cpp
```diff
@@ -488,12 +488,14 @@ RewriteResponse TheorySetsRewriter::preRewrite(TNode node) {
...
-    for(size_t i = 1; i < setNodeIndex; ++i) {
-      insertedElements = nm->mkNode(kind::UNION, 
-                                    insertedElements,
-                                    nm->mkNode(kind::SINGLETON, node[i]));
+    TypeNode elementType = node[setNodeIndex].getType().getSetElementType();
+    Node insertedElements = nm->mkSingleton(elementType, node[0]);
+
+    for (size_t i = 1; i < setNodeIndex; ++i)
+    {
+      Node singleton = nm->mkSingleton(elementType, node[i]);
```
mudathirmahgoub, line 497: This is safe because `elementType` is extracted from the last node for `INSERT` which is a set.

src/theory/sets/theory_sets_type_enumerator.cpp
```diff
@@ -89,7 +89,8 @@ SetEnumerator& SetEnumerator::operator++()
     // get a new element and return it as a singleton set
     Node element = *d_elementEnumerator;
     d_elementsSoFar.push_back(element);
-    d_currentSet = d_nodeManager->mkNode(kind::SINGLETON, element);
+    TypeNode elementType = d_elementEnumerator.getType();
+    d_currentSet = d_nodeManager->mkSingleton(elementType, element);
```
mudathirmahgoub, line 93: This is safe because we are extracting the element type from the element enumerator.

src/theory/sets/theory_sets_rels.cpp
ajreynol: I dont believe this is critical (the type shouldnt matter here).

But regardless, this requires a lot of refactoring :)

test/unit/theory/theory_bags_type_rules_black.h
```diff
@@ -69,8 +69,8 @@ class BagsTypeRuleBlack : public CxxTest::TestSuite
     Node node = d_nm->mkConst(Rational(10));
 
     // node of type Int is not compatible with bag of type (Bag String)
-    TS_ASSERT_THROWS(d_nm->mkNode(BAG_COUNT, node, bag),
-                     TypeCheckingExceptionPrivate&);
+//    TS_ASSERT_THROWS(d_nm->mkNode(BAG_COUNT, node, bag),
```
ajreynol, line 72: Why this change?

ajreynol (changes requested):

test/unit/theory/theory_bags_type_rules_black.h
mudathirmahgoub: It failed at some point with `make units` which was weird because it is irrelevant to the PR. The assertion passes now. 
Perhaps I modified the wrong branch :)

ajreynol (approved):

PR #5180, opened 2020-10-01
Minor simplifications to substitution map class

ajreynol:
This class is an important utility in preprocessing, which we are adding proof support for.  This simplifies the interface of this class with regards to unused interfaces for clarity.


HanielB (approved):
LGTM, but please say at a high level what and why you're deleting in the description of the commit when merging.

PR #5181, opened 2020-10-01
Standardization of Theory

ajreynol:
This cleans up various interfaces of Theory now that all theories have been updated to the new standard. This includes making check non-virtual, standardizing when trigger terms are added to equality engines, and simplifications for collectModelInfo.


4tXJ7f (approved):
Awesome, that's a nice milestone.

PR #5182, opened 2020-10-01
Remove duplicate declarations in th_bv.plf

alex-ozdemir:
I think this is dead code now, but still good to fix.


yoni206 (approved):
LGTM. one question: how did u know which version of concat to delete? (in case of some legacy usage of old proofs, does CVC4 generate the more complex concat?)

PR #5183, opened 2020-10-01
Remove duplicate declarations in th_bv.plf

alex-ozdemir:
I think this is dead code now, but still good to fix.


alex-ozdemir:
Sorry for the repeat.

In answer to question, yoni: I knew the second `concat` was the right one because LFSC implements shadowing :), so the first one just got overwritten.

yoni206 (approved):

PR #5184, opened 2020-10-01
Decouple modules from TheoryArithPrivate

ajreynol:
This decouples the arithmetic rewriter, the operator eliminator and the non-linear extension from TheoryArithPrivate and performs some minor cleanup.

This is in preparation for further refactoring of (lazy) arithmetic operator elimination.


src/theory/arith/theory_arith_private.cpp
```diff
@@ -3641,31 +3586,23 @@ void TheoryArithPrivate::postCheck(Theory::Effort effortLevel)
...
   if(Debug.isOn("arith::print_model")) {
     debugPrintModel(Debug("arith::print_model"));
   }
   Debug("arith") << "TheoryArithPrivate::check end" << std::endl;
+  return emmittedConflictOrSplit;
+}
+
+bool TheoryArithPrivate::isNonlinearIncomplete() const
+{
+  return d_nlIncomplete;
```
nafur, line 3605: Why does this piece of information still live in `TheoryArithPrivate`?

src/theory/arith/theory_arith.cpp
```diff
@@ -101,15 +99,55 @@ void TheoryArith::preRegisterTerm(TNode n)
...
 
 TrustNode TheoryArith::ppRewrite(TNode atom)
 {
   CodeTimer timer(d_ppRewriteTimer, /* allow_reentrant = */ true);
-  return d_internal->ppRewrite(atom);
+  Debug("arith::preprocess") << "arith::preprocess() : " << atom << endl;
+
+  if (options::arithRewriteEq())
+  {
+    if (atom.getKind() == kind::EQUAL && atom[0].getType().isReal())
```
nafur, line 115: Can't we do this for `Int` as well?

nafur (commented):
I'd suggest to move the `d_nlIncomplete` flag to `TheoryArith`...

src/theory/arith/theory_arith_private.cpp
ajreynol: This flag indicates whether `TheoryArithPrivate` is incomplete when used in isolation.  In other words, this tracks if the linear solver encountered any non-linear facts.

This is required when the non-linear extension is disabled.  This would hypothetically catch cases where non-linear arithmetic was accidentally introduced in a linear logic or in any other case where the non-linear extension is disabled.

src/theory/arith/theory_arith.cpp
ajreynol: Int is a subtype of Real, so this already handles Int.

src/theory/arith/theory_arith_private.cpp
nafur: Ah, I see. The current name seems to suggest that the (computation of the) nonlinear extension was incomplete. Can we rename the flag? (like `foundNonlinearity`?)

nafur (approved):
LGTM.

PR #5186, opened 2020-10-02
Add operators bag.from_set, bag.to_set to the theory of bags

mudathirmahgoub:


mudathirmahgoub:
This PR needs to be updated after the following PR is merged 
https://github.com/CVC4/CVC4/pull/5179

src/theory/bags/bags_rewriter.cpp
```diff
@@ -429,6 +430,19 @@ BagsRewriteResponse BagsRewriter::rewriteIsSingleton(const TNode& n) const
...
 
+BagsRewriteResponse BagsRewriter::rewriteFromSet(const TNode& n) const
+{
+  Assert(n.getKind() == BAG_FROM_SET);
+  if (n[0].getKind() == SINGLETON)
+  {
+    // (bag.from_set (singleton (singleton_op Int) x)) = (mkBag x 1)
+    Node one = d_nm->mkConst(Rational(1));
+    Node bag = d_nm->mkNode(MK_BAG, n[0][0], one);
+    return BagsRewriteResponse(bag, Rewrite::IS_SINGLETON_MK_BAG);
```
ajreynol, line 441: Update Rewrite value.

src/theory/bags/bags_rewriter.cpp
```diff
@@ -71,6 +71,7 @@ RewriteResponse BagsRewriter::postRewrite(TNode n)
       case BAG_CHOOSE: response = rewriteChoose(n); break;
       case BAG_CARD: response = rewriteCard(n); break;
       case BAG_IS_SINGLETON: response = rewriteIsSingleton(n); break;
+      case BAG_FROM_SET: response = rewriteFromSet(n); break;
```
ajreynol, line 74: What about BAG_TO_SET?  You should add a rewrite function and I believe there is a dual rewrite rule (bag.to_set (mk_bag x n)) = (singleton x).

ajreynol (changes requested):

src/theory/bags/bags_rewriter.cpp
mudathirmahgoub: Done. I guess reveals some `suspicious` operation :)

src/theory/bags/bags_rewriter.cpp
mudathirmahgoub: I thought about this. But if `n` is a variable, then it can be assigned a negative or zero multiplicity later, which would make the bag empty. As a result, bag.to_set` returns the emptyset. If `n` is a positive constant, then, this is not in a normal form, and  should be handled by `NormalForm::evaluate` which would return `(singleton (singleton_op t) x)`.

ajreynol: Ah right. Although I think there is still a case where the element is not constant and the multiplicity is a postitive constant, where you'd need e.g.
`(bag.to_set (mkBag x 4)) ---> (singleton x)`.

mudathirmahgoub: Done. I missed this case when the element is a variable. Thanks for pointing this out.

ajreynol (approved):
LGTM

PR #5188, opened 2020-10-02
Add arithmetic preprocess module

ajreynol:
This class serves a similar purpose to the strings preprocess module (https://github.com/CVC4/CVC4/blob/master/src/theory/strings/theory_strings_preprocess.h).

It can potentially be used for reducing extended arithmetic operators on demand via lemmas. 

This PR does not change the current behavior, but generalizes the use of operator elimination in TheoryArith to make this possible.


src/theory/arith/arith_preprocess.cpp
```diff
@@ -0,0 +1,67 @@
...
+    return (*it).second;
+  }
+  TrustNode tn = eliminate(atom);
+  if (tn.isNull())
+  {
+    // did not reduce
+    d_reduced[atom] = false;
+    return false;
+  }
+  // tn is of kind REWRITE, turn this into a LEMMA here
```
nafur, line 45: If it can only be `REWRITE`, we might want to assert this here. Why can't it be a conflict?

src/theory/arith/arith_preprocess.h
```diff
@@ -0,0 +1,80 @@
...
+  TrustNode eliminate(TNode n);
+  /**
+   * Reduce assertion. This sends a lemma via the inference manager if atom
+   * contains any extended operators. When applicable, the lemma is of the form:
+   *   atom == d_opElim.eliminate(atom)
+   * This method returns true if a lemma of the above form was added to
+   * the inference manager. Note this returns true even if this lemma was added
+   * on a previous call.
+   */
+  bool reduceAssertion(TNode atom);
```
nafur, line 59: I'm a bit confused on how this is supposed to be used. Shouldn't the user of this method get to know about the reduced node? (i.e. return a `Node` instead of a `bool`)

nafur (changes requested):
I'm not sure about the interface of the new `reduceAssertion`. Can you explain a bit how this is supposed to be used?

src/theory/arith/arith_preprocess.h
ajreynol: I don't see a use case for knowing the Node.  This module has access to the inference manager so it can send the lemma itself, where notice that the reduced form of atom will come back as an asserted literal.

Notice that there is a distinction between this module and the operator elimination module, if you did want to know the reduced form of a Node, you could use that module.

src/theory/arith/arith_preprocess.cpp
ajreynol: It cant be a conflict because the lemma is of the form (= atom REDUCE(atom)) where atom does not simplify to a constant.

I'll add the assertion.

src/theory/arith/arith_preprocess.h
nafur: I see.

nafur (approved):
LGTM.

PR #5191, opened 2020-10-02
sygus-inst: Add more special BV values.

aniemetz:


ajreynol:
May need to test SyGuS problems that rely on BV default grammars.  Either that or add an option that guards this.

aniemetz:
Has a negative impact on the `cegist-cav18` benchmarks, will add only for sygus inst.

mpreiner (approved):

ajreynol (approved):
LGTM

PR #5193, opened 2020-10-03
(proof-new) Add interface for trusted substitution and update ppAssert

ajreynol:
The current work on proof-new involves proofs for preprocessing. The biggest issue currently is that our preprocessing passes do not track proofs for substitutions.

This adds a "trusted substitution" class with is a layer on substitution map.  The proof aspect of this class is not yet implemented, this PR just adds its interface.

This also updates Theory::ppAssert to take a TrustSubstitutionMap instead of a SubstitutionMap, since eventually we will require proofs to be provided for substitutions that are added to this map.


src/theory/trust_substitutions.h
```diff
@@ -0,0 +1,67 @@
...
+  SubstitutionMap& get();
+  /**
+   * Add substitution x -> t, where pg can provide a closed proof of (= x t)
+   * in the remainder of this user context.
+   */
+  void addSubstitution(TNode x, TNode t, ProofGenerator* pg = nullptr);
+  /**
+   * Add substitution x -> t, which was derived from the proven field of
+   * trust node tn. This ensures that we add a substitution with a proof
+   * based on transforming the tn.getProven() to (= x t) if tn has a
```
HanielB, line 45:
```suggestion
   * based on transforming tn.getProven() to (= x t) if tn has a
```

src/theory/trust_substitutions.h
```diff
@@ -0,0 +1,67 @@
...
+   * non-null proof generator; otherwise if tn has no proof generator
+   * we simply add the substitution.
```
HanielB, line 47:
```suggestion
   * non-null proof generator; otherwise we simply add the substitution.
```

src/theory/trust_substitutions.h
```diff
@@ -0,0 +1,67 @@
...
+   */
+  void addSubstitutionSolved(TNode x, TNode t, TrustNode tn);
```
HanielB, line 49: I don't follow the name. Why not overload `addSubstitution`?

HanielB (commented):
My only concern with this PR is that this "addSubstitutionSolved" name seems confusing and in the places where it replaced "addSubstitution" there were no clarifications about it.

src/theory/trust_substitutions.h
ajreynol: This indicates that we had a literal `lit` inside of `tn`, and it was put into solved form `x = t`.

I think this is a bit of a different interface than the one above, which says i can prove `x = t`.  This method is with respect to a previous formula and puts it into solved form. I will add some more comments about this.

HanielB: I see. This "solved form" bit clarifies it. I guess it's not necessary to add comments on the calls.

HanielB (approved):

PR #5203, opened 2020-10-05
Replace calls to SmtEngine with calls to Solver.

abdoo8080:
This PR replaces more calls to SmtEngine functions with calls to corresponding Solver functions in `command.cpp`. The PR also adds `CVC4_API_RECOVERABLE_CHECK` macro to use for recoverable exceptions.

### DIFF
--- a/src/smt/command.cpp
+++ b/src/smt/command.cpp
@@ -1661,12 +1661,8 @@ void GetValueCommand::invoke(api::Solver* solver)
   {
     NodeManager* nm = solver->getNodeManager();
     smt::SmtScope scope(solver->getSmtEngine());
-    std::vector<Node> result;
-    for (const Expr& e :
-         solver->getSmtEngine()->getValues(api::termVectorToExprs(d_terms)))
-    {
-      result.push_back(Node::fromExpr(e));
-    }
+    std::vector<Node> result =
+        api::termVectorToNodes(solver->getValue(d_terms));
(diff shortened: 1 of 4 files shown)


src/smt/command.cpp
```diff
@@ -1661,12 +1661,8 @@ void GetValueCommand::invoke(api::Solver* solver)
   {
     NodeManager* nm = solver->getNodeManager();
     smt::SmtScope scope(solver->getSmtEngine());
-    std::vector<Node> result;
-    for (const Expr& e :
-         solver->getSmtEngine()->getValues(api::termVectorToExprs(d_terms)))
-    {
-      result.push_back(Node::fromExpr(e));
-    }
+    std::vector<Node> result =
```
ajreynol, line 1664: I think we should update this method to use api::Term instead of Node. (In general, command.cpp should only use api::Term now I believe).

ajreynol (changes requested):

src/smt/command.cpp
abdoo8080: Discussed offline. Will be addressed in a different PR.

ajreynol (approved):
LGTM

aniemetz (approved):
LGTM. I'm wondering if we shouldn't guard all things that trigger `RecoverableModalException` in the SmtEngine at the API level (as done here for `getUnsatCore`).

PR #5204, opened 2020-10-05
cmake: Add warning when unit testing is disabled due to assertions being disabled.

aniemetz:
This further fixes the testing instructions for API tests (formerly known as `system tests`)
in INSTALL.md.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -229,8 +229,8 @@ We have 4 categories of tests:
   (label: **example**)
 - **regression tests** (5 levels) in directory `test/regress`
   (label: **regressN** with N the regression level)
-- **system tests** in directory `test/system`
-  (label: **system**)
+- **api tests** in directory `test/api`
+  (label: **api**)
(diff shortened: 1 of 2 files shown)


INSTALL.md
```diff
@@ -229,35 +229,38 @@ We have 4 categories of tests:
...
-                                          # > runs system/ouroborous
+                                          # > runs api/ouroborous
+    ctest -R api/ouroborous$              # run all tests that match '*api/ouroborous'
+                                          # > runs api/ouroborous
 ### Testing Unit Tests
 
 The unit tests are not built by default.
 
+To configure CVC4 with unit tests, assertions must be enabled, and building a
+non-static binary must be configured.
```
HanielB, line 262:
```suggestion
Note that CVC4 can only be configured with unit tests in non-static builds with
assertions enabled.
```

HanielB (approved):
LGTM, just one suggestion

PR #5206, opened 2020-10-05
SyGuS: Add fp.sub to default FP grammar.

aniemetz:


ajreynol (approved):

martin-cs:
Out of interest -- why?  So the results are more obvious than `f + (-g)`?

aniemetz:
So that the size of the term is smaller (from the point of view of the SyGuS solver). Improves performance.

martin-cs:
Sounds good; thanks.

PR #5207, opened 2020-10-05
(proof-new) proofs for ArithCongMan -> ee facts

alex-ozdemir:
Threads proof production through the ArithCongruenceManager's pipeline
of information to the equality engine.

The idea is to define a method:

```
ArithCongruenceManager::assertLitToEqualityEngine(literal, reason, pf)
```

This method asserts a literal to the equality engine with the given
proof and reason. It stores the proof of the literal for the equality
engine's future reference, and is smart about symmetric literals.

This machinery uses proofs *that* are not closed over the reason, as the
equality engine requires.

Other functions in the pipeline:
* assertionToEqualityEngine
* equalsCostant
construct proofs and call assertLitToEqualityEngine.

Future commits will address the ee -> ArithCongruenceManager pipeline,
and the relationship between ArithCongruenceManager and the rest of
arithmetic.


src/theory/arith/congruence_manager.h
```diff
@@ -155,9 +155,28 @@ class ArithCongruenceManager {
...
+                                 ArithVar s,
+                                 TNode reason,
+                                 std::shared_ptr<ProofNode> pf);
+
+  /** Check for proof for this or a symmetric fact
+   *
+   * @returns whether this or a symmetric fact has a proof.
+   */
+  bool hasProofFor(TNode f) const;
+  /** Sets the proof for this fact and the symmetric one. */
```
ajreynol, line 178: Can you say a bit more about this method: when/who should call this? It is SAT context dependent?

ajreynol (approved):
LGTM, one minor documentation request.

src/theory/arith/congruence_manager.h
alex-ozdemir: Good idea. I dropped a pointer here to the documentation for the proof generators, whose documentation describes the closure properties of the different kinds of proofs at play in `ArithCongruenceManager`.

PR #5208, opened 2020-10-06
bv-to-int: change order of passes

### LINKED ISSUE #5095: (solve-bv-as-int) Fatal failure at preprocessing/passes/non_clausal_simp.cpp:128

yoni206:
Closes #5095 and replaces #5110.
There are two tests in #5095 that produce two different assertion failures when using bv-to-int. 
The first happens because the substitution map wasn't applied after the pass.
The second happens because `div` (that is introduced in the pass) is not rewritten using `witness`.
Both problems are solved by making sure that `apply-substs`, `theory-preprocess` and `ite-removal` are executed after bv-to-int. 
The two tests from #5095 are included as regressions.


src/smt/process_assertions.cpp
```diff
@@ -219,8 +209,23 @@ bool ProcessAssertions::apply(Assertions& as)
   if (options::solveBVAsInt() != options::SolveBVAsIntMode::OFF)
   {
     d_passes["bv-to-int"]->apply(&assertions);
+    // after running bv-to-int, we need to immediately run
+    // theory-preprocess and ite-removal so that newlly created
+    // terms and assertions are normalized (e.g., div is expanded).
+    d_passes["theory-preprocess"]->apply(&assertions);
+    d_passes["ite-removal"]->apply(&assertions);
```
ajreynol, line 216: Is this necessary? I thought that ITE removal was subsumed by theory-preprocess.  Can you try removing this line?

ajreynol (changes requested):

src/smt/process_assertions.cpp
yoni206: right!  deleted in https://github.com/CVC4/CVC4/pull/5208/commits/2868064bce934d1641b97c7cbe0ed00bdee45bab

ajreynol (approved):
LGTM

PR #5209, opened 2020-10-06
Add operator MakeBagOp for constructing bags

mudathirmahgoub:
This PR removes subtyping rules for bags and add operator `MakeBagOp` similar to `SingletonOp`


src/expr/node_manager.cpp
```diff
@@ -961,6 +961,17 @@ Node NodeManager::mkSingleton(const TypeNode& t, const TNode n)
   return singleton;
 }
 
+Node NodeManager::mkBag(const TypeNode& t, const TNode n, const TNode m)
```
mudathirmahgoub, line 964: If we expose SINGLETON, and MK_BAG in the Solver class, I don't think we need mkBag in or mkSingleton. We can combine both in Solver::mkOp<T>(Sort s), and then use the normal function Solver::mkTerm(Op op, Term child)

src/api/cvc4cppkind.h
```diff
@@ -1832,11 +1832,23 @@ enum CVC4_PUBLIC Kind : int32_t
    */
   MEMBER,
   /**
-   * The set of the single element given as a parameter.
+   * From the API, use Term Solver::mkSingleton(Sort s, Term t)
```
mudathirmahgoub, line 1835: With `mkSingleton` we don't need to expose SINGLETON in the api. It should only be internal. However, it is used by the parser. So we can add a special token for it. 
Alternatively we can expose it  in the API and remove mkSingleton as discussed below.

src/api/cvc4cppkind.h
```diff
@@ -1832,11 +1832,23 @@ enum CVC4_PUBLIC Kind : int32_t
...
    * Create with:
-   *   mkTerm(Kind kind, Term child)
+   *   NodeManager::mkConst(SingletonOp(t))
+   *
+   * Construct a singleton set from an element given as a parameter.
+   * Parameters: 2
+   *   -[1]: SingletonOp
+   *   -[2]: Single element
+   * Create with:
+   *   mkNode(kind::SINGLETON, op, n);
```
ajreynol, line 1851: The user of the API doesn't need to know about these details of SingletonOp.  This documentation should reference the solver's `mkSingleton(n)` and simply warn not to use `mkTerm(SINGLETON, n)`.

src/expr/node_manager.h
```diff
@@ -578,12 +578,22 @@ class NodeManager {
...
 
+  /**
+  * Create a bag from the given element n along with its multiplicity m.
+  * @param t the element type of the returned bag.
+  *          Note that the type of n needs to be a subtype of t.
+  * @param n the element that is used to to construct the bag
+  * @param m the multiplicity of the element n
+  * @return a bag that contains m occurrences of n.
+  */
+  Node mkBag(const TypeNode& t, const TNode n, const TNode m);
```
ajreynol, line 595: The right name for this method is probably (and unfortunately) `mkMkBag`, since you are asking to make a node of MK_BAG kind.

src/api/cvc4cppkind.h
mudathirmahgoub: Then the user does not need to know about the SINGLETON operator. We can write don't use this kind, but then the user would wonder why is it defined here if it is not needed. I tried to delete it, but the parsers for smt, cvc, tptop complain

src/expr/node_manager.h
mudathirmahgoub: We need a better name. Or another reason to use mkOp<T>(Sort s) :)

mudathirmahgoub: `buildMKBag`, `buildMakeBag`, `constructMkBag`, `createMkBag`. 
Actually MkBag itself is ugly.

src/api/cvc4cppkind.h
ajreynol: We need SINGLETON in the API since getting the kind of a singleton node should return SINGLETON.

src/expr/node_manager.h
ajreynol: mkMkBag is consistent with our naming scheme, either that or we should rename MK_BAG.

src/api/cvc4cppkind.h
ajreynol: It is needed, the user will want to know the kind of singleton Terms.

src/expr/node_manager.cpp
ajreynol: We need this because the internal code should never call the API.

ajreynol (changes requested):
Should we remove mkSingleton from cvc4cpp.h in this PR? Or will this be separate?

test/unit/theory/theory_bags_type_rules_white.h
```diff
@@ -99,9 +101,16 @@ class BagsTypeRuleBlack : public CxxTest::TestSuite
   void testToSetOperator()
   {
     vector<Node> elements = getNStrings(1);
-    Node bag = d_nm->mkNode(MK_BAG, elements[0], d_nm->mkConst(Rational(10)));
+    Node bag = d_nm->mkBag(d_nm->stringType(), elements[0], d_nm->mkConst(Rational(10)));
     TS_ASSERT_THROWS_NOTHING(d_nm->mkNode(BAG_TO_SET, bag));
     TS_ASSERT(d_nm->mkNode(BAG_TO_SET, bag).getType().isSet());
+    std::cout<<"Rational(4, 4).isIntegral() " << d_nm->mkConst(Rational(4,4)).getType()<<  std::endl;
```
ajreynol, line 107: Ensure clang-formatted

mudathirmahgoub: Done by cleaning up these debugging lines

ajreynol (approved):

mudathirmahgoub:
> Should we remove mkSingleton from cvc4cpp.h in this PR? Or will this be separate?

This PR can be merged before `mkInteger` since it does not remove `mkSingleton`. Removing `mkSingleton` depends on merging `mkInteger` PR.

PR #5210, opened 2020-10-06
(proof-new) Make pnm protected and forward it to the nonlinear extension.

nafur:
This PR moved `d_pnm` in `Theory` from private to protected and then forwards it to the nonlinear extension constructor.


ajreynol (approved):
LGTM

PR #5214, opened 2020-10-06
Added proofFormatMode flag and trivially used flag

scottviteri:


src/smt/proof_manager.cpp
```diff
@@ -118,6 +121,10 @@ void PfManager::printProof(std::shared_ptr<ProofNode> pfn, Assertions& as)
   // TODO (proj #37) according to the proof format, print the proof node
   // leanPrinter(out, fp.get());
   std::ostream& out = *options::out();
+  if (options::proofFormatMode() == options::ProofFormatMode::LEAN) {
+    d_lpfpp->process(fp);
+    proof::leanPrinter(out, fp);
+  }
```
HanielB, line 127: Note that you should be clang-formatting.

src/proof/lean/lean_printer.h
```diff
@@ -14,23 +14,36 @@
...
+
+#include <iostream>
+#include "expr/proof_node.h"
+
+namespace CVC4 {
+namespace proof {
 
 enum LeanRule
 {
+  // in what format should I put these lean rules
```
HanielB, line 28: Let's try to follow the enumeration in `expr/proof_rule.h`. So an enum value for each lean of the Lean calculus.

scottviteri:
Woops doesn't compile

src/proof/lean/lean_printer.h
```diff
@@ -14,23 +14,37 @@
...
   // should print theorem statement
+  out << "hello world"
+      << "\n";
 
   // should traverse proof node, print each as a proof step, according to the
   // LEAN_RULE id in the LeanRule enum
   out << pfn;
 }
+
+#endif
```
HanielB, line 48: you should move this to the end of the file

src/proof/lean/lean_printer.h
Lachnitt: Can we make this an enum class?

HanielB: Yes, we should. So it should change to `enum class LeanRule : uint32_t`.

PR #5216, opened 2020-10-07
Get correct NodeManagerScope for toStrings in API

makaimann:
Gets the correct `NodeManagerScope` for getting the `toString` of a term. The following program had strange output:
```
#include "cvc4/api/cvc4cpp.h"
using namespace CVC4::api;
using namespace std;
int main()
{
  Solver s;
  Term x = s.mkConst(s.getIntegerSort(), "x");
  cout << "x = " << x << endl;
  Solver s2;
  cout << "x = " << x << endl;
  return 0;
}
```

It was outputting:
```
x = x
x = var_267
```

Fixing the scope makes the output `x = x` both times, as expected.

Thanks @4tXJ7f for the guidance.


4tXJ7f (changes requested):
Thanks a lot for doing this. Could we do the same for `Sort` (https://github.com/CVC4/CVC4/blob/0c540b4b96a100ba1d280c700c6c360efeb24fc2/src/api/cvc4cpp.cpp#L1063) and maybe add simple unit tests in https://github.com/CVC4/CVC4/blob/master/test/unit/api/term_black.h and https://github.com/CVC4/CVC4/blob/master/test/unit/api/sort_black.h? It would be good to make sure that future refactorings don't break this functionality (we'll at some point get rid of `NodeManagerScope`).

makaimann:
Good point, thanks for the feedback! Addressed your comments in 777d7eeb8006814b3c4e508a26982fe783586be0 and expanded the sort test in 1bbd654e16d0e79cfcba232b280cc915675a99ed. Would we expect to have printing issues for sorts due to the scope as well? I tried it locally, and at least for the bit-vector and uninterpreted sorts I tested, it didn't seem to matter if we set the scope or not. Is there a more complicated sort I should test?

4tXJ7f (approved):
LGTM, thanks for the quick update. Heh, I was suprised to hear that `Sort::toString()` didn't fail in a similar manner. It turns out that `Sort` still uses `Expr`-level `Type` (https://github.com/CVC4/CVC4/blob/0c540b4b96a100ba1d280c700c6c360efeb24fc2/src/api/cvc4cpp.h#L649), which does the scoping [here](https://github.com/CVC4/CVC4/blob/0c540b4b96a100ba1d280c700c6c360efeb24fc2/src/expr/type.cpp#L232). Your test will be helpful for whoever changes the internal representation of `Sort` from `Type` to `TypeNode`.

4tXJ7f (changes requested):
Sorry, I forgot about `Op`. We should do the same thing there: https://github.com/CVC4/CVC4/blob/a69b6eb561fadb4c2c9f09b700950af7ce43b378/src/api/cvc4cpp.cpp#L1482 . It is a wrapper around `Node` for the operators that are not just a simple `Kind`.

makaimann:
Oh right, good catch! Just addressed that in the last commit. Should I leave the `NodeManagerScope` code in the `Sort::toString()` implementation or take it out? Either way, I'll leave the test for it as you suggested.

4tXJ7f (approved):
LGTM, thanks for the updates. Leaving the `NodeManagerScope` in `Sort::toString` is the right thing to do because we will need it there soon.

PR #5217, opened 2020-10-07
(proof-new) Indexed root predicates

nafur:
This PR adds a new node type called `INDEXED_ROOT_PREDICATE`.
It will be used for proofs generated by the cad subsolver. It roughly works as follows:

General structure: `(INDEXED_ROOT_PREDICATE k=unsigned var~0 poly)`

Example: `(INDEXED_ROOT_PREDICATE k=1 (>= y 0.0) (+ (* 1.0 (^ x 2.0)) (* 1.0 y)))`

It states that `y` should be `>=` than the first (k'th) root of the polynomial `x*x+y` (in `y`). Note that it can only be evaluated with respect to an assignment for all the remaining variables within the polynomial.

### DIFF
--- a/src/theory/arith/kinds
+++ b/src/theory/arith/kinds
@@ -80,6 +80,13 @@ operator LEQ 2 "less than or equal, x <= y"
 operator GT 2 "greater than, x > y"
 operator GEQ 2 "greater than or equal, x >= y"
 
+constant INDEXED_ROOT_PREDICATE_OP \
+	::CVC4::IndexedRootPredicate \
+	::CVC4::IndexedRootPredicateHashFunction \
+	"util/arith.h" \
+	"operator for the indexed root predicate; payload is an instance of the CVC4::IndexedRootPredicate class"
+parameterized INDEXED_ROOT_PREDICATE INDEXED_ROOT_PREDICATE_OP 2 "indexed root predicate; first parameter is a INDEXED_ROOT_PREDICATE_OP, second is a real variable, third is a polynomial compared to zero"
+
 operator IS_INTEGER 1 "term-is-integer predicate (parameter is a real-sorted term)"
 operator TO_INTEGER 1 "convert term to integer by the floor function (parameter is a real-sorted term)"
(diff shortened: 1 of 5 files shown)


src/theory/arith/kinds
```diff
@@ -80,6 +80,13 @@ operator LEQ 2 "less than or equal, x <= y"
 operator GT 2 "greater than, x > y"
 operator GEQ 2 "greater than or equal, x >= y"
 
+constant INDEXED_ROOT_PREDICATE_OP \
+	::CVC4::IndexedRootPredicate \
+	::CVC4::IndexedRootPredicateHashFunction \
+	"util/arith.h" \
+	"operator for the indexed root predicate; payload is an instance of the CVC4::IndexedRootPredicate class"
```
ajreynol, line 87: I'm not sure about the issue of "polynomial compared to zero".  If we allow rewriting of applications of this node, I would expect that the third parameter is an arbitrary arithmetic literal (that is in turn interpreted as a polynomial compared to zero).

ajreynol (changes requested):

src/theory/arith/kinds
nafur: Ah right. I changed my mind about there I feel more comfortable having the sign condition and eventually settled for the second argument (with the variable) instead of the polynomial, but forgot to update the documentation.

src/theory/arith/theory_arith_type_rules.h
```diff
@@ -130,6 +130,34 @@ class IAndTypeRule
...
+  inline static TypeNode computeType(NodeManager* nodeManager,
+                                     TNode n,
+                                     bool check)
+  {
+    IndexedRootPredicate info = n.getOperator().getConst<IndexedRootPredicate>();
+
+    if (check)
+    {
+      TypeNode t1 = n[0].getType(check);
+      if (!t1.isBoolean())
```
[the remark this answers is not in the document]
nafur: See other comment.

src/theory/arith/kinds
ajreynol: I see, ok.

ajreynol (approved):

PR #5219, opened 2020-10-07
Make sure conflicts are not rewritten (in arithmetic)

nafur:
The arithmetic inference manager did rewrite conflicts, which lead to nodes from the conflict not being assumptions (but rewritten assumptions) triggering an assertion. This rewrite is now removed.
Furthermore rewrites triggering the same assertion within the icp subsolver have been refactored to avoid this issue.

### DIFF
--- a/src/theory/arith/nl/icp/icp_solver.cpp
+++ b/src/theory/arith/nl/icp/icp_solver.cpp
@@ -350,8 +350,7 @@ void ICPSolver::check()
     std::vector<Node> lemmas = generateLemmas();
     for (const auto& l : lemmas)
     {
-      d_im.addPendingArithLemma(Rewriter::rewrite(l),
-                                InferenceId::NL_ICP_PROPAGATION);
+      d_im.addPendingArithLemma(l, InferenceId::NL_ICP_PROPAGATION);
     }
   }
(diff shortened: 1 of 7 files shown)


src/theory/arith/nl/icp/icp_solver.cpp
```diff
@@ -350,7 +350,8 @@ void ICPSolver::check()
     std::vector<Node> lemmas = generateLemmas();
     for (const auto& l : lemmas)
     {
-      d_im.addPendingArithLemma(l, InferenceId::NL_ICP_PROPAGATION);
+      d_im.addPendingArithLemma(Rewriter::rewrite(l),
```
ajreynol: Does it make sense to always rewrite within addPendingArithLemma instead of making a special call here?

ajreynol (approved):
LGTM, one question about where to rewrite.

src/theory/arith/nl/icp/icp_solver.cpp
nafur: I just removed it. I think we want to (always) rewrite within `addPendingArithLemma` to improve caching in the arithmetic inference manager.

PR #5220, opened 2020-10-07
(proof-new) Update the strings inference manager for proofs

ajreynol:
This updates the inference manager in strings in two ways:
(1) It now inherits from InferenceManagerBuffered, meaning that the custom methods for process the current pending lemma/fact vectors are removed in favor of the standard ones.
(2) It has support for proof generation via the InferProofCons utility.

This PR standardizes three methods `processFact`, `processLemma`, and `processConflict` which take `InferInfo` and processes any string-specific behavior pertaining to processing facts, lemmas and conflicts with the inference manager.  These methods are intended to preserve the previous behavior of this class.


ajreynol:
Note this is pending performance testing, which I'm running currently.

ajreynol:
Results 60 second timeout on QF_SLIA SMT-LIB:
```
               CVC4-100720-sim_strings_d   32153   4959.32   30601   30906.3     62754     63768
                   CVC4-100720_strings_d   32177   5312.16   30615   31270.1     62792     63768
             CVC4-100720-2-sim_strings_d   32180   5397.29   30589   30547.8     62769     63768




For unsatisfiable : 
                   CVC4-100720_strings_d       X      27       0
               CVC4-100720-sim_strings_d       3       X       0
             CVC4-100720-2-sim_strings_d       3      27       X

For satisfiable : 
                   CVC4-100720_strings_d       X      21      29
               CVC4-100720-sim_strings_d       7       X      18
             CVC4-100720-2-sim_strings_d       3       6       X
```
`CVC4-100720-2-sim_strings_d ` = this PR, `CVC4-100720_strings_d` = master, `CVC4-100720-sim_strings_d` = an earlier version of this PR with lemma caching.

We lose a handful of benchmarks overall (-29+3 sat, +3-0 unsat).  Most of the lost sat benchmarks are from pyex, where many are close to timeout, and none are solved in less than 30 seconds by master.

src/theory/strings/inference_manager.cpp
```diff
@@ -113,9 +129,13 @@ void InferenceManager::sendInference(const std::vector<Node>& exp,
...
 {
-  eq = eq.isNull() ? d_false : Rewriter::rewrite(eq);
-  if (eq == d_true)
+  if (eq.isNull())
+  {
+    eq = d_false;
+  }
+  else if (Rewriter::rewrite(eq) == d_true)
   {
+    // if trivial, return
```
4tXJ7f, line 138: I actually think that the negation of this if statement should be an assertion. If we really try to send an inference where the conclusion rewrites to true, we probably created some bogus conclusion.

src/theory/strings/inference_manager.cpp
```diff
@@ -258,239 +271,167 @@ void InferenceManager::addToExplanation(Node lit, std::vector<Node>& exp) const
...
 {
-  std::vector<TNode> antec_exp;
-  // copy to processing vector
-  std::vector<Node> aconj;
-  for (const Node& ac : a)
-  {
-    utils::flattenOp(AND, ac, aconj);
-  }
-  for (const Node& apc : aconj)
+  // get the fact(s), where there are multiple if the conclusion is an AND
```
4tXJ7f, line 318:
```suggestion
  // Get the fact(s). There are multiple facts if the conclusion is an AND
```

src/theory/strings/theory_strings.cpp
```diff
@@ -623,15 +614,17 @@ void TheoryStrings::notifyFact(TNode atom,
...
+      iiPrefixConf.d_id = Inference::PREFIX_CONFLICT;
+      iiPrefixConf.d_conc = d_false;
+      utils::flattenOp(AND, pc, iiPrefixConf.d_ant);
       Trace("strings-conflict")
-          << "CONFLICT: Eager prefix : " << conflictNode << std::endl;
+          << "CONFLICT: Eager prefix : " << pc << std::endl;
       ++(d_statistics.d_conflictsEagerPrefix);
-      d_im.conflict(conflictNode);
+      // call the inference manager to send the conflict
+      d_im.processConflict(iiPrefixConf);
```
4tXJ7f, line 627: I think that it would be better to make the pending conflict store an `InferInfo`. It is a bit confusing that a pending conflict can only be a prefix conflict and may change in the future.

4tXJ7f (changes requested):
Looking great. Just some minor discussion points.

src/theory/strings/inference_manager.cpp
ajreynol: I think there are several places (e.g. inferences via cd simplification) that rely on this as a mechanism for filtering.  If this were an assertion, then this filtering would have to be done in multiple places.

src/theory/strings/theory_strings.cpp
ajreynol: Done.

4tXJ7f (approved):
LGTM, thanks for the updates.

PR #5221, opened 2020-10-07
(proof-new) Add lazy proof set utility

ajreynol:
This is a common pattern that is required in several places in preprocessing.


HanielB (approved):
LGTM

PR #5222, opened 2020-10-07
Provide API version of some SMT Commands.

abdoo8080:
This PR provides an SMT version of the following SMT commands:
- `get-instantiations`
- `block-model`
- `block-model-values`
- `get-qe`
- `get-qe-disjunct`

`command.cpp` now calls those functions instead of their SmtEngine counterparts. In addition, `SEXPR` is now an API kind.


src/api/cvc4cpp.h
```diff
@@ -3094,6 +3098,52 @@ class CVC4_PUBLIC Solver
...
 
+  /**
+   * Do quantifier elimination.
+   * SMT-LIB: ( get-qe <q> )
+   * Requires a logic that supports quantifier elimination. Currently, the only
+   * logics supported by quantifier elimination is LRA and LIA.
+   * @param q a quantified formula of the form:
+   *   Q x1...xn. P( x1...xn, y1...yn )
+   * where P( x1...xn, y1...yn ) is a quantifier-free formula
+   * @param strict whether to output warnings, such as when an unexpected logic
```
ajreynol, line 3109: Let's remove this argument.  Setting this to false is for internal use only, and always be fixed to `true` for users of the API.

src/api/cvc4cpp.h
```diff
@@ -3094,6 +3098,52 @@ class CVC4_PUBLIC Solver
...
+   *   - If Q is exists, let A^Q_n be the formula
+   *       A ^ ~ret^Q_1 ^ ... ^ ~ret^Q_n
+   *     where for each i=1,...n, formula ret^Q_i is the result of calling
+   *     getQuantifierEliminationDisjunct for q with the set of assertions
+   *     A^Q_{i-1}. Similarly, if Q is forall, then let A^Q_n be
+   *       A ^ ret^Q_1 ^ ... ^ ret^Q_n
+   *     where ret^Q_i is the same as above. In either case, we have
+   *     that ret^Q_j will eventually be true or false, for some finite j.
+   */
+  Term getQuantifierEliminationDisjunct(api::Term q, bool strict = true) const;
```
ajreynol, line 3145: Ditto, remove the strict argument.

src/api/cvc4cpp.h
```diff
@@ -2868,12 +2868,16 @@ class CVC4_PUBLIC Solver
...
+
   /**
    * Assert a formula.
    * SMT-LIB: ( assert <term> )
    * @param term the formula to assert
+   * @param isUnsatCore determines whether to include this asserted formula in
+   * an unsat core (if one is later requested)
    */
-  void assertFormula(Term term) const;
+  void assertFormula(Term term, bool isUnsatCore = true) const;
```
ajreynol, line 2880: I believe we decided that this would be a more general name like `isTracked`.  You should then say that including this asserted formula in the unsat is an *example* of what is implied by setting this true.  I believe there were other examples of what this would impact (e.g. getAssignment).  @4tXJ7f suggested this?

ajreynol (changes requested):
This looks good, a few initial comments.

src/api/cvc4cpp.h
```diff
@@ -2868,12 +2868,16 @@ class CVC4_PUBLIC Solver
    */
   Term simplify(const Term& t);
 
+  Term expandDefinitions(Term t) const;
```
aniemetz, line 2871: This needs to be documented.

test/unit/api/solver_black.h
```diff
@@ -141,6 +143,16 @@ class SolverBlack : public CxxTest::TestSuite
...
+  void testBlockModel1();
+  void testBlockModel2();
+  void testBlockModel3();
+  void testBlockModel4();
+  void testBlockModelValues1();
+  void testBlockModelValues2();
+  void testBlockModelValues3();
+  void testBlockModelValues4();
+  void testBlockModelValues5();
+
```
aniemetz, line 155: Test for `expandDefinitions` is missing.

aniemetz (changes requested):

aniemetz:
Do we really need to export `expandDefinitions`? I feel like this should be internal only.

src/api/cvc4cpp.h
abdoo8080: > Do we really need to export `expandDefinitions`? I feel like this should be internal only.

I discussed this with Andy and he agrees. I removed this change. In addition, I will remove `ExpandDefinitionsCommand` class and `--expand-definitions` option in a separate PR.

test/unit/api/solver_black.h
abdoo8080: Not applicable anymore.

src/api/cvc4cpp.h
abdoo8080: Discussed offline. Removed the option.

aniemetz (approved):

ajreynol (approved):

PR #5223, opened 2020-10-07
Simplify approach for collapsed selectors over non-closed enumerable types

ajreynol:
This simplifies the approach for wrongly-applied selectors whose range type involves e.g. uninterpreted sorts.  These cases are particularly tricky since uninterpreted constants should never appear in facts or lemmas.

Previously, the rewriter had special cases for either making a ground term or value, and a purify step was used to eliminate the occurrences of uninterpreted sorts afterwards.

This PR leverages `mkGroundTerm` in this inference instead, and makes the rewriter uniformly use `mkGroundValue`.  

This also required making the type enumerator for datatypes more uniform.  In particular, we require that the first value enumerated by the type enumerator is of the same shape as the term required by `mkGroundTerm`, or else debug-check-model will fail.  This is due to the fact that now the inference for wrongly applied selectors uses terms while the rewriter uses values.  The type enumerator is updated so that recursive codatatypes also first take `mkGroundValue` as the zero term, when they are well-founded.  Previously this was skipped since the codatatype enumerator uses a different algorithm for computing its ground terms.

This is in preparation for further work on optimizations and proofs for datatypes.


src/theory/datatypes/type_enumerator.cpp
```diff
@@ -201,7 +201,20 @@ Node DatatypesEnumerator::getTermEnum( TypeNode tn, unsigned i ){
...
+   // Start with the ground term constructed via mkGroundValue, which does
+   // a traversal over the structure of the datatype to find a finite term.
+   // Notice that mkGroundValue may be dependent upon extracting the first
+   // value of type enumerators for *other non-datatype* subfield types of
+   // this datatype. Since datatypes can not be embedded in non-datatype
+   // types (e.g. (Array D D) cannot be a subfield type of datatype D), this
+   // call is guaranteed to avoid infinite recursion. It is important that we
+   // start with this term, since it has the same shape as the one returned by
+   // TypeNode::mkGroundTerm for d_type, which avoids debug check model
+   // failures.
```
HanielB, line 213: say why failures are avoided

src/theory/datatypes/type_enumerator.cpp
```diff
@@ -201,7 +201,20 @@ Node DatatypesEnumerator::getTermEnum( TypeNode tn, unsigned i ){
...
+   d_zeroTerm = d_datatype.mkGroundValue(d_type);
+   // Only use the zero term if it was successfully constructed. This may
+   // fail for codatatype types whose only values are infinite.
+   d_zeroTermActive = !d_zeroTerm.isNull();
```
HanielB, line 217: Say in the PR description why you moved this to be general rather than just for non-codatatypes.

HanielB (approved):
LGTM, just a few comment requests

PR #5227, opened 2020-10-08
bv2int: bvand translation code move

yoni206:
This PR is essentially a code move. 
It moves the code that translates `bvand` from the `bv_to_int` preprocessing pass to a new class `IAndHelper`, in the new files `nl_iand_utils.{h,cpp}`. The goal is to have this code in a place which can be shared between the bv2int preprocessing pass and the `iand` solver.

Future PRs will:
1. Optimize this utility
2. Make use of it in the iand solver

The code in `nl_iand_utils.{h,cpp}` is almost completely a move from `bv_to_int.{h,cpp}`. The changes are all minor, and include, e.g., the following changes:
1. the node manager is no longer a member, and so is the constant 0.
2. using the `oneBitAnd` function directly, without a function pointer.

### DIFF
--- a/src/preprocessing/passes/bv_to_int.h
+++ b/src/preprocessing/passes/bv_to_int.h
@@ -75,7 +75,7 @@
 #include "context/context.h"
 #include "preprocessing/preprocessing_pass.h"
 #include "preprocessing/preprocessing_pass_context.h"
-#include "theory/arith/nl/nl_iand_utils.h"
+#include "theory/arith/nl/iand_table.h"
 
 namespace CVC4 {
 namespace preprocessing {
@@ -284,7 +284,7 @@ class BVToInt : public PreprocessingPass
   Node d_one;
   
   // helper class for handeling bvand translation
-  theory::arith::nl::IAndUtils d_iandUtils;
+  theory::arith::nl::IAndTable d_iandTable;
 };
 
 }  // namespace passes
--- a/src/preprocessing/passes/bv_to_int.h
+++ b/src/preprocessing/passes/bv_to_int.h
@@ -282,7 +282,8 @@ class BVToInt : public PreprocessingPass
    */
   Node d_zero;
   Node d_one;
-
+  
+  // helper class for handeling bvand translation
   theory::arith::nl::IAndHelper d_iandHelper;
 };
 
--- a/src/preprocessing/passes/bv_to_int.h
+++ b/src/preprocessing/passes/bv_to_int.h
@@ -75,6 +75,7 @@
 #include "context/context.h"
 #include "preprocessing/preprocessing_pass.h"
 #include "preprocessing/preprocessing_pass_context.h"
+#include "theory/arith/nl/nl_iand_utils.h"
 
 namespace CVC4 {
 namespace preprocessing {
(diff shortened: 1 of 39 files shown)


src/theory/arith/nl/nl_iand_utils.h
```diff
@@ -0,0 +1,105 @@
...
+   * @param y is an integer operand that correspond to the second original
+   *        bit-vector operand.
+   * @param bvsize is the bit width of the original bit-vector variables.
+   * @param granularity is specified in the options for this preprocessing
+   *        pass.
+   * @param f is a pointer to a boolean function that corresponds
+   *        to the original bitwise operation.
+   * @return A node that represents the operation, as described above.
+   */
+  Node createBitwiseNode(Node x, Node y, uint64_t bvsize, uint64_t granularity);
```
ajreynol, line 79: These should be static.

src/theory/arith/nl/nl_iand_utils.h
```diff
@@ -0,0 +1,105 @@
...
+#include <vector>
+
+#include "expr/node.h"
+
+namespace CVC4 {
+namespace theory {
+namespace arith {
+namespace nl {
+
+class IAndHelper
```
ajreynol, line 28: I'd call this `IAndUtils`.

ajreynol (changes requested):

src/preprocessing/passes/bv_to_int.h
```diff
@@ -354,6 +282,9 @@ class BVToInt : public PreprocessingPass
    */
   Node d_zero;
   Node d_one;
+  
+  // helper class for handeling bvand translation
```
ajreynol, line 286:
```suggestion
  /** helper class for handeling bvand translation */
```

src/theory/arith/nl/iand_table.cpp
```diff
@@ -0,0 +1,157 @@
...
+  return nm->mkConst<Rational>(intpow2(k));
+}
+
+bool oneBitAnd(bool a, bool b) { return (a && b); }
+
+Node IAndTable::createITEFromTable(
+    Node x,
+    Node y,
+    uint64_t granularity,
+    std::map<std::pair<uint64_t, uint64_t>, uint64_t> table)
```
ajreynol, line 45:
```suggestion
    const std::map<std::pair<uint64_t, uint64_t>, uint64_t>& table)
```

ajreynol: Note that if you don't do this, it will be copied.

src/theory/arith/nl/iand_table.h
```diff
@@ -0,0 +1,110 @@
...
+   * @param table a function from pairs of integers to integers.
+   *        The domain of this function consists of pairs of
+   *        integers between 0 (inclusive) and 2^{bitwidth} (exclusive).
+   * @return An ite term that represents this table.
+   */
+  Node createITEFromTable(
+      Node x,
+      Node y,
+      uint64_t granularity,
+      std::map<std::pair<uint64_t, uint64_t>, uint64_t> table);
```
ajreynol, line 102:
```suggestion
      const std::map<std::pair<uint64_t, uint64_t>, uint64_t>& table);
```

ajreynol (changes requested):

PR #5229, opened 2020-10-08
reset-assertions: Remove all non-global symbols in the parser

### LINKED ISSUE #5163: Model soundness / parser bug related to (reset-assertions)
Hi, for this formula with (reset-assertions), CVC4 gives an unsound result. 
```
[596] % z3 small.smt2
unsat
[597] % cvc4 small.smt2
sat
[598] % 
[598] % cat small.smt2
(set-logic ALL)
(define-fun a () Bool false)
(reset-assertions)
(assert a)
(check-sat)
[599] %
```

OS: Ubuntu 18.04
Commit: 69e1472

4tXJ7f:
Fixes #5163. When `:global-declarations` is disabled (the default), then
`(reset-assertions)` should remove all declared and defined symbols. Before
this commit, we would remove the defined function from `SmtEngine` but the
parser would not remove it from its symbol table because the symbol was defined
at (what it considered) level 0. Level 0, however, is reserved for global
symbols that we never want to remove.

This commit adds an additional global `pushScope()`/`popScope()`, similar to
what we have in `SmtEngine`. As a result, user-defined symbols are now defined
at level 1 and get removed from the `SymbolTable` when `(reset-assertions)` is
called. The commit also makes sure that symbols defined by the logic are
asserted at level 0, so that they are not removed by `(reset-assertions)`. It
adds a flag to `defineType` to ignore duplicate definitions because some
symbols are defined by multiple logics, which leads to a failing assertion when
inserting them both at level 0 in the hashmap of the `SymbolTable`. E.g.
strings and integer arithmetic both define `Int`. The commit also fixes
existing unit tests that fail with these stricter semantics of
`(reset-assertions)`.


ajreynol (approved):
Looks like there is a build error in debug, otherwise LGTM.

PR #5231, opened 2020-10-09
use right arith explanation fn to fix regressions

alex-ozdemir:
Use the Node-returning variants of
Constraint::externalExplainByAssertions in TheoryArithPrivate locations
that aren't computing proofs.

The problem is that the TrustNode returning variant requires that arith
has stored an externally valid literal for the constraint, which is not
always the case.

This is what we did on proof-new, I just forgot.

Note: I'm still running `make check`. I'll comment when it is done.


alex-ozdemir:
`make check` passes. That was faster than I remember...

ajreynol (approved):
LGTM, thanks for the quick fix.

PR #5240, opened 2020-10-11
Eliminate uses of Expr in SmtEngine interface

ajreynol:
This eliminates all use of Expr in the SmtEngine except in a few interfaces (setUserAttribute, getAssignment) whose signature is currently in question.

The next steps will be to (1) eliminate Expr from the smt/model.h interface, (2) replace Type by TypeNode in Sort.


abdoo8080 (approved):
You can remove `exprVectorToNodes` function in `smt_engine.cpp` line 119 I added a while ago. Otherwise, LGTM.

PR #5242, opened 2020-10-11
(proof-new) Change merge policy for proof node updater

ajreynol:
This updates the proof node updater with a `mergeSubproofs` field which automatically merges subproofs in the same SCOPE that prove the same thing.  This is currently enabled by default, it is now configurable and disabled by default.


HanielB (approved):
LGTM

PR #5243, opened 2020-10-11
(proof-new) New rules for Booleans

ajreynol:
This adds 2 new rules for convenience to the Boolean checker.

### DIFF
--- a/src/theory/booleans/proof_checker.cpp
+++ b/src/theory/booleans/proof_checker.cpp
@@ -13,6 +13,8 @@
  **/
 
 #include "theory/booleans/proof_checker.h"
+#include "expr/skolem_manager.h"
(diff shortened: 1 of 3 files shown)


src/theory/booleans/proof_checker.cpp
```diff
@@ -212,6 +217,18 @@ Node BoolProofRuleChecker::checkInternal(PfRule id,
...
+    Assert(children.size() == 2);
+    if (children[1].getKind() == Kind::NOT && children[0] == children[1][0])
+    {
+      return NodeManager::currentNM()->mkConst(false);
+    }
+    else
+    {
+      return Node::null();
+    }
+  }
```
HanielB, line 231: no-else-after-return

PR #5244, opened 2020-10-11
(proof-new) Simplifications for proof rule checker interface

ajreynol:
Some of the interfaces in the proof rule checker are unnecessary now and can be deleted.

This also updates STRING_TRUST to have pedantic level 2.

### DIFF
--- a/src/expr/proof_checker.cpp
+++ b/src/expr/proof_checker.cpp
@@ -30,39 +30,6 @@ Node ProofRuleChecker::check(PfRule id,
   return checkInternal(id, children, args);
 }
 
-Node ProofRuleChecker::checkChildrenArg(PfRule id,
-                                        const std::vector<Node>& children,
-                                        Node arg)
-{
-  return check(id, children, {arg});
-}
-Node ProofRuleChecker::checkChildren(PfRule id,
-                                     const std::vector<Node>& children)
-{
-  return check(id, children, {});
-}
-Node ProofRuleChecker::checkChild(PfRule id, Node child)
-{
-  return check(id, {child}, {});
-}
-Node ProofRuleChecker::checkArg(PfRule id, Node arg)
-{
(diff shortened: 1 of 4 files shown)


src/theory/strings/proof_checker.cpp
```diff
@@ -53,7 +53,7 @@ void StringProofRuleChecker::registerTo(ProofChecker* pc)
   pc->registerChecker(PfRule::STRING_CODE_INJ, this);
   pc->registerChecker(PfRule::STRING_SEQ_UNIT_INJ, this);
   // trusted rules
-  pc->registerTrustedChecker(PfRule::STRING_TRUST, this, 1);
+  pc->registerTrustedChecker(PfRule::STRING_TRUST, this, 2);
```
HanielB, line 56: Mention this change in the PR description as well

src/expr/proof_checker.cpp
```diff
@@ -194,17 +161,20 @@ Node ProofChecker::checkDebug(PfRule id,
   // since we are debugging, we want to treat trusted (null) checkers as
   // a failure.
   Node res = checkInternal(id, cchildren, args, expected, out, false);
-  Trace(traceTag) << "ProofChecker::checkDebug: " << id;
-  if (res.isNull())
-  {
-    Trace(traceTag) << " failed, " << out.str() << std::endl;
-  }
-  else
+  if (Trace.isOn(traceTag))
```
HanielB, line 164: Update the documentation of `checkDebug` to say that the check is done if we give an active trace.

HanielB (approved):
LGTM, just some minor doc requests

src/expr/proof_checker.cpp
ajreynol: Actually this was just an optimization, the traceTag doesn't impact whether the check is run (usually you are looking to see the return value).

PR #5245, opened 2020-10-11
(proof-new) Generalize preprocess proof generator

ajreynol:
This class is of general use, not just as the overall maintainer of proofs of preprocessing, but also locally within preprocessing passes.  This generalizes the interface slightly and also does some minor cleaning.

### DIFF
--- a/src/smt/preprocess_proof_generator.cpp
+++ b/src/smt/preprocess_proof_generator.cpp
@@ -21,9 +21,13 @@
 namespace CVC4 {
 namespace smt {
 
-PreprocessProofGenerator::PreprocessProofGenerator(context::UserContext* u,
-                                                   ProofNodeManager* pnm)
-    : d_pnm(pnm), d_src(u), d_helperProofs(u)
+PreprocessProofGenerator::PreprocessProofGenerator(ProofNodeManager* pnm,
+                                                   context::Context* c,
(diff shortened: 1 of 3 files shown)


src/smt/preprocess_proof_generator.cpp
```diff
@@ -21,9 +21,13 @@
...
 namespace smt {
 
-PreprocessProofGenerator::PreprocessProofGenerator(context::UserContext* u,
-                                                   ProofNodeManager* pnm)
-    : d_pnm(pnm), d_src(u), d_helperProofs(u)
+PreprocessProofGenerator::PreprocessProofGenerator(ProofNodeManager* pnm,
+                                                   context::Context* c,
+                                                   std::string name)
+    : d_pnm(pnm),
+      d_src(c ? c : &d_context),
```
HanielB, line 28: I'm confused about where `d_context` is set?

ajreynol: d_context is a non-pointer data member with a default constructor, one could put `d_context()` above but its not necessary.

HanielB: I see. I forgot how this worked. :)

HanielB (approved):

PR #5246, opened 2020-10-11
(proof-new) More features for SMT proof post-processor

ajreynol:
This updates the SMT proof postprocessor with additional stats. It also adds the feature to handle conjunctions as substitution, e.g. a child proof concluding (and (= x t) (= y s)) is interpreted as a substitution for x, y, whereas previously we insisted they must be provided separately.


src/theory/builtin/proof_checker.h
```diff
@@ -84,13 +84,26 @@ class BuiltinProofRuleChecker : public ProofRuleChecker
...
-                              TNode& subs,
-                              MethodId ids = MethodId::SB_DEFAULT);
+  static bool getSubstitutionForLit(Node exp,
+                                    TNode& var,
+                                    TNode& subs,
+                                    MethodId ids = MethodId::SB_DEFAULT);
+  /**
+   * Get substitution for formula exp. Adds to vars/subs to the substitution
+   * specified by exp for the substitution method ids, which may be multiple
+   * substitutions if exp is of kind AND. The vector from are the literals
```
HanielB, line 97:
```suggestion
   * substitutions if exp is of kind AND. The vector "from" are the literals
```
Confusing to read otherwise. :)

HanielB, line 97: Say why this case is restricted for ids == SB_DEFAULT

HanielB (approved):
LGTM, just a minor request

PR #5247, opened 2020-10-11
Guard uses of printing approximations for constants

ajreynol:
Errors with these methods were encountered while working on new techniques for arithmetic preprocessing.

Also, there is obvious inefficiency since approximations for constants are being computed when certain traces are disabled.


src/theory/arith/arith_utilities.cpp
```diff
@@ -106,7 +106,11 @@ bool isTranscendentalKind(Kind k)
 
 Node getApproximateConstant(Node c, bool isLower, unsigned prec)
 {
-  Assert(c.isConst());
+  if (!c.isConst())
+  {
+    Assert(false) << "getApproximateConstant: non-constant input " << c;
+    return Node::null();
+  }
```
nafur, line 113: I really don't like this. If we think that `c` should really be a constant here, we should assert this. If it is not, why would we make the production build robust to it?

src/theory/arith/arith_utilities.cpp
```diff
@@ -178,7 +182,12 @@ Node getApproximateConstant(Node c, bool isLower, unsigned prec)
 
 void printRationalApprox(const char* c, Node cr, unsigned prec)
 {
-  Assert(cr.isConst());
+  if (!cr.isConst())
+  {
+    Assert(false) << "printRationalApprox: non-constant input " << cr;
+    Trace(c) << cr;
+    return;
+  }
```
nafur, line 190: ditto.

nafur (changes requested):
Just one remark about assertions...

src/theory/arith/arith_utilities.cpp
ajreynol: In theory, the code should always be robust to this (in both debug and production).  Moreover we shouldn't *require* a failure if we are in production, since this is an internal error.  This is the common use of "Assert" instead of "AlwaysAssert".

nafur (approved):
LGTM.

PR #5249, opened 2020-10-12
(proof-new) Added addTrustedConflict() for arithmetic inference manager

nafur:
This PR adds a new method `addTrustedConflict` for the arithmetic inference manager.

PR #5250, opened 2020-10-12
(proof-new) Infrastructure for CAD proofs

nafur:
This PR prepares adding proofs for the CAD solver. It adds the necessary proof rules and a proof checker, and passes the proof node manager to where we need it.


nafur:
This PR depends on #5249 to actually make sense (https://github.com/CVC4/CVC4/pull/5250/files#diff-772af8b86c2602ebb3d3af0dd6ac8a45R101)

ajreynol (approved):
LGTM, I think I agree about the stratification of proof checker you're doing here.  We haven't done this before (a submodule of a theory register its own proof rules) but I think it makes sense in this case.

PR #5252, opened 2020-10-12
(proof-new) Make circuit-propagator proof producing

nafur:
This PR make the `CircuitPropagator` (mostly) proof producing.
It is sufficient to pass regressions, but not complete yet:
- not all cases are covered yet
- there is still an issue with the lifetime of proofs (that @ajreynol will debug)
- the code is only partially cleaned up


src/theory/booleans/circuit_propagator_proofs.h
```diff
@@ -0,0 +1,580 @@
+/*********************                                                        */
+/*! \file circuit_propagator_proofs.h
```
ajreynol, line 2: I'd call this file proof_circuit_propagator.h / cpp to follow our general naming conventions of proof-producing classes.

src/theory/booleans/circuit_propagator_proofs.h
```diff
@@ -0,0 +1,580 @@
...
+    {
+      lits.emplace_back(negate ? (*i).notNode() : Node(*i));
+    }
+  }
+  return lits;
+}
+
+}  // namespace
+
+struct CircuitPropagatorProver
```
ajreynol, line 59: ProofCircuitPropagator would follow our naming conventions better.  Also this should be a class (and cleaned up later...).

src/theory/booleans/proof_checker.cpp
```diff
@@ -173,6 +173,7 @@ Node BoolProofRuleChecker::checkInternal(PfRule id,
       // literal to be removed from "second" clause. They will be negated
       if (i > 0)
       {
+        elim.insert(args[i - 1].notNode());
```
ajreynol, line 176: Not sure if we want this change. @HanielB ?

ajreynol (changes requested):
A few high level things

src/theory/booleans/circuit_propagator_proofs.h
nafur: Yes, there will be cleanup. Also adding a cpp, etc. (it is only included by the circuit_propagator.cpp, so it's not so bad having everything in the header for the meantime)

ajreynol (approved):

PR #5253, opened 2020-10-12
bv-to-int: apply substs before and after

### LINKED ISSUE #5230: (solve-bv-as-int) Fatat failure at preprocessing/passes/bv_to_int.cpp:992, 177, 689
cvc e5913461e
~~~~
$ cat xx.smt2
(set-logic QF_UFBV)
(set-option :check-unsat-cores true)
(set-option :solve-bv-as-int sum)
(declare-const v0 Bool)
(declare-const v5 Bool)
(declare-const v8 Bool)
(declare-const v9 Bool)
(check-sat-assuming ((! (or (distinct ((_ sign_extend 0) #b1001110) (bvlshr #b1001110 #b1001110) ((_ sign_extend 0) #b1001110) #b1001110) v9 (= #b1001110 ((_ sign_extend 0) #b1001110) #b1001110 #b1001110 ((_ sign_extend 0) #b1001110)) v5 (distinct ((_ sign_extend 0) #b1001110) (bvlshr #b1001110 #b1001110) ((_ sign_extend 0) #b1001110) #b1001110)) :named IP_183) (! (or v0 v8) :named IP_211)))
~~~~

~~~~
$ cvc xx.smt2
Fatal failure within CVC4::Node CVC4::preprocessing::passes::BVToInt::createShiftNode(std::vector<CVC4::NodeTemplate<true> >, uint64_t, bool) at /home/peisen/test/tofuzz/CVC4/src/preprocessing/passes/bv_to_int.cpp:994
Check failure

 !y.isConst()

Aborted
~~~~

yoni206:
In #5208 we made `bv-to-int` run before `apply-substs`.
The reason was that `bv-to-int` introduces new nodes, that need to be processed by apply-substs.
This solved #5095 .

However, when doing that, I neglected the fact that `bv-to-int` actually relies on `apply-substs`, e.g., when assuming that some bit-vector operators are eliminated (such as shift by const).This was revealed by #5230 .

My suggestion (implemented in this PR) is to run `apply-substs` both before (in order to satisfy the assumptions of `bv-to-int` regarding eliminated operators) and after (in order to process the introduced assertions).

The three tests from #5230 are added to regressions, and so this PR closes #5230 if merged.


yoni206:
Subsumed by #5263.

PR #5254, opened 2020-10-13
(proof-new) Prove lemmas in Constraint

alex-ozdemir:
Includes:
* T/F splitting lemmas for any arith constraint
* Unate lemmas produced early on

The hard part is proving the unate lemmas. In general, they are all implied by 2-constraint farkas proofs, so we ultimately map them all down to `proveOr`, which constructs that proof.

`make check` was happy with this change. Hopefully the CI agrees :).

### DIFF
--- a/src/theory/arith/theory_arith_private.cpp
+++ b/src/theory/arith/theory_arith_private.cpp
@@ -1063,7 +1063,7 @@ bool TheoryArithPrivate::AssertDisequality(ConstraintP constraint){
 
   if(!split && c_i == d_partialModel.getAssignment(x_i)){
     Debug("arith::eq") << "lemma now! " << constraint << endl;
-    outputLemma(constraint->split());
+    outputTrustedLemma(constraint->split());
     return false;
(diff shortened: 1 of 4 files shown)


ajreynol:
So btw, CI on master does not test proofs yet, so testing on proof-new is the only way currently to be sure its correct.

I'm assuming these changes are coming from proof-new directly essentially?

src/theory/arith/theory_arith_private.cpp
```diff
@@ -1918,6 +1918,12 @@ void TheoryArithPrivate::outputConflicts(){
   }
 }
 
+void TheoryArithPrivate::outputTrustedLemma(TrustNode lemma)
+{
+  Debug("arith::channel") << "Arith trusted lemma: " << lemma << std::endl;
+  (d_containing.d_out)->lemma(lemma.getNode());
```
alex-ozdemir, line 1924: This line is different from `proof-new`.

ajreynol (approved):

PR #5256, opened 2020-10-13
Fix miscellaneous warnings

ajreynol:
Mostly in cardinality extension, which was cleaned in the previous PR.


aniemetz:
I feel like my changes in #5305 are cleaner and should go in first, you still compare `size_t` with `uint32_t`.

ajreynol:
I disagree, assertCardinality on your PR is taking a size_t, which is intuitively not the right type.  The first argument to `assertCardinality` should be an unsigned integer that originates from constant rational (see this pR).  Comparing size_t and uint32_t is not a warning, but regardless this should be fixed by a cast of uint32_t to size or vice versa if you think this is necessary.

src/theory/uf/cardinality_extension.cpp
```diff
@@ -873,8 +875,8 @@ void SortModel::assertCardinality(int c, bool val)
         }
       }
       // we assert it positively, if its beyond the bound, abort
-      if (options::ufssAbortCardinality() != -1
-          && c >= options::ufssAbortCardinality())
+      if (options::ufssAbortCardinality() >= 0
+          && static_cast<int>(c) >= options::ufssAbortCardinality())
```
aniemetz, line 879: This is ugly. This is better since cardinality can't be negative anyways:
```suggestion
          && c >= static_cast<uint32_t>(options::ufssAbortCardinality()))
```

src/theory/uf/cardinality_extension.cpp
```diff
@@ -1141,13 +1135,14 @@ bool SortModel::checkLastCall()
     }
   }
   RepSet* rs = m->getRepSetPtr();
-  int nReps = (int)rs->getNumRepresentatives(d_type);
+  size_t nReps = rs->getNumRepresentatives(d_type);
   if( nReps!=(d_maxNegCard+1) ){
     Trace("uf-ss-warn") << "WARNING : Model does not have same # representatives as cardinality for " << d_type << "." << std::endl;
     Trace("uf-ss-warn") << "   Max neg cardinality : " << d_maxNegCard << std::endl;
     Trace("uf-ss-warn") << "   # Reps : " << nReps << std::endl;
     if( d_maxNegCard>=nReps ){
```
aniemetz, line 1143: Include if/else statements in formatting.

src/theory/uf/cardinality_extension.cpp
```diff
@@ -852,7 +852,8 @@ void SortModel::assertCardinality(int c, bool val)
       bool doCheckRegions = !d_hasCard;
       bool prevHasCard = d_hasCard;
       d_hasCard = true;
-      if( !prevHasCard || c<d_cardinality ){
+      if (!prevHasCard || c < static_cast<uint32_t>(d_cardinality))
```
aniemetz, line 855: `d_cardinality` is already a `uint32_t`, no cast necessary.

src/theory/uf/cardinality_extension.cpp
```diff
@@ -862,7 +863,8 @@ void SortModel::assertCardinality(int c, bool val)
       }
       //should check all regions now
       if( doCheckRegions ){
```
aniemetz, line 865: Format.

src/theory/uf/cardinality_extension.cpp
```diff
@@ -1056,7 +1049,8 @@ void SortModel::addCliqueLemma(std::vector<Node>& clique)
 {
   Assert(d_hasCard);
   Assert(d_cardinality > 0);
-  while( clique.size()>size_t(d_cardinality+1) ){
+  while (clique.size() > d_cardinality + 1)
```
aniemetz, line 1052: This is were you compare against `size_t`. You can leave it as it is, though (this one is safe).

aniemetz (changes requested):

aniemetz (approved):

PR #5260, opened 2020-10-13
(proof-new) Use proper lazy proof construction in circuit-propagator

nafur:
We now use LazyCDProofChain internally as well to properly put together proofs. Also fixes a few more nullptr issues.

### DIFF
--- a/src/expr/lazy_proof_chain.cpp
+++ b/src/expr/lazy_proof_chain.cpp
@@ -94,6 +94,12 @@ std::shared_ptr<ProofNode> LazyCDProofChain::getProofFor(Node fact)
           << "LazyCDProofChain::getProofFor: Call generator " << pg->identify()
           << " for chain link " << cur << "\n";
       std::shared_ptr<ProofNode> curPfn = pg->getProofFor(cur);
+      if (curPfn == nullptr)
+      {
+        Trace("lazy-cdproofchain")
+            << "LazyCDProofChain::getProofFor: No proof found, skip\n";
+        continue;
+      }
       // map node whose proof node must be expanded to the respective poof node
       toConnect[cur] = curPfn;
       if (!rec)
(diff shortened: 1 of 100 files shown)


src/expr/lazy_proof_chain.cpp
```diff
@@ -107,7 +107,10 @@ std::shared_ptr<ProofNode> LazyCDProofChain::getProofFor(Node fact)
           << "\n";
       // retrieve free assumptions and their respective proof nodes
       std::map<Node, std::vector<std::shared_ptr<ProofNode>>> famap;
-      expr::getFreeAssumptionsMap(curPfn, famap);
+      if (curPfn != nullptr)
+      {
+        expr::getFreeAssumptionsMap(curPfn, famap);
+      }
```
HanielB, line 113: let's move this test to before checking `!rec`. If `curPfn` is `null` we mark cur as post-visited and skip. Only if it is not null we add it to `toConnect`. This will also remove the necessity of the test you added in the end.

src/expr/lazy_proof_chain.cpp
```diff
@@ -185,9 +188,12 @@ std::shared_ptr<ProofNode> LazyCDProofChain::getProofFor(Node fact)
...
         << "\n";
-    Trace("lazy-cdproofchain-debug")
-        << "LazyCDProofChain::getProofFor: ...expand to " << *npfn.second.get()
-        << "\n";
+    if (npfn.second != nullptr)
+    {
+      Trace("lazy-cdproofchain-debug")
+          << "LazyCDProofChain::getProofFor: ...expand to "
+          << *npfn.second.get() << "\n";
+    }
```
HanielB, line 196: Remove, as explained above.

src/expr/lazy_proof_chain.cpp
nafur: Is 2459f857ddca0318cc8cd10ae271d40d72509d77 what you had in mind?

HanielB: Almost. Just need to also do `visited[cur] = true;` when `curPfn` is null.

ajreynol: Actually id merge this with the `rec` condition. This also means that we cache toConnect[cur] being nullptr.

HanielB: Currently it does not cache null. When `rec` is false though we do cache. I'd not like to cache null, as it'd require another test after the visiting.

ajreynol: We already have that test actually (https://github.com/CVC4/CVC4/blob/proof-new/src/expr/lazy_proof_chain.cpp#L225).  I added it.  I'd prefer caching nullptr here since it makes the invariant of the data structure cleaner.

PR #5262, opened 2020-10-13
[proof-new] Fixing resolution proof checker

HanielB:
Previously the binary resolution checker was:

- Checking applications in which for a pivot `(not l)` the literal `(not l)` would be eliminated from the first clause and `l` from the second because double negation was handled implicitly. Now whether the binary resolution is such that the pivot is removed as is from the first clause and negated from the second, or the other way around, is marked via an argument.
- Not producing `false` the remaining set of literals after resolution was empty.

This commit also updates the informal description of the rule accordingly, as well as to clarify the behavior when the pivot does not occur properly in the clauses (in which case the rule application corresponds to weakening).


src/expr/proof_rule.h
```diff
@@ -243,24 +243,38 @@ enum class PfRule : uint32_t
...
-  //   G_j = (not F_i)
+  //   - C1 and C2 are nodes viewed as clauses, i.e., either an OR node with
+  //     each children viewed as a literal or a node viewed as a literal. Note
+  //     that an OR node could also be a literal.
+  //   - id is either true or false
+  //   - L is the pivot of the resolution, which occurs as is (resp. under a
+  //     NOT) in C1 and negatively (as is) in C2 if id = true (id = false).
+  //   C is a clause resulting from collecting all the literals in C1, minus the
+  //   first occurrence of the pivot, and C2, minus the first occurrence of the
+  //   pivot, according to the policy above.
```
nafur, line 259: This is not correct, strictly speaking. It should be something like "minus the first occurrence of the pivot or its negation, depending on id".

src/theory/booleans/proof_checker.cpp
```diff
@@ -80,17 +80,45 @@ Node BoolProofRuleChecker::checkInternal(PfRule id,
...
+      {
+        if (pivots[i] != lits[j])
+        {
+          disjuncts.push_back(lits[j]);
+        }
+        else
         {
-          disjuncts.push_back(children[i][j]);
+          // just eliminate first occurrence
+          pivots[i] = Node::null();
```
nafur, line 121: Is the `mkNode(kind::OR` robust to getting an empty or a singleton list of children? I don't think so, and we explicitly handle this case at the end of `CHAIN_RESOLUTION`...

src/theory/booleans/proof_checker.cpp
```diff
@@ -159,23 +187,50 @@ Node BoolProofRuleChecker::checkInternal(PfRule id,
...
     {
-      std::unordered_set<Node, NodeHashFunction> elim;
+      std::unordered_map<Node, unsigned, NodeHashFunction> elim;
       // literals to be removed from "first" clause
       if (i < childrenSize - 1)
       {
-        elim.insert(args.begin() + i, args.end());
+        for (unsigned j = (2*i), argsSize = args.size(); j < argsSize; j = j + 2)
+        {
+          // whether pivot should occur as is or negated depends on the id of
```
nafur, line 204: Maybe it's too early in the morning, but I can't follow what you're doing here.

In `elim`, you are collecting all pivot nodes (negated or not) starting at the current `i`, and then you add the current pivot node (negated). Finally, you collect all literals of the current clause and add them (minus what is in `elim`) into `clauseNodes`.

So what confuses me is the following: why do you compute `elim` in every loop from scratch? It seems to me that this removes pivots multiple times from the clauses. Assume you have three pivot (with four clauses) you would put the last pivot three times in `elim`. Why should the pivot used for the resolution of the third and fourth clause be eliminated from the first clause?

nafur (changes requested):
I'm not sure about the code for `CHAIN_RESOLUTION`. I guess this needs some more comments...

src/theory/booleans/proof_checker.cpp
HanielB: Indeed, we have to handle this case here as well. I'l add.

src/theory/booleans/proof_checker.cpp
HanielB: You're right. This is yet another misconception because I inadvertently wrote a checker for resolution + some factoring when I should have written one for just resolution. This is good though because the checking will be much simpler now. Thanks for noticing!

HanielB: After internal discussion we've decided to change the chain resolution rule in another PR, as significant changes will be necessary.

src/expr/proof_rule.h
```diff
@@ -246,14 +246,25 @@ enum class PfRule : uint32_t
...
-  //   G_j = (not F_i)
+  //   - C1 and C2 are nodes viewed as clauses, i.e., either an OR node with
+  //     each children viewed as a literal or a node viewed as a literal. Note
+  //     that an OR node could also be a literal.
+  //   - id is either true or false
+  //   - L is the pivot of the resolution, which occurs as is (resp. under a
+  //     NOT) in C1 and negatively (as is) in C2 if id = true (id = false).
+  //   C is a clause resulting from collecting all the literals in C1, minus the
+  //   first occurrence of the pivot or its negation, and C2, minus the first
+  //   occurrence of the pivot or its negation, according to the policy above.
```
nafur, line 262: We should mention here that we simplify a "singleton clause" to a literal.

nafur (approved):
Just one comment about the documentation of the proof rule, otherwise LGTM.

src/expr/proof_rule.h
HanielB: Updated. Also took the opportunity to clarify about the empty clause case.

nafur (approved):
LGTM.

PR #5264, opened 2020-10-14
run_regression.py to fail on invalid requirements

yoni206:
Currently, the following test is skipped when running `ctest`:
```
 ; REQUIRES: symfpuuuu
 ; EXPECT: sat
 (set-logic ALL)
 (assert true)
 (check-sat)
```

I think it is better for such a test to fail, because otherwise it might never run without anyone noticing (this almost happened to me here: https://github.com/CVC4/CVC4/pull/5263/commits/c7e277b364cb800fd2003763cc3703f8528c9891).

This PR makes such tests to fail by checking whether the `REQUIRES` value is valid. No regressions fail as a result.

Co-authored-by: Mathias Preiner <mathias.preiner@gmail.com>


test/regress/run_regression.py
```diff
@@ -313,15 +316,21 @@ def run_regression(unsat_cores, proofs, dump, use_skip_return_code, wrapper,
         return (EXIT_SKIP if use_skip_return_code else EXIT_OK)
 
     for req_feature in requires:
-        if req_feature.startswith("no-"):
-            inv_feature = req_feature[len("no-"):]
```
mpreiner, line 317: Add `is_negative = True` to the if block and replace `inv_feature` with `req_feature`.

mpreiner (approved):
LGTM, thanks for the changes!

mpreiner:
Btw. make sure to format the code with yapf

yoni206:
yapf'ed

PR #5265, opened 2020-10-14
bv2int: implementing the iand-sum mode

yoni206:
There are 3 potential modes to lazily treat the `iand` operator:
(1) by value (typical CEGAR loop)
(2) by sum (lazily expanding each `iand` node into a sum of ITEs)
(3) by bit-wise equalities (lazily expanding each `iand` node into bit-wise equalities)

This PR implements (2).

The relevant option is added to existing tests, and a new test is added. In a few other tests, some options are removed to make them run faster.


test/regress/regress1/nl/iand-native-2.smt2
```diff
@@ -0,0 +1,18 @@
+; COMMAND-LINE: --iand-mode=value
+; COMMAND-LINE: --iand-mode=sum --bvand-integer-granularity=1
```
ajreynol, line 2: I think some of these large amounts of variants seem a bit excessive, do we need to test for multiple granularities here?  I'd prefer if only a small number of benchmarks varied to the integer granularity.

ajreynol (changes requested):
One minor request for regressions, I'd like to be a bit more conservative in the number of variants you test for some of these.

Otherwise, the PR looks good to me.

ajreynol (approved):
LGTM

PR #5266, opened 2020-10-14
(proof-new) Missing proofs for circuit-propagator

nafur:
This PR implements the missing proofs for the circuit propagator and adds a regressions that does some basic checks (as the missing cases were the one that were not checked by the existing regressions).


test/regress/regress0/preprocess/circuit-prop.smt2
```diff
@@ -0,0 +1,53 @@
+; COMMAND-LINE: --proof-new-req --check-proofs-new
```
ajreynol, line 1:
```suggestion
; COMMAND-LINE:
```

ajreynol: The first option won't ever be put in master and the second will eventually be built in to run_regression.py.

ajreynol (changes requested):

ajreynol (approved):

PR #5267, opened 2020-10-14
(proof-new) TheoryArithPrivate: farkas lemma proof

alex-ozdemir:
Use a farkas proof to prove one of arithmetic's lemmas.

These changes were checked-out directly from `proof-new`, without modification.

### DIFF
--- a/src/theory/arith/theory_arith_private.cpp
+++ b/src/theory/arith/theory_arith_private.cpp
@@ -4599,7 +4599,7 @@ bool TheoryArithPrivate::rowImplicationCanBeApplied(RowIndex ridx, bool rowUp, C
       Node clause = flattenImplication(implication);
       std::shared_ptr<ProofNode> clausePf{nullptr};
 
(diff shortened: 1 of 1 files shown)


src/theory/arith/theory_arith_private.cpp
```diff
@@ -4597,7 +4597,63 @@ bool TheoryArithPrivate::rowImplicationCanBeApplied(RowIndex ridx, bool rowUp, C
       }
       Node implication = implied->externalImplication(explain);
       Node clause = flattenImplication(implication);
-      outputLemma(clause);
+      std::shared_ptr<ProofNode> clausePf{nullptr};
+
+      if (options::proofNew())
```
ajreynol, line 4602:
```suggestion
      if (isProofEnabled())
```

ajreynol (changes requested):
One minor request

ajreynol (approved):
LGTM

PR #5268, opened 2020-10-14
Remove some Commands from the API.

abdoo8080:
This PR removes `Solver::getAssignment` command from the API as there is no way to assign names to terms in the API. It also removes `ExpandDefinitionsCommand`, an internal functionality in CVC4.


ajreynol (approved):
LGTM, assuming @aniemetz agrees.

aniemetz:
What is the alternative for (get-assignment), then?

aniemetz:
We might argue that `getValue` is sufficient, and that we just don't have an API command that corresponds to `(get-assignment)`. I guess I'd be fine with this but first want to hear your opinions.

ajreynol:
The consensus from the past meetings is that expression names will not be included in the API.  This means that get-assignment will be implemented by the planned "ParserState" object, which will store a mapping between names and expressions and use calls to Solver::getValue.

aniemetz (approved):

Issue #5269, opened 2020-10-14
Assertion failure within bool CVC4::theory::sets::TheorySetsRels::areEqual


mudathirmahgoub:
The following file [rels.txt](https://github.com/CVC4/CVC4/files/5380367/rels.txt) generates an assertion failure

```
Fatal failure within bool CVC4::theory::sets::TheorySetsRels::areEqual(CVC4::Node, CVC4::Node) at src/theory/sets/theory_sets_rels.cpp:1144
Check failure

 a.getType() == b.getType()

Aborted (core dumped)
```
in the assertion below
```
  bool TheorySetsRels::areEqual( Node a, Node b ){
    Assert(a.getType() == b.getType());
```

ajreynol:
Fixed by #5270.

PR #5270, opened 2020-10-14
Fix issue #5269

mudathirmahgoub:
This PR fixes issue #5269 of  unnecessary calling `TheorySetsRels::areEqual` in a product relation (which compares the rightmost element of the first child with the leftmost element in the second child). This may violate an assertion in `TheorySetsRels::areEqual`  that the types of the 2 elements should be the same as shown below:
```
[Theory::Rels] Start composing members for relation = (product |this/BankAccount | (product univInt |this/Time |))
[Theory::Rels] r1_rmost: srj_39 of type Atom
[Theory::Rels] r2_lmost: skv_3 of type UInt
Fatal failure within bool CVC4::theory::sets::TheorySetsRels::areEqual(CVC4::Node, CVC4::Node) at src/theory/sets/theory_sets_rels.cpp:1149
Check failure

 a.getType() == b.getType()
```


ajreynol (changes requested):
Can you add the regression?

mudathirmahgoub:
> Can you add the regression?

Done. It is not straightforward to reduce the regression. It is added as regress 2 since it takes a little bit to finish.

ajreynol (approved):

ajreynol:
LGTM, nice catch.

PR #5274, opened 2020-10-15
Add mkInteger to the API

mudathirmahgoub:
This PR adds mkInteger to the API and update mkReal to ensure that the returned term has real sort. 
The parsers are modified to parse numbers like "0" "5" as integers and "0.0" and "15.0" as reals. 
This means subtyping is effectively eliminated from all theories except arithmetic.
Other changes:
  - Term::isValue is removed which was introduced to support parsing for constant arrays. It is no longer needed after this PR. 
  - Benchmarks are updated to match the changes in the parsers

Co-authored-by: Andrew Reynolds andrew-reynolds@uiowa.edu


src/api/cvc4cpp.cpp
```diff
@@ -1614,12 +1629,20 @@ size_t Term::getNumChildren() const
   {
     return d_node->getNumChildren() + 1;
   }
+  if(isCastedReal())
```
aniemetz, line 1632: Why do we need this here? Shouldn't `getNumChildren` have the correct value?

src/api/cvc4cpp.cpp
```diff
@@ -1731,7 +1754,7 @@ bool Term::isNull() const { return isNullHelper(); }
 bool Term::isValue() const
 {
   CVC4_API_CHECK_NOT_NULL;
-  return d_node->isConst();
+  return d_node->isConst() || isCastedReal();
```
aniemetz, line 1757: Again, why? Shouldn't `isConst` return true for consts independently from `isCastedReal`?

src/api/cvc4cpp.cpp
```diff
@@ -3692,24 +3715,64 @@ Term Solver::mkPi() const
...
+Term Solver::mkInteger(int64_t val) const
+{
+  CVC4_API_SOLVER_TRY_CATCH_BEGIN;
+  Term integer = mkValHelper<CVC4::Rational>(CVC4::Rational(val));
+  Assert(integer.getSort() == getIntegerSort());
+  return integer;
+  CVC4_API_SOLVER_TRY_CATCH_END;
+}
+
+Term Solver::mkInteger(int64_t num, int64_t den) const
```
aniemetz, line 3737: Why would you need this for `mkInteger`?

src/api/cvc4cpp.cpp
ajreynol: We should discuss this approach.  Basically, we are using `(to_real n)` for integer constant `n` to represent the real constant `n`. This means that we need the API to treat `(to_real n)` as a constant, meaning that the Node and Term have a different opinion about the # children. In particular, `(to_real n)` has 1 child according to the internal API and 0 children according to the external API.

src/api/cvc4cpp.cpp
ajreynol: Agreed, I'm not sure why this is necessary.

src/api/cvc4cpp.cpp
ajreynol: For the same reasons I describe above.

src/api/cvc4cpp.cpp
mudathirmahgoub: It is not necessary and it is removed. I thought of some cases where the user may not want to compute the result of integer division, but I guess the user should either use c++ integer division, or cvc4 division operator.

src/parser/tptp/Tptp.g
```diff
@@ -698,17 +698,17 @@ definedFun[CVC4::ParseOp& p]
...
+                          MK_TERM(api::PLUS, n, SOLVER->mkInteger(1))),
                   // if decPart == 0.5, round to nearest even integer:
                   // result is: to_int(n/2 + .5) * 2
                   MK_TERM(api::MULT,
                           MK_TERM(api::TO_INTEGER,
                                   MK_TERM(api::PLUS,
                                           MK_TERM(api::DIVISION,
                                                   n,
-                                                  SOLVER->mkReal(2)),
+                                                  SOLVER->mkInteger(2)),
```
HanielB, line 709: Is it necessary that this change happens?

src/parser/tptp/Tptp.g
```diff
@@ -698,17 +698,17 @@ definedFun[CVC4::ParseOp& p]
                   MK_TERM(api::GT, decPart, SOLVER->mkReal(1, 2)),
                   // if decPart > 0.5, round up
                   MK_TERM(api::TO_INTEGER,
-                          MK_TERM(api::PLUS, n, SOLVER->mkReal(1))),
+                          MK_TERM(api::PLUS, n, SOLVER->mkInteger(1))),
```
HanielB, line 701: Is it necessary that this change happens?

src/parser/tptp/Tptp.g
```diff
@@ -677,7 +677,7 @@ definedFun[CVC4::ParseOp& p]
       api::Term formals = MK_TERM(api::BOUND_VAR_LIST, n);
       api::Term expr =
           MK_TERM(api::ITE,
-                  MK_TERM(api::GEQ, n, SOLVER->mkReal(0)),
+                  MK_TERM(api::GEQ, n, SOLVER->mkInteger(0)),
```
HanielB, line 680: Is it necessary that this change happens?

src/parser/tptp/Tptp.g
```diff
@@ -627,7 +627,7 @@ definedFun[CVC4::ParseOp& p]
       api::Term formals = MK_TERM(api::BOUND_VAR_LIST, n, d);
       api::Term expr = MK_TERM(api::DIVISION, n, d);
       expr = MK_TERM(api::ITE,
-                     MK_TERM(api::GEQ, expr, SOLVER->mkReal(0)),
+                     MK_TERM(api::GEQ, expr, SOLVER->mkInteger(0)),
```
HanielB, line 630: Is it necessary that this change happens?

src/parser/tptp/Tptp.g
```diff
@@ -604,7 +604,7 @@ definedFun[CVC4::ParseOp& p]
       api::Term formals = MK_TERM(api::BOUND_VAR_LIST, n, d);
       api::Term expr = MK_TERM(api::DIVISION, n, d);
       expr = MK_TERM(api::ITE,
-                     MK_TERM(api::GEQ, d, SOLVER->mkReal(0)),
+                     MK_TERM(api::GEQ, d, SOLVER->mkInteger(0)),
```
HanielB, line 607: Is it necessary that this change happens?

src/parser/tptp/Tptp.g
mudathirmahgoub: No, but I think it is better to have `Rational(2)` here than `(TO_REAL Rational(2))`

src/parser/tptp/Tptp.g
mudathirmahgoub: ditto

src/parser/tptp/Tptp.g
ajreynol: So what @mudathirmahgoub is right: after this PR, it is more *efficient* to use mkInteger here since it avoids a cast.  However, it is less intuitive, since this means that we are assuming that DIVISION is polymorphic, allowing Real divided by Int.

Alternatively we could just multiply by 0.5 here :-P

src/parser/tptp/Tptp.g
ajreynol: Same issue: performance (change) vs intuitiveness (original).

src/api/cvc4cpp.cpp
mudathirmahgoub: It seems this is not enough to satisfy the following [assertion](https://github.com/CVC4/CVC4/pull/5274/files#diff-6e779ce1b72d75a5502238049f07e8ad68d1ebb6c8f4cfc68bef2b194f1c495bR794) which fails. 
`isValue` returns true for the term `((as const (Array Real Real)) 0)`, because the node is a const, however
it returns `false` for `(store ((as const (Array Real Real)) 0) 1.0 2.0)` which is equivalent to the non-constant node `(store ((as const (Array Real Real)) 0) (to_real 1) (to_real 2))`.
I guess the same situation applies for sets (inserting casted integers into a set of reals). 

Possible suggestions I can think of to fix this. 
1- Handle this special case for the `STORE` operator, by converting `TO_REAL` to integer nodes
2- Make  `Term::isCastedReal` or `Node::isValue` recursive as `Term::isConst` in nodes.
3- Modify  `Node::isConst` to treat TO_REAL nodes as const (which Andy wants to avoid at all costs)

src/parser/tptp/Tptp.g
HanielB: I'd say efficiency is of no consequence for this so I'd go for what is more clear.

mudathirmahgoub: Done by going for what is more clear

src/api/cvc4cpp.cpp
mudathirmahgoub: After discussing this with Andy, it is became clear that having 2 different implementations of `Term::isValue` and `Term::isConst` is a bad idea. Then this question arised, does the user need `isValue` at all? So we decided to remove it until there is a convincing scenario that shows its value.

src/api/cvc4cpp.cpp
```diff
@@ -3691,24 +3708,54 @@ Term Solver::mkPi() const
...
+  return integer;
+  CVC4_API_SOLVER_TRY_CATCH_END;
+}
+
 Term Solver::mkReal(const std::string& s) const
 {
   CVC4_API_SOLVER_TRY_CATCH_BEGIN;
-  return mkRealFromStrHelper(s);
+  Term real = mkRealFromStrHelper(s);
+  if (real.getSort() == getIntegerSort())
```
ajreynol, line 3734: Can you make this part of mkRealFromStrHelper?  I would like this decision to use TO_REAL to be controlled a single piece of code.  You'll probably need to add an argument "asReal" to mkRealFromStrHelper.

src/api/cvc4cpp.h
```diff
@@ -1151,6 +1144,10 @@ class CVC4_PUBLIC Term
    */
   Kind getKindHelper() const;
 
+  /**
+   * returns true if the current term is a casted real
```
ajreynol, line 1148: Can you explain this in more detail?

ajreynol (changes requested):

src/api/cvc4cpp.h
mudathirmahgoub: Done

src/api/cvc4cpp.cpp
mudathirmahgoub: Done by calling  `ensureRealSort`

ajreynol (approved):
LGTM, although I'd like at least one more reviewer (marking "complex").  We should also discuss this at length in the meeting tomorrow.

ajreynol:
@mudathirmahgoub  can you update the PR description with the reason for this PR (to eliminate subtyping), and the summary of the approach we are taking?

src/api/cvc4cpp.cpp
aniemetz: As discussed offline, sounds good.

src/api/cvc4cpp.cpp
```diff
@@ -3691,24 +3706,56 @@ Term Solver::mkPi() const
...
   CVC4_API_SOLVER_TRY_CATCH_BEGIN;
-  return mkRealFromStrHelper(s);
+  Term rational = mkRealFromStrHelper(s);
+  return ensureRealSort(rational);
   CVC4_API_SOLVER_TRY_CATCH_END;
 }
 
+Term Solver::ensureRealSort(const Term t) const
+{
+  Assert(t.getSort() == getIntegerSort() || t.getSort() == getRealSort());
```
aniemetz, line 3738: This shouldn't be an exception unless `mkReal` guards and checks that the string passed in is indeed a Real and not an Integer sort. `mkRealFromStrHelper` only checks if it is a string that represents either an Int, Real or rational value.

src/parser/cvc/Cvc.g
```diff
@@ -2207,18 +2200,19 @@ simpleTerm[CVC4::api::Term& f]
       std::stringstream strRat;
       strRat << r;
       f = SOLVER->mkReal(strRat.str());
+      // This is no longer required
```
aniemetz, line 2203: If it's no longer required, it should be removed.

aniemetz (changes requested):

src/api/cvc4cpp.cpp
mudathirmahgoub: `ensureRealSort` is a private function. So for now, only `mkReal` functions call it. Still, if it does not affect the performance, I prefer to have this check to prevent passing terms with wrong sort.

src/parser/cvc/Cvc.g
mudathirmahgoub: will do

src/api/cvc4cpp.cpp
aniemetz: We discussed and resolved this offline.

src/api/cvc4cpp.cpp
```diff
@@ -3108,16 +3124,20 @@ Term Solver::mkValHelper(T t) const
...
+  {
+    CVC4::Rational r = s.find('/') != std::string::npos
+                           ? CVC4::Rational(s)
+                           : CVC4::Rational::fromDecimal(s);
+    return mkValHelper<CVC4::Rational>(r);
+  }
+  catch (const std::invalid_argument& e)
+  {
+    std::stringstream message;
+    message << "Could not construct a number from string argument '" << s << "'"
```
aniemetz, line 3137:
```suggestion
    message << "Cannot construct Real or Int from string argument '" << s << "'"
```

test/unit/api/solver_black.h
```diff
@@ -744,12 +744,13 @@ void SolverBlack::testMkPi() { TS_ASSERT_THROWS_NOTHING(d_solver->mkPi()); }
 
 void SolverBlack::testMkReal()
```
aniemetz, line 745: Create a test `testMkInteger` where you more or less duplicate `testMkReal` and convert all `mkReal` tests to `mkInteger`. Do *not* convert any `mkReal` test in `testMkInteger` but leave as is.
This applies to both `testMkReal` functions.

aniemetz (changes requested):

src/api/cvc4cpp.cpp
mudathirmahgoub: Done

test/unit/api/solver_black.h
mudathirmahgoub: Makes sense. Done

aniemetz (approved):

PR #5283, opened 2020-10-16
bv2int: caching introduced terms

### LINKED ISSUE #5281: (solve-bv-as-int) Invalid model on QF_UF formula
Hi for the following formula, cvc4 1a1aea5
~~~~
(set-option :solve-bv-as-int sum)
(declare-fun uf6 (Bool Bool Bool Bool Bool Bool) Bool)
(declare-fun uf7 (Bool Bool Bool Bool Bool Bool Bool) Bool)
(assert (uf7 true true true true true true true))
(push 1)
(assert (uf6 true true true true (uf7 true true true true true true true) true))
(check-sat)
~~~~
~~~~
$./cvc4 --check-models -i -q xx.smt2
Fatal failure within void CVC4::SmtEngine::checkModel(bool) at /home/peisen/test/tofuzz/CVC4/src/smt/smt_engine.cpp:1755
assertion:     (uf7 true true true true true true true)
simplifies to: false
expected `true'.
Aborted
~~~~

yoni206:
The bv-to-int preprocessing pass has a cache that maps each BV node to its translated integer node.
In case the BV node is a function symbol, it's translated integer node is a fresh function symbol.
On current master, if we later process this fresh function symbol in the preprocessing pass, we again create a fresh function symbol as its translation. This is unsound, and was discovered in #5281, whose benchmark is added to regressions.
Closes #5281 if merged.


ajreynol (approved):
LGTM

PR #5285, opened 2020-10-16
Catch more cases of nested recursion in datatypes

### LINKED ISSUE #5280: Fatal failure at theory/theory_model_builder.cpp:886
Hi for the following formula, cvc4 1a1aea5503
~~~~
$cat f.smt2
(set-logic ALL)
(declare-datatype ty0 ((Emp) (Container (v2 (Set ty0)))))
(declare-fun v1 () ty0)
(assert (not ((_ is Emp) v1)))
(assert (= (v2 v1) (singleton v1)))
(check-sat)

$./cvc4 f.smt2
Fatal failure within bool CVC4::theory::TheoryEngineModelBuilder::buildModel(CVC4::theory::TheoryModel*) at /home/peisen/test/tofuzz/CVC4/src/theory/theory_model_builder.cpp:886
Check failure

 false

Aborted
~~~~

ajreynol:
Fixes #5280. 

Previously we were checking for nested recursive datatypes in expandDefinitions.  This does not catch cases where the only terms of a malformed nested recursive datatype are variables.  The proper place to check is in preRegisterTerm.  The benchmark from that issue now gives an error.

### DIFF
--- a/test/regress/regress0/datatypes/issue5280-no-nrec.smt2
+++ b/test/regress/regress0/datatypes/issue5280-no-nrec.smt2
@@ -1,0 +1,7 @@
+; EXPECT: (error "")
+(set-logic ALL)
+(declare-datatype ty0 ((Emp) (Container (v2 (Set ty0)))))
+(declare-fun v1 () ty0)
+(assert (not ((_ is Emp) v1)))
+(assert (= (v2 v1) (singleton v1)))
+(check-sat)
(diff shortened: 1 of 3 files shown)


test/regress/regress0/datatypes/issue5280-no-nrec.smt2
```diff
@@ -0,0 +1,7 @@
+; EXPECT: (error "")
```
yoni206, line 1: A suggestion: might be a good idea to duplicate this test with `dtNestedRec` turned on, and then to expect no errors. Up to you if that's necessary.

yoni206 (approved):
LGTM, with one suggestion for your consideration.

test/regress/regress0/datatypes/issue5280-no-nrec.smt2
ajreynol: This leads to a memout.  This actually crashed my machine.

This should not be a test, since --dt-nested-rec is not recommended.

PR #5287, opened 2020-10-16
Expand `seq.nth` lazily

yoni206:
Our first support for `seq.nth` eliminated it eagerly during `expandDefinitions`.
This PR changes that, by eliminating it lazily, as done with other extended string operators.


ajreynol (approved):
LGTM, thanks for the updates.

Btw, I am planning to improve check-model so that these check model failures will not be an issue.  The reason they are an issue is because now SEQ_NTH is a partial operator that is not eliminated eagerly at expandDefinitions. The same issue must be addressed for e.g. INTS_DIVISION.

PR #5289, opened 2020-10-17
 (proof-new) Witness axiom reconstruction for purification skolems

ajreynol:
This formalizes the proofs of existentials that justify purification variables, e.g. those with witness form `witness x. x = t` for the term `t` they purify. 

This PR generalizes EXISTS_INTRO to do this, and makes some simplifications to SkolemManager.


src/smt/witness_form.cpp
```diff
@@ -92,6 +93,17 @@ Node WitnessFormGenerator::convertToWitnessForm(Node t)
...
           ProofGenerator* pg = skm->getProofGenerator(exists);
+          if (pg == nullptr)
+          {
+            // it may be a purification skolem
+            pg = convertExistsInternal(exists);
+            if (pg == nullptr)
+            {
+              Trace("witness-form")
+                  << "WitnessFormGenerator: No proof generator for " << exists
+                  << std::endl;
```
HanielB, line 104: Don't we want to use some kind of default proof here?

src/theory/quantifiers/proof_checker.cpp
```diff
@@ -43,8 +44,25 @@ Node QuantifiersProofRuleChecker::checkInternal(
...
+      return Node::null();
+    }
+    // substitution must contain only the variable of the existential
+    for (const std::pair<const Node, Node>& s : subs)
+    {
+      if (s.first != exists[0][0])
+      {
+        return Node::null();
+      }
+    }
```
HanielB, line 64: maybe check this first? The match should be the most expensive check.

ajreynol: `subs` is computed by the match :)

src/smt/witness_form.cpp
ajreynol: Yes, this is just a warning, we fill in via WITNESS_AXIOM below when none is provided.

src/theory/quantifiers/proof_checker.cpp
HanielB: Indeed that makes sense! :)

src/smt/witness_form.cpp
HanielB: I mean what is gonna be the justification of the existential? Do I understand correctly that it'll be an open leaf?

ajreynol: It's justified by a trusted rule "WITNESS_AXIOM".

HanielB: Ah, I see. I forgot about this feature of `addLazyStep`. Maybe add a comment to clarify this?

HanielB (approved):

PR #5291, opened 2020-10-17
(proof-new) Refactoring cyclic checks

ajreynol:
This PR makes it so that cyclic checks are only performed eager when proofNewEagerChecking is true.

It refactors the existing cyclic check in ProofNodeToSExpr to have a more consistent style, and adds a cyclic check to getFreeAssumptions.


HanielB (approved):
LGTM

PR #5299, opened 2020-10-19
(proof-new) Update add lazy step interface in LazyCDProof

ajreynol:
Ensuring closed proofs should not be enabled by default, it is actually not used very often as a whole.  Moreover, the "trust id" argument is the most useful argument and hence should come as the 3rd argument.  This updates all uses of addLazyStep for the change in interface, also changes term conversion generator which had a similar issue with default arguments.

Notice that some calls to addLazyStep were checking closed but without providing a debug string, these I've left alone (they no longer check closed).

### DIFF
--- a/src/theory/trust_substitutions.h
+++ b/src/theory/trust_substitutions.h
@@ -74,7 +74,7 @@ class TrustSubstitutionMap
   ProofGenerator* addSubstitutionSolved(TNode x, TNode t, TrustNode tn);
   /**
    * Add substitutions from trust substitution map t. This adds all
-   * substitutions for t and carries over its information about proofs.
+   * substitutions from the map t and carries over its information about proofs.
    */
   void addSubstitutions(TrustSubstitutionMap& t);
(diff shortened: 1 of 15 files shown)


src/theory/trust_substitutions.h
```diff
@@ -68,11 +68,13 @@ class TrustSubstitutionMap
...
    * we simply add the substitution.
+   *
+   * @return The proof generator that can prove (= x t).
    */
-  void addSubstitutionSolved(TNode x, TNode t, TrustNode tn);
+  ProofGenerator* addSubstitutionSolved(TNode x, TNode t, TrustNode tn);
   /**
    * Add substitutions from trust substitution map t. This adds all
-   * substitutions from the map t and carries over its information about proofs.
+   * substitutions for t and carries over its information about proofs.
```
HanielB, line 77: revert

HanielB (approved):
LGTM, just a minor request

PR #5300, opened 2020-10-19
(proof-new) Updates to assertions pipeline and preprocess generator

ajreynol:
This updates and improves assertions pipeline and preprocess generator so that they avoid cyclic proofs and have better infrastructure for catching pedantic failures.

This is in preparation for making the non-clausal-simplification pass proof producing.


nafur (approved):
LGTM.

PR #5303, opened 2020-10-19
Split CheckModels utility to its own file

ajreynol:
This utility will be undergoing major additions to make it more robust. Thus it should be moved to its own file.

There are no major code changes in this PR, a few things were changed to be consistent to the coding standard.


nafur (approved):
LGTM.

PR #5190, opened 2020-10-02
Integer: GMP: Move implementation of member functions to .cpp file.

aniemetz:
This moves the GMP implementation of member functions of class Integer from
the header to the .cpp file. This only moves code, and adds documentation for
previously undocumented or poorly documented functions.

I ran a performance check on the cluster on non-incremental QF_BV, QF_BVFP,
QF_FP, QF_LIA, QF_LIRA and QF_LRA (BV and FP logics heavily rely on class
Integer since class BitVector is implemented on top of Integer). This change has
no impact on performance. Diff below is cluster noise.

```
                                                            status  result  solved    sat  unsat   best  timeout  memout  error  uniq  disagr   time_cpu     memory
config
2020-10-01-int-pre-moved-impl/default/non_incremental_QF_BV     ok   41713   37947  13236  24711  33311     3729      37      0     8       0  1347418.0  7441889.7
2020-10-01-int-moved-impl/default/non_incremental_QF_BV         ok   41713   37946  13238  24708   4643     3730      37      0     7       0  1347379.3  7434748.7
                                                                                   
                                                              status  result  solved    sat  unsat   best  timeout  memout  error  uniq  disagr  time_cpu    memory
config
2020-10-01-int-moved-impl/default/non_incremental_QF_BVFP         ok   17231   17225  14048   3177    297        6       0      0     0       0    6714.2  269669.3
2020-10-01-int-pre-moved-impl/default/non_incremental_QF_BVFP     ok   17231   17225  14048   3177  16928        6       0      0     0       0    6718.9  269757.9
                                                                                   
                                                            status  result  solved    sat  unsat   best  timeout  memout  error  uniq  disagr  time_cpu    memory
config
2020-10-01-int-moved-impl/default/non_incremental_QF_FP         ok   40318   40227  20129  20098     59       85       6      0     1       0   31515.6  282395.6
2020-10-01-int-pre-moved-impl/default/non_incremental_QF_FP     ok   40318   40226  20129  20097  40168       86       6      0     0       0   31674.2  279882.4
                                                                                   
                                                             status  result  solved   sat  unsat  best  timeout  memout  error  uniq  disagr  time_cpu     memory
config
2020-10-01-int-moved-impl/default/non_incremental_QF_LIA         ok    6947    4089  2627   1462   586     2858       0      0     5       0  973671.1  2605520.8
[...]

PR #5304, opened 2020-10-19
Integer: CLN: Move implementation of member functions to .cpp file.

aniemetz:
This moves the CLN implementation of member functions of class Integer from
the header to the .cpp file. This only moves code, and adds documentation for
previously undocumented or poorly documented functions.

Analogous to #5190.

### DIFF
--- a/src/util/integer_cln_imp.cpp
+++ b/src/util/integer_cln_imp.cpp
@@ -14,16 +14,14 @@
  ** [[ Add lengthier description here ]]
  ** \todo document this file
  **/
-#include "util/integer.h"
-
 #include <sstream>
 #include <string>
 
 #include "cvc4autoconfig.h"
-
+#include "util/integer.h"
 
 #ifndef CVC4_CLN_IMP
-#  error "This source should only ever be built if CVC4_CLN_IMP is on !"
+#error "This source should only ever be built if CVC4_CLN_IMP is on !"
 #endif /* CVC4_CLN_IMP */
 
 #include "base/check.h"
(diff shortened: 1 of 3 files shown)


src/util/integer_cln_imp.cpp
```diff
@@ -53,41 +151,252 @@ Integer Integer::setBit(uint32_t i, bool value) const
...
+
+int Integer::sgn() const
+{
+  cln::cl_I sgn = cln::signum(d_value);
+  return cln::cl_I_to_int(sgn);
+}
+
+bool Integer::strictlyPositive() const { return sgn() > 0; }
+
+bool Integer::strictlyNegative() const { return sgn() < 0; }
```
nafur, line 362:
```suggestion
bool Integer::strictlyNegative() const { return minusp(d_value); }
```

src/util/integer_cln_imp.cpp
```diff
@@ -98,104 +407,159 @@ void Integer::readInt(const cln::cl_read_flags& flags,
...
     // including CLN v1.3.2.
-    // See http://www.ginac.de/CLN/cln.git/?a=commit;h=4a477b0cc3dd7fbfb23b25090ff8c8869c8fa21a for details.
+    // See
+    // http://www.ginac.de/CLN/cln.git/?a=commit;h=4a477b0cc3dd7fbfb23b25090ff8c8869c8fa21a
+    // for details.
     size_t pos = s.find_first_not_of('0');
-    if(pos == std::string::npos) {
+    if (pos == std::string::npos)
+    {
       d_value = read_integer(flags, "0", NULL, NULL);
```
nafur, line 420:
```suggestion
      d_value = cln::read_integer(flags, "0", NULL, NULL);
```

src/util/integer_cln_imp.cpp
```diff
@@ -98,104 +407,159 @@ void Integer::readInt(const cln::cl_read_flags& flags,
...
-    } else {
+    }
+    else
+    {
       const char* cstr = s.c_str();
       const char* start = cstr + pos;
       const char* end = cstr + s.length();
       d_value = read_integer(flags, start, end, NULL);
```
nafur, line 427:
```suggestion
      d_value = cln::read_integer(flags, start, end, NULL);
```

nafur (approved):
Just a few minor improvements (using direct calls to `cln` instead of custom checks that may be a bit slower.

aniemetz:
@nafur I feel like there is room for improvement in terms of performance both in the CLN and BMP Integer implementation. I'll add the changes. Additionally, eventually we also should have a closer look at the GMP implementation with that regard.

aniemetz:
On second thought, let me add the changes in a new PR. This one only moves code.

PR #5307, opened 2020-10-19
(proof-new) Fixes for proofs in rewriter

ajreynol:
This PR fixes proofs in the rewriter so that we use attributes for marking whether a node has been rewritten with proofs instead of consulting the provided TConvProofGenerator for hasRewriteStep.  This additionally marks TRUST_REWRITE steps if a rewriter happens to be nondeterministic (e.g. quantifiers).  It furthermore decouples rewriteWithProof from reconstruction for theory rewrite steps.

The proof postprocessor is additionally updated so that extended equality REWRITE steps are converted to THEORY_REWRITE steps with identifier RW_REWRITE_EQ_EXT for consistency, since eliminating REWRITE should result in THEORY_REWRITE only.  This required generalizing the argument to THEORY_REWRITE to be a MethodId instead of a Boolean.

### DIFF
--- a/src/theory/rewriter.cpp
+++ b/src/theory/rewriter.cpp
@@ -31,6 +31,13 @@ using namespace std;
 namespace CVC4 {
 namespace theory {
 
+/** Attribute true for nodes that have been rewritten with proofs enabled */
+struct RewriteWithProofsAttributeId
+{
+};
+typedef expr::Attribute<RewriteWithProofsAttributeId, bool>
+    RewriteWithProofsAttribute;
+
 // Note that this function is a simplified version of Theory::theoryOf for
 // (type-based) theoryOfMode. We expand and simplify it here for the sake of
 // efficiency.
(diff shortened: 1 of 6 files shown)


src/theory/rewriter.cpp
```diff
@@ -195,7 +203,7 @@ Node Rewriter::rewriteTo(theory::TheoryId theoryId,
 
   // Check if it's been cached already
   Node cached = getPostRewriteCache(theoryId, node);
-  if (!cached.isNull() && (tcpg == nullptr || tcpg->hasRewriteStep(node)))
+  if (!cached.isNull() && (tcpg == nullptr || node.getAttribute(rpfa)))
```
HanielB, line 206: why is it done like this now?

ajreynol: This was actually incorrect if we registered a rewrite for a term without registering the proofs for its children.  We should only mark the attribute after the post-traversal of a term.

PR #5311, opened 2020-10-20
Handle rewrite to bool in BoundInference

nafur:
The BoundInference class did not properly handle rewrites that yield constant Booleans. This PR fixes this issue by explicitly checking for this case.


ajreynol (approved):
LGTM, is there a regression to add for this?

nafur:
> LGTM, is there a regression to add for this?

Unfortunately not one that is easily reproduced on master...

PR #5313, opened 2020-10-20
(proof-new) Use RESOLUTION in circuit propagator

nafur:
This PR updates the circuit propagator to use `RESOLUTION` (instead of `CHAIN_RESOLUTION`) where appropriate.


HanielB (approved):
LGTM

PR #5317, opened 2020-10-21
(proof-new) Updates to lazy proof chain

ajreynol:
Support for a default proof generator, which is optionally not called recursively.

This is required for preprocessing.


src/expr/lazy_proof_chain.cpp
```diff
@@ -207,7 +229,10 @@ std::shared_ptr<ProofNode> LazyCDProofChain::getProofFor(Node fact)
     // update each assumption proof node
     for (std::shared_ptr<ProofNode> pfn : it->second)
     {
-      d_manager->updateNode(pfn.get(), npfn.second.get());
+      if (npfn.second != nullptr)
+      {
+        d_manager->updateNode(pfn.get(), npfn.second.get());
+      }
```
HanielB, line 235: Since we require that that `npfn.second` is not null, revert this change.

HanielB (approved):
LGTM, just one minor request

PR #5301, opened 2020-10-19
(proof-new) Add proofs for circuit propagator

nafur:
This PR adds code to generate proofs for inferences within the (non-clausal) circuit propagator.
It consists of many small methods that each implement simple derivations about single Boolean connectives. It uses already existing proof rules, and thus no separate proof checker is required.
A second PR will use these rules within the circuit propagator class.

PR #5318, opened 2020-10-21
(proof-new) Make circuit propagator proof producing

nafur:
This PR uses the proofs from #5301 to actually produce proofs from the circuit propagator.


src/theory/booleans/circuit_propagator.h
```diff
@@ -107,9 +102,9 @@ class CircuitPropagator
    * discovered by the propagator are put in the substitutions vector used in
    * construction.
    *
-   * @return true iff conflict found
+   * @return a proof for a conflict, or nullptr otherwise
```
ajreynol, line 105:
```suggestion
   * @return a trust node encapsulating the proof for a conflict, or the null trust node otherwise
```

ajreynol: Comment that the trust node has kind LEMMA and proves false.

src/theory/booleans/circuit_propagator.h
```diff
@@ -95,9 +86,13 @@ class CircuitPropagator
 
   bool getNeedsFinish() { return d_needsFinish; }
 
-  std::vector<Node>& getLearnedLiterals() { return d_learnedLiterals; }
+  std::vector<TrustNode>& getLearnedLiterals() { return d_learnedLiterals; }
 
-  void finish() { d_context.pop(); }
+  void finish()
```
ajreynol, line 91: Can you move this implementation to cpp?  Also maybe comment on what this does.

ajreynol (changes requested):
A few minor requests, otherwise looks good.

ajreynol (approved):
LGTM

PR #5320, opened 2020-10-21
(proof-new) arith: dedup literals in flattenImpl

alex-ozdemir:
TheoryArithPrivte uses flattenImplication to turn implications into
clauses. This commit ensures that said clause does not have duplicate
terms. This is important because when we write proofs related to the
clause, the proof machinery doesn't like duplicates.


ajreynol (approved):
LGTM

PR #5322, opened 2020-10-21
Implement bags evaluator

mudathirmahgoub:
This PR implements NormalForm::evaluate for bags


src/theory/bags/normal_form.cpp
```diff
@@ -12,26 +12,628 @@
...
+  Assert(n.isConst()) << "node " << n << " is not in a normal form"
+                      << std::endl;
+  std::map<Node, Rational> elements;
+  if (n.getKind() == EMPTYBAG)
+  {
+    return elements;
+  }
+  while (n.getKind() == kind::UNION_DISJOINT)
+  {
+    Assert(n[0].getKind() == kind::MK_BAG);
```
mudathirmahgoub, line 178: https://github.com/CVC4/CVC4/pull/5294#discussion_r509273946
Perhaps this assertion is not necessary. `n` is asserted to be constant in line 169

src/theory/bags/normal_form.cpp
```diff
@@ -12,26 +12,628 @@
...
+    Node n =
+        nm->mkBag(elementType, it->first, nm->mkConst<Rational>(it->second));
+    bag = nm->mkNode(UNION_DISJOINT, n, bag);
+  }
+  return bag;
+}
+
+Node NormalForm::evaluateMakeBag(TNode n)
+{
+  Assert(n.getKind() == MK_BAG && !n.isConst()
```
mudathirmahgoub, line 215: > Can you give a comment here why this is the case? I assume it is because you only call this on non-constants.

Done

src/theory/bags/normal_form.cpp
```diff
@@ -12,26 +12,628 @@
...
+  // --------
+  // - (bag.is_singleton (emptyBag String)) = false
+  // - (bag.is_singleton (MK_BAG "x" 1)) = true
+  // - (bag.is_singleton (MK_BAG "x" 4)) = false
+  // - (bag.is_singleton (union_disjoint (MK_BAG "x" 1) (MK_BAG "y" 1))) = false
+
+  NodeManager* nm = NodeManager::currentNM();
+  Node falseNode = nm->mkConst(false);
+  Node trueNode = nm->mkConst(true);
+  if (n[0].getKind() == EMPTYBAG)
```
mudathirmahgoub, line 577: > Can remove these 4 lines.

What do you mean?

ajreynol: you don't need to check emptybag, this should just be part of the default "return false".

src/theory/bags/normal_form.cpp
```diff
@@ -12,26 +12,630 @@
...
+    case INTERSECTION_MIN: return evaluateIntersectionMin(n);
+    case DIFFERENCE_SUBTRACT: return evaluateDifferenceSubtract(n);
+    case DIFFERENCE_REMOVE: return evaluateDifferenceRemove(n);
+    case BAG_CHOOSE: return evaluateChoose(n);
+    case BAG_CARD: return evaluateCard(n);
+    case BAG_IS_SINGLETON: return evaluateIsSingleton(n);
+    case BAG_FROM_SET: return evaluateFromSet(n);
+    case BAG_TO_SET: return evaluateToSet(n);
+    default:
+    {
```
ajreynol, line 108: Can delete these braces.

src/theory/bags/normal_form.cpp
mudathirmahgoub: Done

src/theory/bags/normal_form.cpp
mudathirmahgoub: Done

ajreynol (approved):
LGTM

PR #5324, opened 2020-10-22
Apply alpha equivalence to annotated quantified formulas

ajreynol:
Previously, alpha equivalence was not applied to quantified formulas with attributes, likely motivated by keeping alpha equivalent quantified formulas with different user patterns.  It's not clear this is a good a idea.  Moreover, this also implies that quantified formulas with user-provided names (which happens in e.g. Boogie benchmarks) also do not have alpha equivalence applied.  

This makes it so that we apply alpha equivalence regardless of annotations.

FYI @barrettcw


HanielB (commented):
Neither of these cases sound like a good idea to me. I'd like to always reduce alpha-equivalent quantifiers to the same one. But what happens in the cases you said? If I have two alpha-equivalent quantifiers with different user-triggers, the triggers of one are just lost? If that's the case it'd be good to give a warning. We could also merge the triggers.

ajreynol:
The first one we see is the one that is kept.  There isn't really a way to "merge" triggers.  This would only be possible by introducing a third quantified formula with all triggers and saying the 3 are equivalent, which seems wasteful as well.

I don't think this warning should be on by default, the user probably doesn't care about this level of detail by default.  (Btw, these kind of issues may happen thousands of times for inputs e.g. with nested quantifiers that have triggers).

HanielB:
For user-provided triggers I don't think it's good to silently delete their triggers. But indeed maybe better to only warn with some kind of strict option, as this would probably be most useful to "benchmark cleaning".

ajreynol:
I can add a warning when -v is given, this is done via `Notice() <<`.

HanielB (approved):
LGTM now, thanks!

PR #5325, opened 2020-10-22
Change datatypes lemma policy for equalities not coming from instantiate

ajreynol:
Recently, the policy for lemma vs fact changed for instantiate rule in datatypes to ensure that other theories were notified of terms having external finite type. This PR further refines the policy so that the instantiate rule is the *only* rule whose conclusion is an equality that is sent out as a lemma.  

This means that equalities from unification and collapsing selectors are always kept internal.  The justification for this is that the instantiate rule is applied for all terms with the proper policy, and hence it already covers all cases where we may need to introduce new terms.


ajreynol:
Improved performance on the original Facebook set:

| Benchmark                    | old   | new   |
|------------------------------|-------|-------|
| AccountLimits.smt2           | 10.87 | 10.06 |
| Authenticator.smt2           | 0.18  | 0.18  |
| Compare.smt2                 | 0.06  | 0.06  |
| CoreAddresses.smt2           | 1.15  | 1.13  |
| DesignatedDealer.smt2        | 33.43 | 30.26 |
| DualAttestationLimit.smt2    | 0.55  | 0.52  |
| Empty.smt2                   | 0.03  | 0.03  |
| Event.smt2                   | 1.21  | 1.22  |
| FixedPoint32.smt2            | 0.19  | 0.19  |
| LibraBlock.smt2              | 9.49  | 7.37  |
| LibraTimestamp.smt2          | 2.79  | 3.31  |
| LibraTransactionTimeout.smt2 | 0.69  | 0.72  |
| LibraVersion.smt2            | 12.91 | 10.99 |
| LibraVMConfig.smt2           | 23.06 | 13.81 |
| Offer.smt2                   | 0.35  | 0.23  |
| Option.smt2                  | 1.41  | 1.22  |
| RegisteredCurrencies.smt2    | 72.31 | 45.41 |
| Roles.smt2                   | 9.97  | 9.32  |
| Signature.smt2               | 0.11  | 0.03  |
| Signer.smt2                  | 0.1   | 0.03  |
| Testnet.smt2                 | 3.13  | 1.92  |
| Unhosted.smt2                | 1.26  | 0.95  |
| ValidatorConfig.smt2         | 10.82 | 8.11  |
| Vector.smt2                  | 0.04  | 0.04  |

barrettcw (approved):
LGTM

PR #5326, opened 2020-10-22
Use theoryof-mode=type for QF_NRA

nafur:
We've observed that our QF_NRA solving benefits significantly from `--theoryof-mode=type` across the board (of other options that influence QF_NRA solving). This PR sets `type` as the new default for QF_NRA solving.


ajreynol (approved):
LGTM

PR #5328, opened 2020-10-22
Add bitwise refinement mode for IAND

makaimann:
Adds an option to do "bitwise" comparisons in the lazy IAND solver. Instead of creating an exact match for the value of a term using a sum, this would lazily fix groups of bits using integer extracts (divs and mods) when the abstract and concrete values differ at those bits.

For example, with a 1-bit granularity, you might learn a lemma like:

```
((_ iand 4) x y), value = 1
  actual (2, 3) = 2
  bv-value = #b0001
  bv-actual (#b0010, #b0011) = #b0010
IAndSolver::Lemma: (let ((_let_1 ((_ iand 4) x y))) 
         (and (and true 
                  (= (mod _let_1 2) (ite (and (= (mod x 2) 1) (= (mod y 2) 1)) 1 0))) 
                  (= (mod (div _let_1 2) 2) (ite (and (= (mod (div x 2) 2) 1) (= (mod (div y 2) 2) 1)) 1 0)))) 
        ; BITWISE_REFINE
```

which is just forcing the bottom two bits of the `iand` operator result to implement bitwise-AND semantics.


src/theory/arith/nl/iand_solver.cpp
```diff
@@ -215,17 +200,7 @@ Node IAndSolver::mkIOr(unsigned k, Node x, Node y) const
...
-  return ret;
-}
-
-Node IAndSolver::iextract(unsigned i, unsigned j, Node n) const
-{
-  NodeManager* nm = NodeManager::currentNM();
-  //  ((_ extract i j) n) is n / 2^j mod 2^{i-j+1}
-  Node n2j = nm->mkNode(INTS_DIVISION_TOTAL, n, twoToK(j));
-  Node ret = nm->mkNode(INTS_MODULUS_TOTAL, n2j, twoToK(i - j + 1));
+  Node ret = nm->mkNode(MINUS, d_iandTable.twoToKMinusOne(k), x);
```
yoni206, line 203: Maybe this function can also go to the helper file `iand_table`.

yoni206 (commented):
Looks good! Thanks for doing this, and also for cleaning up things.
Regarding the questions:

> does it make sense to have these things in IAndTable? The iextract and related code doesn't actually use the table
I think maybe we should rename `IAndTable`. The purpose of this file is to be a utility file that is shared between `iand_solver.cpp` and `bv_to_int.cpp`. Hopefully, most of the code for the bitwise option can be used for the preprocessing pass as well.

> should we have command lines for the different modes in every bv_to_int_* test?
No. For sure we need them for the `iand_native*` tests (I think there are two of these). And perhaps only for one or two of the `bv_to_int` ones.

> should I add the files from master...yoni206:bv_to_int_with_iand_merged ?
Do you mean the tests files? If so, then no. They are redundant. 

> I should probably support granularity: would you prefer that here or in a different PR?
I think here would be better. Another thing I would add to this branch is using the bit-wise option in the preprocessing pass. This way we can be sure that we are sharing the right functions between the lazy and eager modes. What do you think? I think with both granularities and bv_to_int support we will still be under 200 diff lines.

src/theory/arith/nl/iand_solver.cpp
makaimann: Just to clarify, are you thinking that all the `mk*` methods that are currently in `iand_solver` should go into the common file?

makaimann:
Hey @yoni206, I renamed `IAndTable` to `IAndUtils`, added support for granularities, and added a `BITWISE` option for the eager preprocessing pass as well. I think it's ready for another review whenever you have a chance. I asked one clarifying question above, and depending on your answer to that I'll make more changes. But otherwise, it's pretty close imo.

src/preprocessing/passes/bv_to_int.cpp
```diff
@@ -440,16 +440,38 @@ Node BVToInt::translateWithChildren(Node original,
...
+                                      options::BVAndIntegerGranularity());
+      }
+      else
+      {
+        Assert(options::solveBVAsInt() == options::SolveBVAsIntMode::BITWISE);
+        uint64_t granularity = options::BVAndIntegerGranularity();
+
+        Node x = translated_children[0];
+        Node y = translated_children[1];
+        returnNode = d_nm->mkNode(kind::IAND, x, y);
```
yoni206, line 460: `kind::IAND` is parametric, so u need to first create an operator and then use it in the node:
https://github.com/CVC4/CVC4/blob/a6c001e078767a2de6f36a2fa1333b98e39a6ec8/src/preprocessing/passes/bv_to_int.cpp#L426

src/preprocessing/passes/bv_to_int.cpp
```diff
@@ -440,16 +440,38 @@ Node BVToInt::translateWithChildren(Node original,
...
+        for (uint64_t j = 0; j < bvsize; j += granularity)
+        {
+          high_bit = j + granularity - 1;
+          // don't let high_bit pass bvsize
+          if (high_bit >= bvsize)
+          {
+            high_bit = bvsize - 1;
+          }
+          d_bitwiseAssertions.insert(returnNode.eqNode(
+              d_iandUtils.createBitwiseIAndNode(x, y, high_bit, j)));
```
yoni206, line 473: It seems to me like maybe the entire loop can be implemented inside `createBitwiseIAndNode`, as u have a similar loop in the `iand` solver. Is it possible to push more into that function?

src/theory/arith/nl/iand_solver.cpp
```diff
@@ -259,7 +234,53 @@ Node IAndSolver::sumBasedLemma(Node i)
...
+    // don't let high_bit pass bvsize
+    if (high_bit >= bvsize)
+    {
+      high_bit = bvsize - 1;
+    }
+
+    // check if the abstraction differs from the concrete one on these bits
+    if (bvAbsI.extract(high_bit, j) != bvConcI.extract(high_bit, j))
+    {
+      bitIAnd = d_iandUtils.createBitwiseIAndNode(x, y, high_bit, j);
```
yoni206, line 278: see comment above about `createBitwiseIAndNode` including the loop.

src/theory/arith/nl/iand_utils.cpp
```diff
@@ -146,7 +154,35 @@ Node IAndTable::createBitwiseNode(Node x,
...
+  uint64_t granularity = high - low + 1;
+  Assert(granularity <= 8);
+  // compute the table for the current granularity if needed
+  if (d_bvandTable.find(granularity) == d_bvandTable.end())
+  {
+    computeAndTable(granularity);
+  }
+  const std::map<std::pair<int64_t, int64_t>, uint64_t>& table =
+      d_bvandTable[granularity];
+  return createITEFromTable(
```
yoni206, line 171: I am a bit confused. in the bit-wise mode, isn't the idea to have a big conjunction rather than a big ite?

test/regress/regress3/bv_to_int_check_bvsgt_bvlshr0_4bit.smt2.minimized.smt2
```diff
@@ -1,6 +1,7 @@
 ; COMMAND-LINE:  --solve-bv-as-int=bv 
 ; COMMAND-LINE:  --solve-bv-as-int=sum     
 ; COMMAND-LINE:  --solve-bv-as-int=iand --iand-mode=sum    
+; COMMAND-LINE:  --solve-bv-as-int=iand --iand-mode=bitwise
```
yoni206, line 4: I would also suggest adding this to:
https://github.com/CVC4/CVC4/blob/master/test/regress/regress1/nl/iand-native-1.smt2
https://github.com/CVC4/CVC4/blob/master/test/regress/regress1/nl/iand-native-2.smt2

src/preprocessing/passes/bv_to_int.cpp
makaimann: Oops, thanks for catching that! Would it be hard to add a unit test for this pass? Maybe I should do that.

src/preprocessing/passes/bv_to_int.cpp
makaimann: Yeah that's a good point. It's very similar, the only difference is that it's lazy in the `iand` solver so it would need to be able to check for each range of bits whether or not the abstraction needs to be refined. I'm not sure what the right interface would be for that, so I just left it outside.

src/theory/arith/nl/iand_utils.cpp
makaimann: I'm not sure tbh, in the original branch that we based this on, it was an ITE. What were you envisioning for a big conjunction?

test/regress/regress3/bv_to_int_check_bvsgt_bvlshr0_4bit.smt2.minimized.smt2
makaimann: Thanks, added with 7ca81d943baa8118665670245fda4368461d66b1

src/preprocessing/passes/bv_to_int.cpp
makaimann: Fixed with 9720d2985d30aa3f864e0f3e8d5107f7bcdc417b

makaimann:
Looks like it's returning `unknown` for the `iand-native-1.smt2` test. I'm not sure if it's just because the encoding is not good, or if there's actually a bug. I printed out the lemmas and they make sense to me. It even has some of the same counterexamples as other methods, but then ends up returning unknown. Any thoughts on where to look into this more?

src/preprocessing/passes/bv_to_int.cpp
yoni206: ok that makes sense. maybe then document it and explicitly mention that it is lazy and in what sense?

makaimann:
Hey @ajreynol, sorry to bug you, but I think we could use your insight when you get a chance. This PR adds support for bitwise refinement (the lazy version was already implemented in an old experimental iand branch) for both lazy and eager modes. It also refactors a bit for better code sharing between the two modes. For some reason, I'm getting an unknown result when using the new bitwise refinement in the iand solver with:

`./bin/cvc4 --solve-bv-as-int=iand --iand-mode=bitwise ../test/regress/regress1/nl/iand-native-1.smt2`

Specifically it adds the same lemma twice and then hits [this case](https://github.com/CVC4/CVC4/blob/e7924fc22d220909c1b3eb0e41a1cdb624a69be7/src/theory/arith/nl/nonlinear_extension.cpp#L543). The log output using `-t iand-check` is attached here: [iand-check.log](https://github.com/CVC4/CVC4/files/5625475/iand-check.log). Interestingly, the eager version returns `sat` as expected: `./bin/cvc4 --solve-bv-as-int=bitwise ../test/regress/regress1/nl/iand-native-1.smt2`.

It's definitely possible I am misusing a data structure and/or made a mistake with the lemmas but it’s concerning that we get the same abstract counterexample. The added lemma does look right to me. The bitwise lemmas are produced [here](https://github.com/makaimann/CVC4/blob/bac2e3216cb70af764ea734b9d366a64539d1a79/src/theory/arith/nl/iand_solver.cpp#L241). Could you point me in the right direction for resolving this?

yoni206:
To clarify: 
In the bv-to-int preprocessing pass we translate `(bvand x y)` to `(iand k x' y')`, where `k` is the bit-width, and `x'` and `y'` are the integer translations of `x` and `y`.
Eager means that we add all the equations as extra assertions already in the preprocessing pass.
Lazy means that we don't add any extra assertions, and rely on the CEGAR loop at the iand solver.

ajreynol:
I believe there is an issue with whether the lemmas are being marked as "preprocess". 

For details, there is a flag that determines whether lemmas need to have preprocessing done on them:  https://github.com/CVC4/CVC4/blob/master/src/theory/output_channel.h#L41

For the lemmas you are emitting, this flag must be set.  This is due to the fact that the lemmas have div/mod in them.  If you use `-t theory::assertions-model` you will see that div/mod appear on the assertion stack.  

This should be easy to fix, I'm currently investigating.

ajreynol:
Indeed, this is the issue, unfortunately the lemma property field is currently hardcoded here: https://github.com/CVC4/CVC4/blob/master/src/theory/arith/inference_manager.cpp#L75
You'll have to thread the LemmaProperty::PREPROCESS through this method.  Let me know if you have any questions.
FYI @nafur

makaimann:
That did the trick, thanks @ajreynol! Much appreciated.

src/preprocessing/passes/bv_to_int.cpp
makaimann: Does this make sense? 33dca66629606eb6ecbf69077d5aedd2ed184080

makaimann:
After online discussion, we decided to move the eager mode part of this PR to a separate one. For now, I just reverted the relevant commits.

src/preprocessing/passes/bv_to_int.h
```diff
@@ -48,7 +48,7 @@
  ** Sigma_{i=0}^{b-1}(bvand s[(i+1)*g, i*g] t[(i+1)*g, i*g])*2^(i*g)
  **
  ** More details and examples for this case are described next to
- ** the function createBitwiseNode.
```
yoni206, line 51: Actually this line and the one before it can be deleted because this function has moved to another file.

src/theory/arith/nl/iand_solver.cpp
```diff
@@ -259,7 +246,53 @@ Node IAndSolver::sumBasedLemma(Node i)
...
+  Node lem = d_true;
+
+  // compare each bit to bvI
+  Node cond;
+  Node bitIAnd;
+  unsigned high_bit;
+  for (unsigned j = 0; j < bvsize; j += granularity)
+  {
+    high_bit = j + granularity - 1;
+    // don't let high_bit pass bvsize
```
yoni206, line 281: Is it always enough to decrement by 1? In the sum-based scheme I have a decrement loop.
If it is enough, can you add an assertion that says that the difference is at most 1?

src/theory/arith/nl/iand_utils.h
```diff
@@ -81,10 +84,34 @@ class IAndTable
...
    * @return A node that represents the operation, as described above.
    */
-  Node createBitwiseNode(Node x, Node y, uint64_t bvsize, uint64_t granularity);
+  Node createSumNode(Node x, Node y, uint64_t bvsize, uint64_t granularity);
+
+  /** Create a bitwise integer And node for two integers x and y for bits
+   * between hgih and low Example for high = 0, low = 0 (e.g. granularity 1)
+   *    ite(x[0] == 1 & y[0] == 1, #b1, #b0)
+   *
+   *  For higher granularities it makes use of computeAndTable
```
yoni206, line 93: Can you use `computeAndTable` for all granularities, including 1? Or is there any disadvantage?

test/regress/regress1/nl/iand-native-1.smt2
```diff
@@ -1,5 +1,6 @@
 ; COMMAND-LINE: --iand-mode=value --no-check-models
 ; COMMAND-LINE: --iand-mode=sum --bvand-integer-granularity=1 --finite-model-find --no-check-models
+; COMMAND-LINE:  --solve-bv-as-int=iand --iand-mode=bitwise
```
yoni206, line 3: I think we should test other granularities.
My suggestion is to add the following `COMMAND-LINE` to the following test:

https://github.com/CVC4/CVC4/blob/master/test/regress/regress3/bv_to_int_and_or.smt2

```
--solve-bv-as-int=iand --iand-mode=bitwise --bvand-integer-granularity=1
--solve-bv-as-int=iand --iand-mode=bitwise --bvand-integer-granularity=2
--solve-bv-as-int=iand --iand-mode=bitwise --bvand-integer-granularity=4
--solve-bv-as-int=iand --iand-mode=bitwise --bvand-integer-granularity=5
--solve-bv-as-int=iand --iand-mode=bitwise --bvand-integer-granularity=6
```
(the last one is intended to check whether decrementing by 1 is enough in a comment above).

yoni206 (changes requested):
Thanks for doing this! Looks good. A few comments.

src/theory/arith/nl/iand_solver.cpp
makaimann: Hmm... I'm not sure I follow. I'm probably missing something, what are you worried about? All I'm trying to handle is if your granularity surpasses the size when you count up. Say the granularity is 4 and this is bitwidth 6. Then first you get `extract 3 0`. Then you would get `extract 7 4` but then we just edit this to be `extract 5 4`.

yoni206: OK, I see. It should be OK. To make sure, maybe you could include tests with:
1. granularity 4 and bitwidth 6
2. granularity 5 and bitwidth 3

src/preprocessing/passes/bv_to_int.h
makaimann: Good catch, addressed here c4674e6d2b5717c44ae09a7d317c8c2cc9f2ad3d

src/theory/arith/nl/iand_utils.h
makaimann: Good point! It's actually just an incorrect comment, addressed with c4674e6d2b5717c44ae09a7d317c8c2cc9f2ad3d

test/regress/regress1/nl/iand-native-1.smt2
makaimann: Cool, added with b75799ab04315f836e963f352a415b9dbf4f593b

yoni206 (approved):
Thanks for doing this, and for all the changes! LGTM.

Issue #4370, opened 2020-04-21
Fatal failures at src/theory/uf/equality_engine_iterator.cpp:104


rainoftime:
Hi,
For the following formula:
~~~~
(declare-const v6 Bool)
(declare-const i4 Int)
(declare-const i7 Int)
(declare-const i10 Int)
(declare-const i11 Int)
(declare-fun st0 () (Set Int))
(declare-fun st1 () (Set Int))
(declare-fun st3 () (Set Int))
(declare-fun st4 () (Set Int))
(declare-fun st5 () (Set Int))
(declare-fun st7 () (Set Int))
(declare-fun st8 () (Set Int))
(declare-fun st12 () (Set Int))
(declare-const i12 Int)
(declare-const v19 Bool)
(declare-const i16 Int)
(declare-const i17 Int)
(declare-const v71 Bool)
(assert (> (card (intersection (union st7 st1) st1)) 0))
(assert (! (or (subset st1 st7) (subset st7 st4) (subset st3 st5)) :named IP_216))
(check-sat-assuming ((! (or v71 (= (- (abs i11)) 402) (=> v6 (<= 731 396))) :named IP_112) (! (or (or (subset (union (union st7 st1) st1) st7) (> (- i4 33 i4 0) 0) (= (xor true (distinct 24 (abs 449)) true true (> 85 396)) (<= 469 33) true false (< 731 396)) (< (* i17 i16 i4 (* (abs i11) (- i4 33 i4 0) i12 i7 i12)) 182) (subset st7 (intersection (intersection st4 st0) st12)) (subset (union (union st3 st8) st8) st5) (> 85 396) false (=> true false) false false) v19 (> 986 i10)) :named IP_116)))
~~~~

CVC4 throws out a fatal failure at equality_engine.cpp:2387 (--sets-ext)
~~~~
Fatal failure within CVC4::theory::eq::EqClassIterator& CVC4::theory::eq::EqClassIterator::operator++() at /home/peisen/test/tofuzz/CVC4/src/theory/uf/equality_engine.cpp:2387
Check failure

 d_start == d_ee->getEqualityNode(d_current).getFind()

Aborted (core dumped)

~~~~

OS: Ubuntu 16.04
Commit: ad7907adf

>

nafur:
It seems this issue has been fixed in the meantime:
```
$ bin/cvc4 --sets-ext test.smt2 
test.smt2:1.13: No set-logic command was given before this point.
[...]

PR #5331, opened 2020-10-23
Fix related to preregistering boolean term variables in strings

ajreynol:
We should only add trigger predicates for string predicates, and not arbitrary Boolean terms (which can now occur since we are handling parametric sequences). 

This avoids a debug assertion failure reported on as a followup to #4370.  In that benchmark BOOLEAN_TERM_VARIABLE was being added as a trigger predicate.


src/theory/strings/term_registry.cpp
```diff
@@ -203,8 +203,11 @@ void TermRegistry::preRegisterTerm(TNode n)
   }
   else if (tn.isBoolean())
   {
-    // Get triggered for both equal and dis-equal
-    ee->addTriggerPredicate(n);
+    if (k==STRING_STRCTN || k == STRING_LEQ)
```
yoni206, line 206: Perhaps you could document in a comment why you are restricting only to these two predicates and excluding the rest of the predicates?

yoni206 (approved):
LGTM. One comment for your consideration before merging.

PR #5332, opened 2020-10-23
[proof-new] Fix n-ary kind handling in EqEngine explanations

HanielB:
Adding missing cases of kinds that are n-ary but whose partial applications are
not true terms in the equality engine. This in the case not only for APPLY_UF
but also for the APPLY_* kinds for datatypes.


src/theory/uf/equality_engine.cpp
```diff
@@ -1739,7 +1745,7 @@ void EqualityEngine::addTriggerPredicate(TNode predicate) {
     return addTriggerEquality(predicate);
   }
   Assert(d_congruenceKinds.tst(predicate.getKind()))
-      << "No point in adding non-congruence predicates, kind is " << predicate.getKind();
```
ajreynol, line 1742: Revert this

HanielB: done

ajreynol (approved):

PR #5333, opened 2020-10-23
Proof new

Lachnitt:
This PR adds a veriT post processor and printer. It also introduces a veriT option to the proofFormatMode flag.


src/CMakeLists.txt
```diff
@@ -138,6 +138,9 @@ libcvc4_add_sources(
   proof/lean/lean_post_processor.cpp
   proof/lean/lean_post_processor.h
   proof/lean/lean_printer.h
+  proof/veriT/veriT_post_processor.cpp
```
ajreynol, line 141: As a style comment, I think we should use `verit` not `veriT` here.

src/options/smt_options.toml
```diff
@@ -108,6 +108,10 @@ header = "options/smt_options.h"
 [[option.mode.LEAN]]
   name       = "lean"
   help       = "Output Lean proof"
+[[option.mode.VERIT]]
+  name       = "veriT"
```
ajreynol, line 112: Ditto.

ajreynol (changes requested):
This looks great, one initial style comment.

src/expr/proof_checker.cpp
```diff
@@ -107,6 +107,11 @@ Node ProofChecker::check(
     Assert(expected.isNull() || expected == args[0]);
     return expected;
   }
+  if (id == PfRule::VERIT_RULE)
+  {
+    Assert(args.size() > 1);
```
HanielB, line 112:
```suggestion
    Assert(args.size() > 1 && args[0].getType().isReal());
```
Actually it'd be better to check "isInteger" but I'm not sure that by creating a rational constant this will work. You could try both?

src/expr/proof_rule.h
```diff
@@ -1104,6 +1104,14 @@ enum class PfRule : uint32_t
...
   LEAN_RULE,
 
+  //================================================ Place holder for veriT rules
+  // ======== veriT rule
+  // Children: (P1 ... Pn)
+  // Arguments: (id, Q, A1, ..., Am)
+  // ---------------------
+  // Conclusion: (Q)
+  VERIT_RULE,
+
```
HanielB, line 1114: Kill trailing space

src/smt/proof_manager.cpp
```diff
@@ -32,6 +34,8 @@ PfManager::PfManager(context::UserContext* u, SmtEngine* smte)
           d_pnm.get(), u, "smt::PreprocessProofGenerator")),
       d_pfpp(new ProofPostproccess(d_pnm.get(), smte, d_pppg.get())),
       d_lpfpp(new proof::LeanProofPostprocess(d_pnm.get())),
+      d_vpfppcb(d_pnm.get()),
+      d_vpfpp(new proof::VeriTProofPostprocess(d_pnm.get(),d_vpfppcb)),
```
HanielB, line 38:
```suggestion
      d_vpfpp(nullptr),
```
Them in the body of the constructor check whether the proof format is the veriT one, in which case you reset `d_vpfpp`. 

Another thing is that I'd like the proof manager to only have a link to the postprocessor. I understand the callback as an attribute is necessary because of how you made the postprocessor extend proof node updater. I now think that it's better not to do it and follow the pattern of the other postprocessors (their own class, whose `process` method uses a local proof node updater).

src/CMakeLists.txt
HanielB: Agreed. The directory should also be called `verit`.

src/proof/veriT/veriT_proof_rule.h
```diff
@@ -0,0 +1,141 @@
...
+#ifndef CVC4__PROOF__VERIT_PROOF_RULE_H
+#define CVC4__PROOF__VERIT_PROOF_RULE_H
+
+
+namespace CVC4 {
+
+namespace proof {
+
+enum class VeriTRule : uint32_t
+{
```
HanielB, line 28: Kill trailing space.

src/proof/veriT/veriT_proof_rule.h
```diff
@@ -0,0 +1,141 @@
...
+  // Arguments:
+  // --------------
+  // Conclusion: F
+  //
+  //Each subproof in veriT begins with an anchor command. The outermost application of 
+  //anchor will not be printed.
+  ANCHOR,
+  ASSUME,
+  UNDEFINED, //TODO: delete later
+
```
HanielB, line 43: ditto

src/proof/veriT/veriT_proof_rule.h
```diff
@@ -0,0 +1,141 @@
...
+  DIV_SIMPLIFY,
+  PROD_SIMPLIFY,
+  UNARY_MINUS_SIMPLIFY,
+  MINUS_SIMPLIFY,
+  SUM_SIMPLIFY,
+  COMP_SIMPLIFY,
+  NARY_ELIM,
+  TMP_AC_SIMP,
+  TMP_BFUN_ELIM,
+  TMP_SKOLEMIZE, //SHOULD NOT BE USED, SO DELETE IT AT SOME POINT?
```
HanielB, line 127: I don't think we will need it.

src/proof/veriT/veriT_proof_rule.cpp
```diff
@@ -0,0 +1,127 @@
...
+namespace proof {
+
+//TODO: Put in cpp, rename, change static
+const char* veriTRuletoString(VeriTRule id)
+{
+  switch (id)
+  {
+  case VeriTRule::UNDEFINED: return "undefined";
+  case VeriTRule::ASSUME: return "assume";
+  case VeriTRule::ANCHOR: return "anchor";
```
HanielB, line 29: Kill trailing space

src/proof/veriT/veriT_proof_rule.cpp
```diff
@@ -0,0 +1,127 @@
...
+  case VeriTRule::UNDEFINED: return "undefined";
+  case VeriTRule::ASSUME: return "assume";
+  case VeriTRule::ANCHOR: return "anchor";	
+  case VeriTRule::INPUT: return "input";
+  case VeriTRule::TRUE: return "true";
+  case VeriTRule::FALSE: return "false";
+  case VeriTRule::NOT_NOT: return "not_not";
+  case VeriTRule::AND_POS: return "and_pos";
+  case VeriTRule::AND_NEG: return "and_neg";    
+  case VeriTRule::OR_POS: return "or_pos";
```
HanielB, line 36: ditto

src/proof/veriT/veriT_proof_rule.cpp
```diff
@@ -0,0 +1,127 @@
...
+  case VeriTRule::TRUE: return "true";
+  case VeriTRule::FALSE: return "false";
+  case VeriTRule::NOT_NOT: return "not_not";
+  case VeriTRule::AND_POS: return "and_pos";
+  case VeriTRule::AND_NEG: return "and_neg";    
+  case VeriTRule::OR_POS: return "or_pos";  
+  case VeriTRule::OR_NEG: return "or_neg";
+  case VeriTRule::XOR_POS1: return "xor_pos1";    
+  case VeriTRule::XOR_POS2: return "xor_pos2";    
+  case VeriTRule::XOR_NEG1: return "xor_neg1";
```
HanielB, line 40: ditto

src/proof/veriT/veriT_proof_rule.cpp
```diff
@@ -0,0 +1,127 @@
...
+
+namespace CVC4 {
+
+namespace proof {
+
+//TODO: Put in cpp, rename, change static
+const char* veriTRuletoString(VeriTRule id)
+{
+  switch (id)
+  {
```
HanielB, line 26: This should be clang-formatted.

src/proof/veriT/veriT_printer.h
```diff
@@ -0,0 +1,115 @@
...
+namespace proof {
+
+
+//TEMP ad-hoc code to test conversion
+static int veriTPrintInternal(std::ostream& out,
+			      std::shared_ptr<ProofNode> pfn, int i){
+  int temp = i;
+  std::vector<int> childIds;
+  i++;
+
```
HanielB, line 36: kill trailing space

src/proof/veriT/veriT_post_processor.h
```diff
@@ -0,0 +1,98 @@
...
+{
+ public:
+  VeriTProofPostprocessCallback(ProofNodeManager* pnm);
+  ~VeriTProofPostprocessCallback() {}
+  /**
+   * Initialize, called once for each new ProofNode to process. This initializes
+   * static information to be used by successive calls to update.
+   */
+  void initializeUpdate();
+  bool shouldUpdate(std::shared_ptr<ProofNode> pn,
```
HanielB, line 44: kill trailing space

src/proof/veriT/veriT_post_processor.h
```diff
@@ -0,0 +1,98 @@
...
+		    const std::vector<Node>& children,
+		    const std::vector<Node>& args,
+		    CDProof& cdp);
+};
+
+/**
+ * The proof postprocessor module. This postprocesses a proof node into one
+ * using the rules from the veriT calculus.
+ */
+class VeriTProofPostprocess : public ProofNodeUpdater
```
HanielB, line 78: kill trailing space

src/proof/veriT/veriT_post_processor.h
```diff
@@ -0,0 +1,98 @@
...
+
+#include "expr/proof_node_updater.h"
+#include "proof/veriT/veriT_proof_rule.h"
+
+namespace CVC4 {
+
+namespace proof {
+
+/**
+ * A callback class used by the veriT convereter for post-processing proof nodes
```
HanielB, line 31:
```suggestion
 * A callback class used by the veriT converter for post-processing proof nodes
```

src/proof/veriT/veriT_post_processor.h
```diff
@@ -0,0 +1,98 @@
...
+  void initializeUpdate();
+  bool shouldUpdate(std::shared_ptr<ProofNode> pn, 	
+		    bool& continueUpdate) override;
+  /** Update the proof rule application. */
+  bool update(Node res,
+              PfRule id,
+              const std::vector<Node>& children,
+              const std::vector<Node>& args,
+              CDProof* cdp,
+	      bool& continueUpdate) override;
```
HanielB, line 52: I see that you are using tabs. :) Clang-format should remove them (and I think that trailing spaces as well, at least in .cpp/.h files) but still, maybe better to avoid introducing them in the first place.

src/proof/veriT/veriT_post_processor.h
```diff
@@ -0,0 +1,98 @@
...
+ private:
+  /** The proof node manager */
+  ProofNodeManager* d_pnm;
+  NodeManager* d_nm;
+  /**
+   * This method adds a new step to the proof applying the veriT rule.
```
HanielB, line 58: Clarify why the method is necessary. Its main reason for existence is to add the proper number id to which the `rule` corresponds in the enumeration when adding the step, right?

src/proof/veriT/veriT_post_processor.h
```diff
@@ -0,0 +1,98 @@
...
+  VeriTProofPostprocessCallback(ProofNodeManager* pnm);
+  ~VeriTProofPostprocessCallback() {}
+  /**
+   * Initialize, called once for each new ProofNode to process. This initializes
+   * static information to be used by successive calls to update.
+   */
+  void initializeUpdate();
+  bool shouldUpdate(std::shared_ptr<ProofNode> pn, 	
+		    bool& continueUpdate) override;
+  /** Update the proof rule application. */
```
HanielB, line 46: Say how this method is doing the update, namely that it is splitting on the given rule and translating it into a proof node in terms of the veriT rule.

src/proof/veriT/veriT_post_processor.cpp
```diff
@@ -0,0 +1,1163 @@
...
+    std::cout << "," << children[i];
+  }
+  std::cout << "},{";
+
+  if(args.size() > 0)
+    std::cout << args[0];
+
+  for(int i=1;i<args.size();i++){
+    std::cout << "," << args[i];
+  }
```
HanielB, line 78: Just FYI, you can give vector of nodes directly to the printer. They will be printed as `[ n0 ... nm ]`.

src/proof/veriT/veriT_post_processor.cpp
```diff
@@ -0,0 +1,1163 @@
...
+    // premises: VP1 VP2
+    // args: ()
+    case PfRule::SPLIT:{
+      Node vp1 = d_nm->mkNode(kind::OR,(d_nm->mkNode(kind::NOT,d_nm->mkNode(kind::NOT,d_nm->mkNode(kind::NOT,args[0]))),args[0]));
+      Node vp2 = d_nm->mkNode(kind::OR,(d_nm->mkNode(d_nm->mkNode(kind::NOT,d_nm->mkNode(kind::NOT,d_nm->mkNode(kind::NOT,args[0]))),d_nm->mkNode(kind::NOT,args[0])))); 
+      
+      return addVeriTStep(vp2,VeriTRule::NOT_NOT,{},{},*cdp) 
+	     && addVeriTStep(vp1,VeriTRule::NOT_NOT,{},{},*cdp) 
+             && addVeriTStep(res,VeriTRule::RESOLUTION,{vp1,vp2},{},*cdp);
+    }
```
HanielB, line 246: Kind of beautiful! I like how we derive excluded-middle via double-negation elimination. :) Made me remember of starting to learn logic.

src/proof/veriT/veriT_post_processor.cpp
```diff
@@ -0,0 +1,1163 @@
...
+    // ...
+    //
+    // proof rule: resolution
+    // proof term: (VP:(or (and F1 ... Fn))
+    // premises: (VPn:(or (and F1 ... Fn) (not Fn))) Fn
+    // args: () 
+    case PfRule::AND_INTRO:{
+      bool success = true;
+      std::vector<Node> neg_Nodes;
+      neg_Nodes.push_back(d_nm->mkNode(kind::AND,children));
```
HanielB, line 364: you could use `res` directly here rather than creating the conjunction, right?

src/proof/veriT/veriT_post_processor.cpp
```diff
@@ -0,0 +1,1163 @@
...
+      for(int i = 0; i < children.size(); i++){
+        neg_Nodes.push_back(d_nm->mkNode(kind::NOT,children[i]));
```
HanielB, line 366:
```suggestion
        neg_Nodes.push_back(children[i].notNode());
```

src/proof/veriT/veriT_post_processor.cpp
```diff
@@ -0,0 +1,1163 @@
...
+      }
+
+      Node and_neg_Node = d_nm->mkNode(kind::OR,neg_Nodes);	
+      success = success && addVeriTStep(and_neg_Node,VeriTRule::AND_NEG,{},{},*cdp);
```
HanielB, line 370: If you failed here you'd just return `false`, no? No need for the `success` variable.

src/proof/veriT/veriT_post_processor.cpp
```diff
@@ -0,0 +1,1163 @@
...
+
+      //TODO: It seems as if all this is unnecessary since the resolution rule in VeriT can do a 
+      //chain or resolution steps. Thus, this simplifies to a single resolution step and there is no need 
+      //to build up all these clauses.
```
HanielB, line 374: That's right. You could directly do a resolution in which the first premise is `and_neg_Node` and the others are the elements in `children`. One thing that we should aim for with the use of `VeriTRule::RESOLUTION` is to keep the correct order. This is one the first things that should be updated in the proofnomicon (as not being able to assume the proper order in resolution makes it NP-complete to check...).

HanielB (commented):
This is great, @Lachnitt ! I have some miscellaneous minor comments. Other than that the main thing is to apply clang-format in the PR.

src/CMakeLists.txt
Lachnitt: Done

src/proof/veriT/veriT_post_processor.h
Lachnitt: Yes, since these steps (adding the id, conclusion and optionally add other arguments to a vector) have to be executed for nearly every proof rule this functionality is encapsulated in this function. I might extend it to add SEXPR  to differentiate  (cl F1 F2) and (cl (or F1 F2)).

src/smt/proof_manager.cpp
Lachnitt: @scottviteri this is something also concerning lean

Lachnitt: Done

src/proof/veriT/veriT_post_processor.cpp
Lachnitt: Yes, sorry AND_INTRO was not done yet.

src/proof/veriT/veriT_post_processor.cpp
Lachnitt: @HanielB What is the proper order? E.g. in the AND_INTRO case:
`(or (and F1 ... Fn) (not F1) ... (not Fn))`
There is no way to reorder this to:
`(or (not F1) ... (not Fn) (and F1 ... Fn))`
What is possible (not necessary here but ,maybe somewhere else) is to reorder the order of the children (which are the premises in veriT), so if you had
`(or (and F1 ... Fn) (not F2) (not F1) ... (not Fn))`
then  
`Children: (P2:F2 P1:F1 ... Pn:Fn)`

HanielB: This is a good question. :) The "order" I was worried about was when there are more than two clauses being resolved and you have C1 resolved w/ C2 resolved w/ C3 recorded in the proof as C2, C1, C3, say. The  "order" you exemplified above is for the literals in the clauses themselves, which is less serious (if clauses are viewed as sets, this does not add overhead, while in my example the issue is search the correct clause to resolve).

Currently the veriT proof format is agnostic to both "orders", but IIRC the Isabelle reconstruction assumes the correct order for the clauses being resolved, most of the time. So we should respect that. Does that make sense?

src/expr/proof_checker.cpp
```diff
@@ -107,6 +107,11 @@ Node ProofChecker::check(
     Assert(expected.isNull() || expected == args[0]);
     return expected;
   }
+  if (id == PfRule::VERIT_RULE)
```
ajreynol, line 110: This shouldn't be here, let's move this to theory/builtin/proof_checker.cpp.

Lachnitt: Done

src/expr/proof_checker.cpp
Lachnitt: isInteger works.

ajreynol (approved):
LGTM, I think we should merge this to proof-new now.  @HanielB  if you have more suggestions later we can either work on top of proof-new or address when the code is merged to master.

HanielB:
Sounds good to me. I was going to merge it anyways. :)

PR #5334, opened 2020-10-24
Refactor DeclareSygusVarCommand and SynthFunCommand to use the API.

abdoo8080:
This PR is part of migrating commands. `DeclareSygusVarCommand` and `SynthFunCommand` now call public API function instead of their corresponding SmtEngine functions. Those two commands don't fully initialize the solver anymore. Some operations in `SygusInterpol::solveInterpolation`, which creates an interpolation sub-solver, depend on the solver being fully initialized and were affected by this change. Those operations are now executed by the main solver instead of the sub-solver, which is not fully initialized by the time they are needed.


src/theory/quantifiers/sygus/sygus_interpol.cpp
```diff
@@ -319,6 +319,13 @@ bool SygusInterpol::solveInterpolation(const std::string& name,
                                        const TypeNode& itpGType,
                                        Node& interpol)
 {
+  collectSymbols(axioms, conj);
```
ajreynol, line 322: Can you give a quick comment on what this is done here (based on what you said in the PR)?

ajreynol (changes requested):
One minor change, otherwise looks good to me.

src/theory/quantifiers/sygus/sygus_interpol.cpp
abdoo8080: OK. I wrote an explanation for reordering the instructions. Do you think it's clear?

ajreynol (approved):
LGTM

PR #5338, opened 2020-10-24
(proof-new) Fix true explanations of propagations in pf equality engine

ajreynol:
This ensures we construct proper proofs for propagations whose conclusions are of the form (=> true lit).  Literals may be propagated with "true" as an explanation in datatypes, e.g. discriminators for datatypes with 1 constructor.


src/theory/uf/proof_equality_engine.cpp
```diff
@@ -408,9 +408,20 @@ TrustNode ProofEqEngine::ensureProofForFact(Node conc,
...
   }
-  // scope the proof constructed above, and connect the formula with the proof
-  // minimize the assumptions
+  // Scope the proof constructed above, and connect the formula with the proof
+  // minimize the assumptions.
   pf = d_pnm->mkScope(pfBody, scopeAssumps, true, true);
+  // If we have no assumptions, and are proving an explanation for propagation
+  if (scopeAssumps.empty() && tnk == TrustNodeKind::PROP_EXP)
+  {
+    // Must add "true" as an explicit argument. This is to ensure that the
```
HanielB, line 417: Say here why we have to do this.

HanielB (approved):
LGTM, just a couple documentation comments.

src/theory/uf/proof_equality_engine.cpp
ajreynol: There is a full paragraph explaining this here?

HanielB: The next line says that you have to generate `(=> true F)`. But it does not say why this should be the case.

PR #5339, opened 2020-10-24
(proof-new) Proofs for skolemization

ajreynol:
This adds support for proofs in the quantifiers module that performs skolemization.

Also fixes a bug in the proof checker for skolemization.


src/theory/quantifiers/proof_checker.cpp
```diff
@@ -97,6 +97,7 @@ Node QuantifiersProofRuleChecker::checkInternal(
     else
     {
       std::vector<Node> echildren(children[0][0].begin(), children[0][0].end());
+      echildren[1] = echildren[1].notNode();
```
HanielB, line 100: what is this for?

ajreynol: The rule was incorrect before, it was converting `~forall x. P` to `exists x. P` when it should have been `exists x. ~P`.

HanielB: Ah, I see.

HanielB (approved):

PR #5340, opened 2020-10-24
(proof-new) Add datatypes proof checker

ajreynol:
This adds the proof rules and proof checker for datatypes.

### DIFF
--- a/src/expr/proof_rule.h
+++ b/src/expr/proof_rule.h
@@ -681,22 +681,25 @@ enum class PfRule : uint32_t
   // Arguments: (i)
   // ----------------------------------------
   // Conclusion: (= ti si)
+  // where C is a constructor.
   DT_UNIF,
   // ======== Instantiate
   // Children: none
   // Arguments: (t, n)
   // ----------------------------------------
   // Conclusion: (= ((_ is C) t) (= t (C (sel_1 t) ... (sel_n t))))
-  // where C is the n^th constructor of the type of T.
+  // where C is the n^th constructor of the type of T, and (_ is C) is the
+  // discriminator (tester) for C.
   DT_INST,
   // ======== Collapse
   // Children: none
   // Arguments: ((sel_i (C_j t_1 ... t_n)))
   // ----------------------------------------
   // Conclusion: (= (sel_i (C_j t_1 ... t_n)) r)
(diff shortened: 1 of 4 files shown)


src/expr/proof_rule.h
```diff
@@ -674,6 +674,49 @@ enum class PfRule : uint32_t
   // ---------------------
   // Conclusion: F
   ARRAYS_TRUST,
+  
+  //================================================= Datatype rules
+  // ======== Unification
+  // Children: (P:(= (C t1 ... tn) (C s1 ... sn)))
+  // Arguments: (i)
+  // ----------------------------------------
+  // Conclusion: (= ti si)
```
nafur, line 683: Could you add a bit of general information here? I guess it mostly boils done to a very basic explanation of what DT is all about. I guess `C` stands for "Constructor"?

src/theory/datatypes/proof_checker.cpp
```diff
@@ -0,0 +1,135 @@
...
+        || children[0][1].getKind() != kind::APPLY_CONSTRUCTOR
+        || children[0][0].getOperator() != children[0][1].getOperator()
+        || !getUInt32(args[0], i))
+    {
+      return Node::null();
+    }
+    if (i >= children[0][0].getNumChildren())
+    {
+      return Node::null();
+    }
```
nafur, line 56: Why did you put the last condition in a separate if?

nafur (approved):
LGTM, but I'd really like a bit more documentation on the proof rules.

src/theory/datatypes/proof_checker.cpp
ajreynol: To me it gets a little weird following the code if one condition above (`getUInt32(args[0], i)`) is responsible for setting `i` and then another condition uses it.  Should work but relies on order of operations within an if-condition, anyways, for clarity I've broken at that point.

nafur: I see. So first of: this is safe as the evaluation order for `||` and `&&` is left-to-right (to allow for short-circuiting, in contrast to most other operator / function evaluations). If we find this confusing, we should find the first condition to be confusing as well, though, as `children[0].getKind() != kind::EQUAL` ensures that the access to `children[0][0]` is safe.
No strong opinion on my side, though...

PR #5345, opened 2020-10-26
Enable --nl-cad by default

nafur:
This PR enables `--nl-cad` for QF_NRA (and QF_UFNRA) by default to improve nonlinear arithmetic solving.
Furthermore, it takes care of disabling it when libpoly is not available. It also adds a fix to the CadSolver that avoids incorrect SATs in the presence of theory combination.


nafur:
Benchmarks for master vs. this PR with 20min timeout:
```
                              status  result  solved   sat  unsat  best  timeout  memout  error  uniq  disagr    time_cpu     memory
benchmark config                                                                                                                    
QF_NRA    10230124-all-cad        ok   11489    9120  3698   5422  8125     2262     107      0   518       0   2884160.8  1791211.4
          10230124-all-master     ok   11489    8602  3211   5391   995     2559     107      0     0       0   3234834.1  1789011.5
QF_UFNRA  10230124-all-cad        ok      37      20     9     11    16       17       0      0     0       0     22121.1    25816.8
          10230124-all-master     ok      37      20     9     11     4       16       0      0     0       0     21110.2    25817.7

```

src/smt/set_defaults.cpp
```diff
@@ -1437,6 +1437,36 @@ void setDefaults(LogicInfo& logic, bool isInternalSubsolver)
...
+    }
+#endif
+  }
+#ifndef CVC4_USE_POLY
+  if (options::nlCad())
+  {
+    if (options::nlCad.wasSetByUser())
+    {
+      std::stringstream ss;
+      ss << "Cannot use " << options::nlCad.getName() << " without --poly.";
```
mpreiner, line 1458:
```suggestion
      ss << "Cannot use " << options::nlCad.getName() << " without configuring with --poly.";
```

src/smt/set_defaults.cpp
```diff
@@ -1437,6 +1437,36 @@ void setDefaults(LogicInfo& logic, bool isInternalSubsolver)
...
+      throw OptionException(ss.str());
+    }
+    else
+    {
+      Notice() << "Cannot use --" << options::nlCad.getName()
+               << " without --poly." << std::endl;
```
mpreiner, line 1464:
```suggestion
               << " without configuring with --poly." << std::endl;
```

mpreiner (approved):
LGTM, two minor suggestions.

nafur:
There are two regressions failing. The first issue already exists on master and is merely uncovered by this PR. It relates to IAND and was communicated to @yoni206 already.
The other seems to originate from the fact that we now produce proper models for nonlinear problems. Instead of `SUCCESS` we now print `((x (witness ((BOUND_VARIABLE_580 Real)) (and (= (* __z __z) 2.0) (not (>= (* (- 1.0) __z) (/ (- 5) 4))) (not (>= __z (/ 3 2)))))))` in response to a `(get-value (x))` query.

PR #5347, opened 2020-10-27
Make mkGroundTerm deterministic

ajreynol:
This ensures mkGroundTerm is deterministic using attributes.  

This was uncovered by failures in the proof checker for datatypes.


src/theory/builtin/theory_builtin_type_rules.cpp
```diff
@@ -0,0 +1,50 @@
+/*********************                                                        */
+/*! \file theory_builtin_type_rules.cpp
+ ** \verbatim
+ ** Top contributors (to current version):
+ **   Morgan Deters, Andrew Reynolds, Dejan Jovanovic
```
yoni206, line 5:
```suggestion
 ** Andrew Reynolds
```

yoni206 (approved):
LGTM. One comment inside + a question: Do you think you could extract the failure discovered in the proof checker for datatypes into a test that ensures that this issue is fixed? If not, I guess such a test will exist once proof-new is merged, but it would be nice to have a test specifically for this, if that's possible.

ajreynol:
The failure is not really isolated.  This is a prerequisite for getting any application of the "collapse selector" rule for datatypes working.  Once proof-new is merged to master and enabled in our regressions, this will be tested numerous times on `make regress`.

PR #5350, opened 2020-10-27
Disable --nl-cad option by default

nafur:
This PR disables the `--nl-cad` option by default. It was erroneously enabled with #5345 but leads to errors on, e.g., QF_NIA. Enabling for QF_NRA is done in set_defaults.


ajreynol (approved):

ajreynol (changes requested):
There is a timeout on CI

ajreynol (approved):

PR #5352, opened 2020-10-27
Add rewrites for div/mod in the arithmetic rewriter

ajreynol:
This adds some basic rewrites for integer div/mod in the rewriter.

This is in preparation for improved preprocessing and rewriting for NIA problems with heavy use of div/mod.


src/theory/arith/rewrites.h
```diff
@@ -0,0 +1,68 @@
...
+namespace theory {
+namespace arith {
+
+/**
+ * Types of rewrites used by arithmetic
+ */
+enum class Rewrite : uint32_t
+{
+  NONE,
+  CONST_EVAL,
```
yoni206, line 32: How do u feel about documenting the rules here as ___->___ rules (similar to the cpp)?

src/theory/arith/arith_rewriter.cpp
```diff
@@ -763,42 +761,67 @@ RewriteResponse ArithRewriter::rewriteDiv(TNode t, bool pre){
...
+    }
+  }
+  return RewriteResponse(REWRITE_DONE, t);
+}
 
-  //Leaving the function as before (INTS_MODULUS can be handled),
-  // but restricting its use here
+RewriteResponse ArithRewriter::rewriteIntsDivModTotal(TNode t, bool pre)
+{
+  if (pre)
```
yoni206, line 792: Is this a standard pattern in the rewriter? If not (ant maybe even if yes) can you document something about it?

src/theory/arith/rewrites.cpp
```diff
@@ -0,0 +1,50 @@
...
+    case Rewrite::MOD_TOTAL_BY_CONST: return "MOD_TOTAL_BY_CONST";
+    case Rewrite::DIV_TOTAL_BY_CONST: return "DIV_TOTAL_BY_CONST";
+    case Rewrite::DIV_MOD_BY_ZERO: return "DIV_MOD_BY_ZERO";
+    case Rewrite::MOD_BY_ONE: return "MOD_BY_ONE";
+    case Rewrite::DIV_BY_ONE: return "DIV_BY_ONE";
+    case Rewrite::DIV_MOD_PULL_NEG_DEN: return "DIV_MOD_PULL_NEG_DEN";
+    case Rewrite::MOD_OVER_MOD: return "MOD_OVER_MOD";
+    case Rewrite::MOD_CHILD_MOD: return "MOD_CHILD_MOD";
+    case Rewrite::DIV_OVER_MOD: return "DIV_OVER_MOD";
+    default: return "?";
```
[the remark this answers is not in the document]
ajreynol: If the value is unexpected or uninitialized.  This is never expected to be reached, but indicates that something went wrong.

src/theory/arith/rewrites.h
ajreynol: Done.

src/theory/arith/arith_rewriter.cpp
ajreynol: Its not standard but the arithmetic rewriter does it.  This is combining the prerewrite and postrewrites into one function, with a Boolean flag to indicate which we are in. I'll comment.

src/theory/arith/rewrites.cpp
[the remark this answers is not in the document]
yoni206: Then maybe Assert false before returning `?` so that we at least catch this in debug?

yoni206 (approved):
Thanks for the changes. LGTM. 
I have made one more followup comment about the switch block, for your consideration before merging.

PR #5355, opened 2020-10-27
Convert symbol table from Expr-level to Term-level

ajreynol:
This task is left over from parser migration.

This PR also drops support for a case of symbol overloading, in particular symbols (constructors, selectors) for parametric datatypes cannot be overloaded.  One regression is disabled as a result.


abdoo8080 (approved):
LGTM!

PR #5356, opened 2020-10-27
(proof-new) Separate explanation stages in proof equality engine

ajreynol:
This fixes potential cycles in assertLemma commands in the proof equality engine by using separate proof data structures for the topmost 2 stages of proof generation for equality engine proofs.  This fix is required to support datatype proofs for lemmas.

This PR also removes support for the ProofStepBuffer interface in ProofEqEngine, since it is unused, and suffers similar issues and would require a different unique fix.


ajreynol:
Note: relies on https://github.com/CVC4/CVC4/pull/5338, review that first.

HanielB (approved):
LGTM

PR #5361, opened 2020-10-29
(proof-new) Proofs for regular expression elimination

ajreynol:
Adds support for proofs for regular expression elimination in TheoryStrings::ppRewrite via a coarse-grained rule RE_ELIM.
This rule is updated by this PR to distinguish whether we apply aggressive elimination. Aggressive elimination is not currently supported, since it is non-deterministic due to generating fresh BOUND_VARIABLE.


4tXJ7f (approved):
LGTM

PR #5363, opened 2020-10-29
Update api::Sort to use TypeNode instead of Type

ajreynol:
This is work towards removing the old API.

This makes TypeNode the backend for Sort instead of Type.

It also updates a unit test for methods `isUninterpretedSortParameterized` and `getUninterpretedSortParamSorts` whose implementation was previously buggy due to the implementation of Type-level `SortType`.


src/api/cvc4cpp.cpp
```diff
@@ -49,6 +49,7 @@
 #include "expr/node_manager.h"
 #include "expr/sequence.h"
 #include "expr/type.h"
```
aniemetz, line 51: Can we remove this? Or not yet?

src/api/cvc4cpp.cpp
```diff
@@ -945,13 +946,26 @@ std::ostream& operator<<(std::ostream& out, const Result& r)
 /* -------------------------------------------------------------------------- */
 
 Sort::Sort(const Solver* slv, const CVC4::Type& t)
```
aniemetz, line 948: Shouldn't this argument be a `TypeNode`?

src/api/cvc4cpp.cpp
```diff
@@ -1077,52 +1081,54 @@ std::string Sort::toString() const
...
+}
+const CVC4::TypeNode& Sort::getTypeNode(void) const { return *d_type; }
 
 /* Constructor sort ------------------------------------------------------- */
 
 size_t Sort::getConstructorArity() const
 {
   CVC4_API_CHECK(isConstructor()) << "Not a constructor sort: " << (*this);
-  return ConstructorType(*d_type).getArity();
+  return d_type->getNumChildren() - 1;
```
aniemetz, line 1097: Why the `-1`?

src/api/cvc4cpp.cpp
ajreynol: Not yet, there is a annoying case where we need to use ExprManager mkVar + flags in the parser, which requires passing a type.

src/api/cvc4cpp.cpp
ajreynol: There are current interfaces for both Type and TypeNode, due to the issue I mentioned in the previous comment.  (This is the same situation we are currently in with Term which has constructors for Expr/Node).

This will take a few more PRs to eliminate.

src/api/cvc4cpp.cpp
ajreynol: The last child is the return type, the arity is the number of argument types.

aniemetz (approved):

PR #5366, opened 2020-10-30
Remove mkSingleton from the API

mudathirmahgoub:
This PR removes mkSingleton from the API after removing subtyping from set theory and introducing `mkInteger` to the API. 
Instead the user needs to use `Solver::mkTerm(api::SINGLETON, element)` where `element` has the right type. 
Internally `NodeManager::mkSingleton(type, element)` is still needed to determine the type of the set. 
Other changes:
- Renamed `Solver::mkTerm(Op op, ..` functions to `Solver::mkTermFromOp(Op op, ...`
- Added `mkTermFromOp` to the python API


src/api/cvc4cpp.cpp
```diff
@@ -4177,17 +4161,34 @@ Term Solver::mkTerm(Kind kind) const
...
   CVC4_API_SOLVER_CHECK_TERM(child);
   checkMkTerm(kind, 1);
-
-  Node res = getNodeManager()->mkNode(extToIntKind(kind), *child.d_node);
+  Node res;
+  switch (kind)
+  {
+    case SINGLETON:
+    {
+      TypeNode type = (*child.d_node).getType();
```
ajreynol, line 4180: Can you comment here that the type of the Node and the type of the Term should be the same? This relies on the fact we don't special case Term::getSort().

ajreynol (approved):
LGTM, one request for a comment.

makaimann (commented):
Hey @mudathirmahgoub! I'm going to focus on the addition of the `mkTermFromOp` method, since that's related to the python bindings. I looked through the changes a bit, and I believe I understand why it's necessary. Correct me if I'm wrong, but I believe it's because the `SINGLETON` kind is only handled [here](https://github.com/mudathirmahgoub/CVC4/blob/043424e96e62b5b3a79b6e694d86c30d90248920/src/api/cvc4cpp.cpp#L3276).

Introducing `mkTermFromOp` certainly works to avoid this issue, but it seems like overkill to me. Personally, I'd prefer to maintain the invariant that a non-indexed `Op` can be always used in place of a `Kind`. E.g. `mkTerm(Op(PLUS), x, y)` works, so `mkTerm(Op(SINGLETON), one)` should work as well. Thus, an alternative solution would be to change what happens for non-indexed operators here (and related places):
https://github.com/CVC4/CVC4/blob/d301b5175e39f82df9c179be1da7eaea892f7795/src/api/cvc4cpp.cpp#L4283

to something like:
```
return mkTermHelper(op.d_kind, child);
```

But, I'd be interested to get other opinions. Maybe @aniemetz is the right person to weigh in? If we decide to go that route, you could make the change in this PR, or I'd be happy to make the change in a separate PR.

mudathirmahgoub:
> Hey @mudathirmahgoub! I'm going to focus on the addition of the `mkTermFromOp` method, since that's related to the python bindings. I looked through the changes a bit, and I believe I understand why it's necessary. Correct me if I'm wrong, but I believe it's because the `SINGLETON` kind is only handled [here](https://github.com/mudathirmahgoub/CVC4/blob/043424e96e62b5b3a79b6e694d86c30d90248920/src/api/cvc4cpp.cpp#L3276).
> 
> Introducing `mkTermFromOp` certainly works to avoid this issue, but it seems like overkill to me. Personally, I'd prefer to maintain the invariant that a non-indexed `Op` can be always used in place of a `Kind`. E.g. `mkTerm(Op(PLUS), x, y)` works, so `mkTerm(Op(SINGLETON), one)` should work as well. Thus, an alternative solution would be to change what happens for non-indexed operators here (and related places):
> 
> https://github.com/CVC4/CVC4/blob/d301b5175e39f82df9c179be1da7eaea892f7795/src/api/cvc4cpp.cpp#L4283
> 
> to something like:
> 
> ```
> return mkTermHelper(op.d_kind, child);
> ```
> 
> But, I'd be interested to get other opinions. Maybe @aniemetz is the right person to weigh in? If we decide to go that route, you could make the change in this PR, or I'd be happy to make the change in a separate PR.

Thanks for the review. This solution works too. I don't have personal preference. The APIs need to be consistent and close to each other. Requested review from  @aniemetz

ajreynol:
I agree with Makai.  Changing the control path from making terms from ops would make things more robust and ensure that singleton is constructed properly, either in the case that it is done via Op or Kind.

ajreynol (changes requested):
I don't think we should change mkTerm -> mkTermFromOp.

mudathirmahgoub:
> I agree with Makai. Changing the control path from making terms from ops would make things more robust and ensure that singleton is constructed properly, either in the case that it is done via Op or Kind.

Done by refactoring `mkTerm(Op& op,..` to use mkTermHelper(Kind, ..) if the operator is not indexed.

makaimann (approved):
Looks good to me. Thanks for the changes!

ajreynol (approved):
LGTM

PR #5368, opened 2020-10-30
Avoid dynamic allocation in symbol table implementation

ajreynol:
Encountered this while debugging the new symbol manager.  This is not essential but probably a good idea to refactor.


abdoo8080 (approved):
LGTM!

PR #5376, opened 2020-11-01
Do not regress explanations of datatype lemmas

ajreynol:
This modifies datatypes to not regress explanations for lemmas.  This avoids segfaults in some corner cases of sygus (see attached) and leads to slightly better performance on Facebook verification benchmarks.


nafur (approved):
LGTM.

PR #5377, opened 2020-11-01
Update strings proxy variable map to be context independent

### LINKED ISSUE #5374: (strings-lazy-pp) Fatal failure at src/theory/strings/term_registry.cpp:370
For this formula, CVC4 42fe4ae
~~~~
(set-logic QF_UFSLIA)
(declare-fun str0 () String)
(declare-fun str4 () String)
(declare-fun str15 () String)
(push 1)
(assert (str.in_re (str.replace str0 "gpQrbuIlpcirZXw" "") (str.to_re "")))
(push 1)
(check-sat)
(pop 1)
(pop 1)
(assert (distinct str15 str4 (str.replace str0 "gpQrbuIlpcirZXw" "") ""))
(push 1)
$ cvc toreport/cvc_str.smt2 -i
sat
Fatal failure within CVC4::theory::TrustNode CVC4::theory::strings::TermRegistry::getRegisterTermLemma(CVC4::Node) at /home/peisen/test/tofuzz/CVC4/src/theory/strings/term_registry.cpp:370
Check failure

 d_proxyVarToLength.find(nc) != d_proxyVarToLength.end()

Aborted
~~~~

ajreynol:
This makes strings proxy variables map to be context-independent.  They should be context independent since we are using attributes to mark proxy variables, which are context-independent.  This led to the crash reported on #5374 since proxy variables would persist across multiple user contexts, but would be missing in the user-context dependent map.

This fixes #5374.


4tXJ7f (approved):
LGTM

PR #5380, opened 2020-11-02
Add symbol manager

ajreynol:
This add the symbol manager class, which is a Term-level utility, separate of the API.  This class manages things like expression and assertion names, which is intentionally done outside the solver.  

The symbol manager is intended to live at the same level as the Solver.  When parsing input, the symbol manager will be used to model any interaction of e.g. named expressions and assertions.  It also stores the symbol table of the parser.

This PR adds the basic interface for the symbol manager and passes it to the parser.

Later PRs will migrate the functionality for named expression out of e.g. SmtEngine and into SymbolManager.  Commands will take Solver+SymbolManager instead of Solver. This will allow the parser to be fully migrated to the new API.

Marking "complex" since this impacts further design of the parser and the code that lives in src/main.

FYI @4tXJ7f


src/parser/symbol_manager.cpp
```diff
@@ -0,0 +1,77 @@
...
+  {
+    // already named assertion, or we are not naming an assertion
+    if (!isAssertion || d_namedAsserts.find(t) != d_namedAsserts.end())
+    {
+      return false;
+    }
+  }
+  d_namedAsserts.insert(t);
+  d_names[t] = name;
+  return true;
```
abdoo8080, line 38: I think you need to guard against the case where `t` is not already named and `isAssertion` is `false`.

abdoo8080 (changes requested):
Looks good overall. I left few comments below. Also, my laptop will appreciate it if you reduce the number of `#include "parser/symbol_manager.h"` in header files 🙂.

src/main/command_executor.h
```diff
@@ -21,17 +21,14 @@
 #include "api/cvc4cpp.h"
 #include "expr/expr_manager.h"
 #include "options/options.h"
+#include "parser/symbol_manager.h"
```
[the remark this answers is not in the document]
ajreynol: Yes. So I'm following the principle here that wherever `api/cvc4cpp.h` is safe to include, symbol manager is also safe to include.

src/parser/symbol_manager.cpp
ajreynol: Will fix.

src/parser/symbol_manager.h
```diff
@@ -0,0 +1,99 @@
...
+   * all terms in ts that has names. Terms that do not have names are not
+   * included.
+   *
+   * @param ts The terms
+   * @param names The name list
+   * @param areAssertions Whether we only wish to include assertion names
+   */
+  void getExpressionNames(const std::vector<api::Term>& ts,
+                          std::vector<std::string>& names,
+                          bool areAssertions = false) const;
```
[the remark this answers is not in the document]
ajreynol: This method is in preparation for unsat core generation from names.

If the user wanted more detailed information about what specific expressions had names, then they should use the more fine grained interface above.

I'll add a comment about this.

src/parser/symbol_manager.cpp
```diff
@@ -0,0 +1,77 @@
...
+{
+  if (d_names.find(t) != d_names.end())
+  {
+    // already named assertion
+    return false;
+  }
+  if (isAssertion)
+  {
+    d_namedAsserts.insert(t);
+  }
```
nafur, line 36:
```suggestion
```

src/parser/symbol_manager.cpp
```diff
@@ -0,0 +1,77 @@
...
+namespace CVC4 {
+namespace parser {
+
+SymbolManager::SymbolManager(api::Solver* s) : d_solver(s) {}
+
+SymbolTable* SymbolManager::getSymbolTable() { return &d_symtabAllocated; }
+
+bool SymbolManager::setExpressionName(api::Term t,
+                                      const std::string& name,
+                                      bool isAssertion)
```
nafur, line 26:
```suggestion
                                      const std::string& name)
```

src/parser/symbol_manager.cpp
```diff
@@ -0,0 +1,77 @@
...
+  {
+    d_namedAsserts.insert(t);
+  }
+  d_names[t] = name;
+  return true;
+}
+
+bool SymbolManager::getExpressionName(api::Term t,
+                                      std::string& name,
+                                      bool isAssertion) const
```
nafur, line 43:
```suggestion
                                      std::string& name) const
```

src/parser/symbol_manager.cpp
```diff
@@ -0,0 +1,77 @@
...
+    return false;
+  }
+  if (isAssertion)
+  {
+    // must be an assertion name
+    if (d_namedAsserts.find(t) == d_namedAsserts.end())
+    {
+      return false;
+    }
+  }
```
nafur, line 57:
```suggestion
```

src/parser/symbol_manager.cpp
```diff
@@ -0,0 +1,77 @@
...
+  name = it->second;
+  return true;
+}
+
+void SymbolManager::getExpressionNames(const std::vector<api::Term>& ts,
+                                       std::vector<std::string>& names,
+                                       bool areAssertions) const
```
nafur, line 64:
```suggestion
                                       std::vector<std::string>& names) const
```

src/parser/symbol_manager.cpp
```diff
@@ -0,0 +1,77 @@
...
+{
+  for (const api::Term& t : ts)
+  {
+    std::string name;
+    if (getExpressionName(t, name, areAssertions))
```
nafur, line 69:
```suggestion
    if (getExpressionName(t, name))
```

src/parser/symbol_manager.h
```diff
@@ -0,0 +1,102 @@
...
+   * @param name The given name
+   * @param isAssertion Whether t is being given a name in an assertion
+   * context. In particular, this is true if and only if there was an assertion
+   * command of the form (assert (! t :named name)).
+   * @return true if the name was set. This method may return false if t
+   * already has a name.
+   */
+  bool setExpressionName(api::Term t,
+                         const std::string& name,
+                         bool isAssertion = false);
```
nafur, line 55:
```suggestion
                         const std::string& name);
```

src/parser/symbol_manager.h
```diff
@@ -0,0 +1,102 @@
...
+   *
+   * @param t The term
+   * @param name The argument to update with the name of t
+   * @param isAssertion Whether we only wish to get the assertion name of t
+   * @return true if t has a name. If so, name is updated to that name.
+   * Otherwise, name is unchanged.
+   */
+  bool getExpressionName(api::Term t,
+                         std::string& name,
+                         bool isAssertion = false) const;
```
nafur, line 66:
```suggestion
   * @return true if t has a name. If so, name is updated to that name.
   * Otherwise, name is unchanged.
   */
  bool getExpressionName(api::Term t,
                         std::string& name) const;
```

src/parser/symbol_manager.h
```diff
@@ -0,0 +1,102 @@
...
+   * specific terms had names, then use the more fine grained interface above,
+   * per term.
+   *
+   * @param ts The terms
+   * @param names The name list
+   * @param areAssertions Whether we only wish to include assertion names
+   */
+  void getExpressionNames(const std::vector<api::Term>& ts,
+                          std::vector<std::string>& names,
+                          bool areAssertions = false) const;
```
nafur, line 83:
```suggestion
   */
  void getExpressionNames(const std::vector<api::Term>& ts,
                          std::vector<std::string>& names) const;
```

src/parser/symbol_manager.h
```diff
@@ -0,0 +1,102 @@
...
+  /** The API Solver object. */
+  api::Solver* d_solver;
+  /**
+   * The declaration scope that is "owned" by this symbol manager.
+   */
+  SymbolTable d_symtabAllocated;
+  /** Map terms to names */
+  std::map<api::Term, std::string> d_names;
+  /** The set of terms with assertion names */
+  std::set<api::Term> d_namedAsserts;
```
nafur, line 96:
```suggestion
```

nafur (changes requested):
As per offline discussion:
The case that the whole `isAssertion` stuff is supposed to handle is not valid SMTLIB anyway (defining names multiple times, once as nested term and once as assertion). Thus we should remove it completely.

src/parser/symbol_manager.cpp
ajreynol: I disagree.  We need it to distinguish:
```
(assert t1)
(assert (and (! t1 :named foo) t2))
```
and

```
(assert (! t1 :named foo))
(assert (and t1 t2))
```

nafur (approved):
Ah you are right. Of course...

LGTM.

aniemetz (approved):

abdoo8080 (approved):

PR #5389, opened 2020-11-03
Simplify sygus solver and sygus stream

ajreynol:
This simplifies the sygus solver based on the fact that verification lemmas are always processed in a separate subsolver.

In particular, this means that the implementation of `--sygus-stream` can be simplified signficantly.


src/theory/quantifiers/sygus/synth_conjecture.cpp
```diff
@@ -669,22 +611,22 @@ bool SynthConjecture::doCheck(std::vector<Node>& lems)
...
-    {
-      // if we were successful, we immediately print the current solution.
-      // this saves us from introducing a verification lemma and a new guard.
-      printAndContinueStream(terms, candidate_values);
-      // streaming means now we immediately are looking for a new solution
-      d_hasSolution = false;
-      return false;
-    }
+    // if we were successful, we immediately print the current solution.
+    // this saves us from introducing a verification lemma and a new guard.
```
HanielB, line 621: this should be updated, right?

HanielB (approved):
This looks great. :) Just a minor comment.

PR #5390, opened 2020-11-03
Migrate some documentation about attributes

ajreynol:
From old wiki.


src/expr/attribute.h
```diff
@@ -37,6 +37,44 @@ namespace CVC4 {
 namespace expr {
 namespace attr {
 
+/**
+ * Attributes are roughly speaking [almost] global hash tables from Nodes
+ * (TNodes) to data. Attributes can be thought of as additional fields used to
+ * extend NodeValues. Attributes are mutable, and come in both sat
+ * context-dependent and non-context dependent varieties. Attributes live only
+ * as long as the node itself does. If a Node is garbage-collected Attribute's
```
aniemetz, line 45:
```suggestion
 * as long as the node itself does. If a Node is garbage-collected, Attributes
```

src/expr/attribute.h
```diff
@@ -37,6 +37,44 @@ namespace CVC4 {
 namespace expr {
 namespace attr {
 
+/**
+ * Attributes are roughly speaking [almost] global hash tables from Nodes
+ * (TNodes) to data. Attributes can be thought of as additional fields used to
+ * extend NodeValues. Attributes are mutable, and come in both sat
```
aniemetz, line 43:
```suggestion
 * extend NodeValues. Attributes are mutable and come in both sat
```

src/expr/attribute.h
```diff
@@ -37,6 +37,44 @@ namespace CVC4 {
...
+ * ```
+ * n.getAttribute(InstLevelAttribute());
+ * ```
+ *
+ * To check whether the attribute has been set:
+ * ```
+ * n.hasAttribute(InstLevelAttribute());
+ * ```
+ *
+ * To separate Attributes of the same type in the same table each of the
```
aniemetz, line 71:
```suggestion
 * To separate Attributes of the same type in the same table, each of the
```

src/expr/attribute.h
```diff
@@ -37,6 +37,44 @@ namespace CVC4 {
...
+ * structures struct InstLevelAttributeId {}; is given a different unique value
```
aniemetz, line 72:
```suggestion
 * structures `struct InstLevelAttributeId {};` is given a different unique value
```

aniemetz (approved):
Some minor things, else LGTM.

PR #5391, opened 2020-11-03
Add constants from equality engine evaluation to model

### LINKED ISSUE #5330: Invalid model for string formulas
Hi, for the following formula (**str.in_re, str.replace, str.to_re**),
cvc4 a7aba0c858
~~~~
(declare-fun str1 () String)
(declare-fun str18 () String)
(assert (str.in_re (str.replace str18 str1 str18) (str.to_re "pANjvthXNyBbIgIlkC")))
(check-sat)
~~~~

~~~~
$ cvc4 --strings-exp --check-models f.smt2

Fatal failure within void CVC4::smt::CheckModels::checkModel(CVC4::smt::Model*, CVC4::context::CDList<CVC4::NodeTemplate<true> >*, bool) at /home/peisen/test/tofuzz/CVC4/src/smt/check_models.cpp:258
assertion:     (str.in_re (str.replace str18 str1 str18) (str.to_re "pANjvthXNyBbIgIlkC"))
simplifies to: false
expected `true'.
Aborted (core dumped)
~~~~

4tXJ7f:
Fixes #5330. The issue mentions to related model checking failures:

- When the theory of strings computes a model, there is a step that
  chooses a constant that is different from other constants of the same
  length in other equivalence classes. In the example, the issue was
  that the constant `"A"` was introduced by doing evaluation in the
  equality engine of the theory of strings. The constant was then not
  added to the term set and was skipped while asserting the equality
  engine to the theory model. As a result, an equivalence class was
  assigned `"A"` even though it already existed, which made the model
  invalid. The fix ensures that all constants in the equality engine are
  added to the theory model. It should be ok to handle the issue during
  model construction because other theories shouldn't have to care about
  the constants that we computed internally within the theory of
  strings.
- When an equivalence class has a normal form that consists of a single
  `seq.unit`, then we need to make sure that the value for the argument
  is consistent with the rest of the model and we have to make sure that
  we choose different values for different equivalence classes. The
  commit adds code for retrieving the value of the argument to
  `seq.unit` from the model and adds the resulting constant to the
  theory model to block it for other equivalence classes. Previously,
  this was not done and we ended up with two different equivalence
  classes being assigned the same constant.


ajreynol:
Can you check whether this fixes https://github.com/CVC4/CVC4/issues/5381 as well?  If so, can you add that as a regression?

4tXJ7f:
Ah sorry, forgot to mention. I tested it but unfortunately, it didn't fix that issue. We get a different model now but it is still invalid.

ajreynol:
Ah ok, I'll look into the other issue.

ajreynol (approved):
LGTM

PR #5392, opened 2020-11-03
Add support for printing `re.loop` and `re.^`

4tXJ7f:
In the new SMT-LIB string standard, `re.loop` and `re.^` are indexed
operators but the printer was not updated to print them correctly. This
commit adds support for printing `re.loop` and `re.^`.


ajreynol (approved):
LGTM

PR #5393, opened 2020-11-03
Add scope methods constructing types in API

ajreynol:
This addresses the nightly failures.  It ensures a node manager is in scope when constructing type nodes.

It also simplifies the construction of atomic type nodes to avoid an extra TypeNode(...) constructor.


4tXJ7f (approved):
LGTM

PR #5397, opened 2020-11-05
Add simple substitution utility

ajreynol:
A few new algorithms for CEGQI and single invocation sygus will use this utility for managing transformations.

### DIFF
--- a/src/expr/subs.cpp
+++ b/src/expr/subs.cpp
@@ -1,0 +1,177 @@
+/*********************                                                        */
+/*! \file subs.cpp
+ ** \verbatim
+ ** Top contributors (to current version):
+ **   Andrew Reynolds
+ ** This file is part of the CVC4 project.
+ ** Copyright (c) 2009-2020 by the authors listed in the file AUTHORS
+ ** in the top-level source directory and their institutional affiliations.
+ ** All rights reserved.  See the file COPYING in the top-level source
+ ** directory for licensing information.\endverbatim
+ **
+ ** \brief Simple substitution utility
+ **/
+
+#include "expr/subs.h"
+
+#include "theory/rewriter.h"
+
+namespace CVC4 {
+
+bool Subs::empty() const { return d_vars.empty(); }
+
+size_t Subs::size() const { return d_vars.size(); }
+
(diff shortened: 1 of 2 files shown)


src/expr/subs.cpp
```diff
@@ -0,0 +1,177 @@
...
+ ** directory for licensing information.\endverbatim
+ **
+ ** \brief Simple substitution utility
+ **/
+
+#include "expr/subs.h"
+
+#include "theory/rewriter.h"
+
+namespace CVC4 {
```
mpreiner, line 19: should this be only in the CVC4 namespace or in the `expr` namespace too?

src/expr/subs.h
```diff
@@ -0,0 +1,85 @@
...
+ **/
+
+#ifndef CVC4__EXPR__SUBS_H
+#define CVC4__EXPR__SUBS_H
+
+#include <map>
+#include <vector>
+#include "expr/node.h"
+
+namespace CVC4 {
```
mpreiner, line 22: Ditto namespace.

src/expr/subs.h
```diff
@@ -0,0 +1,85 @@
...
+  void add(Node v, Node s);
+  /** Add vs -> ss to the substitution */
+  void add(const std::vector<Node>& vs, const std::vector<Node>& ss);
+  /** Add eq[0] -> eq[1] to the substitution */
+  void addEquality(Node eq);
+  /** Append the substitution s to this */
+  void append(Subs& s);
+  /** Return the result of this substitution on n */
+  Node apply(Node n, bool doRewrite = false) const;
+  /** Return the result of the reserve of this substitution on n */
```
mpreiner, line 57:
```suggestion
  /** Return the result of the reverse of this substitution on n */
```

src/expr/subs.cpp
ajreynol: Many things in src/expr/ are intentionally only in the CVC4:: namespace.  This goes against the policy of namespaces matching folder directories, but is intentional since it is less verbose.  This utility in particular is preferred to be used as `Subs` instead of `expr::Subs`.

mpreiner (approved):

PR #5401, opened 2020-11-07
Simplify handling of subtypes in smt2 printer

ajreynol:
This makes major simplifications to how subtypes are enforced in the smt2 printer. 

It is now the principle that the smt2 prints things faithfully to the AST, regardless of whether it conforms to the smt2 standard.

It also fixes the current smt2 printing of `to_real`.

Conversely, this removes a hack from GetValueCommand which forced casting via x -> (/ x 1).  This is now properly handled in Solver::getValue.

Some regressions change expected output as a result.  Notice that internally generated Node may not conform to the smt2 standard, but user-level terms will.

### DIFF
--- a/src/printer/smt2/smt2_printer.cpp
+++ b/src/printer/smt2/smt2_printer.cpp
@@ -111,11 +111,9 @@ static bool stringifyRegexp(Node n, stringstream& ss) {
   return true;
 }
 
-// force_nt is the type that n must have
 void Smt2Printer::toStream(std::ostream& out,
                            TNode n,
-                           int toDepth,
-                           TypeNode force_nt) const
+                           int toDepth) const
 {
   // null
   if(n.getKind() == kind::NULL_EXPR) {
@@ -395,7 +393,7 @@ void Smt2Printer::toStream(std::ostream& out,
     if(n.getNumChildren() != 0) {
       for(unsigned i = 0; i < n.getNumChildren(); ++i) {
 	      out << ' ';
-              toStream(out, n[i], toDepth, TypeNode::null());
+              toStream(out, n[i], toDepth);
       }
       out << ')';
     }
@@ -427,7 +425,7 @@ void Smt2Printer::toStream(std::ostream& out,
           << smtKindString(is_int ? kind::TO_INTEGER : kind::DIVISION,
                            d_variant)
           << " ";
-      toStream(out, type_asc_arg, toDepth, TypeNode::null());
+      toStream(out, type_asc_arg, toDepth);
       if (!is_int)
       {
         out << " 1";
@@ -440,8 +438,7 @@ void Smt2Printer::toStream(std::ostream& out,
       out << "(as ";
       toStream(out,
                type_asc_arg,
-               toDepth < 0 ? toDepth : toDepth - 1,
-               TypeNode::null());
+               toDepth < 0 ? toDepth : toDepth - 1);
       out << " " << force_nt << ")";
     }
     return;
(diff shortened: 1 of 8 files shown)


src/printer/smt2/smt2_printer.cpp
```diff
@@ -59,7 +59,48 @@ bool isVariant_2_6(Variant v) { return v == smt2_6_variant; }
...
+{
+  bool neg = r.sgn() < 0;
+  // Print the rational, possibly as decimal.
+  // Notice that we print (/ (- 5) 3) instead of (- (/ 5 3)),
+  // the former is compliant with real values in the smt lib standard.
+  if (r.isIntegral())
+  {
+    if (neg)
+    {
+      out << (v == sygus_variant ? "-" : "(- ") << -r;
```
mudathirmahgoub, line 72: Could you write a comment here as well like sygus requires `-n` instead of `(- n)` for negative integers.

src/printer/smt2/smt2_printer.cpp
```diff
@@ -404,15 +441,18 @@ void Smt2Printer::toStream(std::ostream& out,
...
   Node type_asc_arg;
-  if (n.getKind() == kind::APPLY_TYPE_ASCRIPTION)
+  TypeNode force_nt;
+  if (k == kind::APPLY_TYPE_ASCRIPTION)
   {
     force_nt = n.getOperator().getConst<AscriptionType>().getType();
     type_asc_arg = n[0];
   }
-  else if (!force_nt.isNull() && n.getType() != force_nt)
+  else if (k == kind::CAST_TO_REAL)
```
mudathirmahgoub, line 452: Shouldn't we do the same for `kind::TO_REAL` here?

ajreynol: No, TO_REAL is an operator that should print as `(to_real t)` for any term t.

src/printer/smt2/smt2_printer.cpp
ajreynol: This is outdated actually, ill remove.

src/printer/smt2/smt2_printer.cpp
```diff
@@ -472,13 +518,9 @@ void Smt2Printer::toStream(std::ostream& out,
 
   bool stillNeedToPrintParams = true;
   bool forceBinary = false; // force N-ary to binary when outputing children
-  bool parametricTypeChildren = false;   // parametric operators that are (op t1 ... tn) where t1...tn must have same type
-  bool typeChildren = false;  // operators (op t1...tn) where at least one of t1...tn may require a type cast e.g. Int -> Real
   // operator
-  Kind k = n.getKind();
   if (n.getNumChildren() != 0 && k != kind::INST_PATTERN_LIST
-      && k != kind::APPLY_TYPE_ASCRIPTION && k != kind::CONSTRUCTOR_TYPE
-      && k != kind::CAST_TO_REAL)
```
mudathirmahgoub, line 481: I remember I added `CAST_TO_REAL` here to avoid extra parentheses for `CAST_TO_REAL` which is hidden from the user. I guess this is no longer needed since the printer is faithful to the AST.

ajreynol: Yes, that's right.  I simplified this, which notice also applies to APPLY_TYPE_ASCRIPTION (which is a cast for datatypes).

src/printer/smt2/smt2_printer.cpp
```diff
@@ -1046,7 +1011,26 @@ void Smt2Printer::toStream(std::ostream& out,
...
+  {
+    Assert(tn.isReal());
+    // probably due to subtyping integers and reals, cast it
+    nasc = NodeManager::currentNM()->mkNode(kind::CAST_TO_REAL, n);
+  }
+  else
+  {
+    nasc = n;
+  }
+  toStream(out, nasc, -1);
```
mudathirmahgoub, line 1028: Did you mean `toDepth -1` here? If not, then `toDepth` argument is never used here.

ajreynol: sure, ill change.

mudathirmahgoub (approved):
Thanks for simplifying the complex function `Smt2Printer::toStream` from 922 to 905 lines !!

PR #5403, opened 2020-11-09
Properly reset interrupt for incremental calls to minisat

nafur:
The minisat solver stores whether it has been interrupted in `asynch_interrupt` and expects it to be reset before another call to `solve()`. `MinisatSatSolver::solve()` failed to do this, leading to incorrect `unknown` results as observed in CVC4/cvc4-projects#106. The alternative `MinisatSatSolver::solve(unsigned long& resource)` already did the correct thing.
Fixes CVC4/cvc4-projects#106.


ajreynol (approved):
LGTM

PR #5405, opened 2020-11-09
Add proper support for the declare-heap command for separation logic

### LINKED ISSUE #4926: (finite-model-find) Fatal failure at CVC4/src/expr/type_checker_template.cpp:51 (unspecified sep logic)
Hi,
For this formula, CVC4  throws out a fatal failure:
```
[605] % cvc4 -q small.smt2
sat
[606] % cvc4 --finite-model-find -q small.smt2
Fatal failure within static CVC4::TypeNode CVC4::expr::TypeChecker::computeType(CVC4::NodeManager*, CVC4::TNode, bool) at CVC4/src/expr/type_checker_template.cpp:51
Unhandled case encounteredNULL
Aborted
[607] % 
[607] % cat small.smt2
(declare-sort a 0)
(declare-datatypes ((b 0)) (((c (v Int) (h a)))))
(declare-const u Int)
(declare-const e a)
(define-fun f ((g a) (d Int)) Bool (_ emp a b))
(assert (f e u))
(check-sat)
[608] %
```

OS: Ubuntu 18.04
Commit: c110363

### LINKED ISSUE #5343: Fatal failure at src/theory/sep/theory_sep.cpp on separation logic formula
cvc4 0a4ee1a throwas a fatal failure at /src/theory/sep/theory_sep.cpp:941
~~~~
$ cvc4 -q  xx.smt2 
Fatal failure within CVC4::TypeNode CVC4::theory::sep::TheorySep::getReferenceType(CVC4::Node) at /CVC4/src/theory/sep/theory_sep.cpp:941
Check failure

 !d_type_ref.isNull()

Aborted (core dumped)

$ cat xx.smt2
(set-logic QF_ALL_SUPPORTED)
(assert (wand true true))
(check-sat)
~~~~

ajreynol:
This adds proper support for the `(declare-heap (T U))`  command, which declares the type of the heap in separation logic.  This command is part of the standard for separation logic.

This previous was handled in an ad-hoc way where the type of the heap would be inferred on demand. This was a poor solution and has led to a number of issues related to when the heap is inferred.

Fixes #5343, fixes #4926.

Work towards https://github.com/CVC4/cvc4-wishues/issues/22.

### DIFF
--- a/src/smt/smt_engine.cpp
+++ b/src/smt/smt_engine.cpp
@@ -1513,7 +1513,7 @@ void SmtEngine::checkUnsatCore() {
 
   // set up separation logic heap if necessary
   TypeNode sepLocType, sepDataType;
-  if (coreChecker->getSepHeapTypes(sepLocType, sepDataType))
+  if (getSepHeapTypes(sepLocType, sepDataType))
   {
     coreChecker->declareSepHeap(sepLocType, sepDataType);
   }
--- a/src/smt/smt_engine.cpp
+++ b/src/smt/smt_engine.cpp
@@ -66,6 +66,7 @@
 #include "theory/quantifiers/quantifiers_attributes.h"
 #include "theory/rewriter.h"
 #include "theory/theory_engine.h"
+#include "theory/smt_engine_subsolver.h"
 #include "util/hash.h"
 #include "util/random.h"
 #include "util/resource_manager.h"
@@ -1462,6 +1463,14 @@ void SmtEngine::declareSepHeap(TypeNode locT, TypeNode dataT)
   te->declareSepHeap(locT, dataT);
 }
 
+bool SmtEngine::getSepHeapTypes(TypeNode& locT, TypeNode& dataT)
+{
+  SmtScope smts(this);
+  finishInit();
+  TheoryEngine* te = getTheoryEngine();
+  return te->getSepHeapTypes(locT, dataT);
+}
+
 Node SmtEngine::getSepHeapExpr() { return getSepHeapAndNilExpr().first; }
 
 Node SmtEngine::getSepNilExpr() { return getSepHeapAndNilExpr().second; }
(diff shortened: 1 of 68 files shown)


src/smt/smt_engine.cpp
```diff
@@ -1483,21 +1506,28 @@ void SmtEngine::checkUnsatCore() {
...
+  std::unique_ptr<SmtEngine> coreChecker;
+  initializeSubsolver(coreChecker);
+  coreChecker->getOptions().set(options::checkUnsatCores, false);
+
+  // set up separation logic heap if necessary
+  TypeNode sepLocType, sepDataType;
+  if (getSepHeapTypes(sepLocType, sepDataType))
+  {
+    coreChecker->declareSepHeap(sepLocType, sepDataType);
+  }
```
nafur, line 1519: Why this change?

src/theory/sep/theory_sep.cpp
```diff
@@ -69,6 +69,21 @@ TheorySep::~TheorySep() {
...
 
+void TheorySep::declareSepHeap(TypeNode locT, TypeNode dataT)
+{
+  if (!d_type_ref.isNull())
+  {
+    TypeNode te1 = d_loc_to_data_type.begin()->first;
+    std::stringstream ss;
+    ss << "ERROR: declaring heap constraints for two different types : ";
+    ss << locT << " -> " << dataT << ", but we already have ";
+    ss << d_type_ref << " -> " << d_type_data;
```
nafur, line 80: This error message seems confusing to me. Either we only allow a single declare-heap command (then the message should say so) or we allow multiple declare-heap commands if the types are the same (then we should check this and not print an error)
The tests below seem to imply that we can only have a single declare-heap command...

nafur (changes requested):
The error message is a bit confusing...

src/smt/smt_engine.cpp
ajreynol: We now require that the type of the separation logic heap is declared.  The subsolver for checking unsat cores sets up a fresh copy of the SMT solver.  This copy must carry over the declared heap type.

src/theory/sep/theory_sep.cpp
ajreynol: Yeah, you're right. I'll make this clearer.

nafur (approved):
LGTM

PR #5409, opened 2020-11-10
(proof-new) Proofs for non-clausal simplification

ajreynol:
Adds proof support in non-clausal simplification, connecting the proofs from circuit propagator.


src/preprocessing/passes/non_clausal_simp.cpp
```diff
@@ -115,41 +126,42 @@ PreprocessingPassResult NonClausalSimp::applyInternal(
...
+      std::make_shared<TrustSubstitutionMap>(
+          u, d_pnm, "NonClausalSimp::cprop", PfRule::PREPROCESS_LEMMA);
+  SubstitutionMap& cps = constantPropagations->get();
+  // new substitutions
+  std::shared_ptr<TrustSubstitutionMap> newSubstitutions =
+      std::make_shared<TrustSubstitutionMap>(
+          u, d_pnm, "NonClausalSimp::newSubs", PfRule::PREPROCESS_LEMMA);
+  SubstitutionMap& nss = newSubstitutions->get();
+
   SubstitutionMap::iterator pos;
```
nafur, line 143: This is horrible. This is one iterator used everywhere in this method...

src/preprocessing/passes/non_clausal_simp.cpp
```diff
@@ -280,34 +282,19 @@ PreprocessingPassResult NonClausalSimp::applyInternal(
   // r' another constant propagation, then l'[l/r] -> r' should be a
   //    constant propagation too
   // 4. each lhs of constantPropagations is different from each rhs
-  for (pos = newSubstitutions.begin(); pos != newSubstitutions.end(); ++pos)
+  for (pos = nss.begin(); pos != nss.end(); ++pos)
```
nafur, line 285:
```suggestion
  for (const auto& pos: nss)
```
And replace `(*pos)` by `pos`.

src/preprocessing/passes/non_clausal_simp.cpp
```diff
@@ -280,34 +282,19 @@ PreprocessingPassResult NonClausalSimp::applyInternal(
...
     Assert(top_level_substs.apply((*pos).first) == (*pos).first);
     Assert(top_level_substs.apply((*pos).second) == (*pos).second);
-    Assert(newSubstitutions.apply(newSubstitutions.apply((*pos).second))
-           == newSubstitutions.apply((*pos).second));
+    Node app = nss.apply((*pos).second);
+    Assert(nss.apply(app) == app);
   }
-  for (pos = constantPropagations.begin(); pos != constantPropagations.end();
-       ++pos)
+  for (pos = cps.begin(); pos != cps.end(); ++pos)
```
nafur, line 293: Ditto.

src/preprocessing/passes/non_clausal_simp.cpp
```diff
@@ -355,10 +340,11 @@ PreprocessingPassResult NonClausalSimp::applyInternal(
   TheoryModel* m = d_preprocContext->getTheoryEngine()->getModel();
   Assert(m != nullptr);
   NodeManager* nm = NodeManager::currentNM();
-  for (pos = newSubstitutions.begin(); pos != newSubstitutions.end(); ++pos)
+  for (pos = nss.begin(); pos != nss.end(); ++pos)
```
nafur, line 343: Ditto.

src/preprocessing/passes/non_clausal_simp.cpp
```diff
@@ -415,17 +393,12 @@ PreprocessingPassResult NonClausalSimp::applyInternal(
   }
   learned_literals.clear();
 
-  for (pos = constantPropagations.begin(); pos != constantPropagations.end();
-       ++pos)
+  for (pos = cps.begin(); pos != cps.end(); ++pos)
```
nafur, line 396: Ditto.

nafur (changes requested):
I know this stuff was not introduced with this PR, but I find it bad enough to change it right here and now :-)

ajreynol:
I've updated to use local iterators. In this instance I'd prefer not to use auto since its a non-standard object that is being iterated over and it is clearer to me to have the type.

nafur (approved):
LGTM now.

PR #5410, opened 2020-11-10
Pass symbol manager to commands

ajreynol:
This PR passes the symbol manager to Command::invoke.

There are no behavior changes in this PR. This is in preparation for reimplementing several features in the parser related to symbols.


nafur (approved):
LGTM.
(assuming that some of the commands will eventually use the symbol manager...)

PR #5411, opened 2020-11-10
Fix preregistration in TheorySep before declare-heap

ajreynol:
Followup to fix for #5343. This catches cases where separation logic constraints are preregistered to TheorySep before the heap has been declared, which should also result in an error.


nafur (approved):
LGTM.

Issue #4987, opened 2020-09-01
Non standard SMT-LIB v2.6 response with the `(get-model)` command


timothee-haudebourg:
Hi,

I'm using the `--output-lang=smtlib2.6` option with CVC4 1.8 and expect a standard SMT-LIB v2.6 reponse to the `(get-model) command. However the output doesn't seem to comply to the standard (which I've found here http://smtlib.cs.uiowa.edu/papers/smt-lib-reference-v2.6-r2017-07-18.pdf).

For instance, consider the following SMT-LIB v2.6 input:
```
(set-logic QF_UF)
(declare-sort Sort0 0)
(declare-fun f1 (Sort0) Bool)
(check-sat)
(get-model)
```

Here is the CVC4 output for the last command:
```
(model
; cardinality of Sort0 is 1
(declare-sort Sort0 0)
(declare-fun groundTerm_1 () Sort0)
(define-fun f1 ((BOUND_VARIABLE_279 Sort0)) Bool false)
)
```

However according to the SMT-LIB v2.6 spec, the output of the `(get-model)` command should be of the form:
```
<get_model_response> ::= ( <model_response>* )
    <model_response> ::= ( define-fun <function_def> ) | ( define-fun-rec <function_def> )
                         ( define-funs-rec ( <function_dec>^{n+1} ) ( <term>^{n+1} ) )
```
In particular, the `model` keyword should not be here, the `declare-sort` command should not be here, and I don't even understand why a `declare-fun groundTerm_1` appears (which seems to be quite new, it wasn't here with CVC4 1.7 I think).

So am I missing something here? Am I not reading the right spec for the SMT-LIB v2.6 language? Or is CVC4 not compliant? In which case, should it be fixed? Or where can I find the actual language definition used by CVC4?

aniemetz:
No, you didn't miss anything here. Actually, during SMT-COMP this year, the organizers noticed that none of the participating solvers is compliant with the standard here, all output the model starting with the `model` keyword, CVC4 included. There has been discussion with the SMT-LIB maintainers as to how to deal with this (if there is a good reason to change the standard to have the model response start with a keyword or if the current standard is what should actually be enforced), and no conclusion has been reached yet. @tinelli @barrettcw

timothee-haudebourg:
[...]

PR #5415, opened 2020-11-10
Models standard

yoni206:
This PR relates to #4987 .
Our plan is to:
1. delete the `model` keyword
2. avoid printing extra declarations by default
3. wrap UF values with `as` expressions.

This PR does step 1, and fixes regressions accordingly.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -34,6 +34,9 @@ Changes:
   now restrict bvxnor to only allow two operands in order to avoid confusion
   about the semantics, since the behavior of n-ary operands to bvxnor is now
   undefined in SMT-LIB.
+* SMT-LIB output for `get-model` command now conforms with the standard,
+  and does *not* begin with the keyword `model`. The output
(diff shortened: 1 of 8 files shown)


ajreynol (commented):
I'm not sure if we already want to update how the separation logic model is printed as well (see line 1343).

yoni206:
IMH this is a separate issue because separation logic is not standard, but I'm fine either way.

ajreynol (approved):
LGTM

4tXJ7f (changes requested):
Since this is a user-facing change, we should update NEWS.

yoni206:
Right, @4tXJ7f , thanks!
Addressed here:
https://github.com/CVC4/CVC4/pull/5415/commits/288ca59a0acd4dee03c5155a648cba3e2226de4e

4tXJ7f (approved):
LGTM

PR #5419, opened 2020-11-11
Rewrite witness terms at prerewrite

ajreynol:
Ensures (witness ((x Int)) (= x (+ 1 a)) is rewritten to (+ 1 a), instead of e.g. (witness ((x Int)) (= a (- x 1)).

This is required for supporting purification skolems for arithmetic terms in the new proof architecture.

### DIFF
--- a/src/theory/builtin/theory_builtin_rewriter.cpp
+++ b/src/theory/builtin/theory_builtin_rewriter.cpp
@@ -456,7 +456,7 @@ Node TheoryBuiltinRewriter::rewriteWitness(TNode node)
   Assert(node.getKind() == kind::WITNESS);
   if (node[1].getKind() == kind::EQUAL)
   {
-    for (unsigned i = 0; i < 2; i++)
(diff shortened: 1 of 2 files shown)


src/theory/builtin/theory_builtin_rewriter.cpp
```diff
@@ -468,6 +451,42 @@ Node TheoryBuiltinRewriter::getArrayRepresentationForLambdaRec(TNode n,
   }
 }
 
+Node TheoryBuiltinRewriter::rewriteWitness(TNode node)
+{
+  Assert(node.getKind() == kind::WITNESS);
+  if (node[1].getKind() == kind::EQUAL)
+  {
+    for (unsigned i = 0; i < 2; i++)
```
nafur, line 459:
```suggestion
    for (std::uint64_t i = 0; i < 2; i++)
```

nafur (approved):
Just one tiny suggestion concerning the usage of fixed-width integer types...

PR #5420, opened 2020-11-11
Move symbol manager to src/expr/

ajreynol:
This is required since symbol manager will use context dependent data structures (in its cpp).

This is required since classes in src/parser/ are not allowed to include private headers.


nafur (approved):
LGTM.

PR #5421, opened 2020-11-11
Consolidate basic sygus utilities regarding sygus conjectures

ajreynol:
This is required for new work on generalizing CAV 2015 single invocation techniques.

It adds a new system of marking solutions for synthesis conjectures as attributes, which will be used as a way of eliminating functions from a conjecture while still preserving their solution in a response to `check-synth`.


src/theory/quantifiers/sygus/sygus_utils.cpp
```diff
@@ -0,0 +1,182 @@
...
+                                   const std::vector<Node>& iattrs)
+{
+  Assert(!fs.empty());
+  NodeManager* nm = NodeManager::currentNM();
+  std::vector<Node> ipls;
+  SygusAttribute ca;
+  Node sygusVar = nm->mkSkolem("sygus", nm->booleanType());
+  sygusVar.setAttribute(ca, true);
+  Node instAttr = nm->mkNode(INST_ATTRIBUTE, sygusVar);
+  ipls.push_back(instAttr);
```
HanielB, line 48:
```suggestion
  SygusAttribute ca;
  Node sygusVar = nm->mkSkolem("sygus", nm->booleanType());
  sygusVar.setAttribute(ca, true);
  std::vector<Node> ipls{nm->mkNode(INST_ATTRIBUTE, sygusVar)};
```

src/theory/quantifiers/sygus/sygus_utils.cpp
```diff
@@ -0,0 +1,182 @@
...
+
+void SygusUtils::decomposeSygusConjecture(Node q,
+                                          std::vector<Node>& fs,
+                                          std::vector<Node>& unsf,
+                                          Subs& solvedf)
+{
+  Assert(q.getKind() == FORALL);
+  Assert(q.getNumChildren() == 3);
+  Node ipl = q[2];
+  Assert(ipl.getKind() == INST_PATTERN_LIST);
```
HanielB, line 91: Maybe we wanna rename this at some point? The attribute is used for more than that.

src/theory/quantifiers/sygus/sygus_utils.h
```diff
@@ -0,0 +1,116 @@
...
+                                       std::vector<Node>& unsf,
+                                       Subs& solvedf);
+  /**
+   * Decompose the negated conjecture body.
+   *
+   * This returns a quantifier-free formula corresponding to the (un-negated)
+   * conjecture body. It adds the quantified free variables of the conjecture
+   * to vs.
+   */
+  static Node decomposeConjectureBody(Node conj, std::vector<Node>& vs);
```
HanielB, line 82: It's a bit confusing to have this and the previous method named similarly but doing seemly very different things.

src/theory/quantifiers/sygus/sygus_utils.cpp
ajreynol: Yes, this kind should be renamed (not this PR of course).

src/theory/quantifiers/sygus/sygus_utils.h
ajreynol: Maybe `decomposeSygusConjecture` and `decomposeSygusBody` ?

HanielB: I think what was more prominent was that the first decompose the functions-to-synthesize and the second gets the vars and *negates* the body. Having the fact the body is negated in the name would be helpful, I think. Maybe something like `getNegatedConjectureBodyAndFVs`?

ajreynol: I think that's a bit verbose :) decomposeSygusNegatedBody?

HanielB: That sounds good!

HanielB (approved):

src/theory/quantifiers/sygus/sygus_utils.h
ajreynol: Decided to change the negation in the method, since it was confusing, and call this `decomposeSygusBody` since it is now analogous to the other method (operating on negated conjectures/their bodies).

PR #5423, opened 2020-11-11
(proof-new) Preparations for nonlinear arithmetic proofs

nafur:
This PR does some preparations for proofs for nl-ext lemmas:
- the arithmetic inference manager no longer rewrites lemmas, but only caches based on the rewritten lemma (same for the underlying buffered inference manager)
- a proof checker (that is still empty) is added for nl-ext
- the proof node manager is stored in ExtState so that the subsolver can actually see whether proofs are enabled.


ajreynol (approved):
LGTM

PR #5424, opened 2020-11-11
Make symbol manager context dependent

ajreynol:
This follows a similar style to symbol_table.h/cpp.  This is required since context dependent data structures are cvc4_private, and symbol manager is cvc4_public.

### DIFF
--- a/src/expr/symbol_manager.cpp
+++ b/src/expr/symbol_manager.cpp
@@ -25,9 +25,8 @@ namespace CVC4 {
 
 class SymbolManager::Implementation
 {
-  typedef CDHashMap<api::Term, std::string, api::TermHashFunction>
-      TermStringMap;
-  typedef CDHashSet<api::Term, api::TermHashFunction> TermSet;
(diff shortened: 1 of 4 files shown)


src/expr/symbol_manager.cpp
```diff
@@ -14,15 +14,55 @@
...
 
-SymbolManager::SymbolManager(api::Solver* s) : d_solver(s) {}
+// ---------------------------------------------- SymbolManager::Implementation
 
-SymbolTable* SymbolManager::getSymbolTable() { return &d_symtabAllocated; }
+class SymbolManager::Implementation
+{
+  typedef CDHashMap<api::Term, std::string, api::TermHashFunction>
+      TermStringMap;
+  typedef CDHashSet<api::Term, api::TermHashFunction> TermSet;
```
nafur, line 30:
```suggestion
  using TermStringMap = CDHashMap<api::Term, std::string, api::TermHashFunction>;
  using TermSet = CDHashSet<api::Term, api::TermHashFunction>;
```

nafur (approved):
LGTM. Just one minor suggestion.

PR #5429, opened 2020-11-12
(proof-new) Add proof for NL_SIGN lemma.

nafur:
This PR adds a proof for the NL_SIGN lemma, and does a bit of cleanup.


src/theory/arith/nl/ext/monomial_check.h
```diff
@@ -26,7 +26,7 @@ namespace nl {
 class MonomialCheck
 {
  public:
-  MonomialCheck(ExtState* data);
+  MonomialCheck(ExtState* data, context::Context* ctx);
```
ajreynol, line 29: Can you call `ctx` -> `u` ?  The point is it should be clear that this is the user context. (Although this class could be a generic utility, its more important to me to make its standard use case clear.)

ajreynol (changes requested):

src/expr/proof_rule.h
```diff
@@ -1102,6 +1102,13 @@ enum class PfRule : uint32_t
   // ---------------------
   // Conclusion: (Q)
   INT_TRUST,
+  //======== Multiplication with zero factor
```
ajreynol, line 1105: This rule is a special case of MACRO_SR_PRED_INTRO.  We should consider just using MACRO_SR_PRED_INTRO for uniformity here.

src/theory/arith/nl/ext/monomial_check.cpp
```diff
@@ -311,8 +312,17 @@ int MonomialCheck::compareSign(
   {
     if (mvaoa.getConst<Rational>().sgn() != 0)
     {
-      Node lemma = av.eqNode(d_data->d_zero).impNode(oa.eqNode(d_data->d_zero));
-      d_data->d_im.addPendingArithLemma(lemma, InferenceId::NL_SIGN);
+      Node prem = av.eqNode(d_data->d_zero);
+      Node conc = oa.eqNode(d_data->d_zero);
+      Node lemma = prem.impNode(conc);
+      LazyCDProof* proof = nullptr;
+      if (d_data->d_proof)
```
ajreynol, line 319: It's clearer if a condition like this is `isProofEnabled()`, can you define a method for checking this and use that here instead? This is the style we've been trying to follow throughout.

ajreynol (changes requested):
One minor style comment, otherwise, looks good.

PR #5431, opened 2020-11-12
(proof-new) Make sure CAD proofs are properly closed.

nafur:
This PR makes sure that CAD proofs are properly reset for the next theory call and closes them with a scope using the minimal infeasible subset.


ajreynol (changes requested):

ajreynol (approved):
LGTM

PR #5432, opened 2020-11-12
Model declarations printing options

yoni206:
This PR relates to #4987 .
Our plan is to:

1. delete the model keyword
2. avoid printing extra declarations by default
3. wrap UF values with as expressions.

This PR is step 2. It allows the user to choose the model printing style in case of uninterpreted elements: either using datatypes, or using `declare-sort`s and `declare-fun`s or just using `declare-fun`s.

The default option, which is closest to the standard, is just using `declare-fun`s. In step 3, these will be replaced by abstract values using `as`.


src/options/smt_options.toml
```diff
@@ -412,13 +412,24 @@ header = "options/smt_options.h"
...
+  help_mode  = "uninterpreted elements in models printing  modes."
+[[option.mode.DtEnum]]
+  name = "dtenum"
+  help = "print uninterpreted elements as datatype enumerations, where the sort is the datatype"
+[[option.mode.DeclSortAndFun]]
+  name = "decl-sort-and-fun"
+  help = "print uninterpreted elements declare-fun, and also include a declare-sort for the sort"
+[[option.mode.DeclFun]]
+  name = "decl-fun"
+  help = "print uninterpreted elements declare-fun, but don'tinclude a declare-sort for the sort"
```
ajreynol, line 432:
```suggestion
  help = "(default) print uninterpreted elements declare-fun, but don't include a declare-sort for the sort"
```

ajreynol (changes requested):

src/options/smt_options.toml
```diff
@@ -412,13 +412,24 @@ header = "options/smt_options.h"
...
+  read_only  = true
+  help       = "determines how to print uninterpreted elements in models"
+  help_mode  = "uninterpreted elements in models printing  modes."
+[[option.mode.DtEnum]]
+  name = "dtenum"
+  help = "print uninterpreted elements as datatype enumerations, where the sort is the datatype"
+[[option.mode.DeclSortAndFun]]
+  name = "decl-sort-and-fun"
+  help = "print uninterpreted elements declare-fun, and also include a declare-sort for the sort"
+[[option.mode.DeclFun]]
```
ajreynol, line 430: My understanding the SMT-LIB standard was that we wouldn't include declare-fun either, is this right?  In this case, I'd expect another option, which would be "None" for don't print anything.

ajreynol (changes requested):

src/options/smt_options.toml
yoni206: I agree, but my plan was to hold off with the "none" option until we resolve the `as` issue. Printing neither `declare-fun` nor `as` is not very intuitive in my opinion.

ajreynol: Ah ok, thats fine.  Then the default will change to None afterwards?

yoni206: yes

ajreynol (approved):
LGTM

PR #5434, opened 2020-11-12
Add more features to symbol manager

ajreynol:
This is in preparation for having the symbol manager manage expression names at the parser level instead of inside SmtEngine.

This adds some necessary features regarding scopes and global declarations.

This PR still does not change the behavior of the parser.


src/expr/symbol_manager.cpp
```diff
@@ -57,12 +65,19 @@ class SymbolManager::Implementation
...
+  CDO<bool> d_hasPushedScope;
 };
 
 bool SymbolManager::Implementation::setExpressionName(api::Term t,
                                                       const std::string& name,
                                                       bool isAssertion)
 {
+  // cannot name subexpressions under quantifiers
+  PrettyCheckArgument(
+      !d_hasPushedScope.get(), name, "cannot name function in a scope");
```
nafur, line 80: Isn't this a restriction compared to SMTLIB?

nafur (approved):
Just one question, otherwise LGTM.

src/expr/symbol_manager.cpp
ajreynol: This is to be determined.  Andres and I decided that we should (at the moment) not support this, since it is significantly more complicated to support and is sometimes nonsensical, since :named within quantifier bodies should never be supported.

PR #5436, opened 2020-11-13
(proof-new) Enable proofs for datatypes

ajreynol:
This enables initial proof support in datatypes, which includes connecting the inference-to-proof constructor to the inference manager and making the proper calls to the inference manager using TheoryDatatypes.


src/theory/datatypes/inference_manager.cpp
```diff
@@ -65,6 +76,34 @@ void InferenceManager::process()
...
 }
 
+void InferenceManager::sendDtLemma(Node lem,
+                                   InferId id,
+                                   LemmaProperty p,
+                                   bool doCache)
+{
+  if (isProofEnabled())
+  {
+    processDtLemma(lem, Node::null(), id);
```
nafur, line 86: Just for readability (and deal with stats in one place):
```suggestion
    if (processDtLemma(lem, Node::null(), id))
    {
      d_inferenceLemmas << id;
    }
```

src/theory/datatypes/inference_manager.cpp
```diff
@@ -78,42 +117,106 @@ bool InferenceManager::sendLemmas(const std::vector<Node>& lemmas)
...
+    d_lemPg->setProofFor(lem, pn);
+  }
+  // use trusted lemma
+  TrustNode tlem = TrustNode::mkTrustLemma(lem, d_lemPg.get());
+  if (!trustedLemma(tlem))
+  {
+    Trace("dt-lemma-debug") << "...duplicate lemma" << std::endl;
+    return false;
+  }
+  d_inferenceLemmas << id;
```
nafur, line 162:
```suggestion
```
Moved to sendDtLemma.

nafur (approved):
Just one suggestion.

src/theory/datatypes/inference_manager.cpp
ajreynol: Actually, this would not keep stats if proofs were disabled, right?

nafur: it would. If proofs are disabled, then stats are collected immediately below.

ajreynol: So I see the issue and now remember why I did it like this.  These are two different methods, this one sends lemmas immediately, the other one processed lemmas that were buffered.  The two control paths are disjoint when proofs are disabled.  This change would make it so that lemmas that were buffered and processed would not be counted when proofs are disabled.

PR #5437, opened 2020-11-13
(proof-new) Add proofs for factoring lemmas.

nafur:
This PR adds proofs for lemmas from the factoring subsolver from the nonlinear extension.


src/theory/arith/nl/ext/factoring_check.cpp
```diff
@@ -120,7 +121,13 @@ void FactoringCheck::check(const std::vector<Node>& asserts,
           sum = Rewriter::rewrite(sum);
           Trace("nl-ext-factor")
               << "* Factored sum for " << x << " : " << sum << std::endl;
-          Node kf = getFactorSkolem(sum);
+
+          LazyCDProof* proof = nullptr;
```
ajreynol, line 125: Any reason why this proof is not just allocated more local within getFactorSkolem?

ajreynol (changes requested):

ajreynol (approved):

PR #5439, opened 2020-11-13
(proof-new) Add proofs for NL_TANGENT_PLANE

nafur:
This PR adds proofs for the NL_TANGENT_PLANE lemmas.
It introduces two new proof rules with proof checkers.


src/expr/proof_rule.h
```diff
@@ -1102,6 +1102,32 @@ enum class PfRule : uint32_t
   // ---------------------
   // Conclusion: (Q)
   INT_TRUST,
+  //======== Multiplication tangent plane
```
ajreynol, line 1105: Do these have to be 2 different rules?  You could merge the cases of these 2 rules. Also, do you need 4 cases or can you permute x/y/a/b to replicate the cases?  Maybe 3 cases instead of 4?  Meaning 6 cases overall, in one rule?

ajreynol (changes requested):

src/expr/proof_rule.h
nafur: Observe that the two rules prove two opposite implications. We should use a single rule that infers the equivalence and also learn this equivalence as lemma...

ajreynol (approved):

PR #5440, opened 2020-11-13
(proof-new) Prepare transcendental solver for proofs

nafur:
This PR does a couple of relatively minor refactorings to prepare the transcendental subsolver for proofs:
- Use `ExtState` instead of manually holding the inference manager and the model.
- Don't collect lemmas and send them in a conjunction, but send them individually.
- Fix a bug in thew in monotonicity lemma
- Allow to retrieve the actual degree of taylor approximations that was used


ajreynol (approved):
LGTM

PR #5442, opened 2020-11-13
Cleaning up scopes in preparation for symbol manager

ajreynol:
This changes the default of Parser::pushScope and prepares symbol manager further for maintaining expression names.


nafur (approved):
LGTM.

PR #5444, opened 2020-11-14
Add let binding utility

ajreynol:
This utility will replace the dagification visitor for printing lets for Node in the smt2 printer, and will be used further in e.g. LFSC proof printing.


src/printer/let_binding.cpp
```diff
@@ -0,0 +1,216 @@
...
+      }
+      else
+      {
+        d_count[cur] = 0;
+        visit.insert(visit.end(), cur.begin(), cur.end());
+      }
+    }
+    else
+    {
+      if ((*it).second == 0)
```
mpreiner, line 175: Maybe add a comment why you push `cur` onto `d_visitList`. Can you also move this statement to the else block in line 169?

src/printer/let_binding.cpp
```diff
@@ -0,0 +1,216 @@
...
+      visit.pop_back();
+    }
+  } while (!visit.empty());
+}
+
+void LetBinding::convertCountToLet()
+{
+  Assert(d_thresh > 0);
+  // Assign ids for those whose d_count is >= d_thresh, traverse in d_visitList
+  // in order so that deeper proofs are assigned lower identifiers, which
```
mpreiner, line 189: deeper proofs?

src/printer/let_binding.cpp
```diff
@@ -0,0 +1,216 @@
...
+      continue;
+    }
+    else if (d_letMap.find(n) != d_letMap.end())
+    {
+      // already letified, perhaps at a lower context
+      continue;
+    }
+    itc = d_count.find(n);
+    Assert(itc != d_count.end());
+    if ((*itc).second >= d_thresh)
```
mpreiner, line 206:
```suggestion
    if (itc->second >= d_thresh)
```

src/printer/let_binding.cpp
ajreynol: No, it can't be moved there, since the order is important: we can only add to d_visitList once all subterms of cur are in d_visitList.

src/printer/let_binding.cpp
ajreynol: Will fix, this was copied from the proof letifier accidentally.

src/printer/let_binding.h
```diff
@@ -0,0 +1,150 @@
...
+   * @param letTop Whether we letify n itself
+   * @return the converted node.
+   */
+  Node convert(Node n, const std::string& prefix, bool letTop = true) const;
+
+ private:
+  /**
+   * Compute the count of sub nodes in pn, store in pcount. Additionally,
+   * store each node in the domain of pcount in an order in visitList
+   * such that visitList[i] does not contain sub visitList[j] for j>i.
```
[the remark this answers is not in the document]
ajreynol: This changed names, will fix.

mpreiner (approved):

PR #5449, opened 2020-11-16
Refactor tangent plane lemmas

nafur:
The original lemma only proposed an implication for the tangent plane lemmas, while our implementation adds the inverse implication as well (in a somewhat verbose way).
This PR replaces this by simply constructing the equivalence.

### DIFF
--- a/src/theory/arith/nl/ext/tangent_plane_check.cpp
+++ b/src/theory/arith/nl/ext/tangent_plane_check.cpp
@@ -122,11 +122,12 @@ void TangentPlaneCheck::check(bool asWaitingLemmas)
               Node ab =
                   nm->mkNode(d == 1 || d == 3 ? Kind::GEQ : Kind::LEQ, b, b_v);
               Node conc = nm->mkNode(d <= 1 ? Kind::LEQ : Kind::GEQ, t, tplane);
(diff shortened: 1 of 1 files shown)


src/theory/arith/nl/ext/tangent_plane_check.cpp
```diff
@@ -115,76 +115,20 @@ void TangentPlaneCheck::check(bool asWaitingLemmas)
...
               Node ab =
                   nm->mkNode(d == 1 || d == 3 ? Kind::GEQ : Kind::LEQ, b, b_v);
               Node conc = nm->mkNode(d <= 1 ? Kind::LEQ : Kind::GEQ, t, tplane);
-              Node tlem = nm->mkNode(Kind::OR, aa.negate(), ab.negate(), conc);
+              Node tlem =
+                  nm->mkNode(Kind::EQUAL, nm->mkNode(Kind::AND, aa, ab), conc);
               Trace("nl-ext-tplanes")
                   << "Tangent plane lemma : " << tlem << std::endl;
-              tplaneConj.push_back(tlem);
+              d_data->d_im.addPendingArithLemma(
```
ajreynol, line 129: Can you keep/update the comment from below?

ajreynol (changes requested):

ajreynol (approved):

PR #5451, opened 2020-11-16
Use symbol manager for get assignment

ajreynol:
This replaces the previous methods for get-assignment which involving making a DefinedNamedFunctIonCommand and storing the expression names map in the SmtEngine.

Note: we now limit :named terms to those not beneath quantifiers and let-bindings.

### DIFF
--- a/src/smt/command.cpp
+++ b/src/smt/command.cpp
@@ -1663,15 +1663,15 @@ void GetAssignmentCommand::invoke(api::Solver* solver, SymbolManager* sm)
 {
   try
   {
-    std::vector<std::pair<Expr, Expr>> assignments =
-        solver->getSmtEngine()->getAssignment();
-    vector<SExpr> sexprs;
-    for (const auto& p : assignments)
+    std::map<api::Term, std::string> enames = sm->getExpressionNames();
+    std::vector<SExpr> sexprs;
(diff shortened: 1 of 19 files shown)


ajreynol:
Marking "do not merge" for now until we decide if the restriction of not supporting :named beneath quantifiers and let-bindings is acceptable.

src/smt/command.cpp
```diff
@@ -1663,20 +1612,29 @@ void GetAssignmentCommand::invoke(api::Solver* solver, SymbolManager* sm)
...
-      vector<SExpr> v;
-      v.emplace_back(SExpr::Keyword(p.first.toString()));
-      v.emplace_back(SExpr::Keyword(p.second.toString()));
-      sexprs.emplace_back(v);
+      terms.push_back(e.first);
+      names.push_back(e.second);
+    }
+    // Must use vector version of getValue to ensure error is thrown regardless
+    // of whether terms is empty.
+    std::vector<api::Term> values = solver->getValue(terms);
```
nafur, line 1625: Just to be sure:
```suggestion
    std::vector<api::Term> values = solver->getValue(terms);
    Assert(values.size() == names.size());
```

nafur (approved):
One minor suggestions, otherwise LGTM.

PR #5455, opened 2020-11-17
Fix tangent plane lemmas

### LINKED ISSUE #5452: (nl-ext-tplanes) refutational unsoundness bug
```
[633] % timeout -s 9 30 cvc4 small.smt2
Killed
[634] % cvc4 --nl-ext-tplanes small.smt2
unsat
sat
[635] % 
[635] % cat small.smt2
(set-logic ALL)
(declare-fun a () Real)
(assert (< (/ a a (- 3 (* a (+ a 1)))) 0))
(check-sat)
(reset)
(set-logic ALL)
(define-fun a () Real (- 3.0))
(assert (< (/ a a (- 3 (* a (+ a 1)))) 0))
(check-sat)
[636] %
```
OS: Ubuntu 18.04
Commit: 2b504cd

nafur:
The previous refactoring of tangent plane lemmas introduced incorrect lemmas.
This PR fixes this issue and actually generated the lemmas described in the comment.
Fixes #5452.


ajreynol (approved):
Does the benchmark from the issue time out now?  If its solved, add it here.

nafur:
> Does the benchmark from the issue time out now? If its solved, add it here.

It times out, unfortunately...

PR #5465, opened 2020-11-18
(proof-new) Fix CAD proofs (again)

nafur:
The CAD solver occasionally removed redundant intervals which sometimes leads to smaller conflicts.
However, proofs for intervals are eagerly constructed and were not pruned properly.
Such proofs are not closed as these obsolete proof branches assume assertions that are not scoped if they are no longer part of the conflict.
This PR fixes this issue by adding a generic pruning mechanism to `LazyTreeProofGenerator` and using it appropriately for CAD proofs.

Furthermore, the CAD solver did simply reset its instance of `LazyTreeProofGenerator` for every new theory call. This does not work and is also fixed by this PR. `LazyTreeProofGenerator` no longer allows for resets and now is a proper `ProofGenerator` itself and `CADProofGenerator` issues a new `LazyTreeProofGenerator` for every theory call.


ajreynol (approved):
LGTM

PR #5466, opened 2020-11-18
Fix for sygusToBuiltinEval for non-ground composite terms

ajreynol:
There was a bug in the method for computing the evaluation of a sygus term applied to arguments.

The case that was wrong was for  (DT_SYGUS_EVAL t c1 ... cn) where `t` contained a subterm `(op t1 ... tn)` where:
(1) `(op t1 ... tn)` is a non-ground sygus datatype term
(2) `op` was an operator of the form `(lambda ((z T)) (g z xi))` where `xi` is a variable from the formal argument list of the function to synthesize.
In this case, `xi` was not getting replaced by `ci`.

This bug appears when using sygus repair for grammars with composite term that involve variables from the formal argument list of the synth-fun.


src/theory/datatypes/sygus_datatype_utils.cpp
```diff
@@ -37,12 +37,17 @@ Node applySygusArgs(const DType& dt,
                     Node n,
                     const std::vector<Node>& args)
 {
+  // optimization: if n is just a sygus bound variable, return immediately
```
HanielB, line 40: please describe the two different cases

src/theory/datatypes/sygus_datatype_utils.h
```diff
@@ -122,18 +122,18 @@ Node mkSygusTerm(Node op,
...
- *   C2 / (lambda w1. w1+x)
- *   C3 / (lambda w1. w1+(x+y))
- *   C4 / y
+ *   C1 / L1 where L1 is (lambda w1 w2. w1+w2)
+ *   C2 / L2 where L2 is (lambda w1. w1+x)
+ *   C3 / L3 where L3 is (lambda w1. w1+(x+y))
+ *   C4 / L4 where L4 is y
  * Examples of calling this function:
- *   applySygusArgs( dt, C1, (APPLY_UF (lambda w1 w2. w1+w2) t1 t2), { 3, 5 } )
+ *   applySygusArgs( dt, L1, (APPLY_UF L5 t1 t2), { 3, 5 } )
```
HanielB, line 130:
```suggestion
 *   applySygusArgs( dt, L1, (APPLY_UF L1 t1 t2), { 3, 5 } )
```

src/theory/datatypes/sygus_datatype_utils.h
```diff
@@ -122,18 +122,18 @@ Node mkSygusTerm(Node op,
...
  *     ... returns (APPLY_UF (lambda w1 w2. w1+w2) t1 t2).
- *   applySygusArgs( dt, C2, (APPLY_UF (lambda w1. w1+x) t1), { 3, 5 } )
+ *   applySygusArgs( dt, L2, (APPLY_UF L4 t1), { 3, 5 } )
```
HanielB, line 132: You really wanted to have L4 or was it supposed to be L2?

HanielB (approved):
LGTM, just minor comments. BTW don't we have a benchmark that we could add as a regression?

ajreynol:
The benchmark that exercises the issue times out.

PR #5468, opened 2020-11-18
Use symbol manager for unsat cores

ajreynol:
This PR uses the symbol manager for handling names for unsat cores.

This replaces interfaces in SmtEngine for managing expression names.  It removes the SetExpressionName command.


nafur (approved):
LGTM.
I find it kind of weird that UnsatCores have two modes (named or just nodes) and they don't really interact... but I guess this is fine.

PR #5474, opened 2020-11-19
(proof-new) Add new templated CDProofSet

nafur:
`LazyCDProofSet` provides a convenient way to dynamically allocate proofs. Unfortunately, this class is hard-coded to use `LazyCDProof` objects.
This PR replaces it by a templated `CDProofSet`, uses it instead of `LazyCDProofSet` everywhere and consequently removes `LazyCDProofSet`.


ajreynol (approved):
LGTM, this is great.  Btw, we should get some of these changes into master soon.  Can you do that after we merge this to proof-new?

nafur:
> LGTM, this is great. Btw, we should get some of these changes into master soon. Can you do that after we merge this to proof-new?

Sure.

PR #5480, opened 2020-11-19
Add declare model symbol methods to SymbolManager and Model

ajreynol:
This is in preparation for the symbol manager determining which symbols are printed in the model.


src/expr/symbol_manager.cpp
```diff
@@ -150,6 +166,36 @@ SymbolManager::Implementation::getExpressionNames(bool areAssertions) const
...
+       ++it)
+  {
+    declareSorts.push_back(*it);
+  }
+  for (TermList::const_iterator it = d_declareTerms.begin();
+       it != d_declareTerms.end();
+       ++it)
+  {
+    declareTerms.push_back(*it);
+  }
```
nafur, line 184:
```suggestion
  declareSorts.insert(declareSorts.end(), d_declareSorts.begin(), d_declareSorts.end());
  declareTerms.insert(declareTerms.end(), d_declareTerms.begin(), d_declareTerms.end());
```

nafur (changes requested):
One request. A general question: why not return sorts and terms separately in the SymbolManager as well, just like in the Model?

ajreynol:
I've split the access functions in symbol manager for sorts and terms.

nafur (approved):
LGTM.

PR #5487, opened 2020-11-20
(proof-new) Replace LazyCDProofSet by generic CDProofSet

nafur:
This PR replaces the `LazyCDProofSet`, that was hardcoded to use `LazyCDProof` objects, by a templated `CDProofSet`.


ajreynol (approved):
LGTM

PR #5489, opened 2020-11-20
(proof-new) Some new proofs for nonlinear multiplication

nafur:
This PR adds three new proofs for nonlinear multiplication, each adding a new proof rule.


src/expr/proof_rule.h
```diff
@@ -1102,6 +1102,30 @@ enum class PfRule : uint32_t
   // ---------------------
   // Conclusion: (Q)
   INT_TRUST,
+  //======== Multiplication sign inference
+  // Children: none
+  // Arguments: (f1, ..., fk, m)
+  // ---------------------
+  // Conclusion: (=> (and f1 ... fk) (~ m 0))
+  // Where f1, ..., fk are variables compared to zero (less, greater or not equal), m is a monomial from these variables, and ~ is the comparison (less or greater) that results from the signs of the variables.
```
ajreynol, line 1110: Format?

src/expr/proof_rule.h
```diff
@@ -1102,6 +1102,30 @@ enum class PfRule : uint32_t
...
+  // All variables with even exponent in m should be given as not equal to zero while all variables with odd exponent in m should be given as less or greater than zero.
+  ARITH_MULT_SIGN,
+  //======== Multiplication with positive factor
```
ajreynol, line 1113: I'm actually curious if we can use Alex's ARITH_SCALE_SUM_UPPER_BOUNDS for these two rules.

ajreynol (changes requested):

src/expr/proof_rule.h
ajreynol: Note this would require a SCOPE to conclude the implication as needed.

nafur: I think that the `ARITH_SCALE_SUM_UPPER_BOUNDS` rule does something different, though I can't completely follow the documentation. I'm pretty sure, though, that it only deals with linear terms and there is no multiplication between variables involved.
`ARITH_MULT_SIGN` on the other hand is explicitly about nonlinear monomials.
There may be corner cases where both are applicable, but not for the general case.

ajreynol: Would be good to get thoughts from @alex-ozdemir.

src/theory/arith/nl/ext/proof_checker.cpp
```diff
@@ -45,7 +48,123 @@ Node ExtProofRuleChecker::checkInternal(PfRule id,
...
+  else if (id == PfRule::ARITH_MULT_NEG)
+  {
+    Assert(children.empty());
+    Assert(args.size() == 3);
+    Node mult = args[0];
+    Node orig = args[1];
+    Kind rel = args[2].getKind();
+    Assert(rel == Kind::EQUAL || rel == Kind::DISTINCT || rel == Kind::LT
+           || rel == Kind::LEQ || rel == Kind::GT || rel == Kind::GEQ);
+    Kind rel_inv = reverseRelationKind(rel);
```
alex-ozdemir, line 157: I believe this crashes on `DISTINCT`.

src/expr/proof_rule.h
alex-ozdemir: These are not captured by ARITH_SCALE_SUM_UPPER_BOUNDS. As Gereon suggests, ARITH_SCALE_SUM_UPPER_BOUNDS requires the scalars to be rational constants, and both ARITH_MULT_SIGN and ARITH_MULT_POS can apply to non-constant multiplicands.

If we want to get the arith rules down to a nicer core, there are two steps that make sense to me:

   1. make ARITH_MULT_SIGN a macro on top of ARITH_MULT_POS and ARITH_MULT_NEG.
   2. define an ARITH_SUM rule, let this:
```
// Children: ((><1 a1 b1), ..., (><n an bn))
// Arguments: none
// Conclusion: (>< (+ a1 ... an) (+ b1 ... bn))
// where each ><i is a comparator, and >< is the "sum" of those comparators.
```
and then we could redefine ARITH_SCALE_SUM_UPPER_BOUNDS as a macro on top of ARITH_SUM, ARITH_MULT_POS, and ARITH_MULT_NEG.

Probably no need to do either in this PR.

src/theory/arith/nl/ext/proof_checker.cpp
```diff
@@ -45,7 +48,123 @@ Node ExtProofRuleChecker::checkInternal(PfRule id,
...
+    {
+      auto it = exps.find(v);
+      if (it == exps.end())
+      {
+        exps.emplace(v, 1);
+      }
+      else
+      {
+        ++it->second;
+      }
```
alex-ozdemir, line 71: Fun fact about `std::map`: you can access missing entries using `operator[]` and it default-initializes (i.e. equivalent to `{}`-initilizing) the value for that key if there is no value there already.

ergo, `exps[v]++` suffices here.

Perhaps this shouldn't be used though... I'm not sure whether most C++ devs know this is how the language works.

alex-ozdemir (commented):
Left a few comments inline.
Sorry for the delay. Didn't realize I had been pinged on this PR.

ajreynol (approved):
LGTM, we should do the macro elimination that Alex mentions, but this isn't high priority right now.

PR #5491, opened 2020-11-20
Fix quantifiers scope issue in strings preprocessor

### LINKED ISSUE #5483: (strings-lazy-pp) Assertion failure at src/theory/theory_engine.cpp:1404
CVC4 f2ed7b1aebc
~~~~
(set-logic QF_SLIA)
(declare-fun _substvar_21_ () String)
(declare-fun _substvar_29_ () String)
(assert (xor true true true true (str.<= _substvar_21_ _substvar_29_) true true))
(push 1)
~~~~

~~~~
./cvc4 xx.smt2 -i 
Fatal failure within CVC4::theory::LemmaStatus CVC4::TheoryEngine::lemma(CVC4::theory::TrustNode, CVC4::theory::LemmaProperty, CVC4::theory::TheoryId, CVC4::theory::TheoryId) at /home/CVC4/src/theory/theory_engine.cpp:1404
Check failure

 !expr::hasFreeVar(lemma)

Aborted
~~~~

ajreynol:
Leads to free variables in assertions when using `str.<=` whose reduction uses EXISTS not FORALL.

Fixes #5483.


4tXJ7f (approved):
LGTM

PR #5493, opened 2020-11-20
Fix compiler warnings.

aniemetz:


ajreynol (approved):

src/theory/arith/theory_arith_private.cpp
```diff
@@ -3478,7 +3478,6 @@ bool TheoryArithPrivate::postCheck(Theory::Effort effortLevel)
   }
   d_statistics.d_avgUnknownsInARow.addEntry(d_unknownsInARow);
 
-  size_t nPivots = options::useFC() ? d_fcSimplex.getPivots() : d_dualSimplex.getPivots();
```
mpreiner, line 3481: Shouldn't `nPivots` be actually the upper bound for the loop below, i.e, `i < nPivots`?

aniemetz: @nafur

mpreiner: The code was added here: 
https://github.com/CVC4/CVC4/pull/4763/files#diff-26108011619ab0aae57344114f75c1ad1e596272b61950ab6167ab37c5b1b6a4R3482

mpreiner: Looks like @nafur forgot to change it to `i < nPivots`

aniemetz: Addressed in #5495.

PR #5496, opened 2020-11-20
Eliminate uses of SExpr from the parser.

abdoo8080:
This is a step towards migrating Commands to the API. SExpr is an internal module that's not used by the API but is used by the parser and some commands. This PR replaces SExpr with terms of kind SEXPR in the parser and strings in commands (which brings them more inline with the API).


src/smt/command.cpp
```diff
@@ -45,6 +45,29 @@ using namespace std;
 
 namespace CVC4 {
 
+std::string sexprToString(api::Term sexpr)
```
ajreynol, line 48: Can you say a bit how this is different from how the e.g. smt2 printer prints this term? (I assume this is doing some special case that you wouldn't get with sexpr.toString()?)  Also, is this file the appropriate place for this function?

ajreynol (changes requested):

src/smt/command.cpp
abdoo8080: Added a comment to explain the difference. As for the appropriate place for the method, I simply picked `command.h` because I need the method for some commands and the parser has access to `command.h`. I am fine with other locations though.

ajreynol: I'm still thinking whether its necessary.  Notice that to drop the parentheses, you can use `t[0].toString()` if you know `t` has one child.
In fact, is this method equivalent to `t.getNumChildren()==1 ? t[0].toString() : t.toString()`?

ajreynol: I'm still pretty confused by the method.

Why would we construct a constant string with parenthesis around it?  I would expect that term to be an sexpr with one child so that we always expect a term of kind SEXPR as input to this method.  However, then this method seems to be equivalent to `t.getNumChildren()==1 ? t[0].toString() : t.toString()` assuming t is an SEXPR.

ajreynol: Discussed offline, I believe this method is fine for now.

ajreynol (approved):

ajreynol:
LGTM assuming @aniemetz is okay with the changes related to API exceptions.

aniemetz (approved):

PR #5499, opened 2020-11-20
Change UF ho to ppRewrite instead of expand definition

ajreynol:
UF uses expandDefinitions to convert fully applied HO_APPLY to APPLY_UF.  The more appropriate place to do this is in Theory::ppRewrite.


test/regress/regress1/ho/NUM925^1.p
```diff
@@ -1,4 +1,4 @@
-% COMMAND-LINE:  --uf-ho
+% COMMAND-LINE:  --uf-ho --ho-elim
```
HanielB, line 1: Why did you need to add the option?

ajreynol: Because we answer unknown otherwise.

HanielB: But we were solving them without the pass before? Why is that?

ajreynol: This appears to be random.

HanielB: I see.

HanielB (approved):

PR #5512, opened 2020-11-23
Fix a bug with synth-fun printer

### LINKED ISSUE #5448: NULL pointer dereference on verbosity command
Hi, for the following formula, CVC4 a71274b992
~~~~
(set-logic NIA)
(synth-inv InvF ((x1 Int)))
(check-synth)

$ /cvc4 x.sl  -vvv
Invoking: (set-option :incremental false)

Invoking: (set-logic NIA)

CVC4 suffered a segfault in DEBUG mode.
Offending address is 0xe0
Looks like a NULL pointer was dereferenced.
Invoking: Aborted
~~~~

abdoo8080:
This PR fixes #5448. `SynthFunCommand::toStream` used to call `d_grammar->resolve` even when `d_grammar` is a `nullptr`. This PR fixes the issue and modifies the signature of `Printer::toStreamCmdSynthFun` to make it clear that grammar is an optional argument.


src/printer/printer.h
```diff
@@ -137,7 +137,7 @@ class Printer
                                    const std::vector<Node>& vars,
                                    TypeNode range,
                                    bool isInv,
-                                   TypeNode sygusType) const;
+                                   TypeNode* sygusType = nullptr) const;
```
ajreynol, line 140: I would prefer not using TypeNode*, which is conceptually a double pointer.  You can do `TypeNode sygusType = TypeNode::null()` here.

ajreynol (changes requested):
I'd keep the signature the same and use TypeNode::null() instead of nullptr, but the fix looks correct to be.

ajreynol (approved):

ajreynol:
@abdoo8080 can you also add the regression from the issue?

PR #5516, opened 2020-11-23
Use symbol manager for printing responses get-model

ajreynol:
This makes symbol manager be in charge of determining which sorts and terms to print in response to get-model.  This eliminates the need for the parser to call ExprManager::mkVar (and similar methods) with custom flags.  

This requires significant simplifications to the printers for models, where instead of a NodeCommand, we take a Sort or Term to print in the model.

This is one of the last remaining steps for migrating the parser to the new API.

The next step will be to remove a lot of the internal infrastructure for managing expression names, commands to print in models, node commands, node listeners, etc.


src/printer/ast/ast_printer.cpp
```diff
@@ -272,12 +280,9 @@ void AstPrinter::toStreamCmdDefineFunction(std::ostream& out,
 }
 
 void AstPrinter::toStreamCmdDeclareType(std::ostream& out,
-                                        const std::string& id,
-                                        size_t arity,
                                         TypeNode type) const
```
abdoo8080, line 283: Reformat here and in other files.

src/parser/parser.cpp
```diff
@@ -379,10 +377,11 @@ api::Sort Parser::mkUnresolvedTypeConstructor(
...
   Debug("parser") << "newSortConstructor(P)(" << name << ", " << params.size()
                   << ")" << std::endl;
-  api::Sort unresolved =
-      api::Sort(d_solver,
-                d_solver->getExprManager()->mkSortConstructor(
-                    name, params.size(), ExprManager::SORT_FLAG_PLACEHOLDER));
+  // api::Sort unresolved =
+  //    api::Sort(d_solver,
+  //              d_solver->getExprManager()->mkSortConstructor(
+  //                  name, params.size(), ExprManager::SORT_FLAG_PLACEHOLDER));
```
abdoo8080, line 383: Why are these lines commented out and not removed?

src/printer/printer.h
```diff
@@ -266,18 +264,28 @@ class Printer
...
+                                 const smt::Model& m,
+                                 TypeNode tn) const = 0;
+
+  /**
+   * To stream model term. This prints the appropriate output for type
+   * tn declared via declare-fun.
+   */
+  virtual void toStreamModelTerm(std::ostream& out,
+                                 const smt::Model& m,
+                                 Node n) const = 0;
```
abdoo8080, line 281: Those methods don't seem to be called from outside the printers. Consider making them internal.

test/unit/parser/parser_black.h
```diff
@@ -259,8 +259,6 @@ class Cvc4ParserTest : public CxxTest::TestSuite, public ParserBlack {
     tryGoodInput("a : INT = 5; a: INT;"); // decl after define, compatible
     tryGoodInput("a : TYPE; a : INT;"); // ok, sort and variable symbol spaces distinct
     tryGoodInput("a : TYPE; a : INT; b : a;"); // ok except a is both INT and sort `a'
-    //tryGoodInput("a : [0..0]; b : [-5..5]; c : [-1..1]; d : [ _ .._];"); // subranges
-    tryGoodInput("a : [ _..1]; b : [_.. 0]; c :[_..-1];");
```
abdoo8080, line 263: Why did you remove those two lines? You don't need to explain it in the code. I just want to know.

abdoo8080 (changes requested):
I like where we are going with this PR.

test/unit/parser/parser_black.h
ajreynol: It was a comedy of errors that this worked previously :)

Essentially, this is a testing a feature (subrange types) that was supported in cvc3 but we never added support for in cvc4.  Moreover, the cvc parser incorrectly did not issue official define-fun commands until I fixed this a few PRs ago. Once this was fixed, this still did not trigger an issue since the parser was using the old API and did not complain about making variables of null types.  Now, we call the new API, which correctly recognized that this is not a supported input.

src/printer/printer.h
ajreynol: They need to be virtual protected functions, because the printer has a generic way of printing models that calls these commands (which is then specific to the format).

src/parser/parser.cpp
ajreynol: I'll remove

src/printer/ast/ast_printer.h
```diff
@@ -165,9 +163,21 @@ class AstPrinter : public CVC4::Printer
...
+   * To stream model sort. This prints the appropriate output for type
+   * tn declared via declare-sort or declare-datatype.
+   */
+  void toStreamModelSort(std::ostream& out,
+                         const smt::Model& m,
+                         TypeNode tn) const override;
+
+  /**
+   * To stream model term. This prints the appropriate output for type
+   * tn declared via declare-fun.
```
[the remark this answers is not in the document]
ajreynol: Yes

src/printer/printer.h
```diff
@@ -266,18 +264,28 @@ class Printer
...
 
   /** write model response to command using another language printer */
   void toStreamUsing(OutputLanguage lang,
                      std::ostream& out,
-                     const smt::Model& m,
-                     const NodeCommand* c) const
+                     const smt::Model& m) const
   {
-    getPrinter(lang)->toStream(out, m, c);
+    getPrinter(lang)->toStream(out, m);
```
nafur, line 288: Move to cpp?

nafur (approved):
Just a minor comment, otherwise LGTM.

abdoo8080 (approved):
LGTM.

PR #5518, opened 2020-11-23
Fix regular expression consume for nested star

### LINKED ISSUE #5510: Refutational soundness bug in QF_S
```
[614] % cvc4 small.smt2
unsat
sat
[615] % 
[615] % cat small.smt2
(set-logic QF_S)
(declare-fun x () String)
(declare-fun y () String)
(assert (str.in_re (str.++ x "B" y) (re.* (re.++ (str.to_re "A") (re.union (re.* (str.to_re "A")) (str.to_re "B"))))))
(assert (str.in_re y (str.to_re "A")))
(check-sat)
(reset)
(set-logic QF_S)
(define-fun x () String "A")
(define-fun y () String "A")
(assert (str.in_re (str.++ x "B" y) (re.* (re.++ (str.to_re "A") (re.union (re.* (str.to_re "A")) (str.to_re "B"))))))
(assert (str.in_re y (str.to_re "A")))
(check-sat)
[616] %
```
OS: Ubuntu 18.04
Commit: 36af095

ajreynol:
The issue was caused by simple regular expression consume being too aggressive when a recursive call was made to the method.  In particular, we were assuming that the body of the star must be unrolled to fully consume a string, when it can be skipped if we are not at top level.

Fixes #5510.

### DIFF
--- a/src/theory/strings/regexp_entail.cpp
+++ b/src/theory/strings/regexp_entail.cpp
@@ -252,8 +252,10 @@ Node RegExpEntail::simpleRegexpConsume(std::vector<Node>& mchildren,
             {
               if (children_s.empty())
               {
-                // check if beyond this, we can't do it or there is nothing
-                // left, if so, repeat
+                // Check if beyond this, we hit a conflict. In this case, we
+                // must repeat.  Notice that we do not treat the case where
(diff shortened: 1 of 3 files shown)


4tXJ7f (approved):
LGTM

PR #5525, opened 2020-11-25
Fix transcendental secant plane lemmas

nafur:
The refactoring of the transcendental solver introduced a subtle issue that lead to incorrect secant plane lemmas.
This PR fixes this issue, so that we now produce the correct lemmas again.


src/theory/arith/nl/transcendental/transcendental_state.cpp
```diff
@@ -294,12 +294,12 @@ std::pair<Node, Node> TranscendentalState::getClosestSecantPoints(TNode e,
...
 
 Node TranscendentalState::mkSecantPlane(
-    TNode arg, TNode b, TNode c, TNode approx, TNode approx_c)
+    TNode arg, TNode b, TNode c, TNode approx_b, TNode approx_c)
 {
   NodeManager* nm = NodeManager::currentNM();
   // Figure 3 : P(l), P(u), for s = 0,1
-  Node approx_b =
-      Rewriter::rewrite(approx.substitute(d_taylor.getTaylorVariable(), b));
+  // Node approx_b =
```
ajreynol, line 301: Is this really not needed? If so delete.

src/theory/arith/nl/transcendental/transcendental_state.h
```diff
@@ -89,119 +89,120 @@ struct TranscendentalState
...
-    Node d_pi_2;
-    /** -PI/2 */
-    Node d_pi_neg_2;
-    /** -PI */
-    Node d_pi_neg;
-    /** the concrete lower and upper bounds for PI */
-    Node d_pi_bound[2];
+                      int concavity);
+
+  Node d_true;
```
ajreynol, line 107: May be a formatting issue here.

src/theory/arith/nl/transcendental/transcendental_state.cpp
nafur: Did so, and moved the comment to where it now belongs.

src/theory/arith/nl/transcendental/transcendental_state.h
nafur: Yes, but the new formatting is the right one.

ajreynol (approved):

PR #5528, opened 2020-11-25
Move expand definitions to its own file

ajreynol:
We are changing our policy on how expand definitions is handled during preprocessing.

This will require some additions to expand definitions handling. This PR makes a standalone module for expanding definitions.


nafur (approved):
LGTM.

PR #5531, opened 2020-11-25
Fix missed case of theory preprocessing

ajreynol:
This fixes a rare case of theory preprocessing where rewriting a quantified formula would introduce a term that would not be preprocessed.  This led to solution unsoundness on a branch where the relationship between expand definitions and preprocessing is modified.

This is required for updating to the new policy for expand definitions.


4tXJ7f (approved):
Can we add a regression that exercises this?

ajreynol:
https://github.com/CVC4/CVC4/blob/master/test/regress/regress1/quantifiers/issue3481.smt2 exercises this behavior.  On master before this PR, there are terms that are not theory preprocessed (but this is benign since they are processed at expandDefinitions).  This PR is critical to continue to answer correctly for this benchmark after the planned changes.

PR #5532, opened 2020-11-25
Fully decouple SmtEngine and the Expr layer

ajreynol:
This removes the remaining dependencies of SmtEngine on the Expr layer. It now takes a NodeManager instead of a ExprManager.


4tXJ7f (approved):
LGTM, nice :)

PR #5536, opened 2020-11-26
Remove includes for old API from internal code

ajreynol:
The only code including the old API now are in parser/ and main/ which will require further untangling.


4tXJ7f (approved):
Since expr/array.h has been deleted, you also need to remove it from [here](https://github.com/CVC4/CVC4/blob/865d1ee48de8e4a21d1e97c707be46c34918367d/src/CMakeLists.txt#L1140), which is why the CI fails. LGTM otherwise.

PR #5539, opened 2020-11-27
Refactor transcendental solver

nafur:
This PR does another round of refactoring on the transcendental solver. It cleans up some variable names, introduces an enum type for `Convexity` and passes both the intended taylor degree and the actual taylor degree (which will be needed for proofs).


ajreynol (approved):
LGTM, thanks for all this refactoring.

PR #5544, opened 2020-11-30
Do not expand theory definitions at the beginning of preprocessing

ajreynol:
This updates the preprocessor so that expand definitions does not expand theory symbols at the beginning of preprocessing.

This also restores the previous expandDefinitions method in arithmetic, which is required for correctly interpreting division by zero in models, but should not be applied at the beginning of preprocessing.  Moreover it ensures that only partial operators are eliminated in arithmetic expandDefinitions, which required an additional argument `partialOnly` to arith::OperatorElim.

This adds `-q` to suppress warnings for many quantified regressions which now emit warnings with `--check-model`.  This will be addressed later as part of https://github.com/CVC4/cvc4-wishues/issues/43.

The purpose of this PR is two-fold:
(1) Currently our responses to `get-value` are incorrect for partial operators like `div`, `mod`, `seq.nth` since partial operators can be left unevaluated.
(2) The preprocessor should have the opportunity to rewrite and eliminate extended operators before they are expanded. This is required for addressing performance issues for non-linear arithmetic.  It is also required for ensuring that trigger selection can be done properly for datatype selectors (to be addressed on a later PR).


src/theory/bv/theory_bv.cpp
```diff
@@ -200,7 +200,16 @@ Theory::PPAssertStatus TheoryBV::ppAssert(
   return d_internal->ppAssert(tin, outSubstitutions);
 }
 
-TrustNode TheoryBV::ppRewrite(TNode t) { return d_internal->ppRewrite(t); }
+TrustNode TheoryBV::ppRewrite(TNode t)
```
yoni206, line 203: Does this change the default behavior for the BV theory? That is, when will `texp` be null?

src/theory/fp/theory_fp.cpp
```diff
@@ -453,6 +453,12 @@ TrustNode TheoryFp::expandDefinition(Node node)
 TrustNode TheoryFp::ppRewrite(TNode node)
 {
   Trace("fp-ppRewrite") << "TheoryFp::ppRewrite(): " << node << std::endl;
+  // first, see if we need to expand definitions
```
yoni206, line 456: ditto.

src/theory/bv/theory_bv.cpp
ajreynol: It impacts when expand definitions is applied during preprocessing.  It was previously the case that expandDefinitions (for BV) is applied at the beginning of preprocessing during expandDefinitions, now it is applied at the end (during theory-preprocess / ppRewrite).

yoni206 (commented):
I am really looking forward to this getting merged :) I think it will really enable better div/mod solving, and in general simpler pps.

Two comments:
1. For BV does it change the default behavior (see comments inline)? If so, perhaps it is worth to verify with @aniemetz and @mpreiner before merging?
2. I am not sure about adding `-q` to that many tests. Does it make sense to first add the PR that will fix that, and only then merge this PR without the `-q`s?

ajreynol:
For (1), to summarize: viewing the preprocessor as a black box, this does not impact BV (or other theories).  

For (2), Fixing the `-q` requires significant changes to the way `--check-models` works.  The order of this PR is intentional.  It is better to fix the preprocessor first, otherwise the design of `--check-models` is a moving target (since the preprocessor currently is not ideal).

yoni206:
(2) -- got it. SG.
(1) -- Would it make sense to check with other "preprocessing authors" (here or in tomorrow's meeting)?

ajreynol:
Yes, this PR indeed needs some discussion at the meeting.

src/theory/bv/theory_bv.cpp
aniemetz: So this only affects preprocessing passes then, right?

src/theory/datatypes/theory_datatypes.cpp
```diff
@@ -431,6 +431,8 @@ void TheoryDatatypes::preRegisterTerm(TNode n)
 {
   Trace("datatypes-prereg")
       << "TheoryDatatypes::preRegisterTerm() " << n << endl;
+  // external selectors should be preprocessed away by now
+  Assert (n.getKind()!=APPLY_SELECTOR);
```
mpreiner, line 435: Looks like this is not formatted.

src/theory/bv/theory_bv.cpp
ajreynol: It impacts preprocessing and `get-value`/`get-model`.

nafur (approved):
LGTM. Have you run benchmarks?

ajreynol:
I haven't run benchmarks yet.

yoni206 (approved):
In the meeting we decided it is ok to merge after SMT-LIB run with and without debug assertions. 
LGTM, but lets keep the `do not merge` flag until after runs are done.

nafur:
Current results (with 5s timeout) for non-incremental: [results.txt](https://github.com/CVC4/CVC4/files/5653546/results.txt)

ajreynol:
Based on @nafur 's results, there are no wrong answers or errors, no serious changes in performance (run with a 5 second timeout, the status changes are in QF_BV but everything is close to timeout).  This should be ready to merge.

yoni206:
Thanks, @nafur for all the testing for this!

PR #5545, opened 2020-11-30
Making a proof to check

yoni206:
A simple proof is now checked by both lfsc checkers (though with `trust`s).


src/proof/lfsc/lfsc_printer.cpp
```diff
@@ -371,7 +372,7 @@ void LfscPrinter::printInternal(std::ostream& out,
                                 const std::map<Node, uint32_t>& letMap)
 {
   // TODO: smt2 printer, dag thresh 0 print?
-  Node nc = Letify::convert(n, letMap, "@t");
+  Node nc = Letify::convert(n, letMap, "t");
```
ajreynol, line 375:
```suggestion
  Node nc = Letify::convert(n, letMap, "__t");
```

ajreynol (approved):

PR #5551, opened 2020-12-01
Use mkAnd where the number of children may be less than two.

### LINKED ISSUE #5550: Assertion error at src/./expr/node_builder.h:960 on UFC formula
cvc4 6489cf590b441
~~~~
(set-logic UFC)
(declare-sort a 0)
(declare-fun b () a)
(assert (not (fmf.card b 1)))
(check-sat)
~~~~

~~~~
Fatal failure within CVC4::expr::NodeValue* CVC4::NodeBuilder<N>::constructNV() [with unsigned int nchild_thresh = 10] at /home/peisen/test/tofuzz/CVC4/src/./expr/node_builder.h:960
Check failure

 getNumChildren() >= kind::metakind::getLowerBoundForKind(getKind())
Nodes with kind AND must have at least 2 children (the one under construction has 1)
Aborted

~~~~

nafur:
An `AND` was constructed from a vector that may only hold a single or no element.
This PR uses `mkAnd` instead.
Fixes #5550 .


ajreynol (changes requested):
Thanks for fixing this, can you add the regression too?

ajreynol (approved):

PR #5556, opened 2020-12-01
Add regression from #4927

### LINKED ISSUE #4927: (check-unsat-cores, check-models) Segmentation fault on QF_A formula
Hi,
For this formula, CVC4  throws out a segmentation fault:
```
[520] % cvc4 -q small.smt2
unsat
[521] % cvc4 -q --check-unsat-cores small.smt2
unsat
[522] % cvc4 -q --check-unsat-cores --check-models small.smt2
CVC4 suffered a segfault.
Offending address is 0x30
Looks like a NULL pointer was dereferenced.
Aborted
[523] % 
[523] % cat small.smt2
(declare-const a (Array Bool Bool))
(declare-const b (Array Bool Bool))
(declare-const c Bool)
(declare-const d Bool)
(declare-const e Bool)
(assert (= a (store b c d)))
(assert (distinct e d (ite e c d)))
(assert (= e c (select b d)))
(check-sat)
[524] %
```

OS: Ubuntu 18.04
Commit: c110363

nafur:
The error from #4927 has been fixed in the meantime, this PR adds the example as regression.
Closes #4927.


HanielB (approved):
LGTM

PR #5561, opened 2020-12-01
Use new let binding for cvc printer

ajreynol:
Also changes names slightly to avoid accidental uses of toStream, which can lead to infinite loops if the wrong version is used.


src/printer/cvc/cvc_printer.cpp
```diff
@@ -942,17 +924,17 @@ void CvcPrinter::toStream(std::ostream& out,
     // Quantifiers
     case kind::FORALL:
       out << "(FORALL";
-      toStream(out, n[0], depth, false);
+      toStreamNode(out, n[0], depth, true, lbind);
```
mpreiner, line 927: Just to make sure that this is not a copy&pase: Should this be really `true`? The original call was `false`.

src/printer/cvc/cvc_printer.cpp
```diff
@@ -942,17 +924,17 @@ void CvcPrinter::toStream(std::ostream& out,
...
       out << " : ";
-      toStream(out, n[1], depth, false);
+      toStreamNodeWithLetify(out, n[1], depth, true, lbind);
```
mpreiner, line 929: Ditto.

src/printer/cvc/cvc_printer.cpp
```diff
@@ -942,17 +924,17 @@ void CvcPrinter::toStream(std::ostream& out,
...
       out << ')';
       // TODO: user patterns?
       return;
     case kind::EXISTS:
       out << "(EXISTS";
-      toStream(out, n[0], depth, false);
+      toStreamNode(out, n[0], depth, true, lbind);
```
mpreiner, line 935: Ditto.

src/printer/cvc/cvc_printer.cpp
```diff
@@ -942,17 +924,17 @@ void CvcPrinter::toStream(std::ostream& out,
...
       out << " : ";
-      toStream(out, n[1], depth, false);
+      toStreamNodeWithLetify(out, n[1], depth, true, lbind);
```
mpreiner, line 937: Ditto.

src/printer/cvc/cvc_printer.cpp
ajreynol: Yep, there were inconsistencies which I intentionally fixed here.

mpreiner (approved):

PR #5562, opened 2020-12-01
(proof-new) Proofs for expand definitions

ajreynol:


src/smt/expand_definitions.cpp
```diff
@@ -43,6 +43,17 @@ Node ExpandDefs::expandDefinitions(
...
+  return trn.isNull() ? Node(n) : trn.getNode();
+}
+
+TrustNode ExpandDefs::expandDefinitions(
+    TNode n,
+    std::unordered_map<Node, Node, NodeHashFunction>& cache,
+    bool expandOnly,
+    TConvProofGenerator* tpg)
+{
+  Node orig = n;
```
nafur, line 56:
```suggestion
  const TNode orig = n;
```
orig is never written to, and `n` is a `TNode` already as well.

src/smt/expand_definitions.cpp
```diff
@@ -252,27 +317,53 @@ Node ExpandDefs::expandDefinitions(
...
+                                        d_smt.getUserContext(),
+                                        TConvPolicy::FIXPOINT,
+                                        TConvCachePolicy::NEVER,
+                                        "ExpandDefs::TConvProofGenerator",
+                                        nullptr,
+                                        true));
+  }
+}
+
+bool ExpandDefs::isProofEnabled() const { return d_tpg != nullptr; }
```
nafur, line 366: We never use this method. And given how this class is set up, we probably never will...

nafur (approved):
LGTM, just two suggestions.

PR #5563, opened 2020-12-01
Fix and improve evaluator

ajreynol:
This fixes a segfault in the evaluator for substitutions of the form x -> t where t is not constant. 

This consolidates 2 cases where we did not evaluate children (when we are variable or are an application term with an unevaluated child).  This was problematic previously since we would access children of `currNode` instead of `currNodeVal`.

This also makes our handling of APPLY_UF more consistent, so that it is added to the main switch statement.


4tXJ7f (approved):
LGTM

PR #5567, opened 2020-12-02
google test: api: Migrate op_black.

aniemetz:


test/unit/api/op_black.cpp
```diff
@@ -12,186 +12,169 @@
...
+  EXPECT_EQ(plus, d_solver.mkOp(PLUS));
+  ASSERT_THROW(d_solver.mkOp(BITVECTOR_EXTRACT), CVC4ApiException);
 }
 
-void OpBlack::testGetIndicesString()
+TEST_F(TestApiOpBlack, getIndicesString)
 {
   Op x;
-  TS_ASSERT_THROWS(x.getIndices<std::string>(), CVC4ApiException&);
+  ASSERT_THROW(x.getIndices<std::string>(), CVC4ApiException);
```
makaimann, line 52: One general comment. Do you want to continue the rest of the test even if this fails? In general, it's recommended to use the `EXPECT_*` macros unless it really doesn't make sense to continue the test.

makaimann (approved):
LGTM, thanks for porting this. I just left one minor comment to consider. I also have one more comment and question.

Comment: the `test_api.h` file has a copy-paste error in the header: https://github.com/CVC4/CVC4/blob/95409c284cf683aaf12b68242b503c95241ac01e/test/unit/test_api.h#L2

Question: I pulled this branch and tried to build it. How am I supposed to include GoogleTest? Should I put it somewhere that's globally accessible? I didn't see a setup script for it and it looks like you're not using the CMake infrastructure for pulling the repo when configuring.

test/unit/api/op_black.cpp
aniemetz: I did consider which ones to convert to `EXPECT` and decided to have all the `THROW` and `NO_THROW` as `ASSERT`, that is, not continuing. I guess we could discuss this rule of thumb, it's not set in stone. Past experience has shown me though that (at least for me) it doesn't make much of a difference if it continues or not -- I usually fix the first issue anyways and then continue. Note that `ASSERT` failures only terminate the current test (as in the current function that is executed), not the whole unit test module (as in, execution of the unit test binary, e.g., op_black).

aniemetz:
You have to install Google Test in your system. We decided to not pull it and build it via our cmake infrastructure, because it is kinda hacky. The problem with that is that our compile flags are propagated to building Google Test, which has `-Werror` enabled, so any warnings in the Google Test code with our compile flags result in not being able to build Google Test.

test/unit/api/op_black.cpp
makaimann: I think that rule of thumb makes sense! What about the `ASSERT_TRUE`s? I don't feel strongly about this by the way, just wanted to think about it up front because I ended up using `ASSERT` in smt-switch and now am slowly changing things to `EXPECT`. You make a good point about it not making much of a difference though, because you usually need to fix everything eventually anyway. I also tend to rebuild/rerun the test after every fix.

aniemetz: There I tried to use `EXPECT` if it's independent from what comes after. For example, if you just check a value or something. But if you assert the size of, let's say, a vector, and then access the elements assuming the size (e.g., we assert that size is 2 and then access with `foo[0]` `foo[1]`) I used assert, because if that assertion fails, follow-up statements might crash.

PR #5575, opened 2020-12-02
Remove Record object and convert to Node-level API

ajreynol:
Required for detangling NodeManager from the Expr layer.


src/expr/record.h
```diff
@@ -53,40 +53,7 @@ struct CVC4_PUBLIC RecordUpdateHashFunction {
...
-private:
-  FieldVector* d_fields;
-};/* class Record */
-
-struct CVC4_PUBLIC RecordHashFunction {
-  size_t operator()(const Record& r) const;
-};/* struct RecordHashFunction */
-
-std::ostream& operator<<(std::ostream& os, const Record& r) CVC4_PUBLIC;
+typedef std::vector<std::pair<std::string, TypeNode>> Record;
```
mpreiner, line 56:
```suggestion
using Record = std::vector<std::pair<std::string, TypeNode>>;
```

src/expr/node_manager.cpp
```diff
@@ -736,31 +736,33 @@ TypeNode NodeManager::TupleTypeCache::getTupleType( NodeManager * nm, std::vecto
...
-  if( index==rec.getNumFields() ){
+  if (index == rec.size())
+  {
     if( d_data.isNull() ){
-      const Record::FieldVector& fields = rec.getFields();
       std::stringstream sst;
       sst << "__cvc4_record";
-      for(Record::FieldVector::const_iterator i = fields.begin(); i != fields.end(); ++i) {
-        sst << "_" << (*i).first << "_" << (*i).second;
+      for (const std::pair<std::string, TypeNode>& i : rec)
```
mpreiner, line 744: Maybe use `auto` here?

src/expr/node_manager.cpp
```diff
@@ -736,31 +736,33 @@ TypeNode NodeManager::TupleTypeCache::getTupleType( NodeManager * nm, std::vecto
...
       }
       DType dt(sst.str());
       dt.setRecord();
       std::stringstream ssc;
       ssc << sst.str() << "_ctor";
       std::shared_ptr<DTypeConstructor> c =
           std::make_shared<DTypeConstructor>(ssc.str());
-      for(Record::FieldVector::const_iterator i = fields.begin(); i != fields.end(); ++i) {
-        c->addArg((*i).first, TypeNode::fromType((*i).second));
+      for (const std::pair<std::string, TypeNode>& i : rec)
```
mpreiner, line 754: Ditto.

src/expr/node_manager.cpp
ajreynol: I prefer having the types here, so its clear what we are iterating over.

mpreiner (approved):

Issue #4767, opened 2020-07-17
"Fatal failure" when overwriting symbols with `:global-declarations true`


andrewvaughanj:
## Issue

For the following file:

```
(set-option :global-declarations true)
(push 1)
(declare-fun x!0 () Bool)
(assert x!0)
(check-sat)
(get-model)
(pop 1)
(push 1)
(declare-fun x!0 () (_ BitVec 32))
(assert (= x!0 (_ bv0 32)))
(check-sat)
(get-model)
```

CVC4 falls over as follows:

```
[avj@tempvm build]$ ./bin/cvc4 -m -i ../../../blah.smt2
../../../blah.smt2:2.5: No set-logic command was given before this point.
../../../blah.smt2:2.5: CVC4 will make all theories available.
../../../blah.smt2:2.5: Consider setting a stricter logic for (likely) better performance.
../../../blah.smt2:2.5: To suppress this warning in the future use (set-logic ALL).
sat
(model
(define-fun x!0 () Bool true)
)
Fatal failure within void CVC4::context::CDHashMap<Key, Data, HashFcn>::insertAtContextLevelZero(const Key&, const Data&) [with Key = std::__cxx11::basic_string<char>; Data = CVC4::Expr; HashFcn = std::hash<std::__cxx11::basic_string<char> >] at ../src/./context/cdhashmap.h:388
Check failure

 d_map.find(k) == d_map.end()

Aborted (core dumped)
```

`git rev-parse --short HEAD` == e8df6f67c

andrewvaughanj:
Actually, it seems that global declarations also interferes with type-based overloading:

```
(set-option :global-declarations true)
(declare-fun x!0 () Bool)
(declare-fun x!0 () (_ BitVec 32))
```

```
[avj@tempvm build_debug]$ ./bin/cvc4 ../blah.smt2
../blah.smt2:2.12: No set-logic command was given before this point.
../blah.smt2:2.12: CVC4 will make all theories available.
../blah.smt2:2.12: Consider setting a stricter logic for (likely) better performance.
../blah.smt2:2.12: To suppress this warning in the future use (set-logic ALL).
Fatal failure within void CVC4::context::CDHashMap<Key, Data, HashFcn>::insertAtContextLevelZero(const Key&, const Data&) [with Key = std::__cxx11::basic_string<char>; Data = CVC4::Expr; HashFcn = std::hash<std::__cxx11::basic_string<char> >] at ../src/./context/cdhashmap.h:388
[...]

PR #5577, opened 2020-12-02
Refactor handling of global declarations

ajreynol:
This refactors how global declarations are handled in the parser.  In particular, we do *not* push/pop user contexts in the symbol table and manager when global declarations are true, which is an equivalent behavior to declaring symbols globally.

This further refactors to not use ExprManager flags, thus breaking most of the dependencies on the old API.

This is work towards fixing global declarations in the smt2 parser. The parser still does not behave correctly for overloaded symbols + global declarations (e.g. see #4767), which require further refactoring.

This is also work towards migrating the parser not to depend on the old API.  There are a few miscellaneous things to change after this PR, but we are very close to breaking this dependency now.


nafur (approved):
LGTM.

PR #5578, opened 2020-12-02
Fix RoundingMode mapping in API.

### LINKED ISSUE #5524: Solution soundness issue QF_FP formula
~~~~
(set-logic QF_FP)
(declare-fun fpv5 () Float32)
(assert (fp.eq (fp.mul RTP fpv5 fpv5) ((_ to_fp 8 24) RTN 0.04)))
(check-sat)
~~~~
cvc4 25762974521 yields `sat`, but z3 gives `unsat`.
cvc4 gives the following model.
However, after feeding the model to z3, it still yields `unsat`.
~~~~
(define-fun fpv5 () (_ FloatingPoint 8 24) (fp #b0 #b01111100 #b10011001100110011001101))
~~~~

----
Update: MathSAT5 also yields `unsat`

aniemetz:
Fixes #5524.


mpreiner (approved):

ajreynol:
I believe you need to add the regression to the list?

martin-cs (approved):
Thank you for finding and fixing this.  It looked like a rounding mode issue (I probably should have said so shouldn't I?) and I had been absolutely wracking my brains to figure out how we missed this in our verification.

aniemetz:
I introduced it, and after bisecting it was easy to find :-) So no error on your part.

PR #5581, opened 2020-12-03
Models as

yoni206:
This PR relates to #4987 .
Our plan is to:

1. delete the model keyword (done in #5415 )
2. avoid printing extra declarations by default (done in #5432 )
3. wrap UF values with as expressions.

This PR does step 3, fixes a regression accordingly, and adds the formula from #4987  and a variant of it to the regressions.


src/options/smt_options.toml
```diff
@@ -429,7 +429,11 @@ header = "options/smt_options.h"
   help = "print uninterpreted elements declare-fun, and also include a declare-sort for the sort"
 [[option.mode.DeclFun]]
   name = "decl-fun"
-  help = "(default) print uninterpreted elements declare-fun, but don't include a declare-sort for the sort"
+  help = "print uninterpreted elements declare-fun, but don't include a declare-sort for the sort"
+[[option.mode.As]]
```
ajreynol, line 433: I think this option should just be called "none", e.g. do not print uninterpreted sorts in models.

ajreynol (dismissed):

src/options/smt_options.toml
```diff
@@ -429,7 +429,11 @@ header = "options/smt_options.h"
   help = "print uninterpreted elements declare-fun, and also include a declare-sort for the sort"
 [[option.mode.DeclFun]]
   name = "decl-fun"
-  help = "(default) print uninterpreted elements declare-fun, but don't include a declare-sort for the sort"
+  help = "print uninterpreted elements declare-fun, but don't include a declare-sort for the sort"
+[[option.mode.None]]
+  name = "none"
+  help = "(default) print uninterpreted elements in `as` terms."
```
ajreynol, line 435:
```suggestion
  help = "(default) do not print uninterpreted elements in models."
```

ajreynol (changes requested):

ajreynol (approved):

PR #5583, opened 2020-12-03
Change generated options to be thread_local.

evmaus:
Per discussion on [this issue](https://github.com/CVC4/cvc4-projects/issues/11), having these options be global causes issues if you have multiple copies of CVC4 running in a single process, even if they are not running on the same thread. This can be detected via TSAN.

This provides a temporary fix which makes all of these options thread_local, so that this scenario is supported.  Also per discussion on [this issue](https://github.com/CVC4/cvc4-projects/issues/11), this is intended as a short term fix until a better options model is designed & implemented.


mpreiner (approved):
LGTM, Thanks a lot for the fixes!

PR #5589, opened 2020-12-03
Make term indices in the strings base solver aware of types

### LINKED ISSUE #5542: Fatal failure at src/theory/strings/base_solver.cpp:169
```
[504] % cvc4 --strings-exp -q small.smt2
Fatal failure within void CVC4::theory::strings::BaseSolver::checkInit() at src/theory/strings/base_solver.cpp:169
Check failure
 count[1] < n.getNumChildren()
Aborted
[505] % 
[505] % cat small.smt2
(declare-fun a () String)
(declare-fun b () String)
(declare-fun c () (Seq Int))
(assert (distinct b (str.++ a a)))
(assert (distinct (seq.unit 0) (seq.extract c 1 1)))
(assert (= (seq.len c) 1))
(check-sat)
[506] 
```
OS: Ubuntu 18.04
Commit: 9f372f0

ajreynol:
This is required for handling inputs that combine strings and sequences.

Fixes #5542.


src/theory/strings/base_solver.cpp
```diff
@@ -55,6 +57,7 @@ void BaseSolver::checkInit()
     if (!tn.isRegExp())
     {
       Node emps;
+      std::map<Kind, TermIndex>& tti = d_termIndex[tn];
```
yoni206, line 60: Can you add a comment that says what is `tti`?

yoni206 (approved):
LGTM, one suggestion inline.

PR #5595, opened 2020-12-03
Use NodeDfsIterable for BVToInt::makeBinary

alex-ozdemir:
Replaces the manual dag traversal in `BVToInt::makeBinary`with `NodeDfsIterable`.

This is a subset of the changes in #4176, updated to apply to master.

I recommend you view the diff with whitespace changes ignored, by visiting this link: [https://github.com/CVC4/CVC4/pull/5595/files?w=1](https://github.com/CVC4/CVC4/pull/5595/files?w=1).


yoni206 (approved):
Great! Thanks!

BTW:
@makaimann 's smt-switch-walker has a flag that allows to do stuff both on the way down and on the way up in the same tranversal, rather than deciding about the order in advance. This might be useful for your iterator in order to be able to integrate it with more CVC4 traversals.

https://github.com/makaimann/smt-switch/blob/de8870e1c83c9830c1a4e369764a684743adc50e/include/identity_walker.h#L101

PR #5600, opened 2020-12-04
Fix bug in hasBoundVar

ajreynol:
Led to issues while debugging another issue related to free variables in assumptions.


nafur (approved):
LGTM.

PR #5601, opened 2020-12-04
Fix bugs in getFreeVariables

ajreynol:
Did not consider scopes properly, which would fail to say that formulas with escaped variables (both bound and unbound) in formula had free variables.


nafur (approved):
LGTM.

PR #5603, opened 2020-12-04
Add support for BV proofs with the simple bitblasting solver.

mpreiner:


src/theory/bv/bv_solver_simple.h
```diff
@@ -71,6 +74,11 @@ class BVSolverSimple : public BVSolver
 
   /** Bit-blaster used to bit-blast atoms/terms. */
   std::unique_ptr<BBSimple> d_bitblaster;
+
+  /** Proof generator that manages proofs for lemmas generated by this class. */
+  std::unique_ptr<EagerProofGenerator> d_epg;
+
```
ajreynol, line 80:
```suggestion
  /** The proof checker for bitvectors */
```

src/theory/bv/proof_checker.cpp
```diff
@@ -0,0 +1,54 @@
...
+ ** \brief Implementation of bit-vectors proof checker
+ **/
+
+#include "theory/bv/proof_checker.h"
+
+namespace CVC4 {
+namespace theory {
+namespace bv {
+
+BVProofRuleChecker::BVProofRuleChecker() : d_bb(new BBSimple(nullptr)) {}
```
ajreynol, line 21: I'm wondering if it would be better to make d_bb a local variable wheren BV_BITBLAST is.  This would make the proof checker stateless (and hence more trustworthy conceptually).
Is there a reason for having d_bb be a member? E.g. do you need its cache?

ajreynol (changes requested):

src/theory/bv/proof_checker.cpp
mpreiner: The cache is not mandatory and it also works if we make the bitblaster local in the BV_BITBLAST rule. If we ever run into performance issues because of the cache, we can always fix it.

ajreynol (approved):

PR #5612, opened 2020-12-07
[proof-new] Adding MACRO_RESOLUTION rule and updating proof checker

HanielB:
Previously CHAIN_RESOLUTION's definition and checker were not properly capturing its intended behavior as merely an n-ary RESOLUTION rule (i.e., no factoring nor reordering). A new rule, MACRO_RESOLUTION, now captures this behavior: it combines CHAIN_RESOLUTION, FACTORING, and REORDERING. 

This commit also adds a proof checker for the new rule and updates the proof checker of CHAIN_RESOLUTION.


src/theory/booleans/proof_checker.cpp
```diff
@@ -190,21 +191,145 @@ Node BoolProofRuleChecker::checkInternal(PfRule id,
...
+      }
+    }
+    for (unsigned i = 0, childrenSize = children.size(); i < childrenSize; ++i)
+    {
+      // literal to be removed from rhs clause. They will be negated
+      Node rhsElim = Node::null();
+      if (Trace.isOn("bool-pfcheck"))
+      {
+        Trace("bool-pfcheck") << i << ": current lhsElim:\n";
+        for (const std::pair<const Node&, unsigned>& pair : lhsElim)
```
nafur, line 233: ditto.
```suggestion
        for (const auto& pair : lhsElim)
```

src/theory/booleans/proof_checker.cpp
```diff
@@ -190,21 +191,145 @@ Node BoolProofRuleChecker::checkInternal(PfRule id,
...
+        {
+          Trace("bool-pfcheck")
+              << "\t- " << pair.first << " {" << pair.second << "}\n";
+        }
+      }
+      if (i > 0)
+      {
+        unsigned index = 2 * (i - 1);
```
nafur, line 241:
```suggestion
        std::size_t index = 2 * (i - 1);
```

src/theory/booleans/proof_checker.cpp
```diff
@@ -190,21 +191,145 @@ Node BoolProofRuleChecker::checkInternal(PfRule id,
...
+      {
+        lits.insert(lits.end(), children[i].begin(), children[i].end());
+      }
+      else
+      {
+        lits.push_back(children[i]);
+      }
+      Trace("bool-pfcheck") << i << ": clause lits: " << lits << "\n";
+      std::vector<Node> added;
+      for (unsigned j = 0, size = lits.size(); j < size; ++j)
```
nafur, line 264:
```suggestion
      for (std::size_t j = 0, size = lits.size(); j < size; ++j)
```

src/theory/booleans/proof_checker.cpp
```diff
@@ -190,21 +191,145 @@ Node BoolProofRuleChecker::checkInternal(PfRule id,
...
+  if (id == PfRule::MACRO_RESOLUTION)
+  {
+    Assert(children.size() > 1);
+    Assert(args.size() == 2 * (children.size() - 1) + 1);
+    Trace("bool-pfcheck") << "macro_res: " << args[0] << "\n" << push;
+    NodeManager* nm = NodeManager::currentNM();
+    Node trueNode = nm->mkConst(true);
+    Node falseNode = nm->mkConst(false);
     std::vector<Node> clauseNodes;
     for (unsigned i = 0, childrenSize = children.size(); i < childrenSize; ++i)
```
nafur, line 304:
```suggestion
    for (std::size_t i = 0, childrenSize = children.size(); i < childrenSize; ++i)
```

src/theory/booleans/proof_checker.cpp
```diff
@@ -190,21 +191,145 @@ Node BoolProofRuleChecker::checkInternal(PfRule id,
...
     {
       std::unordered_set<Node, NodeHashFunction> elim;
       // literals to be removed from "first" clause
       if (i < childrenSize - 1)
       {
-        elim.insert(args.begin() + i, args.end());
+        for (unsigned j = (2 * i) + 1, argsSize = args.size(); j < argsSize;
```
nafur, line 310:
```suggestion
        for (std::size_t j = (2 * i) + 1, argsSize = args.size(); j < argsSize;
```

src/theory/booleans/proof_checker.cpp
```diff
@@ -190,21 +191,145 @@ Node BoolProofRuleChecker::checkInternal(PfRule id,
...
+            Assert(args[j] == falseNode);
+            elim.insert(args[j + 1].notNode());
+          }
+        }
       }
       // literal to be removed from "second" clause. They will be negated
       if (i > 0)
       {
-        elim.insert(args[i - 1].negate());
+        unsigned index = 2 * (i - 1) + 1;
```
nafur, line 329:
```suggestion
        std::size_t index = 2 * (i - 1) + 1;
```

src/theory/booleans/proof_checker.cpp
```diff
@@ -214,7 +339,7 @@ Node BoolProofRuleChecker::checkInternal(PfRule id,
       // non-unit clauses will not occur themselves in their elimination sets.
       // If they do then they must be unit.
       std::vector<Node> lits;
-      if (children[i].getKind() == kind::OR && elim.count(children[i]) == 0)
+      if (children[i].getKind() == kind::OR && !elim.count(children[i]))
```
nafur, line 342: This introduces an implicit cast from `std::size_t` to bool... I'd rather stay with `== 0`.
```suggestion
      if (children[i].getKind() == kind::OR && elim.count(children[i]) == 0)
```

src/theory/booleans/proof_checker.cpp
```diff
@@ -226,20 +351,50 @@ Node BoolProofRuleChecker::checkInternal(PfRule id,
       std::vector<Node> added;
       for (unsigned j = 0, size = lits.size(); j < size; ++j)
       {
-        if (elim.count(lits[j]) == 0)
+        // only add if literal does not occur in elimination set
+        if (!elim.count(lits[j]))
```
nafur, line 355: ditto.
```suggestion
        if (elim.count(lits[j]) == 0)
```

src/theory/booleans/proof_checker.cpp
```diff
@@ -226,20 +351,50 @@ Node BoolProofRuleChecker::checkInternal(PfRule id,
       std::vector<Node> added;
       for (unsigned j = 0, size = lits.size(); j < size; ++j)
```
nafur, line 352:
```suggestion
      for (std::size_t j = 0, size = lits.size(); j < size; ++j)
```

nafur (changes requested):
Nothing serious, but a bunch of small improvements.

HanielB:
Thanks for the suggestions! I applied them all.

src/theory/booleans/proof_checker.cpp
ajreynol: @HanielB Still need to do this one

ajreynol (changes requested):
One comment, else LGTM

nafur (approved):
Always getting confused by the same code snippet, thus suggesting a new comment.
Otherwise LGTM.

src/theory/booleans/proof_checker.cpp
```diff
@@ -190,21 +191,147 @@ Node BoolProofRuleChecker::checkInternal(PfRule id,
...
+        rhsElim = args[index] == trueNode ? args[index + 1].notNode()
+                                          : args[index + 1];
+        Trace("bool-pfcheck") << i << ": rhs elim: " << rhsElim << "\n";
+      }
+      // only add to conclusion nodes that are not in elimination set. First get
+      // the nodes.
+      //
+      // Since unit clauses can also be OR nodes, we rely on the invariant that
+      // non-unit clauses will not occur themselves in their elimination sets.
+      // If they do then they must be unit.
```
nafur, line 252: First: I always feel "unit clause" is reserved to SAT solvers, and we should rather use "singleton clause"...

Then I'm always confused by what the code below does. Suggestion for a different (possibly better) comment:

```suggestion
      // Since a Node can not hold an OR with a single child we need to take of such singleton clauses.
      // If the child is not an OR, it is a singleton clause and we only take the child itself.
      // Otherwise, it can only be a singleton clause if the child itself is used as a resolution literal, that is if the child is in lhsElim or rhsElim (and then the negation of the child is in lhsElim).
```

src/expr/proof_rule.h
```diff
@@ -273,15 +273,18 @@ enum class PfRule : uint32_t
   //   to resolution but rather to a weakening of the clause that did not have a
   //   literal eliminated.
   RESOLUTION,
-  // ======== Chain Resolution
-  // Children: (P1:(or F_{1,1} ... F_{1,n1}), ..., Pm:(or F_{m,1} ... F_{m,nm}))
-  // Arguments: (L_1, ..., L_{m-1})
+  // ======== N-ary Resolution
+  // Children: (P1:C_1, ..., Pm:C_n)
+  // Arguments: (id_1, L_1, ..., id_{n-1}, L_{n-1})
```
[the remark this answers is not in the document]
HanielB: Bad name choice done in a hurry. I'll change to `polarity`.

src/theory/booleans/proof_checker.cpp
HanielB: I'll change accordingly, thanks!

ajreynol (approved):

PR #5614, opened 2020-12-07
Fix and reenable fact vs lemma optimization in datatypes

ajreynol:
This corrects an issue where terms internal to datatypes were not getting properly registered e.g. as part of the indices that determine the care graph, due to a context-independent cache being used (when a SAT-context-dependent one was required).

This reenables the fact vs lemma optimization in datatypes, as it is conjectured to be correct.


yoni206 (approved):

yoni206:
LGTM.
How about adding this test (requiring symfpu)?
```
(set-logic ALL)
(declare-datatypes ((V 0) (A 0)) (
  ((I (i Int)) (vec (v A)))
  ((arr (l Int)))))
(declare-fun E (V V) Bool)
(declare-fun eee (A) Bool)
(declare-fun a () V)
(declare-fun aa () A)
(assert (forall ((?y1 A)) (eee ?y1)))
(assert (E a a))
(assert (not (E (I 0) (I 0))))
(assert (= (l aa) (i a)))
(check-sat)
```

yoni206 (approved):
10x 4 adding the regression

PR #5616, opened 2020-12-07
Proper implementation of expand definitions for sequences

ajreynol:
Expand definitions for sequences was wrong in two ways: (1) we replaced str.from_code with a witness term. This led to it being unevaluatable in models. (2) we did not handle seq.nth, meaning its model value was unevaluatable if it was out of bounds. Now it evaluates the value of the uninterpreted function we replace with.

This corrects both issues and adds a regression to demonstrate both kinds of terms evaluate correctly.

To do this, I added a helper function to skolem cache as well as a new (internal-only) kind SEQ_NTH_TOTAL.  Notice applications of this kind should only be used for model evaluation.

Notice this fixes several check-model warnings in the regressions.  It still does not fix others since other things must be corrected for model evaluation (e.g. expandDefinitions must be applied on theory assertions for --debug-check-models).  This will be done in later PRs.


yoni206 (commented):
Looks good. Is there a reasonable place to document what is the meaning of `SEQ_NTH_TOTAL_OOB`?

ajreynol:
I prefer having the documentation in the same place it is used, which is in the cpp of the rewriter.

yoni206 (approved):

PR #5617, opened 2020-12-07
(proof-new) Add proof checker for transcendental solver

nafur:
This PR adds a separate proof checker for the transcendental solver and moves the proof for pi bounds over.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -413,6 +413,8 @@ libcvc4_add_sources(
   theory/arith/nl/strategy.h
   theory/arith/nl/transcendental/exponential_solver.cpp
   theory/arith/nl/transcendental/exponential_solver.h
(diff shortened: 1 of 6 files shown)


src/theory/arith/nl/ext/proof_checker.h
```diff
@@ -29,15 +29,14 @@ namespace nl {
...
- * considers them to be trusted rules.
+ * This proof checker takes care of all proofs for lemmas from the ext
+ * subsolver.
  */
 class ExtProofRuleChecker : public ProofRuleChecker
 {
  public:
-  ExtProofRuleChecker() {}
-  ~ExtProofRuleChecker() {}
+  ExtProofRuleChecker() = default;
```
ajreynol, line 38: Nit: out of curiosity, what is the reason for this instead of {}?

ajreynol (approved):

PR #5618, opened 2020-12-07
(proof-new) Add new CONG_LEMMA proof rule.

nafur:
We'll need a rule to generate a lemma for a congruence. While we have the CONG rule, we need SCOPE to turn it into a lemma, therefore @ajreynol and I agreed at some point that a CONG_LEMMA rule would be nice to have to avoid using SCOPE.

### DIFF
--- a/src/expr/proof_rule.cpp
+++ b/src/expr/proof_rule.cpp
@@ -107,6 +107,7 @@ const char* toString(PfRule id)
     case PfRule::SYMM: return "SYMM";
     case PfRule::TRANS: return "TRANS";
     case PfRule::CONG: return "CONG";
+    case PfRule::CONG_LEMMA: return "CONG_LEMMA";
     case PfRule::TRUE_INTRO: return "TRUE_INTRO";
     case PfRule::TRUE_ELIM: return "TRUE_ELIM";
     case PfRule::FALSE_INTRO: return "FALSE_INTRO";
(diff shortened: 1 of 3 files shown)


ajreynol:
I'm undecided whether this is a good idea.  This adds more burden on external proof conversion.  Either that or we could eliminate this at post-processing.

HanielB:
I see it as two styles of calculi: one that does stuff via subproofs and another that does it via addition of valid clauses. Internally our preference is for the former, so I don't think this rule fits very well with that. External calculi (such as the veriT or currently the Lean one) that are more "valid-clause oriented" can be handled easily via conversions. To have this rule only for congruence would be odd as well.

nafur:
I find this a very nice way of thinking about it. I'm fine with only/mostly having proofs via subproofs, however this will use `SCOPE` a lot, as we mostly use lemmas in the nonlinear solving part.
Note that for many lemmas we need a new proof rule nevertheless. We can also formulate these as subproofs (and always use an additional `SCOPE`)... is this what we want?

HanielB:
I also like it and we do have several such "shortcut" rules. I just worry a bit about maybe having too much "duplication" and how handling different styles could increase the complexity of postprocessing, as @ajreynol said.

nafur:
We have used a way that avoids CONG_LEMMA.

PR #5619, opened 2020-12-08
Update help message for `--bv-div-zero-const` option

yoni206:
The current help message for `--bv-div-zero-const` only mentions `bvudiv` (in which the result is `-1`) and not `bvurem` (in which the result is the first argument).
I think this is a bit misleading, because in practice, this option controls also the behavior of `bvurem`:
https://github.com/CVC4/CVC4/blob/0309ef4aa7462c6fa2a65c1ef408dc9063bb1f21/src/theory/bv/theory_bv.cpp#L134

This PR is an attempt to provide a more accurate help message.


ajreynol:
This option should be deleted, but this looks fine.

yoni206:
10x.
I saw a long discussion about this on slack that revolved around smt-lib 2.5 support.
Looks like we still support it:
https://github.com/CVC4/CVC4/blob/0ab8a3a7af5b80aa7bcaa028276cdc396aa7a4cb/src/options/language.h#L50

So at least for now I think the message should be improved.

mpreiner (approved):

PR #5624, opened 2020-12-08
Properly implement datatype selector triggers

ajreynol:
This ensures that datatype selectors are properly handled as triggers in E-matching.

This is challenging since selectors in quantified formulas are of kind APPLY_SELECTOR but are theory-preprocessed to APPLY_SELECTOR_TOTAL/APPLY_UF.  Hence, we must match on 2 possible operators, and ones that do not match the operator of the trigger.  This adds a custom candidate generator for this case.

It also removes a deprecated option that is no longer used (in part due to our use of shared selectors).

This is in preparation for further work on optimizing cvc4 on benchmarks from Facebook.

Note that there is not a convienient way to call `expandDefinitions` currently (which is required to get the proper operators to match). This PR calls smt::getCurrentSmtEngine() to do this, although we should find a better solution soon, e.g. adding expandDefinitions to the rewriter.

FYI @barrettcw


src/smt/smt_engine.h
```diff
@@ -510,12 +510,11 @@ class CVC4_PUBLIC SmtEngine
...
 
   /**
-   * Expand the definitions in a term or formula.  No other
-   * simplification or normalization is done.
+   * Expand the definitions in a term or formula.
    *
    * @throw TypeCheckingException, LogicException, UnsafeInterruptException
    */
-  Node expandDefinitions(const Node& e);
+  Node expandDefinitions(const Node& e, bool expandOnly = true);
```
HanielB, line 517: document the optional argument

src/theory/quantifiers/ematching/candidate_generator.h
```diff
@@ -208,6 +212,26 @@ class CandidateGeneratorConsExpand : public CandidateGeneratorQE
...
+  /** reset */
+  void reset(Node eqc) override;
+  /** get next candidate */
+  Node getNextCandidate() override;
+
+ protected:
+  /** the selector operator */
+  Node d_selOp;
+  /** the UF operator */
+  Node d_ufOp;
```
HanielB, line 232: Seems a bit wasteful to keep these duplicates as nodes. Do you think it's worth it to use TNodes for this?

src/theory/quantifiers/ematching/candidate_generator.h
```diff
@@ -208,6 +212,26 @@ class CandidateGeneratorConsExpand : public CandidateGeneratorQE
...
+ * Candidate generator for selector applications, which considers both
+ * internal terms corresponding to correctly and incorrectly applied selectors.
+ */
+class CandidateGeneratorSelector : public CandidateGeneratorQE
+{
+ public:
+  CandidateGeneratorSelector(QuantifiersEngine* qe, Node mpat);
+  /** reset */
+  void reset(Node eqc) override;
+  /** get next candidate */
```
HanielB, line 225: It'd be good to describe how this method works in terms of correctly vs non-correctly applied selectors and so on.

HanielB (approved):
LGTM, just a few comments

src/theory/quantifiers/ematching/candidate_generator.h
ajreynol: This class should only be constructed a handful of times per benchmark.  It's not important to do TNode here.

PR #5625, opened 2020-12-08
(proof-new) Add proofs for secant lemmas for exponential functions

nafur:
This PR adds proofs (via new proof rules) for secant plane lemmas for exponential functions from the transcendental solver.


src/theory/arith/nl/transcendental/proof_checker.cpp
```diff
@@ -26,9 +26,83 @@ namespace arith {
 namespace nl {
 namespace transcendental {
 
+namespace {
+
+/**
+ * Evaluates the first d terms of the Maclaurin series for the exponential
+ * function at t. This is a lower bound on the exponential function.
+ */
+inline Node evalMaclaurinExp(std::uint64_t d, TNode t)
```
ajreynol, line 35: Can we make this code shared between the solver and the proof checker? (Similar questions for the other utility methods).  This is a principle I've been trying to stick to, that the proof checker and the solver should use common utilities.

ajreynol (approved):
LGTM, thanks a lot for refactoring this.

PR #5626, opened 2020-12-08
Make decision engine independent of AssertionsPipeline

ajreynol:
This PR makes decision engine independent of AssertionsPipeline, which consequently allows some of the key PropEngine interfaces to be consolidated.  It also modifies PropEngine to take TrustNode for assertLemma, which is the first step for making PropEngine manage proofs from TheoryEngine.

This is in preparation for modifying the interplay between PropEngine, TheoryEngine, TheoryPreprocessor, and new proposed SAT relevancy heuristic.

There are no intended behavior changes in this PR. 

Marking "major" since this impacts several current directions (including proof-new integration, lazy theory preprocessing, SAT relevancy).


src/prop/prop_engine.cpp
```diff
@@ -123,25 +123,71 @@ PropEngine::~PropEngine() {
...
+  }
+
+  // assert to decision engine
+  if (!removable)
+  {
+    // also add to the decision engine, where notice we don't need proofs
+    std::vector<Node> assertions;
+    assertions.push_back(lem.getProven());
+    std::vector<Node> ppLemmasF;
+    for (const theory::TrustNode& tnl : ppLemmas)
```
HanielB, line 185: Should this be a reference?

src/prop/prop_engine.h
```diff
@@ -99,19 +106,28 @@ class PropEngine
...
+   * @param trn the trust node storing the formula to assert
    * @param removable whether this lemma can be quietly removed based
    * on an activity heuristic (or not)
    */
-  void assertLemma(TNode node, bool negated, bool removable);
+  void assertLemma(theory::TrustNode trn, bool removable);
 
   /**
-   * Pass a list of assertions from an AssertionPipeline to the decision engine.
+   * Assert lemma trn with preprocessing lemmas newLemmas which corresponding
```
HanielB, line 116:
```suggestion
   * Assert lemma trn with preprocessing lemmas newLemmas which correspond
```

src/prop/prop_engine.h
```diff
@@ -99,19 +106,28 @@ class PropEngine
...
+   * to lemmas for skolems in newSkolems.
+   *
+   * @param trn the trust node storing the formula to assert
+   * @param ppLemmas the lemmas from preprocessing and term formula removal on
+   * the proven node of trn
+   * @param ppSkolem the skolem that each lemma in ppLemma constrains. It should
+   * be the case that ppLemmas.size()==ppSkolems.size().
+   * @param removable whether this lemma can be quietly removed based
+   * on an activity heuristic (or not)
```
HanielB, line 125: this "(or not)" is redundant, I think?

HanielB (approved):
LGTM, just some minor comments

src/prop/prop_engine.h
ajreynol: Sure :+1:

src/prop/prop_engine.cpp
ajreynol: What do you mean? It is a reference already?

PR #5627, opened 2020-12-08
Do not check FP at last call if not necessary

ajreynol:
This makes the theory of floating points not request a last call effort check, if that call is known to do nothing.

This should make CVC4 (more) agnostic to whether --symfpu is enabled during configuration.


nafur (approved):
Assuming that `d_abstractionMap` is a good indicator for whether TheoryFp needs to do anything: LGTM.

aniemetz (approved):
LGTM

martin-cs:
@nafur : It's fine :-)

PR #5637, opened 2020-12-09
Fixed a bunch of clang warnings.

nafur:
This PR fixes a bunch of compiler warnings, mostly about incorrect types used in range-based for loops.


src/theory/datatypes/sygus_extension.cpp
```diff
@@ -1196,7 +1196,7 @@ void SygusExtension::registerSymBreakLemma( TypeNode tn, Node lem, unsigned sz,
         sca.d_search_terms[tn].find(d);
     if (itt != sca.d_search_terms[tn].end())
     {
-      for (const TNode& t : itt->second)
+      for (const Node& t : itt->second)
```
aniemetz, line 1199: Why not `(TNode t : itt->second)` here?

src/theory/datatypes/sygus_extension.cpp
```diff
@@ -1469,7 +1469,7 @@ void SygusExtension::incrementCurrentSearchSize( Node m, std::vector< Node >& le
           int new_depth = ((int)itsz->second->d_curr_search_size) - ((int)sz);
           std::map< unsigned, std::vector< Node > >::iterator itt = itc->second.d_search_terms[tn].find( new_depth );
           if( itt!=itc->second.d_search_terms[tn].end() ){
-            for (const TNode& t : itt->second)
+            for (const Node& t : itt->second)
```
aniemetz, line 1472: Ditto

src/theory/quantifiers/term_database.cpp
```diff
@@ -276,14 +276,14 @@ void TermDb::computeUfEqcTerms( TNode f ) {
     ops.insert(ops.end(), d_ho_op_slaves[f].begin(), d_ho_op_slaves[f].end());
   }
   eq::EqualityEngine* ee = d_quantEngine->getMasterEqualityEngine();
-  for (const Node& ff : ops)
+  for (TNode ff : ops)
   {
-    for (const TNode& n : d_op_map[ff])
+    for (const Node& n : d_op_map[ff])
```
aniemetz, line 281: Ditto

src/theory/uf/symmetry_breaker.cpp
```diff
@@ -594,7 +594,7 @@ bool SymmetryBreaker::invariantByPermutations(const Permutation& p) {
   subs.push_back(p1);
   repls.push_back(p1);
   repls.push_back(p0);
-  for (const Node nn : d_phi)
+  for (const Node& nn : d_phi)
```
aniemetz, line 597: Ditto

src/theory/quantifiers/term_database.cpp
HanielB: Maybe it can't be a TNode because of the usage below? Regardless, should the reference be dropped?

src/theory/sep/theory_sep.cpp
```diff
@@ -755,7 +755,7 @@ void TheorySep::postCheck(Effort level)
 
       // get model values
       std::map<int, Node> mvals;
-      for (const std::pair<int, Node>& sub_element : d_label_map[satom][slbl])
+      for (const std::pair<const int, Node>& sub_element : d_label_map[satom][slbl])
```
HanielB, line 758: does this matter?

src/theory/strings/sequences_rewriter.cpp
```diff
@@ -1527,7 +1526,7 @@ Node SequencesRewriter::rewriteSeqNth(Node node)
     if (pos < len)
     {
       std::vector<Node> elements = s.getConst<Sequence>().getVec();
-      ret = elements[pos];
+      const Node& ret = elements[pos];
```
HanielB, line 1529: should the reference be dropped?

src/theory/uf/symmetry_breaker.cpp
HanielB: ditto

src/theory/quantifiers/sygus/enum_stream_substitution.cpp
```diff
@@ -88,7 +88,7 @@ void EnumStreamPermutation::reset(Node value)
       d_var_classes[ti.getSubclassForVar(var)].push_back(var);
     }
   }
-  for (const std::pair<unsigned, std::vector<Node>>& p : d_var_classes)
+  for (const std::pair<const unsigned, std::vector<Node>>& p : d_var_classes)
```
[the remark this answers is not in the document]
ajreynol: Yes this makes a big difference for performance.

src/smt/proof_manager.cpp
```diff
@@ -161,7 +161,7 @@ void PfManager::getAssertions(Assertions& as,
     assertions.push_back(*i);
   }
   NodeManager* nm = NodeManager::currentNM();
-  for (const std::pair<Node, smt::DefinedFunction>& dfn : df)
+  for (const std::pair<const Node, const smt::DefinedFunction>& dfn : df)
```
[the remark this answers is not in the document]
nafur: This is purely about the type (for `dfn`) matching what is in `df`. `df` is a `context::CDHashMap<Node, smt::DefinedFunction, NodeHashFunction>` whose `value_type` ends up to be `std::pair<const Key, const Data>`. If you use anything else, the for loop needs to copy the whole thing.
(The mere fact that we discuss this makes my point for using `auto` more often ;-) )

src/theory/strings/sequences_rewriter.cpp
nafur: `const Node` would entail copying from `elements[pos]` to `ret`. This avoids the copy.

src/theory/datatypes/sygus_extension.cpp
nafur: So this would be the difference of copying a `Node` to a `TNode` versus holding a reference to a `Node`. I'm not sure it makes a difference, and I went with having `const T&` for a container that holds `T` as a generic pattern.

src/smt/proof_manager.cpp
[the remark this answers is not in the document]
HanielB: Would `auto` generate the type with the `const`? At some point I was trying to have auto generate a `const_iterator` rather than an `iterator` as a result of some `find` and apparently it was not possible force this. Is something like this not an issue here?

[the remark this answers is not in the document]
ajreynol: The leftmost "const" is optional and depends on if the iterator can modify the data.  

The "const" before "const Node" should always be there since that is the value_type as Gereon mentions.

src/theory/datatypes/sygus_extension.cpp
aniemetz: Right, and I actually agree that with that rule of thumb. I was just wondering since we decided in one of the meetings to use `TNode` instead of `const Node&` as a rule of thumb (and I don't really agree with this rule of thumb).

src/theory/strings/sequences_rewriter.cpp
aniemetz: Yes, and we should avoid copy whenever possible.

src/smt/proof_manager.cpp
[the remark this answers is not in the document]
HanielB: I see, so in this case `const auto&` is fine. And are there cases where `const auto&` would be different from manually doing `const std::pair<const Key, const Data>&`?

src/theory/datatypes/sygus_extension.cpp
nafur: We said that for function arguments, mainly because it allows to be called with either Node (cheap copy to TNode) or TNode without a lot of overhead. In this case we actually know it's a Node, so it's a bit different.

src/smt/proof_manager.cpp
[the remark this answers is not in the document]
nafur: It depends on the container: for `CDHashMap` it will always be `std::pair<const Key, const Data>`. For `std::map` it would be `std::pair<const Key, Data>`, for `std::vector<T>` it would just be `T`.

aniemetz (approved):

PR #5638, opened 2020-12-09
Add methods to obtain values from api::Term

nafur:
This PR adds methods for `api::Term` to check whether it has some value that we can return as a native data type, and methods to return this data if appropriate.

### DIFF
--- a/src/api/cvc4cpp.cpp
+++ b/src/api/cvc4cpp.cpp
@@ -2104,17 +2104,17 @@ bool checkReal64Bounds(const Rational& r)
          && checkIntegerBounds<std::uint64_t>(r.getDenominator());
 }
 
-bool isRealString(const Node& node)
+bool isReal(const Node& node)
 {
   return node.getKind() == CVC4::Kind::CONST_RATIONAL;
 }
 bool isReal32(const Node& node)
 {
-  return isRealString(node) && checkReal32Bounds(getRational(node));
+  return isReal(node) && checkReal32Bounds(getRational(node));
 }
 bool isReal64(const Node& node)
 {
-  return isRealString(node) && checkReal64Bounds(getRational(node));
+  return isReal(node) && checkReal64Bounds(getRational(node));
 }
 
 bool isInteger(const Node& node)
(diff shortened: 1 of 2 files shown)


src/api/cvc4cpp.h
```diff
@@ -1151,6 +1151,76 @@ class CVC4_PUBLIC Term
   // to the new API. !!!
   const CVC4::Node& getNode(void) const;
 
+  // mkTuple
+  bool isTuple() const;
```
ajreynol, line 1155: I'm not sure I'd insist on `is` functions for each `mk`.

Concretely i dont think we need isTuple, isTrue, isFalse, isPi, isValueReal, isRegexpEmpty, isRegexpSigma, isEmptySet, isEmptyBag, isSepNil, isChar, isEmptySequence, isUniverseSet, isConstArray, isPosInf, or anything below it.

ajreynol: Conversely, I think we should consider:

std::unordered_set<api::Term> isSet / getSet, std::vector<api::Term> isSequence / getSequence

nafur: I added `isX` and `getX` for sets, sequences and tuples.
I would think that we want to have such functions for everything that we can construct via the API... we should discuss this internally.

aniemetz: From a user perspective, I think that we do. How else would you query what kind of value it is?

nafur:
> This PR in part uses the term constructor `Term(const Solver* slv, const CVC4::Node& n);` that is marked as being temporary only. We need some way to create a Term from a Node for the "container getters" like `getTuple()`, `getSet()` or `getSequence()`. Is there any alternative to this constructor?

I just noticed it is just **temporarily public**. Everything is fine if it stays as a private constructor.

src/api/cvc4cpp.cpp
```diff
@@ -2073,6 +2073,365 @@ CVC4::Expr Term::getExpr(void) const
...
+  for (std::size_t i = 1, n = d_node->getNumChildren(); i < n; ++i)
+  {
+    res.emplace_back(Term(d_solver, (*d_node)[i]));
+  }
+  return res;
+}
+
+bool Term::isPosZero() const
+{
+  if (isFloatingPoint())
```
mpreiner, line 2268: API functions should not call API functions since we will get issues in the future when we implement API tracing.

nafur: I think this is fixed for all methods from this PR. There are still cases of this pattern, for example for `Sort`.

nafur:
This has been done in #6496

PR #5639, opened 2020-12-09
Refactor regressions

ajreynol:
This adds a net +82 regressions to regress[0-2] and adds several additional disabled regressions to regress3 and regress4.  This involved fixing the status on several regressions, and ensuring CMakeLists.txt includes all files (exactly once) in the test/regress/ subdirectory.

It also moves several regressions to the proper regression levels (those that take >30 seconds in debug are moved to regress3+).


nafur (approved):
LGTM.

PR #5642, opened 2020-12-09
(proof-new) Make theory preprocessor proofs self contained

ajreynol:
Previously, there was a block of code in TheoryEngine::lemma for processing the proofs from theory preprocessing.  

It is much cleaner if this block was self contained in TheoryPreprocessor.

This is required for the planned move of TheoryPreprocessor from TheoryEngine -> PropEngine.


HanielB (approved):
LGTM

PR #5644, opened 2020-12-09
Fix compiler warnings.

aniemetz:


ajreynol (approved):

src/theory/strings/sequences_rewriter.cpp
```diff
@@ -1527,7 +1526,7 @@ Node SequencesRewriter::rewriteSeqNth(Node node)
     if (pos < len)
     {
       std::vector<Node> elements = s.getConst<Sequence>().getVec();
-      ret = elements[pos];
+      Node ret = elements[pos];
```
nafur, line 1529: To avoid the copy? :-)
```suggestion
      const Node& ret = elements[pos];
```

PR #5649, opened 2020-12-10
Add eager bitwise integer AND encoding

makaimann:
This is the eager counterpart to https://www.github.com/CVC4/CVC4. It adds a bitwise mode for the eager bv-to-int preprocessing pass. This model will create an `iand` node over the integer translation of the arguments for any bitwise AND. However, instead of relying on the CEGAR loop in the IAND solver, it will eagerly add all the bitwise equality axioms using an integer extract (with `div` and `mod`). 

NOTE: this will probably still call the `IAND` solver; hopefully that isn't an issue.


yoni206 (commented):
Thanks for the eager option!
It looks good to me!
Before merging, though, I would like to verify with @ajreynol that this is the best way we can simulate an eager translation. @ajreynol -- what do you think?

In a bit more detail:
with the new option, this preprocessing pass will replace `(bvand x y)` by `(iand k x' y')` (assuming `x'` and `y'` are the translations of `x` and `y`). Additionally, it will add to the assertion pipeline all the constraints regarding `(iand k x' y')` -- that is -- the values of all its bits (using divs and mods).

Since we are using the `iand` operator, the `iand` solver might be invoked, but I am not sure what it will do -- all the refinements that it would potentially add are already in the assertions.

yoni206:
Cool. Would it be better to get a single Skolem for each `k`, and its type would be `Int x Int -> Int`, rather than a new Skolem for each `k,x',y'` of type `Int`?

ajreynol:
It may be a little bit better.

Also, one thing I'm confused about: For (iand k x' y'), what if k, x', y' contain universal variables?

There is actually probably a more robust place for the code in this PR, which would be in TheoryArith::ppRewrite. This would handle cases where e.g. IAND terms that occur in instantiations are handled during solving.

yoni206:
universal variables are handled uniformly here: https://github.com/CVC4/CVC4/blob/e4fd524b02054a3ac9724f184e55a983cb6cb6b9/src/preprocessing/passes/bv_to_int.cpp#L1045
Should it be different for `iand`?

Regarding moving this to ppRewrite -- would you say the same would hold for the eager variant of `sum` mode? Like this code?
https://github.com/CVC4/CVC4/blob/e4fd524b02054a3ac9724f184e55a983cb6cb6b9/src/preprocessing/passes/bv_to_int.cpp#L445

ajreynol:
The `d_bitwiseAssertions` added in this PR, are we sure this does not contain formulas with free (universal) variables?

If an IAND node occurs beneath a quantified formula, then does it generate formulas that have free variables?
You can check `expr::hasFreeVar` on the formulas in d_bitwiseAssertions (see https://github.com/CVC4/CVC4/blob/master/src/expr/node_algorithm.h#L85).

It's not clear to me at first glance if/how this is handled in this PR.

makaimann:
> It's not clear to me at first glance if/how this is handled in this PR.

I didn't think about universal variables at all, so this definitely needs more thought. I'll make the change to use the skolem and talk to @yoni206 about how to handle universal variables. Marking `do not merge` for now.

yoni206:
@ajreynol you are right and I totally missed that!
For the range constraints, I handle them directly when introducing a quantifier:
https://github.com/CVC4/CVC4/blob/e4fd524b02054a3ac9724f184e55a983cb6cb6b9/src/preprocessing/passes/bv_to_int.cpp#L1046

So there is a bit of duplication going on between general range constraints and range constraints under quantifiers (although the main code is shared via a function).

@makaimann we will need to do something similar, probably incorporate additional constraints at the same place, together with range constraints. Let's take about it offline as u suggest.

yoni206:
@ajreynol what about `ppRewrite`? I am not sure I understand why this PR is better over there, and whether other parts of this preprocessing pass should go there as well.

ajreynol:
As a general rule of thumb, if the preprocessing pass introduces new (top-level) constraints, it is better to do this in ppRewrite.

In terms of the preprocessing refactor, ppRewrite may eventually be applied lazier, e.g. after conversion through CNF.

makaimann:
Closing because this was instead added to https://github.com/yoni206/CVC4/tree/bv_to_int_module which will be merged in the future.

PR #5655, opened 2020-12-11
(proof-new) Add bound variable manager

ajreynol:
This is a common utility for constructing canonical bound variables.


HanielB (commented):
I'm a bit confused... how is `d_cacheVals` used?

ajreynol:
It ensures that Node that are associated with bound variable are not deleted.  This is what ensures that bound variable creation is deterministic.

HanielB (approved):

PR #5657, opened 2020-12-12
Use scikit-build for python bindings and add cibuildwheel infrastructure

makaimann:
The main purpose of this PR is to allow building a Python wheel for binary distribution (e.g. through `PyPi` for `pip install pycvc4`). It is work towards an official PyPi package, see https://github.com/CVC4/CVC4/issues/5418.

In particular, it adds a top-level `setup.py` file which is [standard](https://packaging.python.org/tutorials/packaging-projects/) for `pip wheel` installation. This uses [scikit-build](https://github.com/scikit-build/scikit-build) to build CVC4 with python bindings. Essentially `scikit-build` is glue between `CMake` and `setuptools`. This PR also uses [cibuildwheel](https://github.com/joerick/cibuildwheel) to generate the wheels for different platforms automatically in GitHub Actions. In particular `cibuildwheel` handles both different python versions and different operating systems. For linux, it uses the [manylinux](https://github.com/pypa/manylinux) docker images so that the wheel can be expected to run on most modern linux distributions.

I refactored the python directory a bit to work better for building wheels. It also ended up being simpler. Now, the directory is `pycvc4` instead of `python`, and it is the module itself, instead of what we had before which created a new directory and copied files over, all within CMake: https://github.com/CVC4/CVC4/blob/107b1422f9549eb2128729c3fd173441029ba443/src/api/python/CMakeLists.txt#L57

A few issues / things to address:
* I had to update the cmake files which are copied here from `scikit-build`. There was an issue in the old version of `target_link_libraries_with_dynamic_lookup` that made it fail to build on CentOS (e.g. `manylinux`). Specifically it tried to link in the python libraries directly which I learned you [should not do](https://github.com/pypa/manylinux/blob/46241e9debbaf4b045c888af55ec80311bf27d99/pep-513.rst#libpythonxyso1) (except on Windows). That's actually what that `cmake` function is meant to avoid. I was also doing this manually in CMake, which this PR removes. Do you think it would be better to just add `scikit-build` as a dependency explicitly? Then we could more easily stay in line with their improvements.
* The install scripts (such as `macos-dependency-installs.sh`) are there because they're used by `ci.yml` and by [wheels.yml](https://github.com/makaimann/CVC4/blob/0234bd8178d9890d2208d1bcf90730cfb0a05d38/.github/workflows/wheels.yml#L20).
* Right now the `Build Wheels` GitHub Action runs all the time, which is definitely excessive because it takes a long time. I can make it run only on releases but then I worry that it will break between releases. Maybe we can set something up to run it elsewhere?
*  `wheels.yml` doesn't automatically upload to PyPi yet. I can work on that in this PR or make it a separate one. Whichever you prefer.
* I disabled the `windows-latest` build for `wheels.yml`. I wasn't sure if CVC4 compiles well on Windows and thought that could be a separate PR.
* The version number [here](https://github.com/makaimann/CVC4/blob/0234bd8178d9890d2208d1bcf90730cfb0a05d38/setup.py#L27) is disconnected from the one in `CMakeLists.txt` and has to be kept up to date (PyPi will reject repeat uploads). I don't like having it in two places, but the `setup.py` calls `cmake` so there's not really a way to configure the file from `cmake`. What do you think about a `toml` file with the version number that both `cmake` and this `setup.py` file read from?


makaimann:
I'm not sure who the right person is to review this so I just put @mpreiner but I welcome feedback from others. @4tXJ7f maybe? Additionally, I can split this up into multiple PRs if that's preferred. It's not quite done yet and I had some questions (see descriptions). So I'll probably just continue working on it here until it's ready and then split it up at that point if we decide to.

mpreiner:
Thanks for taking care of this!

1) I agree we should require `scikit-build` as dependency if CVC4 is configured with `--python-bindings`.

2) Sounds good! How about using the name `install-dependencies-<platform>.sh` for the scripts?

3) Depends on how much time it takes to build. We could also add it to the production python build?

4) Do we want to upload unstable (i.e. master branch) versions to PyPi?

5) CVC4 does not compile natively on Windows. We cross-compile Windows binaries.

6) I'd keep the version number in the top-level CMakeLists.txt. There are two options I can think of right now: setup.py gets the information from the file or we can somehow generate `setup.py` from a `setup.py.in` to have something like this:

```
setup(
    name="pycvc4",
    version="@CVC4_MAJOR@.@CVC4_MINOR@.@CVC4_RELEASE@",
    license="BSD",
    packages=['pycvc4'],
    package_dir={'pycvc4': 'src/api/pycvc4'},
    cmake_args=['-DBUILD_BINDINGS_PYTHON=ON', '-DBUILD_LIB_ONLY=ON']
)
```

What I don't like about the current approach is that it seems to not use `configure.sh` at all since you have to pass the cmake_args in setup.py?

I'd prefer to have our usual configure/build workflow but have an extra pypi target that is additionally built and uses the configured CVC4 version. What do you think?

makaimann:
Addressed 2 with e1dfb2ed6af948709822f1f74bd21aac0622328b and 1 with aa051398846f27d56ec2f47401dd7d6dabdbcbaf. It looks like Travis is failing but I don't think that has to do with those commits (`poly` is failing to build). I'll look into that more when I get a chance.

3. It takes a while, I can get you better estimates once it's running again
4. I'm not sure. I'd probably recommend only uploading to PyPi on releases, but don't feel that strongly. The version number has to change for PyPi to accept it anyway.
5. Good to know
6. Yeah I'm not very happy with the current set up either. For the version number, we have a cyclic dependency issue: we could generate a `setup.py` in CMake but then from within that `setup.py` we would call `configure.sh` (instead of directly calling cmake like you pointed out). 

I don't have a lot of spare cycles right now, but I'll look into 6 more into this when I get a chance. One thing I could look into is not using `scikit-build` at all (except for the CMake modules) and just creating a `wheel` after compiling CVC4 through the normal flow. Although, every approach for building wheels I've seen always builds it from within a Python `setup.py`. I.e. the `setup.py` is the top-level which then calls all the `CMake` infrastructure stuff. I'm not sure how to go about creating a `wheel` from an already compiled version, but I assume it's possible?

henryiii:
Ideally, setup.py calls the other stuff - if someone gets an SDist from PyPI, it should be able to build if possible; using pyproject.toml to enforce scikit-build is present. See https://scikit-hep.org/developer/packaging . Also, PyPI has a lifetime limit on upload size, so uploading dev builds there is really a bad idea usually. And of course, you cannot upload the same filename twice.

I'd probably read the version number from setup.py; it would be better to drive the configure and build scripts from there (see point above).

Not involved at all, just thought I'd chime in an offer some unsolicited random advice. ;)

makaimann:
Just FYI, I haven't forgotten about this PR. I had several other deadlines come up which are more urgent. I'll revisit this and bring it up to date soon.

makaimann:
Since this PR is old and outdated, I am going to close it and split it into multiple PRs starting with https://github.com/cvc5/cvc5/pull/6543. We can still refer to the conversation above for the new PRs.

PR #5667, opened 2020-12-14
[proof-new] Make prop engine proof producing

HanielB:
This makes the prop engine use the ProofCnfStream rather than the regular
CnfStream when proofs are enabled, as well as use the prop engine proof manager
to produce and check prop engine proofs when requested.

Note this PR also clang-formats the file.

In a next PR Minisat will be made producing, thus completing proof production
for the prop engine.

### DIFF
--- a/src/prop/prop_engine.cpp
+++ b/src/prop/prop_engine.cpp
@@ -124,7 +124,7 @@ void PropEngine::finishInit()
   // CNF stream will ignore it since the SAT solver already had it registered,
   // thus not having True as an assumption for the SAT proof. To solve this
   // issue we track it directly here
-  if (d_pfCnfStream)
+  if (isProofEnabled())
   {
     static_cast<MinisatSatSolver*>(d_satSolver)
         ->getProofManager()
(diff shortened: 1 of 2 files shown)


src/prop/prop_engine.cpp
```diff
@@ -152,17 +169,34 @@ void PropEngine::assertFormula(TNode node) {
   Assert(!d_inCheckSat) << "Sat solver in solve()!";
   Debug("prop") << "assertFormula(" << node << ")" << endl;
   // Assert as non-removable
-  d_cnfStream->convertAndAssert(node, false, false, true);
+  if (d_pfCnfStream)
```
ajreynol, line 172: Can you change this to `isProofEnabled()`? Add helper function if needed.

src/prop/prop_engine.cpp
```diff
@@ -152,17 +169,34 @@ void PropEngine::assertFormula(TNode node) {
...
 
 void PropEngine::assertLemma(theory::TrustNode trn, bool removable)
 {
   Node node = trn.getNode();
   bool negated = trn.getKind() == theory::TrustNodeKind::CONFLICT;
   Debug("prop::lemmas") << "assertLemma(" << node << ")" << endl;
-
   // Assert as (possibly) removable
-  d_cnfStream->convertAndAssert(node, removable, negated);
+  if (d_pfCnfStream)
```
ajreynol, line 190: Ditto.

ajreynol (changes requested):
One minor comment.

ajreynol (approved):

PR #5669, opened 2020-12-14
[proof-new] Making the CDCL(T) Minisat proof producing

HanielB:
This closely follows the old proof code in terms of where Minisat is hooked to the SatProofManager, other than a few places like registering removed clauses and removal of redundant literals. Note that this together with the thorough handling from SatProofManager makes the new SAT proofs perceptibly more robust than the old ones.

This PR also adds some traces to better track what Minisat does.

### DIFF
--- a/src/prop/minisat/core/Solver.cc
+++ b/src/prop/minisat/core/Solver.cc
@@ -55,7 +55,8 @@ namespace {
  */
 bool assertionLevelOnly()
 {
-  return options::unsatCores() && options::incrementalSolving();
+  return (options::proofNew() || options::unsatCores())
+         && options::incrementalSolving();
 }
 
 //=================================================================================================
@@ -221,6 +222,12 @@ Solver::Solver(CVC4::prop::TheoryProxy* proxy,
       propagation_budget(-1),
       asynch_interrupt(false)
 {
+  if (pnm)
+  {
+    d_pfManager.reset(
+        new SatProofManager(this, proxy->getCnfStream(), userContext, pnm));
+  }
+
   if (options::unsatCores())
   {
     ProofManager::currentPM()->initSatProof(this);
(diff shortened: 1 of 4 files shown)


src/prop/minisat/core/Solver.cc
```diff
@@ -920,9 +1036,17 @@ int Solver::analyze(CRef confl, vec<Lit>& out_learnt, int& out_btlevel)
             }
 
             // FIXME: can we do it lazily if we actually need the proof?
-            if (options::unsatCores() && level(var(q)) == 0)
+            if ((options::unsatCores() || isProofEnabled())
```
ajreynol, line 1039: Maybe drop the options check here, analogous to what you did on line 1061.

src/prop/minisat/core/Solver.cc
```diff
@@ -934,13 +1058,30 @@ int Solver::analyze(CRef confl, vec<Lit>& out_learnt, int& out_btlevel)
...
+          }
+          if (isProofEnabled())
+          {
+            d_pfManager->addResolutionStep(ca[confl], p);
+          }
         }
 
     }while (pathC > 0);
     out_learnt[0] = ~p;
+    if (CVC4::options::proofNew())
```
ajreynol, line 1075:
```suggestion
    if (options::proofNew())
```

ajreynol: Should this just be guarded with `Debug.isOn("newproof::sat")`?

ajreynol (changes requested):
A few minor things

src/prop/minisat/core/Solver.cc
HanielB: Yes :)

ajreynol (approved):

PR #5670, opened 2020-12-15
Move ownership of term formula removal to theory preprocessor

ajreynol:
This is work towards refactoring ITE removal (more generally, term formula removal) so that it happens at a configurable time, preferably post-CNF conversion.

This moves the TermFormulaRemover to the TheoryPreprocessor and changes several interfaces as a consequence of this move.

The next step will move the TheoryPreprocessor inside prop::TheoryProxy.

There are no behavior changes to solving in this PR.  One aspect of CheckModels is simplified.


HanielB (approved):
LGTM

PR #5676, opened 2020-12-15
Proper expand definitions for sets

ajreynol:
In the new view, `expandDefinitions` is used for eliminating partial operators and is not called during solving, and `ppRewrite` is called during preprocessing.

For sets, `choose` is a partial operator that should be applied in expand definitions, *and* in `ppRewrite`.  On the other hand, `is_singleton` should not be expanded in expandDefinitions since it is not a partial operator, so it should only be handled in ppRewrite.

It also removes some outdated documentation regarding expandDefinitions with universe set, which is now handled by preventing universe set from occurring in solved substitutions.

This is in preparation for refactoring `check-model`.


mudathirmahgoub (approved):
Looks good to me

PR #5677, opened 2020-12-15
Add getters to retrieve constants from api::Term

nafur:
This PR adds method to obtain constant values from `api::Term` that are either integers or strings.

### DIFF
--- a/src/api/cvc4cpp.cpp
+++ b/src/api/cvc4cpp.cpp
@@ -2165,29 +2165,9 @@ std::string Term::getInteger() const
   return detail::getInteger(*d_node).toString();
 }
 
-bool Term::isChar() const
-{
-  return d_node->getKind() == CVC4::Kind::CONST_STRING
-         && d_node->getConst<String>().getVec().size() == 1;
-}
 bool Term::isString() const
 {
   return d_node->getKind() == CVC4::Kind::CONST_STRING;
-}
-std::string Term::getChar() const
-{
-  CVC4_API_CHECK(d_node->getKind() == CVC4::Kind::CONST_STRING
-                 && d_node->getConst<String>().getVec().size() == 1)
-      << "Term should be a Char when calling getChar()";
-  unsigned codePoint = d_node->getConst<String>().getVec()[0];
-  const char* digits = "0123456789ABCDEF";
-  std::string res;
-  for (; codePoint > 0; codePoint /= 16)
-  {
-    res += digits[codePoint % 16];
-  }
-  std::reverse(res.begin(), res.end());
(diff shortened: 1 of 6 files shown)


src/api/cvc4cpp.h
```diff
@@ -1151,6 +1151,67 @@ class CVC4_PUBLIC Term
...
+  /**
+   * Returns the stored integer in (decimal) string representation. Asserts
+   * isInteger().
+   */
+  std::string getInteger() const;
+
+  /**
+   * Returns true if the term is a single string character.
+   */
+  bool isChar() const;
```
ajreynol, line 1199: I'd prefer not having this method.

src/api/cvc4cpp.h
```diff
@@ -1151,6 +1151,67 @@ class CVC4_PUBLIC Term
...
+  /**
+   * Returns true if the term is a string constant.
+   */
+  bool isString() const;
+  /**
+   * Returns true if the term is the empty sequence.
+   */
+  std::string getChar() const;
```
ajreynol, line 1207: Ditto.

ajreynol (changes requested):
I'd omit the isChar/getChar, otherwise this looks good to me.

ajreynol:
Also, we will need unit tests for each of these methods in https://github.com/CVC4/CVC4/blob/master/test/unit/api/solver_black.cpp

src/api/cvc4cpp.cpp
```diff
@@ -2073,6 +2073,109 @@ CVC4::Expr Term::getExpr(void) const
...
+  return checkIntegerBounds<std::int32_t>(r.getNumerator())
+         && checkIntegerBounds<std::uint32_t>(r.getDenominator());
+}
+bool checkReal64Bounds(const Rational& r)
+{
+  return checkIntegerBounds<std::int64_t>(r.getNumerator())
+         && checkIntegerBounds<std::uint64_t>(r.getDenominator());
+}
+
+bool isReal(const Node& node)
```
ajreynol, line 2102: So this may lead to some confusion.

There are actually two ways of representing real constants at the Term level.  A Term is a constant rational if its Node is CONST_RATIONAL{r} with a non-integer r, or (CAST_TO_REAL CONST_RATIONAL{r}) where r is an integral rational.  This is handled in appropriate places https://github.com/CVC4/CVC4/blob/master/src/api/cvc4cpp.h#L1178.
This is what allows us to hide subtyping.

I'd recommend for now inlining this method into isInteger.

ajreynol (approved):
LGTM

PR #5679, opened 2020-12-15
(proof-new) Use bound variable manager

ajreynol:
This uses BoundVarManager for several key places where bound variables are introduced, including for array extensionality witness terms, string preprocessing, quantifiers rewriting, and skolemization.

This is motivated by making certain steps in solving deterministic for the sake of proofs, and gives a more consistent pattern in general for constructing bound variables.


HanielB:
If I understand correctly this bound variable manager also simplifies stuff unrelated to proofs, like these `getOrMakeBV` whatever functions. It'd be good to add this to the description of the PR.

src/theory/quantifiers/quantifiers_rewriter.cpp
```diff
@@ -987,9 +966,12 @@ bool QuantifiersRewriter::getVarElim(Node n,
...
   if (nk == NOT)
   {
-    return getVarElim(n[0], !pol, args, vars, subs);
+    n = n[0];
+    pol = !pol;
+    nk = n.getKind();
+    Assert(nk != NOT);
   }
-  else if ((nk == AND && pol) || (nk == OR && !pol))
+  if ((nk == AND && pol) || (nk == OR && !pol))
```
HanielB, line 974: `else if`?

HanielB (approved):
LGTM, just a minor comment

src/theory/quantifiers/quantifiers_rewriter.cpp
ajreynol: Actually no, I removed the recursion from the NOT case, so now it should fall through and trigger this.

HanielB: Ah, I did not notice that `nk` is updated.

PR #5680, opened 2020-12-15
Move trigger trie to own file

ajreynol:
Also updates minor things to meet coding standards and makes it so that children in the trie are not dynamically allocated.


HanielB (approved):
LGTM

PR #5681, opened 2020-12-15
Use uint64 utility when parsing tuple selectors in smt2

ajreynol:
The smt2 parser is now 100% independent from the Expr-layer.

### DIFF
--- a/src/parser/smt2/smt2.cpp
+++ b/src/parser/smt2/smt2.cpp
@@ -18,7 +18,6 @@
 #include <algorithm>
 
 #include "base/check.h"
-#include "expr/expr.h"
 #include "options/options.h"
 #include "parser/antlr_input.h"
 #include "parser/parser.h"
@@ -1072,13 +1071,11 @@ api::Term Smt2::applyParseOp(ParseOp& p, std::vector<api::Term>& args)
   else if (p.d_kind == api::APPLY_SELECTOR && !p.d_expr.isNull())
   {
     // tuple selector case
-    std::string indexString = p.d_expr.toString();
-    Integer x = p.d_expr.getExpr().getConst<Rational>().getNumerator();
-    if (!x.fitsUnsignedInt())
+    if (!p.d_expr.isUInt32())
     {
-      parseError("index of tupSel is larger than size of unsigned int");
+      parseError("index of tupSel is larger than size of uint32_t");
     }
-    unsigned int n = x.toUnsignedInt();
+    uint32_t n = p.d_expr.getUInt32();
     if (args.size() != 1)
     {
       parseError("tupSel should only be applied to one tuple argument");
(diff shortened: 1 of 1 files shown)


src/parser/smt2/smt2.cpp
```diff
@@ -1072,13 +1071,11 @@ api::Term Smt2::applyParseOp(ParseOp& p, std::vector<api::Term>& args)
...
-    std::string indexString = p.d_expr.toString();
-    Integer x = p.d_expr.getExpr().getConst<Rational>().getNumerator();
-    if (!x.fitsUnsignedInt())
+    if (!p.d_expr.isUInt32())
     {
-      parseError("index of tupSel is larger than size of unsigned int");
+      parseError("index of tupSel is larger than size of uint32_t");
     }
-    unsigned int n = x.toUnsignedInt();
+    uint32_t n = p.d_expr.getUInt32();
```
nafur, line 1078: I'm not sure why you use `uint32_t` instead of `uint64_t` here: `x` is a `Rational` and thus `uint64_t` would be the more general choice, and `n` is then used as index into `dt[0][n]` which expects `std::size_t` (i.e. `uint64_t` on 64bit platforms).
```suggestion
    uint64_t n = p.d_expr.getUInt64();
```

nafur (changes requested):

src/parser/smt2/smt2.cpp
ajreynol: That's fine, I actually had assumed that size_t was uint32_t.

(FYI, the number in question here will rarely be >5).

src/parser/smt2/smt2.cpp
```diff
@@ -1072,13 +1071,11 @@ api::Term Smt2::applyParseOp(ParseOp& p, std::vector<api::Term>& args)
   else if (p.d_kind == api::APPLY_SELECTOR && !p.d_expr.isNull())
   {
     // tuple selector case
-    std::string indexString = p.d_expr.toString();
-    Integer x = p.d_expr.getExpr().getConst<Rational>().getNumerator();
-    if (!x.fitsUnsignedInt())
+    if (!p.d_expr.isUInt32())
```
nafur, line 1074:
```suggestion
    if (!p.d_expr.isUInt64())
```

src/parser/smt2/smt2.cpp
```diff
@@ -1072,13 +1071,11 @@ api::Term Smt2::applyParseOp(ParseOp& p, std::vector<api::Term>& args)
...
     {
-      parseError("index of tupSel is larger than size of unsigned int");
+      parseError("index of tupSel is larger than size of uint32_t");
```
nafur, line 1076:
```suggestion
      parseError("index of tupSel is larger than size of uint64_t");
```

nafur (changes requested):
Also adapt the check before that.

nafur (approved):
LGTM.

PR #5682, opened 2020-12-15
Simplify synth-fun printer

ajreynol:
Simplifies synth-fun printing to assume that the function-to-synthesize should be printed with the proper name and return type.


abdoo8080 (approved):
LGTM!

PR #5684, opened 2020-12-16
Mark quantifier instantiations as needs justify

ajreynol:
This avoids a solution soundness issue when disabling all NL strategies and using --nl-rlv=always.


nafur (approved):
LGTM.

PR #5685, opened 2020-12-16
Simplify and fix check models

### LINKED ISSUE #5645: Invalid model for datatype formula
cvc4 cc6e04b45
~~~~
(declare-datatypes ((Enum1 0)) (((None) (cons (data Bool)))))
(assert (forall ((var_1 Enum1)) (data var_1)))
(check-sat)
~~~~
~~~~
./cvc4 -i -q --check-models delta.out.smt2  
Fatal failure within void CVC4::smt::CheckModels::checkModel(CVC4::smt::Model*, CVC4::context::CDList<CVC4::NodeTemplate<true> >*, bool) at /home/CVC4/src/smt/check_models.cpp:249
assertion:     (forall ((var_1 Enum1)) (data var_1))
simplifies to: false
expected `true'.
Aborted (core dumped)
~~~~

ajreynol:
Currently `--check-models` is implemented by replaying several preprocessing steps, including theory-specific expand definitions, and then checking whether the result evaluates to true.

However, by having `--check-models` rely on complex preprocessing machinery defeats its purpose, as these steps are part of its trusted base.

Moreover, issue #5645 demonstrates that this may lead to spurious errors where we incorrectly conclude that an input assertion is false, when it is not.

This PR significantly simplifies `--check-models` so that it only relies on define-fun expansion + rewriting + evaluation.  This ensures that `--check-models` is "sound" i.e. it does not falsely report a formula as evaluating to false. As a consequence, this makes `check-models` give warnings more often, i.e. when partial operators are involved, thus `-q` is added to silence warnings on some regressions.

A followup PR will use a *satisfiability check* on the input formula post-expand-definitions to properly implement a trustworthy version of check-models that is robust for partial operators.

Fixes #5645.


nafur (approved):
LGTM.

PR #5687, opened 2020-12-16
Simplify internal API for sygus

ajreynol:
This makes simplifications to the internal sygus API now that the SmtEngine interface is independent of parsing.


abdoo8080 (approved):
LGTM!

PR #5688, opened 2020-12-16
[proof-new] Only use old proof code for unsat cores if no new proofs

HanielB:
This is so that eventually we can compare the performance of the old unsat cores vs the new ones.


ajreynol (approved):
LGTM

PR #5689, opened 2020-12-16
Fix issue with selector triggers

ajreynol:
Code could segfault if the number of wrongly applied selector applications was different from correctly applied ones.


HanielB (commented):
Shouldn't you refactor the use of `resetInstantiationRound`? It's a no-op now.

HanielB (approved):

PR #5690, opened 2020-12-16
Move ownership of theory preprocessor to TheoryProxy

ajreynol:
With this PR, TheoryEngine is independent of theory preprocessing.  All theory preprocessing is handled at the level of PropEngine.

No significant behavior changes in this PR.

The next step will make theory preprocessing not mandatory in preprocessing, and optionally done instead at the time when literals are asserted to TheoryEngine.


src/prop/prop_engine.h
```diff
@@ -188,7 +188,7 @@ class PropEngine
    * that is definitionally equal to it.  The result of this function
    * is that the Node can be queried via getSatValue().
    */
-  void ensureLiteral(TNode n);
+  Node ensureLiteral(TNode n);
```
HanielB, line 191: Document this change

src/prop/prop_engine.h
```diff
@@ -261,6 +261,18 @@ class PropEngine
  private:
   /** Dump out the satisfying assignment (after SAT result) */
   void printSatisfyingAssignment();
+
+  /**
+   * Converts the given formula to CNF and assert the CNF to the SAT solver.
```
HanielB, line 266:
```suggestion
   * Converts the given formula to CNF and asserts the CNF to the SAT solver.
```

src/prop/prop_engine.cpp
```diff
@@ -181,50 +197,94 @@ void PropEngine::assertFormula(TNode node) {
...
-    d_pfCnfStream->convertAndAssert(
-        node, negated, removable, trn.getGenerator());
+    Assert(tplemma.getGenerator() != nullptr);
+    // ensure closed, make the proof node eagerly here to debug
+    tplemma.debugCheckClosed("te-proof-debug", "TheoryEngine::lemma");
+    for (size_t i = 0, lsize = ppLemmas.size(); i < lsize; ++i)
+    {
+      Assert(ppLemmas[i].getGenerator() != nullptr);
+      ppLemmas[i].debugCheckClosed("te-proof-debug", "TheoryEngine::lemma_new");
+    }
```
HanielB, line 223: this should be option-guarded

HanielB (changes requested):

src/prop/prop_engine.h
ajreynol: The documentation was incorrect before, now it is correct.

src/prop/prop_engine.cpp
ajreynol: With proof-new-eager-checking?  This is already built in to debugCheckClosed.

HanielB: Right, but at this point we have this spurious loop if the option is off.

src/prop/prop_engine.h
HanielB: Hum, my interpretation of the current documentation is that `n` is added to some map (which is done in the cnf stream), not that you produce that as a result. I think that the latter should be said explicitly.

ajreynol: I'm not sure I follow.  The documentation states the purpose of the method: it returns a Node that can be queried via getSatValue.

HanielB (approved):

PR #5694, opened 2020-12-17
Improve handling of utf8 encoded inputs

### LINKED ISSUE #5673: Fatal failure at /src/util/string.cpp:37
CVC4 240dad8784
~~~~
(set-info :source |
Generated by: Zähl
|)

~~~~
~~~~
./cvc4  delta.out.smt2 
Fatal failure within CVC4::String::String(const std::vector<unsigned int>&) at /home/CVC4/src/util/string.cpp:37
Check failure

 u < num_codes()

Aborted (core dumped)
~~~~

nafur:
This PR fixes an issue where utf8 encoded inputs are incorrectly parsed into `CVC4::String`. We now use `std::mbtowc` to first turn the char sequence from the `std::string` input into a `std::wstring` and then process this `std::wstring` one charactor (`wchar_t`) at a time.
Fixes #5673


nafur:
While this PR fixes the issue at hand, I'm not sure whether this is the right place to do this conversion or whether there should be a more fundamental change (like using `std::wstring` all the way). Thus marking as question and waiting for some feedback...

nafur:
Now it's getting interesting: I looked at one regression (`bug001.smt2`) and reduced it to
```(declare-fun z () String)
(assert (= "\x4a" z))
(assert (= "\x4A" z))
(check-sat)
```
cvc4 returns `unsat` while z3 finds a model (`z = "J"`), and the method I changed is never called, apparently...

ajreynol (approved):
I think this looks good to me. I'd like @4tXJ7f to look as well.

The method `processAdHocStringEsc` fortunately isn't a very critical one, as it does not relate to string solving given the new standard.  Thus these changes should not impact the solver, and makes us more compatible with SMT-LIB it looks like (SMT-LIB standard 2.6 says utf-8 is allowed as input).

mpreiner (approved):
LGTM

mpreiner:
@ajreynol Changes look good to me, I removed do-no-merge label since this is PR is taking way too long to get in.

PR #5695, opened 2020-12-17
Simplify term formula removal interface

ajreynol:
This no longer needs some methods that were previously used specifically for ITE preprocessing and check-model.


nafur (approved):
LGTM.

PR #5696, opened 2020-12-17
Cache rewritten lemmas in TheoryInferenceManager

nafur:
The TheoryInferenceManager cached lemmas as they came in. This PR always rewrites before consulting the cache, making caching more consistent and robust. This change is coming in from proof-new.


ajreynol (approved):
LGTM

PR #5697, opened 2020-12-17
(proof-new) Prepare nonlinear extension and nl-ext for proofs.

nafur:
This PR prepares the nonlinear extension itself and the nl-ext part for proofs. In particular
- we add a proof checker for nl-ext
- we add a `CDProofSet` for nl-ext


ajreynol (approved):
LGTM

PR #5701, opened 2020-12-17
Eliminate recursion from the internals of term formula removal

ajreynol:
This makes all recursion (applying term formula removal on lemmas introduced by term formula removal) optionally happen at the top level call.

This is in preparation for making term formula removal lazier, in which case we will only apply one step of term formula removal at a time.

One QF_UFNIA regression times out due to changing the search, an option is changed for this benchmark.

### DIFF
--- a/src/smt/term_formula_removal.cpp
+++ b/src/smt/term_formula_removal.cpp
@@ -50,12 +50,71 @@ RemoveTermFormulas::~RemoveTermFormulas() {}
 theory::TrustNode RemoveTermFormulas::run(
     Node assertion,
     std::vector<theory::TrustNode>& newAsserts,
-    std::vector<Node>& newSkolems)
+    std::vector<Node>& newSkolems,
+    bool fixedPoint)
 {
   Node itesRemoved = runInternal(assertion, newAsserts, newSkolems);
+  Assert (newAsserts.size()==newSkolems.size());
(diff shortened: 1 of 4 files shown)


src/smt/term_formula_removal.cpp
```diff
@@ -50,14 +50,77 @@ RemoveTermFormulas::~RemoveTermFormulas() {}
...
+  {
+    return theory::TrustNode::null();
+  }
+  // if running to fixed point, we run each new assertion through the
+  // run lemma method
+  if (fixedPoint)
+  {
+    size_t i = 0;
+    std::unordered_set<Node, NodeHashFunction> processed;
+    while (i < newAsserts.size())
```
HanielB, line 68: why not a `for`?

HanielB (approved):
LGTM, just a minor comment

src/smt/term_formula_removal.cpp
ajreynol: Because newAsserts is being added to in the loop.

HanielB: So the size will change? Having the condition for the `for` as you have for the `while` would amount to the same, no?

ajreynol: i think its clearer this way, since it suggests that a fixed point is being run.

PR #5708, opened 2020-12-21
Add option to track and notify from CNF stream

ajreynol:
This adds functionality to CNF stream to allow e.g. TheoryProxy to be notified when a formula is asserted (not just literals).

This will be required for SAT relevancy.

No behavior changes in this PR.


src/prop/cnf_stream.h
```diff
@@ -41,6 +42,17 @@ namespace prop {
 
 class ProofCnfStream;
 
+/** A policy for how literals for formulas steps are handled in cnf_stream */
```
HanielB, line 45: typo?

HanielB (approved):
LGTM, just a minor comment

PR #5711, opened 2020-12-21
Make theory preprocess rewrite equalities a preprocessing pass

ajreynol:
Some theories rewrite equalities during ppRewrite. An example is arithmetic with the option `arith-rewrite-eq`, which rewrites `(= x y)` to `(and (>= x y) (<= x y))` during theory preprocessing.

This PR makes it so that ppRewrite is only called on equalities in preprocessing, during a new preprocessing pass "TheoryRewriteEq".  On the other hand, ppRewrite is *never* called on new equalities generated in lemmas from TheoryEngine.

In detail, the motivation for this change:

(1) Rewriting equalities during ppRewrite is dangerous since it may break invariants wrt theory combination. In particular, equalities in splitting lemmas originating from theory combination must *not* be theory-preprocessed, or else we may be non-terminating or solution unsound. This can happen if a theory requests a split on (= x y) but is not notified of this atom when another theory rewrites (= x y) during ppRewrite.

(2) After this PR, we can simplify our policy for all lemmas generated, in particular, we can say that *all lemmas must be theory preprocessed before their literals are asserted to TheoryEngine*.  This is now possible as the invariant cannot be broken (theoryRewriteEq is relegated to the preprocessor, which is only applied once).  This will make `LemmaProperty::PREPROCESS` obsolete, which in turn will simplify several lemma caches for nonlinear and quantifiers.  It will also significantly simplify proof production for the theory preprocessor (which maintains two stacks of utilities for preprocessed vs non-preprocessed lemmas).

(3) Simplifications to the above policy will make it significantly easier to implement theory-preprocessing apply when literals are asserted.  It is currently not possible to implement this in a coherent way without tracking which literals were a part of lemmas marked as "do not theory-preprocess".


ajreynol:
Marking "do not merge" since arrays, BV, and datatypes also rewrite equalities in the same way.  I will generalize this preprocessing pass to a "theory-rewrite-eq" pass.

src/preprocessing/passes/theory_rewrite_eq.cpp
```diff
@@ -0,0 +1,113 @@
...
+namespace passes {
+
+TheoryRewriteEq::TheoryRewriteEq(PreprocessingPassContext* preprocContext)
+    : PreprocessingPass(preprocContext, "theory-rewrite-eq"){};
+
+PreprocessingPassResult TheoryRewriteEq::applyInternal(
+    AssertionPipeline* assertions)
+{
+  // apply ppRewrite to all equalities in assertions
+  for (unsigned i = 0, size = assertions->size(); i < size; ++i)
```
nafur, line 32:
```suggestion
  for (std::size_t i = 0, size = assertions->size(); i < size; ++i)
```

nafur (approved):
Just one minor change, LGTM.

PR #5712, opened 2020-12-21
Have unsat core regression agnostic to number of assertions in core

HanielB:
Changing our unsat core infrastructure may change the cores we produce. This can be problematic for regressions that hardcode the number of assertions we get (such as this one). This commit changes such a regression to rather expect any core. It is motivated by exactly the previously described issue occurring after using the new proof infrastructure to generate unsat cores, which yielded a smaller core.

Kudos to @4tXJ7f for the `sed` magic.


nafur (approved):
LGTM.

PR #5716, opened 2020-12-22
Remove quant EPR option

ajreynol:
This was an experimental option used in combination with separation logic.


test/regress/CMakeLists.txt
```diff
@@ -2480,6 +2477,10 @@ set(regression_disabled_tests
   regress1/rels/garbage_collect.cvc
   regress1/sets/setofsets-disequal.smt2
   regress1/sets/finite-type/sets-card-neg-mem-union-2.smt2
+  # no longer support quant-epr + sep
+  regress1/sep/emp2-quant-unsat.smt2
+  regress1/sep/finite-witness-sat.smt2
+  regress1/sep/sep-fmf-priority.smt2
```
HanielB, line 2483: So we have no way of solving these regressions now?

ajreynol: No, i think if i'd support something like this, I'd implement it as a preprocessing pass.  The current implementation is fairly sketchy and hard to maintain.

HanielB: I see.

HanielB (approved):

PR #5717, opened 2020-12-22
Add new interfaces to term formula removal and theory preprocess

ajreynol:
This is in preparation for lazy lemmas for term formula removal.


src/smt/term_formula_removal.cpp
```diff
@@ -81,6 +86,13 @@ theory::TrustNode RemoveTermFormulas::run(
   return theory::TrustNode::mkTrustRewrite(assertion, itesRemoved, d_tpg.get());
 }
 
+theory::TrustNode RemoveTermFormulas::run(Node assertion)
```
nafur, line 89:
```suggestion
theory::TrustNode RemoveTermFormulas::run(TNode assertion)
```
The same in the header, and for the other overload of `run()` as well.

nafur (approved):
Just us `TNode` instead of `Node` for arguments, otherwise LGTM.

PR #5718, opened 2020-12-22
Delete duplicated code

yoni206:
A block of theory combination code appeared twice, one time after another.
This PR removes the second occurrence of the block.


ajreynol (approved):
LGTM

PR #5724, opened 2020-12-23
Dumping unsat cores after check-sat-assuming/QUERY

HanielB:
Previously we were not printing unsat cores when passing the option to dump them if we used the `check-sat-assuming` command or the QUERY command. This commit fixes this. 

It also kills the redundant `dump-synth` option, as it simplifies a bit what is going on in the command executor.


src/main/command_executor.cpp
```diff
@@ -163,7 +171,8 @@ bool CommandExecutor::doCommandSingleton(Command* cmd)
     {
       getterCommands.emplace_back(new GetModelCommand());
     }
-    if (d_options.getDumpProofs() && res.isUnsat())
+    // for now only use this route for old proof
```
ajreynol, line 174: Why do we need this?  I'd expect this behavior to not depend on new vs old proofs.

src/main/command_executor.cpp
```diff
@@ -136,11 +136,17 @@ bool CommandExecutor::doCommandSingleton(Command* cmd)
...
+  if (csa != nullptr)
+  {
+    d_result = res = csa->getResult();
+  }
   const QueryCommand* q = dynamic_cast<const QueryCommand*>(cmd);
   if(q != nullptr) {
     d_result = res = q->getResult();
   }
- const  CheckSynthCommand* csy = dynamic_cast<const CheckSynthCommand*>(cmd);
+  const CheckSynthCommand* csy = dynamic_cast<const CheckSynthCommand*>(cmd);
```
ajreynol, line 149: While you're at it, can you embed the getSynth inside the `if` block following this line?

ajreynol (approved):
LGTM

PR #5725, opened 2020-12-24
[proof-new] Only use old proofs for unsat cores if no proof new

HanielB:
Now the old proofs are used for unsat cores only if `proofNew` is disabled. Later commits will generate unsat cores from the new proofs when requested. Eventually we will compare them and when we confirm the new unsat core generation is better we will delete the old one.

This also does some minor refactoring in some preprocessing. No behavior is changed.


ajreynol (approved):
LGTM

PR #5731, opened 2020-12-31
Add bags inference generator 

mudathirmahgoub:
This PR adds  inference generator for basic bag rules.


src/theory/bags/bag_solver.h
```diff
@@ -0,0 +1,66 @@
...
+class BagSolver
+{
+ public:
+  BagSolver(SolverState& s, InferenceManager& im, TermRegistry& tr);
+  ~BagSolver();
+
+  void postCheck();
+
+ private:
+  /** apply inference rules for union disjoint */
```
ajreynol, line 44: Can you say more about what happens operationally here?  You add pending inferences to the InferenceManager? Or does this actually send the lemmas/facts?

src/theory/bags/infer_info.cpp
```diff
@@ -0,0 +1,100 @@
...
+
+namespace CVC4 {
+namespace theory {
+namespace bags {
+
+const char* toString(Inference i)
+{
+  switch (i)
+  {
+    default: return "?";
```
ajreynol, line 27: Implement this already?

src/theory/bags/infer_info.h
```diff
@@ -0,0 +1,127 @@
...
+   * literals that currently hold in the equality engine.
+   */
+  std::vector<Node> d_premises;
+
+  /**
+   * A list of new skolems introduced as a result of this inference. They
+   * are mapped to by a length status, indicating the length constraint that
+   * can be assumed for them.
+   */
+  std::vector<Node> d_new_skolem;
```
ajreynol, line 96: Nit: use camel case `d_newSkolem`.

ajreynol (changes requested):
This is great progress, just some minor comments.

src/theory/bags/infer_info.cpp
mudathirmahgoub: done

src/theory/bags/infer_info.h
mudathirmahgoub: done

src/theory/bags/bag_solver.h
mudathirmahgoub: I would like to discuss this

ajreynol: I think this beyond the scope of the PR, the design here may change later, but the code is fine for now.

What could be considered is to buffer the inferences so that we prefer e.g. conflicts over lemmas, but this can be addressed later.

ajreynol (approved):

PR #5733, opened 2021-01-04
Make sure polynomials are properly factorized in nl-cad

### LINKED ISSUE #5726: Refutation soundness issue on QF_NRA formula
Hi, for the following formula
~~~~
(set-logic QF_NRA)
(declare-fun r0 () Real)
(declare-fun r1 () Real)
(assert (and (not (<= (* r1 (* r1 (* r1 (* r1 (* r1 r1))))) 0)) 
             (not (<= (* r0 (* r0 (- r0 (+ r0 (* r0 (* r0 (+ (* r1 (- r1 (/ 1 720))) (* r0 r0)))))))) 0))))
(check-sat)
~~~~
cvc4 e0dfc0a343d yields `unsat`, but z3 yields `sat`.
Below is a model given by z3
~~~~
  (define-fun r0 () Real
    (/ 1.0 2048.0))
  (define-fun r1 () Real
    (/ 1.0 1024.0))
~~~~

nafur:
CAD theory (used in nl-cad) requires that polynomials are properly factorized (a finest square-free basis). This PR replaces usage of raw std::vector by a new wrapper `PolyVector` that ensures proper factorization whenever a polynomial is added. This fixes one piece of code that omitted factorization, leading to soundness issues as in #5726.
Fixes #5726.

### DIFF
--- a/src/theory/arith/nl/cad/cdcac.cpp
+++ b/src/theory/arith/nl/cad/cdcac.cpp
@@ -39,16 +39,6 @@ namespace arith {
 namespace nl {
 namespace cad {
 
-namespace {
-/** Removed duplicates from a vector. */
-template <typename T>
-void removeDuplicates(std::vector<T>& v)
-{
-  std::sort(v.begin(), v.end());
-  v.erase(std::unique(v.begin(), v.end()), v.end());
-}
-}  // namespace
-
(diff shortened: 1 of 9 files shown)


ajreynol (approved):
LGTM

PR #5746, opened 2021-01-05
Adding str.len to triggers

yoni206:
This PR adds `str.len` to symbols that are considered for instantiations.
It is motivated by a benchmark that originated from Boogie.
A minimized version of that benchmark is added as a regression.


ajreynol (approved):
LGTM, thanks for adding this.

PR #5747, opened 2021-01-06
strings arith checks preprocessing pass: step 1

yoni206:
For example, `len(s) >= 0` would be replaced by `true`.
This will make use of `CVC4::theory::strings::ArithEntail::check`.

This PR is the first step. It only includes the preprocessing pass infrastructure, with an empty implementation of the main function `StrLenSimplify::simplify`. It also adds the pass to the registry.
The implementation of this function is not complicated, but is left for a future PR in order to keep the PR short.

Future PRs will include an implementation of the main function, tests, and a command line option to turn on the pass.

### DIFF
--- a/src/preprocessing/passes/foreign_theory_rewrite.cpp
+++ b/src/preprocessing/passes/foreign_theory_rewrite.cpp
@@ -1,5 +1,5 @@
 /*********************                                                        */
-/*! \file str_len_simplify.cpp
+/*! \file foreign_theory_rewrite.cpp
  ** \verbatim
  ** Top contributors (to current version):
  **   Yoni Zohar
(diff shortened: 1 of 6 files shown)


ajreynol:
FYI @4tXJ7f .

My initial comment is that I would prefer if the preprocessing pass was more general.  The key feature is that this is a case where one theory (Strings) is rewriting literals from another theory (Arithmetic).

Maybe the preprocessing pass would be called something like "ForeignTheoryRewrite" or something. 

I think we should settle on a good name for this, so that if we are tempted to do something similar in spirit, we can add to this preprocessing pass.

src/preprocessing/passes/foreign_theory_rewrite.cpp
```diff
@@ -0,0 +1,48 @@
...
+namespace preprocessing {
+namespace passes {
+
+using namespace CVC4::theory;
+
+ForeignTheoryRewrite::ForeignTheoryRewrite(PreprocessingPassContext* preprocContext)
+    : PreprocessingPass(preprocContext, "foreign-theory-rewrite"),
+      d_cache(preprocContext->getUserContext()){};
+
+Node ForeignTheoryRewrite::simplify(Node n) { assert(false); }
```
4tXJ7f, line 32:
```suggestion
Node ForeignTheoryRewrite::simplify(Node n) { Unimplemented(); }
```

ajreynol (approved):
LGTM

4tXJ7f (changes requested):

PR #5749, opened 2021-01-06
bv-to-int: avoid binarizing nodes twice

yoni206:
In bv-to-int, we first binarize the applications of associative-commutative operators (like `bvadd` etc.).
With this PR, we first check whether we already binarized a node, and only if we didn't, we perform binarization.


alex-ozdemir (approved):
Pending tests, this looks good to me.

PR #5750, opened 2021-01-06
Strings arith checks preprocessing pass: step 2

yoni206:
For example, len(s) >= 0 would be replaced by true.
This will make use of CVC4::theory::strings::ArithEntail::check.

This PR is the *second* step. It includes the implementation of the main function, as well as unit tests for it.

A subsequent PR will add a user-level option that will turn on this preprocessing pass, as well as regression tests.


src/preprocessing/passes/foreign_theory_rewrite.cpp
```diff
@@ -29,7 +29,72 @@ ForeignTheoryRewrite::ForeignTheoryRewrite(PreprocessingPassContext* preprocCont
...
+       }))
+  {
+    // the node is rewritten, and so GT, LT, LEQ
+    // should be eliminated
+    Assert(current.getKind() != kind::GT);
+    Assert(current.getKind() != kind::LT);
+    Assert(current.getKind() != kind::LEQ);
+
+    // for GEQ, we check whether the node can be simplified
+    if (current.getKind() == kind::GEQ)
```
ajreynol, line 49: Put this in its own method.  Also, you should not stop at GEQ, instead traverse it.  The rewrite should be applied to fixed point, you should also traverse the results of the rewritten term.

ajreynol (changes requested):

test/unit/preprocessing/pass_foreign_theory_rewrite_white.h
```diff
@@ -0,0 +1,115 @@
...
+{
+  ExprManager* d_em;
+  NodeManager* d_nm;
+  SmtEngine* d_smt;
+  SmtScope* d_scope;
+  CircuitPropagator* d_cp;
+  ProofNodeManager* d_pnm;
+  PreprocessingPassContext* d_ppc;
+  PreprocessingPassRegistry* d_ppr;
+  ForeignTheoryRewrite* d_foreignTheoryRewritePP;
```
4tXJ7f, line 52: These should be move into a `private` section at the bottom.

4tXJ7f (changes requested):

ajreynol:
We should also consider using this new utility: https://github.com/CVC4/CVC4/blob/proof-new/src/proof/term_processor.h.

This would allow you to only implement a single Node -> Node method.

test/unit/preprocessing/pass_foreign_theory_rewrite_white.h
yoni206: 10x -- done.
825e55a

src/preprocessing/passes/foreign_theory_rewrite.cpp
yoni206: Thanks! fixed here: b5f7f6c

yoni206:
> We should also consider using this new utility: https://github.com/CVC4/CVC4/blob/proof-new/src/proof/term_processor.h.
> 
> This would allow you to only implement a single Node -> Node method without writing new traversal code.

How much different is it from @alex-ozdemir 's https://github.com/CVC4/CVC4/blob/master/src/expr/node_traversal.h (which i am currently using here)?

src/preprocessing/passes/foreign_theory_rewrite.cpp
```diff
@@ -29,7 +29,72 @@ ForeignTheoryRewrite::ForeignTheoryRewrite(PreprocessingPassContext* preprocCont
...
+         return d_cache.find(tn) != d_cache.end();
+       }))
+  {
+    // the node is rewritten, and so GT, LT, LEQ
+    // should be eliminated
+    Assert(current.getKind() != kind::GT);
+    Assert(current.getKind() != kind::LT);
+    Assert(current.getKind() != kind::LEQ);
+
+    // for GEQ, we check whether the node can be simplified
```
ajreynol, line 89: This check should go in the method.  The code here should not hardcode GEQ.  

You should write a method
`Node doForeignRewrite(Node n)` that is independent of the traversal.

src/preprocessing/passes/foreign_theory_rewrite.cpp
```diff
@@ -29,7 +29,72 @@ ForeignTheoryRewrite::ForeignTheoryRewrite(PreprocessingPassContext* preprocCont
...
+  n = Rewriter::rewrite(n);
+  // traverse the node
+  for (TNode current :
+       NodeDfsIterable(n, VisitOrder::POSTORDER, [this](TNode tn) {
+         return d_cache.find(tn) != d_cache.end();
+       }))
+  {
+    // the node is rewritten, and so GT, LT, LEQ
+    // should be eliminated
+    Assert(current.getKind() != kind::GT);
```
ajreynol, line 85: These checks should also be independent of the traversal code.

src/preprocessing/passes/foreign_theory_rewrite.cpp
```diff
@@ -24,12 +24,120 @@ namespace preprocessing {
...
+      d_cache[current] = result;
+      // (d) the result to toVisit
+      toVisit.push_back(result);
+    }
+    else
+    {
+      // current was already processed
+      // remove from toVisit and skip
+      toVisit.pop_back();
+      continue;
```
ajreynol, line 77: Can remove this line.

src/preprocessing/passes/foreign_theory_rewrite.cpp
```diff
@@ -24,12 +24,120 @@ namespace preprocessing {
...
+      {
+        Assert(d_cache.find(child) != d_cache.end());
+        Assert(!d_cache[child].get().isNull());
+        processedChildren.push_back(d_cache[child]);
+      }
+      Node reconstruction = reconstructNode(current, processedChildren);
+      // (c) process the node and store the result in the cache
+      Node result = foreignRewrite(reconstruction);
+      d_cache[current] = result;
+      // (d) the result to toVisit
```
ajreynol, line 69: You should only readd the node if it changed.

src/preprocessing/passes/foreign_theory_rewrite.cpp
```diff
@@ -24,12 +24,120 @@ namespace preprocessing {
...
+    return n;
+  }
+}
+
+Node ForeignTheoryRewrite::rewriteStringsGeq(Node n)
+{
+  // check if the node can be simplified to true
+  if (theory::strings::ArithEntail::check(n[0], n[1], false))
+  {
+    return NodeManager::currentNM()->mkConst<bool>(true);
```
ajreynol, line 106:
```suggestion
    return NodeManager::currentNM()->mkConst(true);
```

src/preprocessing/passes/foreign_theory_rewrite.h
```diff
@@ -39,8 +39,18 @@ class ForeignTheoryRewrite : public PreprocessingPass
...
-  // the main function that simplifies n
+  /* the main function that simplifies n.
+   * does a traversal on n and call rewriting fucntions.
+   */
   Node simplify(Node n);
+  /* A specific simplification function specific for GEQ
+   * constraints in strings.
+   */
+  static Node rewriteStringsGeq(Node n);
+  // invoke rewrite functions for n
```
ajreynol, line 50: This is the key method in the preprocessing pass, can you give more details?

src/preprocessing/passes/foreign_theory_rewrite.h
```diff
@@ -39,8 +39,18 @@ class ForeignTheoryRewrite : public PreprocessingPass
...
+  static Node foreignRewrite(Node n);
+  // reconstruct a node with different children
```
ajreynol, line 52:
```suggestion
  /** construct a node with the same operator as originalNode whose children are processedChildren */
```

test/unit/preprocessing/pass_foreign_theory_rewrite_white.h
```diff
@@ -0,0 +1,111 @@
...
+using namespace CVC4::theory::booleans;
+using namespace CVC4::smt;
+
+class ForeignTheoryRewriteWhite : public CxxTest::TestSuite
+{
+ public:
+  ForeignTheoryRewriteWhite() {}
+
+  void setUp() override
+  {
```
ajreynol, line 48: This seems really excessive to test this theory preprocessing pass.  Either we should make this a more generic utility, or you can just test a static function of your preprocessing pass.

I would prefer the latter.

ajreynol (changes requested):

src/preprocessing/passes/foreign_theory_rewrite.h
```diff
@@ -39,8 +39,23 @@ class ForeignTheoryRewrite : public PreprocessingPass
...
-  // the main function that simplifies n
+  /* the main function that simplifies n.
+   * does a traversal on n and call rewriting fucntions.
+   */
   Node simplify(Node n);
+  /* A specific simplification function specific for GEQ
+   * constraints in strings.
+   */
+  static Node rewriteStringsGeq(Node n);
+  // invoke rewrite functions for n.
```
ajreynol, line 50: For style, can you use /** */?

src/preprocessing/passes/foreign_theory_rewrite.cpp
```diff
@@ -24,12 +24,122 @@ namespace preprocessing {
...
+Node ForeignTheoryRewrite::reconstructNode(Node originalNode,
+                                           vector<Node> newChildren)
+{
+  // Nodes with no children are reconstructed to themselves
+  if (originalNode.getNumChildren() == 0)
+  {
+    Assert(newChildren.empty());
+    return originalNode;
+  }
+  else
```
ajreynol, line 125: Nit: we tend to avoid using "else" after return statements as they are spurious.  There are several instance of this in this PR.

ajreynol (changes requested):
Almost there, just a few style comments.

ajreynol (approved):
LGTM

src/preprocessing/passes/foreign_theory_rewrite.cpp
```diff
@@ -24,12 +24,113 @@ namespace preprocessing {
...
 using namespace CVC4::theory;
-
 ForeignTheoryRewrite::ForeignTheoryRewrite(PreprocessingPassContext* preprocContext)
     : PreprocessingPass(preprocContext, "foreign-theory-rewrite"),
       d_cache(preprocContext->getUserContext()){};
 
-Node ForeignTheoryRewrite::simplify(Node n) { assert(false); }
+Node ForeignTheoryRewrite::simplify(Node n)
+{
+  vector<Node> toVisit;
```
4tXJ7f, line 33:
```suggestion
  std::vector<Node> toVisit;
```

Not sure where we have a `using namespace std;` but either we should have it in this file or we should just write it out (which I prefer). Same applies to other `std` objects.

4tXJ7f (changes requested):
Just a minor comment.

4tXJ7f (approved):
Thank you for the updates. LGTM

PR #5751, opened 2021-01-07
[proof-new]  Implementing getProof in the API and SMT engine

HanielB:
A proof is represented as a string in `GetProofCommand`. The string is generated by the custom ways in which the SMT engine may choose to print the proof, based on `proof-format-mode` (to be added in subsequent commits).


src/api/cvc4cpp.h
```diff
@@ -3161,6 +3161,15 @@ class CVC4_PUBLIC Solver
    */
   std::vector<Term> getUnsatCore() const;
 
+  /**
```
ajreynol, line 3164: I don't think this shouldn't be exported yet.

We should maybe handle proofs analogously to how we handle models currently.

For now, this means that the command layer should retrieve the SmtEngine from the Solver when processing the GetProofCommand and store the result locally (as a string). See https://github.com/CVC4/CVC4/blob/master/src/smt/command.cpp#L1753.

ajreynol (changes requested):

src/api/cvc4cpp.h
HanielB: Sounds good. I've updated the PR

ajreynol (approved):
LGTM

PR #5753, opened 2021-01-07
Remove dependency on expression layer in TPTP parser

HanielB:


ajreynol (approved):
Can you also remove the includes from Tptp.g?

PR #5754, opened 2021-01-07
Rename strings::InferInfo::getAntecedent to InferInfo::getPremises

mudathirmahgoub:
Changes:
- renamed d_new_skolem to d_newSkolem 
- renamed d_ant to d_premises (antecedent is usually used with consequent, and premise is used with conclusion)


mudathirmahgoub:
Since all fields in InferInfo are public, I do not see the reason for `InferInfo::getAntecedent` other than friendly name for `d_ant`.

ajreynol:
The method `InferInfo::getAntecedent` used to have to append two vectors (until a recent change where two vectors were merged into one). However, it still makes the AND node from the vector, I'd keep it.

ajreynol (approved):
LGTM

PR #5756, opened 2021-01-07
Further simplifications in preparation for removing Expr layer

ajreynol:
This deletes variable flags from NodeManager::mkVar and moves ExprManager sort flags to NodeManager.

These flags are used for determining when a variable or sort should be printed via the old dump infrastructure.  The old dump infrastructure is simplified in this PR accordingly.

This PR should preserve behavior of the previous dumping with a minor exception that the internal trace "declarations" will also included symbols introduced from define-fun. This will be further refactored later.  This is in preparation for removing the includes `expr.h`/`expr_manager.h` from `node_manager.h`.


HanielB (approved):
LGTM, but it'd be good to improve the PR description to describe the changes to the dumping infrastructure.

PR #5757, opened 2021-01-08
Fix issue 5513

mudathirmahgoub:
Fix issue5513 by throwing an exception for unsupported bag operators


ajreynol:
Technically, to fix this, you need to use `LogicException`, see e.g. https://github.com/CVC4/CVC4/blob/master/src/theory/strings/term_registry.cpp#L149.

mudathirmahgoub:
> Technically, to fix this, you need to use `LogicException`, see e.g. https://github.com/CVC4/CVC4/blob/master/src/theory/strings/term_registry.cpp#L149.

Done

ajreynol (approved):

PR #5763, opened 2021-01-11
Seq len quant pp pr 3

yoni206:
For example, len(s) >= 0 would be replaced by true.
This will make use of CVC4::theory::strings::ArithEntail::check.

This PR is the third and final step. It adds the user-facing option that turn this feature on, as well as regression tests.


test/regress/regress0/seq/len_simplify_unknown.smt2
```diff
@@ -0,0 +1,5 @@
+; COMMAND-LINE: -q
```
ajreynol, line 1: I would omit this benchmark.  It isn't very useful to test that something doesn't work, since this will only lead to changing if e.g. the option you give becomes default or another technique solves this by default.

src/options/smt_options.toml
```diff
@@ -680,6 +680,16 @@ header = "options/smt_options.h"
   read_only  = true
   help       = "Force no CPU limit when dumping models and proofs"
 
+[[option]]
```
ajreynol, line 683: Would be curious to test enabling this by default on QF_S / QF_SLIA SMT-LIB.

ajreynol (changes requested):

PR #5774, opened 2021-01-13
(proof-new) Fix proofs for sine secant lemmas.

nafur:
This PR fixes a bug in proof generation for sine secant lemmas which set the incorrect degrees for taylor polynomials.
Additionally, it adds some trace messages to debug transcendental proofs in the proof checker.


ajreynol (approved):
LGTM

PR #5776, opened 2021-01-13
Updates to theory preprocess equality

ajreynol:
This makes 3 changes related to arithmetic preprocessing of equalities which revert to the original behavior of master before https://github.com/CVC4/CVC4/commit/a129c57c758b7161be23ebb12d41c05c72154ece.

For background, the commit https://github.com/CVC4/CVC4/commit/a129c57c758b7161be23ebb12d41c05c72154ece unintentionally changed the default behavior slightly in 3 ways (each corrected in this PR), which led a performance regression on QF_LIA in current master.

The 3 fixes are:
(1) Rewrite equalities should be applied as a post-rewrite, not a pre-rewrite in the theory-rewrite-eq preprocessing pass.  This is particularly important for equalities between ITE terms that contain other equalities recursively.
(2) theory-rewrite-eq should apply after rewriting and just before the normal theory preprocessing.
(3) The arith-brab test should call ppRewrite on the arithmetic equality it *introduces*, as it has a choice of whether to eliminate the equality before the lemma is sent out.


ajreynol:
Results QF_LIA, 300 second timeout:
```
                      CVC4-121020-sid_dj    1919    106929    2921    102469      4840      6947
                          CVC4-122320_dj    1635   93850.4    2665   91292.6      4300      6947
                   CVC4-011321-2-tppp_dj    1913    103675    2939    101938      4852      6947
```
CVC4-121020-sid_dj = previous master before a129c57
CVC4-122320_dj = current master after a129c57
CVC4-011321-2-tppp_dj  = this PR

nafur (approved):
LGTM.

PR #5783, opened 2021-01-19
Refactor bags::SolverState

mudathirmahgoub:
Couple of changes:
- SolverState now keep tracks of elements per bag instead of per type.
- bags::InferInfo now stores multiple conclusions (conjuncts). 
- BagSolver applies upward/downward closures for bag elements


src/theory/bags/inference_generator.cpp
```diff
@@ -261,9 +262,10 @@ Node InferenceGenerator::getMultiplicitySkolem(Node element,
                                                InferInfo& inferInfo)
 {
   Node count = d_nm->mkNode(kind::BAG_COUNT, element, bag);
-  Node skolem = d_state->registerBagElement(count);
+  Node skolem = d_state->registerCountTerm(count);
   eq::EqualityEngine* ee = d_state->getEqualityEngine();
   ee->assertEquality(skolem.eqNode(count), true, d_nm->mkConst(true));
+  // inferInfo.d_conclusions.push_back(skolem.eqNode(count));
```
mudathirmahgoub, line 268: This commented line fails regressions whereas line 266 does not. This needs discussion.

src/theory/bags/bag_solver.cpp
```diff
@@ -53,11 +53,27 @@ void BagSolver::postCheck()
...
+  elements.insert(upwards1.begin(), upwards1.end());
+  return elements;
+}
+
 void BagSolver::checkUnionDisjoint(const Node& n)
 {
   Assert(n.getKind() == UNION_DISJOINT);
-  TypeNode elementType = n.getType().getBagElementType();
-  for (const Node& e : d_state.getElements(elementType))
+  set<Node> elements = getElementsForBinaryOperator(n);
```
ajreynol, line 75:
```suggestion
  std::set<Node> elements = getElementsForBinaryOperator(n);
```

ajreynol: Similar for other places, I prefer always having std:: for standard library structures.

src/theory/bags/infer_info.h
```diff
@@ -81,8 +81,8 @@ class InferInfo : public TheoryInference
   bool process(TheoryInferenceManager* im, bool asLemma) override;
   /** The inference identifier */
   Inference d_id;
-  /** The conclusion */
-  Node d_conclusion;
+  /** The conclusions */
+  std::vector<Node> d_conclusions;
```
ajreynol, line 85: I don't think we should make this change.  If necessary, I would make multiple conclusions into an AND node.  This is what strings does.

This is important since this class will eventually be merged with a standard InferInfo class, which will not have a vector of conclusions.

src/theory/bags/inference_generator.cpp
ajreynol: You need to be sure you are processing pending assertions.  I would remove the use of `ee` here for now.

src/theory/bags/solver_state.h
```diff
@@ -31,24 +31,47 @@ class SolverState : public TheoryState
...
+  Node getCountSkolem(const Node& countTerm);
 
-  std::set<Node>& getBags();
+  /** clear all bags data structures */
+  void reset();
 
-  std::set<Node>& getElements(TypeNode t);
-
-  std::map<Node, Node>& getBagElements(Node B);
+  /** merge the elements of the two registered bags n1, n2 */
```
ajreynol, line 63: I'm confused why you are implementing this method.

I thought the elements of a bag were only being computed at full effort? 

This method is called prior to full effort, when equivalence classes merge.  You should not be implementing something like this (or if you were, the data structures in this class would all have to be SAT-context-dependent and the design of this class would have to change considerably).

For now, I do not think this method will be necessary.

ajreynol (changes requested):

src/theory/bags/solver_state.h
mudathirmahgoub: Initially I started collected all terms at post check, and then ran into the issue of merging elements in an equivalent class. 
```
[Z , (union_max A B), (intersection_min A B)] and 
terms (count x Z), (count y (intersection_min A B))
```
I thought it is simpler to use notifications for now and worry about context-dependence later.

src/theory/bags/infer_info.h
mudathirmahgoub: I suggest changing the standard InferInfo to have a vector of conclusions which is more flexible. 
In some situations, I don't know in advance how many inferences I need. To use AND I have to start with a default conclusion (TRUE) in this case. Then I need to add ugly code to check for this case. 
Empty conclusions is nicely handled in the process function

src/theory/bags/bag_solver.cpp
mudathirmahgoub: I prefer the same. I just forgot during copy/paste

src/theory/bags/inference_generator.cpp
mudathirmahgoub: Done. It was a stupid mistake inside InferInfo::process

src/theory/bags/solver_state.h
```diff
@@ -31,24 +31,43 @@ class SolverState : public TheoryState
...
+   * @pre bag A needs is already registered using registerBag(A)
+   * @return a unique skolem for (bag.count e A)
+   */
+  void registerCountTerm(TNode n);
+  /** get all bag terms */
+  const std::set<Node>& getBags();
+  /**
+   * @pre B is a registered bag
+   * @return all elements associated with bag B so far
+   * Note that associated elements are not necessarily elements in B
```
ajreynol, line 47: Is this documentation still accurate?  I would not expect x to be associated with A in this example.

src/theory/bags/solver_state.h
```diff
@@ -31,24 +31,43 @@ class SolverState : public TheoryState
  public:
   SolverState(context::Context* c, context::UserContext* u, Valuation val);
 
-  void registerClass(TNode n);
+  void registerBag(TNode n);
```
ajreynol, line 34: Can you add some documentation here? When is this method called? What is the effect (is n added to some set of this class i assume?).

src/theory/bags/solver_state.h
```diff
@@ -31,24 +31,43 @@ class SolverState : public TheoryState
...
   Node d_true;
   Node d_false;
+  /** node manager for this solver state */
+  NodeManager* d_nm;
+  /** collection of bag representatives */
   std::set<Node> d_bags;
-  std::map<TypeNode, std::set<Node>> d_elements;
-  /** bag -> element -> multiplicity */
-  std::map<Node, std::map<Node, Node>> d_count;
+  /** bag -> associated elements */
```
ajreynol, line 69: Is the domain of this map representatives?

src/theory/bags/theory_bags.cpp
```diff
@@ -213,34 +221,20 @@ void TheoryBags::presolve() {}
 
 void TheoryBags::eqNotifyNewClass(TNode n)
 {
+  if (n.getType().isBag())
```
ajreynol, line 224: I don't think you actually need this method, right?  This is doing moditifcations to the state which should be SAT-context-dependent, but it we are not using SAT-context-dependent data structures.  This could lead to unexpected behavior.
I would delete the body for now.

src/theory/bags/solver_state.h
```diff
@@ -31,24 +31,43 @@ class SolverState : public TheoryState
...
-  std::set<Node>& getElements(TypeNode t);
-
-  std::map<Node, Node>& getBagElements(Node B);
+  /**
+   * @param n has the form (bag.count e A)
+   * @pre bag A needs is already registered using registerBag(A)
+   * @return a unique skolem for (bag.count e A)
+   */
+  void registerCountTerm(TNode n);
+  /** get all bag terms */
```
ajreynol, line 42: Can you be a bit more precise?  Is this the set of representatives? Or the set of all bag terms?  

Also, is this call only valid after the state has been initialized on a FULL effort check?

src/theory/bags/bag_solver.h
```diff
@@ -43,12 +43,21 @@ class BagSolver
  private:
   /** apply inference rules for MK_BAG operator */
   void checkMkBag(const Node& n);
+  /**
```
ajreynol, line 46: I think I see the purpose of this method, you are doing both upwards and downwards closure?
Maybe specify this.

It's also important to be clear about what "known elements in" means, maybe give a pointer to the documnetation of SolverState which explains this.

ajreynol (changes requested):

src/theory/bags/solver_state.h
mudathirmahgoub: You are right. This is an old documentation. Nonetheless  I think eventually x would be in A in the next postCheck call if there is one. 
I updated the documentation.

src/theory/bags/solver_state.h
mudathirmahgoub: If you mean by representative the node returned by `EqualityEngine::getRepresentative` then no. 
It is the node returned by `TheoryState::getRepresentative`. I am not sure if you can call this a real representative. 
The current code fails if the equality engine version is used because new skolem variables are not asserted yet to the equality engine. I thought this is inefficient and needs to be addressed later. So to answer your question,  I guess this map may store a count term for a bag that is not a representative yet in the equality engine.

src/theory/bags/theory_bags.cpp
mudathirmahgoub: Done. I missed this old code which has no effect because the solver state is reset in each postCheck method

src/theory/bags/solver_state.h
mudathirmahgoub: Done. `SolverState::getBags()` returns only representatives in the equality engine. The function is only valid after initialization during `postCheck`.

src/theory/bags/bag_solver.h
mudathirmahgoub: Done. 
I am not sure what upwards and downwards mean here. For example suppose we have terms `n=(bag x c), (bag.count x n)
  (bag.count y n)`. Does downwards here mean `x` and upwards mean `y`?
  Also see this new [commit](https://github.com/CVC4/CVC4/pull/5783/commits/41e97e95b3948a6c8e1d25c5fc7c70d979f5022a)

ajreynol: Downwards closure means that from (count x (op A B)), you must reason about (count x A) and (count x B).

Upwards closure means that from (count x A) or (count x B), you must reason about (count x (op A B)).

ajreynol (approved):
LGTM

PR #5784, opened 2021-01-19
Implement proofs for arith BRAB lemmas

alex-ozdemir:
All changes:
* Add a Pf type alias for std::shared_ptr<ProofNode> to
  expr/proof_rule.h
* Add an eager proof generator to TheoryArith for preprocessing
  rewrites. Right now those are proven with INT_TRUST. Will eventually
  fix.
* Generate proved lemmas in TheoryArithPrivate::branchIntegerVariable.
* Same for TheoryArithPrivate::roundRobinBranch
* Add EagerProofGenerator::mkTrustedRewrite.
* Add some proofsEnabled methods.

### DIFF
--- a/src/theory/arith/theory_arith.h
+++ b/src/theory/arith/theory_arith.h
@@ -45,6 +45,10 @@ class TheoryArith : public Theory {
 
   TimerStat d_ppRewriteTimer;
 
+  // Used to prove pp-rewrites
+  EagerProofGenerator d_ppPfGen;
+  ProofNodeManager* d_pnm;
+
  public:
   TheoryArith(context::Context* c,
               context::UserContext* u,
@@ -152,6 +156,11 @@ class TheoryArith : public Theory {
   ArithPreprocess d_arithPreproc;
   /** The theory rewriter for this theory. */
   ArithRewriter d_rewriter;
+
+  /**
+   * Proofs enabled?
+   */
+  bool proofsEnabled() const;
 };/* class TheoryArith */
 
 }/* CVC4::theory::arith namespace */
--- a/src/theory/arith/theory_arith.h
+++ b/src/theory/arith/theory_arith.h
@@ -47,7 +47,6 @@ class TheoryArith : public Theory {
 
   /** Used to prove pp-rewrites */
   EagerProofGenerator d_ppPfGen;
-  ProofNodeManager* d_pnm;
 
  public:
   TheoryArith(context::Context* c,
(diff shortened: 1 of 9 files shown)


src/theory/arith/theory_arith.h
```diff
@@ -45,6 +45,10 @@ class TheoryArith : public Theory {
 
   TimerStat d_ppRewriteTimer;
 
+  // Used to prove pp-rewrites
```
ajreynol, line 48:
```suggestion
  /** Used to prove pp-rewrites */
```

src/theory/arith/theory_arith.h
```diff
@@ -45,6 +45,10 @@ class TheoryArith : public Theory {
...
+  EagerProofGenerator d_ppPfGen;
+  ProofNodeManager* d_pnm;
```
ajreynol, line 50: The theory already has this pointer in the base class: https://github.com/CVC4/CVC4/blob/master/src/theory/theory.h#L237.

src/theory/arith/theory_arith.h
```diff
@@ -152,6 +156,11 @@ class TheoryArith : public Theory {
   ArithPreprocess d_arithPreproc;
   /** The theory rewriter for this theory. */
   ArithRewriter d_rewriter;
+
+  /**
+   * Proofs enabled?
+   */
+  bool proofsEnabled() const;
```
ajreynol, line 163: Maybe we should generalize this so that it is in `theory.h`.

src/theory/arith/theory_arith_private.cpp
```diff
@@ -3691,38 +3693,102 @@ Node TheoryArithPrivate::branchIntegerVariable(ArithVar x) const {
...
+    Trace("integers") << "eq: " << eq << "\nto: " << literal << endl;
     d_containing.getOutputChannel().requirePhase(literal, true);
-    lem = nm->mkNode(kind::OR, literal, lem);
+    Node l = nm->mkNode(kind::OR, literal, right);
+    Trace("integers") << "l: " << l << endl;
+    if (proofsEnabled())
+    {
+      Node less = nm->mkNode(kind::LT, var, mkRationalNode(nearest));
+      Node greater = nm->mkNode(kind::GT, var, mkRationalNode(nearest));
+      // TODO: justify. Thread proofs through *ensureLiteral*.
```
ajreynol, line 3716:
```suggestion
      // TODO (project #37): justify. Thread proofs through *ensureLiteral*.
```

ajreynol (changes requested):
Thanks for adding this, some minor comments.

src/theory/arith/theory_arith.h
alex-ozdemir: Ah. Good point! Added to the `Theory` class in b9332110c.

src/theory/arith/theory_arith.h
alex-ozdemir: Ah. Good point! Removed in b9332110c.

ajreynol (approved):

PR #5788, opened 2021-01-19
Fix type issues with relevant domain computation

### LINKED ISSUE #5635: (full-saturate-quant) Assertion error at /src/theory/quantifiers/inst_strategy_enumerative.cpp:317
cvc4 1eac626e6c23
~~~~
(assert (forall ((q36 Real)) (exists ((q37 (Array Real (Array Bool Real)))) (exists ((q38 (Array Real (Array Bool Real)))) (xor (= q37 q38) (> q36 0))))))
(check-sat)
~~~~
~~~~
./cvc4 -i -q  ../delta.out.smt2 
Fatal failure within bool CVC4::theory::quantifiers::InstStrategyEnum::process(CVC4::Node, bool, bool) at /home/CVC4/src/theory/quantifiers/inst_strategy_enumerative.cpp:317
Check failure

 terms[i].isNull() || terms[i].getType().isComparableTo(ftypes[i])

Aborted
~~~~

ajreynol:
This fixes 2 issues with relevant domain type computation, the first involving arithmetic INST_CONSTANT that do not belong to the current quantified formula being solved for in a monomial sum, the second involving parametric datatype selectors.

Fixes #5635.  Both benchmarks on that issue are unsolved (one timeout, one unknown) but throw no assertion failure.


HanielB (approved):
LGTM

PR #5798, opened 2021-01-22
Add interface for getting preprocessed term

ajreynol:
Several places, e.g. in quantifiers, requiring knowing what the theory-preprocessed form of a node is.

This is required for an improvement to our E-matching algorithm, which requires knowing what the preprocessed form of ground subterms of triggers are.

Note that I'm not 100% happy with adding a new interface to Valuation, but at the moment I don't see a better way of doing this.  On the positive side, this interface will make a few other things (e.g. the return value of OutputChannel::lemma) obsolete.


HanielB (approved):
LGTM

PR #5808, opened 2021-01-24
Ensure uses of ground terms in triggers are preprocessed and registered

ajreynol:
This ensures ground terms in triggers are preprocessed and registered in the master equality engine.  This avoids cases where our E-matching algorithm is incomplete where it should not be.

Positive impact (+222-69) on SMT-LIB, 30 second timeout:

```
                           CVC4-011421_d   62411   18850.9    1491   151.309     63902     87041
                    CVC4-012121-2-ppgt_d   62564   17895.7    1491   151.106     64055     87041



For unsatisfiable : 
                           CVC4-011421_d       X      69
                    CVC4-012121-2-ppgt_d     222       X

For satisfiable : 
                           CVC4-011421_d       X       0
                    CVC4-012121-2-ppgt_d       0       X

```


src/theory/quantifiers/ematching/trigger.cpp
```diff
@@ -996,6 +1023,69 @@ int Trigger::getActiveScore() {
...
+      if (cur.getNumChildren() == 0)
+      {
+        visited[cur] = cur;
+      }
+      else if (!quantifiers::TermUtil::hasInstConstAttr(cur))
+      {
+        Node vcur = val.getPreprocessedTerm(cur);
+        gts.push_back(vcur);
+        visited[cur] = vcur;
+      }
```
HanielB, line 1052: Can you document what is going on here? In particular why we know that `cur` is ground?

HanielB (approved):
LGTM, just one clarification. The results are nice. :)

Issue #5809, opened 2021-01-25
CVC4 Crash (core dumped)


farif:
**Describe the bug**
A bug (core-dump crash) is triggered and crash CVC4 when a simple specification is provided that use AUFLIA logic.    
The bug is triggered with develop version.

**Command line arguments**: --lang=sygus2
**CVC4 version/commit**: This is CVC4 version 1.9-prerelease [git master e0dfc0a3]
**Operating system**: Ubuntu 20.04.1 LTS

The specification file is attached that triggers the bug:
fatal failure within CVC4::api::Sort CVC4::SymbolTable::Implementation::lookupType(const string&, const std::vector<CVC4::api::Sort>&) const at /home/farif/Documents/projects/CVC4/src/expr/symbol_table.cpp:535
Unhandled case encounteredCould not instantiate sort
Aborted (core dumped)
farif@farif-machine:~_

farif:
```
lang=sygus2
(set-logic ALIA)
(set-option :sygus-out status-and-def)
(set-option :e-matching false)
(define-sort Stream (T) (Array Int T))
(define-sort IntStream () (Stream Int))
(synth-fun phi ((t IntStream) ) IntStream
   ((<F> IntStream))
   ((<F> IntStream (
       ((as const IntStream) 0)
    ))
   )
)
(check-synth)

```

daniel-larraz:
FYI, it works fine with CVC4 1.8.

farif:
The above issue is reported for the following version:
This is CVC4 version 1.9-prerelease [git master e0dfc0a3]
compiled with GCC version 9.3.0
on Jan  4 2021 18:09:40

HanielB:
Fixed in #5810

PR #5810, opened 2021-01-25
Reestablishing support for define-sort

HanielB:
Presumable broken since [3ed42d7ab](https://github.com/CVC4/CVC4/commit/3ed42d7ab). This extends the API to have a `substitute` method for `Sort` that in needed for doing the `Sort` substitution in the case of `define-sort`.

This fixes issue #5809.


ajreynol (changes requested):
Looks good, thanks for fixing this.  Need to add new unit tests for the methods in https://github.com/CVC4/CVC4/blob/master/test/unit/api/sort_black.cpp

test/unit/api/solver_black.cpp
```diff
@@ -965,6 +965,15 @@ TEST_F(TestApiSolverBlack, declareSort)
   ASSERT_NO_THROW(d_solver.declareSort("", 2));
 }
 
+TEST_F(TestApiSolverBlack, defineSort)
```
ajreynol, line 968: Should test both versions of substitute :)

ajreynol (approved):

PR #5812, opened 2021-01-25
Add proofs to TheoryArithPrivate::propagate

alex-ozdemir:
Specifically, add proofs to conflicts between (a) a propagation from the
congruence manager and (b) a constraint from the main solver.

### DIFF
--- a/src/theory/arith/theory_arith_private.cpp
+++ b/src/theory/arith/theory_arith_private.cpp
@@ -4028,12 +4028,54 @@ void TheoryArithPrivate::propagate(Theory::Effort e) {
 
       outputPropagate(toProp);
     }else if(constraint->negationHasProof()){
-      Node exp = d_congruenceManager.explain(toProp).getNode();
-      Node notNormalized = normalized.getKind() == NOT ?
-        normalized[0] : normalized.notNode();
-      Node lp = flattenAnd(exp.andNode(notNormalized));
+      // The congruence manager can prove: antecedents => toProp,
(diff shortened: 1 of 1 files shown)


src/theory/arith/theory_arith_private.cpp
```diff
@@ -4028,12 +4028,54 @@ void TheoryArithPrivate::propagate(Theory::Effort e) {
...
-      Node notNormalized = normalized.getKind() == NOT ?
-        normalized[0] : normalized.notNode();
-      Node lp = flattenAnd(exp.andNode(notNormalized));
+      // The congruence manager can prove: antecedents => toProp,
+      // ergo. antecedents ^ ~toProp is a conflict.
+      TrustNode exp = d_congruenceManager.explain(toProp);
+      Node notNormalized = normalized.negate();
+      std::vector<Node> ants(exp.getNode().begin(), exp.getNode().end());
+      ants.push_back(notNormalized);
+      Node lp = safeConstructNary(kind::AND, ants);
```
alex-ozdemir, line 4037: Note: I made a slight change to the conflict here.
Before we were proving `flattenAnd(antecedents ^ notNormalized)`.
Now were are proving an and of the enumerated antecendents, with `notNormalized` added on.
The difference is that if there is a conjunction in the antecdent list, it is no longer flattened.
This makes proofs easier, and I don't think it effects solving much.

ajreynol (approved):
LGTM

PR #5818, opened 2021-01-26
Thread proofs through arith channels & similar

alex-ozdemir:
Thread proofs through conflict/lemma storages & channels in TheoryArithPrivate
* d_acTmp
* d_approxCuts
* lemma output channel
* conflict output channel


src/theory/arith/theory_arith_private.h
```diff
@@ -713,8 +713,8 @@ class TheoryArithPrivate {
     return (d_containing.d_valuation).getSatValue(n);
   }
 
-  context::CDQueue<Node> d_approxCuts;
-  std::vector<Node> d_acTmp;
+  context::CDQueue<TrustNode> d_approxCuts;
```
ajreynol, line 716: Can you add some quick doc on these members?

ajreynol (changes requested):
One minor doc request

src/theory/arith/theory_arith_private.h
alex-ozdemir: I have rather limited understanding of how these are used, but I documented the (very little) that I know in aa52080e0

ajreynol: No worries :)

ajreynol (approved):

PR #5819, opened 2021-01-26
Simplify lemma interface

ajreynol:
This makes it so that TheoryEngine::lemma returns `void` not `LemmaStatus`.

Currently, there was only one use of LemmaStatus by theory solvers, which was CEGQI using it as a way of getting the preprocessed form of a lemma.  This makes it so that there is an explicit method in Valuation for getting the preprocessed form of a term + its skolems and their definition assertions.

It also simplifies a few things, e.g. Valuation calls are forwarded to PropEngine instead of going through TheoryEngine.  It fixes a few issues in TermFormulaRemoval related to getSkolems.


src/smt/term_formula_removal.cpp
```diff
@@ -517,23 +517,13 @@ Node RemoveTermFormulas::getSkolemForNode(Node k) const
...
-    {
-      return false;
-    }
-  }
-  // otherwise, traverse it
-  bool ret = false;
   std::unordered_set<TNode, TNodeHashFunction> visited;
   std::unordered_set<TNode, TNodeHashFunction>::iterator it;
+  context::CDInsertHashMap<Node, theory::TrustNode, NodeHashFunction>::
+      const_iterator itl;
```
HanielB, line 526: Why declare this? In the usage below you could directly compare with `d_lemmaCache.find(cur)`

src/prop/theory_proxy.h
```diff
@@ -118,6 +118,14 @@ class TheoryProxy : public Registrar
   theory::TrustNode removeItes(TNode node,
                                std::vector<theory::TrustNode>& newLemmas,
                                std::vector<Node>& newSkolems);
+  /**
+   * Get the skolems within node and their corresponding definitions, store
+   * them in skAsserts and sks respectively. Note that this method does not
```
HanielB, line 123: The order is inverted.

src/prop/theory_proxy.h
```diff
@@ -118,6 +118,14 @@ class TheoryProxy : public Registrar
...
+   * traverse to include skolems that occur in skAsserts.
```
HanielB, line 124: This is a bit confusing. I imagine that `skAsserts` are not in `node`, so one would not expect them to be traversed anyway. The disclaimer might be accompanied of this clarification.

src/prop/prop_engine.h
```diff
@@ -209,6 +208,20 @@ class PropEngine
    * if preprocessing n involves introducing new skolems.
    */
   Node getPreprocessedTerm(TNode n);
+  /**
+   * Same as above, but also compute the skolems in n and the lemmas
+   * corresponding to their definition.
```
HanielB, line 213:
```suggestion
   * Same as above, but also compute the skolems in n and in the lemmas
   * corresponding to their definition.
```

src/prop/prop_engine.cpp
```diff
@@ -453,6 +441,39 @@ Node PropEngine::getPreprocessedTerm(TNode n)
...
+  std::vector<theory::TrustNode> toProcessAsserts;
+  std::vector<Node> toProcess;
+  d_theoryProxy->getSkolems(pn, toProcessAsserts, toProcess);
+  size_t index = 0;
+  // until fixed point is reached
+  while (index < toProcess.size())
+  {
+    theory::TrustNode ka = toProcessAsserts[index];
+    Node k = toProcess[index];
+    index++;
```
HanielB, line 460:
```suggestion
    Node k = toProcess[index++];
```

src/theory/quantifiers/cegqi/inst_strategy_cegqi.cpp
```diff
@@ -373,13 +373,20 @@ void InstStrategyCegqi::registerCounterexampleLemma(Node q, Node lem)
...
-  LemmaStatus status =
-      d_quantEngine->getOutputChannel().lemma(lem, LemmaProperty::PREPROCESS);
-  Node ppLem = status.getRewrittenLemma();
+  // send the lemma
+  d_quantEngine->getOutputChannel().lemma(lem, LemmaProperty::PREPROCESS);
+  // get the preprocessed form of the lemma we just sent
+  std::vector<Node> skolems;
+  std::vector<Node> skAsserts;
+  Node ppLem = d_quantEngine->getValuation().getPreprocessedTerm(
+      lem, skAsserts, skolems);
```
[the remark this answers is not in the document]
ajreynol: I don't think this works since the third argument is not a const-vector.

src/prop/prop_engine.cpp
ajreynol: I prefer 2 separate lines, it's confusing to me whether the index is incremented before or after the index lookup.

src/smt/term_formula_removal.cpp
ajreynol: Ill simplify, this was leftover from another refactor.

src/theory/quantifiers/cegqi/inst_strategy_cegqi.cpp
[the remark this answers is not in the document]
HanielB: You're right. Pity :)

PR #5822, opened 2021-01-26
Use standard conflict mechanism in quantifiers state

ajreynol:
Work towards eliminating dependencies on quantifiers engine, this updates quantifiers module to use the standard SAT-context dependent flag in quantifiers state instead of the one in QuantifiersEngine.  It also eliminates the use of a custom call to theoryEngineNeedsCheck.

### DIFF
--- a/src/theory/quantifiers/quant_util.h
+++ b/src/theory/quantifiers/quant_util.h
@@ -156,6 +156,10 @@ class QuantifiersModule {
   quantifiers::TermDb* getTermDatabase() const;
   /** get currently used term utility object */
   quantifiers::TermUtil* getTermUtil() const;
+  /** get the quantifiers state */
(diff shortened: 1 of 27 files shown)


src/theory/quantifiers/quant_util.h
```diff
@@ -156,6 +156,10 @@ class QuantifiersModule {
   quantifiers::TermDb* getTermDatabase() const;
   /** get currently used term utility object */
   quantifiers::TermUtil* getTermUtil() const;
+  /** get the quantifiers state */
+  quantifiers::QuantifiersState& getState();
+  /** get the quantifiers state */
```
mpreiner, line 161:
```suggestion
  /** get the quantifiers inference manager */
```

mpreiner (approved):
LGTM, one minor typo

PR #5824, opened 2021-01-27
Use standard equality engine information in quantifiers state

ajreynol:
This refactors quantifiers so that it uses the standard interfaces for setting up an equality engine and using it via TheoryState.

This eliminates the need for several special interfaces including `getMasterEqualityEngine`, `CombinationEngine::getCoreEqualityEngine`, and most uses of `EqualityQuery`.


src/theory/ee_manager_distributed.cpp
```diff
@@ -66,45 +79,31 @@ void EqEngineManagerDistributed::initializeTheories()
       // theory said it doesn't need an equality engine, skip
       continue;
     }
+    if (esi.d_useMaster)
+    {
+      // they said they want to use the master equality engine
```
mpreiner, line 84: Who are "they"? :-)

src/theory/quantifiers/term_database.h
```diff
@@ -204,59 +191,53 @@ class TermDb : public QuantifiersUtil {
...
-  * Wrt evaluateTerm, this version does not construct new terms, and
-  * thus is less aggressive.
-  */
-  TNode getEntailedTerm(TNode n,
-                        std::map<TNode, TNode>& subs,
-                        bool subsRep,
-                        EqualityQuery* qy = NULL);
+   * If possible, returns a term n' such that:
+   * (1) n' exists in the current equality engine (as specified by the state),
+   * (2) n * subs = n' is entailed in the current context, where * is denotes
```
mpreiner, line 214:
```suggestion
   * (2) n * subs = n' is entailed in the current context, where * denotes
```

mpreiner (approved):
LGTM, two minor comments.

PR #5825, opened 2021-01-27
Add option for whether to filter candidate rewrite pairs

ajreynol:
This option should be disabled for the new sygus reconstruction algorithm on https://github.com/CVC4/CVC4/pull/5779.


abdoo8080 (approved):
LGTM!

PR #5826, opened 2021-01-28
Remove regex header from cvc4cpp.cpp

mudathirmahgoub:
This PR replaces the heavy hammer  std::regex_match for checking numbers with string operations

### DIFF
--- a/test/unit/api/term_black.cpp
+++ b/test/unit/api/term_black.cpp
@@ -685,14 +685,16 @@ TEST_F(TestApiTermBlack, getInteger)
   Term int10 = d_solver.mkInteger("18446744073709551615");
   Term int11 = d_solver.mkInteger("18446744073709551616");
   Term int12 = d_solver.mkInteger("-0");
-  Term int13 = d_solver.mkInteger("012");
-  Term int14 = d_solver.mkInteger("0000");
 
   ASSERT_THROW(d_solver.mkInteger(""), CVC4ApiException);
   ASSERT_THROW(d_solver.mkInteger("-"), CVC4ApiException);
   ASSERT_THROW(d_solver.mkInteger("-1-"), CVC4ApiException);
   ASSERT_THROW(d_solver.mkInteger("0.0"), CVC4ApiException);
   ASSERT_THROW(d_solver.mkInteger("-0.1"), CVC4ApiException);
+  ASSERT_THROW(d_solver.mkInteger("012"), CVC4ApiException);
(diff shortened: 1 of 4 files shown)


test/unit/api/term_black.cpp
```diff
@@ -684,6 +684,13 @@ TEST_F(TestApiTermBlack, getInteger)
   Term int9 = d_solver.mkInteger("4294967296");
   Term int10 = d_solver.mkInteger("18446744073709551615");
   Term int11 = d_solver.mkInteger("18446744073709551616");
+  Term int12 = d_solver.mkInteger("-0");
+
+  ASSERT_THROW(d_solver.mkInteger(""), CVC4ApiException);
+  ASSERT_THROW(d_solver.mkInteger("-"), CVC4ApiException);
+  ASSERT_THROW(d_solver.mkInteger("-1-"), CVC4ApiException);
+  ASSERT_THROW(d_solver.mkInteger("0.0"), CVC4ApiException);
+  ASSERT_THROW(d_solver.mkInteger("-0.1"), CVC4ApiException);
```
mpreiner, line 693: What should happen with "-0" or leading zeroes like "012", "0000"?

mudathirmahgoub: Not sure the current API accepts leading zeros without problems. The smt document says this for numerals
```
Numerals. A〈numeral〉is the digit 0 or a non-empty sequence of digits not starting with 0.
```

mpreiner: OK then we should probably be more strict and not allow "01", "00", "-0",...

mudathirmahgoub: We can do that and also strictly follow the SMTLIB standard in the API
For example `(-1)` is not allowed, whereas `(- 1)` is allowed. Should we also apply this to the the API as well?

```java
Term one = d_solver.mkInteger("1");
Term negativeOne = d_solver.mkTerm(UMINUS, one);
```

```java
Term negativeOne = d_solver.mkInteger("-1")
```

mpreiner: No that should both be possible, that's a usability question of `mkInteger`. The API can allow more than what SMT-LIB allows.

mudathirmahgoub: Ok. After this PR leading zeros would be unsupported.

mpreiner (approved):
LGTM, thanks for the changes!

PR #5827, opened 2021-01-28
Eliminate PREPROCESS lemma property

ajreynol:
This is now possible since we always preprocess lemmas.

Note that the LemmaProperty on inferences may be redundant throughout the non-linear solver now.


src/theory/fp/theory_fp.cpp
```diff
@@ -916,8 +916,9 @@ void TheoryFp::preRegisterTerm(TNode node)
 
 void TheoryFp::handleLemma(Node node) {
   Trace("fp") << "TheoryFp::handleLemma(): asserting " << node << std::endl;
-  // Preprocess has to be true because it contains embedded ITEs
-  d_out->lemma(node, LemmaProperty::PREPROCESS);
+  // will be preprocessed went sent, which is important because it contains
```
nafur, line 919:
```suggestion
  // will be preprocessed when sent, which is important because it contains
```

nafur (approved):
Just one typo.

PR #5828, opened 2021-01-28
Reorganize calls to quantifiers engine from SmtEngine layer

ajreynol:
This reorganizes calls to QuantifiersEngine from SmtEngine.  Our policy has changed recently that the SmtEngine layer is allowed to make calls directly to QuantifiersEngine, which eliminates the need for TheoryEngine to have forwarding calls.  This PR makes this policy more consistent.

This also makes quantifier-specific calls more safe by throwing modal exceptions in the cases where quantifiers are present.

Marking "major" since we currently segfault when e.g. dumping instantiations in non-quantified logics.  This PR makes it so that we throw a modal exception.


HanielB (approved):
LGTM

PR #5832, opened 2021-01-29
Add bag inferences for operators: intersection, duplicate_removal, and empty bags

mudathirmahgoub:
This PR adds inferences for operators: intersection, duplicate_removal, and empty bags during post check. 
It also fixes a bug in SolverState::getElements


ajreynol (changes requested):
Can you revert your changes to theory_engine.cpp?  It seems to have been unintentionally formatted.

mudathirmahgoub:
> Can you revert your changes to theory_engine.cpp? It seems to have been unintentionally formatted.

Yes. I didn't realize it until I made the pull request. It is tedious to revert it back with other changes

src/theory/bags/solver_state.h
```diff
@@ -77,6 +85,8 @@ class SolverState : public TheoryState
   std::set<Node> d_bags;
   /** bag -> associated elements */
   std::map<Node, std::set<Node>> d_bagElements;
+  /** Disequal bag terms */
+  std::set<Node> d_deq;
```
ajreynol, line 89: `std::vector` is better for this (its faster).  Same applies for d_bags.

ajreynol (changes requested):
One minor request, otherwise looks good to me.

ajreynol (approved):
Discussed offline, LGTM

PR #5833, opened 2021-01-29
(proof-new) Distinguish pre vs post rewrites in term conversion proof generator

ajreynol:
This is work towards resolving two kinds of failures on proof-new:

(1) Functional issues with proofs from the rewriter, in particular when a term pre-rewrites and post-rewrites to different things,
(2) Conversion issues in theory-preprocessing, where all steps are assumed to be post-rewrites but some are in fact pre-rewrites.  This leads to the term conversion proof generator proving something different than what is expected.

A followup PR will simplify and fix proofs for theory-preprocessing.

### DIFF
--- a/src/theory/rewriter.cpp
+++ b/src/theory/rewriter.cpp
@@ -382,6 +382,8 @@ Node Rewriter::rewriteTo(theory::TheoryId theoryId,
                 << "with proofs: " << rewriteStackTop.d_node << std::endl;
             Trace("rewriter-proof") << " w/o proofs: " << cached << std::endl;
             Node eq = rewriteStackTop.d_node.eqNode(cached);
+            // we make this a post-rewrite, since we are processing a node that
+            // has finished post-rewriting above
             tcpg->addRewriteStep(rewriteStackTop.d_node,
                                  cached,
                                  PfRule::TRUST_REWRITE,
--- a/src/theory/rewriter.cpp
+++ b/src/theory/rewriter.cpp
@@ -386,7 +386,8 @@ Node Rewriter::rewriteTo(theory::TheoryId theoryId,
                                  cached,
                                  PfRule::TRUST_REWRITE,
                                  {},
-                                 {eq});
+                                 {eq},
+                                 false);
             // don't overwrite the cache, should be the same
             rewriteStackTop.d_node = cached;
           }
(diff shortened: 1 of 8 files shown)


src/theory/theory_preprocessor.cpp
```diff
@@ -431,7 +432,7 @@ Node TheoryPreprocessor::rewriteWithProof(Node term)
       Trace("tpp-debug") << "TheoryPreprocessor: addRewriteStep (rewriting) "
                          << term << " -> " << termr << std::endl;
       // always use term context hash 0 (default)
-      d_tpg->addRewriteStep(term, termr, PfRule::REWRITE, {}, {term});
+      d_tpg->addRewriteStep(term, termr, PfRule::REWRITE, {}, {term}, false);
```
HanielB, line 435: ditto

src/theory/theory_preprocessor.cpp
```diff
@@ -480,15 +482,19 @@ Node TheoryPreprocessor::preprocessWithProof(Node term)
...
                          << term << " -> " << termr << std::endl;
       // small step trust
-      d_tpg->addRewriteStep(
-          term, termr, PfRule::THEORY_PREPROCESS, {}, {term.eqNode(termr)});
+      d_tpg->addRewriteStep(term,
+                            termr,
+                            PfRule::THEORY_PREPROCESS,
+                            {},
+                            {term.eqNode(termr)},
+                            false);
```
HanielB, line 497: ditto

src/theory/rewriter.cpp
```diff
@@ -386,7 +386,8 @@ Node Rewriter::rewriteTo(theory::TheoryId theoryId,
                                  cached,
                                  PfRule::TRUST_REWRITE,
                                  {},
-                                 {eq});
+                                 {eq},
+                                 false);
```
HanielB, line 390: that last argument is the default. Moreover, also maybe add some documentation to say why this is a post-rewrite? Does not seem so clear to me.

HanielB (approved):
LGTM! A few requests for doc but that's it.

I'm grad that in the end it was not a very invasive change. :)

src/theory/theory_preprocessor.cpp
ajreynol: This "false" is wrong (see FIXME) :)

src/theory/theory_preprocessor.cpp
ajreynol: Ditto

src/theory/rewriter.cpp
ajreynol: Will add

PR #5834, opened 2021-01-29
[proof-new] Connecting new unsat cores

HanielB:
Allows one to generate unsat cores from the new proof infrastructure. For new this is controlled by a new option, `--check-unsat-cores-new`.


ajreynol (approved):
LGTM

PR #5835, opened 2021-01-29
(proof-new) Refactor theory preprocessing

ajreynol:
This simplifies our algorithm for theory preprocessing. The motivation is two-fold:
(1) Proofs are currently incorrect in rare cases due to incorrectly tracking pre vs post rewrites.
(2) We can avoid traversing Boolean connectives with term formula removal.  Note that this PR relies on the observation that term formula removal leaves Boolean structure unchanged and can apply locally to theory atoms.

Summarizing the change in algorithms:
```

============  OLD
[1]
TRAVERSE(
  prerewrite: 
    if theory atom {
      TRAVERSE(
        prerewrite:
          apply rewriter
        postrewrite:
          apply rewriter
          apply ppRewrite
            if changed
              apply rewriter
              REPEAT traversal
      )
    }
  postrewrite:  // for Boolean connectives
    apply rewriter
)
[2]
apply term formula removal
[3]
apply rewriter

============  NEW
[1] 
apply rewriter
[2] 
TRAVERSE(
  prerewrite:
    if theory atom {
      TRAVERSE(
        prerewrite:
          // nothing
        postrewrite:
          apply rewriter
          apply ppRewrite
            if changed
              apply rewriter
              REPEAT traversal
      )
      apply term formula removal
      apply rewriter
    }
  postrewrite: // for Boolean connectives
    apply rewriter
)
```

Furthermore, notice that our handling of lemmas generated by term formula removal is now more uniform.  In the second algorithm, term formula removal is *not* applied to fixed point anymore.  In other words, we do not remove skolems from lemmas generated by term formula removal recursively.  Instead, since lemmas are theory-preprocessed recursively, we use the fixed point that runs outside the above algorithm instead of relying on the inner fixed point in term formula removal.

This PR resolves several open issues with proof production on proof-new.

I will performance test this change on SMT-LIB.


ajreynol:
Slightly slower on QF_LIA:
```

                           Configuration	#unsat	time	#sat	time	#solved	#total
                          CVC4-020121_dj	1912	104396	2939	101869	4851	6947
                     CVC4-020121-tppr_dj	1897	116935	2925	103842	4822	6947
						
unsatisfiable						
                          CVC4-020121_dj	X	16				
                     CVC4-020121-tppr_dj	1	X				
satisfiable						
                          CVC4-020121_dj	X	15				
                     CVC4-020121-tppr_dj	1	X				


```

This is due to rewriting being a separate step beforehand, which leads to 15% overhead or so on nec benchmarks it seems.

ajreynol:
Opened https://github.com/CVC4/cvc4-wishues/issues/79 to track performance issue.

src/theory/theory_preprocessor.h
```diff
@@ -38,6 +38,33 @@ namespace theory {
...
+ *              REPEAT traversal
+ *      )
+ *      apply term formula removal
+ *      apply rewriter
+ *    }
+ *  postrewrite: // for Boolean connectives
+ *    apply rewriter
+ * )
+ *
+ * Note that the rewriter must be applied beforehand, since
```
HanielB, line 67: Say why :)

src/theory/theory_preprocessor.cpp
```diff
@@ -287,36 +287,55 @@ TrustNode TheoryPreprocessor::theoryPreprocess(TNode assertion)
...
-        << "TheoryPreprocessor::theoryPreprocess(" << assertion
-        << "): processing " << current << endl;
+        << "TheoryPreprocessor::theoryPreprocess processing " << current
+        << endl;
 
     // If node already in the cache we're done, pop from the stack
-    NodeMap::iterator find = d_ppCache.find(current);
-    if (find != d_ppCache.end())
+    NodeMap::iterator find = d_rtfCache.find(current);
+    if (find != d_rtfCache.end())
```
HanielB, line 295:
```suggestion
    if (d_rtfCache.find(current) != d_rtfCache.end())
```

src/theory/theory_preprocessor.cpp
```diff
@@ -356,8 +376,8 @@ TrustNode TheoryPreprocessor::theoryPreprocess(TNode assertion)
              ++child_it)
         {
           TNode childNode = *child_it;
-          NodeMap::iterator childFind = d_ppCache.find(childNode);
-          if (childFind == d_ppCache.end())
+          NodeMap::iterator childFind = d_rtfCache.find(childNode);
+          if (childFind == d_rtfCache.end())
```
HanielB, line 380:
```suggestion
          if (d_rtfCache.find(childNode) == d_rtfCache.end())
```

HanielB (approved):
Intense stuff! This LGTM modulo minor suggestions, given the wishue to deal with the localized performance hit.

PR #5839, opened 2021-01-31
Simplify alpha equivalence module

ajreynol:
This class uses a discrimination tree which can compute subsumption as well as alpha-equivalence.  This class was initially designed to also support subsumption. However, we were only using this class for alpha-equivalence and hence (canonized) Node comparison suffices and the class can be simplified.  It also makes minor cosmetic updates to the module.

If we plan to support subsumption checking later, we will write a new module, and use this class https://github.com/CVC4/CVC4/blob/master/src/expr/match_trie.h which is a more mature version of the class deleted by this PR. 

Note: I verified that the new and old implementation was equivalent using a temporary AlwaysAssert.


HanielB (approved):
LGTM

PR #5842, opened 2021-02-01
Avoid calling the printers while converting sexpr to string.

abdoo8080:
This PR modifies `sexprToString` to use `Term::getString` to get string constants instead of `Term::toString`, which depends on the output language. The previous behavior caused CVC4 to crash when AST is picked as the output language.


ajreynol (approved):
LGTM

PR #5844, opened 2021-02-02
Clarifying documentation of `--static-binary`

yoni206:
The documentation regarding `--static-binary` is improved.


configure.sh
```diff
@@ -29,7 +29,8 @@ General options;
 Features:
 The following flags enable optional features (disable with --no-<option name>).
   --static                 build static libraries and binaries [default=no]
-  --static-binary          enable/disable static binaries
+  --static-binary          force using standard libraries statically
```
mpreiner, line 32:
```suggestion
  --static-binary          statically link against system libraries
```

configure.sh
```diff
@@ -29,7 +29,8 @@ General options;
...
+                           (must be disabled for static macOS builds) [default=yes]
```
mpreiner, line 33: I wonder if we should do some more magic in the configure.sh script to automatically macOS and disable --static-binary.

yoni206: Sounds good. If we do that -- can/should we remove `static-binary` option altogether, making it transparent to the user?

mpreiner: No I'd still keep the option. But since for macOS this can't be supported we can automatically disable it on macOS.

mpreiner (approved):

PR #5845, opened 2021-02-02
[proof-new] Fix bug in expansion of MACRO_RESOLUTION

HanielB:
Evaluating the proof infrastructure in SMT-LIB has uncovered a rare case (i.e., not previously in our regressions!!) in which a crowding literal occurs in an OR node that represents a single-literal clause *subsequent* to the last clause that includes the crowding literal. This was leading to the clause that eliminates the crowding literal not being found. The commit fixes this issue by excluding single-literal clauses from those that can introduce crowding literals (which was already marked as necessary but not properly enforced).


ajreynol (approved):
LGTM

PR #5846, opened 2021-02-02
Improvements for NL traces

ajreynol:
This makes it so that `-t nl-ext` is a concise summary of what the nl-ext solver is doing, which I use frequently for debugging.

This is a temporary solution, we should consider a deeper refactoring of traces throughout NL at some point.


nafur (approved):
LGTM.

PR #5849, opened 2021-02-02
Optimization module

mcjuneho:
Initial implementation of an optimization solver as well as unit tests.


src/smt/optimization_solver.h
```diff
@@ -0,0 +1,119 @@
...
+  OBJECTIVE_MAXIMIZE,
+  OBJECTIVE_UNDEFINED
+};
+
+/**
+ * An enum for optimization queries.
+ *
+ * Represents the result of a checkopt query
+ * (unimplemented) OPT_OPTIMAL: if 
+ * value was found
```
ajreynol, line 45:
```suggestion
 * (unimplemented) OPT_OPTIMAL: if value was found
```

src/smt/optimization_solver.h
```diff
@@ -0,0 +1,119 @@
...
+enum CVC4_PUBLIC OptResult
+{
+  // the original set of assertions has result UNKNOWN
+  OPT_UNKNOWN,
+  // the original set of assertions has result UNSAT
+  OPT_UNSAT,
+  // the optimization loop finished and optimal
+  OPT_OPTIMAL,
+
+  // The last two values are here as a preparation for future work
```
ajreynol, line 56: Can you open and reference a CVC4 wishue on this?

src/smt/optimization_solver.h
```diff
@@ -0,0 +1,119 @@
...
+  /** A getter for d_type **/
+  ObjectiveType getType();
+  /** A getter for d_node **/
+  Node getNode();
+
+ private:
+  /** The type of objective this is, either OBJECTIVE_MAXIMIZE OR
+   * OBJECTIVE_MINIMIZE  **/
+  ObjectiveType d_type;
+  /** The node associated to the term that was used to construc the objective.
```
ajreynol, line 80:
```suggestion
  /** The node associated to the term that was used to construct the objective.
```

src/smt/optimization_solver.h
```diff
@@ -0,0 +1,119 @@
...
+   * **/
+  Node d_node;
+  };
+
+  /**
+   * A solver for optimization queries.
+   *
+   * This class is responsible for responding to get-interpol commands. It
```
ajreynol, line 88: Fix this documentation (seems to be copied from interpolation).

src/smt/optimization_solver.cpp
```diff
@@ -0,0 +1,137 @@
...
+
+  return OPT_OPTIMAL;
+}
+
+void OptimizationSolver::activateObj(const Node& obj, const int& type)
+{
+  d_activatedObjective = Objective(obj, (ObjectiveType)type);
+}
+
+Node OptimizationSolver::objectiveGetValue(const Node& obj)
```
ajreynol, line 122: Does this method need to take obj?  It is not used here.  If this method took no arguments, then we can just assume you are looking up the value from the last query.

ajreynol (changes requested):
The overall design looks pretty good.  I have some initial requests.

src/smt/optimization_solver.h
yoni206: I'll do that. 
Something like this?
https://github.com/CVC4/cvc4-wishues/issues/80

ajreynol: Looks good.

src/smt/optimization_solver.h
```diff
@@ -0,0 +1,118 @@
...
+  // the optimization loop finished and optimal
+  OPT_OPTIMAL,
+
+  // The last two values are here as a preparation for future work
+  // in which partial and approximate optimizations will be supported.
+
+  // if the optimized value is only partial
+  OPT_SAT_PARTIAL,
+  // if the solver halted early and value is only approximate
+  OPT_SAT_APPROX
```
ahmed-irfan, line 61: what would be the difference between `OPT_SAT_PARTIAL` and `OPT_SAT_APPROX`?

src/smt/optimization_solver.cpp
```diff
@@ -0,0 +1,137 @@
...
+  initializeSubsolver(optChecker);
+  NodeManager* nm = optChecker->getNodeManager();
+
+  // we need to be in incremental mode for multiple objectives since we need to
+  // push pop we need to produce models to inrement on our objective
+  optChecker->setOption("incremental", "true");
+  optChecker->setOption("produce-models", "true");
+
+  // Move assertions from the parent solver to the subsolver
+  std::vector<Node> axioms = d_parent->getExpandedAssertions();
```
ahmed-irfan, line 61: we shouldn't be calling them axioms.
Give a better name, like `parent_assertions` or `p_assertions`

src/smt/optimization_solver.h
```diff
@@ -0,0 +1,118 @@
...
+   */
+  class OptimizationSolver
+  {
+   public:
+    /** parent is the smt_solver that the user added their assertions to **/
+    OptimizationSolver(SmtEngine* parent);
+    ~OptimizationSolver();
+
+    /** Runs the optimization loop for the activated objective **/
+    OptResult checkOpt();
```
ahmed-irfan, line 100: can it be called multiple times?

If no, how do we make sure?

If yes, the d_parent needs to be reset.

ahmed-irfan (commented):
Thanks Michael. nice work!

I left some comments, please check them.

src/smt/optimization_solver.h
mcjuneho: As far as I understand, OPT_SAT_PARTIAL is for when the solver times out and the answer is not well-bounded
whereas OPT_SAT_APPROX is for when the answer is well-bounded but either not discrete or not converging for some other reason and we return an approximate value.

src/smt/optimization_solver.h
mcjuneho: It can be called multiple times:
If it's the same parent solver: shouldn't have any problems since the opt_solver resets and adds the assertions again
If it's a different parent solver: each parent solver should have their own opt_solver that they make using the constructor so we shouldn't have to reset the d_parent.

src/smt/optimization_solver.h
ahmed-irfan: hmmm, if the solver times out without a single sat result in the optimization loop then the result shouldn't be OPT_SAT_PARTIAL, instead, it should be OPT_UNKNOWN. if it does find at least one sat result in the optimization loop then the result is OPT_SAT_APPROX. So, I don't see a use case for OPT_SAT_PARTIAL.

mcjuneho: Hmm. I see where you're coming from. 
OPT_UNKOWN is reserved for cases where the underlying sat solver times out IE when OPT_UNKNOWN is returned from the solver in the optimization loop. 
OPT_SAT_PARTIAL is returned if we were to get stuck in our optimization loop for some reason, and want to end it prematurely. The underlying solver could return only SAT, but our solver may never converge in which case we would want to return OPT_SAT_PARTIAL.
please lmk if you want more info and we can set up a time to meet.

yoni206:
May I suggest we just delete these options for now, and bring them back when we have a test that uses them?

ahmed-irfan:
> May I suggest we just delete these options for now, and bring them back when we have a test that uses them?

I agree. 

We should remove OPT_SAT_PARTIAL.

src/smt/optimization_solver.h
ahmed-irfan: In the last case which you mentioned, it should be ok to return OPT_SAT_APPROX.

ajreynol (approved):
LGTM, thanks for the updates.

I guess SMT-LIB parser support will be added next?

ahmed-irfan (approved):
LGTM as well.

Thanks.

mpreiner:
Note: CxxTest was recently removed and all unit tests were migrated to googletest. It should be pretty straightforward to migrate ` test/unit/theory/theory_int_opt_white.h` to googletest.

aniemetz (changes requested):
Requesting changes because of the unit test migration. I want to have a look before it gets merged in.

mpreiner:
The changes to `.github/workflows/ci.yml` should be reverted.

mcjuneho:
> The changes to `.github/workflows/ci.yml` should be reverted.

Worked with Yoni a bit on this, but with the ci.yml file from master I'm getting this error when I push to the optimization_subsolver branch:

 ! [remote rejected]     optimization_subsolver -> optimization_subsolver (refusing to allow a Personal Access Token to create or update workflow `.github/workflows/ci.yml` without `workflow` scope)
error: failed to push some refs to 'https://github.com/mcjuneho/CVC4.git'

But I can push fine when I use an older version of my ci.yml file. Do you have any advice on how to proceed?

Thanks!

mpreiner:
> > The changes to `.github/workflows/ci.yml` should be reverted.
> 
> Worked with Yoni a bit on this, but with the ci.yml file from master I'm getting this error when I push to the optimization_subsolver branch:
> 
> ! [remote rejected] optimization_subsolver -> optimization_subsolver (refusing to allow a Personal Access Token to create or update workflow `.github/workflows/ci.yml` without `workflow` scope)
> error: failed to push some refs to 'https://github.com/mcjuneho/CVC4.git'
> 
> But I can push fine when I use an older version of my ci.yml file. Do you have any advice on how to proceed?
> 
> Thanks!

Try to rebase on current master or merge master.

mcjuneho:
> > > The changes to `.github/workflows/ci.yml` should be reverted.
> > 
> > 
> > Worked with Yoni a bit on this, but with the ci.yml file from master I'm getting this error when I push to the optimization_subsolver branch:
> > ! [remote rejected] optimization_subsolver -> optimization_subsolver (refusing to allow a Personal Access Token to create or update workflow `.github/workflows/ci.yml` without `workflow` scope)
> > error: failed to push some refs to 'https://github.com/mcjuneho/CVC4.git'
> > But I can push fine when I use an older version of my ci.yml file. Do you have any advice on how to proceed?
> > Thanks!
> 
> Try to rebase on current master or merge master. If that doesn't work you can also give me access to your repo and I'll try to fix it.

Thanks for the help! Got it figured out. Some funky stuff with GitHub's access tokens.

test/unit/theory/CMakeLists.txt
```diff
@@ -29,3 +29,4 @@ cvc4_add_unit_test_white(theory_strings_skolem_cache_black theory)
 cvc4_add_unit_test_white(theory_strings_word_white theory)
 cvc4_add_unit_test_white(theory_white theory)
 cvc4_add_unit_test_white(type_enumerator_white theory)
+cvc4_add_unit_test_white(theory_int_opt_white theory)
```
yoni206, line 32: plz move right after `theory_engine_white`.

test/unit/theory/theory_int_opt_white.cpp
```diff
@@ -0,0 +1,137 @@
+#include <iostream>
+#include "api/cvc4cpp.h"
```
yoni206, line 2: plz remove

test/unit/theory/theory_int_opt_white.cpp
```diff
@@ -0,0 +1,137 @@
...
+#include "expr/node_manager.h"
+#include "expr/node_value.h"
+#include "smt/smt_engine.h"
+#include "smt/smt_engine_scope.h"
+#include "test_smt.h"
+
+namespace CVC4 {
+
+using namespace theory;
+using namespace expr;
```
yoni206, line 16: can u remove this?

test/unit/theory/theory_int_opt_white.cpp
```diff
@@ -0,0 +1,137 @@
...
+
+class TestTheoryIntOptWhite : public TestSmt
+{
+ protected:
+  void SetUp() override
+  {
+    TestSmt::SetUp();
+    d_optslv = new OptimizationSolver(d_smtEngine.get());
+
+    //d_slv = new api::Solver();
```
yoni206, line 31: remove plz

test/unit/theory/theory_int_opt_white.cpp
```diff
@@ -0,0 +1,137 @@
+#include <iostream>
+#include "api/cvc4cpp.h"
+#include "smt/optimization_solver.h"
+#include "expr/expr_manager.h"
+#include "expr/node.h"
+#include "expr/node_builder.h"
+#include "expr/node_manager.h"
```
aniemetz, line 7: Remove (included in test_smt.h).

test/unit/theory/theory_int_opt_white.cpp
```diff
@@ -0,0 +1,137 @@
...
+#include "expr/node_value.h"
+#include "smt/smt_engine.h"
+#include "smt/smt_engine_scope.h"
```
aniemetz, line 10: Ditto

test/unit/theory/theory_int_opt_white.cpp
```diff
@@ -0,0 +1,137 @@
+#include <iostream>
+#include "api/cvc4cpp.h"
+#include "smt/optimization_solver.h"
+#include "expr/expr_manager.h"
```
aniemetz, line 4: Remove

test/unit/theory/theory_int_opt_white.cpp
```diff
@@ -0,0 +1,137 @@
...
+
+using namespace theory;
+using namespace expr;
+using namespace context;
+using namespace kind;
+using namespace smt;
+
+namespace test {
+
+class TestTheoryIntOptWhite : public TestSmt
```
aniemetz, line 23:
```suggestion
class TestTheoryWhiteIntOpt : public TestSmt
```

test/unit/theory/theory_int_opt_white.cpp
```diff
@@ -0,0 +1,137 @@
...
+  {
+    TestSmt::SetUp();
+    d_optslv = new OptimizationSolver(d_smtEngine.get());
+
+    //d_slv = new api::Solver();
+    d_smtEngine->setOption("produce-assertions", "true");
+    d_smtEngine->setOption("incremental", "true");
+    d_intType.reset(new TypeNode(d_nodeManager->integerType()));
+  }
+  OptimizationSolver* d_optslv;
```
aniemetz, line 36:
```suggestion
  std::unique_ptr<OptimizationSolver> d_optslv;
```

test/unit/theory/theory_int_opt_white.cpp
```diff
@@ -0,0 +1,137 @@
...
+
+namespace test {
+
+class TestTheoryIntOptWhite : public TestSmt
+{
+ protected:
+  void SetUp() override
+  {
+    TestSmt::SetUp();
+    d_optslv = new OptimizationSolver(d_smtEngine.get());
```
aniemetz, line 29:
```suggestion
    d_optslv.reset(new OptimizationSolver(d_smtEngine.get()));
```

aniemetz (changes requested):

test/unit/theory/theory_int_opt_white.cpp
mcjuneho: Hi Aina, Thanks for the quick review. Just to check, would it be fine to leave in the smt/optimization_solver.h include since I use the class in the test or should I remove it / put the include somewhere else? Thanks again.

aniemetz: Only remove line 4 (including the expr_manager).

test/unit/theory/theory_int_opt_white.cpp
aniemetz: And here both lines 9 and 10.

test/unit/theory/theory_int_opt_white.cpp
aniemetz: And as @yoni206 pointed out, also line 2.

yoni206:
and also the API, no?

test/unit/theory/theory_int_opt_white.cpp
```diff
@@ -0,0 +1,126 @@
...
+#include "test_smt.h"
+
+namespace CVC4 {
+
+using namespace theory;
+using namespace smt;
+
+namespace test {
+
+class TestTheoryWhiteIntOpt : public TestSmt
```
aniemetz, line 12: Now that I think about it, does the test go through? I feel like you would need `TestSmtNoFinishInit` here and then call `d_smtEngine->finishInit()` after setting the options (after line 21).

mcjuneho: Strangely the test passes, but I can definitely change it to TestSmtNoFinishInit and call finishInit after.

test/unit/theory/theory_int_opt_white.cpp
mpreiner: The filename also needs to be renamed then.

test/unit/theory/theory_int_opt_white.cpp
mcjuneho: I've made the change to NoFinishInit. Just as a note: the test broke when I remove the options I needed, so it seems like the options were somehow getting set correctly. Just wanted to let you know in case this behavior is unexpected.

aniemetz: Hm, but "incremental" is not needed? So the while loop in `checkOpt` is not entered at all in the tests? Can you construct a test where the while loop is covered, too? You will have to reenable "incremental" for that.

mcjuneho: I realized that we don't need incremental for the parent solver, we just need it for the subsolver (d_optslv). Since the subsolver sets its own options we don't have to worry about setting incremental for the parent. We do need produce_assertions, however, since that allows us to push the assertions down from the parent to the subsolver.

aniemetz (approved):

PR #5850, opened 2021-02-02
[proof-new] Fix MACRO_RESOLUTION expansion for singleton clause corner case

HanielB:
Evaluating the proof infrastructure in SMT-LIB has uncovered a rare case (i.e., not previously in our regressions!!) in which the resulting clause from crowding literal elimination is a singleton. This commit makes the expansion code robust to that and adds an example of a problematic benchmark as a regression.

Also improves a bit tracing and some comments.


test/regress/regress1/proofs/macro-res-exp-singleton-after-elimCrowd.smt2
```diff
@@ -0,0 +1,384 @@
+; COMMAND-LINE: --proof-new --check-proofs-new
```
ajreynol, line 1: Could we just enable this benchmark without specifically asking for proofs? (We will be testing proofs on all regressions eventually anyways).

The same applies to the regression you added on the earlier PR.

ajreynol (changes requested):

test/regress/regress1/proofs/macro-res-exp-singleton-after-elimCrowd.smt2
HanielB: Indeed! I'll fix.

PR #5856, opened 2021-02-04
Adding an option to optimize polite combination for datatypes

yoni206:
This PR makes the optimization introduced in bbca987e023b2dbf386a62731b94e41c06f32526 optional.


src/options/smt_options.toml
```diff
@@ -11,6 +11,16 @@ header = "options/smt_options.h"
   read_only  = true
   help       = "dump preprocessed assertions, etc., see --dump=help"
 
+[[option]]
```
ajreynol, line 14: Can you call this dtPoliteOptimize dt-polite-optimize?  Also, this should be in datatype_options.

ajreynol (changes requested):

src/options/smt_options.toml
yoni206: done. also made it default to true.

ajreynol (approved):
LGTM, thanks for adding this.

PR #5859, opened 2021-02-05
Do not traverse quantifiers in term formula removal

ajreynol:
Our current policy for term formula removal leaves quantifier bodies unchanged. This simplifies the code based on this observation.


HanielB (approved):
LGTM

PR #5860, opened 2021-02-05
Avoid spurious traversal of terms during preregistration

ajreynol:
This simplifies a few places in the code which unecessarily traverse terms during preregistration (which already traverses terms).


nafur (approved):
LGTM.

PR #5866, opened 2021-02-05
Fix dumping of assertions for monolithic preprocessing passes

HanielB:
Previously we were unable to dump assertions in `ProcessAssertions::apply` for the definition-expansion, simplify and repeat-simplify passes.


HanielB:
Actually there is more to be done for this to work properly. I'll update the PR soon.

ajreynol (approved):

PR #5867, opened 2021-02-05
Use quantifiers inference manager for lemma management

ajreynol:
Towards eliminating dependencies on quantifiers engine.

This replaces the custom implementation of lemma management in quantifiers engine with the standard one.

It makes a few minor changes to the standard inference manager classes to ensure the same behavior for quantifiers.

Note that some minor changes in behavior are introduced in this PR, such as being more consistent about caching/rewriting lemmas.  This should not have any major impact.


src/theory/inference_manager_buffered.h
```diff
@@ -52,10 +52,19 @@ class InferenceManagerBuffered : public TheoryInferenceManager
    * non-null, pg must be able to provide a proof for lem for the remainder
    * of the user context. Pending lemmas are sent to the output channel using
    * doPendingLemmas.
+   *
+   * @param lem The lemma to send
+   * @param p The property of the lemma
+   * @param pg The proof generator which can provide a proof for lem
+   * @param checkCache Whether we want to check that the lemma is already in
+   * the cache.
+   * @return true if the lemma was added to the list of pending lemmas. This
```
mpreiner, line 61:
```suggestion
   * @return true if the lemma was added to the list of pending lemmas and
```

src/theory/inference_manager_buffered.h
```diff
@@ -52,10 +52,19 @@ class InferenceManagerBuffered : public TheoryInferenceManager
...
+   * returns false if the lemma is already cached.
```
mpreiner, line 62:
```suggestion
   * false if the lemma is already cached.
```

src/theory/quantifiers/quant_util.h
```diff
@@ -185,10 +185,11 @@ class QuantifiersUtil {
...
   virtual ~QuantifiersUtil(){}
   /* reset
-  * Called at the beginning of an instantiation round
-  * Returns false if the reset failed. When reset fails, the utility should have
-  * added a lemma via a call to qe->addLemma. TODO: improve this contract #1163
-  */
+   * Called at the beginning of an instantiation round
+   * Returns false if the reset failed. When reset fails, the utility should
+   * have added a lemma via a call to d_qim.addPendingLemma.
+   * TODO: improve this contract #1163
```
mpreiner, line 191: Is this TODO/issue still valid?

ajreynol: No I'll remove.

mpreiner (approved):

PR #5874, opened 2021-02-06
Remove support for inst closure

ajreynol:
This was a feature implemented for "Deciding Local Theory Extensions via E-matching" CAV 2015 that is not used anymore, and will be a burden to maintain further with potential changes to term database.

It also simplifies the TermDatabase::addTerm method (which changed indentation).


HanielB (approved):
LGTM

PR #5876, opened 2021-02-08
Optimize get skolems method

ajreynol:
This method is used to determine which skolems have definitions according to term formula removal.  This optimizes the method using attributes similar to how expr::getFreeVariables works.

It will be used heavily for SAT relevancy. We should also consider updating the justification heuristic to use this method.


HanielB (approved):
LGTM

PR #5879, opened 2021-02-08
[quantifiers] Fix prenex computation

### LINKED ISSUE #5693: (pre-skolem-quant) Assertion error at src/expr/node_algorithm.cpp:378
CVC4 80e0246
~~~~
(set-option :pre-skolem-quant true)
(declare-fun v7 () Bool)
(assert (forall ((q49 Bool) (q55 Bool)) (xor v7 (exists ((q49 Bool)) (xor v7 q49 q49)) v7 (= q55 q49))))
(push)
~~~~
~~~~
./cvc4 -i -q --check-models ../delta.out.smt2 
Fatal failure within bool CVC4::expr::getFreeVariablesScope(CVC4::TNode, std::unordered_set<CVC4::NodeTemplate<true>, CVC4::NodeHashFunction>&, std::unordered_set<CVC4::NodeTemplate<false>, CVC4::TNodeHashFunction>&, bool) at /home/CVC4/src/expr/node_algorithm.cpp:378
Check failure

 scope.find(cn) == scope.end()

Aborted (core dumped)
~~~~

HanielB:
Previously our prenex computation could generate quantifiers of the form `forall x y y. F`, which would lead to an assertion failure in `getFreeVariablesScope`, as it assumes that no shadowing occurs. This commit makes the prenex computation take a set rather than a vector, thus avoiding duplications of prenexed variables. It also changes `mkForall` to take a constant vector, since it does not modify the given vector.

Fixes #5693


ajreynol (approved):
LGTM, thanks for the fix.

test/regress/regress1/quantifiers/prenex-scholl-smt08_RNDPRE_RNDPRE_4_6.smt2
```diff
@@ -0,0 +1,20 @@
+; EXPECT unsat
```
ajreynol, line 1:
```suggestion
; EXPECT: unsat
```

ajreynol: Note this is actually unecessary, since there is also (set-info :status unsat) in the benchmark below.

PR #5880, opened 2021-02-08
Fix disequality between seq.unit terms

### LINKED ISSUE #5666: Invalid model for Seq formula (seq.prefixof, seq.suffixof, seq.unit)
cvc4 d16b689f85e
~~~~
(declare-fun i5 () Int)
(declare-fun seq2 () (Seq Int))
(assert (< 1 i5))
(assert (xor true true (seq.prefixof seq2 seq2) (seq.suffixof (seq.unit 2) (seq.unit i5))))
(check-sat)
~~~~
~~~~
./cvc4 -i -q --strings-exp --check-models delta.out.smt2 
Fatal failure within void CVC4::smt::CheckModels::checkModel(CVC4::smt::Model*, CVC4::context::CDList<CVC4::NodeTemplate<true> >*, bool) at /home/CVC4/src/smt/check_models.cpp:249
assertion:     (xor (xor (xor true true) (str.prefixof seq2 seq2)) (str.suffixof (seq.unit 2) (seq.unit i5)))
simplifies to: false
expected `true'.
Aborted
~~~~

ajreynol:
This adds a missing inference for disequality between seq.unit terms, which was not handled previously.

Fixes #5666.


test/regress/regress0/strings/issue5666-orig-unit-deq.smt2
```diff
@@ -0,0 +1,7 @@
+(set-logic ALL)
+(set-info :status sat)
+(declare-fun i5 () Int)
+(declare-fun seq2 () (Seq Int))
+(assert (< 1 i5))
+(assert (xor true true (seq.prefixof seq2 seq2) (seq.suffixof (seq.unit 2) (seq.unit i5))))
```
yoni206, line 6: is this `xor true true` needed?

ajreynol: Apparently not

src/theory/strings/core_solver.cpp
```diff
@@ -1987,6 +1987,64 @@ void CoreSolver::processDeq(Node ni, Node nj)
...
+        continue;
+      }
+      Node u = nfc.d_nf[0];
+      // if the other side is constant like
+      NormalForm& nfo = i == 0 ? nfnj : nfni;
+      if (nfo.d_nf.size() == 0 || !utils::isConstantLike(nfo.d_nf[0]))
+      {
+        break;
+      }
+      Node v = nfo.d_nf[0];
```
yoni206, line 2011: Can you add a comment about what is `v` expected to be?

test/regress/CMakeLists.txt
```diff
@@ -1108,6 +1108,8 @@ set(regress_0_tests
   regress0/strings/issue5428-re-diff-assoc.smt2
   regress0/strings/issue5542-strings-seq-mix.smt2
   regress0/strings/issue5608-eager-pp.smt2
+  regress0/strings/issue5666-orig-unit-deq.smt2
+  regress0/strings/issue5666-unit-deq.smt2
```
yoni206, line 1112: do you think there is any way to trigger this bug for pure strings (e.g. with strings of length 1)? If no, never mind. If yes, maybe you can add a regression?

yoni206 (approved):
Looks good, some comments for your consideration before merging.

test/regress/CMakeLists.txt
ajreynol: No, this is specific to the case of seq.unit.  (Disequalities between constant strings of length one are trivially satisfied).

PR #5883, opened 2021-02-09
Remove track instantiations infrastructure

ajreynol:
This was used in the old method for unsat core tracking for instantiation lemmas, which will soon be subsumed.

This is also work towards eliminating the dependencies on quantifiers engine from Instantiate.


HanielB (approved):
Awesome. :)

PR #5884, opened 2021-02-09
cmake: Make Python3 default and improve toml error messages.

mpreiner:
`./configure.sh` will now fail if Python3 is not installed on the system. Since Python2 is now deprecated the user has to explicitly enable it via `--python2`. This commit also removes the `--python3` configure flag.


configure.sh
```diff
@@ -44,8 +44,7 @@ The following flags enable optional features (disable with --no-<option name>).
   --coverage               support for gcov coverage testing
   --profiling              support for gprof profiling
   --unit-testing           support for unit testing
-  --python2                prefer using Python 2 (also for Python bindings)
-  --python3                prefer using Python 3 (also for Python bindings)
+  --python2                prefer using Python 2 (deprecated)
```
makaimann, line 47: Could consider changing the message to something like `force use of Python2`. If you decide to change it, I think it would make sense to change the description of the CMake variable as well.

makaimann (approved):
Looks good, thanks for the change! I left a minor comment for consideration.

configure.sh
mpreiner: Good point. Changed it in configure.sh and CMakeLists.txt

PR #5885, opened 2021-02-09
Fix open proof for factoring lemma

ajreynol:
We need to add to the current proof, regardless of whether we have used the factor skolem previously.

Fixes bugs found by @HanielB on SMT-LIB runs.


nafur (approved):
LGTM. Good catch!

PR #5887, opened 2021-02-10
Add assumption-based unsat cores.

mpreiner:
This PR adds an assumption-based unsat cores option. If enabled it will disable proof logging in the SAT solver and adds input assertions as assumptions to the SAT solver. When an unsat core is requested we extract the unsat core in terms of the unsat assumption in the SAT solver. Assumption-based unsat cores use the proof infrastructure to map the input assumptions back to the original assertions.


src/prop/prop_engine.cpp
```diff
@@ -593,10 +618,30 @@ std::shared_ptr<ProofNode> PropEngine::getProof()
   {
     return nullptr;
   }
+  if (options::unsatCores()
```
mpreiner, line 621: I'm not sure if this really is the right place to return this since we can't distinguish between returning a proof step for the unsat core or the proof itself.

src/prop/prop_engine.cpp
```diff
@@ -593,10 +618,30 @@ std::shared_ptr<ProofNode> PropEngine::getProof()
...
+      && options::unsatCoresMode() == options::UnsatCoresMode::ASSUMPTIONS)
+  {
+    std::vector<Node> core;
+    getUnsatCore(core);
+    CDProof cdp(d_pnm);
+    Node fnode = NodeManager::currentNM()->mkConst(false);
+    cdp.addStep(fnode, PfRule::SAT_REFUTATION, core, {});
+    return cdp.getProofFor(fnode);
+  }
```
HanielB, line 630: I think this should not be in `getProof`, but rather in a different call in the prop engine, like `getRefutation`. As it is if one is in both unsat core and proof mode the actual prop engine proof would never be obtained. So the distinction would be made actually in `SmtEngine::getUnsatCoreInternal`. 

The same applies to the changes in `convertAndAssert` above, which is also making it impossible to get both proofs and unsat cores (from assumptions).

mpreiner: You can't call `convertAndAssert` if you do assumption-based unsat cores.

HanielB: Hum, so it should be the case that assumption-based unsat cores are incompatible with proofs. That's fine, we can issue warnings in `set_defaults`.

src/smt/set_defaults.cpp
```diff
@@ -289,6 +289,14 @@ void setDefaults(LogicInfo& logic, bool isInternalSubsolver)
     disableProofs = true;
   }
 
+  if (options::unsatCores()
```
ajreynol, line 292: This isn't going to quite work I don't think, since this disables *all* proof production, including the allocation of proof node manager. We also need to keep proof production in the preprocessor.

I think I would recommend deleting this block of code and then catching what we need to do in smt::ProofManager.

ajreynol (changes requested):

src/expr/proof_rule.h
```diff
@@ -617,6 +617,15 @@ enum class PfRule : uint32_t
...
   CNF_ITE_NEG3,
 
+  // ========= SAT Refutation for assumption-based unsat cores
+  // Children: (P1, ..., Pn)
+  // Arguments: none
+  // ---------------------
+  // Conclusion: false
+  // Note: P1, ..., Pn corresponds to the unsat core determined by the SAT
+  // solver.
+  SAT_REFUTATION,
```
HanielB, line 627: Rather than this rule we can just use a trust rule. There will be no such checking anyway when we are just producing unsat cores.

src/smt/set_defaults.cpp
mpreiner: Oh yeah, I was stupid... We should probably just issue a warning that the user can't produce proofs, but of course the code should not be disabled...

src/prop/minisat/core/Solver.cc
```diff
@@ -223,12 +225,16 @@ Solver::Solver(CVC4::prop::TheoryProxy* proxy,
       propagation_budget(-1),
       asynch_interrupt(false)
 {
+  d_unsatCoresProofs =
```
HanielB, line 228: I think this option is fine, but when we delete the old proof code we would delete it, since in the unsat cores + proofs mode what is necessary is what is already guarded by `isProofsEnabled`

src/prop/minisat/core/Solver.h
```diff
@@ -398,6 +398,9 @@ class Solver {
      */
     vec<bool> theory;
 
+    /** Indicates if unsat cores should be generated via proofs. */
```
HanielB, line 401:
```suggestion
    /** Indicates if unsat cores should be generated via the old proof infrastructure. */
```

src/expr/proof_rule.h
ajreynol: I disagree, TRUST should never be used, it should always be a specific reason, albeit if it is just an identifier.

Notice that this is the reason TRUST is not on master.

ajreynol: My recommendation: this rule should exist. The documentation should indicate that this is a "trusted" rule.  See e.g. https://github.com/CVC4/CVC4/blob/master/src/expr/proof_rule.h#L216.

HanielB: I had not noticed that it was not. :) That's fine then.

test/regress/run_regression.py
```diff
@@ -346,6 +346,10 @@ def run_regression(check_unsat_cores, check_proofs, dump, use_skip_return_code,
                '--incremental' not in all_args and \
                '--unconstrained-simp' not in all_args:
                 extra_command_line_args += ['--check-unsat-cores']
+                if '--no-check-unsat-cores-new' not in all_args and \
+                   '--produce-proofs' not in all_args:
+                    extra_command_line_args += ['--check-unsat-cores-new']
+                    extra_command_line_args += [['--check-unsat-cores-new', '--unsat-cores-mode=assumptions']]
```
ajreynol, line 352: I'm guessing this block won't go into master like this, right?

test/regress/run_regression.py
```diff
@@ -358,7 +362,10 @@ def run_regression(check_unsat_cores, check_proofs, dump, use_skip_return_code,
 
         # Create a test case for each extra argument
         for extra_arg in extra_command_line_args:
-            command_line_args_configs.append(all_args + [extra_arg])
+            if isinstance(extra_arg, list):
```
ajreynol, line 365: Can you comment what this is doing?

ajreynol (approved):
A few questions, otherwise LGTM.

Also, a regression is still failing, do you know what the issue is?

src/smt/set_defaults.cpp
```diff
@@ -391,6 +391,27 @@ void setDefaults(LogicInfo& logic, bool isInternalSubsolver)
...
+      Notice() << "Proof production and assumption-based unsat cores are "
+               << "not compatible. Proofs cannot be generated." << std::endl;
+    }
+    else
+    {
+      Notice() << "Proof production not enabled, but is required for unsat "
+               << "cores. Disabling assumption-based unsat cores." << std::endl;
+      options::unsatCoresMode.set(options::UnsatCoresMode::PROOF);
+    }
+  }
```
HanielB, line 413: For now the assumptions mode would require --produce-unsat-cores-new, right? There are already checks relative to behavior of that option and --produce-proofs/--check-proofs, so I think this code block can be deleted.

PR #5889, opened 2021-02-10
[proof-new] Adding a proof-producing ensure literal method

HanielB:
The `ensureLiteral` method in `CnfStream` may apply CNF conversion to the given literal (for example if it's an `IFF`), which needs to be recorded in the proof. This commit adds a proof-producing `ensureLiteral` to `ProofCnfStream`, which is called by the prop engine if proofs are enabled.

This commit also simplifies the interfaces on `ensureLit` and `convertAtom` by removing the `preRegistration` flag, which was never used.


src/prop/proof_cnf_stream.h
```diff
@@ -84,6 +84,14 @@ class ProofCnfStream : public ProofGenerator
    * node are saved in d_proof. */
   void convertPropagation(theory::TrustNode ttn);
 
+  /**
+   * Ensure that the given node will have a designated SAT literal that is
+   * definitionally equal to it.  The result of this function is that the Node
+   * can be queried via getSatValue(). Essentially, this is like a "convert-but-
+   * don't-assert" version of convertAndAssert().
+   */
+  void ensureLiteral(TNode n, bool noPreregistration = false);
```
ajreynol, line 93: Can you also delete `noPreregistration`? It is never set to true.

src/prop/cnf_stream.h
```diff
@@ -203,6 +203,9 @@ class CnfStream {
   SatLiteral handleAnd(TNode node);
   SatLiteral handleOr(TNode node);
 
+  /** Stores the literal of the given node in d_literalToNodeMap. */
```
ajreynol, line 206: Say that n should already have a literal via getLiteral.

ajreynol (changes requested):

src/prop/cnf_stream.h
```diff
@@ -203,6 +203,13 @@ class CnfStream {
   SatLiteral handleAnd(TNode node);
   SatLiteral handleOr(TNode node);
 
+  /** Stores the literal of the given node in d_literalToNodeMap.
+   *
+   * Note that n must already have a literal associated to it in
+   * d_nodeToLiteralMap.
+   */
+  void ensureExistingLiteral(TNode n);
```
ajreynol, line 211: To avoid confusion with ensureLiteral, let's call this `ensureMappingForLit` or something along these lines.  Also can this be private (ProofCnfStream is friend?).

ajreynol (changes requested):

src/prop/cnf_stream.h
HanielB: Currently there are no private components to CnfStream, just protected ones. Indeed `ProofCnfStream` is a friend. I'd say we can change this in another PR.

ajreynol (approved):

PR #5890, opened 2021-02-10
Simplify interface for preprocessor

ajreynol:
This does a minor simplification to the interface for preprocessor.  It has a `postProcess` call that is called after another unrelated call at the `SmtSolver` level (notifyPreprocessedAssertions) is made.  This makes it so that `Preprocessor::postProcess` is inlined as the last step of `process`.


HanielB (approved):
LGTM

PR #5894, opened 2021-02-11
Add InferenceId member to TheoryInference

nafur:
This PR adds a new InferenceId member to the TheoryInference class.
All classes derived from TheoryInference are adapted accordingly.


ajreynol (approved):
LGTM, its a bit unfortunately that we need TheoryInference::setId(...) but I think its ok for now.

nafur:
> LGTM, its a bit unfortunately that we need TheoryInference::setId(...) but I think its ok for now.

I agree. But some code needs a significant refactoring before we can get rid of setId(), but we should try to do this at some point.

PR #5895, opened 2021-02-11
(proof-new) Option to not automatically consider symmetry in CDProof

ajreynol:
There are compelling use cases not to automatically introduce SYMM steps in CDProof, e.g. in CDProofs used within ProofNodeUpdater for external conversions.


HanielB (approved):
LGTM

PR #5898, opened 2021-02-11
Add new IntegralHistogramStat

nafur:
This PR adds a new statistics class that improves on HistogramStat if we know that the type is integral (also supporting enums).
Instead of using a comparably slow std::map like HistogramStat, IntegralHistogramStat uses a std::vector that is resized appropriately.
The integral values are simply cast to std::int64_t and used as indices into the vector.


mpreiner (changes requested):
LGTM, but should include unit tests to test the base functionality.

mpreiner (approved):
LGTM

PR #5904, opened 2021-02-12
Add tuple projection operator

mudathirmahgoub:
This PR adds tuple projection operator to the theory of data types. 
It Also adds helper functions for selecting elements from a tuple.


src/api/cvc4cpp.h
```diff
@@ -2559,6 +2559,26 @@ class CVC4_PUBLIC Solver
...
+  /**
+   * returns a tuple that represents
+   * (mkTuple ((_ tupSel i_1) t) ... ((_ tupSel i_n) t))
+   * where each i_j is less than the length of t
+   * @param t a node that has a tuple type
+   * @param indices of the projected elements
+   * @return a tuple that represents the elements with indices {i_1, ..., i_n}
+   * in tuple t
+   */
+  Term tupleProject(Term t, std::vector<uint32_t> indices) const;
```
ajreynol, line 2580:
```suggestion
  Term tupleProject(const std::vector<uint32_t>& indices, Term t) const;
```

ajreynol: I think t should come after indices to be consistent with tupleSelect.

src/theory/datatypes/project_op.cpp
```diff
@@ -0,0 +1,61 @@
...
+}
+
+ProjectOp::ProjectOp(std::vector<uint32_t> indices)
+    : d_indices(std::move(indices))
+{
+}
+
+ProjectOp::ProjectOp(const ProjectOp& op) : d_indices(std::move(op.d_indices))
+{
+  // ToDo: test this
```
ajreynol, line 51: What to test here?

ajreynol: Also, use "TODO", and if this stays, we should link a cvc4-projects issue.

src/theory/datatypes/project_op.h
```diff
@@ -0,0 +1,59 @@
...
+class TypeNode;
+
+/**
+ * The class is an operator for kind project used to project elements in a tuple
+ * It stores the indices of projected elements
+ */
+class ProjectOp
+{
+ public:
+  ProjectOp(std::vector<uint32_t> indices);
```
ajreynol, line 34:
```suggestion
  ProjectOp(const std::vector<uint32_t>& indices);
```

src/theory/datatypes/theory_datatypes_type_rules.h
```diff
@@ -571,6 +571,60 @@ class MatchBindCaseTypeRule
...
+             << std::endl;
+          throw TypeCheckingExceptionPrivate(n, ss.str());
+        }
+      }
+    }
+    TypeNode tupleType = n[0].getType(check);
+    std::vector<TypeNode> types;
+    DType dType = tupleType.getDType();
+    DTypeConstructor constructor = dType[0];
+    for (const uint32_t& index : indices)
```
ajreynol, line 620:
```suggestion
    for (uint32_t index : indices)
```

src/theory/datatypes/theory_datatypes_type_rules.h
```diff
@@ -571,6 +571,60 @@ class MatchBindCaseTypeRule
...
+        std::stringstream ss;
+        ss << "PROJECT expects a tuple for " << n[0] << ". Found" << tupleType;
+        throw TypeCheckingExceptionPrivate(n, ss.str());
+      }
+
+      // make sure all indices are less than the length of the tuple type
+      DType dType = tupleType.getDType();
+      DTypeConstructor constructor = dType[0];
+      size_t numArgs = constructor.getNumArgs();
+      for (const uint32_t& index : indices)
```
ajreynol, line 604:
```suggestion
      for (uint32_t index : indices)
```

ajreynol (changes requested):

src/api/cvc4cpp.h
mudathirmahgoub: Done. More importantly, do we need this function? If we are introducing the project operator, the user would have 2 options to project a tuple

src/theory/datatypes/project_op.cpp
mudathirmahgoub: Removed TODO here

src/theory/datatypes/theory_datatypes_type_rules.h
mudathirmahgoub: Done

src/theory/datatypes/theory_datatypes_type_rules.h
mudathirmahgoub: Done

src/theory/datatypes/project_op.h
mudathirmahgoub: I am not sure about the best option to pass a vector to a constructor. Clang-tidy complained about const and & and suggested using pass by value and std::move in the implementation. There is some discussion [here](https://stackoverflow.com/questions/34799259/c11-pass-vector-to-constructor)

src/api/cvc4cpp.h
```diff
@@ -2559,6 +2559,26 @@ class CVC4_PUBLIC Solver
...
+   * @param index represents the index of the selected element in the tuple
+   * @param tuple a node that has a tuple type
+   * @pre sort of tuple is (Tuple T1 ... Tn) and index < n
+   * @return the element at the given index in tuple
+   */
+  Term tupleSelect(uint32_t index, Term tuple) const;
+
+  /**
+   * returns a tuple that represents
+   * (mkTuple ((_ tupSel i_1) t) ... ((_ tupSel i_n) t))
```
ajreynol, line 2573:
```suggestion
   * returns a tuple ((_ project i_1 ... i_n) t) which note is equivalent to
   * (mkTuple ((_ tupSel i_1) t) ... ((_ tupSel i_n) t))
```

src/theory/datatypes/project_op.h
```diff
@@ -0,0 +1,58 @@
...
+ ** \verbatim
+ ** Top contributors (to current version):
+ **   Mudathir Mohamed
+ ** This file is part of the CVC4 project.
+ ** Copyright (c) 2009-2020 by the authors listed in the file AUTHORS
+ ** in the top-level source directory and their institutional affiliations.
+ ** All rights reserved.  See the file COPYING in the top-level source
+ ** directory for licensing information.\endverbatim
+ **
+ ** \brief a class for MK_BAG operator
```
ajreynol, line 12: Fix.

ajreynol (changes requested):
Also, for 0-argument projection, is the syntax `((_ project) t)` or should it just be `(project t)`?  I'm not sure if SMT-LIB permits indexed symbols with 0 indices.

src/theory/datatypes/project_op.h
mudathirmahgoub: Done. I wonder why MK_BAG is here. Perhaps this file was written in 2020 :)

src/api/cvc4cpp.h
mudathirmahgoub: There is some redundancy here. The implementation of this function actually returns a `tuple` term, not a `project` term. 
Currently If the user wants a project term, he should write
```cpp
  Op op = d_solver.mkOp(PROJECT, indices);
  Term term = d_solver.mkTerm(op, tuple);
```
The term would be rerwritten later to be the same as term returned by `Solver::tupleProject` .
It is unfortunate that `Solver::tupleProject` is similar to `Solver::tupleSelect` which is needed because we don't have a `tupleSelect` operator

mudathirmahgoub:
> Also, for 0-argument projection, is the syntax `((_ project) t)` or should it just be `(project t)`? I'm not sure if SMT-LIB permits indexed symbols with 0 indices.

We have another issue as well. How do we represent an empty tuple. Below is the output of the regression 

```smt2
(declare-fun v () Tuple) ...
---------------------------------------------------------------------------------------------------------------------
(
...
(define-fun v () Tuple mkTuple)
...
)
```

src/theory/datatypes/project_op.h
ajreynol: That's interesting, I was unaware of this. @4tXJ7f maybe could comment, my style has favored const & in this case.

src/api/cvc4cpp.h
ajreynol: Rewriting is an internal detail for solving.  This documentation is for the purpose of the user, when you use this method, it should create a tuple projection term.  Thus it should say what is created, not what it internally rewrites to.

src/theory/datatypes/project_op.h
ajreynol: > > Also, for 0-argument projection, is the syntax `((_ project) t)` or should it just be `(project t)`? I'm not sure if SMT-LIB permits indexed symbols with 0 indices.
> 
> We have another issue as well. How do we represent an empty tuple. Below is the output of the regression
> 
> ```smt
> (declare-fun v () Tuple) ...
> ---------------------------------------------------------------------------------------------------------------------
> (
> ...
> (define-fun v () Tuple mkTuple)
> ...
> )
> ```
> 
> Cesare suggested something like
> 
> ```smt
> (define-fun v () Tuple emptyTuple) ; Unit data type has only one constructor
> ```

I think 0-argument tuples should use the same syntax as the tuple constructor, which is `mkTuple`, although we had discussed changing this to `tuple`.
Thus, I prefer the current implementation.

ajreynol:
> > Also, for 0-argument projection, is the syntax `((_ project) t)` or should it just be `(project t)`? I'm not sure if SMT-LIB permits indexed symbols with 0 indices.
> 
> We have another issue as well. How do we represent an empty tuple. Below is the output of the regression
> 
> ```smt
> (declare-fun v () Tuple) ...
> ---------------------------------------------------------------------------------------------------------------------
> (
> ...
> (define-fun v () Tuple mkTuple)
> ...
> )
> ```
> 
> Cesare suggested something like
> 
> ```smt
> (define-fun v () Tuple emptyTuple) ; Unit data type has only one constructor
> ```

I think 0-argument tuples should use the same syntax as the tuple constructor, which is mkTuple, although we had discussed changing this to tuple.
Thus, I prefer the current implementation.

src/api/cvc4cpp.h
```diff
@@ -2559,6 +2559,15 @@ class CVC4_PUBLIC Solver
   Term mkTuple(const std::vector<Sort>& sorts,
                const std::vector<Term>& terms) const;
 
+  /**
```
ajreynol, line 2562: I think we shouldn't include this yet either.

ajreynol (changes requested):

src/api/cvc4cpp.h
mudathirmahgoub: Done

ajreynol (approved):
LGTM

src/theory/datatypes/project_op.h
nafur: I agree with @mudathirmahgoub: his solution always copies into the function argument and then moves it within the constructor (which is constant, like 6 pointer assignments). `const &`, on the other hand, tries to avoid the copy for the argument, but then copies within the constructor.
You pay a small price (one additional move), but there is a catch: if you regularly call this method with something else (for example an initializer list), then this will actually save a full copy.

PR #5905, opened 2021-02-12
Properly set up equality engine for BV bitblast solver.

mpreiner:
Theory BV now sets up the default equality engine for BV solvers that do not use their own equality engine like e.g. the BV bitblast solver.  This commit also adds the missing equality engine pieces to the BV bitblast solver (`getEqualityStatus`, `explain`).


ajreynol (approved):
LGTM

PR #5913, opened 2021-02-17
Add statistic for InferenceId to TheoryInferenceManager.

nafur:
This PR uses the IntegralHistogramStat to obtain statistics about the sent inferences within the TheoryInferenceManager. All theories are adapted to provide a proper name (prefix) for the name of the statistic.


src/theory/theory_inference_manager.h
```diff
@@ -432,6 +435,8 @@ class TheoryInferenceManager
   uint32_t d_numCurrentLemmas;
   /** The number of internal facts added since the last call to reset. */
   uint32_t d_numCurrentFacts;
+  /** Statistics for the inference ids sent via this inference manager. */
+  IntegralHistogramStat<InferenceId> d_inferenceStats;
```
ajreynol, line 439: I actually think we will want to split this into 3 parts (lemma vs fact vs conflict).  Note that both in strings and in datatypes, there are inferences that may be sent as lemmas or as facts. FYI @yoni206 .

src/theory/theory_inference_manager.cpp
```diff
@@ -34,11 +36,18 @@ TheoryInferenceManager::TheoryInferenceManager(Theory& t,
       d_lemmasSent(t.getUserContext()),
       d_numConflicts(0),
       d_numCurrentLemmas(0),
-      d_numCurrentFacts(0)
+      d_numCurrentFacts(0),
+      d_inferenceStats(name + "::inferenceIds")
```
ajreynol, line 40: If we split this into 3, maybe this would be `::inferencesFact`, `::inferencesLemma` and `::inferencesConflict`.

ajreynol (changes requested):

ajreynol (approved):

PR #5914, opened 2021-02-17
Compute fact or lemma in datatypes prior to buffering

ajreynol:
This is necessary for the planned refactoring of TheoryInference::process. This forces datatypes to decide lemma vs. fact prior to buffering inferences.

FYI @yoni206


nafur (commented):
LGTM.

nafur (approved):
LGTM.

PR #5916, opened 2021-02-17
Move methods from term util to quantifiers registry

ajreynol:
Towards eliminating dependencies on quantifiers engine, and eliminating the TermUtil class.

Note that `QuantifiersModule` had to be moved to its own file to avoid circular include dependencies.


mpreiner (approved):
LGTM

PR #5919, opened 2021-02-17
Eliminate non-static members in term util

ajreynol:
This makes it so that TermUtil is now a collection of static methods.  Further refactoring will make this a standalone file of utility methods. 

This breaks all dependencies on the TermUtil object in QuantifiersEngine.  It also starts breaking some of the depenendencies on quantifiers engine in sygus.


mpreiner (approved):
LGTM

PR #5924, opened 2021-02-18
Fix rewrite for contains over replace

### LINKED ISSUE #5915: Soundness bug in string logic
cvc4 commit: bdc1b222
```
$ cvc4 --strings-exp --produce-models -q bug.smt2
sat
(
(define-fun x () String "AB")
(define-fun y () String "AB")
)
$ z3release bug.smt2
unsat
$ cat bug.smt2
(declare-fun x () String)
(declare-fun y () String)
(assert (= (str.replace (str.replace x "B" (str.++ "B" "B")) "B" (str.++ y "B")) (str.++ y "B")))
(check-sat)
(get-model)
```
CVC4 with `--check-models` option
```
Fatal failure within void CVC4::smt::CheckModels::checkModel(CVC4::smt::Model*, CVC4::context::CDList<CVC4::NodeTemplate<true> >*, bool) at CVC4/src/smt/check_models.cpp:129
assertion:     (let ((_let_1 (str.++ y "B"))) (= (str.replace (str.replace x "B" (str.++ "B" "B")) "B" _let_1) _let_1))
simplifies to: false
expected `true'.
Aborted
```

ajreynol:
Fixes model soundness issue (fixes #5915).

FYI @4tXJ7f


HanielB (approved):
LGTM

PR #5926, opened 2021-02-18
Simplify interface to instantiate

ajreynol:
Does not support InstMatch interfaces anymore, which are spurious.


src/theory/quantifiers/inst_match_trie.h
```diff
@@ -314,7 +273,7 @@ class InstMatchTrieOrdered
    */
   bool addInstMatch(quantifiers::QuantifiersState& qs,
                     Node q,
-                    InstMatch& m,
+                    std::vector<Node>& m,
```
HanielB, line 276: I think this might be passed as `const`?

src/theory/quantifiers/inst_match_trie.h
```diff
@@ -327,7 +286,7 @@ class InstMatchTrieOrdered
    */
   bool existsInstMatch(quantifiers::QuantifiersState& qs,
                        Node q,
-                       InstMatch& m,
+                       std::vector<Node>& m,
```
HanielB, line 289: ditto

src/theory/quantifiers/inst_match_trie.h
```diff
@@ -314,7 +273,7 @@ class InstMatchTrieOrdered
    */
   bool addInstMatch(quantifiers::QuantifiersState& qs,
                     Node q,
-                    InstMatch& m,
+                    std::vector<Node>& m,
                     bool modEq = false)
   {
     return d_imt.addInstMatch(qs, q, m, modEq, d_imtio);
```
HanielB, line 279: Since you're cleaning this, pass the implementations to the `cpp` file.

HanielB (approved):
LGTM

PR #5927, opened 2021-02-18
Cleanup of inferences in arithmetic theory

nafur:
This PR cleans up several issues in the arithmetic theory mostly related to its usage of InferenceId and the TheoryInferenceManager:
- remove the `ArithLemma` class and uses `SimpleTheoryLemma` instead
- only use `NlLemma` if we actually need it
- use proper InferenceIds everywhere
- remove unused code in the nonlinear extension


ajreynol (approved):
LGTM

PR #5931, opened 2021-02-19
Fill in missing inference ids in datatypes theory

ajreynol:
Also updates its inference manager to not track stats since the standard ones are now used.

This also sets up some dependencies in the sygus extension which will be used to implement InferenceId for the sygus extension, to be done on a separate PR.


src/theory/datatypes/inference_manager.h
```diff
@@ -45,16 +44,16 @@ class InferenceManager : public InferenceManagerBuffered
...
    * mustCommunicateFact.
-   * @param i The inference, used for stats and as a hint for constructing
-   * the proof of (conc => exp)
    */
   void addPendingInference(Node conc,
                            Node exp,
-                           bool forceLemma = false,
-                           InferenceId i = InferenceId::UNKNOWN);
+                           InferenceId id,
+                           bool forceLemma = false);
```
nafur, line 56: I think we wanted to put the `id` before `exp`...

nafur (approved):
Just one minor issue.

src/theory/datatypes/inference_manager.h
ajreynol: ah, right.

PR #5937, opened 2021-02-20
Fix datatypes inference manager when proofs are enabled

ajreynol:
Accidentally was not sending lemmas in one interface when proofs are enabled due to recent refactoring.


nafur (approved):
LGTM.

PR #5939, opened 2021-02-20
(proof-new) Option to automatically add SYMM steps during proof node update

ajreynol:
Required for work on external proof conversions.


HanielB (approved):
LGTM

PR #5946, opened 2021-02-22
(proof-new) Add explicit rewrite to iand solver.

nafur:
This PR adds explicit rewrites to the iand solver where previously the inference manager would have rewritten the lemmas.


ajreynol (changes requested):
What is the purpose of this?  Lemmas are already rewritten when they are sent to TheoryEngine.  This may be unnecessary.

nafur:
Discussed offline: these rewrites are probably obsolete, and should be removed from proof-new instead in this case.

nafur:
Rewrites seem to be obsolete.

PR #5949, opened 2021-02-22
(proof-new) Add new arithmetic kind INDEXED_ROOT_PREDICATE.

nafur:
This PR introduces a new arithmetic kind for indexed root predicates.
An indexed root predicate compares a real variable to the k'th root of a given polynomial as follows:
Let `IRP_k(x ~ 0, p)` an indexed root predicate with `k` a non-negative number, `x` some real variable, `~` an arithmetic relation (e.g. `=`, `<`, ...), and `p` a polynomial over `x` (and possibly other variables).
If `p` contains variables apart from `x`, we can only evaluate the expression over a suitable assignment for at least these variables.
The evaluation of this expression is equivalent to computing the k'th real root of `p` in `x` (with all other variables evaluated over a given assignment) and comparing this real root to zero (according to the relation symbol `~`).

Note that we currently do not intend to use this structure for solving, but require it for representing and printing CAD proofs.

### DIFF
--- a/src/util/arith.h
+++ b/src/util/arith.h
@@ -1,0 +1,58 @@
+/*********************                                                        */
+/*! \file arith.h
+ ** \verbatim
+ ** Top contributors (to current version):
+ **   Aina Niemetz, Andres Noetzli, Dejan Jovanovic
+ ** This file is part of the CVC4 project.
+ ** Copyright (c) 2009-2020 by the authors listed in the file AUTHORS
+ ** in the top-level source directory and their institutional affiliations.
+ ** All rights reserved.  See the file COPYING in the top-level source
+ ** directory for licensing information.\endverbatim
+ **
+ ** \brief Utils for arithmetic kinds.
+ **
+ ** Some utils for arithmetic kinds, for example indexed root predicates.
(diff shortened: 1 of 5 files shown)


nafur:
I feel this PR should add some proper documentation on what this kind means (similar to what I explained in the comment of the PR), but I'm not sure where this should go...

src/util/arith.h
```diff
@@ -0,0 +1,58 @@
+/*********************                                                        */
```
ajreynol, line 1: Should we just call this file index_root_predicate.h?

src/util/arith.h
```diff
@@ -0,0 +1,58 @@
...
+/*! \file arith.h
+ ** \verbatim
+ ** Top contributors (to current version):
+ **   Aina Niemetz, Andres Noetzli, Dejan Jovanovic
```
ajreynol, line 5: Update.

src/util/arith.h
nafur: Yes, we should probably do that.

ajreynol (approved):

PR #5951, opened 2021-02-22
Explanation of failure for instantiate, use in enumerative instantiation

ajreynol:
This makes enumerative instantiation generalize the failures in Instantiate::addInstantiate and increment its enumeration accordingly.

This leads to (+104-6) using enumerative instantiation only on SMT-LIB quantified benchmarks, 60 second timeout.

This is in preparation for further improvements to enumerative instantiation.


src/theory/quantifiers/instantiate.h
```diff
@@ -146,6 +146,29 @@ class Instantiate : public QuantifiersUtil
...
+   * If terms = { 4, 0 }, then this method will return false since
+   *   4>5 => P(4,0)
+   * is entailed true based on rewriting. This method may additionally set
+   * failMask to "10", indicating that x's value was critical, but y's value
+   * was not. In other words, all instantiations including { x -> 4 } will also
+   * lead to this method returning false.
+   *
+   * The parameter expFull is whether try to set all bits of the fail mask to
+   * 0. If this argument is true, then we only try to set a suffix of the
+   * bits to false.
```
HanielB, line 163: If it's false? And clarify why only a suffix, since it's not clear from the doc that you're testing in reverse order.

src/theory/quantifiers/instantiate.h
```diff
@@ -146,6 +146,29 @@ class Instantiate : public QuantifiersUtil
                         bool mkRep = false,
                         bool modEq = false,
                         bool doVts = false);
+  /**
+   * Same as above, but we also compute a vector failMask which contains
+   * information about what values in terms led to the instantiation not being
```
HanielB, line 151: "contains information" kinda indicates that it's more than binary. Maybe say something like "failMask marks which values in terms *did not* contribute to the instantiation ..."? The important information is what terms were irrelevant, as by default we say all were.

src/theory/quantifiers/instantiate.h
```diff
@@ -146,6 +146,29 @@ class Instantiate : public QuantifiersUtil
...
+   * added when this method returns false.  For example, if q is the formula
+   *   forall xy. x>5 => P(x,y)
+   * If terms = { 4, 0 }, then this method will return false since
+   *   4>5 => P(4,0)
+   * is entailed true based on rewriting. This method may additionally set
+   * failMask to "10", indicating that x's value was critical, but y's value
+   * was not. In other words, all instantiations including { x -> 4 } will also
+   * lead to this method returning false.
+   *
```
HanielB, line 160: It'd be good to say a bit how the method is trying to set the mask by checking one variable at a time in reverse order. And that this matter because of how the mask is used by the user of this method.

src/theory/quantifiers/instantiate.h
```diff
@@ -147,9 +147,9 @@ class Instantiate : public QuantifiersUtil
                         bool modEq = false,
                         bool doVts = false);
   /**
-   * Same as above, but we also compute a vector failMask which contains
-   * information about what values in terms led to the instantiation not being
-   * added when this method returns false.  For example, if q is the formula
+   * Same as above, but we also compute a vector failMask indicates which values
```
HanielB, line 150: "indicates -> indicating" or "which indicates the values"

HanielB (approved):

PR #5952, opened 2021-02-22
(proof-new) Add proofs for sine lemmas in the transcendental solver

nafur:
This PR adds proofs for the lemmas related to the sine function in the transcendental solver.
It introduces several new proof rules with corresponding proof checkers and produces proofs in the sine solver.


ajreynol (approved):
LGTM

PR #5953, opened 2021-02-22
Require length-in-conclusion form for strings inferences

### LINKED ISSUE #5940: Refutational soundness bug on string formula
```
$cvc4 -q bug.smt2                                                                  
unsat                                                                              
(error "Cannot get model unless immediately preceded by SAT/NOT_ENTAILED or UNKNOWN response.")
$z3release bug.smt2                                                                
sat                                                                                
(                                                                                  
  (define-fun z () Int                                                             
    0)                                                                             
  (define-fun y () String                                                          
    "")                                                                            
  (define-fun x () String                                                          
    "")                                                                            
)                                                                                  
$cat bug.smt2                                                                      
(declare-fun x () String)                                                          
(declare-fun y () String)                                                          
(declare-fun z () Int)                                                             
(assert (= "B" (str.replace (str.substr "A" 0 z) "" (str.replace "B" 
             (str.substr "B" 0 0) (str.substr "A" 0 z)))))
(check-sat)                                                                        
(get-model)                                                                        
$cvc4 -q feed_model.smt2                                                           
sat                                                                                
$cat feed_model.smt2                                                               
(define-fun z () Int 0)                                                            
(define-fun y () String "")                                                        
(define-fun x () String "")                                                        
(assert (= "B" (str.replace (str.substr "A" 0 z) "" 
             (str.replace "B" (str.substr "B" 0 0) (str.substr "A" 0 z)))))
(check-sat)
```
Commit: d278cfe0

ajreynol:
Previously, it was optional whether to put length constraints in conclusion for deq string inferences.  However, due to optimizations in skolem caching, it is now required to guard these length constraints.

It furthermore changes the policy to not ignore the lengths for the registered skolem, since it may be shared elsewhere.

Fixes #5940.

Notice that proof-new already requires this option to be enabled when proofs are enabled.  Hence, this will simplify proof-new when merged.


4tXJ7f (approved):
LGTM, thanks for the fix.

PR #5954, opened 2021-02-22
[proof-new] Optionally print conclusion in the AST proof

HanielB:
Adds an option to optionally print conclusion in the AST proof.


src/options/smt_options.toml
```diff
@@ -139,6 +139,15 @@ header = "options/smt_options.h"
   read_only  = true
   help       = "output proofs after every UNSAT/VALID response"
 
+[[option]]
```
ajreynol, line 142: Let's move this to proof_options.

ajreynol (changes requested):

ajreynol (approved):

PR #5956, opened 2021-02-22
(proof-new) Add proofs for exponential functions

nafur:
This PR adds proofs for lemmas concerned with the exponential function. If also adds the necessary proof rules and corresponding proof checker.


ajreynol (approved):
LGTM

PR #5957, opened 2021-02-22
(proof-new) Add proof for transcendental secant lemmas

nafur:
This PR adds proofs for secant lemmas in the transcendental lemmas for both exponential and sine functions.
It also adds proof rules and corresponding proof checkers.


ajreynol (approved):
LGTM

PR #5958, opened 2021-02-22
(proof-new) Fix regular expression unfolding under substitution

ajreynol:
This case was previously unhandled and exercised by a recently added regression.


HanielB (approved):
LGTM

PR #5959, opened 2021-02-22
Switch to C++17.

mpreiner:


nafur (approved):
LGTM. Yay! 🎉

src/expr/node.h
```diff
@@ -1451,8 +1451,13 @@ NodeTemplate<ref_count>::substitute(Iterator substitutionsBegin,
...
   // otherwise compute
-  Iterator j = find_if(substitutionsBegin, substitutionsEnd,
-                       bind2nd(first_equal_to<typename Iterator::value_type::first_type, typename Iterator::value_type::second_type>(), *this));
+  Iterator j = find_if(
+      substitutionsBegin,
+      substitutionsEnd,
+      std::bind(first_equal_to<typename Iterator::value_type::first_type,
+                               typename Iterator::value_type::second_type>(),
+                std::placeholders::_1,
+                *this));
```
nafur, line 1460: How about this?
```suggestion
  Iterator j = find_if(
      substitutionsBegin,
      substitutionsEnd,
      [this](const auto& subst){ return subst.first == *this; });
```

nafur (changes requested):
My feeling is that lambdas are usually superior to std::bind in modern C++.
Not only are they far more readable, they are also usually more efficient: lambdas end up to be regular functions that can be inlined.

nafur:
So the next issue: dynamic exception specification was removed with c++17 (i.e. `void f() throw(...)`). As far as I can tell, we didn't use them anyway, except for a few commented out instances in `delta_rational.h`. With this removal, `std::set_unexpected` et al were removed as well, and we can simply remove them from our codebase. Also, we might want to consider marking functions `noexcept`.

nafur (approved):
LGTM.

ajreynol:
We should performance test SMT-LIB quantifiers, where substitute can be a bottleneck

ajreynol:
Nevermind, the changed variant of substitute is not the commonly used one.

PR #5961, opened 2021-02-23
[proof-new] Fix handling of removable clauses in proof cnf stream

HanielB:
Previously the `removable` information was not being communicated from the proof cnf stream to the cnf stream, which is the one that actually uses this when asserting clauses into the SAT solver. This commit fixes this by having the proof cnf stream directly use the cnf stream `d_removable` attribute.


ajreynol (approved):
LGTM

PR #5963, opened 2021-02-23
[proof-new] Fix dangling pointer in SAT proof generation

HanielB:
When a clause is being explained, the negation of each of its literals, other than the one it propagates, needs to be explained. This process may lead to the creation of new clauses in the SAT solver (because if a literal being explained was propagated and an explanation was not yet given, it will then be computed and added). This may lead to changes in the memory where clauses are, which may break the reference to the original clause being explained. To avoid this issue we store the literals in the reason before we start explaining their negations. We then iterate over them rather than over the clause, as before.

### DIFF
--- a/src/prop/sat_proof_manager.cpp
+++ b/src/prop/sat_proof_manager.cpp
@@ -352,9 +352,9 @@ void SatProofManager::explainLit(
     printClause(reason);
     Trace("sat-proof") << "\n";
   }
+#ifdef CVC4_ASSERTIONS
   // pedantically check that the negation of the literal to explain *does not*
   // occur in the reason, otherwise we will loop forever
-#ifdef CVC4_ASSERTIONS
   for (unsigned i = 0; i < size; ++i)
   {
     AlwaysAssert(~MinisatSatSolver::toSatLiteral(reason[i]) != lit)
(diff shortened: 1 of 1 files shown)


src/prop/sat_proof_manager.cpp
```diff
@@ -354,20 +354,32 @@ void SatProofManager::explainLit(
...
+  // Since explainLit calls can reallocate memory in the
+  // SAT solver, we directly get the literals we need to explain so we no
+  // longer depend on the reference to reason
+  std::vector<Node> toExplain{children.back().begin(), children.back().end()};
   NodeManager* nm = NodeManager::currentNM();
   Trace("sat-proof") << push;
   for (unsigned i = 0; i < size; ++i)
   {
-    SatLiteral curr_lit = MinisatSatSolver::toSatLiteral(reason[i]);
+    // pedantically make sure that the reason stays the same
```
ajreynol, line 376: What part of the code is making sure the reason stays the same?  Its confusing to have this before the assertions block, makes it seem like this check is only done when assertions are enabled, this isnt the case though?

ajreynol (changes requested):

src/prop/sat_proof_manager.cpp
HanielB: I should move it into the assertion block. The expectation is that the reason does not change. I just wanted to be extra careful.

ajreynol (approved):

PR #5964, opened 2021-02-23
(proof-new) Add proof generator for CAD solver

nafur:
This PR adds a proof generator for the CAD solver, including two new proof rules. The code is not yet used, but will be integrated into the CAD solver itself in another PR.


ajreynol (approved):
LGTM

PR #5965, opened 2021-02-23
(proof-new) Add proof for nonlinear monomial bounds check

nafur:
This PR adds proofs for a nonlinear refinement lemma that deals with multiplication of inequalities with some term. This lemma is split into two proof rules for positive or negative factors.


ajreynol (approved):
LGTM

PR #5980, opened 2021-02-23
Enable proofs on CI for proof enabled builds

ajreynol:


src/theory/builtin/proof_checker.cpp
```diff
@@ -422,7 +422,8 @@ Node BuiltinProofRuleChecker::checkInternal(PfRule id,
            || id == PfRule::THEORY_PREPROCESS_LEMMA
            || id == PfRule::THEORY_EXPAND_DEF || id == PfRule::WITNESS_AXIOM
            || id == PfRule::THEORY_LEMMA || id == PfRule::THEORY_REWRITE
-           || id == PfRule::TRUST_SUBS || id == PfRule::TRUST_SUBS_MAP)
+           || id == PfRule::TRUST_REWRITE || id == PfRule::TRUST_SUBS
+           || id == PfRule::TRUST_SUBS_MAP)
```
nafur, line 426: Did you mean to include this change? It seems unrelated to the PR.

nafur (approved):
LGTM, except for one comment.

src/theory/builtin/proof_checker.cpp
ajreynol: Yes, it's actually required to get CI to pass. Making this PR revealed that master had not taken this change from proof-new yet.

ajreynol:
FYI, we need to fix a debug failure on regress1 before merging this. @HanielB is investigating.  Also, we should wait before the nightlies pass again, which @mpreiner is working on.

test/regress/run_regression.py
```diff
@@ -382,6 +382,15 @@ def run_regression(unsat_cores, proofs, dump, use_skip_return_code,
         if extra_command_line_args:
             command_line_args_configs.append(all_args +
                                              extra_command_line_args)
+            
+        extra_command_line_args_proof = []
+        if proofs and re.search(r'^(unsat|valid)$', expected_output):
+            if '--no-check-proofs' not in all_args and \
+               '--check-proofs' not in all_args:
```
4tXJ7f, line 389: IIRC, we were combining proofs and unsat cores in the same call. This change will call the solver three times instead of two (at least on the unsat benchmarks). Are we ok with this?

test/regress/run_regression.py
```diff
@@ -473,7 +482,7 @@ def main():
 
     timeout = float(os.getenv('TEST_TIMEOUT', 600.0))
 
-    return run_regression(args.enable_proof, args.with_lfsc, args.dump,
+    return run_regression(args.enable_proof, args.enable_proof, args.dump,
```
4tXJ7f, line 485: I think that it would be good to remove the separate argument for unsat cores from `run_regression()` and the `--with-lfsc` from the argument parser above.

4tXJ7f (changes requested):

test/regress/run_regression.py
ajreynol: Yes, this was the intention.  In the future, unsat cores and proofs will be fairly disjoint (apart from preprocessing).

ajreynol:
Note this is subsumed by https://github.com/CVC4/CVC4/pull/6088.

PR #5981, opened 2021-02-24
(proof-new) Add proofs for CAD solver

nafur:
This PR adds proofs for the CAD solver, based on the proof generator from the previous PR.
Note that the level of detail of these CAD proofs is significantly higher than for other proofs. Making these proofs more fine-grained and maybe at some point accessible to proof checkers is probably still quite a bit of work.
Thus, the CAD proof rules are both trusted rules for now.


ajreynol (approved):
LGTM

PR #5990, opened 2021-02-24
Avoid spurious shadowing check for free variables computation in terms with WITNESS subterms

HanielB:
Achieved by not traversing witness terms when computing free variables.


ajreynol:
While WITNESS terms should not contain free variables, it is wrong to assume that they don't, as this PR does.  We should determine the root of the shadowing issue on proof-new, unfortunately this just ignores it.

HanielB:
This was indeed the product of cowardice. Let's fix this properly.

PR #5992, opened 2021-02-25
Move (optional) rewrite from TrustSubstitutionMap to SubstitutionMap.

### LINKED ISSUE #5943: Check failure at src/preprocessing/passes/non_clausal_simp.cpp:235
commit:  d278cfe0

```
$ cvc4 --check-models --produce-models --incremental --strings-exp -q bug.smt2
Fatal failure within virtual CVC4::preprocessing::PreprocessingPassResult CVC4::preprocessing::passes::NonClausalSimp::applyInternal(CVC4::preprocessing::AssertionPipeline*) 
at CVC4/src/preprocessing/passes/non_clausal_simp.cpp:235
Check failure

 cps.apply(t) == t

Aborted

$ cat bug.smt2
(declare-fun s () String)
(assert (and (and (= (str.len (str.at s 3)) 4) (not (not (= (ite (not (= (ite (not (= (str.at (str.at s 3) 
(ite (not (not (= (ite (<= (str.len (str.at s 1)) 3) 1 0) 0))) 1 0)) (str.at s 5))) 1 0) 0)) 1 0) 0)))) (not (not (= (str.len s) 3)))))
(check-sat)
```

nafur:
This PR adds optional rewriting to the SubstitutionMap class. Before, only the new TrustSubstitutionMap added optional rewriting, leading to unexpected inconsistencies between the two. In particular, cases exist where the substitution and the rewriting cancel each other (see #5943).
This PR moves the optional rewriting from TrustSubstitutionMap into SubstitutionMap. While the former enables it by default, it is disabled by default for the latter and thus does not change current behavior.
We now use this new option in an assertion in the non-clausal simplification.
Fixes #5943.


ajreynol (changes requested):
Add regression?

ajreynol (approved):

PR #5996, opened 2021-02-25
Refactor collection of debug and trace tags

### LINKED ISSUE #5921: Build failure with CVC4 in a long path
**Describe the bug**
It seems that if CVC4 is attempted to be built in a long enough path, then building fails with a `Argument list too long` error. The issue seems to stem from the `Debug_tags.tmp-<xxxxx>.sh` and `Trace_tags.tmp-<xxxxx>.sh` files produced by cmake. Unsure about the exact length, but a fresh clone into `/home/verylongpathevenlongername/makeiteverlongerifpossible/andonemorelevelofmakingitevenlonger` triggers the issue. However, one level up does not trigger the issue.

On the two systems tested, `getconf ARG_MAX` yields 2097152.

**Command line arguments**: n/a
**CVC4 version/commit**: c6210af1db67701495efa263207b91064a3bcd0b
**Operating system**: Tested on both Archlinux and a clean Ubuntu 18.04 docker container.

**`configure.sh` options**
No special options passed.

**`configure.sh` output**
[cvc4_configure.txt](https://github.com/CVC4/CVC4/files/6002004/cvc4_configure.txt)

nafur:
We have a mechanism to collect all debug and trace tags used throughout the code base to allow checking for valid tags.
This mechanism relies on a collection of more or less readable shell scripts.
#5921 hinted to a problem with the current setup, as it passes all source files via command line.
This PR refactors this setup so that the scripts collect the files internally, and only the base directory is passed on the command line.
As I was touching this code anyway, I ported everything to python and combined it into a single script, in the hope to make it more maintainable.
Fixes #5921.


src/base/CMakeLists.txt
```diff
@@ -51,42 +49,16 @@ file(GLOB_RECURSE source_files
...
-  OUTPUT ${CMAKE_CURRENT_BINARY_DIR}/Trace_tags
-  COMMAND ${gentags_script} Trace
-  DEPENDS ${gentags_script} ${CMAKE_CURRENT_BINARY_DIR}/Trace_tags.tmp
-)
-
 add_custom_command(
   OUTPUT ${CMAKE_CURRENT_BINARY_DIR}/Debug_tags.h
-  COMMAND ${genheader_script} ${CMAKE_CURRENT_LIST_DIR} Debug
-  DEPENDS mktagheaders ${genheader_script} ${CMAKE_CURRENT_BINARY_DIR}/Debug_tags
+  COMMAND ${PYTHON_EXECUTABLE} ${collect_tags_script} ${CMAKE_CURRENT_BINARY_DIR}/ Debug ${PROJECT_SOURCE_DIR}
```
mpreiner, line 54: The script should only look for tags in `${PROJECT_SOURCE_DIR}/src` otherwise you may have issues when working with multiple git worktrees in `${PROJECT_SOURCE_DIR}`.

src/base/CMakeLists.txt
```diff
@@ -51,42 +49,16 @@ file(GLOB_RECURSE source_files
...
+  DEPENDS ${collect_tags_script} ${source_files}
 )
 
 add_custom_command(
   OUTPUT ${CMAKE_CURRENT_BINARY_DIR}/Trace_tags.h
-  COMMAND ${genheader_script} ${CMAKE_CURRENT_LIST_DIR} Trace
-  DEPENDS mktagheaders ${genheader_script} ${CMAKE_CURRENT_BINARY_DIR}/Trace_tags
+  COMMAND ${PYTHON_EXECUTABLE} ${collect_tags_script} ${CMAKE_CURRENT_BINARY_DIR}/ Trace ${PROJECT_SOURCE_DIR}
```
mpreiner, line 60: Ditto.

src/base/collect_tags.py
```diff
@@ -0,0 +1,62 @@
...
+def parse_args():
+    """Parse command line arguments and return them."""
+    ap = argparse.ArgumentParser(description='collect debug and trace tags')
+    ap.add_argument('destdir', help='where to put the results')
+    ap.add_argument('type', choices=['Debug', 'Trace'], help='which tags to collect')
+    ap.add_argument('basedir', help='where to look for source file')
+    return ap.parse_args()
+
+
+def collect_tags(basedir: str):
```
mpreiner, line 20: Not sure if we should use type hints since this would require a minimum version of Python 3.5.

src/base/collect_tags.py
```diff
@@ -0,0 +1,62 @@
...
+    """Collect all tags used in filed within the given base directory.
+    Return them sorted lexicographically."""
+    tags = set()
+    for ext in ['.cc', '.cpp', '.g', '.h']:
+        for filename in glob.iglob(f'{basedir}/**/*{ext}', recursive=True):
```
mpreiner, line 25: Same with f-strings (requires at least Python 3.6).

mpreiner (approved):
LGTM, thanks for the changes!

PR #5997, opened 2021-02-25
Datatypes lemmas: share only external types

yoni206:
Forcing lemmas in datatypes used to be done only for external types.
This was changed to consider all types, which is not needed.
This PR brings back the restriction to external types.


src/theory/datatypes/theory_datatypes.cpp
```diff
@@ -1562,7 +1562,7 @@ void TheoryDatatypes::instantiate( EqcInfo* eqc, Node n ){
   // may contribute to conflicts due to cardinality (good examples of this are
   // regress0/datatypes/dt-param-card4-bool-sat.smt2 and
   // regress0/datatypes/list-bool.smt2).
-  bool forceLemma = true;
+  bool forceLemma = dt.involvesExternalType();
```
ajreynol, line 1565: Can you make this the `else` condition instead of computing/overwriting?

ajreynol (changes requested):

ajreynol (approved):

PR #6007, opened 2021-02-26
google test: theory: Migrate type_enumerator_white.

aniemetz:


ajreynol (approved):
Just FYI, this is a really annoying unit test (since it hardcodes the behavior of type enumerators, which have no obligation to give terms in an order). This perhaps should be rewritten at some point, if issues arise in it again.

PR #6010, opened 2021-02-26
Optionally permit creation of non-flat function types

ajreynol:
This is required for creating the representation of closues in LFSC, which are of the form `((forall x T) P)` where notice that `forall` has non-flat function type `(-> Int Sort (-> Bool Bool))`.


HanielB (approved):
LGTM

PR #6017, opened 2021-02-26
google test: util: Migrate check_white.

aniemetz:


yoni206 (approved):
LGTM.
Somehow this time GH knew that this is a move and was able to show the diff between the new file and the old file. Is it just luck? If not, it would be great if the other migration PRs look like this.

aniemetz:
This depends on the number of changes. All migration PRs are created the same (from my side), but if there's a lot of changes, git just "discards" the old one and treats the moved one as "new". Nothing we can do about this.

yoni206:
maybe two commits? one that moves the file and one that makes the changes?
Then I can review the PR by filtering only according to the second commit?

aniemetz:
That doesn't change anything. That's how the PR is created (no separate commit for the move). But it wouldn't change if git can match it or not.

aniemetz:
The issue is not the move (that's just a rename), but that there are so many changes.

PR #6018, opened 2021-02-26
google test: util: Migrate configuration_black.

aniemetz:


mudathirmahgoub (approved):
There is a conflict in test/unit/util/CMakeLists.txt, but the PR looks good to me.

PR #6030, opened 2021-03-01
Separate term tuple enumeration for quantifier instantiation

MikolasJanota:
Re-factoring out the code from `inst_strategy_enumerative` into a separate class. Some additional tricks to avoid duplicate instantiations, most notably, before instantiation, a tuple is checked if it's not a super-tuple of some tuple that had earlier resulted in a useless instantiation.

### DIFF
--- a/src/theory/quantifiers/index_trie.cpp
+++ b/src/theory/quantifiers/index_trie.cpp
@@ -1,0 +1,117 @@
+/*********************                                                        */
+/*! \file index_trie.cpp
+ ** \verbatim
+ ** Top contributors (to current version):
+ **   Mikolas Janota
+ ** This file is part of the CVC4 project.
+ ** Copyright (c) 2009-2020 by the authors listed in the file AUTHORS
+ ** in the top-level source directory and their institutional affiliations.
+ ** All rights reserved.  See the file COPYING in the top-level source
+ ** directory for licensing information.\endverbatim
+ **
+ ** \brief Implementation of a trie that store subsets of tuples of term indices
+ ** that are not yielding  useful instantiations. of quantifier instantiation.
+ ** This is used in the term_tuple_enumerator.
+ **/
+#include "theory/quantifiers/index_trie.h"
+
+namespace CVC4 {
+namespace theory {
+namespace quantifiers {
+
+void IndexTrie::add(const std::vector<bool>& mask,
+                    const std::vector<size_t>& values)
+{
+  const size_t cardinality = std::count(mask.begin(), mask.end(), true);
+  if (d_ignoreFullySpecified && cardinality == mask.size())
+  {
+    return;
+  }
+
+  d_root = addRec(d_root, 0, cardinality, mask, values);
+}
+
+void IndexTrie::freeRec(IndexTrieNode* n)
+{
+  if (!n)
+  {
+    return;
+  }
+  for (auto c : n->d_children)
+  {
+    freeRec(c.second);
+  }
+  freeRec(n->d_blank);
+  delete n;
+}
+
+bool IndexTrie::findRec(const IndexTrieNode* n,
+                        size_t index,
+                        const std::vector<size_t>& members,
+                        size_t& nonBlankLength) const
+{
+  if (!n || index >= members.size())
+  {
+    return true;  // all elements of members matched
+  }
+  if (n->d_blank && findRec(n->d_blank, index + 1, members, nonBlankLength))
+  {
+    return true;  // found in the blank branch
+  }
+  nonBlankLength = index + 1;
+  for (const auto& c : n->d_children)
+  {
+    if (c.first == members[index]
+        && findRec(c.second, index + 1, members, nonBlankLength))
+    {
+      return true;  // found in the matching subtree
+    }
+  }
+  return false;
+}
+
+IndexTrieNode* IndexTrie::addRec(IndexTrieNode* n,
+                                 size_t index,
+                                 size_t cardinality,
+                                 const std::vector<bool>& mask,
+                                 const std::vector<size_t>& values)
+{
+  if (!n)
+  {
+    return nullptr;  // this tree matches everything, no point to add
+  }
+  if (cardinality == 0)  // all blanks, all strings match
+  {
+    freeRec(n);
+    return nullptr;
+  }
+
+  Assert(index < mask.size());
+
+  if (!mask[index])  // blank position in the added vector
+  {
+    auto blank = n->d_blank ? n->d_blank : new IndexTrieNode();
+    n->d_blank = addRec(blank, index + 1, cardinality, mask, values);
+    return n;
+  }
+  Assert(cardinality);
+
+  for (auto& edge : n->d_children)
+  {
+    if (edge.first == values[index])
+    {
+      // value already amongst the children
+      edge.second =
+          addRec(edge.second, index + 1, cardinality - 1, mask, values);
+      return n;
+    }
+  }
+  // new child needs to be added
+  auto child =
+      addRec(new IndexTrieNode(), index + 1, cardinality - 1, mask, values);
+  n->d_children.push_back(std::make_pair(values[index], child));
+  return n;
+}
+}  // namespace quantifiers
+}  // namespace theory
+}  // namespace CVC4
--- a/src/theory/quantifiers/index_trie.h
+++ b/src/theory/quantifiers/index_trie.h
@@ -1,0 +1,110 @@
+/*********************                                                        */
+/*! \file index_trie.h
+ ** \verbatim
+ ** Top contributors (to current version):
+ **   Mikolas Janota
+ ** This file is part of the CVC4 project.
+ ** Copyright (c) 2009-2021 by the authors listed in the file AUTHORS
+ ** in the top-level source directory and their institutional affiliations.
+ ** All rights reserved.  See the file COPYING in the top-level source
+ ** directory for licensing information.\endverbatim
+ **
+ ** \brief Implementation of a trie that store subsets of tuples of term indices
+ ** that are not yielding  useful instantiations. of quantifier instantiation.
+ ** This is used in the term_tuple_enumerator.
+ **/
+#ifndef CVC4__THEORY__QUANTIFIERS__INDEX_TRIE_H
+#define CVC4__THEORY__QUANTIFIERS__INDEX_TRIE_H
+#include <algorithm>
+#include <utility>
+#include <vector>
+
+#include "base/check.h"
+
+namespace CVC4 {
+namespace theory {
+namespace quantifiers {
+
+/** A single node of the IndexTrie. */
+struct IndexTrieNode
+{
+  std::vector<std::pair<size_t, IndexTrieNode*>> d_children;
+  IndexTrieNode* d_blank = nullptr;
+};
+
+/** Trie of  sequences indices, used to check for subsequence membership.
+ *
+ * The  data structure stores tuples of indices where some elements may be
+ * left blank. The objective is to enable checking whether a given, completely
+ * filled in, tuple has a  sub-tuple  present in the data structure.  This is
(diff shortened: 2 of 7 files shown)


src/theory/quantifiers/index_trie.h
```diff
@@ -0,0 +1,76 @@
...
+ **/
+#ifndef INDEX_TRIE_H_45
+#define INDEX_TRIE_H_45
+#include <algorithm>
+#include <utility>
+#include <vector>
+
+#include "base/check.h"
+
+namespace CVC4 {
```
ajreynol, line 24: Our namespaces (mostly) follow directory structure, so this should use 3 nested namespace, e.g. https://github.com/ajreynol/CVC4/blob/master/src/theory/quantifiers/alpha_equivalence.h#L24

src/theory/quantifiers/index_trie.h
```diff
@@ -0,0 +1,76 @@
...
+struct IndexTrieNode
+{
+  std::vector<std::pair<size_t, IndexTrieNode*>> d_children;
+  IndexTrieNode* d_blank = nullptr;
+};
+
+/** Trie  of indices, used to check subsumption.
```
ajreynol, line 31: Can you document this class a bit more? It's not clear what the use case is at first glance.

src/theory/quantifiers/inst_strategy_enumerative.cpp
```diff
@@ -175,189 +176,52 @@ void InstStrategyEnum::check(Theory::Effort e, QEffort quant_e)
...
+  ttec.d_quantEngine = d_quantEngine;
+  ttec.d_rd = d_rd;
+  ttec.d_fullEffort = fullEffort;
+  ttec.d_increaseSum = options::fullSaturateSum();
+  ttec.d_isRd = isRd;
+  std::unique_ptr<TermTupleEnumeratorInterface> enumerator(
+      mkTermTupleEnumerator(quantifier, &ttec));
+  std::vector<Node> terms;
+  std::vector<bool> failMask;
+  Instantiate* const ie = d_quantEngine->getInstantiate();
```
ajreynol, line 198:
```suggestion
  Instantiate* ie = d_quantEngine->getInstantiate();
```
As a style preference, I don't add const to a pointer like this.

src/theory/quantifiers/term_tuple_enumerator.h
```diff
@@ -0,0 +1,57 @@
...
+ ** Copyright (c) 2009-2020 by the authors listed in the file AUTHORS
+ ** in the top-level source directory and their institutional affiliations.
+ ** All rights reserved.  See the file COPYING in the top-level source
+ ** directory for licensing information.\endverbatim
+ **
+ ** \brief Implementation of an enumeration of tuples of terms for the purpose
+ *of quantifier instantiation.
+ **/
+#ifndef TERM_TUPLE_ENUMERATOR_H_7640
+#define TERM_TUPLE_ENUMERATOR_H_7640
```
ajreynol, line 16:
```suggestion
#ifndef CVC4__THEORY__QUANTIFIERS__TERM_TUPLE_ENUMERATOR_H
#define CVC4__THEORY__QUANTIFIERS__TERM_TUPLE_ENUMERATOR_H
```

src/theory/quantifiers/term_tuple_enumerator.h
```diff
@@ -0,0 +1,57 @@
...
+#include <vector>
+
+#include "smt/smt_statistics_registry.h"
+#include "theory/quantifiers/relevant_domain.h"
+#include "theory/quantifiers_engine.h"
```
ajreynol, line 21: Can you forward declare QuantifiersEngine and then include it in the cpp?  We are trying to eliminate our use of this include.

src/theory/quantifiers/term_tuple_enumerator.cpp
```diff
@@ -0,0 +1,499 @@
...
+    {
+      Trace(trace) << "_ ";
+    }
+  }
+  Trace(trace) << "]" << std::endl;
+}
+
+namespace theory {
+namespace quantifiers {
+/**
```
ajreynol, line 66: I generally prefer larger documentation to be in the h file, perhaps this could this go above the interface class?

src/theory/quantifiers/index_trie.h
```diff
@@ -0,0 +1,76 @@
+/*********************                                                        */
+/*! \file index_trie.cpp
```
ajreynol, line 2:
```suggestion
/*! \file index_trie.h
```

src/theory/quantifiers/index_trie.h
```diff
@@ -0,0 +1,76 @@
...
+ ** in the top-level source directory and their institutional affiliations.
+ ** All rights reserved.  See the file COPYING in the top-level source
+ ** directory for licensing information.\endverbatim
+ **
+ ** \brief Implementation of a trie that store subsets of tuples of term indices
+ ** that are not yielding  useful instantiations. of quantifier instantiation.
+ ** This is used in the term_tuple_enumerator.
+ **/
+#ifndef INDEX_TRIE_H_45
+#define INDEX_TRIE_H_45
```
ajreynol, line 17:
```suggestion
#ifndef CVC4__THEORY__QUANTIFIERS__INDEX_TRIE_H
#define CVC4__THEORY__QUANTIFIERS__INDEX_TRIE_H
```

src/theory/quantifiers/index_trie.cpp
```diff
@@ -0,0 +1,113 @@
...
+ ** All rights reserved.  See the file COPYING in the top-level source
+ ** directory for licensing information.\endverbatim
+ **
+ ** \brief Implementation of a trie that store subsets of tuples of term indices
+ ** that are not yielding  useful instantiations. of quantifier instantiation.
+ ** This is used in the term_tuple_enumerator.
+ **/
+#include "theory/quantifiers/index_trie.h"
+
+namespace CVC4 {
```
ajreynol, line 18: This should be updated to 3 nested namespace CVC4, theory, quantifiers.

src/theory/quantifiers/term_tuple_enumerator.cpp
```diff
@@ -0,0 +1,499 @@
...
+#include "base/output.h"
+#include "smt/smt_statistics_registry.h"
+#include "theory/quantifiers/index_trie.h"
+#include "theory/quantifiers/quant_module.h"
+#include "theory/quantifiers/term_util.h"
+#include "util/statistics_registry.h"
+
+namespace CVC4 {
+
+// TODO: kosher?
```
ajreynol, line 33: You can delete this TODO, this is fine as a utility in the cpp.

src/theory/quantifiers/term_tuple_enumerator.cpp
```diff
@@ -0,0 +1,499 @@
...
+  bool nextCombinationSum();
+  /** Find the next lexicographically smallest combination of terms that change
+   * on the change prefix and their sum is equal to d_currentStage. */
+  bool nextCombinationMax();
+  /** Set up terms for given variable.  */
+  virtual size_t prepareTerms(size_t variableIx) = 0;
+  /** Get a given term for a given variable.  */
+  virtual Node getTerm(size_t variableIx,
+                       size_t term_index) CVC4_WARN_UNUSED_RESULT = 0;
+};  // namespace quantifiers
```
ajreynol, line 156:
```suggestion
};
```

ajreynol (changes requested):
Excited to have this in.  I have some initial requests related to style.

src/theory/quantifiers/inst_strategy_enumerative.cpp
MikolasJanota: I feel like this is somewhat inconsistent. Wouldn't it make more sense if either you never use `const` or you try to use it as much as possible? Note, however, that `const` is already used in the code base in this particular situation (see `grep -r '\* \<const\>' src `).

ajreynol: You're right, there are some inconsistencies on this point (although most code that uses this style are old).  This is a minor point, but the reason why I don't like `const` in this case is that non-constant methods are being called on `ie` (e.g. https://github.com/CVC4/CVC4/pull/6030/files#diff-9d139abfe6d05092d511906633375cde3335cb3bb8f00be3d72f4b46a066e23aR210), so it's a bit confusing to say we have a "const pointer" here, I guess this means that the value of the pointer does not change (but the underlying class may)?

FYI, the reference on style guidelines is here: https://github.com/CVC4/CVC4/wiki/Code-Style-Guidelines although subtle points like this are generally up for debate.

MikolasJanota: Yes, you're right it just means that the local variable isn't going to change while the contents pointed to may change. It's just as saying `const int i  = 0;`. Anyhow, I'll get rid of it.

ajreynol:
Results from 60s timeout run on all quantified logics involving UF or A:
```
                     CVC4-030221-mjt_nem   59592   51285.1    1492   157.734     61084     87041
                 CVC4-030221-mjt_nem_fss   59534   46138.9    1492   157.786     61026     87041
                         CVC4-030221_nem   59416   52241.1    1492   158.277     60908     87041




For unsatisfiable : 
                         CVC4-030221_nem       X      12     242
                     CVC4-030221-mjt_nem     188       X     269
                 CVC4-030221-mjt_nem_fss     360     211       X
```
CVC4-030221-mjt = this PR,
CVC4-030221 = master,
nem = --no-e-matching --full-saturate-quant,
nem_fss = --no-e-matching --full-saturate-quant --fs-sum

Summary is this looks great, the new implementation is a clear win on the old (+188,-12). Moreover, `--fs-sum` is nicely orthogonal (+211-269) but not better than default.

Thus this LGTM.  FYI @HanielB .

HanielB:
Results are nice! Great to have this. :)

src/theory/quantifiers/index_trie.h
```diff
@@ -0,0 +1,110 @@
+/*********************                                                        */
+/*! \file index_trie.h
+ ** \verbatim
+ ** Top contributors (to current version):
+ **   Mikolas Janota
+ ** This file is part of the CVC4 project.
+ ** Copyright (c) 2009-2020 by the authors listed in the file AUTHORS
```
ajreynol, line 7:
```suggestion
 ** Copyright (c) 2009-2021 by the authors listed in the file AUTHORS
```

src/theory/quantifiers/term_tuple_enumerator.h
```diff
@@ -0,0 +1,89 @@
+/*********************                                                        */
+/*! \file  term_tuple_enumerator.cpp
+ ** \verbatim
+ ** Top contributors (to current version):
+ **   Mikolas Janota
+ ** This file is part of the CVC4 project.
+ ** Copyright (c) 2009-2020 by the authors listed in the file AUTHORS
```
ajreynol, line 7:
```suggestion
 ** Copyright (c) 2009-2021 by the authors listed in the file AUTHORS
```

src/theory/quantifiers/term_tuple_enumerator.cpp
```diff
@@ -0,0 +1,501 @@
+/*********************                                                        */
+/*! \file  term_tuple_enumerator.cpp
+ ** \verbatim
+ ** Top contributors (to current version):
+ **   Mikolas Janota
+ ** This file is part of the CVC4 project.
+ ** Copyright (c) 2009-2020 by the authors listed in the file AUTHORS
```
ajreynol, line 7:
```suggestion
 ** Copyright (c) 2009-2021 by the authors listed in the file AUTHORS
```

ajreynol (changes requested):
Last minor requests.

ajreynol (approved):

PR #6031, opened 2021-03-01
Clean up includes to reduce compile times

nafur:
This PR cleans up a ton of includes, based on the suggestions of iwyu.
Mostly, it removes includes from header files in favor of forward declarations and adds includes to  source files.


nafur:
I did some experiments on my local machine, comparing compilation of this PR and master with clang in debug mode, and with gcc in production mode (both with `make -j6`).
As we can see the cpu time spent on compilation was reduced by almost 2 minutes for clang/debug and even 4:30 minutes for gcc/production.

`clang debug`
```
PR
real	5m40,002s
user	31m9,598s
sys	1m34,578s

master
real	6m4,541s
user	33m4,545s
sys	1m36,145s
```
`g++ production`
```
PR
real	8m29,927s
user	44m59,269s
sys	2m16,042s

master
real	9m17,933s
user	49m26,726s
sys	2m23,784s
```

ajreynol (commented):
This looks great to me, I agree with the changes to quantifiers engine, and to the use of forward declarations. Much earlier revisions of QuantifiersEngine had forward declarations done in this way.  Notice that most of those forward declarations will disappear entirely once the dependencies of quantifiers are cleaned up.

I'm not an expert on the changes to the tag collection, so it'd be good to have another review.

nafur:
> I'm not an expert on the changes to the tag collection, so it'd be good to have another review.

This is part of another PR and will vanish from (the diff of) this PR once #5996 is merged.

ajreynol (approved):
LGTM, thanks for doing this.

Issue #5769, opened 2021-01-12
Compilation fails on Apple M1 (ANTLR)


mpreiner:
Issue originally submitted by Stephen Siegel on cvc-users list.

```
I ran into this problem trying to build CVC4 1.8 from source on macOS 11.1 (Apple M1)…

[ 93%] Building CXX object src/parser/CMakeFiles/cvc4parser.dir/antlr_line_buffered_input.cpp.o
/opt/sw/CVC4-1.8/src/parser/antlr_line_buffered_input.cpp:291:10: error: cast from
      pointer to smaller type 'ANTLR3_MARKER' (aka 'int') loses information
  while ((ANTLR3_MARKER)(input->nextChar) != seekPoint) {
         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
make[2]: *** [src/parser/CMakeFiles/cvc4parser.dir/antlr_line_buffered_input.cpp.o] Error 1

I figured out the problem.   The script get-antlr-3.4 attempts to guess whether the architecture is 32 or 64 bit, but for arm64 it guesses incorrectly (if MACHINE_TYPE is not 'x86_64' it concludes it is 32 bit).  The solution is to read the output from this script:

 ============== WARNING ====================
The script guessed that this machine is 32 bit.
If antlr should be built as 64 bit (i.e. -m64),
please rerun the script as
MACHINE_TYPE="x86_64" ./get-antlr3.4

[note: it should be get-antlr-3.4].   After doing what it said, the build completed and all tests passed.   It might be a good idea to update this script as arm64 becomes more widely used.

```

mpreiner:
We decided to add explicit machine type checks for ARM/Apple M1. If a machine type could not be recognized we should fail instead of printing a warning.

nafur:
This issue should be resolved by #6033.

PR #6033, opened 2021-03-01
Add aarch64 (ARM64) cross-compile support.

### LINKED ISSUE #1479: lib vs lib64 when downloading antlr-3.4
Both contrib/get-antlr-3.4 and cvc4's configure assume that LIBANTLR3C_INSTALL_DIR will contain a sub-folder "lib". However, (at least one) 64 bit installation left behind only "lib64". This prevents contrib/get-antlr-3.4 from finishing and configure from confirming antlr-3.4 is in place.

mpreiner:
This commit adds support for cross-compiling for aarch64 platforms and simplifies cross-compilation handling for Windows. The configure script now automatically downloads and cross-compiles the required dependencies ANTLR3 and GMP when passing option `--arm64` or `--win64`.

Fixes #1479 #5769.


nafur (approved):
LGTM.

contrib/get-antlr-3.4
```diff
@@ -43,63 +47,58 @@ install_bin "$ANTLR_HOME_DIR/bin/antlr3"
 setup_dep \
   "https://www.antlr3.org/download/C/libantlr3c-3.4.tar.gz" \
   "$ANTLR_HOME_DIR/libantlr3c-3.4"
-cd "$ANTLR_HOME_DIR/libantlr3c-3.4"
+cd "$ANTLR_HOME_DIR/libantlr3c-3.4" || exit 1
+
+# Add aarch64 architecture
+sed 's/avr32 \\$/avr32 | aarch64 \\/' config.sub > config.sub.new
```
aniemetz, line 53: Why not use `sed -i`?

src/theory/quantifiers/fmf/bounded_integers.cpp
```diff
@@ -786,7 +786,7 @@ bool BoundedIntegers::getBoundElements( RepSetIterator * rsi, bool initial, Node
         {
           long rr = range.getConst<Rational>().getNumerator().getLong()+1;
           Trace("bound-int-rsi")  << "Actual bound range is " << rr << std::endl;
-          for( unsigned k=0; k<rr; k++ ){
+          for( long k=0; k<rr; k++ ){
```
aniemetz, line 789: Format. Can you use fixed-width types here?

aniemetz (changes requested):

contrib/get-antlr-3.4
mpreiner: to make it portable for macOS sed.

src/theory/quantifiers/fmf/bounded_integers.cpp
mpreiner: I'd rather not use fixed-width types in this PR. I wanted to keep the change minimal so that ARM64 compiles with -Werror.

aniemetz (approved):

nafur (approved):

jakevossen5:
Not sure if this has the same reasoning, but will you be able to remove [this line](https://github.com/CVC4/homebrew-cvc4/blob/de029b990d5af6b79b1976acc7e205f2143ab519/Formula/cvc4.rb#L23) if it now works on arm64?

PR #6035, opened 2021-03-02
Remove uses of SExpr class.

abdoo8080:
This PR is a step towards removing `SExpr` class. It replaces `SExpr` with `std::string` for `set-info` and `set-option` commands.


ajreynol (approved):
Thanks for the PR.

The main question to consider is whether `SmtEngine::setOption` should take a `std::string` or a `Node` (of kind SEXPR) for the value argument.  The reason for std::string is that the API takes a string and thus we are just propagating that interface.  On the other hand, the SMT-LIB standard says that options can be attributes with values (which can be SEXPR).  

Thus, to me, it depends on whether we are happy with the decision for the API to take a string for `setOption`, or whether it should take a `Term` of SEXPR kind.  I'm guessing the choice to `take std::string` in `Solver::setOption` is intentional, if so, then this PR looks good to me.  Perhaps @aniemetz can weigh in?

aniemetz:
Yes, the choice was intentional. Originally, we didn't export kind SEXPR to the API. I still think it's easier to use if we keep the strings and don't require the user to create terms. I'd only switch to terms if it is absolutely necessary.

src/api/cvc4cpp.cpp
```diff
@@ -5594,6 +5594,11 @@ void Solver::resetAssertions(void) const
   CVC4_API_SOLVER_TRY_CATCH_END;
 }
 
+Node stringToSexpr(const std::string& value)
```
aniemetz, line 5597: What is this for?

abdoo8080: At first, we wanted to modify the `SmtEngine` API for the two commands to accept sexpr nodes instead of SExpr objects. However, we decided that would be an overkill when the API only allows `std::string`. This function would have translated strings to sexpr nodes and is not necessary anymore. Deleted.

aniemetz (approved):

PR #6037, opened 2021-03-02
More cleanup of includes to reduce compilation times

nafur:
Similar to #6031, this PR implements suggestions from iwyu to reduce the number of includes in header files by introducing forward declarations and moving includes to source files.

### DIFF
--- a/src/prop/prop_engine.cpp
+++ b/src/prop/prop_engine.cpp
@@ -42,8 +42,6 @@
 #include "util/resource_manager.h"
 #include "util/result.h"
 
-using namespace std;
-
 namespace CVC4 {
 namespace prop {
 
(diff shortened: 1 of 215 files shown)


nafur:
Compilation times (clang with debug):

`master`:
```
real	5m39,831s
user	30m55,109s
sys	1m30,545s
```

this PR:
```
real	5m17,991s
user	29m4,435s
sys	1m29,101s
```

src/prop/prop_engine.cpp
```diff
@@ -41,6 +41,8 @@
 #include "util/resource_manager.h"
 #include "util/result.h"
 
+using namespace std;
```
HanielB, line 44: why this addition? I'd rather we were removing these.

nafur: I removed one from a header file, but didn't feel like immediately fixing the whole source file.

HanielB: I see. Yeah, in that case better to leave it like this.

aniemetz: As long as it gets removed eventually. Else, I'd rather have it done in this PR since you already touched it.

nafur: Okay, I've removed it.

ajreynol (approved):

PR #6038, opened 2021-03-02
Remove obsolete dependency on CxxTest.

aniemetz:


mpreiner (changes requested):
Some parts that can be removed but are not in this PR:

https://github.com/CVC4/CVC4/blob/master/configure.sh#L77
https://github.com/CVC4/CVC4/blob/master/cmake/FindCxxTest.cmake
https://github.com/CVC4/CVC4/blob/master/CMakeLists.txt#L71

mpreiner:
Can you also remove cxxtest from the CI workflows: https://github.com/CVC4/CVC4/blob/master/.github/workflows/ci.yml#L67

mpreiner (approved):

PR #6040, opened 2021-03-03
(proof-new) Simplifications to arithmetic operator elimination and preprocessing

ajreynol:
Due to refactoring in theory preprocessor, which does fixed point preprocessing on created lemmas, several things can be simplified in arithmetic operator elimination.

This is required for further simplification to witness forms in the internal proof calculus.


nafur (approved):
LGTM.

PR #6042, opened 2021-03-03
Add cmake scripts for iwyu targets.

nafur:
This PR adds some utility targets that simplify the usage of iwyu (include-what-you-use) on our code base.


cmake/IWYU.cmake
```diff
@@ -0,0 +1,36 @@
...
+if(IWYU_PATH)
+    # iwyu is available
+
+    # add a target to inspect number of times headers are included.
+    add_custom_target(iwyu-list-includes
+        COMMAND grep -hor "\"#include .*\""  ${PROJECT_SOURCE_DIR}/src | sort | uniq -c | sort -n
+    )
+
+    # find the standard library directory
+    file(GLOB_RECURSE LLVM_INCLUDE /usr/lib*/llvm*/**/stddef.h)
```
mpreiner, line 12: I'd not use GLOB_RECURSE here. Finding a directory should be possible via `find_path`.

cmake/IWYU.cmake
```diff
@@ -0,0 +1,36 @@
...
+    if(LLVM_INCLUDE)
+        list(GET LLVM_INCLUDE 0 LLVM_INCLUDE)
+        get_filename_component(LLVM_INCLUDE "${LLVM_INCLUDE}" DIRECTORY)
+        set(LLVM_INCLUDE "-I${LLVM_INCLUDE}")
+    endif()
+
+    # add a target to run iwyu on all files (within the compilation database)
+    add_custom_target(iwyu-all
+        COMMAND ${IWYU_PATH} -o clang -p . -- -Xiwyu --cxx17ns ${LLVM_INCLUDE}
+    )
```
mpreiner, line 22: What happens if `LLVM_INCLUDE` is empty?

cmake/IWYU.cmake
nafur: I tried and failed. While `find_path` has some support for wildcards, it is very limited and does not cover this case. Even if we turn `lib*` into `lib` (and `lib64`), I still could not get it to work.

cmake/IWYU.cmake
nafur: On my machine, it complains about not finding `stddef.h`. I'm not really sure why that is, though.

cmake/IWYU.cmake
```diff
@@ -0,0 +1,36 @@
+find_program(IWYU_PATH iwyu_tool)
```
mpreiner, line 1:
```suggestion
find_program(IWYU_PATH NAMES iwyu_tool iwyu-tool)
```

On Arch Linux the tool is called `iwyu-tool`

cmake/IWYU.cmake
```diff
@@ -0,0 +1,36 @@
...
+
+if(IWYU_PATH)
+    # iwyu is available
```
mpreiner, line 4:
```suggestion
    # iwyu is available
    message(STATUS "Found IWYU: ${IWYU_PATH}")
```

cmake/IWYU.cmake
mpreiner: The IWYU documentation suggests to use `clang -print-resource-dir` to find the correct clang directory (https://github.com/include-what-you-use/include-what-you-use#how-to-install).

cmake/IWYU.cmake
```diff
@@ -0,0 +1,36 @@
...
+    file(GLOB subdirs ${PROJECT_SOURCE_DIR}/src/*)
+    foreach(dir ${subdirs})
+        # add a target for every subdirectory
+        file(GLOB_RECURSE source_files
+            ${dir}/*.cpp
+            ${dir}/*.cc
+        )
+        get_filename_component(dirname ${dir} NAME)
+        add_custom_target(iwyu-${dirname}
+            COMMAND ${IWYU_PATH} -o clang -j4 -p . ${source_files} -- -Xiwyu --cxx17ns ${LLVM_INCLUDE}
```
mpreiner, line 33: Not sure if `-j4` should be hard coded. If we want to use all available threads we can use `CTEST_NTHREADS`, which is defined in the top-level `CMakeLists.txt`

nafur: So what I would like to achieve is that we can do `make -j8 iwyu-all`. I've not found a way (yet), and simply removed the `-j4` for now.

cmake/IWYU.cmake
```diff
@@ -0,0 +1,40 @@
...
+    # iwyu is available
+    message(STATUS "Found IWYU: ${IWYU_PATH}")
+
+    # add a target to inspect number of times headers are included.
+    add_custom_target(iwyu-list-includes
+        COMMAND grep -hor "\"#include .*\""  ${PROJECT_SOURCE_DIR}/src | sort | uniq -c | sort -n
+    )
+
+    # find the standard library directory
+    execute_process(COMMAND clang -print-resource-dir
```
mpreiner, line 13: Might be a good idea to first check via `find_program(CLANG_PATH clang)` if clang is installed on the system and use `${CLANG_PATH}` in this command.

mpreiner (approved):
LGTM, thanks for the changes!

PR #6043, opened 2021-03-03
New C++ API: Clean up usage of internal Result.

aniemetz:
This disables the temporarily available internals of Result.
It further changes the interface for getUnknownExplanation, which now
returns an enum value instead of a string.


mpreiner (approved):
LGTM

PR #6049, opened 2021-03-04
Reimplement time limit mechanism for windows

### LINKED ISSUE #5034: Command line argument tlimit option is not working properly in windows
**Describe the bug**
`tlimit` option was working in previous Windows versions. 
```
This is CVC4 version 1.8-prerelease
compiled with GCC version 5.3.1 20160211
on Apr 14 2020 11:41:12
```
With this new version `cvc4-2020-08-18-win64-opt.exe`, the option `tlimit` no longer works with the test file below. It works fine in Linux. 

[test.txt](https://github.com/CVC4/CVC4/files/5182513/test.txt)

**Command line arguments**:
```
cvc4-2020-08-18-win64-opt.exe test.txt --lang=smt2 --tlimit=5000
```
**CVC4 version/commit**:
```
bin>cvc4-2020-08-18-win64-opt.exe --version
This is CVC4 version 1.9-prerelease
compiled with GCC version 5.3.1 20160211
on Aug 18 2020 11:44:08
```
**Operating system**: 
```
>systeminfo

OS Name:                   Microsoft Windows 10 Pro
OS Version:                10.0.19041 N/A Build 19041
OS Manufacturer:           Microsoft Corporation
OS Configuration:          Standalone Workstation
OS Build Type:             Multiprocessor Free
System Type:               x64-based PC
Processor(s):              1 Processor(s) Installed.
                           [01]: Intel64 Family 6 Model 44 Stepping 2 GenuineIntel ~2395 Mhz
Total Physical Memory:     16,367 MB
Available Physical Memory: 7,530 MB
Virtual Memory: Max Size:  20,904 MB
Virtual Memory: Available: 3,969 MB
Virtual Memory: In Use:    16,935 MB
```
**`configure.sh` options**
```
https://cvc4.cs.stanford.edu/downloads/builds/win64-opt/unstable/cvc4-2020-08-18-win64-opt.exe
```

**`configure.sh` output**
```
https://cvc4.cs.stanford.edu/downloads/builds/win64-opt/unstable/cvc4-2020-08-18-win64-opt.exe
```

nafur:
As noted in #5034, `--tlimit` is not working properly on windows. It turns out that the timer mechanism provided by the windows API are not suitable for our use case. Thus, this PR implements a generic std::thread-based timer mechanism which is used whenever the POSIX timers (`setitimer`) are not available. It also adds some documentation on the timer options and the reasons we ended up with this.
Fixes #5034.


nafur:
I've tested this new implementation on linux only. Given that it only uses standard C++ primitives, I'm reasonably sure that it works on windows as well, but we should only merge it when someone has actually tested it on windows...

mudathirmahgoub:
I tried to compile for Windows, but it seems std::thread in c++11 is not supported yet by MinGW which is a surprise. 
See https://stackoverflow.com/questions/21211980/mingw-error-thread-is-not-a-member-of-std
```
src/main/time_limit.cpp: In function ‘CVC4::main::TimeLimit CVC4::main::install_time_limit(const CVC4::Options&)’:
src/main/time_limit.cpp:88:8: error: ‘thread’ is not a member of ‘std’
   88 |   std::thread t([ms]()
```

mudathirmahgoub:
I tested in Windows and it works fine now with posix thread model.

src/main/time_limit.cpp
```diff
@@ -12,12 +12,40 @@
...
  **
  ** Implementation of time limits that are imposed by the --tlimit option.
+ **
+ ** There are various strategies to implement time limits, with different
+ ** advantages and disadvantages:
+ **
+ ** std::thread: we can spawn a new thread which waits for the time limit.
+ ** Unless we use std::jthread (from C++20), std::thread is not interruptible
+ ** and thus we need a synchronization mechanism so that the main thread can
+ ** communicate to the timer thread that is wants to finish. Apparently, this
```
mudathirmahgoub, line 22:
```suggestion
 ** communicate to the timer thread that it wants to finish. Apparently, this
```

CMakeLists.txt
```diff
@@ -352,6 +352,10 @@ enable_testing()
 #-----------------------------------------------------------------------------#
 # Check options, find packages and configure build.
 
+# required for time limits on non-POSIX systems
+find_package(Threads REQUIRED)
```
mpreiner, line 356: If this is only required for the Windows cross-compilation we should guard this and not require it for all builds.

mudathirmahgoub (approved):
LGTM for c++17. Thanks for resolving the issue. 
I guess We should get back to this when we move to c++20.

CMakeLists.txt
nafur: I agree. I've added a check for the availability of `setitimer` and moved this into `ConfigureCVC4.cmake`

PR #6051, opened 2021-03-04
(proof-new) Replace witness form by original form in the internal proof calculus 

ajreynol:
This makes a simplification to the internal proof calculus.  In particular, purification skolems are no longer are special case of witness skolems.  The concept of "witness form" is replaced in most places by "original form".

This is required for fixing two issues:
(1) variable shadowing issues in skolemization.
(2) complications to rules like string reduction which previously would require external bookkeeping of bound variables introduced for witness terms.  This made the LFSC proof conversion extremely cumbersome for e.g. string reductions.

In this PR, the main changes:
- The internals of SkolemManager are changed to use original form vs witness form when necessary. This eliminates the need to do variable renaming in SkolemManager::skolemize.
- the rule WITNESS_INTRO is replaced by SKOLEM_INTRO
- MACRO_SR_* rules use original form
- Proof post processing is simplified

These changes imply that ppRewrite should not return WITNESS terms.  Followup PRs will modify arithmetic preprocessing so that its ppRewrite returns skolems instead.

### DIFF
--- a/src/expr/skolem_manager.cpp
+++ b/src/expr/skolem_manager.cpp
@@ -22,16 +22,7 @@ using namespace CVC4::kind;
 
 namespace CVC4 {
 
-/**
- * Attribute for associating terms to a unique bound variable.  This
- * is used to construct canonical bound variables e.g. for constructing
- * bound variables for witness terms in the skolemize method below.
- */
-struct WitnessBoundVarAttributeId
-{
-};
-typedef expr::Attribute<WitnessBoundVarAttributeId, Node>
-    WitnessBoundVarAttribute;
+// witness, original are analogous, but share skolems
 
 // Attributes are global maps from Nodes to data. Thus, note that these could
 // be implemented as internal maps in SkolemManager.
@@ -45,11 +36,15 @@ struct SkolemFormAttributeId
 };
 typedef expr::Attribute<SkolemFormAttributeId, Node> SkolemFormAttribute;
 
-// Maps terms to their purify skolem variables
-struct PurifySkolemAttributeId
+struct OriginalFormAttributeId
 {
 };
-typedef expr::Attribute<PurifySkolemAttributeId, Node> PurifySkolemAttribute;
+typedef expr::Attribute<OriginalFormAttributeId, Node> OriginalFormAttribute;
+
+struct SkolemLemmaAttributeId
+{
+};
+typedef expr::Attribute<SkolemLemmaAttributeId, Node> SkolemLemmaAttribute;
 
 Node SkolemManager::mkSkolem(Node v,
                              Node pred,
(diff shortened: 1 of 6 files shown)


ajreynol:
The documentation is above SkolemManager::skolemize (https://github.com/CVC4/CVC4/blob/master/src/expr/skolem_manager.h#L108).  Do you mean you want this documentation inlined in proof_rule.h as well?

src/expr/skolem_manager.cpp
```diff
@@ -22,16 +22,7 @@ using namespace CVC4::kind;
...
- * Attribute for associating terms to a unique bound variable.  This
- * is used to construct canonical bound variables e.g. for constructing
- * bound variables for witness terms in the skolemize method below.
- */
-struct WitnessBoundVarAttributeId
-{
-};
-typedef expr::Attribute<WitnessBoundVarAttributeId, Node>
-    WitnessBoundVarAttribute;
+// witness, original are analogous, but share skolems
```
HanielB, line 25: this seems out of place

src/expr/skolem_manager.cpp
```diff
@@ -59,25 +49,31 @@ Node SkolemManager::mkSkolem(Node v,
...
+  // We do not currently insist that pred does not contain witness terms
   Assert(v.getKind() == BOUND_VARIABLE);
   // make the witness term
   NodeManager* nm = NodeManager::currentNM();
   Node bvl = nm->mkNode(BOUND_VAR_LIST, v);
-  // translate pred to witness form, since pred itself may contain skolem
-  Node predw = getWitnessForm(pred);
-  // make the witness term, which should not contain any skolem
-  Node w = nm->mkNode(WITNESS, bvl, predw);
+  // Make the witness term, where notice that pred may contain skolem. We do
```
HanielB, line 57:
```suggestion
  // Make the witness term, where notice that pred may contain skolems. We do
```

src/expr/skolem_manager.cpp
```diff
@@ -176,20 +162,16 @@ Node SkolemManager::mkPurifySkolem(Node t,
...
-  // directly.
-  if (t.getKind() == WITNESS)
-  {
-    return getOrMakeSkolem(getWitnessForm(t), prefix, comment, flags);
-  }
-  Node v = NodeManager::currentNM()->mkBoundVar(t.getType());
-  Node k = mkSkolem(v, v.eqNode(t), prefix, comment, flags);
-  t.setAttribute(psa, k);
+  Node to = getOriginalForm(t);
+  // We do not currently insist that to does not contain witness terms
```
HanielB, line 166: seems out of place

src/expr/skolem_manager.cpp
```diff
@@ -59,25 +49,31 @@ Node SkolemManager::mkSkolem(Node v,
...
   // make the witness term
   NodeManager* nm = NodeManager::currentNM();
   Node bvl = nm->mkNode(BOUND_VAR_LIST, v);
-  // translate pred to witness form, since pred itself may contain skolem
-  Node predw = getWitnessForm(pred);
-  // make the witness term, which should not contain any skolem
-  Node w = nm->mkNode(WITNESS, bvl, predw);
+  // Make the witness term, where notice that pred may contain skolem. We do
+  // not recursively convert pred to witness form, since witness terms should
+  // be treated as opaque. Moreover, the use of witness forms leads to
```
HanielB, line 59: What is the impact of this in terms of `getWitnessForm`? If one has a skolem and wants to get its witness form without skolems it's their responsibility to do the recursive transformation?

ajreynol: Yes, which notice would be quite complicated, it would involve variable shadowing issues.

HanielB: I see. It'd be good to add a note on that for getWitnessForm documentation, saying that since it's not recursive one can get witness forms with skolems and there is no guarantee that the witnesses of these skolems can be substituted as is without leading to shadowing.

HanielB (approved):
LGTM

PR #6053, opened 2021-03-04
Add initial bit-blaster for proof logging.

aniemetz:
This adds a bit-blaster wrapper around BBSimple for proof logging.
It iteratively bit-blasts in DFS post order. The actual proof logging
will be added in the following.


mpreiner (approved):
LGTM

PR #6054, opened 2021-03-04
New C++ Api: Clean up usage of internal types in Term.

aniemetz:
This disables the temporarily available internals of Term.
Note: getExpr() is still available and will be disabled when the API is
      fully converted to Nodes.


ajreynol (approved):
LGTM

PR #6058, opened 2021-03-04
Simplify theory preprocessing

ajreynol:
Theory preprocessing now theory-preprocesses lemmas until fixed point.  This eliminates the old code for rewriting them only, which is no longer necessary as theory-preprocessing subsumes rewriting.


HanielB (approved):
LGTM

PR #6060, opened 2021-03-04
google test: Remove obsolete Expr test fixtures.

aniemetz:


test/unit/expr/attribute_black.cpp
```diff
@@ -30,7 +30,7 @@ using namespace smt;
 
 namespace test {
 
-class TestExprBlackAttribute : public TestExprBlack
+class TestExprBlackAttribute : public TestNode
```
ajreynol, line 33: Should these also be renamed TestExpr -> TestNode?

ajreynol (changes requested):

test/unit/expr/attribute_black.cpp
aniemetz: Right now the naming scheme follows the pattern Test<subfolder>(Black|White)<testname>. I was considering to rename it, but thought of postponing it to when we rename `src/expr` to `src/node`.

ajreynol (approved):

PR #6064, opened 2021-03-05
(proof-new) Disabling optimizations for unoptimized strings skolem caching

ajreynol:
This PR ensures that several optimizations are *not* performed in the reference implementation of skolem sharing (d_useOpts=false).  This is to ensure that the many of the rules in the strings proof checker do not depend on the rewriter.

These errors were caught by the LFSC proof checker.


HanielB (approved):
LGTM.

PR #6068, opened 2021-03-05
Remove SMT-LIB 2.5 and 2.0 support.

mpreiner:
This commit removes parser and printer support for old SMT-LIB standards and also converts all regression tests to 2.6.


ajreynol (approved):
LGTM, there may be other things to delete, but let's address this in a separate PR.

PR #6069, opened 2021-03-05
Remove partial UDIV/UREM operators.

mpreiner:
This commit removes the partial UDIV/UREM operator handling. BITVECTOR_UDIV and BITVECTOR_UREM are now total.


ajreynol (approved):
LGTM

PR #6074, opened 2021-03-08
Fix justification heuristic again

### LINKED ISSUE #6073: (strings-lazy-pp)Assertion error at CVC4/src/./expr/node.h:1102
59d9aad4839e64e0f
~~~~
(declare-fun v1 () Bool)
(declare-fun i2 () Int)
(declare-fun str1 () String)
(declare-fun str6 () String)
(declare-fun str7 () String)
(declare-fun str8 () String)
(assert (= 17 (str.to_int str1)))
(push)
(assert (or (= (str.++ str7 str1) (ite (= 1 (+ i2 (str.to_int str1))) (str.++ str6 str6 str7) str8)) v1))
(push)
(assert v1)
(check-sat)
~~~~
~~~~
/cvc4 -q -i delta.out.smt2 --strings-exp
Fatal failure within CVC4::NodeTemplate<ref_count>::NodeTemplate(const CVC4::NodeTemplate<ref_count>&) [with bool ref_count = false] at /home/CVC4/src/./expr/node.h:1102
Check failure

 d_nv->d_rc > 0
TNode constructed from TNode with rc == 0
Aborted (core dumped)
~~~~

nafur:
This PR replaces all TNode types in datatypes by Node within justification heuristic.
Fixes #6073.
Unfortunately, the example from #6073 times out now, thus there is no new regression.

### DIFF
--- a/src/decision/justification_heuristic.h
+++ b/src/decision/justification_heuristic.h
@@ -72,14 +72,11 @@ class JustificationHeuristic : public ITEDecisionStrategy {
    * directly. Doesn't have ones introduced during during term removal.
    */
   context::CDList<Node> d_assertions;
-  //TNode is fine since decisionEngine has them too
 
   /** map from skolems introduced in term removal to the corresponding assertion
    */
   SkolemMap d_skolemAssertions;
-  // 'key' being TNode is fine since if a skolem didn't exist anywhere,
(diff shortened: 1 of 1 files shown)


ajreynol (approved):
LGTM

PR #6078, opened 2021-03-08
contrib: Do not use HOST env variable for cross-compilation host target.

mpreiner:
On some systems HOST is set by default to the host name. This commit adds a `--host` option for the get-antlr-3.4 and get-gmp-dev scripts. It further simplifies the machine type check in get-antlr-3.4 and updates the GMP version to 6.2.1.


nafur (approved):
LGTM.

PR #6082, opened 2021-03-09
ContextObj::destroy(): Guard against invalid use.

aniemetz:
We do not allow that a context object is invalid on destruction.
This guards against invalid use as described in #2607. Note that #2607
proposed to skip invalid objects on destruction. We now rather do not
allow for such a case to occur at all.


src/context/context.cpp
```diff
@@ -238,21 +239,26 @@ ContextObj* ContextObj::restoreAndContinue()
 
 void ContextObj::destroy()
 {
+  /* The object to destroy may never be invalid. */
```
4tXJ7f, line 242: Slightly more details would be useful here. How about:

> The object to destroy must be valid, i.e. its current state must belong to a scope. We remove the object and its previous versions from their respective scopes below. If this assertion is failing, you may have created an object at a non-zero level and let it outlive the destruction of that level.

4tXJ7f (approved):
LGTM, thanks for the update.

mpreiner (approved):

aniemetz:
Just for the record, we ran this on SMT-LIB and the assertion did not fail.

PR #6088, opened 2021-03-09
ci: Enable checking of proofs + unsat cores.

mpreiner:
This commit refactors the `run_regression.py` script and adds options for enabling/disabling checking of proofs and unsat cores. Both options are enabled by default and disabled for each corresponding CI build.


test/regress/run_regression.py
```diff
@@ -370,18 +343,25 @@ def run_regression(unsat_cores, proofs, dump, use_skip_return_code,
...
                '--unconstrained-simp' not in all_args:
                 extra_command_line_args += ['--check-unsat-cores']
+            if check_proofs and \
+               '--no-check-proofs' not in all_args and \
+               '--check-proofs' not in all_args:
+                extra_command_line_args += ['--check-proofs']
         if '--no-check-abducts' not in all_args and \
-            '--check-abducts' not in all_args:
+            '--check-abducts' not in all_args and \
+            '(get-abduct)' in benchmark_content:
```
ajreynol, line 359:
```suggestion
            'get-abduct' in benchmark_content:
```
get-abduct has arguments, so I believe this won't catch anything.

ajreynol (changes requested):
Thanks for doing this, one comment.

ajreynol (commented):
This looks good to me, although I believe there are regress[1-2] failures on debug (that we are fixing https://github.com/CVC4/CVC4/pull/6051), can you check locally if debug fails `make regress`?
We should probably wait to merge before these issues are fixed, I'll update you when we're ready.

ajreynol (approved):
LGTM, thanks for the fixes.

PR #6089, opened 2021-03-09
Remove logic request

ajreynol:
This removes use of the logic request utility.

It generally bad practice to change the logic dynamically, e.g. during preprocessing, since it makes it so that CVC4 does not properly initialize.  We now insist that logic is changed upfront in set_defaults.

This is in preparation for the `smt::Env` class, which will change the ownership of logic.


src/smt/set_defaults.cpp
```diff
@@ -602,13 +608,29 @@ void setDefaults(LogicInfo& logic, bool isInternalSubsolver)
...
+        Notice() << "Enabling UF because " << logic << " requires it."
+                 << std::endl;
+      }
       log.enableTheory(THEORY_UF);
       logic = log;
       logic.lock();
     }
   }
+  if (options::arithMLTrick())
+  {
```
nafur, line 622: This may end up to enable integers in cases where the old solution would not.

src/smt/set_defaults.cpp
```diff
@@ -581,6 +581,12 @@ void setDefaults(LogicInfo& logic, bool isInternalSubsolver)
     logic = log;
     logic.lock();
   }
+  if (options::bvAbstraction())
+  {
+    // bv abstraction may require UF
+    Notice() << "Enabling UF because bvAbstraction requires it." << std::endl;
+    needsUf = true;
```
nafur, line 588: UF was only enabled if we use `bitblastmode::LAZY`, not it is always enabled with `bvAbstraction`.

nafur (commented):
This seems to widen the logic more eagerly than previously. Should we benchmark this?

src/smt/set_defaults.cpp
ajreynol: The latter I believe is implied by the former. @mpreiner @aniemetz is this right?

mpreiner: That's right `bvAbstraction` is only used by `--bv-solver=lazy`.

ajreynol:
Btw, we don't need to benchmark this as both these options (I believe) are expert-only and moreover candidates to be deleted.

src/smt/set_defaults.cpp
aniemetz: And will thus be removed.

aniemetz:
I can confirm that bvAbstraction will be deleted.

aniemetz (approved):

PR #6091, opened 2021-03-09
Improved handing of 'lib64' vs. 'lib' for glpk-cut-log and antlr-3.4

### LINKED ISSUE #1479: lib vs lib64 when downloading antlr-3.4

aytey:
PR #6033 fixed #1479 for `antlr-3.4`, such that relocating the statically-compiled archive worked correctly on systems where the install directory is `lib64` and not `lib` (e.g., such as openSUSE).

However, `glpk-cut-log` has the same issue.

This PR makes two changes:

- Moves the logic for calculating the library path from an archive name into `get-script-header.sh`
- Updates both `get-antlr-3.4` and `get-glpk-cut-log` to use this this common logic

This has been tested on openSUSE Tumbleweed 20210223.


contrib/get-script-header.sh
```diff
@@ -85,3 +85,17 @@ function install_bin
   [ ! -d "$INSTALL_BIN_DIR" ] && mkdir -p "$INSTALL_BIN_DIR"
   cp "$1" "$INSTALL_BIN_DIR"
 }
+
+function get_lib_dir
```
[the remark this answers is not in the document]
andrewvaughanj: Done!

contrib/get-script-header.sh
```diff
@@ -85,3 +85,30 @@ function install_bin
...
+   #
+   lib_name="$1"
+   lib_dir="$(dirname "$(find "$INSTALL_DIR" -name "${lib_name}")")"
+   echo ${lib_dir}
+}
+
+function rename_installed_lib
+{
+   #
+   # This function takes uses the calculated library
```
mpreiner, line 105:
```suggestion
   # This function uses the calculated library
```

andrewvaughanj: Done!

mpreiner (approved):
Awesome, thanks for the changes!

PR #6094, opened 2021-03-09
New C++ Api: Add missing argument checks in Solver functions.

aniemetz:
This adds missing checks to guard that Term and Sort arguments are
associated with the solver object that is called.


mpreiner (approved):
LGTM

PR #6095, opened 2021-03-09
Use Assert instead of assert.

mpreiner:
This commit replaces all uses of assert with Assert from `base/check.h` to ensure that all assertions get checked in production builds with enabled assertions.


aniemetz (approved):

aniemetz:
It seems that you missed some.

test/api/ouroborous.cpp
```diff
@@ -24,6 +24,7 @@
  ** below, in SMT-LIBv2 form (but they're good for all languages).
  **/
 
+#include <cassert>
```
nafur, line 27: Why are we using raw `assert` here?

aniemetz: Because this is a test that uses CVC4 but is not part of CVC4.

aniemetz (approved):

PR #6098, opened 2021-03-10
ci: Replace debug builds with assertion enabled production builds.

mpreiner:
This will help reduce the CI times and ccache sizes.


.github/workflows/ci.yml
```diff
@@ -37,15 +37,15 @@ jobs:
             os: ubuntu-latest
             exclude_regress: 1-4
 
-          - name: debug
-            config: debug --symfpu --lfsc --no-debug-symbols --editline
-            cache-key: debug
+          - name: production-dbg
+            config: production --assertions --tracing --unit-testing --symfpu --lfsc --editline
+            cache-key: production-dbg
```
nafur, line 42: This merges the caches for production and production+assertions builds, which makes sense for our dependencies.
Can it worsen the performance of the ccache setup? (significantly, under realistic circumstances)

mpreiner: I decided to separate the caches for production and production+assertions since the assertions build will trigger rebuilds for almost every source file anyways (any file that contains an Assert). The dependencies are cached separately (`deps/install` gets cached).

nafur (approved):
LGTM.

PR #6099, opened 2021-03-10
Use CVC4_ASSERTIONS instead of NDEBUG.

mpreiner:
Ensures that all checks are performed in production builds with enabled assertions.


nafur (approved):
LGTM. What about the `NDEBUG` in minisat?

mpreiner:
I'll replace the NDEBUG in the MiniSat code base too.

PR #6102, opened 2021-03-10
(proof-new) Update ppRewrite to use skolem lemmas

ajreynol:
This is required for proofs. The internal calculus no longer uses witness forms for reasoning, and hence we cannot return witness terms in ppRewrite.  This is required to fix a debug failure, as well as making life easier on external proof conversions.

As a result of this PR, for example, given (P a) as input to ppRewrite, previous we returned:

`(P (witness ((x T)) (A x)))`

now we return:

`(P k)`, with skolem lemma `( (A k), k )`

Followup PRs will remove the use of WITNESS in ppRewrite (e.g. in sets and strings); this PR modifies arithmetic to not return WITNESS in response to ppRewrite.

### DIFF
--- a/src/theory/arith/operator_elim.cpp
+++ b/src/theory/arith/operator_elim.cpp
@@ -479,9 +479,21 @@ Node OperatorElim::mkWitnessTerm(Node v,
   NodeManager* nm = NodeManager::currentNM();
   SkolemManager* sm = nm->getSkolemManager();
   // we mark that we should send a lemma
-  Node k = sm->mkSkolem(
-      v, pred, prefix, comment, NodeManager::SKOLEM_DEFAULT, this, true);
-  // TODO: (project #37) add to lems
+  Node k =
+      sm->mkSkolem(v, pred, prefix, comment, NodeManager::SKOLEM_DEFAULT, this);
+  TNode tv = v;
+  TNode tk = k;
+  Node lem = pred.substitute(tv, tk);
+  if (d_pnm != nullptr)
+  {
+    TrustNode tlem =
+        mkTrustNode(lem, PfRule::THEORY_PREPROCESS_LEMMA, {}, {lem});
+    lems.push_back(SkolemLemma(tlem, k));
+  }
+  else
+  {
+    lems.push_back(SkolemLemma(TrustNode::mkTrustLemma(lem), k));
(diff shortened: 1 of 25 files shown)


src/theory/theory_preprocessor.cpp
```diff
@@ -122,6 +121,23 @@ TrustNode TheoryPreprocessor::preprocessInternal(
     }
     Trace("tpp-debug") << "TheoryPreprocessor::preprocess: finish" << std::endl;
   }
+
+  if (procLemmas)
+  {
+    // Also must preprocess the new lemmas. This is especially important for
+    // formulas containing witness terms whose bodies are not in preprocessed
+    // form, as term formula removal introduces new lemmas for these that
+    // require theory-preprocessing.
```
nafur, line 130: Is this the reason why we can't simply push the lemmas to the inference manager?

nafur (approved):
Two questions, otherwise LGTM.

src/theory/arith/operator_elim.cpp
```diff
@@ -479,9 +479,21 @@ Node OperatorElim::mkWitnessTerm(Node v,
...
   SkolemManager* sm = nm->getSkolemManager();
   // we mark that we should send a lemma
-  Node k = sm->mkSkolem(
-      v, pred, prefix, comment, NodeManager::SKOLEM_DEFAULT, this, true);
-  // TODO: (project #37) add to lems
+  Node k =
+      sm->mkSkolem(v, pred, prefix, comment, NodeManager::SKOLEM_DEFAULT, this);
+  TNode tv = v;
+  TNode tk = k;
+  Node lem = pred.substitute(tv, tk);
```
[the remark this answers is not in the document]
ajreynol: Unfortunately this doesn't work, substitute requires TNode. 
I've had to do this annoying pattern numerous times before, I'm wondering if it is time to add a variant of substitute that takes Node.

src/theory/theory_preprocessor.cpp
ajreynol: Yes, because this is being done at a time when the inference manager is not ready.  In particular, ppRewrite is called during preprocessing where the SAT solver has not yet received any assertions.

PR #6106, opened 2021-03-10
Improve arithmetic proofs

nafur:
The proof rules for `ARITH_MULT_POS` and `ARITH_MULT_NEG` were complex than necessary in that they incorporated a rewriting step. This PR removes rewriting from these rules, making them cleaner and easier to understand.
The proof now applies these simpler rule and uses `MACRO_SR_PRED_TRANSFORM` to prove the lemma that is actually added.


alex-ozdemir (approved):
Great!

PR #6109, opened 2021-03-10
(proof-new) Clean up uses of witness with skolem lemmas

ajreynol:
This ensures we do not return WITNESS terms in response to ppRewrite.  This makes standard utility methods in SkolemLemma to help make this easy.

It also removes TheorySetsPrivate::expandDefinitions, which was returning WITNESS in response to expandDefinitions, which is no longer permitted.


nafur (approved):
LGTM.

PR #6112, opened 2021-03-10
Letify quantifier bodies independently

ajreynol:
This fixes a subtle bug with how quantifier bodies are letified.

It makes our letification more conservative so that let symbols are never pushed inside quantifier bodies, instead quantifier bodies are letified independently of the context.


src/printer/let_binding.h
```diff
@@ -69,6 +69,19 @@ namespace CVC4 {
  * d_letList[i] does not contain subterm d_letList[j] for j>i.
  * It is intended that d_letList contains only unique nodes. Each node
  * in d_letList is mapped to a unique identifier in d_letMap.
+ *
+ * Notice that this class will *not* use introduce let symbols when converting
```
HanielB, line 73:
```suggestion
 * Notice that this class will *not* use introduced let symbols when converting
```

HanielB (approved):
LGTM

PR #6113, opened 2021-03-10
Clean up ownership of Datatypes in NodeManager.

aniemetz:
This is in preparation for deleting the Expr layer.
This further fixes incorrect handling of ownership for the datatypes
owned by the NodeManager.


src/expr/node_manager.h
```diff
@@ -442,7 +434,7 @@ class NodeManager
    * would lead to memory leaks. Thus TypeNode are given a DatatypeIndexConstant
    * which is used as an index to retrieve the DType via this call.
    */
-  const DType& getDTypeForIndex(size_t index) const;
+  const DType& getDType(size_t index) const;
```
ajreynol, line 437: I'd prefer keeping the original name here, since it is indicative of the use of an index for the lookup.  This makes the method more descriptive and distinguishes it from typical uses e.g. TypeNode::getDType.

ajreynol (changes requested):
One minor request, otherwise this looks great.

ajreynol (approved):

PR #6115, opened 2021-03-11
Fix compile warnings when compiling with GLPK.

mpreiner:
Fixes -Wshadow and -Wsuggest-override wranings.


src/theory/arith/approx_simplex.cpp
```diff
@@ -580,14 +581,17 @@ static CutInfoKlass fromGlpkClass(int klass){
...
-  , d_mipProb(NULL)
-  , d_solvedRelaxation(false)
-  , d_solvedMIP(false)
+ApproxGLPK::ApproxGLPK(const ArithVariables& var,
+                       TreeLog& l,
+                       ApproximateStatistics& s)
+    : ApproximateSimplex(var, l, s),
+      d_inputProb(NULL),
+      d_realProb(NULL),
+      d_mipProb(NULL),
```
aniemetz, line 592: s/NULL/nullptr

aniemetz (approved):
Just one minor thing. Btw, the whole ArithVar being an Index which is a uint32_t makes it hard to immediately see that the loops don't create copies of complex types...

aniemetz:
> Btw, the whole ArithVar being an Index which is a uint32_t makes it hard to immediately see that the loops don't create copies of complex types...

I didn't mean to say that you should fix this (I know that was already there), I was just complaining :-)

mpreiner:
> > Btw, the whole ArithVar being an Index which is a uint32_t makes it hard to immediately see that the loops don't create copies of complex types...
> 
> I didn't mean to say that you should fix this (I know that was already there), I was just complaining :-)

Yes I know, I kept the same behavior and just fixed the warnings.

PR #6126, opened 2021-03-12
Replace HistogramStat by IntegralHistogramStat

nafur:
This PR uses IntegralHistogramStat instead of HistogramStat when appropriate, that is everywhere.


aniemetz:
Does this mean that HistogramStat is now obsolete?

nafur:
> Does this mean that HistogramStat is now obsolete?

I think so, yes.

mpreiner (approved):
If `HistogramStat` is not used anymore then we can maybe delete it (unless it will be useful in the future).

nafur:
> If `HistogramStat` is not used anymore then we can maybe delete it (unless it will be useful in the future).

I'll remove it for now. After all, reimplementing it if we need it at some point is not that difficult...

PR #6137, opened 2021-03-12
[proof-new] Making alpha-equivalence reduction proof-producing

HanielB:
Previously we would generate `THEORY_LEMMA` proof nodes for this transformation. This showed up in an example from the Isabelle/HOL integration (minimized and added as a regression).


HanielB:
Closing since this is subsumed by newer changes.

PR #6139, opened 2021-03-12
cmake: Generate cvc4_export.h and set visibility to hidden.

mpreiner:
cmake will automatically generate an export header cvc4_export.h, which makes sure that the correct export features are
defined depending on the compiler and target platform. The macro CVC4_EXPORT replaces CVC4_PUBLIC and its usage is reduced by 2/3. Visibility is set to `default` for debug builds.

Fixes https://github.com/CVC4/cvc4-projects/issues/96


src/options/mkoptions.py
```diff
@@ -168,23 +168,23 @@
 
 
 TPL_OPTION_STRUCT_RW = \
-"""extern struct CVC4_PUBLIC {name}__option_t
+"""extern struct  {name}__option_t
```
nafur, line 171:
```suggestion
"""extern struct {name}__option_t
```

src/options/mkoptions.py
```diff
@@ -168,23 +168,23 @@
...
 {{
   typedef {type} type;
   type operator()() const;
   bool wasSetByUser() const;
   void set(const type& v);
   const char* getName() const;
-}} thread_local {name} CVC4_PUBLIC;"""
+}} thread_local {name} ;"""
```
nafur, line 178:
```suggestion
}} thread_local {name};"""
```

src/options/mkoptions.py
```diff
@@ -168,23 +168,23 @@
...
 
 TPL_OPTION_STRUCT_RO = \
-"""extern struct CVC4_PUBLIC {name}__option_t
+"""extern struct  {name}__option_t
```
nafur, line 181:
```suggestion
"""extern struct {name}__option_t
```

src/options/mkoptions.py
```diff
@@ -168,23 +168,23 @@
...
 {{
   typedef {type} type;
   type operator()() const;
   bool wasSetByUser() const;
   const char* getName() const;
-}} thread_local {name} CVC4_PUBLIC;"""
+}} thread_local {name} ;"""
```
nafur, line 187:
```suggestion
}} thread_local {name};"""
```

src/options/mkoptions.py
```diff
@@ -258,9 +258,9 @@
 TPL_DECL_MODE_FUNC = \
 """
 std::ostream&
-operator<<(std::ostream& os, {type} mode) CVC4_PUBLIC;"""
+operator<<(std::ostream& os, {type} mode) ;"""
```
nafur, line 261:
```suggestion
operator<<(std::ostream& os, {type} mode);"""
```

src/options/mkoptions.py
```diff
@@ -258,9 +258,9 @@
...
 
-TPL_IMPL_MODE_FUNC = TPL_DECL_MODE_FUNC[:-len(" CVC4_PUBLIC;")] + \
+TPL_IMPL_MODE_FUNC = TPL_DECL_MODE_FUNC[:-len(" ;")] + \
```
nafur, line 263: Not sure what this is doing, but this seems wrong now.

nafur (approved):
LGTM, just a few minor formatting issues.
Do we have some documentation on what should be exported?

mpreiner:
> LGTM, just a few minor formatting issues.
> Do we have some documentation on what should be exported?

Ideally, we'd only need to export the public API and the parser. However, currently there is no clean separation between the public API and the main driver + parser. I initially removed all CVC4_PUBLIC and then added CVC4_EXPORT to classes/functions that are currently needed by the main driver + parsers. That means that the majority of CVC4_EXPORT in `src/` (excluding `src/api`) should be removed once we have a clean separation.

PR #6147, opened 2021-03-15
Make nonlinear extension account for relevant term set

### LINKED ISSUE #5662: Invalid model on NRA formula
```
[585] % cvc4 --check-models -q small.smt2
Fatal failure within void CVC4::smt::CheckModels::checkModel(CVC4::smt::Model*, CVC4::context::CDList<CVC4::NodeTemplate<true> >*, bool) at src/smt/check_models.cpp:249
assertion:   (distinct a (* d h))
simplifies to: false
expected `true'.
Aborted
[586] % 
[586] % cat small.smt2
(set-logic NRA)
(declare-fun a () Real)
(declare-fun b () Real)
(declare-fun c () Real)
(declare-fun d () Real)
(declare-fun j () Real)
(declare-fun e () Real)
(declare-fun f () Real)
(declare-fun k () Real)
(declare-fun g () Real)
(declare-fun h () Real)
(assert (forall ((i Real)) (xor (and (or (and (or (and (or (and (or
  (and (> 0.0 (* a e) (* c e)) (>= 0 k)) (<= g 0)) (= (* b j)
  2.0))))) (> f k)) (>= 0.0 k))) (>= 0 k))))
(assert (distinct a (* d h)))
(check-sat)
[587] %
```
Commit: a49efc8

ajreynol:
This fixes a subtle issue with non-linear and theory combination.

It currently could use irrelevant terms preregistered to the ExtTheory for its model-based refinement. This means that the non-linear extension could accidentally modify values for terms that are not included in its term set, violating theory combination.

In particular, in the minimized example from #5662, (* a k) was a stale term in ExtTheory, and non-linear extension falsely believed that `a` was a term whose model value could be modified, since moreover `a` was not a shared term.  In reality, `a` was not a shared term since it *only* was registered to UF.

Fixes #5662.


nafur (approved):
LGTM.

PR #6148, opened 2021-03-15
New C++ Api: Comprehensive guards for member functions of class Grammar.

aniemetz:


src/api/cvc4cpp.cpp
```diff
@@ -3096,68 +3096,72 @@ Grammar::Grammar(const Solver* slv,
...
+  //////// all checks before this line
   d_ntsToTerms[ntSymbol].push_back(rule);
+  ////////
+  CVC4_API_TRY_CATCH_END;
 }
 
 void Grammar::addRules(const Term& ntSymbol, const std::vector<Term>& rules)
 {
+  CVC4_API_TRY_CATCH_BEGIN;
+  //////// all checks before this line
```
ajreynol, line 3119: Is this out of place?

ajreynol (changes requested):
Would be good to have @abdoo8080 have a quick look as well.

src/api/checks.h
```diff
@@ -201,6 +201,103 @@ namespace api {
...
+#define CVC4_API_CHECK_TERMS(terms)                                            \
+  do                                                                           \
+  {                                                                            \
+    size_t i = 0;                                                              \
+    for (const auto& s : terms)                                                \
+    {                                                                          \
+      CVC4_API_ARG_AT_INDEX_CHECK_NOT_NULL("term", s, terms, i);               \
+      CVC4_API_ARG_AT_INDEX_CHECK_EXPECTED(                                    \
+          this->d_solver == s.d_solver, "term", terms, i)                      \
+          << "a term associated with the solver this term is associated with"; \
```
abdoo8080, line 234:
```suggestion
          << "a term associated with the solver this object is associated with"; \
```
It would be great to pass and print the name of the class as well.

src/api/cvc4cpp.cpp
aniemetz: Indeed!

src/api/checks.h
```diff
@@ -201,6 +201,103 @@ namespace api {
...
+#define CVC4_API_CHECK_TERMS_MAP(map)                                 \
+  do                                                                  \
+  {                                                                   \
+    size_t i = 0;                                                     \
+    for (const auto& p : map)                                         \
+    {                                                                 \
+      CVC4_API_ARG_AT_INDEX_CHECK_NOT_NULL("term", p.first, map, i);  \
+      CVC4_API_ARG_AT_INDEX_CHECK_EXPECTED(                           \
+          this->d_solver == p.first.d_solver, "term", map, i)         \
+          << "a term associated with this solver object";             \
```
abdoo8080, line 254:
```suggestion
          << "a term associated with the solver this object is associated with";             \
```

src/api/checks.h
```diff
@@ -201,6 +201,103 @@ namespace api {
...
+      CVC4_API_ARG_AT_INDEX_CHECK_NOT_NULL("sort", p.second, map, i); \
+      CVC4_API_ARG_AT_INDEX_CHECK_EXPECTED(                           \
+          this->d_solver == p.second.d_solver, "sort", map, i)        \
+          << "a sort associated with this solver object";             \
```
abdoo8080, line 258: Ditto.

src/api/checks.h
```diff
@@ -201,6 +201,103 @@ namespace api {
...
+#define CVC4_API_CHECK_TERMS_WITH_SORT(terms, sort)                            \
+  do                                                                           \
+  {                                                                            \
+    size_t i = 0;                                                              \
+    for (const auto& t : terms)                                                \
+    {                                                                          \
+      CVC4_API_ARG_AT_INDEX_CHECK_NOT_NULL("term", t, terms, i);               \
+      CVC4_API_ARG_AT_INDEX_CHECK_EXPECTED(                                    \
+          this->d_solver == t.d_solver, "term", terms, i)                      \
+          << "a term associated with this solver object";                      \
```
abdoo8080, line 277: Ditto.

abdoo8080 (approved):
Pointed out some typos. Otherwise, LGTM!

src/api/checks.h
aniemetz: I'd rather not pass the class name as an argument unless we really have to -- and there's no nice automatic way of doing this....

aniemetz: I mean, I can introduce macros per class, that's one thing I wanted to avoid, but if we want to get the class name in there it actually might make sense.

ajreynol (approved):

src/api/checks.h
abdoo8080: > I mean, I can introduce macros per class, that's one thing I wanted to avoid, but if we want to get the class name in there it actually might make sense.

I would prefer not to print the class name if it comes to that 😐

PR #6149, opened 2021-03-15
Eliminate more uses of SExpr.

abdoo8080:
This PR eliminates all remaining uses of `SExpr` outside of statistics.


src/smt/smt_engine.cpp
```diff
@@ -1965,59 +1965,90 @@ void SmtEngine::setIsInternalSubsolver() { d_isInternalSubsolver = true; }
...
+
+  if (atom == "true")
+  {
+    return nm->mkConst<bool>(true);
+  }
+  else if (atom == "false")
+  {
+    return nm->mkConst<bool>(false);
+  }
+  else
```
ajreynol, line 2039: Nit: you can delete this `else`

src/smt/command.cpp
```diff
@@ -1641,15 +1640,14 @@ void GetAssignmentCommand::invoke(api::Solver* solver, SymbolManager* sm)
...
-    std::vector<SExpr> sexprs;
+    std::vector<api::Term> sexprs;
     for (size_t i = 0, nterms = terms.size(); i < nterms; i++)
     {
-      std::vector<SExpr> ss;
-      ss.emplace_back(SExpr::Keyword(names[i]));
-      ss.emplace_back(SExpr::Keyword(values[i].toString()));
-      sexprs.emplace_back(ss);
+      // Put the name of the expression inside of a variable name for printing
+      api::Term name = solver->mkVar(solver->getBooleanSort(), names[i]);
```
ajreynol, line 1647: I assume its wrong to use a string constant here?  Why is that?

ajreynol (changes requested):

src/smt/command.cpp
abdoo8080: Modified the comment. Please take another look.

ajreynol (approved):
LGTM

PR #6151, opened 2021-03-15
[proof-new] Disabling proofs on regressions with known bug

HanielB:
This is so that we can use CI in master for proofs.


ajreynol (changes requested):
Actually, can you also include those taking more than 60 seconds on debug?  I think there were a few ones that become quite slow.

ajreynol (approved):

PR #6160, opened 2021-03-16
Make SEXPR simply typed

ajreynol:
Currently, SEXPR applications are given a parametric type SEXPR_TYPE applied to the types of its arguments.  This means that SEXPR that are type checked consume roughly double the memory.  This issue arises in practice when printing proofs in the internal calculus. 

There is no need to have SEXPR_TYPE as a parametric type, this PR makes SEXPR simply typed.

Also moves some implementation of TypeNode methods to type_node.cpp.


src/expr/type_node.cpp
```diff
@@ -656,7 +645,33 @@ bool TypeNode::isSortConstructor() const {
...
+
+bool TypeNode::isConstructor() const
+{
+  return getKind() == kind::CONSTRUCTOR_TYPE;
+}
+
+bool TypeNode::isSelector() const { return getKind() == kind::SELECTOR_TYPE; }
+
+bool TypeNode::isTester() const { return getKind() == kind::TESTER_TYPE; }
+
```
HanielB, line 674: You should mention these changes in the description.

HanielB (approved):
LGTM, just one minor comment

PR #6161, opened 2021-03-16
Clean up remaining raw uses of output channel

ajreynol:
After this PR, with only a few exceptions, all calls to output channel for lemmas, conflicts are made through inference manager.

This is work towards standardizing the statistics for theories.


src/theory/sep/theory_sep.cpp
```diff
@@ -128,19 +128,7 @@ Node TheorySep::mkAnd( std::vector< TNode >& assumptions ) {
...
-    Debug("sep") << "TheorySep::propagateLit(" << literal
-                 << "): already in conflict" << std::endl;
-    return false;
-  }
-  bool ok = d_out->propagate(literal);
-  if (!ok) {
-    d_state.notifyInConflict();
-  }
-  return ok;
+  return d_im.propagateLit(literal);
```
nafur, line 131: Can we remove this function and have everyone call `d_im` directly?

nafur (approved):
Just one suggestion.
Also: I think we should benchmark...

src/theory/sep/theory_sep.cpp
ajreynol: Not quite yet, the problem is that some theories like arithmetic have custom propagate methods.

ajreynol:
Virtually no difference on FP logics, SMT-LIB 60 second timeout:
```
                       CVC4-031721-rii_d   29249   10556.7   48804   4743.83     78053     80089
                           CVC4-031721_d   29249   10519.5   48805   4787.59     78054     80089



For unsatisfiable : 
                           CVC4-031721_d       X       0
                       CVC4-031721-rii_d       0       X

For satisfiable : 
                           CVC4-031721_d       X       1
                       CVC4-031721-rii_d       0       X
```

PR #6163, opened 2021-03-17
(proof-new) Fixes to set defaults

ajreynol:
This copies most of the remaining changes to set_defaults.cpp from proof-new.

In particular, it recognizes when proofs must be disabled.

This is required to fix the regressions (locally) and the nightlies.


mpreiner (approved):
LGTM

PR #6167, opened 2021-03-18
 Function types are always first-class

ajreynol:
This eliminates one of the dependencies of TypeNode on options. In particular, this makes it so that function types are always considered first-class.

This change makes it so that higher-order types can always be constructed; logic exceptions should occur if they are used in constraints.


src/theory/uf/theory_uf.cpp
```diff
@@ -224,6 +226,18 @@ TrustNode TheoryUF::ppRewrite(TNode node, std::vector<SkolemLemma>& lems)
...
     }
   }
+  else if (k == kind::APPLY_UF)
+  {
+    // check for higher-order
+    // logic exception if higher-order is not enabled
+    if (isHigherOrderType(node.getOperator().getType()) && !options::ufHo())
+    {
+      std::stringstream ss;
+      ss << "UF recieved an application whose operator has higher-order type "
```
aniemetz, line 236:
```suggestion
      ss << "UF received an application whose operator has higher-order type "
```

aniemetz:
Unit tests are failing, maybe this check is not strict enough?

ajreynol:
The unit tests should be updated to *not* expect errors in several places now it looks like.

In other words, this PR is allowing for more possible use cases (intentionally).

src/theory/uf/theory_uf.h
```diff
@@ -162,7 +162,8 @@ class TheoryUF : public Theory {
                     const TNodeTrie* t2,
                     unsigned arity,
                     unsigned depth);
-
+  /** Is t a higher order type? */
```
HanielB, line 165: Say that this is checking explicitly for tn being a function type and the criterion is one of the arguments having a function type.

src/api/checks.h
```diff
@@ -429,16 +429,18 @@ namespace api {
...
+#define CVC4_API_SOLVER_CHECK_CODOMAIN_SORT(sort)           \
+  do                                                        \
+  {                                                         \
+    CVC4_API_ARG_CHECK_NOT_NULL(sort);                      \
+    CVC4_API_CHECK(this == sort.d_solver)                   \
+        << "Given sort is not associated with this solver"; \
+    CVC4_API_ARG_CHECK_EXPECTED(sort.isFirstClass(), sort)  \
+        << "first-class sort as codomain sort";             \
+    CVC4_API_ARG_CHECK_EXPECTED(!sort.isFunction(), sort)   \
+        << "function sort as codomain sort";                \
```
aniemetz, line 443: Does this mean that functions are first-class?

ajreynol: Yes, functions are first-class, but they regardless should never be the return type of a function.  This is a restriction we want to keep.

src/theory/uf/theory_uf.cpp
```diff
@@ -224,6 +226,18 @@ TrustNode TheoryUF::ppRewrite(TNode node, std::vector<SkolemLemma>& lems)
       return TrustNode::mkTrustRewrite(node, ret, nullptr);
     }
   }
+  else if (k == kind::APPLY_UF)
```
aniemetz, line 229: When can it happen that we have both cases? And -- do we still need `HO_APPLY`?

ajreynol: Both are needed.  APPLY_UF is for fully applied function applications, HO_APPLY is for partially applied function applications. Sometimes a fully applied function term needs to be equated to an HO_APPLY term as well.  This gets into the details of this paper: https://homepage.divms.uiowa.edu/~ajreynol/cade19b.pdf

aniemetz (approved):

src/theory/uf/theory_uf.cpp
HanielB: But if we lift the requirement of having APPLY_UF with higher-order stuff we could do away with HO_APPLY, no?

HanielB: (Not that it should be done in this PR...)

ajreynol: These two issues are not related actually.

That has to do with partial vs total applications.  This PR has to do with functions as arguments.

HanielB: Can we have a term with APPLY_UF have a functional return type? If we can't then I see why it's still necessary to have HO_APPLY.

ajreynol: It's true that this would eliminate the first barrier for not needing HO_APPLY. But regardless, the reason we need HO_APPLY is because of function equality.  We won't do congruence over equality between APPLY_UF *operators*. Unless there is yet more magic in the equality engine we don't realize. :)

HanielB: haha yes, I think the big motivation is practicality, but I wanted to be sure that we don't need them, we just want them. :)

PR #6162, opened 2021-03-17
Refactor / reimplement statistics

nafur:
This PR refactors how we collect statistics.
It splits the current statistic values into the values and a proxy object. The actual values now live inside the registry (making the ownership model way easier) while the proxy object are handed to whoever wants to collect a new statistic.
It also extends the C++ API to obtain and inspect the statistics.
To change the ownership, this PR needs to touch every single statistic in the whole codebase and change how it is registered.

PR #6170, opened 2021-03-18
Move statistics from the driver into the SmtEngine

nafur:
This PR does some refactoring on how we handle statistics in the driver, and some minor cleanup along the way.
- The SmtEngine now has dedicated statistics for the data collected within the driver and provides utility functions to set them.
- The driver pushes the collected statistics to the SmtEngine instead of adding them itself to the statistics registry.
- The node manager no longer holds a statistics registry (that nobody used anymore anyway)
- The command executor no longer holds a pointer to the SmtEngine itself. The pointer is not necessary and seems to become stale after we call reset on the command executor (which, luckily, we don't seem to do usually)

The main motivation for this change is to make the whole statistics infrastructure private to the library and not exporting it to the outside world.

### DIFF
--- a/src/main/signal_handlers.cpp
+++ b/src/main/signal_handlers.cpp
@@ -40,7 +40,6 @@
 #include "options/options.h"
 #include "smt/smt_engine.h"
 #include "util/safe_print.h"
-#include "util/statistics.h"
 
 using CVC4::Exception;
 using namespace std;
@@ -61,10 +60,6 @@ void print_statistics()
 {
   if (pOptions != NULL && pOptions->getStatistics() && pExecutor != NULL)
   {
-    if (pTotalTime != NULL && pTotalTime->running())
-    {
-      pTotalTime->stop();
-    }
     pExecutor->safeFlushStatistics(STDERR_FILENO);
   }
 }
--- a/src/smt/smt_engine.cpp
+++ b/src/smt/smt_engine.cpp
@@ -513,15 +513,6 @@ CVC4::SExpr SmtEngine::getInfo(const std::string& key) const
   if (key == "all-statistics")
   {
     vector<SExpr> stats;
-    StatisticsRegistry* sr = getNodeManager()->getStatisticsRegistry();
-    for (StatisticsRegistry::const_iterator i = sr->begin(); i != sr->end();
-         ++i)
-    {
-      vector<SExpr> v;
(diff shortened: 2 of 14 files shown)


src/main/signal_handlers.cpp
```diff
@@ -61,10 +60,6 @@ void print_statistics()
 {
   if (pOptions != NULL && pOptions->getStatistics() && pExecutor != NULL)
   {
-    if (pTotalTime != NULL && pTotalTime->running())
-    {
-      pTotalTime->stop();
-    }
```
nafur, line 67: In case you wonder: this has become completely unnecessary (because the timers gracefully handle this case). After all, we only stop this single timer, but do not care about other timers that may still be running elsewhere...

src/smt/smt_engine.cpp
```diff
@@ -429,6 +430,17 @@ const std::string& SmtEngine::getFilename() const
 {
   return d_state->getFilename();
 }
+
+void SmtEngine::setResultStatistic(api::Result result) {
+    std::stringstream ss;
+    ss << result;
+    d_stats->d_driverResult.set(ss.str());
```
mpreiner, line 437:
```suggestion
    d_stats->d_driverResult.set(result.toString());
```

src/smt/smt_engine.h
```diff
@@ -24,6 +24,7 @@
 #include <string>
 #include <vector>
 
+#include "api/cvc4cpp.h"
```
mpreiner, line 27: This is an absolute no go and should not be needed. The public header should never be included in the internal code.

mpreiner (changes requested):
Can you check if `CVC4_EXPORT` is still needed in `src/util/statistics_registry.h` and `src/util/statistics_base.h`?

src/smt/smt_engine.h
ajreynol: Agreed, perhaps its for `api::Result` below, you should probably extract the relevant data and make it take only internal data.

nafur:
> Can you check if `CVC4_EXPORT` is still needed in `src/util/statistics_registry.h` and `src/util/statistics_base.h`?

Most probably it is not. However, these files will be replaced anyway with #6162...

mpreiner (approved):

PR #6172, opened 2021-03-18
Make ResetCommand go through APISolver

nafur:
This PR changes `ResetCommand` to not call `SmtEngine::reset` but instead reconstruct the `api::Solver` object. This makes `SmtEngine::reset` obsolete and removes it, and moves the original options from `SmtEngine` to `api::Solver`.
The motivation for this change is twofold:
- The `ResetCommand` should not use `SmtEngine` directly, but only through the API (i.e. `Solver`).
- As soon as the statistics data lives within the registry, we need to re-register statistics after resetting the `SmtEngine`. We can now do this within the `api::Solver` constructor.


aniemetz:
Iirc we deleted this a while ago.

ajreynol:
So I think this needs further discussion.

Currently, ResetCommand needs to be fixed: it should not have the authority to call SmtEngine::reset. 

I think there are a few alternatives, including (1) what Gereon is proposing, (2) the CommandExecutor deletes the *Solver* when a reset is issued.

aniemetz:
IMO (2) is also the cleanest way to handle this.

nafur:
I've changed this PR to reconstruct the Solver object instead.
I noticed that `SmtEngine::reset()` seems obsolete now (but implements the very same pattern: https://github.com/CVC4/CVC4/blob/master/src/smt/smt_engine.cpp#L1811). Maybe we should remove it, and move `d_originalOptions` up to the `Solver` class?

ajreynol:
Yes, I agree SmtEngine::reset could be removed and d_originalOptions could be moved up to Solver.

ajreynol:
Although it works, I'm still a bit wary of ResetCommand deleting an object (Solver) that it doesn't own.  

I'd still prefer handling this at a higher level, although I'm open for discussion on this.

ajreynol (approved):
LGTM

aniemetz:
We discussed this offline and came to the conclusion that the proposed solution is quite elegant and probably the best solution for now.

PR #6173, opened 2021-03-18
Move statistics registry to env class

nafur:
This PR moves the statistics registry from `SmtEngine` to `Env`.


ajreynol (approved):
Nice, happy to have this moved to where it belongs.

PR #6174, opened 2021-03-18
Simplify strings term registration

### LINKED ISSUE #6072: Invalid model for string formula
59d9aad4839
~~~~
(declare-fun i4 () Int)
(declare-fun str5 () String)
(assert (= (str.++ str5 (str.from_int i4)) (str.++ (str.from_int 15) (str.++ str5 (str.from_int 1)))))
(push)
(check-sat)
(pop)
(check-sat)
~~~~
~~~~
cvc4 -q -i --check-models --strings-exp delta.out.smt2 
sat
Fatal failure within void CVC4::smt::CheckModels::checkModel(CVC4::smt::Model*, CVC4::context::CDList<CVC4::NodeTemplate<true> >*, bool) at /home/CVC4/src/smt/check_models.cpp:131
assertion:     (= (str.++ str5 (str.from_int i4)) (str.++ (str.from_int 15) (str.++ str5 (str.from_int 1))))
simplifies to: false
expected `true'.
Aborted (core dumped
~~~~

ajreynol:
String terms may enter into the equality engine without appearing in assertions, due to eager context-dependent simplification internal to the equality engine (--strings-eager-eval).  In rare cases, we did not catch when a new string constant appeared in the equality engine, meaning it would not be marked as relevant leading to bogus models in incremental mode.

This makes it so that proxy variables are stored in a user-context dependent manner, which impacts when terms marked as having a proxy variable are registered. 

The PR also simplifies our policies for when string terms are registered slightly.

Fixes #6072.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -565,8 +565,8 @@ if(BUILD_BINDINGS_PYTHON)
 endif()
 
 if(BUILD_BINDINGS_JAVA)
-  message(FATAL_ERROR
-    "Java bindings for the new API are not implemented yet.")
+  add_subdirectory(src/api/java)
+  message(WARNING "Java API is currently under development.")
 endif()
 
 #-----------------------------------------------------------------------------#
--- a/cmake/CVC4Config.cmake.in
+++ b/cmake/CVC4Config.cmake.in
@@ -18,7 +18,9 @@ if(NOT TARGET CVC4::cvc4)
   include(${CMAKE_CURRENT_LIST_DIR}/CVC4Targets.cmake)
 endif()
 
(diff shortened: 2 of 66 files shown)


4tXJ7f (approved):
LGTM

HanielB (approved):
LGTM too

ajreynol:
Slight performance hit on smt-lib strings 60 second timeout, +22-74, will investigate.

ajreynol:
It seems that eager registration for terms from internal equalities is a good idea.  I'm keeping this for now, although I've simplified the code.

The actual bug in this issue was related to how we store proxy variables.

With the updated PR, we are +2-3 (essentially noise) on SMT-LIB strings 60 second timeout.

PR #6176, opened 2021-03-18
BitVector: Change setBit to set the bit in place.

aniemetz:
Creating BitVectors (and deleting them) is in general expensive because
of the underlying multi-precision Integer. If possible, unnecessary
constructions and desctructions of BitVectors should be avoided.
The most common use case for `setBit` is that for an existing BitVector,
a given bit should be set to a certain value. Not doing this in place
generates unnecessary constructions and destructions of BitVectors.


mpreiner (approved):
LGTM

PR #6177, opened 2021-03-18
fp: Use setBit instead of bv or in conversion from Real.

aniemetz:
This is a minor optimization that goes into effect as soon as #6176 goes
in.


martin-cs (approved):
Fair.

PR #6179, opened 2021-03-18
Guard for non-unique skolems in term formula removal

### LINKED ISSUE #6132: Check failure at src/context/cdinsert_hashmap.h:144
Commit: 91d701d

```
$cvc4 -q --strings-exp delta.out.smt2                                              
Fatal failure within void CVC4::context::InsertHashMap<Key, Data, HashFcn>::push_back(
const Key&, const Data&) [with Key = CVC4::NodeTemplate<true>; 
Data = CVC4::theory::TrustNode; HashFcn = CVC4::NodeHashFunction] 
at  /src/./context/cdinsert_hashmap.h:144
Check failure                                                                      
                                                                                   
 !contains(k)                                                                      
                                                                                   
[1]    12977 abort      cvc4 -q --strings-exp delta.out.smt2                       
$cat delta.out.smt2                                                                
(declare-fun str () String)                                                        
(assert (= 0 (ite (= str (str.from_code 
             (ite (= 0 (ite (> (str.len (str.from_int (str.len str))) 1) 1 0)) 1 0))) 1 0)))
(check-sat)
```

ajreynol:
In rare cases, we may reuse skolems for different terms (those that are the same up to purification) due to recent changes in how skolem are generated.  This guards for this case in the term formula remover, which avoids assertion failures in cd insert hash map.

Fixes #6132.

### DIFF
--- a/src/smt/term_formula_removal.cpp
+++ b/src/smt/term_formula_removal.cpp
@@ -293,11 +293,10 @@ Node RemoveTermFormulas::runCurrent(std::pair<Node, uint32_t>& curr,
       d_skolem_cache.insert(node, skolem);
 
       // Notice that in very rare cases, two different terms may have the
-      // same purification skolem. An example is the pair of terms
-      // (ite (ite A B C) D E) and (ite k D E), where k is the skolem
-      // introduced to eliminate (ite A B C). For such cases, for simplicity,
-      // we repeat the work of constructing the assertion and proofs below.
-      // This is so that the proof for the new form of the lemma is used.
+      // same purification skolem (see SkolemManager::mkPurifySkolem) For such
+      // cases, for simplicity, we repeat the work of constructing the
+      // assertion and proofs below. This is so that the proof for the new form
+      // of the lemma is used.
 
       // The new assertion
       newAssertion = nodeManager->mkNode(
--- a/src/smt/term_formula_removal.cpp
+++ b/src/smt/term_formula_removal.cpp
@@ -505,7 +505,7 @@ Node RemoveTermFormulas::runCurrent(std::pair<Node, uint32_t>& curr,
       newLem = theory::TrustNode::mkTrustLemma(newAssertion, d_lp.get());
 
       // store in the lemma cache, if it is not already there.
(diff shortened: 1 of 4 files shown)


src/smt/term_formula_removal.cpp
```diff
@@ -292,6 +292,12 @@ Node RemoveTermFormulas::runCurrent(std::pair<Node, uint32_t>& curr,
           "a variable introduced due to term-level ITE removal");
       d_skolem_cache.insert(node, skolem);
 
+      // Notice that in very rare cases, two different terms may have the
+      // same purification skolem. An example is the pair of terms
+      // (ite (ite A B C) D E) and (ite k D E), where k is the skolem
+      // introduced to eliminate (ite A B C). For such cases, for simplicity,
+      // we repeat the work of constructing the assertion and proofs below.
```
HanielB, line 299: Say why this is not an issue.

src/smt/term_formula_removal.cpp
```diff
@@ -497,8 +503,11 @@ Node RemoveTermFormulas::runCurrent(std::pair<Node, uint32_t>& curr,
 
       newLem = theory::TrustNode::mkTrustLemma(newAssertion, d_lp.get());
 
-      // store in the lemma cache
-      d_lemmaCache.insert(skolem, newLem);
+      // store in the lemma cache, if it is not already there.
+      if (d_lemmaCache.find(skolem)==d_lemmaCache.end())
```
HanielB, line 507: clang-format

src/smt/term_formula_removal.cpp
HanielB: I think the information I was looking for here is why it's not an issue to have two different terms with the same purification skolem, which is because this will only happen when their original form is the same. As it is one wonders.

ajreynol: Ok, I think this documentation doesn't belong here then. I'll add this to `skolem_manager.h`.

HanielB (approved):
LGTM

src/smt/term_formula_removal.cpp
HanielB: That's better, thanks!

PR #6183, opened 2021-03-19
FP: Add documentation for FloatingPointLiteral constructors.

aniemetz:


martin-cs (approved):
Sounds good.  I wonder whether we should change the interface to make this kind of thing clearer?

aniemetz:
I was wondering about that too. Maybe make it private and have `FloatingPoint` be friend of `FloatingPointLiteral`?

PR #6187, opened 2021-03-22
Add skolem definition manager

ajreynol:
This creates a central utility for managing "skolem definitions", e.g. mapping between skolems and the lemmas that define their behavior.

This utility is taken from the satRlv branch.  It will also be used for the new implementation of the justification decision heuristic.

Note that this PR takes some helper functions out of term formula removal (e.g. `hasSkolems`)  Prior to this PR, these helper functions were incorrect since term formula removal does *not* account for all introduced skolems.  For instance, `Theory::ppRewrite` may introduce skolems directly.  This PR consolidates these cases into the new class, which is called from PropEngine when lemmas and assertions are added.  At the moment, the only use of this method is for CEGQI, which needs to do its own tracking of skolems in certain literals.

It also makes some minor reorganization to prop engine.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -209,6 +209,8 @@ libcvc4_add_sources(
   prop/sat_solver_factory.h
   prop/sat_solver_types.cpp
   prop/sat_solver_types.h
+  prop/skolem_def_manager.cpp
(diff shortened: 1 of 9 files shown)


4tXJ7f (approved):
LGTM

PR #6194, opened 2021-03-23
Deleting old LFSC signatures

HanielB:


mpreiner:
`ENABLE_PROOFS` should also be removed in `configure.sh`

configure.sh
```diff
@@ -411,8 +411,6 @@ cmake_opts=""
 [ $ninja != default ] && cmake_opts="$cmake_opts -G Ninja"
 [ $muzzle != default ] \
   && cmake_opts="$cmake_opts -DENABLE_MUZZLE=$muzzle"
-[ $proofs != default ] \
-  && cmake_opts="$cmake_opts -DENABLE_PROOFS=$proofs"
```
mpreiner, line 415: Sorry that was wrong from my side. This should still be kept.

HanielB: The issue is in the test directory actually, I think, let me check

HanielB: So I reverted this change but I think we actually want to delete this. The current proof infrastructure is such that there is almost nothing to compile out. I can do a separate PR to remove from the source code the tests. What do you think? Also @ajreynol ?

mpreiner: `ENABLE_PROOFS` also enabled the `CVC4_PROOF` compile flag, which is still used in the code.

HanielB: Right. What I mean is removing all of this in another PR.

ajreynol: I think this gets back to the question of whether we want to be able to compile out proof support.  I still think it might be a useful feature in the future to have this.

However, I am in favor of eliminating our current dependence on this flag, as it is I think only used for the old proof infrastructure.

HanielB: I did a proposal in #6195, PTAL.

ajreynol (approved):
LGTM

PR #6195, opened 2021-03-23
Removing unused build options and deprecated proof compile flag

HanielB:


HanielB:
As things are, CI will always fail because the `production-dbg` build uses the `--lfsc` configure flag.

test/regress/run_regression.py
```diff
@@ -12,8 +12,7 @@
 """
 Usage:
 
-    run_regression.py [--enable-proof] [--with-lfsc] [--dump]
-        [--use-skip-return-code] [wrapper] cvc4-binary
+    run_regression.py [--dump] [--use-skip-return-code] [wrapper] cvc4-binary
         [benchmark.cvc | benchmark.smt | benchmark.smt2 | benchmark.p]
```
mpreiner, line 16: This is outdated and doesn't reflect the current set of options. Maybe remove?

mpreiner (approved):
LGTM except for one minor comment.

mpreiner:
> As things are, CI will always fail because the `production-dbg` build uses the `--lfsc` configure flag.

Why not remove the `--lfsc` from CI then? :-)

test/unit/api/solver_black.cpp
```diff
@@ -1423,7 +1423,7 @@ TEST_F(TestApiBlackSolver, getUnsatCore3)
   {
     d_solver.assertFormula(t);
   }
-  Result res = d_solver.checkSat();
+  CVC4::api::Result res = d_solver.checkSat();
```
mpreiner, line 1426: Is this change really necessary? There is a `using namespace api` at the top of the file.

HanielB: I copied what was going on in the other files. There is some ambiguity on which `Result` to use because of the namespaces.

PR #6196, opened 2021-03-23
Add Solver.java to the Java API

mudathirmahgoub:
PR changes:
 - Add Solver.java and relation JNI c files
 - Update FindJUnit to download JUnit5
 - Add Java unit tests


src/api/java/CMakeLists.txt
```diff
@@ -9,20 +9,59 @@
 ## directory for licensing information.
 ##
 
-# create a directory for the cvc package
-file(MAKE_DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}/cvc")
+# copy cvc libraries to the resource directory
+file(GLOB CVC_LIBS "${CMAKE_INSTALL_PREFIX}/lib/*.so*")
+file(COPY ${CVC_LIBS} DESTINATION "${CMAKE_CURRENT_BINARY_DIR}/src/main/resources")
```
mpreiner, line 14: Is there any library other than `libcvc4.so` that you want to include?

src/api/java/CMakeLists.txt
```diff
@@ -9,20 +9,59 @@
 ## directory for licensing information.
 ##
 
-# create a directory for the cvc package
-file(MAKE_DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}/cvc")
+# copy cvc libraries to the resource directory
+file(GLOB CVC_LIBS "${CMAKE_INSTALL_PREFIX}/lib/*.so*")
```
mpreiner, line 13: I'd not recommend using glob here, it should be possible to get the location of `libcvc4` via the generator expressions `$<TARGET_FILE:cvc4>`

src/api/java/CMakeLists.txt
```diff
@@ -9,20 +9,59 @@
...
+file(COPY ${CVC_LIBS} DESTINATION "${CMAKE_CURRENT_BINARY_DIR}/src/main/resources")
+file(COPY "${CMAKE_CURRENT_LIST_DIR}/cvc" DESTINATION "${CMAKE_CURRENT_BINARY_DIR}/src/main/java")
+file(COPY "/home/mudathir/Desktop/CVC4/mudathir/test/api/java/cvc" DESTINATION "${CMAKE_CURRENT_BINARY_DIR}/src/test/java")
```
mpreiner, line 16: Looks like the wrong path here.

src/api/java/CMakeLists.txt
```diff
@@ -9,20 +9,59 @@
...
+
+# copy gradle files
+file(COPY "${CMAKE_CURRENT_LIST_DIR}/gradle" DESTINATION ${CMAKE_CURRENT_BINARY_DIR})
+file(COPY "${CMAKE_CURRENT_LIST_DIR}/gradlew" DESTINATION ${CMAKE_CURRENT_BINARY_DIR})
+file(COPY "${CMAKE_CURRENT_LIST_DIR}/build.gradle" DESTINATION ${CMAKE_CURRENT_BINARY_DIR})
```
mpreiner, line 21: Why do they have to be copied over?

src/api/java/CMakeLists.txt
```diff
@@ -9,20 +9,59 @@
...
+  "${PYTHON_EXECUTABLE}"
+  "${CMAKE_CURRENT_LIST_DIR}/genkinds.py"
+  --kinds-header "${PROJECT_SOURCE_DIR}/src/api/cvc4cppkind.h"
+  --kinds-file-prefix "${CMAKE_CURRENT_BINARY_DIR}/src/main/java/cvc/Kind"
   DEPENDS
-    genkinds.py
+  genkinds.py
   COMMENT
-    "Generate Kind.java"
+  "Generate Kind.java"
```
mpreiner, line 35: Please keep the indentation.

src/api/java/CMakeLists.txt
```diff
@@ -9,20 +9,59 @@
...
 )
+
+
+find_package(Java)
```
mpreiner, line 39:
```suggestion
find_package(Java REQUIRED)
```

src/api/java/CMakeLists.txt
```diff
@@ -9,20 +9,59 @@
...
+include(UseJava)
+find_package(JNI)
```
mpreiner, line 41:
```suggestion
find_package(JNI REQUIRED)
```

src/api/java/CMakeLists.txt
```diff
@@ -9,20 +9,59 @@
...
+
+message(STATUS "JNI_INCLUDE_DIRS=${JNI_INCLUDE_DIRS}")
+include_directories(${JNI_INCLUDE_DIRS})
```
mpreiner, line 44: Do not use `include_directories` but `target_include_directories`.

src/api/java/CMakeLists.txt
```diff
@@ -9,20 +9,59 @@
...
+
+message(STATUS "JNI_LIBRARIES=${JNI_LIBRARIES}")
+link_directories(${JNI_LIBRARIES})
```
mpreiner, line 47: Better use `target_link_libraries`.

src/api/java/CMakeLists.txt
```diff
@@ -9,20 +9,59 @@
...
+
+include_directories("${CMAKE_INSTALL_PREFIX}/include")
```
mpreiner, line 49: Ditto.

src/api/java/CMakeLists.txt
```diff
@@ -9,20 +9,59 @@
...
+link_directories("${CMAKE_INSTALL_PREFIX}/lib")
```
mpreiner, line 50: Ditto.

mpreiner (changes requested):
Some initial build system related comments.

src/api/java/CMakeLists.txt
mudathirmahgoub: No, but for some reason, `libcvc4.so` does not work, only `libcvc4.so.7` works.
Actually I have to account for Mac and Windows, but for now I am focused on Linux

src/api/java/CMakeLists.txt
mudathirmahgoub: Yes. Missed this :)

src/api/java/CMakeLists.txt
mudathirmahgoub: Gradle generate these build files to avoid versions problem. Also `gradlew` allows users to run it, and it figures out the right version of Gradle to download (which needs internet).

src/api/java/CMakeLists.txt
mudathirmahgoub: Will do. clang-format does not work here. Is there any script to format this ?

src/api/java/CMakeLists.txt
mudathirmahgoub: Ok

src/api/java/CMakeLists.txt
mudathirmahgoub: Ok

src/api/java/CMakeLists.txt
mudathirmahgoub: Done

src/api/java/CMakeLists.txt
mudathirmahgoub: Done

src/api/java/CMakeLists.txt
mudathirmahgoub: These are installed jdk libraries. So target_link_libraries failed here as well.

src/api/java/CMakeLists.txt
mudathirmahgoub: Fixed

src/api/java/CMakeLists.txt
mudathirmahgoub: Done

src/api/java/CMakeLists.txt
mudathirmahgoub: I guess I misunderstood what is meant by target. It is now working.

src/api/java/CMakeLists.txt
mudathirmahgoub: Done

src/api/java/CMakeLists.txt
mudathirmahgoub: done

src/api/java/CMakeLists.txt
```diff
@@ -9,8 +9,17 @@
 ## directory for licensing information.
 ##
 
-# create a directory for the cvc package
-file(MAKE_DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}/cvc")
+# copy cvc libraries to the resource directory
+file(GLOB CVC_LIBS "${CMAKE_INSTALL_PREFIX}/lib/*.so*")
```
nafur, line 13: I actually see another issue here: this code is executed at configure time (i.e. when you run `configure.sh` or `cmake`). At this point in time, there may not even be any `*.so` files. And even if they exist (because we have built them at some point in the past) they may (and probably will) be outdated.
I think that all this code should go into a custom command or target (so that it is executed at build time), just like generating the Kind file or calling gradle.

mudathirmahgoub: The jni includes cvc installed files and libraries. Since these are generated after `make install` which is separate from the default build, the java api is build is not run when `make` is executed for now.
The `.so` files are generated with the default build, however, the include files are not as far as I know.

mpreiner: Yes, that's a good point. This should all be doable with generator expressions.

src/api/java/CMakeLists.txt
mpreiner: Again, this should be solved by just using `$<TARGET_FILE:cvc4>`

src/api/java/CMakeLists.txt
```diff
@@ -20,9 +29,35 @@ add_custom_target(
...
+include(UseJava)
+find_package(JNI REQUIRED)
+
+# specify the output destination of cvcJavaApi library
+set(CMAKE_LIBRARY_OUTPUT_DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}/src/main/resources")
+
+add_library(cvcJavaApi SHARED jni/cvc_Solver.cpp)
+
+target_include_directories(cvcJavaApi PUBLIC ${JNI_INCLUDE_DIRS})
+target_include_directories(cvcJavaApi PUBLIC "${CMAKE_INSTALL_PREFIX}/include")
```
mpreiner, line 50: Why exactly is this needed?

src/api/java/CMakeLists.txt
```diff
@@ -20,9 +29,35 @@ add_custom_target(
...
+target_link_libraries(cvcJavaApi ${JNI_LIBRARIES})
+target_link_libraries(cvcJavaApi cvc4)
+
+# run gradle
+add_custom_target(
+  gradleBuild
+  POST_BUILD
+  COMMAND
+    "cd ${CMAKE_CURRENT_BINARY_DIR}"
```
mpreiner, line 59: There should be no need for the `cd`. You should be already in `CMAKE_CURRENT_BINARY_DIR`.

mudathirmahgoub: Yes. This custom command does not execute for some reason. I added this line to see if it fixes the issue, but it did not. I removed it now.

src/api/java/CMakeLists.txt
mudathirmahgoub: Because of this [line](https://github.com/CVC4/CVC4/pull/6196/files#diff-31a7cb3b95442b3150f94f788c11f51614a5a4577d97e29f003b4ca2fc481af4R3)

mpreiner: Yes, but for that you actually need to install the CVC4 headers to a given prefix. Why not add use the headers from the source tree with `target_include_directories(cvcJavaApi PUBLIC "${PROJECT_SOURCE_DIR}/src")`?

src/api/java/CMakeLists.txt
mudathirmahgoub: Fixed

src/api/java/CMakeLists.txt
mudathirmahgoub: Changed to use `$<TARGET_FILE:cvc4>`

src/api/java/CMakeLists.txt
mudathirmahgoub: Changed so that only `build.gradle` is copied to the build dir. `gradlew` is invoked from the src directory now

src/api/java/CMakeLists.txt
mudathirmahgoub: Changed the cmake to use generater expressions.

src/api/java/CMakeLists.txt
mudathirmahgoub: Done

src/api/java/CMakeLists.txt
mudathirmahgoub: Changed to use `src/api`

src/api/java/CMakeLists.txt
```diff
@@ -9,8 +9,55 @@
...
-file(MAKE_DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}/cvc")
+# create directories
+file(MAKE_DIRECTORY
+  "${CMAKE_CURRENT_BINARY_DIR}/src/main/resources"
+  "${CMAKE_CURRENT_BINARY_DIR}/src/main/java"
+  "${CMAKE_CURRENT_BINARY_DIR}/src/test/java"
+  )
+
+# find java
+find_package(Java)
```
nafur, line 20: I would guess this is `REQUIRED` as well.

src/api/java/CMakeLists.txt
```diff
@@ -9,8 +9,55 @@
...
+# find jni package
+find_package(JNI REQUIRED)
+
+# specify the output destination of cvcjni library
+set(CMAKE_LIBRARY_OUTPUT_DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}/src/main/resources")
```
nafur, line 25: Why do we want to change this? Wouldn't this change all targets from here on (also those not related to java)?

src/api/java/gradlew
```diff
@@ -0,0 +1,185 @@
+#!/usr/bin/env sh
```
nafur, line 1: [This comment also applies to the `gradle-wrapper.jar`]

If I understand this correctly (I've never used gradle myself) these files have been generated by you using gradle.
Even though gradle describes this as a "one-time process", we should make sure that we can actually reproduce this step. I'd suggest to handle this similar to ANTLR3 (which also downloads the jar file on demand).

src/api/java/CMakeLists.txt
```diff
@@ -9,8 +9,55 @@
...
+
+target_include_directories(cvcjni PUBLIC ${JNI_INCLUDE_DIRS})
+target_include_directories(cvcjni PUBLIC ${PROJECT_SOURCE_DIR}/src)
+target_include_directories(cvcjni PUBLIC ${CMAKE_BINARY_DIR}/src/)
+target_link_libraries(cvcjni ${JNI_LIBRARIES})
+target_link_libraries(cvcjni cvc4)
+
+
+# copy libraries and java files
+add_custom_target(copy-java-api-files
```
nafur, line 37: Why are we copying these files?
Is it so that gradle finds them? Can't we make gradle tell where they are?

(I'd rather like to change gradle so that it plays nicely with cmake instead of modifying cmake to make gradle work in its default config)

nafur (commented):
Some comments...

src/api/java/CMakeLists.txt
mudathirmahgoub: It is implied by find_package(JNI REQUIRED). So far this line is redundant. I added it hoping it would fix the Mac build in github,  but it didn't.

src/api/java/CMakeLists.txt
mudathirmahgoub: Actually this is no longer needed. It is removed now

src/api/java/CMakeLists.txt
mudathirmahgoub: Cmake is used now for compiling java files. So copying is no longer needed.

src/api/java/jni/cvc_Solver.cpp
```diff
@@ -0,0 +1,49 @@
...
+/*
+ * Class:     cvc_Solver
+ * Method:    deletePointer
+ * Signature: (J)V
+ */
+JNIEXPORT void JNICALL Java_cvc_Solver_deletePointer(JNIEnv*,
+                                                     jclass,
+                                                     jlong pointer)
+{
+  std::cout << "Deleting Solver pointer: " << pointer << std::endl;
```
4tXJ7f, line 28: Remove

src/api/java/jni/cvc_Solver.h
```diff
@@ -0,0 +1,37 @@
+/* DO NOT EDIT THIS FILE - it is machine generated */
```
4tXJ7f, line 1: It would be good to generate this file on the fly.

src/api/java/jni/cvcJavaApi.h
```diff
@@ -0,0 +1,24 @@
+#ifndef CVC__JAVA_API_H
+#define CVC__JAVA_API_H
+
+#include <iostream>
```
4tXJ7f, line 4: Unused include

src/api/java/jni/cvcJavaApi.h
```diff
@@ -0,0 +1,24 @@
+#ifndef CVC__JAVA_API_H
+#define CVC__JAVA_API_H
```
4tXJ7f, line 2: Are we going to have more code in this file? We should use the usual namespaces and include guards here.

src/api/java/cvc/CVCApiException.java
```diff
@@ -0,0 +1,9 @@
+package cvc;
```
4tXJ7f, line 1: All the files should have the usual boilerplate comments.

4tXJ7f, line 1: The package name should probably be `cvc5` to be consistent with the C++ API.

4tXJ7f (changes requested):
Thanks a lot for working on this. Some initial comments.

src/api/java/jni/cvc_Solver.cpp
mudathirmahgoub: Done

src/api/java/jni/cvc_Solver.h
mudathirmahgoub: Done

src/api/java/jni/cvcJavaApi.h
mudathirmahgoub: It was used with std::cout for debugging. But now it is not used.

mudathirmahgoub: Removed

src/api/java/cvc/CVCApiException.java
mudathirmahgoub: I am in favor of just `cvc`. Any way the package name is too short and we need a proper long name like 
`com...cvc5`. This can be done in future PR when the name is finalized.

src/api/java/jni/cvcJavaApi.h
mudathirmahgoub: Perhaps more cases need to be added to this file. What would be a nice namespace for the java api
Something like `cvc5::api::java` ? 
What do you mean by guards?

src/api/java/cvc/CVCApiException.java
mudathirmahgoub: Changing package name to cvc5, would affect the manually written [files](https://github.com/CVC4/CVC4/pull/6196/files#diff-31a7cb3b95442b3150f94f788c11f51614a5a4577d97e29f003b4ca2fc481af4R13). 
JNI function names include the package name, which would need to be updated for future cvc releases.

mudathirmahgoub: I am convinced now with cvc5 in case the user wants two jar files cvc5.jar and cvc6.jar in the future.
Changed the package name to cvc5.

src/api/java/CMakeLists.txt
```diff
@@ -2,28 +2,93 @@
...
+add_dependencies(cvc5jni generate-jni-headers)
+
+target_include_directories(cvc5jni PUBLIC ${JNI_INCLUDE_DIRS})
+target_include_directories(cvc5jni PUBLIC ${PROJECT_SOURCE_DIR}/src)
+target_include_directories(cvc5jni PUBLIC ${CMAKE_BINARY_DIR}/src/)
+target_include_directories(cvc5jni PUBLIC ${CMAKE_CURRENT_BINARY_DIR}/jni)
+target_link_libraries(cvc5jni ${JNI_LIBRARIES})
+target_link_libraries(cvc5jni cvc4)
+
+set(CVC5_JAR "cvc5-${CVC4_MAJOR}.${CVC4_MINOR}.${CVC4_RELEASE}.jar" CACHE STRING "cvc5 jar" FORCE)
```
mpreiner, line 81: This variable seems to be overwritten in line 93.

mpreiner (commented):
Some more build system comments.

cmake/FindJUnit.cmake
```diff
@@ -1,36 +1,54 @@
...
+# JUnit_FOUND - should be true for testing
+# JUnit_JAR   - absolute path to JUnit5 jar file
 
-find_package(Java REQUIRED)
-include(UseJava)
-find_package(Hamcrest REQUIRED)
+include(deps-helper)
 
-find_jar(JUnit_JAR NAMES junit junit4)
+find_jar(JUnit_JAR junit-platform-console-standalone-1.7.1.jar)
```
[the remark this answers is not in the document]
mudathirmahgoub: Perhaps not. I just chose the latest version that supports junit5. 
There are some earlier versions, but I am not sure which ones support junit5. 
I am sure we don't want junit4 which is not compatible with the written tests. 
I have changed the command to include look for known places, although there is no specific path for junit5.
For junit4 there is this [link] (https://junit.org/junit4/faq.html#started_1)

cmake/FindJUnit.cmake
```diff
@@ -1,36 +1,54 @@
...
+    JUnit
+    PREFIX ${DEPS_PREFIX}
+    URL https://repo1.maven.org/maven2/org/junit/platform/junit-platform-console-standalone/1.7.1/junit-platform-console-standalone-1.7.1.jar
+    URL_HASH SHA1=99245bde65d028a8b8ff604be26e929ab6ff2e58
+    DOWNLOAD_NO_EXTRACT ON
+    CONFIGURE_COMMAND ""
+    BUILD_COMMAND ""
+    INSTALL_COMMAND ${CMAKE_COMMAND} -E copy
+    <SOURCE_DIR>/../junit-platform-console-standalone-1.7.1.jar
+    <INSTALL_DIR>/share/java/junit-platform-console-standalone-1.7.1.jar
```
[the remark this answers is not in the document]
mudathirmahgoub: done

cmake/FindJUnit.cmake
```diff
@@ -1,36 +1,54 @@
...
+
+set(JUnit_FOUND TRUE)
+
+mark_as_advanced(JUnit_JAR)
+mark_as_advanced(JUnit_FOUND)
+
+if(JUnit_FOUND_SYSTEM)
+  message(STATUS "Found JUnit: ${JUnit_JAR}")
+else()
+  message(STATUS "Building JUnit: ${JUnit_JAR}")
```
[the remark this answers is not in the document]
mudathirmahgoub: done

src/api/java/CMakeLists.txt
```diff
@@ -9,21 +9,86 @@
...
+  "${PYTHON_EXECUTABLE}"
+  "${CMAKE_CURRENT_BINARY_DIR}/genkinds.py"
+  --kinds-header "${PROJECT_SOURCE_DIR}/src/api/cpp/cvc5_kind.h"
+  --kinds-file-prefix "${CMAKE_CURRENT_BINARY_DIR}/src/main/java/cvc/Kind"
   DEPENDS
-    "${CMAKE_CURRENT_BINARY_DIR}/genkinds.py"
+  "${CMAKE_CURRENT_BINARY_DIR}/genkinds.py"
   COMMENT
-    "Generate Kind.java"
+  "Generate Kind.java"
```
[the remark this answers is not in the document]
mudathirmahgoub: Done. The editor (clion) keeps messing with cmake files.

src/api/java/CMakeLists.txt
```diff
@@ -9,21 +9,86 @@
 ## directory for licensing information.
 ##
 
-# create a directory for the cvc package
-file(MAKE_DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}/cvc")
+# create directories
+file(MAKE_DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}/src/main/java/cvc/")
```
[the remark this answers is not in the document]
mudathirmahgoub: changed to cvc5

src/api/java/CMakeLists.txt
```diff
@@ -9,21 +9,86 @@
...
-    "Generate Kind.java"
+  "Generate Kind.java"
 )
+
+# find java
+find_package(Java COMPONENTS Development REQUIRED)
+include(UseJava)
+
+# specify java source files
+set(java_files
```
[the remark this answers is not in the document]
mudathirmahgoub: Done

src/api/java/CMakeLists.txt
```diff
@@ -9,21 +9,86 @@
...
-    "${CMAKE_CURRENT_BINARY_DIR}/genkinds.py"
-    --kinds-header "${PROJECT_SOURCE_DIR}/src/api/cpp/cvc5_kind.h"
-    --kinds-file-prefix "${CMAKE_CURRENT_BINARY_DIR}/cvc/Kind"
+  "${PYTHON_EXECUTABLE}"
+  "${CMAKE_CURRENT_BINARY_DIR}/genkinds.py"
+  --kinds-header "${PROJECT_SOURCE_DIR}/src/api/cpp/cvc5_kind.h"
+  --kinds-file-prefix "${CMAKE_CURRENT_BINARY_DIR}/src/main/java/cvc/Kind"
   DEPENDS
-    "${CMAKE_CURRENT_BINARY_DIR}/genkinds.py"
+  "${CMAKE_CURRENT_BINARY_DIR}/genkinds.py"
```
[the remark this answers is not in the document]
mudathirmahgoub: Done. Thanks for the example.

src/api/java/CMakeLists.txt
mudathirmahgoub: Fixed by introducing a new new variable `CVC5_JAR_PATH`.

src/api/java/CMakeLists.txt
```diff
@@ -49,7 +57,7 @@ add_custom_target(
   ALL
   COMMAND
   # generate jni header files
-  ${Java_JAVAC_EXECUTABLE} -h jni ${java_files} -d jni
+  ${Java_JAVAC_EXECUTABLE} -h jni ${JAVA_FILES} -d jni
   COMMENT "Generate jni headers"
   VERBATIM
 )
```
mpreiner: We should be able to do a `add_custom_command` here too in order to avoid unnecessary rebuilds of the JNI headers. They should only be rebuilt if `JAVA_FILES` changed.

mudathirmahgoub: Will do

mudathirmahgoub: Done

cmake/FindJUnit.cmake
```diff
@@ -1,36 +1,59 @@
...
-  VERSION_VAR JUnit_VERSION)
+if(NOT JUnit_FOUND_SYSTEM)
+  include(ExternalProject)
 
-mark_as_advanced(JUnit_JAR JUnit_JAR_DEPS)
+  # Download junit generator jar
+  ExternalProject_Add(
+    JUnit
+    PREFIX ${DEPS_PREFIX}
+    URL https://repo1.maven.org/maven2/org/junit/platform/junit-platform-console-standalone/1.7.1/junit-platform-console-standalone-1.7.1.jar
```
4tXJ7f, line 36: It would be good to define a variable with the version. `JUNIT_VERSION` should contain the version that we found/installed (see [GMP](https://github.com/CVC4/CVC4/blob/master/cmake/FindGMP.cmake#L45)).

src/api/java/CMakeLists.txt
```diff
@@ -2,28 +2,112 @@
...
 ## in the top-level source directory and their institutional affiliations.
 ## All rights reserved.  See the file COPYING in the top-level source
 ## directory for licensing information.
 ##
 
-# create a directory for the cvc package
-file(MAKE_DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}/cvc")
+# create directories
+file(MAKE_DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}/src/main/java/cvc5/")
+file(MAKE_DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}/jni")
```
4tXJ7f, line 14: This directory is for the generated JNI files, right? I think that it would be more consistent to have them in `src/api/java/jni`.

src/api/java/CMakeLists.txt
```diff
@@ -2,28 +2,112 @@
...
 ## Copyright (c) 2009-2021 by the authors listed in the file AUTHORS
 ## in the top-level source directory and their institutional affiliations.
 ## All rights reserved.  See the file COPYING in the top-level source
 ## directory for licensing information.
 ##
 
-# create a directory for the cvc package
-file(MAKE_DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}/cvc")
+# create directories
+file(MAKE_DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}/src/main/java/cvc5/")
```
4tXJ7f, line 13: If we stored the Java files in `src/api/java/cvc5` would Java complain about the folder structure?

src/api/java/CMakeLists.txt
```diff
@@ -2,28 +2,112 @@
...
+include(UseJava)
+
+# specify java source files
+set(JAVA_FILES
+  ${CMAKE_CURRENT_LIST_DIR}/cvc5/CVC5ApiException.java
+  ${CMAKE_CURRENT_LIST_DIR}/cvc5/CVC5ApiRecoverableException.java
+  ${CMAKE_CURRENT_LIST_DIR}/cvc5/IPointer.java
+  ${CMAKE_CURRENT_LIST_DIR}/cvc5/Solver.java
+  ${CMAKE_CURRENT_LIST_DIR}/cvc5/Utils.java
+  ${CMAKE_CURRENT_BINARY_DIR}/src/main/java/cvc5/Kind.java
```
4tXJ7f, line 51: Nit: could use `${JAVA_KIND_FILE}` here.

src/api/java/cvc5/Solver.java
```diff
@@ -0,0 +1,48 @@
...
+   * @throws CVC5ApiException
+   */
+  public void setLogic(String logic) throws CVC5ApiException
+  {
+    setLogic(pointer, logic);
+  }
+
+  private native void setLogic(long solverPointer, String logic) throws CVC5ApiException;
+
+  // endregion
```
4tXJ7f, line 47: It looks like `// region ...` is missing. Maybe remove the `// endregion`?

4tXJ7f (changes requested):
Some more minor comments

src/api/java/cvc/CVCApiException.java
4tXJ7f: I actually meant [this](https://github.com/CVC4/CVC4/blob/master/src/api/cpp/cvc5.cpp#L1-L32) part of our files :)

src/api/java/CMakeLists.txt
mudathirmahgoub: I guess  `src/api/java/cvc5` can work, but my I understand generated files should be in the `build` directory. Also files in the build directory  are not tracked by git, so generated files would not appear with `git status`.

src/api/java/CMakeLists.txt
mudathirmahgoub: Right. Done.

src/api/java/cvc5/Solver.java
mudathirmahgoub: remnant of old code . Removed

src/api/java/CMakeLists.txt
nafur: Note that the prefix is still `CMAKE_CURRENT_BINARY_DIR`, which is within the build directory.

mudathirmahgoub: Oh, I misunderstood this with the  main source files. Named `main` to `api`.

src/api/java/CMakeLists.txt
mudathirmahgoub: Moved jni files to `src/api/java/jni`

cmake/FindJUnit.cmake
```diff
@@ -1,36 +1,59 @@
...
-include(UseJava)
-find_package(Hamcrest REQUIRED)
+include(deps-helper)
 
-find_jar(JUnit_JAR NAMES junit junit4)
+find_jar(JUnit_JAR junit-platform-console-standalone
+  PATHS
+    ${DEPS_BASE}/share/java
+    $ENV{PATH} $ENV{HOME} $ENV{CLASSPATH} $ENV{JUNIT_HOME}
+  VERSIONS 1.7.1
```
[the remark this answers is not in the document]
mudathirmahgoub: Added variable JUNIT_VERSION to the cmake file and removed the hard coded 1.7.1.

cmake/FindJUnit.cmake
mudathirmahgoub: Done

src/api/java/cvc/CVCApiException.java
mudathirmahgoub: Done. Once these java files are stable and handle all needed functionalities, they could also be auto generated.

4tXJ7f (approved):
LGTM, thanks for all the updates!

src/api/java/CMakeLists.txt
```diff
@@ -13,21 +13,106 @@
...
+message(STATUS "JAVA_AWT_INCLUDE_PATH: ${JAVA_AWT_INCLUDE_PATH}")
+
+add_library(cvc5jni SHARED jni/cvc5_Solver.cpp)
+add_dependencies(cvc5jni generate-jni-headers)
+
+target_include_directories(cvc5jni PUBLIC ${JNI_INCLUDE_DIRS})
+target_include_directories(cvc5jni PUBLIC ${PROJECT_SOURCE_DIR}/src)
+target_include_directories(cvc5jni PUBLIC ${CMAKE_BINARY_DIR}/src/)
+target_include_directories(cvc5jni PUBLIC ${JNI_DIR})
+target_link_libraries(cvc5jni ${JNI_LIBRARIES})
```
mpreiner, line 102:
```suggestion
target_link_libraries(cvc5jni PRIVATE ${JNI_LIBRARIES})
```

src/api/java/CMakeLists.txt
```diff
@@ -13,21 +13,106 @@
...
+target_link_libraries(cvc5jni cvc5)
```
mpreiner, line 103:
```suggestion
target_link_libraries(cvc5jni PRIVATE cvc5)
```

src/api/java/CMakeLists.txt
```diff
@@ -13,21 +13,106 @@
...
+add_jar(cvc5jar
+  SOURCES
+  ${JAVA_FILES}
+  VERSION ${CVC5_MAJOR}.${CVC5_MINOR}.${CVC5_RELEASE}
+  OUTPUT_NAME cvc5
 )
+
+add_dependencies(cvc5jar generate-java-kinds cvc5jni cvc5)
+
+set(CVC5_JAR_PATH "${CMAKE_CURRENT_BINARY_DIR}/${CVC5_JAR}" CACHE STRING "cvc5 jar" FORCE)
```
mpreiner, line 117: This should not be needed since the location of `cvc5jar` can be queried via `get_target_property(CVC5_JAR_PATH cvc5jar JAR_FILE)` when you need it for the tests.

(https://cmake.org/cmake/help/latest/module/UseJava.html)

src/api/java/CMakeLists.txt
```diff
@@ -13,21 +13,106 @@
...
+set(CVC5_JNI_PATH "${CMAKE_CURRENT_BINARY_DIR}" CACHE STRING "cvc5 jni path" FORCE)
```
mpreiner, line 118: You should be able to get this information by something like `get_filename_component(CVC5_JNI_PATH ${CVC5_JAR_PATH} DIRECTORY)` when you need it for the tests.

(https://cmake.org/cmake/help/latest/command/get_filename_component.html)

src/api/java/cvc5/CVC5ApiRecoverableException.java
```diff
@@ -0,0 +1,24 @@
...
+ ** Top contributors (to current version):
+ **   Mudathir Mohamed
+ ** This file is part of the CVC5 project.
+ ** Copyright (c) 2009-2021 by the authors listed in the file AUTHORS
+ ** in the top-level source directory and their institutional affiliations.
+ ** All rights reserved.  See the file COPYING in the top-level source
+ ** directory for licensing information.\endverbatim
+ **
+ ** \brief class CVC5ApiRecoverableException for cvc5 recoverable API exceptions
+ **/
```
mpreiner, line 13: Replace with new header format.

test/unit/api/java/CMakeLists.txt
```diff
@@ -0,0 +1,54 @@
+#####################
+## CMakeLists.txt
+## Top contributors (to current version):
+##   Mudathir Mohamed
+## This file is part of the CVC5 project.
+## Copyright (c) 2009-2021 by the authors listed in the file AUTHORS
+## in the top-level source directory and their institutional affiliations.
+## All rights reserved.  See the file COPYING in the top-level source
+## directory for licensing information.
+##
```
mpreiner, line 10: Replace with new header format.

test/unit/api/java/CMakeLists.txt
```diff
@@ -0,0 +1,54 @@
...
+    -d . # specify the output directory for the generated .class files
+  COMMENT "Build junit tests"
+  VERBATIM
+)
+
+# make sure junit jar file is present
+add_dependencies(build-junit-tests JUnit-EP-jar)
+# make sure cvc.jar file is built first
+add_dependencies(build-junit-tests cvc5jar)
+
```
mpreiner, line 37: Here is the place where you should retrieve the information for `CVC5_JNI_PATH` and `CVC5_JAR_PATH` as described above.

mpreiner (commented):
LGTM overall. I have some final comments. We should be good to go after they got adressed.

src/api/java/CMakeLists.txt
```diff
@@ -13,21 +13,106 @@
...
+  ${JNI_DIR}/cvc5_Solver.h
+)
+
+# generate jni headers
+add_custom_command(
+  OUTPUT
+    ${JNI_HEADERS}
+  BYPRODUCTS
+    ${JNI_HEADERS}
+  ALL
```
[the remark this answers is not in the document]
mudathirmahgoub: Done

src/api/java/CMakeLists.txt
mudathirmahgoub: Done

src/api/java/CMakeLists.txt
mudathirmahgoub: Done

src/api/java/CMakeLists.txt
mudathirmahgoub: Thanks. Changed to use `get_filename_component`.

src/api/java/cvc5/CVC5ApiRecoverableException.java
mudathirmahgoub: Done.

test/unit/api/java/CMakeLists.txt
mudathirmahgoub: Done

test/unit/api/java/CMakeLists.txt
mudathirmahgoub: For some reason `get_target_property(CVC5_JAR_PATH cvc5jar JAR_FILE)` does not work and this error happens
`get_target_property() called with non-existent target "cvc5jar".`

mpreiner: Are you able to call `get_target_property()` in `src/api/java/CMakeLists.txt`?

mudathirmahgoub: Yes. It works fine there.

mpreiner: Hmm that's weird since you obviously don't have a problem finding `cvc5jar` for `add_dependencies(build-junit-tests cvc5jar)`. Ok let's move the `get_target_property()` to the other file.

mudathirmahgoub: Don.e My guess, add_jar target is local and not global if this makes sense.

mpreiner (approved):
LGTM, thanks for the changes!

PR #6201, opened 2021-03-24
Refactor our integration of LFSC

nafur:
This PR further decouples LFSC from the remaining build system, but makes it more convenient to use at the same time.
The `contrib/get-lfsc-checker` script is now completely standalone and installs LFSC and the necessary signatures to `deps/`, as well as wrapper scripts to conveniently run `lfscc` from there and directly check `cvc4` proofs.
The goal would be to automatically use LFSC in our regressions as well.

### DIFF
--- a/contrib/get-lfsc-checker
+++ b/contrib/get-lfsc-checker
@@ -85,15 +85,14 @@ SIGS="\$SIG_DIR/core_defs.plf \\
     \$SIG_DIR/strings_programs.plf \\
     \$SIG_DIR/strings_rules.plf \\
     \$SIG_DIR/quantifiers_rules.plf"
-$BASE_DIR/bin/lfscc \$SIGS \$@ >& temp-lfsc.txt
+$BASE_DIR/bin/lfscc \$SIGS \$@ >& lfsc.out
 
 # recover macros for applications of arity 1,2,3, and simplify builtin syntax for constants
-#sed -i 's/(f_ite [^ \)]*)/f_ite/g' temp-lfsc.txt
-#sed -i 's/(\\ [^ ]* (\\ [^ ]* (\\ [^ ]* (apply (apply (apply f_\([^ ]*\) [^ ]*) [^ ]*) [^ ]*))))/\1/g; s/(\\ [^ ]* (\\ [^ ]* (apply (apply f_\([^ ]*\) [^ ]*) [^ ]*)))/\1/g; s/(\\ [^ ]* (apply f_\([^ ]*\) [^ ]*))/\1/g; s/(var \([^ ]*\) [^ \)]*)/var_\1/g; s/(int \([^ \)]*\))/\1/g; s/emptystr/""/g; s/int\.//g' temp-lfsc.txt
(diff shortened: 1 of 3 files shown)


nafur:
Marking as draft as it does not seem to work yet.
Note that I slightly changed how the script is called, as we can not assume (or automatically detect) the location of cvc4, hence it is now passed to the wrapper script:
```
$ deps/bin/cvc4-lfsc-check.sh ~/cvc4_proof_prs/build_debug/bin/cvc4 test/regress/regress0/chained-equality.smt2
=== Generate with CVC4 /home/gereon/cvc4_proof_prs/build_debug/bin/cvc4 test/regress/regress0/chained-equality.smt2
=== End generate with CVC4
=== Check with lfsc proof.plf
Error: /home/gereon/cvc4_prs/deps/share/lfsc/signatures/boolean_programs.plf at 46:30-46:31
The head of an application is missing a type in code.
1. the application: ((\ t (apply f_not t)) premise)
=== Finshed check with lfsc
```
Could @ajreynol check whether this is to be expected?

Also, I'd really like to get rid of this `sed` mess.

ajreynol:
@nafur That error is occurring since we are using a slightly older LFSC version than what is necessary. I think we should update the used LFSC version to the latest version as it incorporates some fixes from @alex-ozdemir.

Can you try updating the LFSC commit?

ajreynol:
Note the sed stuff is not essential but could be helpful for debugging, it recovers macros in LFSC error messages so that error messages are readable.

contrib/get-lfsc-checker
```diff
@@ -1,18 +1,102 @@
...
-mkdir build
-cd build
-cmake -DCMAKE_INSTALL_PREFIX="$INSTALL_DIR" ..
-make install -j$(nproc)
+echo "=== Finished check with lfsc"
+EOF
+chmod +x $BASE_DIR/bin/lfsc-check.sh
 
-echo
-echo ===================== Now configure CVC4 with =====================
```
ajreynol, line 17: Can you perhaps add a similar one line help message that is printed at the end of get-lfsc-checker about the next intended step for the user?  I don't think this is done in the updated script?

ajreynol (changes requested):

nafur:
I addressed your comments.
With the most recent LFSC version, my example seems to work fine.

contrib/get-lfsc-checker
```diff
@@ -1,18 +1,108 @@
...
+
+echo "=== Finished check with lfsc"
+EOF
+chmod +x $BASE_DIR/bin/lfsc-check.sh
+
+popd
+
+echo ""
+echo "========== How to use LFSC =========="
+echo "Check a generated proof:"
```
ajreynol, line 105: Can we also say how to generate an LFSC proof from cvc4 (without checking?)  Does this need another simple script?

ajreynol (changes requested):
One more comment. Otherwise, this looks good to me.

contrib/get-lfsc-checker
nafur: Yes, I just added that third script.

ajreynol (approved):
LGTM

PR #6202, opened 2021-03-24
Refactor ANTLR3 dependency

nafur:
This PR refactors our first, and arguably most fragile, dependency.
Right now all dependencies need to be manually installed (by calling the appropriate `contrib/get-X` script). For optional dependencies, we additionally need to enable them when calling the configure script (or via `ccmake`).
This PR is the first step in refactoring all dependencies to be automatically build (if required) as an external project.

Note that this not only eliminates the need to call contrib scripts, but also simplifies cross compilation: as all dependencies are now built within the build folders, every build folder has its own copy which may use different toolchains.


nafur:
There is one issue: in an existing checkout where ANTLR has already been build in the old deps folder, the new find script probably still finds the old ANTLR build.

cmake/FindANTLR3.cmake
```diff
@@ -0,0 +1,104 @@
...
+## FindANTLR.cmake
+## Top contributors (to current version):
+##   Gereon Kremer, Mathias Preiner, Aina Niemetz
+## This file is part of the CVC4 project.
+## Copyright (c) 2009-2021 by the authors listed in the file AUTHORS
+## in the top-level source directory and their institutional affiliations.
+## All rights reserved.  See the file COPYING in the top-level source
+## directory for licensing information.
+##
+# Find ANTLR
```
mpreiner, line 11:
```suggestion
# Find ANTLR3
```

cmake/deps-helper.cmake
```diff
@@ -0,0 +1,46 @@
+# where to build dependencies
+set(DEPS_PREFIX "${CMAKE_BINARY_DIR}/deps")
+# base path to installed dependencies
+set(DEPS_BASE "${CMAKE_BINARY_DIR}/deps")
```
mpreiner, line 4: Shouldn't this be something like `${CMAKE_BINARY_DIR}/deps/install` otherwise it is the same as `DEPS_PREFIX`?

cmake/FindDummy.cmake
```diff
@@ -0,0 +1,102 @@
...
+#   Dummy_FOUND_SYSTEM - we use the system version
+# Exported targets:
+#   Dummy - an imported library target
+
+# provides some utility definitions
+include(deps-helper)
+
+# initial setup
+find_package(Java REQUIRED)
+include(UseJava)
```
mpreiner, line 26: I'd not include this in the helper (I fear that this will be copied all over the place even though Java will not be needed ;-)).

cmake/FindDummy.cmake
```diff
@@ -0,0 +1,102 @@
+#####################
+## FindDummy.cmake
```
mpreiner, line 2: Maybe rename to .cmake.template just to make sure that this is not a cmake file that can be included.

cmake/FindDummy.cmake
```diff
@@ -0,0 +1,102 @@
...
+if(NOT Dummy_FOUND_SYSTEM)
+    # we install the library with an external project
+    include(ExternalProject)
+
+    # check for cases where the external project is known to fail
+    # it might make sense to specify a reason
+    fail_if_cross_compiling("Windows" "" "Dummy" "some reason")
+    fail_if_cross_compiling("" "arm" "Dummy" "some reason")
+
+    # declare some relase, version, tag, commit
```
mpreiner, line 57:
```suggestion
    # declare some release, version, tag, commit
```

cmake/FindANTLR3.cmake
```diff
@@ -0,0 +1,104 @@
...
+
+    # Download, build and install antlr3 runtime
+    ExternalProject_Add(
+        ANTLR3-EP-runtime
+        DEPENDS ANTLR3-EP-config.guess
+        PREFIX ${DEPS_PREFIX}
+        URL https://www.antlr3.org/download/C/libantlr3c-3.4.tar.gz
+        URL_HASH SHA1=faa9ab43ab4d3774f015471c3f011cc247df6a18
+        CONFIGURE_COMMAND ${CMAKE_COMMAND} -E copy <SOURCE_DIR>/../config.guess <SOURCE_DIR>/config.guess
+        COMMAND sed "s/avr32 \\\\/avr32 | aarch64 \\\\/" <SOURCE_DIR>/config.sub > <SOURCE_DIR>/config.sub.new
```
mpreiner, line 70: Can you check if this is still needed? Since config.guess is now up-to-date it should actually recognize aarch64 architectures.

cmake/FindANTLR3.cmake
```diff
@@ -0,0 +1,104 @@
...
+        COMMAND ${CMAKE_COMMAND} -E copy <SOURCE_DIR>/config.sub.new <SOURCE_DIR>/config.sub
+        COMMAND ${CMAKE_COMMAND} -E copy_directory <SOURCE_DIR>/include include/
+        COMMAND <SOURCE_DIR>/configure --with-pic --disable-antlrdebug --prefix=<INSTALL_DIR> --srcdir=<SOURCE_DIR> --disable-shared --enable-static --enable-64bit --host=${TOOLCHAIN_PREFIX}
```
mpreiner, line 73: 64-bit should not always be enabled. It should depend on the target architecture.

cmake/FindANTLR3.cmake
```diff
@@ -0,0 +1,104 @@
...
+    )
+
+    set(ANTLR3_BINARY java -cp "${DEPS_BASE}/share/java/antlr-3.4-complete.jar" org.antlr.Tool)
```
mpreiner, line 76:
```suggestion
    set(ANTLR3_BINARY ${Java_JAVA_EXECUTABLE} -cp "${DEPS_BASE}/share/java/antlr-3.4-complete.jar" org.antlr.Tool)
```

mpreiner (changes requested):

cmake/deps-helper.cmake
nafur: You can do either, I guess. Note that all checkouts and build folders etc are in `${CMAKE_BINARY_DIR}/deps/src/`, thus I felt it was okay to avoid this additional folder... but I don't really care.

cmake/FindDummy.cmake
nafur: Good point!

cmake/FindDummy.cmake
nafur: :+1:

cmake/FindANTLR3.cmake
nafur: I checked and the arm (cross-compilation) build failed for me without this sed command.

cmake/FindANTLR3.cmake
nafur: Right. I simply removed the option for now. If we plan to support 32bit builds in any way, we should do more changes, though. For example, the toolchains should define `CMAKE_SYSTEM_PROCESSOR` (https://cmake.org/cmake/help/latest/variable/CMAKE_SYSTEM_PROCESSOR.html)... but we should do this in another PR.

nafur: Okay. We can't simply remove it. :-)

cmake/deps-helper.cmake
mpreiner: It probably just matters for caching in CI. Do we have to cache the `${CMAKE_BINARY_DIR}/deps/` or only the `${CMAKE_BINARY_DIR}/deps/{lib,include,..}` directories?

nafur: Right now we need to cache `deps/src` as well, i.e. `${CMAKE_BINARY_DIR}/deps/`.
It's even worse than "not being cached": removing `deps/src` from a succesful build breaks the subsequent build for weird reasons...

It may be possible to change this (maybe via `BUILD_BYPRODUCTS`?) but it seems rather unlikely to me. Note that the `src` directly also contains the `stamp` files that store the timestamps necessary to determine whether something needs to be updated...
I'd suggest to simply cache the whole directory (just like we cached the whole `deps` directory beforehand)

mpreiner: I just hope that the whole directory won't be too large (it includes all the object files, sources etc.). We just tried to get down the size of the caches so that we can store more ccache caches.

nafur: As per offline discussion: the folder size is of course larger than only the install folder of the previous deps folder, but still okay.

.gitignore
```diff
@@ -12,5 +12,4 @@ generated/
 \#*\#
 *.swp
 /debug/
-/antlr-3.4
-deps/
+/deps/
```
mpreiner, line 15: Is this even required if we have the dependencies in build/?

cmake/deps-helper.cmake
```diff
@@ -0,0 +1,46 @@
...
+# search path hint for find_X family
+list(APPEND CMAKE_PREFIX_PATH "${DEPS_BASE}/deps/")
+list(REMOVE_DUPLICATES CMAKE_PREFIX_PATH)
+# cmake wants include dirs to exist when a target is created
+file(MAKE_DIRECTORY "${DEPS_BASE}/include/")
+
+macro(check_system_version name)
```
mpreiner, line 11: Please document. Where does the `${name}_FIND_VERSION` come from?

cmake/deps-helper.cmake
```diff
@@ -0,0 +1,46 @@
+# where to build dependencies
+set(DEPS_PREFIX "${CMAKE_BINARY_DIR}/deps")
+# base path to installed dependencies
+set(DEPS_BASE "${CMAKE_BINARY_DIR}/deps")
+# search path hint for find_X family
+list(APPEND CMAKE_PREFIX_PATH "${DEPS_BASE}/deps/")
+list(REMOVE_DUPLICATES CMAKE_PREFIX_PATH)
+# cmake wants include dirs to exist when a target is created
+file(MAKE_DIRECTORY "${DEPS_BASE}/include/")
```
mpreiner, line 9: Not sure if I understand why this is needed. What issue did you run into?

mpreiner (changes requested):
Please make sure to format to 80 chars.

.gitignore
nafur: We decided to build "external tools" (like LFSC) in `/deps/`. I'd keep it.

cmake/deps-helper.cmake
nafur: Good catch. This is a leftover from when the external projects were still living outside of the Find scripts.

mpreiner: I was actually referring  to the MAKE_DIRECTORY line :)

nafur: Ah. Well, the issue is what the comment says: when you create a target (at configure time) with an include dir that does not exist yet (because it's only created at build time when the first dependency is actually installed) cmake complains.

mpreiner: You mean because of e.g. this: https://github.com/CVC4/CVC4/pull/6202/files#diff-ab64debc440f89def9a126261fa321af0b17b28764660020f3b9f0934559cd0dR96 ?

nafur: Exactly.

mpreiner: Ok good, then how about changing the comment to:

```
# CMake wants directories specified via INTERFACE_INCLUDE_DIRECTORIES to exist when target property is set.
```

mpreiner (approved):

PR #6204, opened 2021-03-24
Ensure nonlinear extensions properly reconsiders its model

### LINKED ISSUE #6192: non-linear arithmetic Check failure at src/theory/model_manager.cpp:126
Commit: 61b9dadc
```
$cvc4 -q --no-strings-lazy-pp bug.smt2                                          
Fatal failure within void CVC4::theory::ModelManager::postProcessModel(bool) at 
src/theory/model_manager.cpp:126
Check failure                                                                   
                                                                                
 d_modelBuiltSuccess                                                            
                                                                                
[1]    4355 abort      cvc4 -q --no-strings-lazy-pp bug.smt2                    
$cat bug.smt2                                                                   
(declare-fun skoX () Real)                                                      
(assert (< 1.0 (/ 0 0 (+ (- 3) (* skoX skoX)))))                                
(check-sat) 
```

nafur:
In certain cases, the nonlinear extension would not re-check properly but only repeat an (already failed) model repair. The result lemma would then already be in cache and thus not be sent to the solver.
This PR ensures that `modelBasedRefinement` is always run and entirely removes the `d_builtModel` flag that was responsible for this behavior. Additionally, it asserts that the lemma that (tries to) force the model to be reconciled during theory combination is actually sent.
Fixes #6192.


ajreynol:
Should we do a run on e.g. SMT-LIB QF_NRA to test if there are any changed statuses?  Otherwise, this looks good to me.

ajreynol (approved):
LGTM

PR #6205, opened 2021-03-25
Do not use Configuration class in API black tests.

mpreiner:


mpreiner:
FYI: @mudathirmahgoub

mudathirmahgoub (approved):
LGTM. I am not sure why sort_black tests were passing before without SymFPU.

aniemetz (approved):

aniemetz:
> LGTM. I am not sure why sort_black tests were passing before without SymFPU.

I'm wondering about that, too.

PR #6206, opened 2021-03-25
FP: Refactor FloatingPointLiteral in preparation for MPFR.

aniemetz:
This pushes all symfpu specific parts from FloatingPoint into
FloatingPointLiteral. FloatingPoint is now generic. An additional
FloatingPointLiteral implementation using MPFR will be made configurable
similiarly to how we handle Integers with either GMP or CLN backend.


martin-cs (approved):
Seems reasonable.

PR #6212, opened 2021-03-25
Modular bv2int part 1

yoni206:
This PR introduces the header file for a more modular bv-to-int module, that will be called from the preprocessing pass bv-to-int, and possibly from the bit-vector solver after preprocessing.

The header file is basically a copy of the `bv_to_int.h` header file from preprocessing, with some adjustments to increase modularity.

In addition to the header file we also introduce an empty unit test that `#include`s the header, in order to identify compilation errors. The unit test will be populated in subsequent PRs, that will also include implementations.


src/theory/bv/int_blaster.h
```diff
@@ -0,0 +1,365 @@
...
+#include "context/cdhashmap.h"
+#include "context/cdhashset.h"
+#include "context/cdo.h"
+#include "context/context.h"
+#include "options/smt_options.h"
+#include "theory/arith/nl/iand_utils.h"
+
+namespace CVC4 {
+
+using CDNodeMap = context::CDHashMap<Node, Node, NodeHashFunction>;
```
ajreynol, line 86: I'd prefer if this were nested within the class.

src/theory/bv/int_blaster.h
```diff
@@ -0,0 +1,365 @@
...
+ **   Yoni Zohar
+ ** This file is part of the CVC4 project.
+ ** Copyright (c) 2009-2020 by the authors listed in the file AUTHORS
+ ** in the top-level source directory and their institutional affiliations.
+ ** All rights reserved.  See the file COPYING in the top-level source
+ ** directory for licensing information.\endverbatim
+ **
+ ** \brief A translation utility from bit-vectors to integers.
+ **
+ ** Converts bit-vector formulas to integer formulas.
```
ajreynol, line 14: Can you move this to be a comment above the class instead?

src/theory/bv/int_blaster.h
```diff
@@ -0,0 +1,365 @@
...
+  IntBlaster(context::Context* context,
+             options::SolveBVAsIntMode mode,
+             uint64_t granluarity = 1,
+             bool introduceFreshIntVars = true);
+
+  /**
+   * The result is an integer term and is computed
+   * according to the translation specified above.
+   * @param n is a bit-vector term or formula to be translated.
+   * @param lemmas additional lemmas that are needd for the translation
```
ajreynol, line 109:
```suggestion
   * @param lemmas additional lemmas that are needed for the translation
```

src/theory/bv/int_blaster.h
```diff
@@ -0,0 +1,365 @@
...
+   * @param  isLeftShift true iff the desired operation is a left shift.
+   * @return a node representing the shift.
+   *
+   */
+  Node createShiftNode(std::vector<Node> children,
+                       uint64_t bvsize,
+                       bool isLeftShift);
+
+  /**
+   * Adds the constraint 0 <= node < 2^size to lemmas
```
ajreynol, line 147: This can be one line, similar for below.

ajreynol (changes requested):

ajreynol (approved):

PR #6213, opened 2021-03-26
Optimizer for BitVectors

Ouyancheng:
Adds support for BitVector optimization, which is done via offline binary search. Units tests included. 
Also mildly refactors the optimizer architecture.


ajreynol:
As a general comment, I think we need to have a discussion where the code for optimization should go.  I don't think it should be put into this single file, nor should we add specific optimization implementation files to `src/smt/`.  I would suggest we think about a new subdirectory `src/optimization`.

Ouyancheng:
I agree, the file is becoming large...

ajreynol:
After a quick discussion, I think the consensus is that we should create a new directory `src/omt/`, which will have all the optimization code apart from the core solver (OptimizationSolver) which should stay in `src/smt/` where it is now.  Then I'd advocate that the individual OMTOptimizer classes to be in separate files.

Otherwise, the style looks pretty good to me.  Once the above files are moved, I can do another review.

Ouyancheng:
Moved individual optimizers to separate files.

src/omt/bitvector_optimizer.h
```diff
@@ -0,0 +1,52 @@
...
+
+/**
+ * Optimizer for BitVector type
+ */
+class OMTOptimizerBitVector : public OMTOptimizer
+{
+ public:
+  OMTOptimizerBitVector(bool isSigned);
+  virtual ~OMTOptimizerBitVector() = default;
+  virtual std::pair<OptResult, Node> minimize(SmtEngine* parentSMTSolver,
```
ajreynol, line 30:
```suggestion
  std::pair<OptResult, Node> minimize(SmtEngine* parentSMTSolver,
```

src/omt/bitvector_optimizer.h
```diff
@@ -0,0 +1,52 @@
...
+                                              Node target) override;
+  virtual std::pair<OptResult, Node> maximize(SmtEngine* parentSMTSolver,
```
ajreynol, line 32:
```suggestion
  std::pair<OptResult, Node> maximize(SmtEngine* parentSMTSolver,
```

src/omt/integer_optimizer.h
```diff
@@ -0,0 +1,50 @@
...
+
+/**
+ * Optimizer for Integer type
+ */
+class OMTOptimizerInteger : public OMTOptimizer
+{
+ public:
+  OMTOptimizerInteger() = default;
+  virtual ~OMTOptimizerInteger() = default;
+  virtual std::pair<OptResult, Node> minimize(SmtEngine* parentSMTSolver,
```
ajreynol, line 30:
```suggestion
  std::pair<OptResult, Node> minimize(SmtEngine* parentSMTSolver,
```

src/omt/integer_optimizer.h
```diff
@@ -0,0 +1,50 @@
...
+                                              Node target) override;
+  virtual std::pair<OptResult, Node> maximize(SmtEngine* parentSMTSolver,
```
ajreynol, line 32:
```suggestion
  std::pair<OptResult, Node> maximize(SmtEngine* parentSMTSolver,
```

src/smt/optimization_solver.h
```diff
@@ -55,23 +55,43 @@ enum OptResult
...
+
   // The last value is here as a preparation for future work
   // in which pproximate optimizations will be supported.
 
   // if the solver halted early and value is only approximate
   OPT_SAT_APPROX
 };
 
+/**
+ * The optimization objective,
```
ajreynol, line 69: Can you ensure this uses the 80 char column limit better?

src/omt/bitvector_optimizer.cpp
```diff
@@ -0,0 +1,248 @@
...
+    return std::make_pair(OptResult::OPT_UNSAT, value);
+  }
+
+  // value equals to upperBound
+  value = optChecker->getValue(target);
+
+  // this gets the bitvector!
+  BitVector bvValue = value.getConst<BitVector>();
+  unsigned int bvSize = bvValue.getSize();
+  // BitVector bv1 = BitVector::mkOne(bvSize);
```
ajreynol, line 178:
```suggestion
```

src/omt/integer_optimizer.cpp
```diff
@@ -0,0 +1,107 @@
...
+#include "options/smt_options.h"
+#include "smt/smt_engine.h"
+#include "theory/quantifiers/quantifiers_attributes.h"
+#include "theory/smt_engine_subsolver.h"
+
+using namespace CVC4::theory;
+
+namespace CVC4::smt {
+
+std::unique_ptr<SmtEngine> OMTOptimizerInteger::createOptChecker(
```
ajreynol, line 26: It is a bit hard to keep track of all these variants.

It might be better to not have this method and just use the `createOptCheckerWithTimeout` utility directly.

src/omt/bitvector_optimizer.cpp
```diff
@@ -0,0 +1,248 @@
...
+            + aMod2PlusbMod2.arithRightShift(bv1));
+  }
+  else
+  {
+    return (a.logicalRightShift(bv1) + b.logicalRightShift(bv1)
+            + aMod2PlusbMod2.logicalRightShift(bv1));
+  }
+}
+
+std::unique_ptr<SmtEngine> OMTOptimizerBitVector::createOptChecker(
```
ajreynol, line 54: Similar to the integer case, it might be better to use the `createOptCheckerWithTimeout` method directly and delete this method.

ajreynol (changes requested):

src/omt/bitvector_optimizer.h
Ouyancheng: AFAIK Removing virtual will cause the compiler directly call OMTOptimizer::minimize when invoking [unique_ptr<OMTOptimizer>] ->minimize.

src/omt/bitvector_optimizer.h
Ouyancheng: same here.

src/omt/bitvector_optimizer.h
Ouyancheng: I might be wrong, perhaps override has the same effect as virtual...

Ouyancheng: OK I am wrong... specifying virtual in the base class is sufficient...

src/smt/optimization_solver.h
Ouyancheng: Sure~

src/omt/integer_optimizer.cpp
Ouyancheng: Agreed

src/omt/bitvector_optimizer.cpp
Ouyancheng: Resolved

src/omt/bitvector_optimizer.cpp
Ouyancheng: Resolved

src/omt/bitvector_optimizer.h
Ouyancheng: Resolved

src/smt/optimization_solver.h
Ouyancheng: Should be resolved

src/omt/integer_optimizer.cpp
```diff
@@ -0,0 +1,90 @@
...
+    value = optChecker->getValue(target);
+    Assert(!value.isNull());
+    increment = nm->mkNode(incrementalOperator, target, value);
+    optChecker->assertFormula(increment);
+    intermediateSatResult = optChecker->checkSat();
+  }
+  return std::make_pair(OptResult::OPT_OPTIMAL, value);
+}
+
+std::pair<OptResult, Node> OMTOptimizerInteger::minimize(
```
ajreynol, line 77: Should this just be the default implementation of `OMTOptimizer::minimize`?

src/omt/integer_optimizer.cpp
```diff
@@ -0,0 +1,90 @@
...
+    SmtEngine* parentSMTSolver, Node target)
+{
+  return this->optimize(
+      parentSMTSolver, target, ObjectiveType::OBJECTIVE_MINIMIZE);
+}
+std::pair<OptResult, Node> OMTOptimizerInteger::maximize(
```
ajreynol, line 83: Ditto, for maximize.

ajreynol (changes requested):
A few more minor comments.

src/omt/integer_optimizer.cpp
Ouyancheng: Hmm... I think the "less than (LT)" operator and the "LE" operator only works for Integer and Real. So I didn't make it universal... and thus I didn't move OMTOptimizerInteger::optimize to OMTOptimizer either~  
Also using this linear search algorithm on reals might not run into completion -- it keeps finding bad values... I had a testcase for this, but it's removed in this PR since it doesn't run to completion. 

Notice that the testcase for the old framework... not the new refactored one...
`
void testOmtOpenReal()
  {
    Term ub1 = d_slv->mkReal(100);
    Term lb1 = d_slv->mkReal(0);
    Term lb2 = d_slv->mkReal(110);
    Term ub2 = d_slv->mkReal(120);

    Term cost1 = d_slv->mkConst(d_slv->getRealSort(), "cost3"); 
    Term cost2 = d_slv->mkConst(d_slv->getRealSort(), "cost4");
    // 0 <= cost1 <= 100
    d_slv->assertFormula(d_slv->mkTerm(CVC4::api::Kind::LEQ, lb1, cost1));
    d_slv->assertFormula(d_slv->mkTerm(CVC4::api::Kind::LEQ, cost1, ub1));
    // 110 <= cost2 <= 120
    d_slv->assertFormula(d_slv->mkTerm(CVC4::api::Kind::LEQ, lb2, cost2));
    d_slv->assertFormula(d_slv->mkTerm(CVC4::api::Kind::LEQ, cost2, ub2));

    Term cost3 = d_slv->mkTerm(CVC4::api::PLUS, cost1, cost2);

    const ObjectiveType obj_type = OBJECTIVE_MAXIMIZE;
    d_optslv->activateObj(*cost3.d_node, obj_type);

    OptResult r = d_optslv->checkOpt();

    // TS_ASSERT_EQUALS(ub1, ub2);

    TS_ASSERT_EQUALS(d_optslv->objectiveGetValue(), d_nm->mkConst(Rational(220)));
    std::cout << "Result is :" << r << std::endl;
    std::cout << "Optimized value is: " << d_optslv->objectiveGetValue() << std::endl;
  }
`

ajreynol: So the body of this method is a one line method:
```
return this->optimize(
      parentSMTSolver, target, ObjectiveType::OBJECTIVE_MINIMIZE);
```
This itself doesn't contain any reference to anything integer specific.  Indeed, it should be the case that target may be integer-specific (?), but I don't see why this method in its current form needs to be here.

Ouyancheng: Yes, this linear search could be used on Reals, just doesn't run to completion sometimes (for bounded cases)... So I would try not to use it on Reals. 
This is the linear search currently in master and it's only for integers. 
I would figure out an approach for LIRA later on...

ajreynol: I see, ok, I guess there isn't a need to create OMTOptimizer::optimize for now.

ajreynol (approved):

src/omt/integer_optimizer.cpp
Ouyancheng: Thanks for the review Dr. Andrew!

src/omt/omt_optimizer.cpp
```diff
@@ -0,0 +1,69 @@
...
+ ** All rights reserved.  See the file COPYING in the top-level source
+ ** directory for licensing information.\endverbatim
+ **
+ ** \brief The base class for optimizers of individual CVC4 type
+ **/
+
+#include "omt_optimizer.h"
+
+#include "bitvector_optimizer.h"
+#include "integer_optimizer.h"
```
mpreiner, line 18:
```suggestion
#include "omt/integer_optimizer.h"
```

src/omt/omt_optimizer.cpp
```diff
@@ -0,0 +1,69 @@
...
+ ** in the top-level source directory and their institutional affiliations.
+ ** All rights reserved.  See the file COPYING in the top-level source
+ ** directory for licensing information.\endverbatim
+ **
+ ** \brief The base class for optimizers of individual CVC4 type
+ **/
+
+#include "omt_optimizer.h"
+
+#include "bitvector_optimizer.h"
```
mpreiner, line 17:
```suggestion
#include "omt/bitvector_optimizer.h"
```

test/unit/theory/theory_int_opt_white.cpp
```diff
@@ -56,20 +57,23 @@ TEST_F(TestTheoryWhiteIntOpt, max)
   d_smtEngine->assertFormula(upb);
   d_smtEngine->assertFormula(lowb);
 
-  const ObjectiveType max_type = OBJECTIVE_MAXIMIZE;
+  const ObjectiveType max_type = ObjectiveType::OBJECTIVE_MAXIMIZE;
 
   // We activate our objective so the subsolver knows to optimize it
   d_optslv->activateObj(max_cost, max_type);
 
+  std::cerr << "a" << std::endl;
```
mpreiner, line 65: Looks like it doesn't belong here.

mpreiner (changes requested):
Most of the changes are related to moving the code to the `omt` subdirectory. Make sure that you include the headers with the correct path prefixed with omt and the namespace should also be renamed to `omt`.

src/omt/bitvector_optimizer.cpp
```diff
@@ -0,0 +1,242 @@
...
+ ** This file is part of the CVC4 project.
+ ** Copyright (c) 2009-2021 by the authors listed in the file AUTHORS
+ ** in the top-level source directory and their institutional affiliations.
+ ** All rights reserved.  See the file COPYING in the top-level source
+ ** directory for licensing information.\endverbatim
+ **
+ ** \brief Optimizer for BitVector type
+ **/
+
+#include "bitvector_optimizer.h"
```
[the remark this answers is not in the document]
Ouyancheng: they are at the same omt folder, so perhaps no need to explicitly write omt/... ?

src/omt/bitvector_optimizer.cpp
```diff
@@ -0,0 +1,242 @@
...
+                                                const BitVector& b,
+                                                bool isSigned)
+{
+  // computes (a + b) / 2 without overflow
+  // rounding towards -infinity: -1.5 --> -2,  1.5 --> 1
+  // average = (a / 2) + (b / 2) + (((a % 2) + (b % 2)) / 2)
+  uint32_t aMod2 = (uint32_t)(a.isBitSet(0));
+  uint32_t bMod2 = (uint32_t)(b.isBitSet(0));
+  BitVector aMod2PlusbMod2(a.getSize(), uint32_t((aMod2 + bMod2) / 2));
+  BitVector bv1(a.getSize(), (uint32_t)1);
```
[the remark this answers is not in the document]
Ouyancheng: ok...

src/omt/bitvector_optimizer.cpp
```diff
@@ -0,0 +1,242 @@
...
+
+BitVector OMTOptimizerBitVector::computeAverage(const BitVector& a,
+                                                const BitVector& b,
+                                                bool isSigned)
+{
+  // computes (a + b) / 2 without overflow
+  // rounding towards -infinity: -1.5 --> -2,  1.5 --> 1
+  // average = (a / 2) + (b / 2) + (((a % 2) + (b % 2)) / 2)
+  uint32_t aMod2 = (uint32_t)(a.isBitSet(0));
+  uint32_t bMod2 = (uint32_t)(b.isBitSet(0));
```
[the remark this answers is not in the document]
Ouyancheng: sure

src/omt/bitvector_optimizer.cpp
```diff
@@ -0,0 +1,242 @@
...
+  BitVector aMod2PlusbMod2(a.getSize(), uint32_t((aMod2 + bMod2) / 2));
+  BitVector bv1(a.getSize(), (uint32_t)1);
+  if (isSigned)
+  {
+    return (a.arithRightShift(bv1) + b.arithRightShift(bv1)
+            + aMod2PlusbMod2.arithRightShift(bv1));
+  }
+  else
+  {
```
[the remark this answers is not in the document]
Ouyancheng: Coding style issue? I would definitely prefer keeping it in order to make it clear.

test/unit/theory/theory_int_opt_white.cpp
Ouyancheng: ok I will remove.

src/omt/bitvector_optimizer.cpp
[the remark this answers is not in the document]
Ouyancheng: writing omt/... confuses me somehow...

src/omt/bitvector_optimizer.cpp
```diff
@@ -0,0 +1,242 @@
...
+#include "bitvector_optimizer.h"
+
+#include "options/smt_options.h"
+#include "smt/smt_engine.h"
+#include "theory/quantifiers/quantifiers_attributes.h"
+#include "theory/smt_engine_subsolver.h"
+
+using namespace CVC4::theory;
+
+namespace CVC4::smt {
```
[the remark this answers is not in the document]
Ouyancheng: ok I could change this...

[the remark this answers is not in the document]
Ouyancheng: Err... sorry I would not change this... because the files are heavily referencing stuffs in CVC4::smt, changing this will introduce a lot of CVC4::smt::... in the header files. In other words, it's not completely independent from CVC4::smt. 
And introducing a new namespace like this doesn't bring us any real benefit...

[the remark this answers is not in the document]
Ouyancheng: OK, it seems that only stuffs like OptResult and ObjectiveType from CVC4::smt are referred... I could change this...

src/omt/integer_optimizer.cpp
```diff
@@ -0,0 +1,90 @@
...
+  {
+    // if objective is MAX, then assert optimization_target >
+    // current_model_value
+    incrementalOperator = kind::GT;
+  }
+  // Workhorse of linear search:
+  // This loop will keep incrmenting/decrementing the objective until unsat
+  // When unsat is hit,
+  // the optimized value is the model value just before the unsat call
+  while (intermediateSatResult.isSat())
```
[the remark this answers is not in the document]
Ouyancheng: I will add isSat() == Result::SAT here.

src/omt/bitvector_optimizer.cpp
```diff
@@ -0,0 +1,242 @@
...
+BitVector OMTOptimizerBitVector::computeAverage(const BitVector& a,
+                                                const BitVector& b,
+                                                bool isSigned)
+{
+  // computes (a + b) / 2 without overflow
+  // rounding towards -infinity: -1.5 --> -2,  1.5 --> 1
+  // average = (a / 2) + (b / 2) + (((a % 2) + (b % 2)) / 2)
+  uint32_t aMod2 = (uint32_t)(a.isBitSet(0));
+  uint32_t bMod2 = (uint32_t)(b.isBitSet(0));
+  BitVector aMod2PlusbMod2(a.getSize(), uint32_t((aMod2 + bMod2) / 2));
```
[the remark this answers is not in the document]
Ouyancheng: Resolved

src/omt/bitvector_optimizer.cpp
```diff
@@ -0,0 +1,242 @@
...
+      OMTOptimizer::createOptCheckerWithTimeout(parentSMTSolver, false);
+  NodeManager* nm = optChecker->getNodeManager();
+  Result intermediateSatResult = optChecker->checkSat();
+  // Model-value of objective (used in optimization loop)
+  Node value;
+  if (intermediateSatResult.isUnknown())
+  {
+    return std::make_pair(OptResult::OPT_UNKNOWN, value);
+  }
+  if (!intermediateSatResult.isSat())
```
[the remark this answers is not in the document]
Ouyancheng: Resolved

src/omt/integer_optimizer.cpp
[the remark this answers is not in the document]
Ouyancheng: the loop condition is now (intermediateSatResult.isSat() == Result::SAT).

Ouyancheng:
Hi Mathias @mpreiner, 

Thanks a lot for your review! 
If the PR looks good to you, could you please take off the <do not merge> tag? 
I have synchronized the PR with master and I think the namespace is all good now~ 

Thanks again! 

Regards, 
Yancheng Ou

mpreiner (approved):
LGTM, thanks for all the changes!

Ouyancheng:
> LGTM, thanks for all the changes!

Thanks Mathias~!

PR #6215, opened 2021-03-26
Refactor dependencies for external SAT solvers

nafur:
This PR refactors how we obtain, build and use the external SAT solvers used by CVC4: CaDiCaL, CryptoMiniSat and Kissat.
All three contrib scripts are removed and instead an external project is integrated into the cmake find scripts.


cmake/FindCaDiCaL.cmake
```diff
@@ -13,15 +13,80 @@
...
+    check_system_version("CaDiCaL")
+endif()
+
+if(NOT CaDiCaL_FOUND_SYSTEM)
+    include(CheckSymbolExists)
+    include(ExternalProject)
+
+    fail_if_include_missing("sys/resource.h" "CaDiCaL")
+
+    # TODO(mpreiner): Use a more recent version?
```
mpreiner, line 45: Yes, but we'll do that in a separate commit (we'll do that when preparing for SMT-COMP).

cmake/FindCaDiCaL.cmake
```diff
@@ -13,15 +13,80 @@
...
+
+    ExternalProject_Add(
+        CaDiCaL-EP
+        PREFIX ${DEPS_PREFIX}
+        BUILD_IN_SOURCE ON
+        URL https://github.com/arminbiere/cadical/archive/refs/tags/rel-${CaDiCaL_VERSION}.tar.gz
+        URL_HASH SHA1=9de1176737b74440921ba86395fe5edbb3b131eb
+        CONFIGURE_COMMAND mkdir -p <SOURCE_DIR>/build
+        # avoid configure script, prepare the makefile manually
+        COMMAND sed -e "s,@CXX@,${CMAKE_CXX_COMPILER}," -e "s,@CXXFLAGS@,${CXXFLAGS}," -e "s,@MAKEFLAGS@,-j1," <SOURCE_DIR>/makefile.in > <SOURCE_DIR>/build/makefile
```
mpreiner, line 66: Is `-jN` passed to external projects when we build CVC4 with `make -jN`?

mpreiner, line 66: Indent `COMMAND` since it belongs to `CONFIGURE_COMMAND`?

cmake/FindCaDiCaL.cmake
```diff
@@ -13,15 +13,80 @@
...
+        BUILD_COMMAND make -C <SOURCE_DIR>/build libcadical.a
+        INSTALL_COMMAND ${CMAKE_COMMAND} -E copy <SOURCE_DIR>/build/libcadical.a <INSTALL_DIR>/lib/libcadical.a
+        COMMAND ${CMAKE_COMMAND} -E copy <SOURCE_DIR>/src/cadical.hpp <INSTALL_DIR>/include/cadical.hpp
```
mpreiner, line 69: Ditto.

cmake/FindCryptoMiniSat.cmake
```diff
@@ -13,16 +13,78 @@
 # CryptoMiniSat_INCLUDE_DIR - the CryptoMiniSat include directory
 # CryptoMiniSat_LIBRARIES - Libraries needed to use CryptoMiniSat
 
+include(deps-helper)
 
 find_path(CryptoMiniSat_INCLUDE_DIR NAMES cryptominisat5/cryptominisat.h)
```
mpreiner, line 18: CryptoMiniSat exports CMake targets when installed. We should use `find_package(cryptominisat5)`with the specific version we want in this case.

cmake/FindCryptoMiniSat.cmake
```diff
@@ -13,16 +13,78 @@
...
+    )
+    ExternalProject_Add_Step(
+        CryptoMiniSat-EP
+        cleanup
+        DEPENDEES install
+        COMMAND ${CMAKE_COMMAND} -E remove <BINARY_DIR>/cryptominisat5_simple
+        COMMAND ${CMAKE_COMMAND} -E remove <INSTALL_DIR>/bin/cryptominisat5_simple
+    )
+
+    set(CryptoMiniSat_INCLUDE_DIR "${DEPS_BASE}/include/")
```
mpreiner, line 70: Again, we should use the installed CMake target for CryptoMiniSat.

mpreiner (changes requested):
Make sure to format to 80 chars.

cmake/FindCaDiCaL.cmake
nafur: A `COMMAND` generally belongs to the `X_COMMAND` that comes right before it. (see https://cmake.org/cmake/help/latest/module/ExternalProject.html#external-project-definition at "Miscellaneous Options:")

mpreiner: I know, that's why I suggested to indent it to make it clearer, but if you prefer it this way I'm fine with it.

nafur: If used properly, the external projects share the jobserver of the whole build. (Aka: yes).
In this particular case, I explicitly set `-j1` in the makefile, and also called `make` manually. I don't know why... it is fixed now.

cmake/FindCryptoMiniSat.cmake
nafur: You are right. I'm working on it (see https://github.com/msoos/cryptominisat/pull/645)

cmake/FindCryptoMiniSat.cmake
nafur: I'm not sure about this. The naive way would be to call `find_package` at this point in time. This does not work, though, as this searches for the export file that is only generated after CryptoMiniSat has been installed (i.e. at build time).

mpreiner: Ah yes. Hmm, well then we have to do it this way.

mpreiner (commented):
Can you break lines to 80 chars and make consistent indentation (2 vs. 4 spaces)? I think until now we mostly used 2 spaces for indentation in CMake files.

nafur:
> Can you break lines to 80 chars and make consistent indentation (2 vs. 4 spaces)? I think until now we mostly used 2 spaces for indentation in CMake files.

I've used `cmake-format --tab-size 2 --dangle-parens YES --first-comment-is-literal` now.

mpreiner:
> > Can you break lines to 80 chars and make consistent indentation (2 vs. 4 spaces)? I think until now we mostly used 2 spaces for indentation in CMake files.
> 
> I've used `cmake-format --tab-size 2 --dangle-parens YES --first-comment-is-literal` now.

Nice, didn't know about cmake-format!

src/CMakeLists.txt
```diff
@@ -1128,20 +1128,17 @@ if(USE_ABC)
   target_include_directories(cvc4 PRIVATE ${ABC_INCLUDE_DIR})
 endif()
 if(USE_CADICAL)
-  target_link_libraries(cvc4 ${CaDiCaL_LIBRARIES})
-  target_include_directories(cvc4 PRIVATE ${CaDiCaL_INCLUDE_DIR})
+  target_link_libraries(cvc4 CaDiCaL)
```
mpreiner, line 1131: Link PRIVATE?

src/CMakeLists.txt
```diff
@@ -1128,20 +1128,17 @@ if(USE_ABC)
...
 endif()
 if(USE_CLN)
   target_link_libraries(cvc4 ${CLN_LIBRARIES})
   target_include_directories(cvc4 PUBLIC $<BUILD_INTERFACE:${CLN_INCLUDE_DIR}>)
 endif()
 if(USE_CRYPTOMINISAT)
-  target_link_libraries(cvc4 ${CryptoMiniSat_LIBRARIES})
-  target_include_directories(cvc4 PRIVATE ${CryptoMiniSat_INCLUDE_DIR})
+  target_link_libraries(cvc4 CryptoMiniSat)
```
mpreiner, line 1138: Ditto.

src/CMakeLists.txt
```diff
@@ -1128,20 +1128,17 @@ if(USE_ABC)
...
 endif()
 if(USE_KISSAT)
-  target_link_libraries(cvc4 ${Kissat_LIBRARIES})
-  target_include_directories(cvc4 PRIVATE ${Kissat_INCLUDE_DIR})
+  target_link_libraries(cvc4 Kissat)
```
mpreiner, line 1141: Ditto.

src/CMakeLists.txt
nafur: This needs to be tone for all calls at once... see #6218.

cmake/FindCaDiCaL.cmake
```diff
@@ -1,27 +1,99 @@
...
-## in the top-level source directory and their institutional affiliations.
-## All rights reserved.  See the file COPYING in the top-level source
-## directory for licensing information.
-##
+# FindCaDiCaL.cmake
+# Top contributors (to current version):
+#   Mathias Preiner
+# This file is part of the CVC4 project.
+# Copyright (c) 2009-2021 by the authors listed in the file AUTHORS
+# in the top-level source directory and their institutional affiliations.
```
mpreiner, line 7: Oh, this change will break the update copyright script as it expects `##`. Can you revert the change?

cmake/FindCryptoMiniSat.cmake
```diff
@@ -1,28 +1,102 @@
...
-##
+# FindCryptoMiniSat.cmake
+# Top contributors (to current version):
+#   Mathias Preiner
+# This file is part of the CVC4 project.
+# Copyright (c) 2009-2021 by the authors listed in the file AUTHORS
+# in the top-level source directory and their institutional affiliations.
+# All rights reserved.  See the file COPYING in the top-level source
+# directory for licensing information.
+#
```
mpreiner, line 10: Ditto

cmake/FindKissat.cmake
```diff
@@ -1,27 +1,82 @@
...
-##
+# FindKissat.cmake
+# Top contributors (to current version):
+#   Aina Niemetz
+# This file is part of the CVC4 project.
+# Copyright (c) 2009-2021 by the authors listed in the file AUTHORS
+# in the top-level source directory and their institutional affiliations.
+# All rights reserved.  See the file COPYING in the top-level source
+# directory for licensing information.
+#
```
mpreiner, line 10: Ditto

cmake/FindCaDiCaL.cmake
nafur: done,

cmake/FindCaDiCaL.cmake
```diff
@@ -1,4 +1,4 @@
-#####################
+######################
```
mpreiner, line 1:
```suggestion
#####################
```

cmake/FindCryptoMiniSat.cmake
```diff
@@ -1,4 +1,4 @@
-#####################
+######################
```
mpreiner, line 1:
```suggestion
#####################
```

cmake/FindKissat.cmake
```diff
@@ -1,4 +1,4 @@
-#####################
+######################
```
mpreiner, line 1:
```suggestion
#####################
```

mpreiner (approved):

PR #6217, opened 2021-03-26
Eliminate use of quantifiers engine in enumerative instantiation

ajreynol:
This also makes minor updates to how term tuple enumerators are constructed.


src/theory/quantifiers/inst_strategy_enumerative.cpp
```diff
@@ -182,14 +181,15 @@ bool InstStrategyEnum::process(Node quantifier, bool fullEffort, bool isRd)
...
   ttec.d_fullEffort = fullEffort;
   ttec.d_increaseSum = options::fullSaturateSum();
-  ttec.d_isRd = isRd;
+  // make the enumerator, which is either relevant domain or term database
+  // based on the flag isRd.
   std::unique_ptr<TermTupleEnumeratorInterface> enumerator(
-      mkTermTupleEnumerator(quantifier, &ttec));
+      isRd ? mkTermTupleEnumeratorRd(quantifier, &ttec, d_rd)
+           : mkTermTupleEnumerator(
+                 quantifier, &ttec, d_qstate, d_treg.getTermDatabase()));
```
HanielB, line 192: This is the only place in which you are calling these methods. Maybe you could just inline them?

HanielB (approved):
LGTM, one minor suggestion

src/theory/quantifiers/inst_strategy_enumerative.cpp
ajreynol: Not currently, their implementation is using things that are local to term_tuple_enumerator.cpp by design.

PR #6218, opened 2021-03-26
Refactor SymFPU dependency

nafur:
This PR refactors the contrib script to download SymFPU to a cmake external project.


martin-cs (approved):
How does CVC4 feel about git submodules because that might be an even easier alternative.

nafur:
> How does CVC4 feel about git submodules because that might be an even easier alternative.

The idea has come up. Several devs (including myself) have reservations about git submodules. The main problems I personally see are:
- dependencies that do not live in a git are not suitable to be used as a git submodule (e.g. the ANTLR jar, GLPK). Yes, you can maintain your own git for that, but that really misses the point of easy updating.
- dependencies are managed in multiple places: some stuff lives in git, some in cmake.
- git submodules are somewhat confusing to use: they may be very natural if you have fully understood git and are always aware of the fact that you have submodules. I personally found them pretty confusing, although I tend to be explaining how git works more often than having it explained to me.

martin-cs:
OK; as long as it has been considered.  I find they work well in some cases and I think they would work well for SymFPU but I will go with the group decision.

cmake/FindSymFPU.cmake
```diff
@@ -1,20 +1,59 @@
...
-##
+# FindSymFPU.cmake
+# Top contributors (to current version):
+#   Mathias Preiner
+# This file is part of the CVC4 project.
+# Copyright (c) 2009-2021 by the authors listed in the file AUTHORS
+# in the top-level source directory and their institutional affiliations.
+# All rights reserved.  See the file COPYING in the top-level source
+# directory for licensing information.
+#
```
mpreiner, line 10:
```suggestion
## FindSymFPU.cmake
## Top contributors (to current version):
##   Mathias Preiner
## This file is part of the CVC4 project.
## Copyright (c) 2009-2021 by the authors listed in the file AUTHORS
## in the top-level source directory and their institutional affiliations.
## All rights reserved.  See the file COPYING in the top-level source
## directory for licensing information.
##
```

mpreiner (changes requested):

src/parser/CMakeLists.txt
```diff
@@ -103,6 +103,15 @@ set_target_properties(cvc4parser PROPERTIES SOVERSION ${CVC4_SOVERSION})
...
 target_link_libraries(cvc4parser PUBLIC cvc4)
 target_link_libraries(cvc4parser PRIVATE ANTLR3)
+
+if(USE_CLN)
+  target_link_libraries(cvc4parser PRIVATE ${CLN_LIBRARIES})
+  target_include_directories(cvc4parser PRIVATE $<BUILD_INTERFACE:${CLN_INCLUDE_DIR}>)
+endif()
+
+target_link_libraries(cvc4parser PRIVATE ${GMP_LIBRARIES})
+target_include_directories(cvc4parser PRIVATE $<BUILD_INTERFACE:${GMP_INCLUDE_DIR}>)
```
mpreiner, line 113: Shouldn't be needed anymore since Integer/Rational was eliminated in the TPTP parser (https://github.com/CVC4/CVC4/pull/6239).

src/main/CMakeLists.txt
```diff
@@ -58,6 +58,10 @@ set_target_properties(cvc4-bin
     OUTPUT_NAME cvc4
     RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin)
 target_link_libraries(cvc4-bin cvc4 cvc4parser)
+if(USE_CLN)
+  target_link_libraries(cvc4-bin ${CLN_LIBRARIES})
+  target_include_directories(cvc4-bin PRIVATE $<BUILD_INTERFACE:${CLN_INCLUDE_DIR}>)
+endif()
```
mpreiner, line 64: Why is this needed?

src/parser/CMakeLists.txt
nafur: CI seems to disagree: https://github.com/CVC4/CVC4/pull/6218/commits/f0c324ef95e899b73ebf2674577e68820f5b728c / https://github.com/CVC4/CVC4/runs/2231668107

src/main/CMakeLists.txt
nafur: We have this unfortunate include chain: `smt_engine.h` -> `sexpr.h` -> `integer.h`
This pulls gmp and cln code into various places, including `src/main`, api tests and unit tests

mpreiner (approved):

Issue #6224, opened 2021-03-27
CVC4 can't find Java


andrewvaughanj:
The legacy script `./contrib/get-antlr-3.4` used to work fine on my machine (has `java`, doesn't have `javac`), but now it seems I can't build (after, I guess, #6202):

```
avj@platypus ~/clones/CVC4/master$ ./configure.sh |& tail -n 13
CMake Error at /usr/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:218 (message):
  Could NOT find Java (missing: Java_JAR_EXECUTABLE Java_JAVAC_EXECUTABLE
  Java_JAVAH_EXECUTABLE Java_JAVADOC_EXECUTABLE) (found version "1.8.0_282")
Call Stack (most recent call first):
  /usr/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:582 (_FPHSA_FAILURE_MESSAGE)
  /usr/share/cmake/Modules/FindJava.cmake:331 (find_package_handle_standard_args)
  cmake/FindANTLR3.cmake:98 (find_package)
  src/parser/CMakeLists.txt:12 (find_package)


-- Configuring incomplete, errors occurred!
See also "/home/avj/clones/CVC4/master/build/CMakeFiles/CMakeOutput.log".
See also "/home/avj/clones/CVC4/master/build/CMakeFiles/CMakeError.log".
```

My guess is that this is because:

```
find_package(Java REQUIRED)
```

in https://github.com/CVC4/CVC4/blob/master/cmake/FindANTLR3.cmake#L98 should read:

```
find_package(Java COMPONENTS Runtime REQUIRED)
```

andrewvaughanj:
Closed 🎉

PR #6225, opened 2021-03-27
When building ANTLR via CMake, do not require javac #6224

andrewvaughanj:
As title; attempts to correct #6224.


nafur (approved):
Thanks!
I never thought about this distinction, and apparently we never tested on a system that had a JRE but not a JDK...

Issue #6223, opened 2021-03-27
`configure.sh` has no `--no-ninja` option


andrewvaughanj:
CVC4 typically has the ability to disable features with:

```
The following flags enable optional features (disable with --no-<option name>).
```

However, this doesn't work for `--ninja` and if you want to use Makefiles:

```
avj@platypus ~/clones/CVC4/master$ ./configure.sh --no-ninja
*** configure.sh: invalid option '--no-ninja' (try -h)
```

This gives rise to #6222

andrewvaughanj:
This really isn't a problem and isn't easy to avoid unless we add a `--make` option.

Closing out of pointlessness.

PR #6226, opened 2021-03-27
Add an option to explicitly not use Ninja #6223

andrewvaughanj:
This adds an option to `./configure.sh` to *not* use Ninja.

Note, most other `./configure.sh` options pass `-D<OPTION>=$var` to CMake -- this won't work for `-G Ninja`, so I've had to check `!= OFF`.

This works if, e.g., you have `export CMAKE_GENERATOR=Ninja` set but don't want to use Ninja.

### DIFF
--- a/configure.sh
+++ b/configure.sh
@@ -237,6 +237,7 @@ do
     --arm64) arm64=ON;;
 
     --ninja) ninja=ON;;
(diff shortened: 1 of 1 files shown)


andrewvaughanj:
Actually, this doesn't work (and is pointless).

PR #6231, opened 2021-03-29
Move decision manager into theory inference manager

ajreynol:
This makes it so that the decision manager is accessible from TheoryInferenceManager.

This is work towards breaking circular dependencies in quantifiers, and also helps simplify several other dependencies in e.g. UF and datatypes.


src/theory/theory.cpp
```diff
@@ -127,9 +126,11 @@ void Theory::setQuantifiersEngine(QuantifiersEngine* qe)
 
 void Theory::setDecisionManager(DecisionManager* dm)
 {
-  Assert(d_decManager == nullptr);
   Assert(dm != nullptr);
-  d_decManager = dm;
+  if (d_inferManager!=nullptr)
```
nafur, line 130: Format...

nafur (approved):
Just one minor comment.

PR #6233, opened 2021-03-29
Fix total time statistic

nafur:
Since one of the last changes to the total time collection in the driver, this collection fails when the solver terminates unexpectedly (either by a signal or an exception). This PR fixes this issue and does some cleanup along the way.


src/main/main.h
```diff
@@ -36,7 +38,19 @@ extern const char* progPath;
...
+
+/** Manages a custom timer for the total runtime in RAII-style. */
+class TotalTimer
+{
+    public:
+    TotalTimer(): d_start(std::chrono::steady_clock::now()) {}
+    ~TotalTimer();
+    private:
+    std::chrono::steady_clock::time_point d_start;
+};
```
mpreiner, line 51: clang-format

aniemetz (commented):
LGTM but it seems you forgot to clang-format.

src/main/main.h
nafur: done

src/main/signal_handlers.cpp
```diff
@@ -21,6 +21,7 @@
 #include <string.h>
 
 #include <cerrno>
+#include <chrono>
```
mpreiner, line 24: This include shouldn't be needed, right?

nafur: You are right. I removed it.

mpreiner (approved):

PR #6236, opened 2021-03-29
Fix configuration printing.

aniemetz:


nafur (approved):
LGTM.

cmake/Helpers.cmake
```diff
@@ -131,9 +131,9 @@ endmacro()
 # Helper to print the configuration of a 2-valued or 3-valued option 'var'
 # with prefix 'str'.
 macro(print_config str var)
-  if("${var}" STREQUAL "ON")
+  if(${var} STREQUAL "ON")
     set(OPT_VAL_STR "on")
-  elseif("${var}" STREQUAL "OFF")
```
nafur, line 136: I think we need to keep it as `"${var}"`, in case `${var}` is empty.

aniemetz: Indeed!

PR #6237, opened 2021-03-29
Modular bv2int part 2

yoni206:
This PR introduces the implementations for the modular bv-to-int module, that will be called from the preprocessing pass bv-to-int, and possibly from the bit-vector solver after preprocessing.

This is basically a copy of `bv_to_int.cpp` with some changes to increase modularity.

unit tests for these implementations are planned for a future PR.
Also, the PR does not include any code for the eager bit-wise mode, which is also left for a future PR (by makai).

### DIFF
--- a/src/theory/bv/int_blaster.cpp
+++ b/src/theory/bv/int_blaster.cpp
@@ -947,7 +947,7 @@ Node IntBlaster::translateFunctionSymbol(Node bvUF,
    * Similarly for the domains.
    */
   TypeNode intRange = bvRange.isBitVector() ? d_nm->integerType() : bvRange;
-  for (TypeNode d : bvDomain)
+  for (const TypeNode& d : bvDomain)
   {
     intDomain.push_back(d.isBitVector() ? d_nm->integerType() : d);
   }
--- a/src/theory/bv/int_blaster.cpp
+++ b/src/theory/bv/int_blaster.cpp
@@ -507,40 +507,10 @@ Node IntBlaster::translateWithChildren(
       }
       else
       {
-        Assert(d_mode == options::SolveBVAsIntMode::BITWISE);
-        // Enforce semantics over individual bits with iextract and ites
-        Assert(translated_children.size() == 2);
-        uint64_t granularity = options::BVAndIntegerGranularity();
-
-        Node x = translated_children[0];
-        Node y = translated_children[1];
-        Node iAndOp = d_nm->mkConst(IntAnd(bvsize));
-        Node iAnd = d_nm->mkNode(kind::IAND, iAndOp, x, y);
-        // get a skolem so the IAND solver knows not to do work
-        returnNode = d_nm->getSkolemManager()->mkPurifySkolem(
-            iAnd,
-            "__intblast__iand",
-            "skolem for an IAND node in bitwise mode " + iAnd.toString());
-        addRangeConstraint(returnNode, bvsize, lemmas);
-
-        // eagerly add bitwise lemmas according to the provided granularity
-        uint64_t high_bit;
-        for (uint64_t j = 0; j < bvsize; j += granularity)
-        {
-          high_bit = j + granularity - 1;
-          // don't let high_bit pass bvsize
-          if (high_bit >= bvsize)
-          {
-            high_bit = bvsize - 1;
-          }
-          Node extractedReturnNode =
-              d_iandUtils.iextract(high_bit, j, returnNode);
-          addBitwiseConstraint(
-              extractedReturnNode.eqNode(
-                  d_iandUtils.createBitwiseIAndNode(x, y, high_bit, j)),
-              lemmas);
-        }
+        // no other modes are currently implemented
+        Unreachable();
       }
+
       break;
     }
     case kind::BITVECTOR_SHL:
@@ -765,11 +735,6 @@ Node IntBlaster::translateWithChildren(
     case kind::BOUND_VAR_LIST:
     {
       returnNode = d_nm->mkNode(oldKind, translated_children);
-      if (d_mode == options::SolveBVAsIntMode::BITWISE)
-      {
-        throw OptionException(
-            "--solve-bv-as-int=bitwise does not support quantifiers");
-      }
       break;
     }
     case kind::FORALL:
--- a/src/theory/bv/int_blaster.cpp
+++ b/src/theory/bv/int_blaster.cpp
@@ -42,6 +42,26 @@ Rational intpow2(uint64_t b) { return Rational(Integer(2).pow(b), Integer(1)); }
 
 }  // namespace
 
+
+IntBlaster::IntBlaster(context::Context* c,
+                       options::SolveBVAsIntMode mode,
+                       uint64_t granularity,
+                       bool introduceFreshIntVars)
+    : d_binarizeCache(c),
+      d_eliminationCache(c),
+      d_rebuildCache(c),
+      d_intblastCache(c),
+      d_rangeAssertions(c),
+      d_mode(mode),
+      d_granularity(granularity),
+      d_context(c),
+      d_introduceFreshIntVars(introduceFreshIntVars)
+{
+  d_nm = NodeManager::currentNM();
+  d_zero = d_nm->mkConst<Rational>(0);
+  d_one = d_nm->mkConst<Rational>(1);
+};
+
 void IntBlaster::addRangeConstraint(Node node,
                                     uint64_t size,
                                     std::vector<Node>& lemmas)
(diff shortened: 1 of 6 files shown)


src/theory/bv/int_blaster.cpp
```diff
@@ -0,0 +1,1133 @@
...
+    bool inEliminationCache =
+        (d_eliminationCache.find(current) != d_eliminationCache.end());
+    bool inRebuildCache =
+        (d_rebuildCache.find(current) != d_rebuildCache.end());
+    if (!inEliminationCache)
+    {
+      // current is not the elimination of any previously-visited node
+      // current hasn't been eliminated yet.
+      // eliminate operators from it using rewrite rules
+      Node currentEliminated =
```
ajreynol, line 181: I'm not a huge fan of this, since this will need to be refactored if/when BV rewriter is refactored.

I assume you are doing this because the default rewriter of BV won't suffice?

src/theory/bv/int_blaster.cpp
```diff
@@ -0,0 +1,1133 @@
...
+  kind::Kind_t oldKind = original.getKind();
+  // signed comparisons were supposed to be eliminated by this point.
+  Assert(oldKind != kind::BITVECTOR_SLT);
+  Assert(oldKind != kind::BITVECTOR_SGT);
+  Assert(oldKind != kind::BITVECTOR_SLE);
+  Assert(oldKind != kind::BITVECTOR_SGE);
+  // The following variable will only be used in assertions.
+  CVC4_UNUSED uint64_t originalNumChildren = original.getNumChildren();
+  Node returnNode;
+  // Assert that BITVECTOR_UDIV/UREM were replaced by their
```
ajreynol, line 386: This comment is outdated. There is no _TOTAL versions anymore.

src/theory/bv/int_blaster.cpp
```diff
@@ -0,0 +1,1133 @@
...
+    {
+      returnNode = d_nm->mkNode(oldKind, translated_children);
+      break;
+    }
+    case kind::FORALL:
+    {
+      returnNode = translateQuantifiedFormula(original);
+      break;
+    }
+    case kind::EXISTS:
```
ajreynol, line 732: Can you turn this into an assert at a higher level?  I believe this will cause a compiler warning as is.

src/theory/bv/int_blaster.cpp
```diff
@@ -0,0 +1,1133 @@
...
+  std::vector<TypeNode> bvDomain = tn.getArgTypes();
+  std::vector<TypeNode> intDomain;
+  /**
+   * if the original range is a bit-vector sort,
+   * the new range should be an integer sort.
+   * Otherwise, we keep the original range.
+   * Similarly for the domains.
+   */
+  TypeNode intRange = bvRange.isBitVector() ? d_nm->integerType() : bvRange;
+  for (TypeNode d : bvDomain)
```
ajreynol, line 937:
```suggestion
  for (const TypeNode& d : bvDomain)
```

src/theory/bv/int_blaster.cpp
```diff
@@ -0,0 +1,1133 @@
...
+    Node adjustedChild = castToType(translatedChild, originalChild.getType());
+    builder << adjustedChild;
+  }
+  Node reconstruction = builder.constructNode();
+  // cast to tn in case the reconstruction is a bit-vector.
+  reconstruction = castToType(reconstruction, resultType);
+  return reconstruction;
+}
+
+IntBlaster::IntBlaster(context::Context* c,
```
ajreynol, line 1020: Move the constructor to the top of the file

ajreynol (changes requested):

src/theory/bv/int_blaster.cpp
yoni206: Yes, for example, we eliminate Or (using the relatively new rewrite rule OrEliminate), which the default rewriter does not do. In general, it is expected that int blasting would prefer other rewrites than bit-blasting.

ajreynol: Wouldn't it be simpler to just add the bv2int case for BVOR?

yoni206: And for the rest of these eliminations as well? 
Some of them were written specifically to be used in this pass, e.g.:
```
                                  RewriteRule<SdivEliminateFewerBitwiseOps>,
                                  RewriteRule<SremEliminateFewerBitwiseOps>,
                                  RewriteRule<SmodEliminateFewerBitwiseOps>,
```

Some of them are not turned on by the default rewriter.

We could have all these cases in the main translation, of coruse, but this feels like a rewrite doesn't it?

ajreynol: The rest of the eliminations should be made available if we had access to Theory::expandDefinitions here. 

I have been advocating for moving Theory::expandDefinitions to Rewriter::expandDefinitions.

What else would remain if you were able to call `Rewriter::rewrite( Theory/Rewriter::expandDefinitions( n ) )`?

ajreynol: Btw, I guess you can call `TheoryBVRewriter::eliminateBVSDiv(node);`.

yoni206: In what follows, whenever I say "not called by the bv rewriter" I actually mean "not called from this file: https://github.com/CVC4/CVC4/blob/master/src/theory/bv/theory_bv_rewriter.cpp". If these statements are not equivalent please tell me, i think they are.

Let's start with sdiv, srem, smod:
```
 RewriteRule<SdivEliminateFewerBitwiseOps>
 RewriteRule<SremEliminateFewerBitwiseOps>
 RewriteRule<SmodEliminateFewerBitwiseOps>
```
I don't think it would be good to call `TheoryBVRewriter::eliminateBVSDiv(node)`. It calls the ordinary elimination rules for sdiv, srem and smod. These introduce extraction terms in order to check the sign of the bit-vector. Extraction terms are translated to integers using divs and mods which is expensive. In contrast, I implemented custom sdiv, srem and smod elimination rules called `SdivEliminateFewerBitwiseOps,SremEliminateFewerBitwiseOps,SmodEliminateFewerBitwiseOps` that check the sign of the translated vector simply using the integer `<` operator. I think this approach is better for our translation.
I think this answer is also relevant for your question about `expandDefinition` because i seems like the usual eliminations are called from there as well.

The following elimination rules are not called from the bv rewriter because these operators are not eliminated by it. I added them for this pass, and hence calling the rewriter/expandDef won't help.
```
 RewriteRule<NegEliminate>
 RewriteRule<XorEliminate>
 RewriteRule<OrEliminate>
```

The following elimination rules are not called from the bv rewriter, I don't know why. Hence calling the rewriter won't help.
```
 RewriteRule<CompEliminate>
 RewriteRule<SltEliminate>
```

Now, the following rewrites are in fact being called from the BV rewriter. Does that mean that If I call `Rewriter::rewrite(n)` they will be invoked? If so, I can at least remove all these lines and replace them with assertions that these operators are absent from the processed formula.
```
RewriteRule<UdivZero>
 RewriteRule<XnorEliminate>
 RewriteRule<NandEliminate>
 RewriteRule<NorEliminate>
 RewriteRule<SubEliminate>
 RewriteRule<RepeatEliminate>
 RewriteRule<RotateRightEliminate>
 RewriteRule<RotateLeftEliminate>
 RewriteRule<SleEliminate>
 RewriteRule<SgtEliminate>
 RewriteRule<SgeEliminate>
```

ajreynol: Yes, those that are normally rewritten should be omitted from this list.  
I would conjecture that Comp and Slt *are* eliminated, but through other means.

This means that Neg / Xor / Or are the only cases. I would prefer if your pass explicitly handled these as part of its conversion. I believe this would be more informative, instead of having to have an explicit dependency on external rewrite rules.

ajreynol: Btw, the same applies for Sdiv / Smod / Srem, can you actually verify that your code actually has the opportunity to rewrite these?  The problem is that the standard rewriter can definitely be called before you run this pass, and the rewriter will eliminate all these if I'm not mistaken.

yoni206: > Yes, those that are normally rewritten should be omitted from this list.
> I would conjecture that things like Comp are eliminated, but through other rewrite rules.
OK sure, I will remove them and add assertions.

> This means that Neg / Xor / Or are the only cases (maybe SLT as well). I would prefer if your pass explicitly handled these as part of its conversion. I believe this would be more informative, instead of having to have an explicit dependency on external rewrite rules.
I am happy to do so, but would like to verify that @4tXJ7f is alright with that also, as he advocated for using rewrites some time ago in an offline discussion. 

 

> @ajreynol ajreynol 2 hours ago Member
> Btw, the same applies for Sdiv / Smod / Srem, can you actually verify that your code actually has the opportunity to rewrite these? The problem is that the standard rewriter can definitely be called before you run this pass, and the rewriter will eliminate all these if I'm not mistaken.
You are right -- they are not called! I believe I did verify that they are called when I introduced them but since then we changed (more than once) the place of this preprocessing pass w.r.t. expanding definitions and rewritings. This would be another advantage of the approach that you are suggesting (handling everything in the pass).

yoni206:
Following the discussion above, I will remove the new rewriting rules from the BV rewriter, and will implement the needed eliminations inside the main conversion function. I am therefore closing this PR and will open a new one once that is done.

PR #6242, opened 2021-03-30
Add missing inference ids

ajreynol:
Towards having complete stats on inference ids for each lemma, fact, and conflict.


nafur (approved):
LGTM.

PR #6244, opened 2021-03-30
python bindings in a custom build directory

yoni206:
In current master, the following fails whenever `<name>` contains a `/`:
```
./configure.sh --python-bindings --name=<name>
```
The reason is that `src/api/python/genkinds.py` adds a directory to the python path while relying on the fact that the build directory is located directly under the main repo directory, which is not the case if `<name>` contains a `/`.
This PR fixes this by having `cmake` determine the right directory to add to the python path.


mpreiner (approved):
LGTM

makaimann (approved):
Thanks for the fix! Looks good to me also.

PR #6246, opened 2021-03-31
FP: Factor out symfpu traits.

aniemetz:
This is in preparation for a MPFR floating-point literal implementation.
We will need to have both literal kinds return a symFPU unpacked float
via `getSymUF()` in order to be able to plug it into the fp_converter.
For this, it makes sense to have the traits implemented and to be
included separately, so that they can also be included in the MPFR
implementation.


src/util/CMakeLists.txt
```diff
@@ -9,6 +9,7 @@
 ## directory for licensing information.
 ##
 configure_file(floatingpoint_literal_symfpu.h.in floatingpoint_literal_symfpu.h)
+configure_file(floatingpoint_literal_symfpu_traits.h.in floatingpoint_literal_symfpu_traits.h)
```
mpreiner, line 12: 80 chars

mpreiner (approved):

PR #6247, opened 2021-03-31
Bags: Move implementation of type rules from header to .cpp file.

aniemetz:


aniemetz:
@mudathirmahgoub can you document the header similiarly to https://github.com/CVC4/CVC4/blob/master/src/theory/fp/theory_fp_type_rules.h when this is merged back to master?

mudathirmahgoub (approved):
LGTM

PR #6249, opened 2021-03-31
Rename namespace CVC4 to CVC5.

aniemetz:


aniemetz:
This change made me notice that we still use `Expr` in `examples`. I will not fix this in this PR, but in subsequent PRs.

nafur (approved):
It looks like the PR does what the title says :-)

martin-cs:
So glad that you got this merged 36 minutes before this could be an April fools joke.

PR #6252, opened 2021-03-31
Fix non-linear for unknown case

ajreynol:
As a result of this issue, currently we are incorrectly trying to repair a model when one is not guaranteed to exist, leading to a spurious assertion failure.

Fixes https://github.com/CVC4/CVC4/issues/6228.  The benchmarks on that issue now answer "unknown" without an assertion failure.


nafur (approved):
LGTM.

PR #6253, opened 2021-03-31
kinds: Remove non-existent properties.

aniemetz:


src/theory/arith/kinds
```diff
@@ -8,7 +8,7 @@ theory THEORY_ARITH ::CVC5::theory::arith::TheoryArith "theory/arith/theory_arit
 typechecker "theory/arith/theory_arith_type_rules.h"
 
 properties stable-infinite
-properties check propagate ppStaticLearn presolve notifyRestart
+properties propagate ppStaticLearn presolve notifyRestart
```
ajreynol, line 11:
```suggestion
properties ppStaticLearn presolve notifyRestart
```

ajreynol (changes requested):

src/theory/arith/kinds
aniemetz: `propagate()` is defined in `theory_arith.h`, though?

ajreynol: Oh I see, it looked as though you were deleting the propagate property, I forgot this is not 100% standardized yet.

Nevermind :)

ajreynol (approved):

ajreynol (changes requested):
Actually, this needs a bit more thought.  

I would actually recommend either:
(A) keeping the `check` property but disabling the warnings, or
(B) eliminate the `check` property (but fix it so that checks are called), and verify that it does not degrade performance.

The situation here is that some theories implement check(), others don't, however the code to check if the `check()` method exists is broken since this method is now implemented in the base class, not the derived ones.

ajreynol (approved):
LGTM

PR #6255, opened 2021-04-01
A proposal for python api unit tests

yoni206:
This PR introduces two unit tests for the python api, translated directly from the unit tests for the cpp api.
The goal is to get feedback in order to reach some kind of a pattern/style for python API tests.
Also, i'd be happy to hear if there is any specific cpp api unit test I should translate for this initial attempt (e.g., a test that is more representative or might raise difficulties). For now i just picked the first two solver tests.


test/unit/api/python/test_solver.py
```diff
@@ -0,0 +1,36 @@
+#####################
+## test_term.py
```
makaimann, line 2: minor: filename

test/unit/api/python/test_solver.py
```diff
@@ -0,0 +1,36 @@
...
+import pycvc4
+import sys
+
+from pycvc4 import kinds
+
+class TestSolver:
+    def setup_class(self):
+        self.solver = pycvc4.Solver()
+
+    def test_recoverableException(self):
```
makaimann, line 21: Personal preference here, but I dislike the mix of underscore and camel case. It would also be more pythonic to just use underscores.

makaimann (approved):
Left some minor comments for you to consider, but otherwise it looks good to me. Thanks for working on this! I think this is a fine pattern. It is nice to not repeat the solver instantiation as it is in the existing tests. 

The class is not used very heavily though, so I would also consider [fixtures](https://semaphoreci.com/community/tutorials/testing-python-applications-with-pytest) without a class. This also has the nice side effect of being able to use `solver` as opposed to `self.solver`. Of course, there is some additional flexibility of a class (e.g. inheritance), but I'm not sure we'll need it. What do you think? The documentation on using a class seems relatively sparse as well, so I wonder if it's not too commonly used.

test/unit/api/CMakeLists.txt
```diff
@@ -21,3 +21,8 @@ cvc4_add_unit_test_white(solver_white api)
 cvc4_add_unit_test_black(sort_black api)
 cvc4_add_unit_test_black(term_black api)
 cvc4_add_unit_test_white(term_white api)
+
+# add Python bindings tests if building with Python bindings
+if (BUILD_BINDINGS_PYTHON)
+  add_subdirectory(python)
+endif()
```
mpreiner, line 28: I'm not sure if we should make a python subdirectory in `test/unit/api/python` or do something like `test/python/unit/api`. We already have a `test/api/python` directory. I'd suggest we use `test/<lang>/{unit,api,...}` in order to avoid all the `python` subdirectories. This also has the advantage that we need to check for `pytest` once in `test/python/CMakeLists.txt`.

makaimann: I think the idea was to move the tests from `test/api` but regardless you still make a good point. Eventually there might be multiple directories of Python tests and it makes sense to gather them.

yoni206: yes, I was planning to move the content of `test/api/python` to `test/unit/api/python`. Will we have any other python tests except for unit tests of the python api?

test/unit/api/python/CMakeLists.txt
```diff
@@ -0,0 +1,40 @@
...
+  ${PYTHON_EXECUTABLE} -c "import pytest"
+  RESULT_VARIABLE
+    RET_PYTEST
+  ERROR_QUIET
+)
+
+if(RET_PYTEST)
+    message(FATAL_ERROR
+      "Could not find Python module pytest. Install via `pip install pytest'.")
+endif()
```
mpreiner, line 26: I'd also suggest to change this message to something similar to https://github.com/CVC4/CVC4/blob/master/src/options/CMakeLists.txt#L20 to use `${PYTHON_EXECUTABLE} -m pip install pytest`. Maybe we can add a helper to print this kind of Python module errors.

test/unit/api/CMakeLists.txt
mpreiner: If the only tests will be in `test/unit/api` I'm fine with having a `test/unit/api/{python,java,...}` directory structure.

yoni206: @makaimann @aniemetz do we expect other python tests?

yoni206:
@mpreiner and @makaimann thanks for the comments. I've addressed them, and also edited the test file: `Exception` is thrown also for syntax errors, so I changed the expectation to `Exception` to an expectation to `RuntimeError`, which is what is thrown when there is an exception from the api (thanks makai).

test/unit/api/CMakeLists.txt
makaimann: I’m not sure, but my inclination is to make it flexible enough to add (non-unit) tests in the future. Seems safer to allow the possibility than assume it won’t ever be added imo.

yoni206: OK that makes sense. I have reorganized. Hopefully the way I handled `cmake` is not too silly (one liner `CMakeLists.txt` files that just add the only subdir). Please let me know if there is a better way.

test/python/unit/api/CMakeLists.txt
```diff
@@ -0,0 +1,44 @@
...
+  #   and run it with 'ctest -R "unit/api/python/myapitest"'.
+  add_test (NAME python/unit/api/${name}
+    COMMAND ${PYTHON_EXECUTABLE} -m pytest ${CMAKE_CURRENT_SOURCE_DIR}/${filename}
+    # directory for importing the python bindings
+    WORKING_DIRECTORY ${CMAKE_BINARY_DIR}/src/api/python)
+
+endmacro()
+
+# add specific test files
+cvc4_add_python_api_test(pytest_solver test_solver.py)
```
mpreiner, line 44: This file should be ` test/python/CMakeLists.txt`. You don't need the other CMakeLists.txt files

mpreiner (changes requested):

test/python/unit/api/CMakeLists.txt
yoni206: Thanks! changed and pushed, but also had to give a fuller path to the test file in this line.

test/python/CMakeLists.txt
```diff
@@ -0,0 +1,44 @@
...
+      "Could not find pytest for Python "
+      "version ${PYTHON_VERSION_MAJOR}.${PYTHON_VERSION_MINOR}. "
+      "Make sure to install pytest for this Python version "
+      "via \n`${PYTHON_EXECUTABLE} -m pip install pytest'.\nNote: You need to "
+      "have pip installed for this Python version.")
+endif()
+
+macro(cvc4_add_python_api_test name filename)
+
+  #   we create test target 'unit/api/python/myapitest'
```
mpreiner, line 34:
```suggestion
  #   we create test target 'python/unit/api/myapitest'
```

test/python/CMakeLists.txt
```diff
@@ -0,0 +1,44 @@
...
+  #   and run it with 'ctest -R "unit/api/python/myapitest"'.
```
mpreiner, line 35:
```suggestion
  #   and run it with 'ctest -R "python/unit/api/myapitest"'.
```

mpreiner (commented):
Almost there :-)

yoni206:
:) sorry for missing these, thanks!

mpreiner (approved):
LGTM, thanks for the changes!

PR #6257, opened 2021-04-01
Fix type rule for to_real

### LINKED ISSUE #6208: semantics of to_real when the argument is a real
Hi, I wonder what should be the semantics of`(to_real 0.5)`.  Should it be rejected as `to_real` expects an integer, viewed as `(to_real (to_int 0.5))`, or treated like identity?   Thanks.

```
[600] % cvc4 --strict-parsing small.smt2 
sat
unsat
[601] % z3release small.smt2 
unsat
unsat
[602] % 
[602] % cat small.smt2 
(set-logic QF_NIRA)
(assert (= 0.5 (to_real 0.5)))
(check-sat)
(reset)
(set-logic QF_NIRA)
(assert (= 0.5 (to_real (to_int 0.5))))
(check-sat)
[603] % 
```

ajreynol:
This fixes the type rule for `to_real` to match SMT-LIB: its argument must be an integer.

This required fixing the TPTP parser which has a more relaxed semantics for to_real / to_rat.

This also fixes Solver::isReal, which should return false if we are the integer type.

Fixes #6208.


nafur (approved):
LGTM.

Issue #6260, opened 2021-04-01
Overflow Error in mkBitVector of Python API


NicolaasWeideman:
**Describe the bug**
The Python statement `slv.mkBitVector(32, 2**31)` yields the following error:
```
Traceback (most recent call last):
  File "mwe.py", line 13, in <module>
    bv = slv.mkBitVector(32, 2**31)
  File "cvc4.pxi", line 742, in pycvc4.Solver.mkBitVector
OverflowError: value too large to convert to int
```

I attached a minimal example:
[example.zip](https://github.com/CVC4/CVC4/files/6244805/example.zip)

NicolaasWeideman:
It appears the error is caused by casting the value of the bit vector to a signed integer (see [here](https://github.com/CVC4/CVC4/blob/afaf4413775ff7d6054a5893f1397ad908e0773c/src/api/python/cvc4.pxi#L742)).
Pull request #6261 should fix this.

PR #6261, opened 2021-04-01
Fix Issue #6260 type casting in mkBitVector

NicolaasWeideman:
Update the type casting in `mkBitVector` of the Python API to match the types expected by `Solver::mkBitVector` in [cvc4cpp.cpp](https://github.com/CVC4/CVC4/blob/afaf4413775ff7d6054a5893f1397ad908e0773c/src/api/cvc4cpp.cpp#L5085).
This should fix Issue #6260.


src/api/python/cvc4.pxi
```diff
@@ -736,18 +736,18 @@ cdef class Solver:
         cdef Term term = Term(self)
         if isinstance(size_or_str, int):
             if val is None:
-                term.cterm = self.csolver.mkBitVector(<int> size_or_str)
+                term.cterm = self.csolver.mkBitVector(<uint32_t> size_or_str)
             else:
-                term.cterm = self.csolver.mkBitVector(<int> size_or_str,
-                                                      <int> val)
+                term.cterm = self.csolver.mkBitVector(<uint32_t> size_or_str,
+                                                      <uint64_t> val)
```
mpreiner, line 742: Hmm, casting is not a proper fix here since `size_or_str` can be an arbitrary integer (e.g. `2**128`). We should actually convert the integer to a string and pass the value as string to `mkBitVector`. @makaimann what do you think?

mpreiner: Don't we also have the same issue with `val`?

makaimann: I agree, a better solution is to convert to a string for the value. The size can't be arbitrary so that can stay the same. Thanks for the PR @NicolaasWeideman! 

You'd need to declare this method: https://github.com/NicolaasWeideman/CVC4/blob/9c208b9cac092adc97a98b49d061715d6ad0e89d/src/api/cvc4cpp.h#L2894

here at this line: https://github.com/NicolaasWeideman/CVC4/blob/9c208b9cac092adc97a98b49d061715d6ad0e89d/src/api/python/cvc4.pxd#L189

Then you should just be able to call `str(val)` and pass it to the function. Let me know if I can help.

NicolaasWeideman:
@makaimann could you have a look at the updates, please?
If I understand correctly, `val` is used as the base [here](https://github.com/NicolaasWeideman/CVC4/blob/9c208b9cac092adc97a98b49d061715d6ad0e89d/src/api/python/cvc4.pxi#L750) and therefore casting to `uint32_t` should still be ok?

makaimann:
Yes that’s right, it is used as the base there.

src/api/python/cvc4.pxi
```diff
@@ -736,18 +736,19 @@ cdef class Solver:
...
         if isinstance(size_or_str, int):
             if val is None:
-                term.cterm = self.csolver.mkBitVector(<int> size_or_str)
+                term.cterm = self.csolver.mkBitVector(<uint32_t> size_or_str)
             else:
-                term.cterm = self.csolver.mkBitVector(<int> size_or_str,
-                                                      <int> val)
+                term.cterm = self.csolver.mkBitVector(<uint32_t> size_or_str,
+                                                      <const string &> str(val).encode(),
+                                                      <uint32_t> 10)
```
makaimann, line 743: Very minor: you shouldn’t have to cast the value 10.

makaimann (approved):
Looks great, thanks again for the PR! Just left a minor comment for you to consider.

NicolaasWeideman:
@makaimann thanks for the pointer! I removed the cast.

For the failing checks, the error reported is `GitHub Actions has encountered an internal error when running your job.` Do either of you know how to fix this?

Issue #6203, opened 2021-03-24
Invalid model for string formula


jiwonparc:
commit: 34798fb8

```
$ cvc4 -q --strings-exp --check-models small.smt2
Fatal failure within void CVC4::smt::CheckModels::checkModel(CVC4::smt::Model*, CVC4::context::CDList<CVC4::NodeTemplate<true> >*, bool) at 
CVC4/src/smt/check_models.cpp:131
Internal error detectedSmtEngine::checkModel(): ERRORS SATISFYING ASSERTIONS WITH MODEL:
assertion: (str.in_re (str.++ c (str.substr e (* (mod 0 0) (str.len b)) (str.len d))) (re.opt (str.to_re "ab")))
simplifies to: false
expected `true'.
Run with `--check-models -v' for additional diagnostics.
Aborted
$ cat small.smt2
(declare-fun a () String)
(declare-fun b () String)
(declare-fun c () String)
(declare-fun d () String)
(declare-fun e () String)
(assert (str.in_re (str.++ c (str.substr e (* (mod 0 0) (str.len b)) (str.len d))) (re.opt (str.to_re "ab"))))
(assert (= (str.len c) (+ (str.len e) (* (str.indexof a "a" 0) (str.len b))) 1 (str.len b)))
(check-sat)
```

jiwonparc:
commit: b3c6975a

```
$ cvc4 -q --strings-exp --check-models small.smt2
Fatal failure within void CVC4::smt::CheckModels::checkModel(CVC4::smt::Model*, CVC4::context::CDList<CVC4::NodeTemplate<true> >*, bool)
 at CVC4/src/smt/check_models.cpp:131
Internal error detectedSmtEngine::checkModel(): ERRORS SATISFYING ASSERTIONS WITH MODEL:
assertion:     (let ((_let_1 (str.to_re "A"))) (let ((_let_2 (re.* re.allchar ))) 
(xor (str.in_re a (re.++ _let_2 _let_1 re.allchar  _let_2)) (str.in_re a (re.++ (re.* (str.to_re a)) _let_1 re.allchar )))))
simplifies to: false
expected `true'.
Run with `--check-models -v' for additional diagnostics.
Aborted
$ cat small.smt2
(declare-fun a () String)
(assert
 (xor (str.in_re a (re.++ (re.* re.allchar) (str.to_re "A") re.allchar (re.* re.allchar)))
[...]

PR #6262, opened 2021-04-01
Simplify caching of regular expression unfolding

ajreynol:
This ensures that we always cache regular expressions in a user-dependent manner.  Previously, we were erroneously caching in a SAT-dependent way for very rare cases when non-constant regular expressions were used.  Since we never dependent on current assertions for the unfolding, there is no need to cache in the SAT context.

This fixes the second benchmark from #6203.

This PR also improves our traces for checking models in strings.


4tXJ7f (approved):
LGTM

PR #6265, opened 2021-04-01
Add cache for new dependencies folder.

nafur:
This PR adds caching of the new dependencies folder `build/deps/` for the CI jobs and renames the old deps folder to "auxiliary told".
Note that we need to cache the entirety of `build/deps/` (instead of just the install folder for the old one), otherwise `cmake` will try to rebuild them. Some of the external projects remove unnecessary files in their build to reduce their footprint in the cache.


.github/workflows/ci.yml
```diff
@@ -108,25 +108,34 @@ jobs:
...
 
+
+    # GitHub actions currently does not support modifying an already existing
+    # cache. Hence, we create a new cache for each commit with key
+    # cache-${{ runner.os }}-${{ matrix.cache-key }}-${{ github.sha }}. This
+    # will result in an initial cache miss. However, restore-keys will search
+    # for the most recent cache with prefix
+    # cache-${{ runner.os }}-${{ matrix.cache-key }}-, and if found uses it as
+    # a base for the new cache.
+
```
mpreiner, line 119: This comment only refers to the ccache caching step and not the others since we don't use `${{ github.sha }}` for dependency caching.

.github/workflows/ci.yml
```diff
@@ -108,25 +108,34 @@ jobs:
...
     - name: Restore Dependencies
       id: restore-deps
       uses: actions/cache@v1
+      with:
+        path: build/deps
+        key: ${{ runner.os }}-deps-${{ hashFiles('cmake/Find**') }}-${{ hashFiles('.github/workflows/ci.yml') }}
```
mpreiner, line 125: Do we also have to include deps-helper here?

.github/workflows/ci.yml
```diff
@@ -108,25 +108,34 @@ jobs:
...
+
+    - name: Restore Auxiliary Tools
+      id: restore-aux-tools
+      uses: actions/cache@v1
```
mpreiner, line 129:
```suggestion
      uses: actions/cache@v2
```

Please also update the other cache versions to v2.

mpreiner (changes requested):

.github/workflows/ci.yml
nafur: Added it.

.github/workflows/ci.yml
nafur: I changed the comment to not only describe what this does, but also discuss why we do it, and why we don't do it for the other caches.

.github/workflows/ci.yml
```diff
@@ -108,32 +108,50 @@ jobs:
...
+    #   change, which is when the contrib scripts or the CI config changes.
+    # - ccache changes with (almost) every commit and handles outdated contents
+    #   properly. We thus use `restore-keys` and store a new cache for every
+    #   commit.
+    # - deps (build/deps/) does handle outdated contents gracefully, but does
+    #   not change frequently. We thus use `restore-keys` to restore any recent
+    #   cache, but only store a new cache if the cmake or CI config changes.
+    #
+    # All caches are separated by operating system. Both ccache and deps are
+    # additionally deparated by `cache-key`, i.e. the CI job type, because they
```
mpreiner, line 133:
```suggestion
    # additionally separated by `cache-key`, i.e. the CI job type, because they
```

mpreiner (approved):

PR #6266, opened 2021-04-01
Disable substring component contains in strip endpoints

ajreynol:
Fixes the first benchmark from #6203.


4tXJ7f (approved):
LGTM, thanks for the fix.

PR #6267, opened 2021-04-01
Fix case where RE unfolding generates a trivially true lemma

ajreynol:
An RE unfolding lemma may rewrite to true for tautological RE memberships that our rewriter does not rewrite the membership to true.

An example is `(str.in_re x (re.* (re.union (str.to_re "A") (str.to_re x)))`.

This PR ensures we are robust to these cases.

This fixes benchmarks 3-5 from #6203.  Benchmark 3 is added here, 4-5 time out.


4tXJ7f (approved):
LGTM

PR #6269, opened 2021-04-02
cmake: Do not link against main object library.

mpreiner:
CMake 3.10.2 (default on Ubuntu 18.04) does not allow `target_link_libraries` on object libraries.


aniemetz (approved):

nafur (approved):
LGTM.

PR #6275, opened 2021-04-05
[proof-new] Registering proof checkers uniformly from the SMT solver

HanielB:
Each theory has its own proof checker, responsible for checking the rules pertaining to that theory. The main proof checker uses these specialized checkers. Previously the main proof checker (of the proof node manager used across the SMT solver) was connected to these theory proof checkers during initialization of the theory. This commit adds an interface to the theories for retrieving its proof checker (analogous to how one retrieves the rewriter of a theory) which is used by a new method in the theory engine to register a theory proof checker to a given proof checker according to a theory id.

This is in preparation for the new unsat cores based on proofs.


src/smt/smt_solver.cpp
```diff
@@ -66,6 +66,16 @@ void SmtSolver::finishInit(const LogicInfo& logicInfo)
     theory::TheoryConstructor::addTheory(d_theoryEngine.get(), id);
   }
 
+  // Add the proof checkers for each theory
+  if (d_pnm)
+  {
+    ProofChecker* pc = d_pnm->getChecker();
+    for (theory::TheoryId id = theory::THEORY_FIRST; id < theory::THEORY_LAST;
```
ajreynol, line 73: I'd prefer this loop was inside of TheoryEngine::initializeProofChecker, meaning TheoryEngine::initializeProofChecker should not take a TheoryId.

ajreynol (changes requested):

src/smt/smt_solver.cpp
ajreynol: I don't think there is a reason to be this fine-grained.  I think it'd best to keep it simple, e.g. say a special preprocessing pass used an arithmetic rule.

HanielB: Ah, so you mean we should always just initialize them all?

ajreynol: Yes

HanielB: I see. I'll update.

src/theory/bags/theory_bags.h
```diff
@@ -28,6 +28,9 @@
 #include "theory/theory_eq_notify.h"
 
 namespace cvc5 {
+
+class ProofRuleChecker;
```
ajreynol, line 32: This shouldn't be necessary.

src/theory/fp/theory_fp.h
```diff
@@ -32,6 +32,9 @@
 #include "theory/uf/equality_engine.h"
 
 namespace cvc5 {
+
+class ProofRuleChecker;
```
ajreynol, line 36: Ditto.

src/theory/sep/theory_sep.h
```diff
@@ -32,6 +32,9 @@
 #include "util/statistics_registry.h"
 
 namespace cvc5 {
+
+class ProofRuleChecker;
```
ajreynol, line 36: Ditto.

src/theory/sets/theory_sets.h
```diff
@@ -30,6 +30,9 @@
 #include "theory/uf/equality_engine.h"
 
 namespace cvc5 {
+
+class ProofRuleChecker;
```
ajreynol, line 34: Ditto.

ajreynol (changes requested):
A few more minor requests.

ajreynol (approved):
LGTM

PR #6277, opened 2021-04-05
[proof-new] Fixing SMT post-processor's handling of assumptions

HanielB:
Previously the SMT post-processor would update *any* assumption as long as it
had a proof for it. This can be a problem when one as assumption introduced in a
scope that should not be expanded. This commit fixes the issue by adding the
option of configuring a proof node updater to track scopes and the assumptions
they introduce, which can be used to determine the prood nodes which should be
updated. It also changes the SMT post-processor to only update assumptions that
have not been introduced in some scope.

This commit fixes an issue found by @Lachnitt during the integration of CVC4 and Isabelle.


src/expr/proof_node_updater.cpp
```diff
@@ -125,28 +125,21 @@ void ProofNodeUpdater::processInternal(
       }
       traversing.push_back(cur);
       visit.push_back(cur);
-      if (d_mergeSubproofs)
+      // If we are not the top-level proof, we were a scope, or became a scope
```
ajreynol, line 128: We should still check d_mergeSubproofs here, right?

src/smt/proof_manager.cpp
```diff
@@ -33,7 +33,11 @@ PfManager::PfManager(context::UserContext* u, SmtEngine* smte)
       d_pnm(new ProofNodeManager(d_pchecker.get())),
       d_pppg(new PreprocessProofGenerator(
           d_pnm.get(), u, "smt::PreprocessProofGenerator")),
-      d_pfpp(new ProofPostproccess(d_pnm.get(), smte, d_pppg.get())),
+      d_pfpp(new ProofPostproccess(
+          d_pnm.get(),
+          smte,
+          d_pppg.get(),
+          options::proofFormatMode() != options::ProofFormatMode::VERIT)),
```
ajreynol, line 40: Comment on the reason for this.

ajreynol (changes requested):

src/expr/proof_node_updater.cpp
HanielB: In that case we'd do the tracking only when we are merging subproofs. The way I see it this tracking serves for more than doing the merges correctly, so I didn't want to have it constrained by this.

ajreynol: Ah, I see. 
Yes, I think this was a micro-optimization that we don't need, I agree with the change.

src/smt/proof_manager.cpp
HanielB: done

ajreynol (approved):

PR #6278, opened 2021-04-05
Fix subtyping for sets care graph

### LINKED ISSUE #5705: Invalid model for set constraint
CVC4 39df95de07
~~~~
(set-logic QF_LIRAFS)
(declare-fun s () (Set Real))
(declare-fun t3 () (Set Real))
(assert (or (member 1.0 t3) (member 0.0 s)))
(assert (not (= t3 (setminus s (singleton 1.0)))))
(check-sat)
~~~~
~~~~
./cvc4 --check-models ../delta.out.smt2 
Fatal failure within void CVC4::smt::CheckModels::checkModel(CVC4::smt::Model*, CVC4::context::CDList<CVC4::NodeTemplate<true> >*, bool) at /home/CVC4/src/smt/check_models.cpp:129
assertion:     (not (= t3 (setminus s (singleton 1.0))))
simplifies to: false
expected `true'.
Aborted (core dumped)
~~~~

ajreynol:
We were getting types for set singleton/membership in a way that was unsafe for subtyping, which was leading to incorrectly computing care graphs for sets of reals.

Fixes #5705.


src/theory/sets/theory_sets_private.cpp
```diff
@@ -985,9 +985,20 @@ void TheorySetsPrivate::computeCareGraph()
       {
         Trace("sets-cg-debug") << "...build for " << f1 << std::endl;
         Assert(d_equalityEngine->hasTerm(f1));
-        // break into index based on operator, and type of first argument (since
-        // some operators are parametric)
-        TypeNode tn = f1[0].getType();
+        // break into index based on operator, and the type of the element
+        // type of the proper set, which notice must be safe wrt subtyping.
```
mudathirmahgoub, line 989: `,and the type of the element type of the proper set,`
Do you mean the `get the type of the element from the type of the proper set`?

ajreynol: No, this is the purpose of the fix: we need to consider the element type of the set, not the type of the element in the set.

mudathirmahgoub (approved):
LGTM. This was a tricky one.

PR #6280, opened 2021-04-05
Remove stdPrintAscii option

### LINKED ISSUE #4179: strings-print-ascii soundness bug on QF_S formula
Hi,
For this formula:
```
(set-option :strings-print-ascii true)
(declare-fun a () String)
(assert (= (str.code a) 128))
(check-sat)
```
CVC4  gives an incorrect answer:
```
[555] % cvc4-1.7 --quiet small.smt2
sat
[556] % cvc4 --quiet small.smt2
unsat
[557] % 
[557] % cat small.smt2
(set-option :strings-print-ascii true)
(declare-fun a () String)
(assert (= (str.code a) 128))
(check-sat)
```

OS: Ubuntu 18.04
Commit: 01b2570

ajreynol:
Fixes #4179.


4tXJ7f (approved):
Glad to get rid of this option :)

PR #6281, opened 2021-04-05
Set incomplete if not applying ho extensionality

### LINKED ISSUE #4318: (uf-ho, uf-ho-ext) Invalid model for QF_UFLIA formula
Hi,
For this case, CVC4  gives an invalid model:
```
[917] % cvc4 -q small.smt2
Fatal failure within void CVC4::TheoryEngine::checkTheoryAssertionsWithModel(bool) at /home/suz/software/CVC4/src/theory/theory_engine.cpp:2283
Internal error detectedTHEORY_UF has an asserted fact that the model doesn't satisfy.
The fact: (not (= a c))
Model value: false
Aborted
[918] % 
[918] % cat small.smt2
(set-option :uf-ho true)
(set-option :uf-ho-ext false)
(set-option :check-models true)
(declare-fun a (Int) Int)
(declare-fun b (Int) Int)
(declare-fun c (Int) Int)
(assert (xor (distinct a b) (distinct a c)))
(check-sat)
[919] %
```
It might be related to  https://github.com/CVC4/CVC4/issues/4240 .

OS: Ubuntu 18.04
Commit: 912b650

ajreynol:
If the user manually disables ho-extensionality via expert option --no-uf-ho-ext, we should answer "unknown" instead of "sat" when applicable.

Fixes #4318.


HanielB (approved):
LGTM

PR #6284, opened 2021-04-05
Fix spurious antecedant for symbolic regular expressions

### LINKED ISSUE #6271: (strings-fmf) Invalid model for string formula
~~~~
(set-option :strings-fmf true)
(declare-fun str7 () String)
(declare-fun str8 () String)
(assert (str.in_re str8 (re.++ re.allchar re.allchar (str.to_re str7))))
(check-sat)
~~~~
~~~~
cvc4 -q -i --check-models ~/work/delta.out.smt2 
Fatal failure within void cvc5::smt::CheckModels::checkModel(cvc5::smt::Model*, cvc5::context::CDList<cvc5::NodeTemplate<true> >*, bool) at /home//CVC4/src/smt/check_models.cpp:131
assertion:     (str.in_re str8 (re.++ re.allchar  re.allchar  (str.to_re str7)))
simplifies to: false
expected `true'.
Aborted (core dumped)
~~~~

ajreynol:
Fixes #6271.

This was triggered by recent fixes, this fixes solution soundness issues with symbolic regular expressions due to spuriously included antecedants, which made lemmas SAT-context dependent while being cached as user-context dependent.


4tXJ7f (approved):
LGTM, nice find.

PR #6285, opened 2021-04-05
(proof-new) Proper implementation of proof node cloning

ajreynol:
Previously, we were traversing proof node as a tree, now we use a dag traversal.

This also makes sure that proofs work when we have a external proof conversion and we are in incremental mode. In such cases, the final proof must be cloned to ensure that we do not overwrite proof nodes, which may be reused across multiple check-sat.


HanielB (approved):
LGTM

PR #6289, opened 2021-04-05
Fix issue with lemma during equality engine iterator in sets

### LINKED ISSUE #4370: Fatal failures at src/theory/uf/equality_engine_iterator.cpp:104

ajreynol:
Fixes #4370.

### DIFF
--- a/src/theory/sets/theory_sets_private.cpp
+++ b/src/theory/sets/theory_sets_private.cpp
@@ -286,6 +286,10 @@ void TheorySetsPrivate::fullEffortCheck()
           d_card_enabled = true;
           // register it with the cardinality solver
           d_cardSolver->registerTerm(n);
+          if (d_im.hasSent())
+          {
+            break;
+          }
           // if we do not handle the kind, set incomplete
           Kind nk0 = n[0].getKind();
           // some kinds of cardinality we cannot handle
--- a/test/regress/CMakeLists.txt
+++ b/test/regress/CMakeLists.txt
@@ -1968,6 +1968,7 @@ set(regress_1_tests
   regress1/sets/issue2568.smt2
   regress1/sets/issue2904.smt2
   regress1/sets/issue4124-need-check.smt2
+  regress1/sets/issue4370-2-lemma-ee-iter.smt2
   regress1/sets/issue4391-card-lasso.smt2
   regress1/sets/issue5271.smt2
   regress1/sets/issue5309.smt2
--- a/test/regress/CMakeLists.txt
+++ b/test/regress/CMakeLists.txt
@@ -1969,6 +1969,7 @@ set(regress_1_tests
   regress1/sets/issue2904.smt2
   regress1/sets/issue4124-need-check.smt2
   regress1/sets/issue4370-2-lemma-ee-iter.smt2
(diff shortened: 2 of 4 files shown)


src/theory/sets/theory_sets_private.cpp
```diff
@@ -286,6 +286,10 @@ void TheorySetsPrivate::fullEffortCheck()
           d_card_enabled = true;
           // register it with the cardinality solver
           d_cardSolver->registerTerm(n);
+          if (d_im.hasSent())
+          {
+            break;
```
mudathirmahgoub, line 291: Why does this break solve the issue,  when there is a  new lemma  or fact?

ajreynol: `d_cardSolver->registerTerm` may send a lemma, which then calls Theory::preRegisterTerm, which in turn modifies the equality engine, making the equality engine iterator we are using invalid.

mudathirmahgoub: Thanks for the clarification.

mudathirmahgoub (approved):
LGTM

mudathirmahgoub (approved):

PR #6293, opened 2021-04-06
genkinds: Do not use relative paths to find src directory.

mpreiner:
`sys.path.insert(0, os.path.abspath('../../../../src/api'))` will not work if the build directory is in a subdirectory, see #6244.


mudathirmahgoub (approved):
LGTM. Thanks for removing this relative path. I didn't know you can use CMake variables inside python code.

PR #6294, opened 2021-04-06
parsekinds: Remove DEFAULT_HEADER.

mpreiner:
`DEFAULT_HEADER` in `src/api/parsekinds.py` is essentially unused since both `genkinds.py` scripts pass the kinds header to the script. The current value of `DEFAULT_HEADER` does not work for the scripts since the working directory for genkinds.py is in `src/api/{java,python}`.


makaimann (approved):
LGTM, thanks!

PR #6300, opened 2021-04-07
cmake: Do not always regenerate cvc4kinds.{pxi,pxd}.

mpreiner:
Only regenerate files if dependencies changed. This avoids unnecessary
rebuilds of the Cython code.


makaimann (approved):
LGTM, thanks for the fix! I did notice that there are some `cvc5`s and some `cvc4`s but assumed they’re all correct for the current status of the transition.

mpreiner:
> LGTM, thanks for the fix! I did notice that there are some `cvc5`s and some `cvc4`s but assumed they’re all correct for the current status of the transition.

Yep, that's the current status.

PR #6301, opened 2021-04-07
Add benchmark for issue 5101

### LINKED ISSUE #5101: (default) Model validator issue on QF_AUFLIRA formula
Hi, for the following formula
~~~~
(set-logic QF_AUFLIRA)
(declare-fun _substvar_173_ () Bool)
(declare-const v9 Bool)
(declare-const r0 Real)
(declare-const r7 Real)
(declare-const r10 Real)
(declare-const i17 Int)
(declare-const arr (Array Real Real))
(assert (or _substvar_173_ (> 0.0 r10 (select (store arr r0 (to_real i17)) 817949693.0))))
(assert (or v9 (is_int r7)))
(check-sat)
~~~~

cvc4  5f04a78336 `--check-models` yields
~~~~
Fatal failure within void CVC4::SmtEngine::checkModel(bool) at /home/CVC4-Release/src/smt/smt_engine.cpp:1677
model value for i17
             is (/ (- 1) 2)
value type is     Real
should be of type Int
Aborted
~~~~

In the following model, `v9` is true. Thus the type of i17 is irrelevant for the formula to be satisfiable.
~~~~
sat
(model
(define-fun _substvar_173_ () Bool false)
(define-fun v9 () Bool true)
(define-fun r0 () Real 817949693.0)
(define-fun r7 () Real (/ 1 4))
(define-fun r10 () Real (/ (- 1) 4))
(define-fun i17 () Int (/ (- 1) 2))
(define-fun arr () (Array Real Real) (store ((as const (Array Real Real)) 0) 817949693.0 1.0))
)
~~~~

ajreynol:
Fixes #5101.


nafur (approved):
LGTM.

Issue #6100, opened 2021-03-10
Fatal failure cvc4 1.8 at type_checker_template.cpp:51 with --finite-model-find


robby-phd:
**Describe the bug**

cvc4 1.8 with `--finite-model-find` fails for the attached input.

[cvc4-bug.smt2.gz](https://github.com/CVC4/CVC4/files/6112170/cvc4-bug.smt2.gz)

```bash
$ cvc4 cvc4-bug.smt2
unknown
$ cvc4 --finite-model-find cvc4-bug.smt2
Fatal failure within static CVC4::TypeNode CVC4::expr::TypeChecker::computeType(CVC4::NodeManager *, CVC4::TNode, bool) at /Users/noetzli/repos/CVC4_release/src/expr/type_checker_template.cpp:51
Unhandled case encounteredNULL
```

**Command line arguments**: --finite-model-find
**CVC4 version/commit**: [1.8 release binary [git master 3bc4a541]](https://github.com/CVC4/CVC4/releases/download/1.8/cvc4-1.8-macos-opt)
**Operating system**: macOS 11.2.3

ajreynol:
Hi Robby,

Thanks for the benchmark.  

This is due to the combination of arrays and finite model finding.  There is a sort (Array Int A) where the uninterpreted sort A is assigned cardinality one. This makes the sort (Array Int A) have cardinality one, but certain parts of the code are assuming this is an infinite type. This makes our model builder run out of values, leading to the error.

The solution will be to ensure that we properly handle theory combination for finite model finding + arrays.  Note this issue may be related: https://github.com/CVC4/CVC4/issues/4260

FYI @barrettcw .

Issue #4260, opened 2020-04-10
Fatal failure at theory/theory_model_builder.cpp:796 FMF+Arrays combination


rainoftime:
Hi,
For the following formula:
~~~~
(set-logic QF_AUFNIA)
(set-option :fmf-fun-rlv true)
(set-option :sygus-inference true)
(declare-sort S0 0)
(declare-const arr-6648601738383876656_-5336841048752577678-0 (Array Int S0))
(declare-const arr-6648601738383876656_-5336841048752577678-1 (Array Int S0))
(assert (distinct arr-6648601738383876656_-5336841048752577678-1 arr-6648601738383876656_-5336841048752577678-0))
(check-sat)
~~~~

CVC4 throws out a fatal failure:
~~~~
Fatal failure within virtual bool CVC4::theory::TheoryEngineModelBuilder::buildModel(CVC4::Model*) at /home/CVC4/src/theory/theory_model_builder.cpp:817
Check failure

 !n.isNull()

Aborted (core dumped)
~~~~

OS: Ubuntu 16.04
Commit: 2f8caab

ajreynol:
It looks like this has to do with theory combination of Arrays + UF when finite model finding is enabled.  Changing the logic to quantified and command line --dump-models only triggers the bug.

ajreynol:
```
(set-logic AUFNIA)
(declare-sort S0 0)
(declare-const a (Array Int S0))
(declare-const b (Array Int S0))
(assert (distinct b a))
(check-sat)
```
with `--finite-model-find --dump-models`.

ajreynol:
Works with `--theoryof-mode=type`.

rainoftime:
~~~~
(set-logic AUFLIRA)
(set-option :fmf-fun-rlv true)
(declare-fun uf6 (Bool Bool Bool Bool Bool Bool) Bool)
(declare-sort S0 0)
(assert (uf6 true true true true true (not (exists ((q65 Bool) (q66 Int) (q67 (Array Real S0)) (q68 (Array Real S0)) (q69 Int) (q70 (Array Int Real))) (xor true true true (= q68 q68 q67 q68))))))
(check-sat)
~~~~

PR #6302, opened 2021-04-07
Add cardinality class definition

ajreynol:
This is work towards correcting our computation of whether a type is finite.  Currently, arrays/functions with uninterpreted sorts as element/range types are always considered infinite.  This is incorrect if finite model finding is enabled, since the interpretation of the uninterpreted sort can be one.  This leads to errors during model building due to exhausted values (#4260, #6100).

This PR adds a new concept of a cardinality class, which is required for properly categorizing types with/without finite model finding.

A followup PR will replace TypeNode::isFinite with TypeNode::getCardinalityClass.  Calls to TypeNode::isFinite will be replaced by calls to TheoryState::isTypeFinite, which will properly take cardinality classes into account.


src/util/cardinality_class.h
```diff
@@ -0,0 +1,94 @@
...
+/**
+ * Writes a cardinality class name to a stream.
+ *
+ * @param out The stream to write to
+ * @param c The cardinality class to write to the stream
+ * @return The stream
+ */
+std::ostream& operator<<(std::ostream& out, CardinalityClass c);
+
+/** Take the min class of c1 and c2 */
```
yoni206, line 79: For these two functions, I suggest explicitly defining an ordering over cardinality classes. Maybe mention in the enum that they are listed in their order (for example, it seems arbitrary that UNKNOWN is bigger/smaller than infinite.

src/util/cardinality_class.h
```diff
@@ -0,0 +1,94 @@
...
+/** Take the max class of c1 and c2 */
+CardinalityClass maxCardinalityClass(CardinalityClass c1, CardinalityClass c2);
+/**
+ * Is a type with the given cardinality class finite?
+ *
+ * If fmfEnabled is true, then this method assumes that uninterpreted sorts
+ * have cardinality one. If fmfEnabled is false, then this method assumes that
+ * uninterpreted sorts have infinite cardinality.
+ */
+bool isCardinalityClassFinite(CardinalityClass c, bool fmfEnabled);
```
yoni206, line 90: it might be nice to have this function `static` to clarify that it does not depend on any state of the solver. Also for other functions. Not critical though.

yoni206 (approved):
LGTM, thanks for initiating this, and especially for splitting this feature to several PRs!
Left some minor suggestions for your consideration before merging.

src/util/cardinality_class.h
ajreynol: They are global functions actually (`CardinalityClass` is not a class), so they don't depend on any state.

PR #6303, opened 2021-04-07
Use exceptions when constructing malformed datatypes

### LINKED ISSUE #5150: Bug: Fatal Failure (datatypes parsing)
**Describe the bug**
apparently legal smt2 program causes fatal failure in CVC4 version 1.9-prerelease

```
(set-logic ALL)
(declare-datatype Switch ((On (Bool val)) (Off)))

(define-fun CheckSwitch ((s Switch)) Bool
    (match s ((On val) val) (Off false)))
)

(declare-const x Switch)
(assert (= (CheckSwitch x) false))
(check-sat)
```

**CVC4 version/commit**: Paste CVC4 version/commit here.

This is CVC4 version 1.9-prerelease [git master edf2e3ff]
compiled with GCC version 10.2.0
on Sep 12 2020 17:20:51

**Operating system**: Name/version of the operating system.
Linux 5.7.0-2-amd64 #1 SMP Debian 5.7.10-1 (2020-07-26) x86_64 GNU/Linux

ajreynol:
Fixes #5150.


src/expr/node_manager.cpp
```diff
@@ -599,9 +599,11 @@ std::vector<TypeNode> NodeManager::mkMutualDatatypeTypes(
 
       typeNode = mkTypeNode(kind::PARAMETRIC_DATATYPE, params);
     }
-    AlwaysAssert(nameResolutions.find(dtp->getName()) == nameResolutions.end())
-        << "cannot construct two datatypes at the same time "
-           "with the same name";
+    if (nameResolutions.find(dtp->getName()) != nameResolutions.end())
+    {
+      throw Exception(
```
aniemetz, line 604: Should this be a type checking exception?

ajreynol: I thought the same at first, but TypeCheckingException requires a Node, which we don't have here.

aniemetz (approved):

PR #6307, opened 2021-04-07
Formalize more skolems

ajreynol:
This formalizes more skolems in preparation for moving Theory::expandDefinitions to Rewriter::expandDefinitions.

It also adds proof support for datatypes purification.


HanielB (approved):
LGTM

PR #6311, opened 2021-04-08
Add identifiers for sources of incompleteness

ajreynol:
This PR classifies all internal kinds of incompleteness as identifiers.

It makes it so TheoryEngine records an identifier when its incomplete flag is set.

The next step will be to possibly communicate this value to the user.


nafur (approved):
LGTM.

PR #6313, opened 2021-04-08
Add more missing inference ids

ajreynol:
This also makes the relations solver use the inference manager in the standard way.


src/theory/inference_id.h
```diff
@@ -66,6 +66,13 @@ enum class InferenceId
   ARITH_BB_LEMMA,
   ARITH_DIO_CUT,
   ARITH_DIO_DECOMPOSITION,
+  // unate lemma during presolve
+  ARITH_UNATE,
+  // row implication
+  ARITH_ROW_IMPL,
```
nafur, line 72: Can you elaborate what this means?

src/theory/fp/theory_fp.cpp
```diff
@@ -801,11 +802,13 @@ void TheoryFp::convertAndEquateTerm(TNode node) {
 #ifdef SYMFPUPROPISBOOL
       handleLemma(nm->mkNode(kind::EQUAL, node, converted));
```
nafur, line 803: What about this case?

src/theory/inference_id.h
```diff
@@ -229,6 +236,12 @@ enum class InferenceId
   DATATYPES_SYGUS_MT_POS,
   // ---------------------------------- end datatypes theory
 
+  //-------------------------------------- floating point theory
+  FP_PREPROCESS,
+  FP_EQUATE_TERM,
+  FP_REGISTER_TERM,
```
nafur, line 242: Some docs here as well? Should we split the FP_PREPROCESS into more specific ids?

nafur (changes requested):
Mostly LGTM, but a few questions.

src/theory/inference_id.h
ajreynol: I actually have no idea.  @alex-ozdemir may know since he did proofs for this part?

src/theory/fp/theory_fp.cpp
ajreynol: I wasn't sure about the code that is under this flag.  There is another part of the code under this flag that appears to use a different interface for outputting lemmas.  @aniemetz  what is the status of this `SYMFPUPROPISBOOL`?  Are we planning to maintain this?

src/theory/inference_id.h
ajreynol: I'd prefer this level of granularity for now, as I'm not sure if the ways that FP sends lemmas will be further refactored.  @aniemetz do you have plans to change how FP makes calls to lemma?

src/theory/fp/theory_fp.cpp
aniemetz: That's a very good question. @martin-cs ?

nafur (approved):
As per offline discussion, we merge this PR now and fill in the gaps in the documentation later.

PR #6314, opened 2021-04-08
Add identifiers for extended function reductions

ajreynol:
This adds identifiers for extended function reductions, which are reasons for why an extended term no longer needs to be processed.  The motivation is help understand check-model failures.

This PR adds identifiers to the ExtTheory utility. It also cleans up some unused parts of this utility.  Some blocks of code changed indentation in this class.

### DIFF
--- a/src/theory/ext_theory.h
+++ b/src/theory/ext_theory.h
@@ -163,9 +163,9 @@ class ExtTheoryCallback
  */
 class ExtTheory
 {
-  typedef context::CDHashMap<Node, bool, NodeHashFunction> NodeBoolMap;
+  using NodeBoolMap = context::CDHashMap<Node, bool, NodeHashFunction>;
   using NodeExtReducedIdMap = context::CDHashMap<Node, ExtReducedId, NodeHashFunction>;
-  typedef context::CDHashSet<Node, NodeHashFunction> NodeSet;
+  using NodeSet = context::CDHashSet<Node, NodeHashFunction>;
 
  public:
   /** constructor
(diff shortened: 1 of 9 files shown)


src/theory/ext_theory.h
```diff
@@ -242,6 +279,10 @@ class ExtTheory
   static std::vector<Node> collectVars(Node n);
   /** is n context dependent inactive? */
   bool isContextIndependentInactive(Node n) const;
+  /**
+   * Same as above, but rid is update to the reason if this method returns true.
```
mpreiner, line 283:
```suggestion
   * Same as above, but rid is updated to the reason if this method returns true.
```

src/theory/ext_theory.h
```diff
@@ -232,6 +264,11 @@ class ExtTheory
   bool hasActiveTerm() const;
   /** is n an active extended function term? */
   bool isActive(Node n) const;
+  /**
+   * Same as above, but rid is update to the reason if this method returns
```
mpreiner, line 268:
```suggestion
   * Same as above, but rid is updated to the reason if this method returns
```

src/theory/ext_theory.h
```diff
@@ -118,6 +164,8 @@ class ExtTheoryCallback
 class ExtTheory
 {
   typedef context::CDHashMap<Node, bool, NodeHashFunction> NodeBoolMap;
+  typedef context::CDHashMap<Node, ExtReducedId, NodeHashFunction>
+      NodeExtReducedIdMap;
```
mpreiner, line 168:
```suggestion
  using NodeExtReducedIdMap = context::CDHashMap<Node, ExtReducedId, NodeHashFunction>;
```

mpreiner: Maybe also change the typedef above.

mpreiner (approved):

PR #6318, opened 2021-04-08
Avoid spurious runs in run_regression.py

ajreynol:
The options `--check-synth-sol` and `--check-abducts` are independent from the rest of the solver.  Hence, it is not necessary to run with/without them enabled in our regressions. Running only with them suffices.

This saves roughly 30-40 seconds on regression regress0-2.


HanielB (approved):
LGTM

PR #6321, opened 2021-04-09
Enable expr miner timeout

ajreynol:
We currently were ignoring the option `--sygus-expr-miner-timeout=N`.  This corrects the issue.


HanielB (approved):
LGTM

PR #6322, opened 2021-04-09
Refactor resource manager

nafur:
This PR does another round of refactoring of the resource manager and related code.
- it moves the `Resource` enum out of the `ResourceManager` class
- it treats the resources in a generic way (storing the statistics in a vector) instead of the manual treatment we had before
- weights no longer live in the options, but in the `ResourceManager` and are changed accordingly in the `ResourceManager` constructor
- following the generic treatment of resources, it also removes all the resource-specific options `--x-step` in favor of a generic `--rweight name=weight`
- removed several unused methods from the `ResourceManager`

Note that we handle the `Resource` enum in a way that allows to easily use other enums as additional resources, for example `InferenceId`. The general idea is that we will at some point have sensible default weights (so that the cumulative resources somewhat simulate the solver runtime) and users (almost) never need to modify them.


src/options/smt_options.toml
```diff
@@ -467,175 +467,18 @@ header = "options/smt_options.h"
...
-  name       = "bvSatSimplifyStep"
-  category   = "expert"
-  long       = "bv-sat-simplify-step=N"
-  type       = "unsigned"
-  default    = "1"
-  read_only  = true
-  help       = "amount of resources spent for each sat simplify (bitvectors)"
+  name       = "resourceWeightHolder"
+  category   = "undocumented"
+  type       = "std::vector<std::string>"
```
mpreiner, line 481: Why is this needed?

nafur: I want to store a string for every usage of `--rweight`. Other options that may occur multiple time use handlers, but I want to collect them and then iterate over them later on. I store the `--rweight` values in this other option, and then access this other option later.

mpreiner: OK, how about adding a comment above this option that describes what it is used for and that it is populated by `setResourceWeight`?

nafur: Done.

mpreiner (approved):

PR #6323, opened 2021-04-09
Learn equalities involving Boolean variables

4tXJ7f:
Previously, the circuit propagator was not learning literals of the form `(= x
t)` where `x` is Boolean, since this term was not treated as a theory literal.
This commit changes that, which improves performance significantly, since it
allows the elimination of Boolean variables, which, in turn, can make the
justification heuristic much more effective.

### DIFF
--- a/src/theory/booleans/circuit_propagator.cpp
+++ b/src/theory/booleans/circuit_propagator.cpp
@@ -716,7 +716,9 @@ TrustNode CircuitPropagator::propagate()
                     && (current[0].isVar() && current[1].isVar()));
 
     // If an atom, add to the list for simplification
-    if (atom)
+    if (atom
+        || (current.getKind() == kind::EQUAL
+            && (current[0].isVar() || current[1].isVar())))
     {
       Debug("circuit-prop")
           << "CircuitPropagator::propagate(): adding to learned: "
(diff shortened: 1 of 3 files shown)


src/theory/booleans/circuit_propagator.cpp
```diff
@@ -716,7 +716,9 @@ TrustNode CircuitPropagator::propagate()
                     && (current[0].isVar() && current[1].isVar()));
 
     // If an atom, add to the list for simplification
-    if (atom)
+    if (atom
```
ajreynol, line 719: Can we just refactor the definition of atom to do this?  I actually think the 3rd case of atom above is a stronger version of this?

ajreynol (changes requested):

src/theory/booleans/circuit_propagator.cpp
4tXJ7f: The PR doesn't change `atom` because it is used in a different context below. The idea is that equalities between Boolean variables and terms are treated as theory literals for the purpose of learning them but otherwise as regular Boolean terms.

ajreynol (approved):
LGTM

ajreynol:
Do post the results of an SMT-LIB run here when they are available.

PR #6324, opened 2021-04-09
[proof-new] Optimizing sat proof

HanielB:
For some benchmarks, checking MACRO_RESOLUTION can be up to 80% (!!!) of the running time. This commit introduces a new rule that does not perform checking. The old rule and checker are kept for ground truth. Some miscellaneous minor changes are also made in the PR.

### DIFF
--- a/src/theory/booleans/proof_checker.cpp
+++ b/src/theory/booleans/proof_checker.cpp
@@ -25,6 +25,7 @@ void BoolProofRuleChecker::registerTo(ProofChecker* pc)
   pc->registerChecker(PfRule::SPLIT, this);
   pc->registerChecker(PfRule::RESOLUTION, this);
   pc->registerChecker(PfRule::CHAIN_RESOLUTION, this);
+  pc->registerChecker(PfRule::MACRO_RESOLUTION_TRUST, this);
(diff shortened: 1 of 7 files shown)


src/theory/booleans/proof_checker.cpp
```diff
@@ -25,6 +25,7 @@ void BoolProofRuleChecker::registerTo(ProofChecker* pc)
   pc->registerChecker(PfRule::SPLIT, this);
   pc->registerChecker(PfRule::RESOLUTION, this);
   pc->registerChecker(PfRule::CHAIN_RESOLUTION, this);
+  pc->registerChecker(PfRule::MACRO_RESOLUTION_TRUST, this);
```
ajreynol, line 28: This should be a trusted rule with a pedantic level, e.g. 3, see
https://github.com/CVC4/CVC4/blob/master/src/theory/builtin/proof_checker.cpp#L73

ajreynol (changes requested):
One minor comment, else LGTM

src/theory/booleans/proof_checker.cpp
HanielB: done

ajreynol (approved):

PR #6325, opened 2021-04-09
New C++ Api: Initial layout of Api documentation.

aniemetz:
As discussed offline, this implements the proposed layout.


nafur (approved):
LGTM. Seems like a good basis we can build on.

PR #6328, opened 2021-04-11
Avoid using substitute's input cache after the method call.

abdoo8080:
As it traverses a node, `Node::substitute` populates its input cache with `TNode`s that are not preserved by the `SygusReconstruct` module and maybe destroyed after the method call. This PR fixes a bug where those unsafe `TNode`s are referenced throughout the module by passing the method a temporary copy of the cache instead.


ajreynol (commented):
This is a good find :)

I would recommend a different fix:  Instead of making a fresh copy of subs every time we apply it, can we either (A) replace d_sol with parallel vectors `d_solVars / d_solSubs` and use the substitute using iterators (https://github.com/CVC4/CVC4/blob/master/src/expr/node.h#L238)
(B) use the parallel vectors in addition to tracking d_sol.
Either case, we would not have to do a large copy when applying this substitute, and this would fix the issue since `d_sol` would not be modified.

4tXJ7f:
I agree with Andy but I think that there is an even easier solution, which is to use [this](https://github.com/CVC4/CVC4/blob/master/src/expr/node.h#L507-L513) overload, i.e. changing the call to `sz.substitute(sol.begin(), sol.end())`.

ajreynol (approved):

PR #6329, opened 2021-04-11
Fix sexpr bug with AST output language.

abdoo8080:
When a command is invoked, its result/status may be printed depending on its current verbosity, which (for probably an outdated reason) is stored in `SMTEngine`. One of my previous PRs modified the `SMTEngine` to return the verbosity as an `sexpr`. The modification works correctly when the output language is SMT2, but it breaks down with the AST output language, which prints `sexpr`s in a different way. This PR fixes this bug by making `SMTEngine` return options as strings instead of `sexpr`.


test/regress/regress0/options/ast-and-sexpr.smt2
```diff
@@ -0,0 +1,10 @@
+; EXPECT: (SEXPR (SEXPR check-sat (CONST_RATIONAL 2)) (SEXPR * (CONST_RATIONAL 1)))
+; EXPECT: (:status unknown)
+
+(set-option :output-language ast)
+(set-option :command-verbosity (* 1))
```
ajreynol, line 5: Can you comment in the benchmark (with `;`) what these commands mean?

src/smt/smt_engine.cpp
```diff
@@ -1983,30 +1983,23 @@ Node SmtEngine::getOption(const std::string& key) const
...
-  }
-  try
-  {
-    Integer z(atom);
-    return nm->mkConst(Rational(z));
-  }
-  catch (std::invalid_argument&)
-  {
-    // Fall through to the next case
+  if (atom != "true" && atom != "false") {
```
ajreynol, line 1991: Remember to clang-format

ajreynol (changes requested):

test/regress/regress0/options/ast-and-sexpr.smt2
abdoo8080: Done!

src/smt/smt_engine.cpp
abdoo8080: Done!

ajreynol (approved):

PR #6334, opened 2021-04-12
Strings: Move implementation of type rules from header to .cpp file.

aniemetz:


src/theory/strings/theory_strings_type_rules.cpp
```diff
@@ -0,0 +1,376 @@
...
+ *
+ * This file is part of the cvc5 project.
+ *
+ * Copyright (c) 2009-2021 by the authors listed in the file AUTHORS
+ * in the top-level source directory and their institutional affiliations.
+ * All rights reserved.  See the file COPYING in the top-level source
+ * directory for licensing information.
+ * ****************************************************************************
+ *
+ * [[ Add one-line brief description here ]]
```
ajreynol, line 13: Is this text intentional? To be filled later?

ajreynol (approved):
LGTM, one minor question.

src/theory/strings/theory_strings_type_rules.cpp
aniemetz: Added

PR #6339, opened 2021-04-13
Add InferenceId as resources

nafur:
This PR extends the resource manager to consider `theory::InferenceId` a resource we can spend. This should give us a robust way to have the resource limiting cover a lot of theory reasoning. To make use of this, `TheoryInferenceManager` now spends these resources.
Also, it makes the `ResourceManager` properly use the options from the `Env` class.


src/smt/env.h
```diff
@@ -61,7 +61,7 @@ class Env
   /**
    * Construct an Env with the given node manager.
    */
-  Env(NodeManager* nm);
+  Env(NodeManager* nm, Options* opts);
```
ajreynol, line 64: Since we're adding options via constructor, we can probably remove the `setOptions` method.  This method only existed due to construction order dependencies previously.

ajreynol (changes requested):

src/smt/env.h
nafur: I agree.

ajreynol (approved):

PR #6340, opened 2021-04-13
Improve documentation for API exceptions

nafur:
This PR improves the documentation for exceptions in the C++ API and explores some general techniques for C++ docs with sphinx (like using `:cpp:class:` instead of `:ref:`).


docs/cpp/exceptions.rst
```diff
@@ -0,0 +1,20 @@
+Exceptions
+==========
+
+The CVC5 API communicates certain errors using exceptions.
+We broadly distinguish two types of exceptions: :cpp:class:`CVC4ApiException <cvc5::api::CVC4ApiException>` and :cpp:class:`CVC4ApiRecoverableException <cvc5::api::CVC4ApiRecoverableException>` (which is derived from :cpp:class:`CVC4ApiException <cvc5::api::CVC4ApiException>`).
+
+If any method fails with a :cpp:class:`CVC4ApiRecoverableException <cvc5::api::CVC4ApiRecoverableException>`, the solver behaves as if the failing method was not called and can safely be used.
```
aniemetz, line 7:
```suggestion
If any method fails with a :cpp:class:`CVC4ApiRecoverableException <cvc5::api::CVC4ApiRecoverableException>`, the solver behaves as if the failing method was not called. The solver can still be used safely.
```

aniemetz (commented):
Can you reformat the .rst files to 80 char?

Also, should we always document parameters? The Exception constructors have parameters that are yet undocumented.

nafur:
> Also, should we always document parameters? The Exception constructors have parameters that are yet undocumented.

I'm undecided on this. On the one hand, we should just document every argument. On the other hand, in many cases (like these ones) there is no room for confusion and you just have duplication: the text for the function and the text for the argument essentially say the same thing.

aniemetz:
As discussed offline, we decided to document all parameters with `@param` even if there is no room for confusion and it adds duplication.

nafur:
Added missing documentation.

aniemetz (approved):

PR #6341, opened 2021-04-13
Improve documentation of API kinds

nafur:
This PR improves the documentation of the `api::Kind` enum. Note that the docs for many of the enum values should still be improved. This PR merely makes sure that everything that is already there is actually output (`/*` vs `/**`) and properly rendered (missing spacing between lists, some formulas, etc).


docs/cpp/kind.rst
```diff
@@ -1,10 +1,16 @@
 Kind
 ====
 
+Every :cpp:class:`Term <cvc5::api::Term>` has a kind which represents its type, for example whether it is an equality (:cpp:enumerator:`EQUAL <cvc5::api::Kind::EQUAL>`), a conjunction (:cpp:enumerator:`AND <cvc5::api::Kind::AND>`), or a bitvector addition (:cpp:enumerator:`BITVECTOR_PLUS <cvc5::api::Kind::BITVECTOR_PLUS>`).
+
+Note that the api type :cpp:enum:`cvc5::api::Kind` roughly corresponds to :cpp:enum:`cvc5::Kind`, but is a different type and has a few subtle differences.
```
aniemetz, line 6:
```suggestion
Note that the API type :cpp:enum:`cvc5::api::Kind` roughly corresponds to :cpp:enum:`cvc5::Kind`, but is a different type and has a few subtle differences.
```

aniemetz: I'm not sure if we should even mention that. `cvc5::Kind` should not be accessible from outside at all.

src/api/cpp/cvc5_kind.h
```diff
@@ -78,187 +82,227 @@ enum CVC4_EXPORT Kind : int32_t
...
    * Permitted in bindings and in the lambda and quantifier bodies only.
+   * 
    * Parameters:
-   *   See mkVar().
+   *   - See @ref cvc5::api::Solver::mkVar() "mkVar()".
+   * 
    * Create with:
-   *   mkVar(const std::string& symbol, Sort sort)
-   *   mkVar(Sort sort)
+   *   - `cvc5::api::Solver::mkVar(Sort sort, const std::string& symbol)`
```
aniemetz, line 129: Why do we need prefix `cvc5::api` here but not above, for `mkConst` or `mkTerm`?

src/api/cpp/cvc5_kind.h
```diff
@@ -2845,7 +3405,7 @@ enum CVC4_EXPORT Kind : int32_t
 #endif
 
   /* ----------------------------------------------------------------------- */
-  /* marks the upper-bound of this enumeration */
+  /** marks the upper-bound of this enumeration */
```
aniemetz, line 3408:
```suggestion
  /** Marks the upper-bound of this enumeration. */
```

aniemetz (commented):
Again, 80 chars for .rst.

src/api/cpp/cvc5_kind.h
nafur: We don't need it.

docs/cpp/kind.rst
```diff
@@ -1,10 +1,22 @@
...
 ====
 
+Every :cpp:class:`Term <cvc5::api::Term>` has a kind which represents its type,
+for example whether it is an equality (:cpp:enumerator:`EQUAL
+<cvc5::api::Kind::EQUAL>`), a conjunction (:cpp:enumerator:`AND
+<cvc5::api::Kind::AND>`), or a bitvector addition
+(:cpp:enumerator:`BITVECTOR_PLUS <cvc5::api::Kind::BITVECTOR_PLUS>`).
+
+Note that the API type :cpp:enum:`cvc5::api::Kind` roughly corresponds to
+:cpp:enum:`cvc5::Kind`, but is a different type and has a few subtle
```
aniemetz, line 11:
```suggestion
:cpp:enum:`cvc5::Kind`, but is a different type. It hides internal kinds that
should not be exported to the API, and maps all kinds that we want to export
to its corresponding internal kinds.
```

docs/cpp/kind.rst
```diff
@@ -1,10 +1,22 @@
...
+differences.
```
aniemetz, line 12:
```suggestion
```

aniemetz (approved):

PR #6346, opened 2021-04-13
[proof-new] Fix explanation of literals in SAT proof manager

HanielB:
Prevents exponential behavior in SAT proof generation by not reexplaining previously explained literals. Also fix a potential issue in not previously overwriting rederived resolution chains during solving.


ajreynol (changes requested):
The benchmark in the PR is simply too big, I suggest omitting it.

HanielB:
I've removed the benchmark.

ajreynol (approved):

PR #6348, opened 2021-04-13
Refactor quantifiers macros

ajreynol:
This does some refactoring of quantifiers macros preprocessing pass to use up-to-date utility methods, including lambdas, substitutions, methods for getting free variables.

This is work towards adding proofs for macros.


yoni206 (approved):
Looks good and sounds cool. plz just verify clang-format before merging (there were some places where i was unsure).

PR #6349, opened 2021-04-13
Fix type rule for relations join image

ajreynol:
The join image type rule restricted that an argument was a constant. This is a logic restriction that should not be a part of the type checker.

This is required for not throwing type checking exceptions during proof conversion to LFSC.


mudathirmahgoub (approved):
LGTM. Please fix this compile [error](https://github.com/CVC4/CVC4/pull/6349/checks?check_run_id=2339548078#step:12:558).

src/theory/sets/theory_sets_private.cpp
```diff
@@ -1279,6 +1279,27 @@ void TheorySetsPrivate::preRegisterTerm(TNode node)
...
     break;
+    case kind::JOIN_IMAGE:
+    {
+      // these are logic exceptions, not type checking exceptions
+      if (node[1].getKind() != kind::CONST_RATIONAL)
+      {
+        throw LogicException(
+            "JoinImage cardinality constraint must be a constant");
+      }
+      cvc5::Rational r(INT_MAX);
```
mudathirmahgoub, line 1290: Please add `#include <climits>` to this file.

mudathirmahgoub (changes requested):

mudathirmahgoub (approved):

PR #6350, opened 2021-04-13
Add internal API methods for pool-based instantiation

ajreynol:


src/smt/smt_engine.h
```diff
@@ -496,6 +496,17 @@ class CVC4_EXPORT SmtEngine
 
   /*------------------------- end of sygus commands ------------------------*/
 
+  /**
+   * Declare pool whose initial value is the terms in initValue. A pool is
+   * a variable of type (Set T) that is used in quantifier annotations and does
+   * not occur in constraints.
+   *
+   * @param p The pool to declare, which should be a variable of type (Set T)
```
HanielB, line 504: where is this tested? I tried to find a check through the flow where `p` would be registered/used but did not find it. The documentation makes it seem like this is required, so maybe should be checked sooner rather than later?

ajreynol: Sure, I'll add.

HanielB (approved):
LGTM

PR #6351, opened 2021-04-13
Rename occurrences of CVC4 to CVC5.

aniemetz:
This renames everything but GitHub links and build system related
macros. Switching the build system to cvc5 will be the last step in the
renaming process.


src/util/sexpr.cpp
```diff
@@ -279,7 +279,7 @@ bool SExpr::languageQuotesKeywords(OutputLanguage language) {
       return true;
     case language::output::LANG_AST:
     case language::output::LANG_CVC3:
-    case language::output::LANG_CVC4:
+    case language::output::LANG_CVC:
```
mpreiner, line 282: Not sure if this should really be renamed. @ajreynol  what do you think?

aniemetz: Do we even still support LANG_CVC3 @ajreynol ?

ajreynol: I think LANG_CVC3 can be deleted, it doesn't appear to be used.

aniemetz: Resolved offline, yes we can rename and yes we can remove LANG_CVC3. Will do the latter in a separate PR.

mpreiner (approved):

PR #6354, opened 2021-04-14
Fix libpoly build and use new release

### LINKED ISSUE #4706: Use proper release for libpoly
#4679 integrates libpoly, and the corresponding contrib script uses the current master. As soon as there is a new release of libpoly, we should use a fixed release.

nafur:
This PR fixes the libpoly build: it naively removed the `test/` folder from the source directory to save on cache size.
It also uses the recently published 0.1.9 release of libpoly.
Fixes #4706.


HanielB (approved):
LGTM

PR #6355, opened 2021-04-14
Refactor cmake: auto-download and default-on dependencies

nafur:
This PR changes a few things in how dependencies are handled during configuration:
- `--x-dir` are removed for most dependencies, use the generic `--dep-path` instead
- the cmake `ENABLE_AUTO_DOWNLOAD` determines whether we attempt to download missing dependencies ourselves
- external projects check this option and send an error if it is OFF
- some optional dependencies are enabled by default (CaDiCaL, Poly, SymFPU)

This will essentially fail every call to `./configure.sh` until the user specifies `--auto-download`.


configure.sh
```diff
@@ -295,26 +289,11 @@ do
...
-    --kissat-dir) die "missing argument to $1 (try -h)" ;;
-    --kissat-dir=*) kissat_dir=${1##*=} ;;
-
-    --poly-dir) die "missing argument to $1 (try -h)" ;;
-    --poly-dir=*) poly_dir=${1##*=} ;;
-
-    --symfpu-dir) die "missing argument to $1 (try -h)" ;;
-    --symfpu-dir=*) symfpu_dir=${1##*=} ;;
+    --dep-path) die "missing argument to $1 (try -h)" ;;
+    --dep-path=*) dep_path=${1##*=} ;;
```
mpreiner, line 296: This should support specifying multiple paths `--dep-path=dep1 --dep-path=dep2 ...` since it's not guaranteed that all dependencies are installed in the same path.

cmake/deps-helper.cmake
```diff
@@ -39,6 +39,20 @@ if(CMAKE_VERSION VERSION_GREATER_EQUAL "3.14")
...
+            message(FATAL_ERROR "Could not find the required dependency ${name} \
+in the system. Please install it yourself or use --auto-download to let us \
+download and build it for you.")
+        else()
+            message(FATAL_ERROR "Could not find the optional dependency ${name} \
+in the system. You can disable this dependency with ${disable_option}, install \
+it yourself or use --auto-download to let us download and build it for you.")
+        endif()
+    endif()
+endmacro(check_auto_download)
```
mpreiner, line 54: Should this specify the version that needs to be installed by the user?

nafur: Good point. I'll try to print the required version, if there is one.

mpreiner (approved):
LGTM

PR #6356, opened 2021-04-14
[unsat-cores] Improving new unsat cores

HanielB:
This commit adds a new option to produce unsat cores based on our proof infrastructure (whereas previously we could only do so if we were also checking unsat cores) and the corresponding changes to the default settings to account for it. Since now `options::unsatCores()` and `options::produceProofs()` are incompatible, several parts of the code where we tested if we were in "old unsat cores mode", by testing the former and the negation of the latter options, are updated accordingly.

This commit also changes how SMT engine sets things by disabling proofs in the theory engine if we are in unsat core mode.


src/options/smt_options.toml
```diff
@@ -207,6 +207,13 @@ header = "options/smt_options.h"
   type       = "bool"
   help       = "after UNSAT/VALID, produce and check an unsat core (expensive)"
 
+[[option]]
```
ajreynol, line 210: Can you give a quick comment to say this is temporary?

Also, can you just for sanity's sake say how to enable old vs. new unsat cores, as well as what the values of unsatCores, unsatCoresNew, produceProofs are intended to be after set_defaults?

HanielB: haha yes, this is important indeed. Where would be the best place to put this?

src/prop/theory_proxy.cpp
```diff
@@ -96,9 +96,38 @@ void TheoryProxy::explainPropagation(SatLiteral l, SatClause& explanation) {
 
   theory::TrustNode tte = d_theoryEngine->getExplanation(lNode);
   Node theoryExplanation = tte.getNode();
-  if (cvc5::options::produceProofs())
+  if (options::produceProofs())
   {
-    d_propEngine->getProofCnfStream()->convertPropagation(tte);
+    // no proof tracking for propagations if we're producing unsat cores
+    if (!options::unsatCoresNew())
```
ajreynol, line 102: I think this logic should just be put inside of `d_propEngine->getProofCnfStream()->convertPropagation`.  Is there a reason to have this block of code here?

ajreynol (changes requested):

src/prop/theory_proxy.cpp
HanielB: I think that wouldn't work because the trust node, when we don't have proofs in the theory engine, does not have the explanation. It just contains the propagated node.

ajreynol: Hm, maybe I'd suggest a new method ::convertPropagationTrusted then.

HanielB: I see. Sounds good, will add.

src/smt/proof_manager.h
```diff
@@ -38,6 +38,25 @@ class ProofPostproccess;
...
+ * - If we are producing unast cores using this proof infrastructure, then the
+ *   SmtEngine will have proofs using this proof manager (if --produce-proofs
+ *   was not passed by the user it will be activiated), but these proofs will
+ *   only cover preprocessing and the prop engine, i.e., the theory engine will
+ *   not have proofs.
+ *
+ * - If we are not producing unsat cores then the SmtEngine will have proofs as
+ *   long as --produce-proofs was given.
+ *
+ * - If SmtEngine has been configured in a way that is incompatible with unsat
```
ajreynol, line 57: I don't think this is correct.  Proofs are only disabled if there is an option that is incompatible with *proofs* not *unsat cores*?

HanielB: You're right, it's the other way around: if we are incompatible with proofs the the new unsat cores will be disabled

src/smt/smt_engine.cpp
```diff
@@ -231,6 +231,12 @@ void SmtEngine::finishInit()
     d_asserts->setProofGenerator(pppg);
     // enable it in the SmtSolver
     d_smtSolver->setProofNodeManager(pnm);
+    // if proofs and unsat cores, proofs are used solely for unsat core
+    // production, so we don't generate proofs in the theory engine
+    if (options::unsatCoresNew())
```
ajreynol, line 236: I think we should make this an argument to SmtSolver finishInit and not have a separate `set` method.

ajreynol (changes requested):
Another comment: we probably can't do dump-proof / check-proofs when unsat cores are enabled, right?  We should disable those options, as well as ensuring that SmtEngine::checkProofs is guarded by checkProofs option being enabled.

src/smt/set_defaults.cpp
```diff
@@ -278,11 +293,18 @@ void setDefaults(LogicInfo& logic, bool isInternalSubsolver)
...
+  if (options::unsatCores())
   {
     disableProofs = true;
   }
 
+  // new unsat core specific restrictions for proofs
+  if (options::unsatCoresNew())
+  {
+    // no fine-graininess
+    options::proofGranularityMode.set(options::ProofGranularityMode::OFF);
```
ajreynol, line 305: I would only overwrite this if it was not set by user.

src/options/smt_options.toml
```diff
@@ -207,12 +207,21 @@ header = "options/smt_options.h"
   type       = "bool"
   help       = "after UNSAT/VALID, produce and check an unsat core (expensive)"
 
+# Temporary option until old unsat cores are removed and this becomes the standard
+[[option]]
+  name       = "unsatCoresNew"
+  category   = "regular"
+  long       = "produce-unsat-cores-new"
+  type       = "bool"
+  help       = "turn on unsat core generation using the new proof infrastructure. Note this option will activate the --produce-proofs option and that it is incompatible with the regular unsat cores."
```
ajreynol, line 216: I would omit saying this activates the `--produce-proofs` option since it (should not) be equivalent to enabling it on the command line.  This is just an internal detail.

src/options/smt_options.toml
```diff
@@ -207,12 +207,21 @@ header = "options/smt_options.h"
...
+
+# Temporary option until old unsat cores are removed and this becomes the standard
 [[option]]
   name       = "checkUnsatCoresNew"
   category   = "regular"
   long       = "check-unsat-cores-new"
   type       = "bool"
-  help       = "after UNSAT/VALID, produce and check an unsat core (expensive) using the new proof infrastructure"
+  help       = "after UNSAT/VALID, produce and check an unsat core (expensive) using the new proof infrastructure. Note this option will activate the --produce-proofs option and that it is incompatible with the regular unsat cores."
```
ajreynol, line 224: Ditto.

src/smt/smt_engine.cpp
```diff
@@ -1625,13 +1632,16 @@ void SmtEngine::getInstantiationTermVectors(
 {
   SmtScope smts(this);
   finishInit();
-  if (options::produceProofs() && getSmtMode() == SmtMode::UNSAT)
+  if (options::produceProofs() && !options::unsatCoresNew()
+      && getSmtMode() == SmtMode::UNSAT)
   {
+    Trace("test") << "test1\n";
```
ajreynol, line 1638: Remove

src/smt/smt_engine.cpp
```diff
@@ -1625,13 +1632,16 @@ void SmtEngine::getInstantiationTermVectors(
...
     // minimize instantiations based on proof manager
     getRelevantInstantiationTermVectors(insts);
   }
   else
   {
+    Trace("test") << "test2\n";
```
ajreynol, line 1644: Ditto

src/smt/smt_solver.h
```diff
@@ -72,7 +72,8 @@ class SmtSolver
   /**
    * Create theory engine, prop engine based on the logic info.
    */
-  void finishInit(const LogicInfo& logicInfo);
+  void finishInit(const LogicInfo& logicInfo,
```
ajreynol, line 75: Document the argument.

src/smt/proof_manager.h
```diff
@@ -38,6 +38,25 @@ class ProofPostproccess;
...
+ *   was not passed by the user it will be activated), but these proofs will
+ *   only cover preprocessing and the prop engine, i.e., the theory engine will
+ *   not have proofs.
+ *
+ * - If we are not producing unsat cores then the SmtEngine will have proofs as
+ *   long as --produce-proofs was given.
+ *
+ * - If SmtEngine has been configured in a way that is incompatible with proofs
+ *   and it was set to produce unsat cores using proofs, then unsat core
+ *   production will be disabled.
```
ajreynol, line 59:
```suggestion
 *   then unsat core production will be disabled.
```

ajreynol (changes requested):

src/prop/proof_cnf_stream.cpp
```diff
@@ -571,6 +571,32 @@ void ProofCnfStream::convertPropagation(theory::TrustNode trn)
   d_psb.clear();
 }
 
+void ProofCnfStream::convertPropagationTrusted(Node exp, Node lit)
+{
+  // we need however to register the assumption as this might break the
+  // expansion when finalizing the proof (for trying to explain as a literal
+  // what is actually a clause). Since we have no proofs in theory engine,
+  // we cannot go via tte.getProven(). The explanation is decomposing
```
ajreynol, line 579: Actually I'm confused, tte.getProven() is fine to call for any non-null TrustNode, regardless of whether proofs are enabled.

ajreynol (changes requested):

src/prop/proof_cnf_stream.cpp
HanielB: the issue was not that I couldn't call it, but that `getProven` would give the same result as `getNode` when we are not producing proofs in the SMT engine.

ajreynol: I think this would suggest there is a bug.  What TrustNodeKind are the TrustNode in question?

HanielB: I didn't check the invariants to see if it was a bug of expected behavior. Maybe we have to change [this line](https://github.com/CVC4/CVC4/blob/e6168da0663c73f17994b50f62a11799619b67ba/src/theory/theory_engine.cpp#L1775) to generate a trust node for prop explain?

ajreynol: Yes, the line you reference looks wrong to me.

HanielB: OK, I'll take a look.

ajreynol (approved):

PR #6358, opened 2021-04-14
Add instantiation pool feature to the API

ajreynol:
This adds the command `declare-pool` to the public API. It also adds parsing support for this feature, lifts the internal kinds to public kinds, adds an example regression, and a unit test for the new declare-pool solver method.


src/api/cpp/cvc5.cpp
```diff
@@ -6590,6 +6596,24 @@ Term Solver::getSeparationNilTerm() const
   CVC5_API_TRY_CATCH_END;
 }
 
+Term Solver::declarePool(const std::string& symbol,
```
aniemetz, line 6599: Is `symbol` allowed to be empty? If not, this should be checked.

ajreynol: This should be the same policy as `declare-fun`, which I suppose allows the symbol to be emtpy (`||` in SMT-LIB).
Thus, no check is needed.

aniemetz (commented):
Can you also add a python test for this?

ajreynol:
@aniemetz  Yoni and Ying are working on the python tests.  We don't yet have an analog for solver_black in python, so I'd suggest we commit this so they can do the conversion to python tests.

What do you think @yoni206 ?

yoni206:
SGTM. In case all agree, @Ying1123 , please make sure you pull from master so that your python solver test is up to date with the cpp solver test.

aniemetz:
SGTM.

aniemetz (approved):

PR #6360, opened 2021-04-14
BV: Move implementation of type rules from header to .cpp.

aniemetz:


src/theory/bv/theory_bv_type_rules.cpp
```diff
@@ -0,0 +1,351 @@
...
+    TypeNode t = (*it).getType(check);
+    // NOTE: We're throwing a type-checking exception here even
+    // when check is false, bc if we don't check that the arguments
+    // are bit-vectors the result type will be inaccurate
+    if (!t.isBitVector())
+    {
+      throw TypeCheckingExceptionPrivate(n, "expecting bit-vector terms");
+    }
+    size += t.getBitVectorSize();
+  }
```
mpreiner, line 150: Maybe use foreach loop.

mpreiner (approved):

PR #6361, opened 2021-04-14
python API sorts: adding functions and tests

yoni206:
This PR does the following:
1. removes old python sort API test
2. creates a new python sort API test, obtained by translating the (entire) cpp unit test https://github.com/CVC4/CVC4/blob/master/test/unit/api/sort_black.cpp
3. adds support for bags and datatype selectors/testers domain/codomain in the python API.


test/python/unit/api/test_sort.py
```diff
@@ -0,0 +1,714 @@
...
+
+def test_operators_comparison(solver):
+    try:
+        solver.getIntegerSort() == Sort(solver)
+        solver.getIntegerSort() != Sort(solver)
+        solver.getIntegerSort() < Sort(solver)
+        solver.getIntegerSort() <= Sort(solver)
+        solver.getIntegerSort() > Sort(solver)
+        solver.getIntegerSort() >= Sort(solver)
+    except RuntimeError:
```
makaimann, line 61: Quick question: is there a reason to use the `try` `except` as opposed to just asserting each of those conditions individually? It should fail if any of them throw an exception, right? Is it because you want to abort the whole test?

test/python/unit/api/test_sort.py
```diff
@@ -0,0 +1,714 @@
...
+        pytest.fail()
+
+
+def test_is_boolean(solver):
+    assert (solver.getBooleanSort().isBoolean())
```
makaimann, line 66:
```suggestion
    assert solver.getBooleanSort().isBoolean()
```

Don't need the parentheses for an assert.

test/python/unit/api/test_sort.py
```diff
@@ -0,0 +1,714 @@
...
+    try:
+        Sort(solver).isBoolean()
+    except RuntimeError:
+        pytest.fail()
+
+
+def test_is_integer(solver):
+    assert (solver.getIntegerSort().isInteger())
```
makaimann, line 74: same comment about parentheses for the assert argument (won't comment on all of them).

test/python/unit/api/test_sort.py
```diff
@@ -0,0 +1,714 @@
...
+    assert (not solver.getRealSort().isInteger())
+    try:
```
makaimann, line 76: ditto on the `try`-`except`, won't comment on all of them.

makaimann (changes requested):
Thanks for the PR! Have a couple minor requests. The comments about the `assert` argument are actually a request for a change. It might be fine with one argument, but it's better to not use the parentheses because if you ever do something like `assert (arg1, arg2)` then it's passing a non-empty tuple which will always evaluate to `True`. The other comments are just questions.

test/python/unit/api/test_sort.py
yoni206: I have two weak reasons:
1. I want to distinguish between `RuntimeError` (which is thrown from the API if something bad happens) and other errors like syntax errors, typos, etc. 
2. I want to be similar to the cpp api tests

If you think this is unnecessary, I have no problem removing all these `try`s. 

BTW, @Ying1123  is preparing a PR for the solver_black tests in which she has implemented a `not_throws` function:
https://github.com/Ying1123/CVC4/blob/3770e0edcd93e4659867d5dcba7557d3f5e8d3b4/test/python/unit/api/test_solver.py#L19
used like this:
https://github.com/Ying1123/CVC4/blob/3770e0edcd93e4659867d5dcba7557d3f5e8d3b4/test/python/unit/api/test_solver.py#L44

having this would make us even more similar to the cpp api tests. If we decide to go with these checks, I can either bring her function to this PR, or refactor the tests after her code gets merged. What do u think?

test/python/unit/api/test_sort.py
yoni206: Fixed in 5daa7a5 for all asserts.

test/python/unit/api/test_sort.py
yoni206: dito

test/python/unit/api/test_sort.py
yoni206: yup lets discuss this, see my response above.

test/python/unit/api/test_sort.py
makaimann: Cool, thanks for the explanation! I don't feel super strongly either way. I don't think matching it super closely to C++ is necessary, although there is value in that. There's a bit of tension there between having very similar APIs and allowing Pythonic use. As for `assert_not_throws`, I think it could make sense to have a function like that. I don't quite follow what the `ExpectedException` argument is for though. Because if it's not supposed to throw, why would there be an expected exception?

yoni206: this is again the distinction between `RuntimeError` and other exceptions. `RuntimeError` will go to the `pytest.fail()` line while other exceptions will fail in the failed line, which will create a difference.

It would be simpler, of course, to just write the commands, and if they fail then they fail. The code will be more readable, maybe even `pythonic` (whatever that means :) ) but just different then cpp, that i see you don't think is very important. @Ying1123 was also advocating removing these `try` statements. So why don't we just drop them?

yoni206: dropped here: 29a0821
Would be interested in knowing what u think. Of course, this commit can be reverted.

makaimann: Yeah I think that looks good! I'd vote to stick with that but as mentioned earlier don't mind if you'd rather have some extra error infrastructure like the `assert_not_throws`.

makaimann (approved):
LGTM, thanks for the changes!

PR #6362, opened 2021-04-15
Fix printing of stats when aborted.

nafur:
This PR improves/fixes printing of statistics when the solver has been aborted, i.e. when we use `printSafe()`, and a few other minor issues with the new statistics setup.
- add `toString()` methods for `TypeConstant`, `api::Kind` to allow for automatic printing by `print_safe<>()`
- improve `kindToString()` to avoid `std::stringstream`
- fix newlines between statistics in `printSafe()`
- make printing of histograms consistent
- make `--stats-all`, `--stats-expert` and `--stats-every-check` automatically enable `--stats` (and vice versa)

### DIFF
--- a/src/options/options_handler.cpp
+++ b/src/options/options_handler.cpp
@@ -271,22 +271,22 @@ void OptionsHandler::setStats(const std::string& option, bool value)
 #endif /* CVC5_STATISTICS_ON */
   if (value)
   {
-    if (option == "stats-all")
+    if (option == options::statisticsAll.getName())
     {
       d_options->d_holder->statistics = true;
     }
-    else if (option == "stats-every-query")
+    else if (option == options::statisticsEveryQuery.getName())
     {
       d_options->d_holder->statistics = true;
     }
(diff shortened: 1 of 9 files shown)


src/options/options_handler.cpp
```diff
@@ -259,15 +259,40 @@ void OptionsHandler::setProduceAssertions(std::string option, bool value)
...
+  }
+  else
+  {
+    if (option == "stats")
+    {
+      d_options->d_holder->statisticsAll = false;
+      d_options->d_holder->statisticsEveryQuery = false;
+      d_options->d_holder->statisticsExpert = false;
+    }
+  }
```
mpreiner, line 295: I'd not compare against the long option names here since these strings may change. Instead use the `getName()` method of the corresponding option type:
e.g. 
```option == options::statisticsAll.getName()```

nafur: Fixed.

ajreynol (approved):

PR #6363, opened 2021-04-15
Replace SExpr class by simpler conversion routines

nafur:
This PR finally removes the `SExpr` class. SMT-LIB compatible output is retained by using new on-the-fly conversion to s-expression strings. This finally allows us to remove includes to integer and rational from `smt_engine.h`.
In detail:
- a new set of `toSExpr()` methods is implemented that converts certain types to s-expression strings (without an intermediate class representing s-expressions)
- `SmtEngine::getInfo()` returns a string instead of SExpr and uses the new `toSExpr` methods
- `SmtEngine::getStatistic()` is removed
- `SExpr` class is removed
- `d_commandVerbosity` uses `int` instead of `Integer`


src/util/sexpr.h
```diff
@@ -10,296 +10,131 @@
...
  */
-std::ostream& operator<<(std::ostream& out, PrettySExprs ps);
+template <typename T>
+std::string toSExpr(const T& t)
+{
+  std::stringstream ss;
+  toSExpr(ss, t);
+  return ss.str();
+}
+/**
```
abdoo8080, line 128: Nit: Add a new line.

abdoo8080 (approved):
LGTM! Thanks for the PR.

src/util/sexpr.h
nafur: I'm not sure what you mean. Would you add a newline to `ss`? (I'd not do that, as the caller of this function should decide about the context of this s-expression. The s-expression itself has no trailing newline, IMHO).

abdoo8080: I meant to add a new line between lines 127 and 128 of the code...

nafur: Ah, yes. Changed it.

PR #6364, opened 2021-04-15
Add API documentation for statistics

nafur:
This PR adds documentation for `api::Statistics` and `api::Stat`, as well as further explanations in sphinx.
It also adds a custom css to our sphinx theme that slightly changes how inline code blocks look.


src/api/cpp/cvc5.h
```diff
@@ -2337,10 +2337,12 @@ struct CVC4_EXPORT RoundingModeHashFunction
 
 /**
  * Represents a snapshot of a single statistic value.
- * A value can be of type int64_t, double, std::string or a histogram
+ * A value can be of type `int64_t`, `double`, `std::string` or a histogram
  * (`std::map<std::string, uint64_t>`).
- * The value type can be queried (using `isInt`, `isString`, etc.) and
- * the stored value can be accessed (using `getInt`, `getString`, etc.).
+ * The value type can be queried (using `isInt()`, `isDouble()`, etc.) and
```
aniemetz, line 2342: Shouldn't the references properly link to the functions?

nafur: They do. Note that this is in the header and links are created by doxygen. Doxygen will add links for whatever "sufficiently looks like a function" and it is "sufficiently clear" what the target should be: https://www.doxygen.nl/manual/autolink.html
Here, `isInt()` is enough as the doc is in the context of a class which has a member function of that name...

aniemetz: Afaik you don't have to enclose it in backticks, adding `()` should be enough.

nafur: This is true. I like that code is rendered as code in monospace, without backticks it is rendered as regular text, IIRC.

aniemetz: In your image above it's not, it is regular text in blue and bold (for the links). I'd suggest to not use backticks if we reference a function, but only if it doesn't link anywhere, e.g., `std::string`. This will improve readability in the code itself.

nafur: Just to make sure we are talking about the same thing: we discuss how, e.g., `isInt()` is and should be rendered, right?
Right now, `isInt()` is rendered as a link in a code block (monospace, bold, in a box). Without the backticks, it is only a link.

Comparison: image vs. image

This ultimately is a matter of style / personal preference. I like that code (in a broad sense, including function names) are rendered as code, and I don't think that readability in the source files suffers from a few backticks. We should find a common style, though...

HanielB: I prefer it with the backticks.

mpreiner: Hmm it really depends, if you have lots of these code blocks in a paragraph it can get pretty cluttered with code blocks. I'd use the code blocks only for bigger code blocks. The functions are highlighted anyways due to autolink.

nafur: This PR now includes a custom css file that changes how the code looks like.
Example:
image

aniemetz: Looks good to me. Just for the record, this was the result of a lengthy offline discussion, where we decided, we don't want inline code to be rendered in boxes but use monospace font.

@nafur can you double check that (non-inline) code blocks are still rendered correctly (in a box)?

aniemetz: Checked offline, everything looks fine, code blocks are correctly rendered.

aniemetz (commented):
One minor request, can you add to the css that code block links are still rendered in blue?

aniemetz (approved):

PR #6365, opened 2021-04-15
Split C++ API docs from general docs

nafur:
This PR moves everything related to the C++ API docs into `docs/cpp` to clearly separate the generic setup for the sphinx documentation from everything related to the C++ API.


aniemetz (commented):
cpp.rst was on the top-level on purpose, more like, the entry into the CPP docs. The index of the python docs would also be on the top-level. I don't really see a reason why this should be in `cpp`, I'd rather have the content files there, but the index file on the top level. @mpreiner opinions?

nafur:
I just found "everything in a subfolder" an easier rule compared to "everything but the index file". But I don't really care, as long as the cmake stuff is separated.

aniemetz (approved):
I guess I can live with having it moved to `cpp` :)

PR #6366, opened 2021-04-15
Clean up interaction of resource manager and its options

nafur:
This PR simplifies how the resource manager interacts with the options. Instead of using some notification mechanism, the resource manager simply retrieves the options via `options::xyz()`. This simplifies the options handler, the resource manager interface and the options.
When instructed to do so by the API, the `SmtEngine` now overwrites the respective option instead of calling out to the resource manager.


src/util/resource_manager.cpp
```diff
@@ -222,8 +192,8 @@ uint64_t ResourceManager::getTimeUsage() const { return d_cumulativeTimeUsed; }
 
 uint64_t ResourceManager::getResourceRemaining() const
 {
-  if (d_resourceBudgetCumulative <= d_cumulativeResourceUsed) return 0;
-  return d_resourceBudgetCumulative - d_cumulativeResourceUsed;
+  if (options::cumulativeResourceLimit() <= d_cumulativeResourceUsed) return 0;
```
ajreynol, line 195: Instead of calling options:: here, can we already do this properly by caching the reference to options passed to this manager and looking it up on that object?

ajreynol: Same goes for the rest of the occurrences of `options::` in this file.

ajreynol (changes requested):

src/util/resource_manager.cpp
nafur: As per offline discussion: before doing this, we want to do some major refactoring of the options class. I'd just leave it as it is for now.

ajreynol (approved):

PR #6368, opened 2021-04-15
Build support library from base and context.

mpreiner:
This PR creates a support library from the utilities in base and context, which will be required in the parser as soon as we move the symbol table/manager to the parser.

Note: I decided to always build static libraries from base and context (and optionally enable -fPIC for shared builds) since I'm not sure if we want to have these libraries installed separately. Right now these are considered as cvc5 internal utilities that can be used in all cvc5 libraries, but not outside.


src/base/CMakeLists.txt
```diff
@@ -75,3 +78,14 @@ set_source_files_properties(
...
+
+add_library(cvc4base STATIC ${LIBBASE_SOURCES})
+if(ENABLE_SHARED)
+  set_target_properties(cvc4base PROPERTIES POSITION_INDEPENDENT_CODE ON)
+endif()
+target_compile_definitions(cvc4base
+  PRIVATE
+    -D__BUILDING_CVC4LIB
+    -D__STDC_LIMIT_MACROS
+    -D__STDC_FORMAT_MACROS
```
nafur, line 90: What do these do?

mpreiner: That's what I've been wondering too. It is an old relic from the old build infrastructure, maybe we can get rid of it.

mpreiner: I removed it.

nafur (approved):
LGTM.

PR #6373, opened 2021-04-16
BV: Add proof logging for bit-blasting. 

aniemetz:
Note: This includes #6372.


src/expr/proof_rule.h
```diff
@@ -727,6 +727,47 @@ enum class PfRule : uint32_t
   // ---------------------
   // Conclusion: (= t bitblast(t))
   BV_BITBLAST,
```
HanielB, line 729: We can delete this now?

src/expr/proof_rule.h
```diff
@@ -727,6 +727,47 @@ enum class PfRule : uint32_t
...
+  BV_BITBLAST_ASHR,
+  BV_BITBLAST_ULTBV,
+  BV_BITBLAST_SLTBV,
+  BV_BITBLAST_ITE,
+  BV_BITBLAST_EXTRACT,
+  BV_BITBLAST_REPEAT,
+  BV_BITBLAST_ZERO_EXTEND,
+  BV_BITBLAST_SIGN_EXTEND,
+  BV_BITBLAST_ROTATE_RIGHT,
+  BV_BITBLAST_ROTATE_LEFT,
```
HanielB, line 770: These should all be documented afterwards.

src/theory/bv/bv_solver_simple.cpp
```diff
@@ -69,9 +70,16 @@ BVSolverSimple::BVSolverSimple(TheoryState* s,
...
                                ProofNodeManager* pnm)
     : BVSolver(*s, inferMgr),
-      d_bitblaster(new BBProof(s)),
-      d_epg(pnm ? new EagerProofGenerator(pnm, s->getUserContext(), "")
-                : nullptr)
+      d_tcpg(
+          pnm ? new TConvProofGenerator(pnm,
+                                        nullptr,
+                                        TConvPolicy::ONCE,
+                                        TConvCachePolicy::NEVER,
```
HanielB, line 77: It'd be good to document these

src/expr/proof_rule.h
ajreynol: To keep things concise, I'd prefer giving only one detailed documentation and then (at most) one line per kind here.

See for example: https://github.com/CVC4/CVC4/blob/master/src/expr/proof_rule.h#L217.

HanielB: That sounds good indeed.

aniemetz: Ah right, that was on my mental TODO on Thursday but then my brain went into vacation mode :D

src/theory/bv/bv_solver_simple.cpp
aniemetz: Absolutely! Suggestions?

src/expr/proof_rule.h
aniemetz: It looks like it. @mpreiner and I weren't sure initially if we wouldn't need it for something.

mpreiner: Shouldn't be needed anymore since we now have the fine-grained rules.

src/theory/bv/bv_solver_simple.cpp
```diff
@@ -69,9 +70,16 @@ BVSolverSimple::BVSolverSimple(TheoryState* s,
                                TheoryInferenceManager& inferMgr,
                                ProofNodeManager* pnm)
     : BVSolver(*s, inferMgr),
-      d_bitblaster(new BBProof(s)),
-      d_epg(pnm ? new EagerProofGenerator(pnm, s->getUserContext(), "")
-                : nullptr)
+      d_tcpg(
+          pnm ? new TConvProofGenerator(pnm,
```
ajreynol, line 74: Can maybe use TConvCachePolicy::STATIC here.

src/theory/bv/bv_solver_simple.cpp
HanielB: `TConvPolicy::ONCE` is justified by a converted term being looked at again only once (i.e., after the children are rewritten, the post-visited term may be rewritten again). Since terms are rewritten to terms that may contain themselves, using the FIXPOINT policy could lead to infinite loops.

For the caching policy it's what Andy mentioned on slakc.

src/expr/proof_rule.h
aniemetz: We discussed this offline, will keep it for coarse proofs.

src/expr/proof_rule.h
```diff
@@ -707,6 +707,58 @@ enum class PfRule : uint32_t
...
   BV_BITBLAST,
+  // ======== Bitblast Bit-Vector Constant
+  // Children: none
+  // Arguments: (= t bitblast(t))
+  BV_BITBLAST_CONST,
+  // ======== Bitblast Bit-Vector Variable
+  // Children: none
+  // Arguments: (= t bitblast(t))
+  BV_BITBLAST_VAR,
+  // ======== Bitblast Bit-Vector Terms
```
ajreynol, line 718: Can you add a TODO to change the form of these to only take the LHS of the equality and to make the proof checker compute the conclusion using the bitblast util?

ajreynol (changes requested):

src/expr/proof_rule.h
```diff
@@ -707,6 +707,58 @@ enum class PfRule : uint32_t
   // ---------------------
   // Conclusion: (= t bitblast(t))
   BV_BITBLAST,
+  // ======== Bitblast Bit-Vector Constant
+  // Children: none
+  // Arguments: (= t bitblast(t))
```
ajreynol, line 712: The conclusion is the argument here?

src/expr/proof_rule.h
```diff
@@ -707,6 +707,58 @@ enum class PfRule : uint32_t
...
+  BV_BITBLAST_CONST,
+  // ======== Bitblast Bit-Vector Variable
+  // Children: none
+  // Arguments: (= t bitblast(t))
```
ajreynol, line 716: Ditto, specify conclusion.

src/expr/proof_rule.h
aniemetz: `(= t bitblast(t))`

src/expr/proof_rule.h
```diff
@@ -701,12 +701,72 @@ enum class PfRule : uint32_t
   ARRAYS_TRUST,
 
   //================================================= Bit-Vector rules
+  // Note: bitblast() represents the result of the bit-blasted term as a
+  //       bit-vector consisting of the output bits of the bit-blasted circuit
+  //       representation of the term
```
HanielB, line 706: Reference the method in the code that has the semantics of this function.

src/theory/bv/bitblast/proof_bitblaster.cpp
```diff
@@ -16,19 +16,79 @@
...
+
+BBProof::~BBProof() {}
 
 void BBProof::bbAtom(TNode node)
 {
   std::vector<TNode> visit;
   visit.push_back(node);
   std::unordered_set<TNode, TNodeHashFunction> visited;
+
+  bool fine_proofs =
```
HanielB, line 86: caml case?

src/theory/bv/bitblast/proof_bitblaster.cpp
```diff
@@ -16,19 +16,79 @@
...
+      options::proofGranularityMode() != options::ProofGranularityMode::OFF;
+  bool fine_proofs_enabled = isProofsEnabled() && fine_proofs;
```
HanielB, line 88: ditto

HanielB (approved):
LGTM, just minor comments

src/expr/proof_rule.h
aniemetz: There is no code, just a kind.

src/theory/bv/bitblast/proof_bitblaster.cpp
aniemetz: We only require camelCase for class/struct members.

ajreynol (approved):

PR #6374, opened 2021-04-16
cmake: Build object libraries for base and context.

mpreiner:
cmake complains that the static base and context libraries are not
exported as install targets. Since we do not want to install these
libraries we'll build object libraries instead.


nafur (approved):
LGTM. Why didn't this come up in the CI install check?

PR #6375, opened 2021-04-16
Add guards to disable clang-format around placeholders in templates.

aniemetz:


src/options/options_template.cpp
```diff
@@ -343,10 +350,11 @@ void Options::printLanguageHelp(std::ostream& out) {
  * If you add something that has a short option equivalent, you should
  * add it to the getopt_long() call in parseOptions().
  */
+// clang-format off
 static struct option cmdlineOptions[] = {
   ${cmdline_options}$
-  { NULL, no_argument, NULL, '\0' }
-};/* cmdlineOptions */
+  {NULL, no_argument, NULL, '\0'}};
```
nafur, line 356:
```suggestion
  {nullptr, no_argument, nullptr, '\0'}};
```

nafur, line 356:
```suggestion
  {nullptr, no_argument, nullptr, '\0'}};
```

src/options/options_template.cpp
```diff
@@ -537,10 +548,11 @@ std::string Options::suggestCommandLineOptions(const std::string& optionName)
   return didYouMean.getMatchAsString(optionName.substr(0, optionName.find('=')));
 }
 
+// clang-format off
 static const char* smtOptions[] = {
   ${options_smt}$
-  NULL
-};/* smtOptions[] */
+  NULL};
```
nafur, line 554:
```suggestion
  nullptr};
```

nafur (commented):
Just two minor comments. I have a general question about this, though: why do we need the guards at all? Do we care about clang-format not changing formatting in the generated files?

nafur (approved):
I see. LGTM.

PR #6376, opened 2021-04-16
Remove linking against gmp and cln in tests and parser

nafur:
Finally, we no longer need to link against GMP and CLN for the parser and the tests.
To actually achieve this, this PR also removes some dead code and unused includes from some parser files.


ajreynol (approved):
LGTM

PR #6378, opened 2021-04-16
Fix dependencies for stats options

nafur:
A last-minute edit in a previous PR broke the handling of dependencies between the statistic options.

### DIFF
--- a/src/options/options_handler.cpp
+++ b/src/options/options_handler.cpp
@@ -275,25 +275,26 @@ void OptionsHandler::setStats(const std::string& option, bool value)
     throw OptionException(ss.str());
   }
 #endif /* CVC5_STATISTICS_ON */
+  Assert(option.substr(0, 2) == "--");
+  std::string opt = option.substr(2);
   if (value)
   {
-    if (option == std::string("--") + options::statisticsAll.getName())
+    if (opt == options::statisticsAll.getName())
     {
       d_options->d_holder->statistics = true;
     }
-    else if (option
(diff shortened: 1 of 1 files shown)


src/options/options_handler.cpp
```diff
@@ -277,22 +277,23 @@ void OptionsHandler::setStats(const std::string& option, bool value)
 #endif /* CVC5_STATISTICS_ON */
   if (value)
   {
-    if (option == options::statisticsAll.getName())
+    if (option == std::string("--") + options::statisticsAll.getName())
```
ajreynol, line 280: Nit: this is a bit ugly, can we do option.substr(2) on the left hand side instead? And moreover store this a temp variable above?  I'm actually surprised this method isn't stripping this earlier.

ajreynol (changes requested):

src/options/options_handler.cpp
nafur: I'd rather not do the `substr(2)`. It is not more efficient (both variants create a temporary), but it may do weird stuff if `option` for some reason does not start with `--`. I felt this version is clearer and more robust.

As for the temp variable: what exactly do you want to put in a temp variable?

ajreynol: I'm saying that this method should ideally be refactored later so that option contains `"stats"` only.
The temporary would ensure this is stored at the beginning of this method (or at the beginning of the block where it is used).

nafur: Ah, I see.

ajreynol (approved):

PR #6379, opened 2021-04-16
Make trust substitution map generate proofs lazily

ajreynol:
This is work towards addressing a bottleneck when generating proofs for substitutions from non-clausal simplification.

This makes the proof generation in trust substitution map lazy.

Further work is required to allow an alternative fixpoint semantics for substitutions in the proof checker instead of the current sequential one.


src/theory/trust_substitutions.cpp
```diff
@@ -215,35 +227,34 @@ TrustNode TrustSubstitutionMap::apply(Node n, bool doRewrite)
...
-  }
+  Assert(index <= d_tsubs.size());
   std::vector<Node> csubsChildren;
-  for (const TrustNode& tns : d_tsubs)
+  for (size_t i = 0; i < index; i++)
   {
-    csubsChildren.push_back(tns.getProven());
+    csubsChildren.push_back(d_tsubs[i].getProven());
   }
   std::reverse(csubsChildren.begin(),csubsChildren.end());
```
HanielB, line 251: format

src/theory/trust_substitutions.cpp
```diff
@@ -215,35 +227,34 @@ TrustNode TrustSubstitutionMap::apply(Node n, bool doRewrite)
...
-  d_currentSubs = NodeManager::currentNM()->mkAnd(csubsChildren);
-  if (d_currentSubs.get().getKind() == kind::AND)
+  Node cs = NodeManager::currentNM()->mkAnd(csubsChildren);
+  if (cs.getKind() == kind::AND)
   {
-    d_subsPg->addStep(d_currentSubs, PfRule::AND_INTRO, csubsChildren, {});
+    d_subsPg->addStep(cs, PfRule::AND_INTRO, csubsChildren, {});
   }
```
HanielB, line 256: Since you have the size you could compute this directly rather than check kinds and vector size, if you think it'd matter.

HanielB (approved):
LGTM, just minor comments

PR #6380, opened 2021-04-16
Minor simplifications to substitutions

ajreynol:
In preparation for using this class as part of our proof checker.

This removes an option that was previously used as a hack to try to make check-models work for quantifiers. It also makes supplying a context optional.


HanielB (approved):
LGTM

PR #6383, opened 2021-04-17
Basic setup for examples in documentation

nafur:
This PR adds the setup for adding examples to the documentation in a structured way, and adds three such examples. The idea is to have the same example for multiple languages and show them on one page in multiple tabs. 
In detail, this PR
- adds a sphinx extension that provides a convenient and uniform creation of tabbed code blocks from a list of files
- uses this extension to show a couple of examples
- adds these to the existing documentation

This PR deliberately does not attempt to collect examples in cmake automatically for two reasons: (a) file names are usually different for idiomatic language-specific reasons that at least ugly to handle in cmake and (b) doing it manually allows us to easily add some custom explanations on the example pages.


aniemetz (approved):
Nice!

PR #6391, opened 2021-04-19
Remove references to get-antlr script

polgreen:
This dependency was refactored and the script no longer exists:

 https://github.com/CVC4/CVC4/commit/4b7e392eca42f89d103c5eb74157a44d4b40fe4b#diff-fc1e87ada6cfa1335e3062451e7ac9d13f17482c4660989eeee7d5ec75a620d6

I * think * it's now dealt with by cmake

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -3,7 +3,6 @@ CVC4 prerelease version 1.9
 
 ## Building CVC4
 
-    ./contrib/get-antlr-3.4  # download and build ANTLR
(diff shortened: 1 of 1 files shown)


nafur:
You raise an excellent point, indeed I completely forgot about the `INSTALL` file when refactoring the dependencies setup. However, I'd like to update the file for all dependencies at once and probably include some new information as well. I would thus close this PR and open a new one in the next few days. I'll mention you in the commit message.

PR #6394, opened 2021-04-19
Fully incorporate quantifiers macros into ppAssert / non-clausal simplification

ajreynol:
This PR removes the quantifiers macro preprocessing pass, which had several shortcomings, both in terms of performance and features.

This makes it so that quantifier macros are the (optional) implementation of TheoryQuantifiers::ppAssert.

In other words, quantifiers can now be put into "solved form", `forall x. P(x)` ---> `P = lambda x. true`.

This is part of an effort to improve proofs for preprocessing, as well as centralizing our reason about substitutions for the sake of efficiency.


test/regress/regress0/quantifiers/issue2033-macro-arith.smt2
```diff
@@ -1,4 +1,4 @@
-; COMMAND-LINE: --macros-quant
```
yoni206, line 1: how come this is needed now and was not needed when it was a preprocessing pass? This is a bit of a regression, right? I am guessing that is used to be able to check models and now it cannot?

src/theory/quantifiers/theory_quantifiers.cpp
```diff
@@ -68,6 +70,11 @@ TheoryQuantifiers::TheoryQuantifiers(Context* c,
   // pointer will be retreived by TheoryEngine and set to all theories
   // post-construction.
   d_quantEngine = d_qengine.get();
+
+  if (options::macrosQuant())
+  {
+    d_qmacros.reset(new QuantifiersMacros(d_qreg));
```
yoni206, line 76: I might be completely off here, but shouldn't this `new` get `delete`ed somewhere?

src/CMakeLists.txt
```diff
@@ -82,8 +82,6 @@ libcvc4_add_sources(
   preprocessing/passes/non_clausal_simp.h
   preprocessing/passes/pseudo_boolean_processor.cpp
   preprocessing/passes/pseudo_boolean_processor.h
-  preprocessing/passes/quantifier_macros.cpp
-  preprocessing/passes/quantifier_macros.h
```
yoni206, line 86: I think @4tXJ7f 's general input about this would be nice. If I remember correctly, quantifier macros was once upon a time in `theory/quantifiers` and then it was moved to `preprocessing`, right? Now we are bringing them back to `theory/quantifiers`. Do we expect other preprocessing passes to have the same fate? And can this recreate the problems that made us move all passes to `preprocessing` in the first place?

src/theory/quantifiers/quantifiers_macros.h
```diff
@@ -0,0 +1,78 @@
...
+namespace cvc5 {
+namespace theory {
+namespace quantifiers {
+
+class QuantifiersRegistry;
+
+/**
+ * A utility for inferring macros from quantified formulas. This can be seen as
+ * a method for putting quantified formulas in solved form, e.g.
+ *   forall x. P(x) ---> P = (lambda x. true)
```
yoni206, line 34: Does this still fall under the name `quantifier macros`? maybe the name should be changed?

src/theory/quantifiers/quantifiers_macros.h
```diff
@@ -0,0 +1,78 @@
...
+  Node solve(Node lit, bool reqGround = false);
+
+ private:
+  bool isBoundVarApplyUf(Node n);
+  bool containsBadOp(Node n, Node op, bool reqGround);
+  bool isGroundUfTerm(Node f, Node n);
+  void getMacroCandidates(Node n,
+                          std::vector<Node>& candidates,
+                          std::map<Node, bool>& visited);
+  Node solveInEquality(Node n, Node lit);
```
yoni206, line 63: Can you add documentation? If it was undocumented already in the preprocessing pass then this is a code move and then i realize my request is unfair, but might still be useful :-)
For example, I am not sure what is a "bad op", and what does it mean to be `BoundVarApplyUf`.

yoni206 (commented):
I have some questions/comments, inside, and also a question for @4tXJ7f inside.

src/theory/quantifiers/theory_quantifiers.cpp
ajreynol: Its allocated into a unique_ptr, this is automatic.

src/theory/quantifiers/quantifiers_macros.h
ajreynol: No, this would still be considered quantifier macros.

src/CMakeLists.txt
ajreynol: There are no problems in this PR.  Another way to view this PR is that we are merging quantifier_macros and non_clausal_simplify, the latter subsumes the former.

test/regress/regress0/quantifiers/issue2033-macro-arith.smt2
ajreynol: Yes, this has revealed a more general issue with check-models, that we don't evaluate terms under quantifiers.  This will be addressed on future PRs.

src/CMakeLists.txt
ajreynol: Or another explanation: the original problem was that we were using custom calls to `theory/`. In this PR, we are using an existing interface (ppAssert) which is called from non-clausal simplification.

yoni206 (approved):
Thanks for the changes and the answers!

PR #6397, opened 2021-04-20
Add std::vector<Term> Op:: getIndices() and operator[] for Op

mudathirmahgoub:


ajreynol (changes requested):
I believe we decided today that this method should return std::vector<api::Term> ?

aniemetz:
Why would this return a vector of Terms? Indices are never terms.

ajreynol:
As I understood the discussion in the meeting, this it because we can index by a string or an int, divisible must be indexed by a string since it is arbitrary precision.  Thus, a more general interface than this one is needed.

mudathirmahgoub:
I guess in the end, it was decided to only add operator[i] to Op to return index  at position i. 
So I guess this function std:vector<uint32_t> should support all indexed operators, and not only ` TUPLE_PROJECT`. 
@4tXJ7f please can you verify this?

4tXJ7f (changes requested):
As Andy mentioned, we have decided to return a `Term` object to cover the different cases and we've also decided that just `operator[]` and `getNumIndices()` should be the interface (which mirrors what we have for children of terms). I don't think that we should construct a vector and then index into that, though. Ideally, the helper just takes the index of the index and returns it depending on the kind.

nafur:
Just as a side note: although I very much encourage using vocabulary types (like `std::variant` or `std::optional`), we should avoid them in the API for now. This would force our users to migrate to C++17, and even if they do that there are some problems with `std::variant` in particular (namely using `clang` with a `g++` 7.1/7.2 stdlib, which currently fails one of our nightlies).

mudathirmahgoub:
> As Andy mentioned, we have decided to return a `Term` object to cover the different cases and we've also decided that just `operator[]` and `getNumIndices()` should be the interface (which mirrors what we have for children of terms). I don't think that we should construct a vector and then index into that, though. Ideally, the helper just takes the index of the index and returns it depending on the kind.

Done. The subscript operator now returns a Term. For the Java API, it makes sense to have a function that returns a vector of Terms for the indices. Otherwise, the user has to call the subscript operator n times, which would make n calls to the c++ api, a practice that is not recommended.

mudathirmahgoub:
Another comment, we have a consistency issue. When operators are created, actual types for indices are required, but when  they are returned, their types would be Term. 
```c++
  std::vector<uint32_t> indices = {0, 3, 2, 0, 1, 2};
  Op tuple_project_op = d_solver.mkOp(TUPLE_PROJECT, indices);
  Term zero =  tuple_project_op[0];
```
To be consistent, mkOp should only receives Terms as arguments.

4tXJ7f:
> Another comment, we have a consistency issue. When operators are created, actual types for indices are required, but when they are returned, their types would be Term.
> 
> ```c++
>   std::vector<uint32_t> indices = {0, 3, 2, 0, 1, 2};
>   Op tuple_project_op = d_solver.mkOp(TUPLE_PROJECT, indices);
>   Term zero =  tuple_project_op[0];
> ```
> 
> To be consistent, mkOp should only receives Terms as arguments. Do you agree?

I think that it is ok to have at least a variant of `mkOp` that accepts a vector of integers (this would be consistent with the other `mkOp` functions), since non-integer indices are basically non-existent.

src/api/cpp/cvc5.cpp
```diff
@@ -1877,9 +1886,193 @@ size_t Op::getNumIndices() const
       break;
     default: CVC5_API_CHECK(false) << "Unhandled kind " << kindToString(k);
   }
+  return size;
+}
+
+Term Op::operator[](size_t index) const
```
ajreynol, line 1892: Shouldn't we just use getIndicesHelper() for this method?  Then the implementation here would only be 2 lines?

ajreynol (changes requested):

src/api/cpp/cvc5.cpp
ajreynol: Either that or the opposite (getIndices uses getIndexHelper or something).

mudathirmahgoub: Yes, there is some redundancy here. Personally I prefer to have getIndicesHelper only. But  @4tXJ7f  commented the following for performance. 
```
I don't think that we should construct a vector and then index into that, though. Ideally, the helper just takes the index of the index and returns it depending on the kind.
```
For the same performance reason, `getIndexHelper` would be called `n` times for `getIndices` if there are `n` indices which is inefficient. 
So I implemented them both separately to optimize both of them. However, I agree maintaining them would be annoying with new indices.

mudathirmahgoub: After offline discussion, we decided to implement `getIndexHelper` which is called by both operator[] and getIndices90

ajreynol (approved):

4tXJ7f (approved):
LGTM. Thanks for the updates and sorry about the delay.

PR #6399, opened 2021-04-20
Fix `ANTLR3_COMMAND` for system ANTLR3 JAR

4tXJ7f:
`ANTLR3_COMMAND` was using a fixed path instead of relying on the
`ANTLR3_JAR` variable. If the ANTLR3 JAR was found on the system (for
example due to an existing `deps` folder at the root of the CVC4
folder), then the command would fail because the JAR was not at the
expected location.  This commit changes the command to use the variable
and prints the location of the JAR file to make debugging easier.


nafur (approved):
LGTM. Thanks!

PR #6403, opened 2021-04-20
UF: Move implementation of type rules to cpp.

aniemetz:


aniemetz:
@ajreynol can you document the header similiarly to https://github.com/CVC4/CVC4/blob/master/src/theory/fp/theory_fp_type_rules.h when this is merged back to master?

aniemetz:
https://github.com/CVC4/cvc4-projects/issues/270

ajreynol (approved):

PR #6405, opened 2021-04-20
Reconciling proofs and unsat cores

HanielB:
This commit changes how defaults are set and how the SMT solver is initialized so that proofs can be used fully with (new) unsat cores. Three modes of unsat cores are established now: 

- the upcoming assumption-based cores, which are incompatible with producing proofs (but enable proofs for preprocessing)
- cores based on the SAT proof, which are incompatible with producing proofs (but enable proofs for preprocessing and the SAT solver)
- cores based on the full proof, which are unrestricted

All the modes activate proofs but lead to errors if the user requires proofs but is not in the full proofs mode for cores.


src/smt/set_defaults.cpp
```diff
@@ -92,12 +92,14 @@ void setDefaults(LogicInfo& logic, bool isInternalSubsolver)
     }
   }
   if (options::unsatCoresNew()
+      && options::unsatCoresMode() != options::UnsatCoresMode::FULL_PROOF
```
ajreynol, line 95: This block of code should check `options::unsatCoresNew() && options::produceProofs()`
and set `unsatCoresMode` to FULL_PROOF if so, with a `Notice` if the unsatCoreMode was set by user.

It should be run *after* the next block of code (and any other block of code that could set `produceProofs` to true).

src/smt/set_defaults.cpp
```diff
@@ -92,12 +92,14 @@ void setDefaults(LogicInfo& logic, bool isInternalSubsolver)
...
           || (options::checkProofs() && options::checkProofs.wasSetByUser())
           || (options::dumpProofs() && options::dumpProofs.wasSetByUser())))
   {
-    AlwaysAssert(false) << "Can't properly produce proofs and have the new "
-                           "unsat cores simultaneously.\n";
+    AlwaysAssert(false)
+        << "Can't simultaneously produce proofs and not have the new "
+           "unsat cores in full proof mode .\n";
   }
   if (options::checkProofs() || options::unsatCoresNew()
```
ajreynol, line 104: You should also enable proofs if unsat core mode was set to full proof.

src/options/smt_options.toml
```diff
@@ -200,6 +200,25 @@ header = "options/smt_options.h"
   type       = "bool"
   help       = "turn on unsat core generation"
 
+[[option]]
+  name       = "unsatCoresMode"
+  smt_name   = "unsat-cores-mode"
+  category   = "regular"
+  long       = "unsat-cores-mode=MODE"
+  type       = "UnsatCoresMode"
+  default    = "SAT_PROOF"
```
ajreynol, line 209: Default should be ASSUMPTIONS eventually right?

ajreynol (changes requested):

src/options/smt_options.toml
HanielB: Definitely. When it's merged to master this should be updated.

src/smt/set_defaults.cpp
HanielB: Good point! I changed accordingly.

src/smt/set_defaults.cpp
HanielB: We need to set produceProofs for all unsat core modes, so I think this is already taken care of by your previous suggestion.

src/smt/set_defaults.cpp
```diff
@@ -91,20 +91,25 @@ void setDefaults(LogicInfo& logic, bool isInternalSubsolver)
...
     Notice() << "SmtEngine: setting proof" << std::endl;
     options::produceProofs.set(true);
   }
+
+  if (options::unsatCoresNew() && options::produceProofs())
+  {
+    if (options::unsatCoresMode() != options::UnsatCoresMode::FULL_PROOF
+        && options::unsatCoresMode.wasSetByUser())
+    {
+      Notice() << "Forcing full-proof mode unsat cores mode since proofs are "
```
ajreynol, line 106: I would guard the notice with `wasSetByUser`.

The mode should be set regardless.

src/smt/smt_engine.cpp
```diff
@@ -238,10 +238,13 @@ void SmtEngine::finishInit()
...
-  // if proofs and unsat cores, proofs are used solely for unsat core
-  // production, so we don't generate proofs in the theory engine, which is
-  // communicated via the second argument
-  d_smtSolver->finishInit(logic, options::unsatCoresNew());
+  // if proofs and unsat cores, proofs are used solely for unsat core production
+  // unless the unsat cores are in FULL_PROF mode. Thus we don't generate proofs
+  // in the theory engine, which is communicated via the second argument
+  d_smtSolver->finishInit(
+      logic,
+      options::unsatCoresNew()
```
ajreynol, line 246: I think it may be better to just push this logic into SmtSolver.

The issue here is that this Boolean value isn't being used in a central way, thus SmtSolver::finishInit is probably a more appropriate place to query the options.

ajreynol (changes requested):

ajreynol (changes requested):
One thing that I think is still missing: `produceProofs` should turn `unsatCoresNew` on, right?

If the user just writes `--produce-proofs`, I think we should allow them to get unsat cores, right?

src/smt/set_defaults.cpp
```diff
@@ -91,20 +91,37 @@ void setDefaults(LogicInfo& logic, bool isInternalSubsolver)
...
+
+  if (options::produceProofs() && !options::unsatCoresNew())
+  {
+    if (options::unsatCoresNew.wasSetByUser())
+    {
+      Notice() << "Since proofs are produced, we necessarily also produce "
+                  "unsat cores (in full-proof mode)."
+               << std::endl;
+    }
+    options::unsatCores.set(true);
```
ajreynol, line 121:
```suggestion
    options::unsatCoresNew.set(true);
```

src/smt/set_defaults.cpp
```diff
@@ -91,20 +91,37 @@ void setDefaults(LogicInfo& logic, bool isInternalSubsolver)
...
+    if (options::unsatCoresMode.wasSetByUser())
+    {
+      Notice() << "Forcing full-proof mode unsat cores mode since proofs are "
+                  "being produced."
+               << std::endl;
+    }
+    options::unsatCoresMode.set(options::UnsatCoresMode::FULL_PROOF);
+  }
+
+  if (options::produceProofs() && !options::unsatCoresNew())
```
ajreynol, line 113: I think this block can be combined with the one above.

ajreynol (changes requested):

src/preprocessing/passes/non_clausal_simp.cpp
```diff
@@ -69,8 +69,11 @@ NonClausalSimp::NonClausalSimp(PreprocessingPassContext* preprocContext)
 PreprocessingPassResult NonClausalSimp::applyInternal(
     AssertionPipeline* assertionsToPreprocess)
 {
-  Assert(!options::unsatCores() || isProofEnabled())
-      << "Unsat cores with non-clausal simp only supported with new proofs";
+  Assert(options::unsatCoresMode() <= options::UnsatCoresMode::OLD_PROOF
```
ajreynol, line 72:
```suggestion
  Assert(options::unsatCoresMode() != options::UnsatCoresMode::OLD_PROOF
```

src/smt/set_defaults.cpp
```diff
@@ -52,59 +52,82 @@ void setDefaults(LogicInfo& logic, bool isInternalSubsolver)
...
+    options::unsatCoresMode.set(options::UnsatCoresMode::FULL_PROOF);
   }
-  if (options::unsatCoresNew()
-      && ((options::produceProofs() && options::produceProofs.wasSetByUser())
-          || (options::checkProofs() && options::checkProofs.wasSetByUser())
-          || (options::dumpProofs() && options::dumpProofs.wasSetByUser())))
+
+  // set proofs on if not yet set
+  if (options::unsatCores() && !options::produceProofs()
+      && options::unsatCoresMode() != options::UnsatCoresMode::OLD_PROOF)
```
ajreynol, line 105: I think this is wrong for ASSUMPTIONS unsat cores.

src/smt/set_defaults.cpp
```diff
@@ -368,12 +387,12 @@ void setDefaults(LogicInfo& logic, bool isInternalSubsolver)
   // if we requiring disabling proofs, disable them now
   if (disableProofs && options::produceProofs())
   {
-    if (options::unsatCoresNew())
+    if (options::unsatCoresMode() > options::UnsatCoresMode::OLD_PROOF)
```
ajreynol, line 390: I'd not rely on the order here, let's make the cases explicit.

ajreynol (changes requested):

src/smt/set_defaults.cpp
HanielB: I don't think I see how. It'd have been a scenario in which one e.g. passes `--produce-unsat-cores --unsat-cores-mode=assumptions`, which would require us turning proofs and the unsat core mode would not be changed (since proofs were not on in the test in the code block above). The initialization of the solver would be done as expected, I believe.

ajreynol: So you're saying that this wouldn't happen? Maybe add an assert for this case? E.g. that proofs are enabled when we have unsat cores ASSUMPTIONS.

PR #6406, opened 2021-04-20
cmake: Add optional module name argument to check_python_module.

mpreiner:
Addtionally adds checks for required documentation modules.


nafur (approved):
LGTM.

PR #6407, opened 2021-04-20
Arrays: Move implementation of type rules to cpp.

aniemetz:


aniemetz:
https://github.com/CVC4/cvc4-projects/issues/271

mpreiner (approved):

PR #6409, opened 2021-04-21
Allow in-place construction of `CDList` items

4tXJ7f:
This commit adds `CDList::emplace_back()`, which allows users to create
elements in `CDList` in-place (as opposed to copying the items using
`CDList::push_back()`. This allows `CDList` to be used with
`std::unique_ptr`s, which do not allow copying. Using
`CDList::emplace_back()` could also be more efficient in certain cases.


mpreiner (approved):
LGTM

PR #6410, opened 2021-04-21
Add explicit dependencies for base library

nafur:
This PR adds missing cmake dependencies for the base library. This makes sure that `Debug_tags.h` and `Trace_tags.h` are already present when we start compiling.


nafur:
Whether you see this issue right now depends on how cmake orders independent targets. I, personally, have not observed it on my machine yet. I'm not sure whether the dependency on `gen-gitinfo` is needed, as it may be added automatically as the `BYPRODUCTS` file is part of the sources of `cvc4base`. It should not hurt, though.

mpreiner (approved):

PR #6413, opened 2021-04-21
Add CI jobs to build docs

nafur:
This PR adds CI jobs to automatically build documentation for branches on the main repository and for pull request.
The first job builds the documentation for every push and pr and stores the generated documentation in `cvc5.github.io/docs-ci`. All versions are stored and for every branch and PR a symbolic link to the most recent version is maintained.
Note that the PR jobs are run by the `pull_request_target` trigger that allows access to repository secrets, but runs in the context of the target branch and we thus need to (carefully!) pull in the relevant changes manually.
The second job runs once a week and prunes the docs-ci repository: all directories that have not been touched for a month are removed, and all commits older than a month are squashed into a single commit. This retains the full history for the last month, but effectively removes everything older than that.


nafur:
For this PR to work, we need the repository `cvc5-docs.github.io`, probably a `cvc5-docs` user, and a personal token with (force) push access to `cvc5-docs.github.io` stored as `DOCS_TOKEN`.

.github/workflows/docs_update.yml
```diff
@@ -0,0 +1,86 @@
...
+    paths:
+    - '.ci/**'
+    - 'docs/**'
+    - 'examples/**'
+    - 'src/api/**'
+
+# When run by pull_request_target (in comparison to pull_request), the action
+# has access to the repository secrets so that we can push the generated docs.
+# To allow this, the action is run in the context of the master branch instead
+# of the PR directly. We therefore to obtain the changes from the PR manually,
```
mpreiner, line 19:
```suggestion
# of the PR directly. We therefore obtain the changes from the PR manually,
```

.github/workflows/docs_update.yml
```diff
@@ -0,0 +1,86 @@
...
+            libgmp-dev \
+            doxygen \
+            python3-docutils
+          python3 -m pip install \
+            setuptools toml \
+            sphinxcontrib-bibtex sphinx-tabs sphinx-rtd-theme \
+            breathe
+
+      - name: Configure
+        run: ./configure.sh debug --docs
```
mpreiner, line 51: Why a debug build?

.github/workflows/docs_update.yml
```diff
@@ -0,0 +1,86 @@
...
+            NAME=$(echo "${GITHUB_REF}" | sed -E "s,refs/pull/([0-9]+)/merge,\1, ")
+            echo "Identified PR #$NAME"
+            NAME="pr$NAME"
+            # be careful here, see explanation above!
+            git checkout "${GITHUB_REF}" -- \
+              `git ls-tree "${GITHUB_REF}" --name-only -r docs/ | grep -E ".*\.(rst|bib)"` \
+              `git ls-tree "${GITHUB_REF}" --name-only -r src/api/ | grep -E ".*\.(h|cpp|java|py)"` \
+              `git ls-tree "${GITHUB_REF}" --name-only -r examples/ | grep -E ".*\.(h|cpp|java|py|smt2)"`
+          fi
+          if [ ! -z "$NAME" ]; then
```
mpreiner, line 76:
```suggestion
          if [ -n "$NAME" ]; then
```

nafur:
Everything is set up now, this PR should be good to go. See https://cvc5.github.io/docs-ci/

aniemetz (commented):
What if we have multiple PRs that make changes to docs? Are they distinguished by PR id? Right now it doesn't look like it.

nafur:
> What if we have multiple PRs that make changes to docs? Are they distinguished by PR id? Right now it doesn't look like it.

I'm trying to do this right here (https://github.com/CVC4/CVC4/pull/6413/files#diff-277cba447e0ae243607c3726d2b4296e2c11c4e8ee294a9c28d5ff037a3fa784R75 and following). While I was trying to back this code up with references to the github documentation, I realized this would not work... I'll fix it :-)

nafur:
Okay, it works now (tested with as PR on my fork).

aniemetz (approved):

PR #6414, opened 2021-04-21
Add assertion list utility for justification heuristic

ajreynol:


src/decision/assertion_list.h
```diff
@@ -0,0 +1,105 @@
...
+  AssertionList(context::Context* ac,
+                context::Context* ic,
+                bool useDyn = false);
+  virtual ~AssertionList() {}
+  /** Presolve, which clears the dynamic assertion order */
+  void presolve();
+  /** Add the assertion n */
+  void addAssertion(TNode n);
+  /**
+   * Get the new assertion, increment d_assertionIndex, sets fromIndex to the
```
mpreiner, line 74: Why does a local variable `fromIndex` matter here? Maybe add some more documentation to clarify.

src/decision/assertion_list.cpp
```diff
@@ -0,0 +1,134 @@
...
+    // no decision does not impact the decision order
+    return;
+  }
+  std::vector<TNode>::iterator it =
+      std::find(d_dlist.begin(), d_dlist.end(), n);
+  if (s == DecisionStatus::DECISION)
+  {
+    if (it == d_dlist.end())
+    {
+      // if we just had status on an assertion and it didnt occur id dlist,
```
mpreiner, line 110:
```suggestion
      // if we just had a status on an assertion and it didn't occur in dlist,
```

src/decision/assertion_list.cpp
```diff
@@ -0,0 +1,134 @@
...
+      // then our index should have exhausted dlist
+      Assert(d_dindex.get() == d_dlist.size());
+      if (d_dindex.get() == d_dlist.size())
```
mpreiner, line 113: Why the `if` if you already assert the condition?

src/decision/assertion_list.cpp
```diff
@@ -0,0 +1,134 @@
...
+    // not using dynamic ordering, return
+    return;
+  }
+  if (s == DecisionStatus::NO_DECISION)
+  {
+    // no decision does not impact the decision order
+    return;
+  }
+  std::vector<TNode>::iterator it =
+      std::find(d_dlist.begin(), d_dlist.end(), n);
```
mpreiner, line 105: How long is `d_dlist` expected to be? `std::find` over long vectors may be sub-optimal.

src/decision/assertion_list.h
ajreynol: This was actually leftover from something else I was trying and is not needed anymore, I'll remove.

src/decision/assertion_list.cpp
ajreynol: This is to make the code safe, I'd like it to fail in debug but not execute the branch otherwise (which can lead to segfaults).

src/decision/assertion_list.cpp
ajreynol: `d_dlist` is worst case equal to # assertions, but typically will be roughly equal to maximum assertion level of the SAT solver.

Since d_dlist must be ordered, I believe the alternative is to use an `unordered_set` in addition to this list, which would be double memory but maybe faster lookup.  I could add this, although I have mixed feelings.

What do you think?

mpreiner: Depends on what we are concerned about: speed or memory? :)
I think an additional cache shouldn't hurt too much.

mpreiner (approved):

PR #6416, opened 2021-04-21
(proof-new) Proofs for sets purification lemmas

ajreynol:
This adds proofs for sets purification lemmas, which are of the form `(= t (skolem t))` and `(member t (skolem (singleton t)))`.  Each can be trivially justified in the internal calculus by `MACRO_SR_PRED_INTRO`.


src/theory/sets/term_registry.cpp
```diff
@@ -152,6 +154,21 @@ void TermRegistry::debugPrintSet(Node s, const char* c) const
   }
 }
 
+void TermRegistry::sendSimpleLemmaInternal(Node eq, InferenceId id)
```
mudathirmahgoub, line 157: The node argument here could be an equality node, or a singleton node. Please rename the argument to `n`

src/theory/sets/skolem_cache.cpp
```diff
@@ -35,8 +35,19 @@ Node SkolemCache::mkTypedSkolemCached(
...
+    {
+      Assert(a.getType() == tn);
+      sk = sm->mkPurifySkolem(a, c);
+    }
+    else
+    {
+      sk = sm->mkDummySkolem(c, tn, "sets skolem");
+    }
     d_skolemCache[a][b][id] = sk;
+    d_allSkolems.insert(sk);
```
mudathirmahgoub, line 50: Is this needed? It seems this was working before.  The function `isSkolem` which uses this private field `d_allSkolems` is not used anywhere.

mudathirmahgoub (approved):
LGTM. It seems we don't need the private field `d_allSkolems`, but its existence only incurs memory cost.

src/theory/sets/skolem_cache.cpp
ajreynol: isSkolem is used in sets solver state still.

I initially thought this wasnt needed but this does impact how the skolems are handled.

PR #6420, opened 2021-04-21
Bool: Move implementation of type rules to cpp.

aniemetz:


aniemetz:
https://github.com/CVC4/cvc4-projects/issues/274

src/theory/booleans/theory_bool_type_rules.cpp
```diff
@@ -0,0 +1,79 @@
...
+      {
+        Debug("pb") << "failed type checking: " << *child_it << std::endl;
+        Debug("pb") << "  integer: " << (*child_it).getType(check).isInteger()
+                    << std::endl;
+        Debug("pb") << "  real: " << (*child_it).getType(check).isReal()
+                    << std::endl;
+        throw TypeCheckingExceptionPrivate(n,
+                                           "expecting a Boolean subexpression");
+      }
+    }
```
mpreiner, line 45: How about changing this to a foreach loop?

mpreiner (approved):

PR #6421, opened 2021-04-21
Fix models for sygus-inference, bv2int, real2int

ajreynol:
In each case, previously we were generating a define-fun, what we needed was to do a model substitution.

This actually meant that check-models was giving false positives.  The model was incorrect but check-models succeeded due to its use of expand definitions.


src/preprocessing/passes/bv_to_int.cpp
```diff
@@ -841,7 +841,12 @@ void BVToInt::defineBVUFAsIntUF(Node bvUF, Node intUF)
   // If the result is BV, it needs to be casted back.
   result = castToType(result, resultType);
   // add the function definition to the smt engine.
```
yoni206, line 843: update comment?

src/preprocessing/passes/real_to_int.cpp
```diff
@@ -188,8 +188,7 @@ Node RealToInt::realToIntInternal(TNode n, NodeMap& cache, std::vector<Node>& va
           var_eq.push_back(n.eqNode(ret));
           // ensure that the original variable is defined to be the returned
           // one, which is important for models and for incremental solving.
```
yoni206, line 190:
```suggestion
          // ensure that the original variable is substituted to the returned
          // one, which is important for models and for incremental solving and models.
```

src/preprocessing/preprocessing_pass_context.h
```diff
@@ -81,11 +81,21 @@ class PreprocessingPassContext
   void recordSymbolsInAssertions(const std::vector<Node>& assertions);
 
   /**
-   * Add substitution to theory model.
+   * Add substitution to theory model. This method should only be called if
+   * we have already added the substitution to the top-level substitutions
+   * class. Otherwise, addSubstitution should be called instead.
```
yoni206, line 86: If there is a way to partially check this using an Assert in the code, that would be nice.

src/preprocessing/preprocessing_pass_context.h
```diff
@@ -81,11 +81,21 @@ class PreprocessingPassContext
...
    * @param lhs The node replaced by node 'rhs'
    * @param rhs The node to substitute node 'lhs'
    */
   void addModelSubstitution(const Node& lhs, const Node& rhs);
+  /**
+   * Add substitution to the top-level substitutions and to the theory model.
+   * @param lhs The node replaced by node 'rhs'
+   * @param rhs The node to substitute node 'lhs'
```
yoni206, line 94: maybe also describe pg?

test/regress/regress2/sygus/issue4022-conjecture-gen.smt2
```diff
@@ -1,3 +1,5 @@
+; COMMAND-LINE: -q
```
yoni206, line 1: one day we should grep `-q` in the regression tests and try to eliminate them, like we did once for `--no-check-proof`. But I guess that day is not here yet, and now we are still adding `-q`s :)

yoni206 (commented):
LGTM, thanks a lot for doing this! Left some documentation comments.

src/preprocessing/preprocessing_pass_context.h
ajreynol: Not at the moment, due to the order in which things occur.  In the end, we need to clean up calls to this method so that everything calls methods for adding substitutions that ensure this invariant is maintained.

test/regress/regress2/sygus/issue4022-conjecture-gen.smt2
ajreynol: Yep, exactly, I'm planning to fix check-models soon. The core issue is that bodies of quantified formulas are not evaluated wrt substitutions when they should be.

yoni206 (approved):
Thanks for the changes

PR #6422, opened 2021-04-22
Remove unused stuff from options setup

nafur:
This PR removes some old stuff from our options setup that has not been used in a long time.
Most prominently, this includes the man pages that were still generated, and the `alias` and `links` options, which no longer worked anyway.


mpreiner (approved):
LGTM

PR #6423, opened 2021-04-22
New design in DOT representation, nodes colored based on visions(basic and propositional)

diegodrc:
Conclusion and rule are placed on the same node.
Nodes are colored based on the vision they will belong to.
Signed-off-by: Diego Della Rocca de Camargos <diegodellarocc@gmail.com>


src/proof/dot/dot_printer.h
```diff
@@ -66,10 +65,10 @@ class DotPrinter
   /** Replace all quotes but escaped quotes in given string
    * @param s The string to have the quotes processed.
    */
-  static void cleanQuotes(std::string& s);
+  static std::string sanitizeString(const std::string&);
```
HanielB, line 68: Update the documentation. Also use names for the arguments.

src/proof/dot/dot_printer.h
```diff
@@ -50,9 +50,8 @@ class DotPrinter
    * @param pn the proof node to print
    * @param ruleID the id of the rule to print
    */
-  static void printInternal(std::ostream& out,
-                            const ProofNode* pn,
-                            uint64_t& ruleID);
+  static void printInternal(
+      std::ostream& out, const ProofNode* pn, uint64_t& ruleID, int, bool);
```
HanielB, line 54: Add names for the arguments and add documentation above to describe them. Moreove, use `uint64_t` rather than `int`. We favor unsigned variables when we only care for non-negative values. And we specify explicitly the bit-width to avoid possible issues when running in different architectures.

src/proof/dot/dot_printer.cpp
```diff
@@ -25,55 +25,103 @@
 namespace cvc5 {
 namespace proof {
 
-void DotPrinter::cleanQuotes(std::string& s)
+// guarantee that string does not have unescaped quotes
```
HanielB, line 28: General method documentation should go in the header file.

src/proof/dot/dot_printer.cpp
```diff
@@ -25,55 +25,103 @@
...
+std::string DotPrinter::sanitizeString(const std::string& s)
 {
-  std::string rep("\\\"");
-  for (size_t pos = 0; (pos = s.find("\"", pos)) != std::string::npos;
-       pos += rep.length())
+  std::string newS;
+  newS.reserve(s.size());
```
HanielB, line 32: why do you reverse?

src/proof/dot/dot_printer.cpp
```diff
@@ -25,55 +25,103 @@
...
+  return newS;
 }
 
 void DotPrinter::print(std::ostream& out, const ProofNode* pn)
 {
   uint64_t ruleID = 0;
 
-  out << "digraph proof {\n";
-  DotPrinter::printInternal(out, pn, ruleID);
+  out << "digraph proof {\n\trankdir=\"BT\";\n\tnode [shape=record];\n";
```
HanielB, line 54: Say what these new components of the output do.

src/proof/dot/dot_printer.cpp
```diff
@@ -25,55 +25,103 @@
...
+      if (scopeNumber < 2)
+      {
+        classes << " basic";
+        colors << ", color = blue ";
+      }
+      if (inPropositionalVision)
+      {
+        classes << " propositional";
+        colors << ", fillcolor=aquamarine4, style=filled ";
+      }
```
HanielB, line 103: Are these two supposed to be exclusive (one or the other)? It's a bit confusing as it is.

HanielB (changes requested):
A few comments:

src/proof/dot/dot_printer.cpp
diegodrc: They are not supposed to be exclusive. A node can belong at the same time to the basic vision and the propositional vision, right?

src/proof/dot/dot_printer.cpp
diegodrc: At that moment I know the minimum size of the string, so I allocate memory to it.

HanielB: I misread reserve or reverse, sorry :)

src/proof/dot/dot_printer.cpp
HanielB: They can indeed. Add a comment to say that this is the case.

src/proof/dot/dot_printer.cpp
```diff
@@ -25,55 +25,108 @@
...
+  return newS;
 }
 
 void DotPrinter::print(std::ostream& out, const ProofNode* pn)
 {
   uint64_t ruleID = 0;
 
-  out << "digraph proof {\n";
-  DotPrinter::printInternal(out, pn, ruleID);
+  // The dot attribute rankdir="BT" sets the direction of graph layout, to place
```
HanielB, line 53:
```suggestion
  // The dot attribute rankdir="BT" sets the direction of the graph layout, placing
```

src/proof/dot/dot_printer.cpp
```diff
@@ -25,55 +25,108 @@
...
+  // the root node at the top. The "node [shape..." sets the shape of all nodes
```
HanielB, line 54:
```suggestion
  // the root node at the top. The "node [shape..." attribute sets the shape of all nodes
```

src/proof/dot/dot_printer.h
```diff
@@ -40,19 +40,24 @@ class DotPrinter
...
-   * Print the rule in the format:
-   * "$RULE_ID $RULE_NAME($RULE_ARGUMENTS)" [ shape = "box"];
-   * "$RULE_ID $RULE_NAME($RULE_ARGUMENTS)" -> "$RULE_ID $RULE_CONCLUSION";
-   * and then for each child of the rule print
-   * "$CHILD_ID $CHILD_CONCLUSION" -> "$RULE_ID $RULE_NAME($RULE_ARGUMENTS)";
-   * and then recursively call the function with the child as argument.
+   * Print the nodes of the proof in the format:
+   * $NODE_ID [ label = "{$CONCLUSION|$RULE_NAME($RULE_ARGUMENTS)}",
+   * $COLORS_AND_CLASSES_RELATED_TO_THE_RULE ]; and then for each child of the
+   * node $CHILD_ID -> $NODE_ID; and then recursively call the function with the
```
HanielB, line 46:
```suggestion
   * node $CHILD_ID -> $NODE_ID; and then recursively calls the function with the
```

src/proof/dot/dot_printer.h
```diff
@@ -40,19 +40,24 @@ class DotPrinter
...
+   * child as argument.
    * @param out the output stream
    * @param pn the proof node to print
    * @param ruleID the id of the rule to print
+   * @param scopeCounter counter to check how many SCOPE was already sawed in
+   * the proof
```
HanielB, line 52:
```suggestion
   * @param scopeCounter counter of how many SCOPE were already depth-first traversed in the proof up to this point
```

src/proof/dot/dot_printer.h
```diff
@@ -40,19 +40,24 @@ class DotPrinter
...
+   * @param inPropositionalView flag used to mark the nodes that belong to the
+   * propositional view proof
```
HanielB, line 54:
```suggestion
   * @param inPropositionalView flag used to mark the proof node being traversed was generated by the SAT solver and thus belong to the propositional view 
```

HanielB (changes requested):
Almost there. :) Just a few suggestions on the wording of comments. Make sure you format afterwards, I messed up the formatting with some of the suggestions, I believe.

HanielB (approved):

PR #6424, opened 2021-04-22
Enable strings exp by default for strings specific logics

ajreynol:
One of the main motivations for this PR is to simplify our process for doing SMT-LIB wide runs.


nafur (approved):
LGTM.

PR #6428, opened 2021-04-22
First part of options refactoring

nafur:
This PR does a first round of refactoring and gets rid of a significant portion of generated code. In particular
- it removes `options::optionName.wasSetByUser()` (we still have `Options::wasSetByUser()`)
- it removes `options::optionName.set()` (we still have `Options::set()`)
- it removes `options::optionName.getName()` in favor of `options::optionName.name`
- it removes the specializations of `Options::assign()` and `Options::assignBool()` from the headers
- it eliminates `runHandlerAndPredicates()` and `runBoolPredicates()`

The removed methods are only used in few places with are changed to using `Options::current().X()` instead.

In the future, we also want to get rid of `options::optionName()` and use `Options::operator[]()` instead, and furthermore not use `Options::current()` but use the options from the `Env` object. This PR already adds `Env::getOption()` as a shorthand for `Env::getOptions()[...]` and uses it as a proof of concept within `SmtEngine`.


src/smt/set_defaults.cpp
```diff
@@ -52,91 +52,91 @@ void setDefaults(LogicInfo& logic, bool isInternalSubsolver)
   // implied options
   if (options::debugCheckModels())
   {
-    options::checkModels.set(true);
+    Options::current().set(options::checkModels, true);
```
ajreynol, line 55: Can we already just store a local variable `Options& opts = Options::current()` at the top of this file?

Eventually we will pass this to this method instead of calling Options::current() at all.

src/theory/arith/theory_arith_private.cpp
```diff
@@ -3118,7 +3118,7 @@ bool TheoryArithPrivate::solveRealRelaxation(Theory::Effort effortLevel){
 //               d_qflraStatus = d_attemptSolSimplex.attempt(relaxSolution);
 //             }
 //           }
-//           options::arithStandardCheckVarOrderPivots.set(pass2Limit);
+//           Options::current().set(options::arithStandardCheckVarOrderPivots, pass2Limit);
```
ajreynol, line 3121: We can probably just delete this commented out code.

ajreynol (changes requested):

ajreynol (approved):
LGTM, although I'd like @mpreiner  to have a look at the changes to mkoptions.py

src/printer/printer.cpp
```diff
@@ -133,10 +133,10 @@ Printer* Printer::getPrinter(OutputLanguage lang)
   // Options can be null in certain circumstances (e.g., when printing
   // the singleton "null" expr.  So we guard against segfault
   if(not Options::isCurrentNull()) {
-    if(options::outputLanguage.wasSetByUser()) {
+    if(Options::current().wasSetByUser(options::outputLanguage)) {
       lang = options::outputLanguage();
     }
-    if(lang == language::output::LANG_AUTO && options::inputLanguage.wasSetByUser()) {
+    if(lang == language::output::LANG_AUTO && Options::current().wasSetByUser(options::inputLanguage)) {
```
mpreiner, line 139: clang-format

src/options/options_template.cpp
```diff
@@ -435,7 +435,7 @@ void Options::parseOptionsRecursive(Options* options,
   }
 
   // Having this synonym simplifies the generation code in mkoptions.
-  options::OptionsHandler* handler = options->d_handler;
+  options::OptionsHandler* d_handler = options->d_handler;
```
mpreiner, line 438: Why this name change? `d_handler` is not a data member of Options is it?

src/options/options_handler.cpp
```diff
@@ -159,13 +159,13 @@ void OptionsHandler::checkBvSatSolver(std::string option, SatSolverMode m)
...
     {
       throwLazyBBUnsupported(m);
     }
-    if (!options::bitvectorToBool.wasSetByUser())
+    if (!Options::current().wasSetByUser(options::bitvectorToBool))
     {
-      options::bitvectorToBool.set(true);
+      Options::current().set(options::bitvectorToBool, true);
     }
   }
```
mpreiner, line 170: Since we use this construct quite a lot:

```
if (!wasSetByUser())
{
  set(...)
}
```

How about we have an additional `Options::setDefault` for these cases that only sets an option if it was not set by the user?

src/options/options_handler.cpp
```diff
@@ -198,27 +198,27 @@ void OptionsHandler::checkBitblastMode(std::string option, BitblastMode m)
...
     } else {
-      options::BitblastMode mode = stringToBitblastMode("", "eager");
-      options::bitblastMode.set(mode);
+      options::BitblastMode mode = stringToBitblastMode("eager");
+      Options::current().set(options::bitblastMode, mode);
     }
-    if(!options::bitvectorAigSimplifications.wasSetByUser()) {
-      options::bitvectorAigSimplifications.set("balance;drw");
+    if(!Options::current().wasSetByUser(options::bitvectorAigSimplifications)) {
+      Options::current().set(options::bitvectorAigSimplifications, "balance;drw");
```
mpreiner, line 221: Shouldn't this default actually be specified in the toml file, not here?

mpreiner (commented):
Make sure to clang-format your changes.

src/options/options_template.cpp
nafur: `mkoptions.py` uses the same template in various places. In one place, the handler is called `d_handler`...
I can check whether I can improve on this...

src/options/options_handler.cpp
nafur: Not sure, you tell me ;-)
But given how the relevant options are named and the help message of these options, I'd agree...

src/options/options_template.cpp
nafur: This weirdness was required solely because `parseOptionsRecursive` was a static method for ... no particular reason at all. :-)

src/options/options.h
```diff
@@ -123,7 +120,21 @@ class CVC5_EXPORT Options
   void set(T t, const typename T::type& val) {
     ref(t) = val;
   }
-  
+
+  /**
+   * Set the default value of the given option. Is equivalent to calling `set()`
+   * if `wasSetByUser()` return false. Uses `ref()`, which causes a compile-time
```
mpreiner, line 126:
```suggestion
   * if `wasSetByUser()` returns false. Uses `ref()`, which causes a compile-time
```

mpreiner (approved):
LGTM, thanks for the changes! I found one minor typo.

PR #6429, opened 2021-04-23
Add new substitution apply methods fixpoint, sequential, simultaneous

ajreynol:
This adds various methods for applying substitution as an options to `MACRO_SR_*` rules.  It extends the proof checker and the proof post processor to eliminate based on these types.

It updates the trust substitutions utility used by non-clausal simplification to use fixed-point semantics for substitution, which is highly important for efficiency.

As a result of this PR, we are orders of magnitude faster for checking proofs for problems where non-clausal substitution infers many substitutions.  It also makes our regressions noticeably faster:
```
regress0    = 128.68 sec*proc (1360 tests)
regress1    = 211.34 sec*proc (941 tests)
regress2    = 224.82 sec*proc (112 tests)
```
New:
```
regress0    = 124.76 sec*proc (1360 tests)
regress1    = 187.99 sec*proc (941 tests)
regress2    = 178.80 sec*proc (112 tests)
```


src/theory/trust_substitutions.cpp
```diff
@@ -204,7 +204,11 @@ std::shared_ptr<ProofNode> TrustSubstitutionMap::getProofFor(Node eq)
     }
   }
   Trace("trust-subs-pf") << "...apply eq intro" << std::endl;
-  if (!d_tspb->applyEqIntro(n, ns, pfChildren, d_ids))
+  // We use fixpoint as the substitution-apply identifier. Notice that it
+  // suffices to use SBA_SEQUENTIAL here, but SBA_FIXPOINT is typically
+  // more efficient, since it does not require linear traversals of the term
+  // or quadratic traversals of the range of substitutions.
```
HanielB, line 210: Maybe an example to explain the different asymptotic behavior?

src/theory/builtin/proof_checker.cpp
```diff
@@ -187,48 +195,60 @@ bool BuiltinProofRuleChecker::getSubstitutionFor(Node exp,
...
 Node BuiltinProofRuleChecker::applySubstitution(Node n,
                                                 const std::vector<Node>& exp,
-                                                MethodId ids)
+                                                MethodId ids,
+                                                MethodId ida)
 {
-  Node curr = n;
-  // apply substitution one at a time, in reverse order
+  // Build a simultaneous subsitution, built in reverse order. This is done
+  // for performance reasons, so that only one Node::substitute call is made.
```
HanielB, line 213: the sequential case applies Node::substitute vars.size() times. Clarify.

HanielB (changes requested):

src/theory/builtin/proof_checker.cpp
ajreynol: This comment is outdated, ill remove.

HanielB (approved):
LGTM

PR #6430, opened 2021-04-23
Make sure a ReferenceStat is set to values of the correct type

nafur:
This PR fixes a very subtle issue with setting the values a `ReferenceStat` refers to.
`ReferenceStat::set()` would take a variable by `const&` and then store the pointer to it. When giving it a different, but implicitly convertible, type, the pointer would assume the wrong type and consequently read incorrect values from it.
This PR makes `set()` a template function that explicitly checks that the given type is the correct one.
As we can only export `int64_t` to the API, this forces users of `ReferenceStat` to use `int64_t` stats.


ajreynol (approved):
LGTM

PR #6432, opened 2021-04-23
Use std::hash for API types

nafur:
This PR replaces all `api::*HashFunction` objects by specializations of `std::hash`. `std::hash` is meant to be extended in this way, and it allows for a much cleaner usage of `std::unordered_set` and `std::unordered_map` with these types.


src/api/cpp/cvc5.h
```diff
@@ -889,6 +899,28 @@ class CVC5_EXPORT Op
...
+namespace std {
+/**
+ * Hash function for Ops.
+ */
+template <>
+struct CVC5_EXPORT hash<cvc5::api::Op>
+{
+  size_t operator()(const cvc5::api::Op& t) const;
+};
+}  // namespace std
```
mpreiner, line 921: Should we collect all `std::hash` into one std namespace at the end of the file?

mpreiner (approved):

PR #6436, opened 2021-04-23
Improve getValue for non-evaluated operators

ajreynol:
This makes it so that we attempt evaluation + rewriting on applications of operators that do not always evaluate, and return constants in case the evaluation was successful.

This fixes warnings for check-models on 43 of our regressions, and also uncovered one regression where our model was wrong but check-models silently succeeded. I've opened https://github.com/CVC4/cvc4-projects/issues/276 for fixing the latter.


test/regress/CMakeLists.txt
```diff
@@ -2548,6 +2547,8 @@ set(regression_disabled_tests
   ###
   regress1/bug472.smt2
   regress1/datatypes/non-simple-rec-set.smt2
+  # TODO: fix models (projects #276)
```
yoni206, line 2550: Can't find this project:
https://github.com/orgs/CVC4/projects/276

Just to make sure -- is there a specific issue that corresponds to this TODO? If not, can you create one in the corresponding project?

see guideline here: https://github.com/CVC4/CVC4/wiki/Code-Style-Guidelines#todofixme-comments

yoni206 (approved):
LGTM, please add an issue for the TODO (more details inline the review).

test/regress/CMakeLists.txt
ajreynol: https://github.com/CVC4/cvc4-projects/issues/276

PR #6437, opened 2021-04-23
Use substitutions for implementing defined functions

ajreynol:
This eliminates explicit tracking of defined functions, and instead makes define-fun add to preprocessing substitutions.
In other words, the effect of:
```
(define-fun f X t)
```
is to add `f -> (lambda X t)` to the set of substitutions known by the preprocessor.  This is essentially the same as when
`(= f (lambda X t))` was an equality solved by non-clausal simplification

The motivation for this change is both uniformity and for performance, as fewer traversals of the input formula.

After this PR, the expandDefinitions preprocessing step is replaced by `apply-substs`. 

Further work will reorganize the relationship between the expand definitions module and the rewriter, after which global calls to SmtEngine::expandDefinitions can be cleaned up.


src/smt/assertions.cpp
```diff
@@ -59,8 +61,8 @@ void Assertions::finishInit()
   {
     // In the case of incremental solving, we appear to need these to
     // ensure the relevant Nodes remain live.
-    d_assertionList = new (true) AssertionList(d_userContext);
-    d_globalDefineFunRecLemmas.reset(new std::vector<Node>());
+    d_assertionList = new (true) AssertionList(d_env.getUserContext());
```
nafur, line 64: What exactly does `new (true) ...` do and why do we use it here?

nafur (commented):
A question about new...

src/smt/assertions.cpp
ajreynol: I believe this has something to do with dynamically allocating a context-dependent object.

I'm not sure why we have this strange way of doing this, I can just update to a more normal way of doing this.

nafur (approved):
LGTM now, but I'd really like to have someone else have a look as well.

aniemetz:
I'll have a look tomorrow.

src/preprocessing/preprocessing_pass_context.h
```diff
@@ -97,25 +94,22 @@ class PreprocessingPassContext
...
   /** Pointer to the SmtEngine that this context was created in. */
   SmtEngine* d_smt;
-
-  /** Pointer to the ResourceManager for this context. */
-  ResourceManager* d_resourceManager;
-
-  /* The top level substitutions */
-  theory::TrustSubstitutionMap d_topLevelSubstitutions;
-
+  /** Reference to the env */
```
aniemetz, line 109:
```suggestion
  /** Reference to the environment. */
```

src/smt/assertions.h
```diff
@@ -144,21 +148,24 @@ class Assertions
                   bool inUnsatCore,
                   bool inInput,
                   bool isAssumption,
+                  bool isFunDef,
                   bool maybeHasFv);
-  /** pointer to the user context */
-  context::UserContext* d_userContext;
+  /** reference to the environment */
```
aniemetz, line 153:
```suggestion
  /** Reference to the environment. */
```

src/smt/env.h
```diff
@@ -84,6 +85,9 @@ class Env
   /** Get a pointer to the Rewriter owned by this Env. */
   theory::Rewriter* getRewriter();
 
+  /** Gets a reference to the top-level substitution map */
```
aniemetz, line 88:
```suggestion
  /** Get a reference to the top-level substitution map */
```

src/smt/expand_definitions.h
```diff
@@ -87,8 +81,8 @@ class ExpandDefs
       TConvProofGenerator* tpg);
   /** Reference to the SMT engine */
   SmtEngine& d_smt;
-  /** Reference to resource manager */
-  ResourceManager& d_resourceManager;
+  /** Reference to the env */
```
aniemetz, line 84:
```suggestion
  /** Reference to the environment. */
```

src/smt/proof_manager.h
```diff
@@ -83,29 +78,25 @@ class PfManager
    * The argument pfn is the proof for false in the current context.
    *
    * Throws an assertion failure if pg cannot provide a closed proof with
-   * respect to assertions in as and df. For the latter, entries in the defined
-   * function map correspond to equalities of the form (= f (lambda (...) t)),
-   * which are considered assertions in the final proof.
+   * respect to assertions in as. Note that as includes equalities of the form
```
aniemetz, line 81:
```suggestion
   * respect to assertions in as. Note this includes equalities of the form
```
?

src/theory/trust_substitutions.cpp
```diff
@@ -27,21 +27,31 @@ TrustSubstitutionMap::TrustSubstitutionMap(context::Context* c,
...
+      d_subsPg(nullptr),
+      d_applyPg(nullptr),
+      d_helperPf(nullptr),
       d_name(name),
       d_trustId(trustId),
       d_ids(ids),
       d_eqtIndex(c)
 {
+  setProofNodeManager(pnm);
+}
```
aniemetz, line 41:
```suggestion
}

```

aniemetz (approved):
Just some very minor things, else LGTM.

PR #6438, opened 2021-04-24
Ensure proofs track dependencies for trusted substitutions

ajreynol:
This ensures that dependencies are tracked for all inferred substitutions, even if a trusted step is added.

This is required to ensure unsat cores are sound when we use e.g. non-clausal simplification + unsat cores.


HanielB (approved):
LGTM

PR #6440, opened 2021-04-25
Use fast enumeration by default for Boolean predicate synthesis

ajreynol:
This updates the policy of when to apply smart enumeration: we do so if the grammar has ITE or admits Boolean connective terms.  Previously, we applied smart enumeration for ITE and all Boolean grammars.  However, this is misguided since e.g. partial evaluation unfolding has no opportunity to be effective if the enumerated terms are only Boolean literals.

This significantly improves run time on a challenge benchmark from @makaimann.

### DIFF
--- a/src/theory/quantifiers/sygus/type_info.h
+++ b/src/theory/quantifiers/sygus/type_info.h
@@ -82,6 +82,11 @@ class SygusTypeInfo
    */
   bool hasIte() const;
   /**
+   * Returns true if this sygus type has a constructor whose sygus operator is
+   * a Boolean connective.
+   */
+  bool hasBoolConnective() const;
+  /**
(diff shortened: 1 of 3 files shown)


src/theory/quantifiers/sygus/type_info.h
```diff
@@ -196,6 +201,11 @@ class SygusTypeInfo
    * or is a lambda whose body is ITE.
    */
   bool d_hasIte;
+  /**
+   * Whether this sygus type has a constructors whose sygus operator is a
```
yoni206, line 205:
```suggestion
   * Whether this sygus type has a constructor whose sygus operator is a
```

yoni206 (approved):

PR #6442, opened 2021-04-26
Protect IntStat methods

nafur:
This PR protects two methods of the IntStat class in case statistics are disabled.


nafur:
This should fix one of the issues with our nightlies.

mpreiner (approved):

PR #6446, opened 2021-04-27
Refactor resource manager options

nafur:
This PR refactors how the resource manager options are used. It moves options relevant for the resource manager into its own file (making the number of smt options a bit smaller) and uses the `Options` object directly instead of using the old static accessors.


src/options/resource_manager_options.toml
```diff
@@ -0,0 +1,57 @@
...
+# --rweight is used to override the default of one particular resource weight.
+# It can be given multiple times to override multiple weights. When options are
+# parsed, the resource manager might now be created yet, and it is not clear
+# how an option handler would access it in a reasonable way. The option handler
+# thus merely puts the data in another option that holds a vector of strings.
+# This other option "resourceWeightHolder" has the sole purpose of storing
+# this data in a way so that the resource manager can access it in its
+# constructor.
+[[option]]
+  smt_name   = "resourceWeight"
```
mpreiner, line 46: Shouldn't this be `name`?

nafur: As this option is never accessed via `options::<name>()` (irrespective of the fact that we will drop this syntax at some point), it does not need a name. It is only accessed indirectly via the `resourceWeightHolder` option.

mpreiner: Why do we have an `smt_name` then? `smt_name` is just the name of the option that should be used via the SMT-LIB front-end.

mpreiner (approved):

PR #6447, opened 2021-04-27
Simplify NodeManager mkFunctionType

ajreynol:
Previously, we were conditionally checking whether a function was "flat" e.g. did not have a function type as the range type.

The original motivation for this was to catch cases where the user made a declare-fun that had function return type, which is not permitted.  All these checks are already done at the API level https://github.com/CVC4/CVC4/blob/master/src/api/cpp/cvc5_checks.h#L441.

However, internally we should have no such restriction.  This is required for simplifying the LFSC term processor.


HanielB (approved):
LGTM

PR #6448, opened 2021-04-27
Fix refutational soundness bug in quantifier prenexing

ajreynol:
This bug can be triggered by define-fun, quantifier macros or inferred substitutions whose RHS contain quantified formulas.

This corrects the issue by ensuring that bound variables introduced for prenexing are fresh for distinct quantified formula subterms that may share quantified variables.

This was reported by Geoff Sutcliffe via a TPTP run.


HanielB (approved):
LGTM

PR #6449, opened 2021-04-27
Add support for datatype update

ajreynol:
This removes the special case of TUPLE_UPDATE and RECORD_UPDATE in favor of the more general datatype update.

Datatype update is handled analogously to APPLY_SELECTOR / APPLY_TESTER.

Review/merge https://github.com/CVC4/CVC4/pull/6450 first.


ajreynol:
I'm wondering if APPLY_DT_UPDATE -> APPLY_UPDATER, similarly using the name "Updater" instead of "DatatypeUpdater" in a few places.

My concern was that this is too general, however, thinking a bit more, "TESTER" / "SELECTOR" are already general.  The use of "UPDATER" would be more symmetric.  In some sense, datatypes are a generalization of the other types so using these names may be appropriate.  What do you think @yoni206 @aniemetz ?

yoni206:
I am in favor of having it consistent with the other datatypes components (selectors and testers), so +1

ajreynol:
@yoni206 I've updated the names to be more general, PTAL.

ajreynol:
@shazqadeer

yoni206 (approved):
LGTM, one comment inline to consider.

nblei:
@ajreynol any examples on how to use this?  Its a bit difficult for me to figure out since there does not appear to be an equivalent smtlib2 construction.  The documentation states it takes three parameters (1: Updater, 2: Datatype term, and 3: Value to update a field of the datatype term with), but it says to create with a function which only takes two terms as arguments.

However, two terms as arguments leads to a runtime exception:
```
Invalid kind 'APPLY_UPDATER', expected Terms with kind APPLY_UPDATER must have at least 3 children and at most 3 children (the one under construction has 2)
```

I don't understand what `Updater` is, and how this differs from the third argument (`value to update a field`).

PR #6450, opened 2021-04-27
Add internal support for datatype update

ajreynol:


src/expr/dtype_cons.cpp
```diff
@@ -565,14 +558,24 @@ bool DTypeConstructor::resolve(
       }
       Trace("datatypes-init")
           << "  ...range after parametric substitution " << range << std::endl;
-      arg->d_selector = sm->mkDummySkolem(
-          argName,
-          nm->mkSelectorType(self, range),
-          "is a selector",
-          NodeManager::SKOLEM_EXACT_NAME | NodeManager::SKOLEM_NO_NOTIFY);
     }
+    arg->d_selector = sm->mkDummySkolem(
```
yoni206, line 562: Can you document this block? e.g., why o we need a skolem?

src/expr/node_manager.h
```diff
@@ -34,6 +34,8 @@
 
 namespace cvc5 {
 
+using Record = std::vector<std::pair<std::string, TypeNode>>;
```
yoni206, line 37: can you document how this will be used?

yoni206 (approved):
LGTM. 2 minor comments inside. Also, before merging -- could you check if you could add a unit test? For example, that verifies that the type checker fails for ill-typed terms with the updated?

src/expr/node_manager.h
ajreynol: This will be moved from record.h in the other PR, I'll remove from this PR.

PR #6452, opened 2021-04-27
Fix BV Optimization Boundary Condition when lower bound = upper bound + 1

Ouyancheng:
If lb = ub + 1 then (lb+ub)/2 =pivot == lb since it's rounding to -infinity. 
and lb <= x < pivot will always be UNSAT. 
We need to handle this differently. 
And this only happens in minimization.


src/omt/bitvector_optimizer.cpp
```diff
@@ -33,20 +33,19 @@ BitVector OMTOptimizerBitVector::computeAverage(const BitVector& a,
   // computes (a + b) / 2 without overflow
   // rounding towards -infinity: -1.5 --> -2,  1.5 --> 1
   // average = (a / 2) + (b / 2) + (((a % 2) + (b % 2)) / 2)
+  // For unsigned bitvectors, this algorithm should be equivalent to:
+  // if (!isSigned) {
+  //   return BitVector(a.getSize(),
+  //     (a.toInteger() + b.toInteger()).floorDivideQuotient(Integer(2)));
+  // }
```
yoni206, line 40: Since this is a comment, I suggest using ordinary english/math. Otherwise it looks like commented out code.

yoni206 (commented):
Thanks for fixing this. I have one minor comment inline.
In addition -- it would be good to add a unit test that fails on master and passes with the fix.
Is that possible?

src/omt/bitvector_optimizer.cpp
Ouyancheng: not this time. the test will be added after everything is up.

Ouyancheng: and the comments could be removed.

Ouyancheng:
> Thanks for fixing this. I have one minor comment inline.
> In addition -- it would be good to add a unit test that fails on master and passes with the fix.
> Is that possible?

not this time. the test will be included after everything is up.

src/omt/bitvector_optimizer.cpp
Ouyancheng: this piece of code-like comment is removed. A narrative description is already included in the function.

mpreiner (changes requested):
I agree with @yoni206. This PR should already include a unit test to cover this bug (fails on master, passes with this PR). Further, the title of the PR and the description should describe the bug that has been fixed.

Both
```
I don't know why this bug escaped the last PR review...
But it's not too late to fix it now.
```
and
```
Fix BV Optimizer bug!
```
does not describe the underlying issue at all and is not useful to the reviewer to understand the bug.

aniemetz:
I agree, this PR should include a unit test to cover this bug.

Ouyancheng:
👆I can confirm the test I recently pushed fails on master as of now.

test/unit/theory/theory_bv_opt_white.cpp
```diff
@@ -151,5 +151,31 @@ TEST_F(TestTheoryWhiteBVOpt, signed_max)
             << std::endl;
 }
 
+TEST_F(TestTheoryWhiteBVOpt, min_boundary)
+{
+  Node x = d_nodeManager->mkVar(*d_BV32Type);
+  // do not remove
```
yoni206, line 157: Is this needed? If not, please remove. Otherwise, make the comment clearer.

test/unit/theory/theory_bv_opt_white.cpp
```diff
@@ -151,5 +151,31 @@ TEST_F(TestTheoryWhiteBVOpt, signed_max)
...
+  Node y = d_nodeManager->mkVar(*d_BV32Type);
+
+  // x <= 18
+  d_smtEngine->assertFormula(d_nodeManager->mkNode(
+      kind::BITVECTOR_ULE, d_nodeManager->mkConst(BitVector(32u, 18u)), x));
+  // do not remove, this perturbs the solver the trigger some boundary bug
+  // exists previously
```
yoni206, line 164:
```suggestion
  // this perturbs the solver to trigger some boundary bug
  // that existed previously
```

yoni206 (commented):
Thanks for adding the test and for verifying that it fails on master.
I have two comments, described inline the review.

test/unit/theory/theory_bv_opt_white.cpp
Ouyancheng: It's described in the following statement. just don't remove it.

mpreiner: This comment is not necessary. If the statement would be removed it would result in a compile error. The main point here is that line 165 should not be removed/disabled since this will trigger the described bug.

Ouyancheng: you mean the comment? ok I gonna remove

test/unit/theory/theory_bv_opt_white.cpp
```diff
@@ -151,5 +151,31 @@ TEST_F(TestTheoryWhiteBVOpt, signed_max)
             << std::endl;
 }
 
+TEST_F(TestTheoryWhiteBVOpt, min_boundary)
+{
+  Node x = d_nodeManager->mkVar(*d_BV32Type);
+  // this variable is important on triggering the bug
```
yoni206, line 157:
```suggestion
```
Having this comment here is confusing, because it is not clear what bug you are referring to.  Please remove the comment -- it is clear from lines 163--164 that this test was added due to a bugfix, and it is clear from line 165 that the variable `y` is important, since it is used there.

Ouyancheng: ... removed

yoni206 (approved):
Thank you for the changes.

Ouyancheng:
> I agree with @yoni206. This PR should already include a unit test to cover this bug (fails on master, passes with this PR). Further, the title of the PR and the description should describe the bug that has been fixed.
> 
> Both
> 
> ```
> I don't know why this bug escaped the last PR review...
> But it's not too late to fix it now.
> ```
> 
> and
> 
> ```
> Fix BV Optimizer bug!
> ```
> 
> does not describe the underlying issue at all and is not useful to the reviewer to understand the bug.

fixed

mpreiner (approved):
LGTM, thanks for the changes.

mpreiner:
>  Err... not sure why the PR is still showing changes requested. It's all resolved.

Reviewers who requested changes still have to approve the changes. This does not happen automatically ;-)

Ouyancheng:
> > Err... not sure why the PR is still showing changes requested. It's all resolved.
> 
> Reviewers who requested changes still have to approve the changes. This does not happen automatically ;-)

Ah... that makes sense. Thanks!

PR #6457, opened 2021-04-28
Make sure reference stats are reset properly

nafur:
This PR adds a `reset()` method to the `ReferenceStat` class. It then uses it to properly reset such statistics in the minisat solvers where lifetime is an issue.

### DIFF
--- a/src/prop/bvminisat/bvminisat.cpp
+++ b/src/prop/bvminisat/bvminisat.cpp
@@ -38,9 +38,7 @@ BVMinisatSatSolver::BVMinisatSatSolver(StatisticsRegistry& registry,
   d_statistics.init(d_minisat.get());
 }
 
-
-BVMinisatSatSolver::~BVMinisatSatSolver() {
-}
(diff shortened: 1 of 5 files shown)


nafur:
This addresses the asan issues in our asan nightlies.

src/prop/bvminisat/bvminisat.cpp
```diff
@@ -278,5 +276,24 @@ void BVMinisatSatSolver::Statistics::init(BVMinisat::SimpSolver* minisat){
   d_statEliminatedVars.set(minisat->eliminated_vars);
 }
 
+void BVMinisatSatSolver::Statistics::deinit()
```
mpreiner, line 279: Why a `deinit()` method instead of doing the reset in `~Statistics`?

nafur: I found it way easier to do this manually instead of relying in destruction order, in particular in this situation. This (or rather something equivalent) is done in `~Statistics` anyway. The issue here is that the member (d_statistics) is destructed after the manual destructor has run, which destroys d_minisat manually.

mpreiner (approved):

PR #6459, opened 2021-04-29
Refactor optimization result and objective classes + add preliminary support for multiple objectives

Ouyancheng:
This PR is one part of another PR which is considered complicated to review. more will come later. 

**Preliminary support** means currently only supports single objective. 
But it supports queue-ing up objectives and it always optimize the first. 
Multiple-obj optimization algorithm will be up next. 

1. Refactor optimization result and objective classes 
2. Add preliminary support for multiple objectives
3. The unit tests are now comparing node values instead of node addresses

Sorry for being lengthy, it's not very possible to further break this PR into pieces. 
Hopefully this is easy enough for reviewing. 
The diff view is clean enough.


src/smt/optimization_solver.h
```diff
@@ -72,38 +93,58 @@ enum class OptResult
...
-   * @param n the optimization target node
+   * @param target the optimization target node
    * @param type speficies whether it's maximize/minimize
    * @param bvSigned specifies whether it's using signed or unsigned comparison
    *    for BitVectors this parameter is only valid when the type of target node
    *    is BitVector
    **/
-  Objective(Node n, ObjectiveType type, bool bvSigned = false);
-  ~Objective(){};
+  OptimizationObjective(Node target, ObjectiveType type, bool bvSigned = false)
```
Ouyancheng, line 117: These short function are inlined according to this: https://github.com/CVC4/CVC4/wiki/Code-Style-Guidelines#inline-functions
They are short enough to justify the inline.

src/smt/optimization_solver.h
```diff
@@ -122,32 +163,49 @@ class OptimizationSolver
...
    * @param bvSigned specifies whether we should use signed/unsigned
    *   comparison for BitVectors (only effective for BitVectors)
    *   and its default is false
    **/
-  void activateObj(const Node& obj,
-                   const ObjectiveType type,
-                   bool bvSigned = false);
-  /** Gets the value of the optimized objective after checkopt is called **/
-  Node objectiveGetValue();
+  void pushObjective(const Node target,
```
yoni206, line 184: can't this still be a reference?

src/smt/optimization_solver.h
```diff
@@ -122,32 +163,49 @@ class OptimizationSolver
...
 
  private:
   /** The parent SMT engine **/
   SmtEngine* d_parent;
+
   /** The objectives to optimize for **/
-  Objective d_activatedObjective;
-  /** A saved value of the objective from the last sat call. **/
-  Node d_savedValue;
+  std::vector<OptimizationObjective> d_objectives;
```
yoni206, line 205: Please clarify in the documentation of this field how are multiple objectives handled -- are they computed in the order they are presented in the vector? Are they independent of each other?

src/smt/optimization_solver.cpp
```diff
@@ -16,78 +16,58 @@
...
-OptimizationSolver::~OptimizationSolver() {}
-
-OptResult OptimizationSolver::checkOpt()
-{
-  // Make sure that the objective is not the default one
-  Assert(d_activatedObjective.getType() != ObjectiveType::OBJECTIVE_UNDEFINED);
-  Assert(!d_activatedObjective.getNode().isNull());
-
+  Assert(d_objectives.size() > 0);
+  // just a single objective optimization
```
yoni206, line 29: From this comment and from the code it seems like only single objective optimization is currently supported. Is that right? If so, please add assertions in the code that will not allow to create more than one objective (e.g., here assert that the objective size is 1).

src/smt/optimization_solver.h
```diff
@@ -122,32 +163,49 @@ class OptimizationSolver
...
    **/
-  OptimizationSolver(SmtEngine* parent);
-  ~OptimizationSolver();
+  OptimizationSolver(SmtEngine* parent)
+      : d_parent(parent), d_objectives(), d_results()
+  {
+  }
+  ~OptimizationSolver() = default;
+
+  /** Runs the optimization loop for the pushed objective **/
```
yoni206, line 172: Please specify in this comment whether it is the last pushed objective, or that you are assuming that only a single objective was pushed (in that case assert this in the implementation).

yoni206 (commented):
Thanks for the refactoring. Some comments and questions inline the review.

src/smt/optimization_solver.h
Ouyancheng: Node is already a reference to NodeValue. Using a reference of Node which is already a reference is redundant.

mpreiner: No, it's not redudant since Node is essentially a shared_ptr to NodeValue and you don't want to increment the reference count each time `pushObjective` is called. Instead of using Node use either a `const Node&` or a `const TNode`.

src/smt/optimization_solver.h
Ouyancheng: Currently it only supports single obj.
and this will change as I push more PRs. 
there will be a parameter later specifying how multigoal is handled. But not this time. 

You know I am breaking a complete PR into chunks.

src/smt/optimization_solver.cpp
Ouyancheng: YES. and this is *preliminary* support for multiple objective. 
If this is approved optimization algorithms will be up next.

src/smt/optimization_solver.h
Ouyancheng: the behavior currently is to always optimize the first one. 
I don't want to add this to comments because multiobj algorithms are coming right after so that I don't need to change all this comments.

src/smt/optimization_solver.cpp
Ouyancheng: It allows, just picks the first one and optimize with others no effect.

Ouyancheng: and multiobj algorithms will be up right next.

src/smt/optimization_solver.h
Ouyancheng: The parameter Node will be deallocated by RAII when the function is returned.

Ouyancheng: and storing something in a vector needs strong reference.

src/smt/optimization_solver.h
```diff
@@ -30,40 +29,62 @@ class SmtEngine;
...
+   *   if getType() is not OPTIMAL, it might return an empty node or a node
+   *   containing non-optimal value
+   **/
+  Node getValue() { return d_value; }
 
-  // if the solver halted early and value is only approximate
-  OPT_SAT_APPROX
+ private:
+  ResultType d_type;
+  Node d_value;
```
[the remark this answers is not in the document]
Ouyancheng: done

src/smt/optimization_solver.h
4tXJ7f: > The parameter Node will be deallocated by RAII when the function is returned.

That is correct but orthogonal to what @yoni206 and @mpreiner's point. The point is that creating/destroying the `Node` instance for the parameter is not necessary in the first place. The creation (resp. destruction) of nodes is not free: Each time we do that, we have to increment (resp. decrement) the reference count. Even if you end up storing the parameter in a vector, you still avoid unnecessary work by using a reference (see [this example](https://godbolt.org/z/qMojeP787)).

Ouyancheng: @4tXJ7f Right! I'm porting all of them to TNode now, and the reason is performance-wise you said. Wait a sec for re-compilation on my side. 
To add to your point: The excessive performance of increasing the rc comes from the rc mutex I believe.

Ouyancheng: @4tXJ7f oh, Node is not using shared_ptr internally and it's currently NOT thread-safe. 
Increasing the rc is simply increasing an integer in the current implementation.

Ouyancheng: Ported to TNode for all function parameters (including constructors). Now this issue should be resolved.

src/smt/optimization_solver.h
aniemetz: If you assume it is single you should assert this and comment on this. No matter if a follow up PR will change this behavior, this is what goes into master with this PR, so this should be properly asserted and documented.

src/smt/optimization_solver.h
aniemetz: > @4tXJ7f oh, Node is not using shared_ptr internally and it's currently NOT thread-safe.
That's why @mpreiner wrote "essentially a shared_ptr". Does not change the fact that we do reference counting (albeit "manually" and not via a shared_ptr), which adds overhead.

Ouyancheng: @aniemetz Logical! 
In the current implementation I suspect the only "overhead" in Node's rc is the addition of a few inlined integer instructions... it's minimal and definitely not the bottleneck. 
Whereas std::share_ptr's overhead mainly comes from its mutex.

src/smt/optimization_solver.h
Ouyancheng: detailed in comments

src/smt/optimization_solver.h
```diff
@@ -169,20 +174,23 @@ class OptimizationSolver
   }
   ~OptimizationSolver() = default;
 
-  /** Runs the optimization loop for the pushed objective **/
+  /** Runs the optimization loop for the pushed objective 
+   * FIXME: this function currently supports only single objective 
+   * for multiple pushed objectives it always optimizes the first one. 
+   * Add support for multi-obj later **/
```
yoni206, line 180: Thanks for adding this comment.
If you use FIXME, then an issue should be created:
https://github.com/CVC4/CVC4/wiki/Code-Style-Guidelines#todofixme-comments

In this particular case I don't think all of this is needed, so please just remove the `FIXME:` prefix.

src/smt/optimization_solver.cpp
yoni206: Thanks for adding the comment. Please also add an assertion
`Assert(d_objectives.size() == 1);` and add a comment next to it saying we currently process a single objective. When you change the code so that it processes more objectives, simply remove the assertion.

I saw that you mentioned that this is done: https://github.com/CVC4/CVC4/pull/6459#discussion_r624153355

But I can't find the assertion itself.

yoni206 (commented):
Thank you for the changes. Two more comments regarding single/multiple objectives.

src/smt/optimization_solver.h
Ouyancheng: I can fix this. 
But notice that there's a FIXME like this in NodeValue::inc() in node_value.h . 
You might want to fix that as well.

Ouyancheng: I think a guideline like this is basically preventing people from adding FIXME or TODO, even though they should be. 
This is not a good practice. and I saw FIXMEs in NodeTemplate that are not referencing anything, and I just learned from that. 
Anyways, I will remove this.

Ouyancheng: @barrettcw

Ouyancheng: Fixed

src/smt/optimization_solver.cpp
Ouyancheng: Oh I forgot, sorry. Now it's fixed.

yoni206 (approved):

src/smt/optimization_solver.h
Ouyancheng: The original google guideline is not that restrictive: 
https://google.github.io/styleguide/cppguide.html#TODO_Comments 

> TODOs should include the string TODO in all caps, followed by the name, e-mail address, bug ID, or other identifier of the person or issue with the best context about the problem referenced by the TODO.

mpreiner: The FIXMEs without an associated issue are very old and the guidelines were introduced afterwards. So all FIXME since we introduced the dev guidelines have associated issues. The reason why we made the guidelines more restrictive is to avoid the problem of orphaned FIXMEs/TODOs.

Ouyancheng:
> Resolve comment regarding FIXME.

it's already resolved.

src/smt/optimization_solver.h
```diff
@@ -122,32 +168,52 @@ class OptimizationSolver
...
    **/
-  OptimizationSolver(SmtEngine* parent);
-  ~OptimizationSolver();
+  OptimizationSolver(SmtEngine* parent)
+      : d_parent(parent), d_objectives(), d_results()
+  {
+  }
+  ~OptimizationSolver() = default;
+
+  /** Runs the optimization loop for the pushed objective
```
aniemetz, line 177:
```suggestion
  /**
   * Run the optimization loop for the pushed objective 
```

src/smt/optimization_solver.cpp
```diff
@@ -16,78 +16,56 @@
...
 
-  std::pair<OptResult, Node> optResult;
-  if (d_activatedObjective.getType() == ObjectiveType::OBJECTIVE_MAXIMIZE)
+  OptimizationResult optResult;
+  if (d_objectives[0].getType() == OptimizationObjective::MAXIMIZE)
   {
-    optResult = optimizer->maximize(this->d_parent,
-                                    this->d_activatedObjective.getNode());
+    optResult =
+        optimizer->maximize(this->d_parent, this->d_objectives[0].getTarget());
```
aniemetz, line 39:
```suggestion
        optimizer->maximize(this->d_parent, d_objectives[0].getTarget());
```

src/smt/optimization_solver.cpp
```diff
@@ -16,78 +16,56 @@
...
   }
-  else if (d_activatedObjective.getType() == ObjectiveType::OBJECTIVE_MINIMIZE)
+  else if (d_objectives[0].getType() == OptimizationObjective::MINIMIZE)
   {
-    optResult = optimizer->minimize(this->d_parent,
-                                    this->d_activatedObjective.getNode());
+    optResult =
+        optimizer->minimize(this->d_parent, this->d_objectives[0].getTarget());
```
aniemetz, line 44:
```suggestion
        optimizer->minimize(d_parent, d_objectives[0].getTarget());
```

src/smt/optimization_solver.h
```diff
@@ -122,32 +168,52 @@ class OptimizationSolver
...
+   * for multiple pushed objectives it always optimizes the first one. 
+   * Add support for multi-obj later **/
+  OptimizationResult::ResultType checkOpt();
 
-  /** Runs the optimization loop for the activated objective **/
-  OptResult checkOpt();
   /**
-   * Activates an objective: will be optimized for
-   * @param obj the Node representing the expression that will be optimized for
+   * Pushes an objective
```
aniemetz, line 184:
```suggestion
   * Push an objective.
```

test/unit/theory/theory_bv_opt_white.cpp
```diff
@@ -79,22 +75,18 @@ TEST_F(TestTheoryWhiteBVOpt, signed_min)
...
+  BitVector val = d_optslv->getValues()[0].getValue().getConst<BitVector>();
   std::cout << "opt value is: " << val << std::endl;
 
   // expect the minimum x = -1
-  ASSERT_EQ(d_optslv->objectiveGetValue(),
-            d_nodeManager->mkConst(BitVector(32u, (unsigned)0x80000000)));
-  std::cout << "Passed!" << std::endl;
-  std::cout << "Optimized value is: " << d_optslv->objectiveGetValue()
-            << std::endl;
+  ASSERT_EQ(val, BitVector(32u, (unsigned)0x80000000));
```
aniemetz, line 88:
```suggestion
  ASSERT_EQ(val, BitVector(32u, (uint32_t)0x80000000));
```

test/unit/theory/theory_bv_opt_white.cpp
```diff
@@ -110,20 +102,18 @@ TEST_F(TestTheoryWhiteBVOpt, unsigned_max)
...
-  BitVector val = d_optslv->objectiveGetValue().getConst<BitVector>();
+  BitVector val = d_optslv->getValues()[0].getValue().getConst<BitVector>();
   std::cout << "opt value is: " << val << std::endl;
 
-  ASSERT_EQ(d_optslv->objectiveGetValue(),
-            d_nodeManager->mkConst(BitVector(32u, (unsigned)2)));
-  std::cout << "Optimized value is: " << d_optslv->objectiveGetValue()
-            << std::endl;
+  ASSERT_EQ(d_optslv->getValues()[0].getValue().getConst<BitVector>(),
+            BitVector(32u, (unsigned)2));
```
aniemetz, line 115:
```suggestion
            BitVector(32u, 2u));
```

aniemetz (changes requested):
Some minor things.

PR #6460, opened 2021-04-29
Python API tests for terms -- Part 1

yoni206:
This PR removes the old python api tests for terms from `test/api/python/test_term.py` and replaces it with a new test file `test/python/unit/api/test_term.py`. The new test file is obtained by translating `test/unit/api/term_black.cpp`.

In this PR I only include the tests that pass without requiring any change to the python API. The next PR will add functions to the python API that are currently not supported, along with corresponding tests.


test/python/unit/api/test_term.py
```diff
@@ -0,0 +1,789 @@
+###############################################################################
+# Top contributors (to current version):
+#   Makai Mann, Andres Noetzli
```
makaimann, line 3: You should put yourself here (at least first) :)

test/python/unit/api/test_term.py
```diff
@@ -0,0 +1,789 @@
...
+  z = Term(solver)
+
+  assert x == x
+  assert not x != x
+  assert not x == y
+  assert x != y
+  assert not (x == z)
+  assert x != z
+
+def test_getKind(solver):
```
makaimann, line 42: The other new tests all used underscores and lowercase (https://github.com/CVC4/CVC4/blob/8431e9d49b71774092ca29c85855cbdf5bf09c53/test/python/unit/api/test_sort.py#L64), but these use a combination of underscores and camel case. Either is fine from my point of view, but they should probably be the same.

test/python/unit/api/test_term.py
```diff
@@ -0,0 +1,789 @@
...
+##
+
+
+
+import pytest
+import pycvc5
+from pycvc5 import kinds
+from pycvc5 import Sort, Term
+
+def assert_eq(x, y):
```
makaimann, line 21: Is this just here to match the C++ tests better? I'm not a huge fan of it tbh, because it doesn't add anything but now you can't pass a message like `assert x == y, "Something important"`. Furthermore, then people have to remember to use it (see line 35 where it's not used).

makaimann (approved):
This is great, thanks for doing all this! Left a few nitpicky comments that you can address if you'd like but otherwise looks good to me.

yoni206:
Thanks @makaimann , I am putting a do not merge to make sure I address your comments.

yoni206:
Closing because I accidentally opened this branch on the main repo. Will create another PR from my fork. Sorry!

PR #6468, opened 2021-04-30
Python API tests for terms -- Part 1

yoni206:
This PR removes the old python api tests for terms from `test/api/python/test_term.py` and replaces it with a new test file `test/python/unit/api/test_term.py`. The new test file is obtained by translating `test/unit/api/term_black.cpp`.

In this PR I only include the tests that pass without requiring any change to the python API. The next PR will add functions to the python API that are currently not supported, along with corresponding tests.

Comment: This was originally done in #6460 on the wrong fork. Now it is re-opened, and addresses all comments given there.


makaimann (approved):
LGTM, thanks for working on this!

PR #6472, opened 2021-05-01
FP: Rewrite to_fp conversion from signed bit-vector.

aniemetz:
SymFPU does not allow to_fp conversion from signed bv of size 1. This
adds rewrites for this case.

Rewrites for the constant and the non-constant cases were tested in
isolation.


src/theory/fp/theory_fp_rewriter.cpp
```diff
@@ -753,32 +778,43 @@ RewriteResponse maxTotal(TNode node, bool isPreRewrite)
...
-    BitVector arg(node[1].getConst<BitVector>());
+    BitVector sbv(node[1].getConst<BitVector>());
 
-    FloatingPoint res(param.getSize(), rm, arg, true);
+    NodeManager* nm = NodeManager::currentNM();
 
-    Node lit = NodeManager::currentNM()->mkConst(res);
-    
-    return RewriteResponse(REWRITE_DONE, lit);
+    if (sbv.getSize() == 1)
```
ajreynol, line 789: Can you give a quick comment here why this is a special case?

ajreynol (approved):
One minor comment, else LGTM.

src/theory/fp/theory_fp_rewriter.cpp
aniemetz: It is not. It's just that @martin-cs assumed that it would not make sense to support it. A bit-vector `1` represents value `-1` of size `1`. Nothing special about it.

aniemetz: I can add a comment that this is due to SymFPU not supporting it.

PR #6473, opened 2021-05-01
Add helper functions for multi-objective optimization + refactoring

Ouyancheng:
- add 3 helper functions 
1. judge whether a node is optimizable
2. make strong improvement expression according to optimization objective
3. make weak improvement expression according to optimization objective
- optChecker is now created by optimizationSolver instead of the minimize/maximize functions 
- Slightly refactors function signatures so that they are accepting OptimizationObjective instead of accepting target, type in separate parameters.


src/omt/integer_optimizer.cpp
```diff
@@ -74,17 +71,15 @@ OptimizationResult OMTOptimizerInteger::optimize(
...
+  return this->optimize(optChecker, target, true);
 }
-OptimizationResult OMTOptimizerInteger::maximize(
-    SmtEngine* parentSMTSolver, TNode target)
+OptimizationResult OMTOptimizerInteger::maximize(SmtEngine* optChecker,
+                                                 TNode target)
 {
-  return this->optimize(
-      parentSMTSolver, target, OptimizationObjective::MAXIMIZE);
+  return this->optimize(optChecker, target, false);
```
yoni206, line 82: In case you haven't, please make sure this (and the rest of the changes) are clang-formatted.

src/omt/omt_optimizer.h
```diff
@@ -27,54 +27,95 @@ class OMTOptimizer
...
-   * for BitVectors and it's only valid when the type is BitVector
-   *
-   * @param targetNode the target node for the expression to be optimized
-   * @param isSigned speficies whether to use signed comparison for BitVectors
-   *   and it's only valid when the type of targetNode is BitVector
+   * Returns whether node supports optimization
+   * @param node the target node to check for optimizability
+   * @return whether node supports optimization
+   **/
+  static bool nodeSupportsOptimization(TNode node);
```
yoni206, line 36: Please mention here which types are currently supported.

src/omt/omt_optimizer.cpp
```diff
@@ -17,20 +17,22 @@
...
 using namespace cvc5::theory;
 using namespace cvc5::smt;
 namespace cvc5::omt {
 
-std::unique_ptr<OMTOptimizer> OMTOptimizer::getOptimizerForNode(TNode targetNode,
-                                                                bool isSigned)
+bool OMTOptimizer::nodeSupportsOptimization(TNode node)
+{
+  TypeNode type = node.getType();
+  return (type.isInteger() || type.isBitVector());
```
yoni206, line 28: I see some code that treats also Real and FP, so it is not clear why they are not considered in this line. Please make this clearer in the comments in the code.

src/omt/omt_optimizer.cpp
```diff
@@ -39,32 +41,167 @@ std::unique_ptr<OMTOptimizer> OMTOptimizer::getOptimizerForNode(TNode targetNode
...
+      }
+      else
+      {
+        CVC5_FATAL() << "Target type does not support optimization";
+      }
+    }
+      // there's no default case
+      // because if more cases in OptimizationObjective::ObjectiveType are added
+      // compiler will issue a warning here in compile time instead of pushing
+      // it to run time.
```
yoni206, line 126: Please make this assumption explicit by adding a default case with an assertion of false. Compiler warnings are sometimes missed, and it is better to be explicit with assertion regarding what you are assuming.

yoni206 (changes requested):
Thanks for this refactoring. Some comments inline.

src/omt/omt_optimizer.cpp
Ouyancheng: Ok, but the default case is falling through to the end, which is an unreachable statement as desired.

src/omt/omt_optimizer.cpp
Ouyancheng: they are not supported. the other code adds Real and FP but they really are not supported for optimization.

src/omt/omt_optimizer.h
Ouyancheng: k

src/omt/integer_optimizer.cpp
Ouyancheng: didn't I? will do now

Ouyancheng: they are formatted

src/omt/omt_optimizer.cpp
Ouyancheng: added

src/omt/omt_optimizer.cpp
```diff
@@ -67,6 +67,7 @@ Node OMTOptimizer::mkStrongIncrementalExpression(
     {
       if (targetType.isInteger() || targetType.isReal())
       {
+        Assert(!targetType.isReal());  // remove this if real support is added
```
yoni206, line 70: Thanks for this assertion, but i think that now it is very confusing because you have two assertions one after the other and the first subsumes the second. 

Since right now Reals are not supported, please just remove the Reals, like you do for FP.

It will be very easy to add these cases later, but for now including them only creates confusion.

Ouyancheng: formatted

Ouyancheng:
👆NOTE: f is for "formatted" in the last commit message 😂......

src/omt/omt_optimizer.cpp
```diff
@@ -39,32 +42,143 @@ std::unique_ptr<OMTOptimizer> OMTOptimizer::getOptimizerForNode(TNode targetNode
...
+  TypeNode targetType = objective.getTarget().getType();
+  switch (objective.getType())
+  {
+    case OptimizationObjective::MINIMIZE:
+    {
+      if (targetType.isInteger())
+      {
+        Assert(lhs.getType().isInteger() || lhs.getType().isReal())
+            << lhsTypeError;
+        Assert(rhs.getType().isInteger() || rhs.getType().isReal())
```
yoni206, line 72: Please remove the `isReal()` from these places as well (similarly for the similar places  below e.g. for maximize).

src/omt/omt_optimizer.cpp
```diff
@@ -39,32 +42,143 @@ std::unique_ptr<OMTOptimizer> OMTOptimizer::getOptimizerForNode(TNode targetNode
...
+      {
+        Assert(lhs.getType() == targetType) << lhsTypeError;
+        Assert(rhs.getType() == targetType) << rhsTypeError;
+        return (objective.bvIsSigned())
+                   ? (nm->mkNode(Kind::BITVECTOR_SLT, lhs, rhs))
+                   : (nm->mkNode(Kind::BITVECTOR_ULT, lhs, rhs));
+      }
+      else
+      {
+        CVC5_FATAL() << "Target type does not support optimization";
```
yoni206, line 86: All `CVC5_FATAL` should be replaced with `Unreachable` because it is used in a code that is supposed to be unreachable (this is also guarded by assertions above). 

So please use Unreachable instead for all occurrences. The error message is very useful though, so please still use it.

yoni206 (commented):
Thanks for the changes. There are still a few redundant usages of reals. Details about this and another comment are inline.

src/omt/omt_optimizer.cpp
Ouyancheng: Unreachable doesn't tell you the reasons. 
and Unreachable() is CVC5_FATAL() << "Unreachable code reached"...

src/omt/omt_optimizer.cpp
Ouyancheng: this is desired, because Real numbers are eligible to use LT and GT and they could be compared against integers. 
Not restricted to Integer.

src/omt/omt_optimizer.cpp
yoni206: Right, that's why I ask that you keep the error message so that we see both the reason and that it is unreachable.

Ouyancheng: I wouldn't say it's unreachable, because it could be reached logically if you pass other types to it. 
just a CVC5_FATAL() with an error message is more desired. 
Unreachable are used to tag really unreachable code AFAIK.

src/omt/omt_optimizer.cpp
yoni206: but in the current code, is there any benchmark that will make `lhs.getType()` to be a real? Aren't we guarding to get only bit-vectors and integers?

src/omt/omt_optimizer.cpp
mpreiner: How about `Unimplemented()` then?

src/omt/omt_optimizer.cpp
Ouyancheng: The function is supposed to make an incremental expression if the type is same as obj.getTarget() or is implicitly convertible to that... So it's desired behavior. 
test cases doesn't reflect this.

src/omt/omt_optimizer.cpp
Ouyancheng: I don't plan to support FLOATINGPOINT TBH...

src/omt/omt_optimizer.cpp
yoni206: If the module currently only supports integers and bit-vectors, then handling reals is confusing and redundant. If it is not redundant then please add a unit test that makes `rhs.getType().isReal()` so that this case is properly tested. Otherwise, please remove.

yoni206 (approved):

PR #6477, opened 2021-05-03
Adding functions to the python API and testing them

yoni206:
This PR adds some functions that are missing in the python API, along with unit tests for them.
Subsequent PR will include additional missing functions.

Also includes a `yapf` run to reformat the test file.

Co-authored-by: Makai Mann <makaim@stanford.edu>


src/api/python/cvc5.pxd
```diff
@@ -338,9 +340,11 @@ cdef extern from "api/cpp/cvc5.h" namespace "cvc5::api":
         bint operator==(const Term&) except +
         bint operator!=(const Term&) except +
         Term operator[](size_t idx) except +
+        uint64_t getId() except +
         Kind getKind() except +
         Sort getSort() except +
-        Term substitute(const vector[Term] es, const vector[Term] & reps) except +
+        Term substitute(const Term& e, const Term& replacement) except +
```
makaimann, line 346: Do you really need this in the Python API? You can't overload functions in Python, so you could always just create lists with one element in each. You could even do this internally in the function and handle the case where the user passes single values so it behaves the same in the interface. It's fine to keep it, but I'm just saying it might not be necessary.

makaimann (approved):
Looks good to me, thanks! Only left a very minor comment.

src/api/python/cvc5.pxd
yoni206: Thanks for the feedback!
I was trying to follow a pattern I saw in the existing API, where there are two declarations similar to the c++ API and a single implementation that checks the type:

For example:
https://github.com/CVC4/CVC4/blob/95e763bc9f962d29a9b450fd46507c754dd655a8/src/api/python/cvc5.pxd#L146
https://github.com/CVC4/CVC4/blob/95e763bc9f962d29a9b450fd46507c754dd655a8/src/api/python/cvc5.pxi#L506

Is there an example in the python API for the style you are proposing? I'll be happy to follow that one instead.

makaimann: Got it, makes sense. Looks like there's a mix unfortunately. Here's the other pattern:
https://github.com/CVC4/CVC4/blob/95e763bc9f962d29a9b450fd46507c754dd655a8/src/api/python/cvc5.pxd#L158
https://github.com/CVC4/CVC4/blob/95e763bc9f962d29a9b450fd46507c754dd655a8/src/api/python/cvc5.pxi#L628

Which do you think is better? I'm not sure how much overhead there is in creating the vector and whatnot. I was figuring that if you have to declare the vector for the possibility of list inputs, then you might as well use it in either case. I can volunteer to unify these when we pick one pattern.

makaimann: @mpreiner wondering if you have any thoughts on the best pattern here? You can see the comments above but the short summary is: should we declare overloaded functions from C++ and call the matching one from Python, or should we use a single C++ function. In particular this is for functions that can take single argument(s), or vector(s). One option is to always use the vector version at the Cython -> C++ interface.

mpreiner: I don't see any harm in using the overloaded functions from the C++ API if it makes life easier on the Python side, but in this case we can just use the more general version with vectors.

yoni206: Thanks, @mpreiner !
@makaimann the code is now updated accordingly. Please let me know if you have more comments. If not, feel free to remove the `do not merge` label.

makaimann: Thanks!

PR #6482, opened 2021-05-04
Improve generation of python API documentation

nafur:
This PR makes generating the python part of our API documentation more reliable. If python bindings are enabled, it makes the `docs` target depend on `pycvc5` and renders a warning into the documentation if python bindings are disabled.
To make the dependency on `pycvc5` work properly, it changes the python bindings build so that the library is directly put in the right place and avoid the manual rename, so that the `pycvc5` target actually points to the correct file.
Unfortunately, this means we need to build `libcvc5` when building `docs` with python bindings enabled.

### DIFF
--- a/docs/python/CMakeLists.txt
+++ b/docs/python/CMakeLists.txt
@@ -1,0 +1,21 @@
+###############################################################################
+# Top contributors (to current version):
+#   Gereon Kremer
+#
+# This file is part of the cvc5 project.
+#
+# Copyright (c) 2009-2021 by the authors listed in the file AUTHORS
+# in the top-level source directory and their institutional affiliations.
+# All rights reserved.  See the file COPYING in the top-level source
+# directory for licensing information.
+# #############################################################################
+#
+# Build system configuration for python API documentation.
+##
+
+add_custom_target(docs-python)
+
+if (BUILD_BINDINGS_PYTHON)
+    # Python API docs are generated from built python API
+    add_dependencies(docs-python pycvc5)
+endif()
--- a/docs/python/python.rst
+++ b/docs/python/python.rst
@@ -5,7 +5,7 @@ Python API Documentation
 
     .. warning::
 
-        This documentation was build while python bindings were disabled. Most probably, this part of the documentation is either empty or outdated. Please enable :code:`BUILD_BINDINGS_PYTHON` in :code:`cmake` and build the documentation again.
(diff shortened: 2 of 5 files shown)


nafur:
The warning looks like this: 
image

docs/python/python.rst
```diff
@@ -1,6 +1,12 @@
 Python API Documentation
 ========================
 
+.. only:: not bindings_python
+
+    .. warning::
+
+        This documentation was build while python bindings were disabled. Most probably, this part of the documentation is either empty or outdated. Please enable :code:`BUILD_BINDINGS_PYTHON` in :code:`cmake` and build the documentation again.
```
makaimann, line 8:
```suggestion
        This documentation was built while python bindings were disabled. This part of the documentation is likely either empty or outdated. Please enable :code:`BUILD_BINDINGS_PYTHON` in :code:`cmake` and build the documentation again.
```

makaimann: Would it be outdated or would it always just be empty?

docs/python/CMakeLists.txt
```diff
@@ -0,0 +1,21 @@
...
+# #############################################################################
+#
+# Build system configuration for python API documentation.
+##
+
+add_custom_target(docs-python)
+
+if (BUILD_BINDINGS_PYTHON)
+    # Python API docs are generated from built python API
+    add_dependencies(docs-python pycvc5)
```
makaimann, line 20: Just for my own understanding: is the documentation actually generated from the `pycvc5` python module itself? As in it exercises the python bindings to get the documentation? Or is it just that the `pycvc5` documentation is generated within the `pycvc5` target in CMake? If this question doesn't make any sense, no worries about answering -- I was just curious.

makaimann (approved):
LGTM, nice to have a warning message! I left a few minor comments which are not crucial so I'm approving.

docs/python/python.rst
nafur: This depends on whether you have already build the python bindings in this particular build folder before. If you did, you may get proper documentation (if you didn't change anything about that), you may get something outdated (if you changed the python bindings, but did not build it), you may get nothing (if something changed that breaks how python imports the lib. I guess this can happen).

docs/python/CMakeLists.txt
nafur: Yes. The easiest way to generate documentation for python code with Sphinx is using `autodoc`, which will essentially do introspection on the python module and extract the docstrings. For this to work, the module needs to be working.

PR #6486, opened 2021-05-04
Move env into smt solver, theory engine, prop engine

ajreynol:
This is work towards eliminating singletons.

Also, TheoryModel should use the same substitution map as the preprocessor.  This is work towards unifying these things, which will be done in a future PR.


nafur (approved):
LGTM.

PR #6487, opened 2021-05-04
Simplifications to expand definitions

ajreynol:
This removes the expandOnly flag from expandDefinitions.

The use of expandOnly = true is equivalent to applying top-level substitutions only, which should be done explicitly via `Env::getTopLevelSubstitutions`.  It updates the trust substitutions utility to distinguish `apply` vs `applyTrusted` for convenience for this purpose.

This also breaks several dependencies in e.g. expand definitions module.


aniemetz (approved):
Nice!

PR #6489, opened 2021-05-04
Save block comments associated with each kind when parsing kinds file

makaimann:
This PR adds features to the `KindsParser` for saving and looking up the documentation comment associated with each Kind. This PR does not make use of it yet, but future PRs can query for the comment to automatically add it to language binding documentation (e.g., Python / Java bindings).


mudathirmahgoub (approved):
LGTM. This is a good start towards generating doc for kinds in python and Java. 
Further processing is still needed to  handle  `Create with` paragraphs which are written specifically for the C++ API.  This can be done in future PRs.

PR #6490, opened 2021-05-05
Integrate documentation build with the regular CI workflow

nafur:
The new documentation workflow requires building CVC5 again in a separate workflow, which takes quite some time.
This PR integrates building the documentation with the regular CI workflow.


.github/workflows/ci.yml
```diff
@@ -214,3 +218,79 @@ jobs:
...
+      with:
+        name: documentation
+        path: build/docs/sphinx-gh/
+
+  upload-docs:
+    if: github.event_name == 'push' || github.event_name == 'pull_request_target'
+    name: upload-docs
+    runs-on: ubuntu-latest
+    continue-on-error: true
+    needs: build
```
mpreiner, line 239: Can we make this dependent on `ubuntu:production` only? Then we don't have to wait with the upload until all builds are done.

nafur: It seems this is not possible: https://github.com/CVC4/CVC4/actions/runs/815832121
Also, I've found nothing in the documentation about requiring a single job from a matrix

mpreiner (approved):

PR #6493, opened 2021-05-05
Do not have quantifiers model inherit from theory model

ajreynol:
This is work towards making the initialization of theory engine, theory models, quantifiers engine more flexible.

This makes it so that the specialized quantifiers models classes (FirstOrderModel) do not inherit from TheoryModel.  There is no longer any reason for this, since FirstOrderModel does not have any override methods.

As a result of this PR, there is only one kind of TheoryModel and it is constructed immediately when ModelManager is constructed.

This required refactoring the initialization of when FirstOrderModel is constructed in ModelBuilder classes in quantifiers.

This also avoids the need for casting TheoryModel to FirstOrderModel.

### DIFF
--- a/src/theory/quantifiers_engine.cpp
+++ b/src/theory/quantifiers_engine.cpp
@@ -65,10 +65,10 @@ QuantifiersEngine::QuantifiersEngine(
   // Finite model finding requires specialized ways of building the model.
   // We require constructing the model here, since it is required for
   // initializing the CombinationEngine and the rest of quantifiers engine.
-  if ((options::finiteModelFind() || options::fmfBound())
-      && (options::mbqiMode() == options::MbqiMode::FMC
-          || options::mbqiMode() == options::MbqiMode::TRUST
-          || options::fmfBound()))
+  if (options::fmfBound()
+      || (options::finiteModelFind()
+          && (options::mbqiMode() == options::MbqiMode::FMC
+              || options::mbqiMode() == options::MbqiMode::TRUST)))
   {
     Trace("quant-init-debug") << "...make fmc builder." << std::endl;
     d_builder.reset(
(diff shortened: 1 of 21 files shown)


src/theory/quantifiers_engine.cpp
```diff
@@ -44,21 +44,54 @@ namespace cvc5 {
...
+  Trace("quant-init-debug")
+      << "Initialize model engine, mbqi : " << options::mbqiMode() << " "
+      << options::fmfBound() << std::endl;
+  // Finite model finding requires specialized ways of building the model.
+  // We require constructing the model here, since it is required for
+  // initializing the CombinationEngine and the rest of quantifiers engine.
+  if ((options::finiteModelFind() || options::fmfBound())
+      && (options::mbqiMode() == options::MbqiMode::FMC
+          || options::mbqiMode() == options::MbqiMode::TRUST
+          || options::fmfBound()))
```
mpreiner, line 71: Simplified condition:

```suggestion
  if (options::fmfBound()
      || (options::finiteModelFind()
          && (options::mbqiMode() == options::MbqiMode::FMC
              || options::mbqiMode() == options::MbqiMode::TRUST)))
```

src/theory/theory_model.cpp
```diff
@@ -681,6 +681,19 @@ bool TheoryModel::areDisequal(TNode a, TNode b)
   }
 }
 
+bool TheoryModel::hasUfTerms(Node f) const
+{
+  std::map<Node, std::vector<Node> >::const_iterator it = d_uf_terms.find(f);
+  return it != d_uf_terms.end();
```
mpreiner, line 687:
```suggestion
  return d_uf_terms.find(f) != d_uf_terms.end();
```

src/theory/theory_model.cpp
```diff
@@ -681,6 +681,19 @@ bool TheoryModel::areDisequal(TNode a, TNode b)
...
+}
+
+const std::vector<Node>& TheoryModel::getUfTerms(Node f) const
+{
+  std::map<Node, std::vector<Node> >::const_iterator it = d_uf_terms.find(f);
```
mpreiner, line 692:
```suggestion
  const auto it = d_uf_terms.find(f);
```

mpreiner (approved):

PR #6496, opened 2021-05-06
Add more getters for api::Term

nafur:
This PR adds more getter functions for `api::Term` to retrieve values from constant terms (and terms that the average API use might consider constant).
It also introduces `std::wstring` as regular representation for string constants instead of `std::vector<uint32>` for the SMT-LIB parser and the `String` class.


src/api/cpp/cvc5.cpp
```diff
@@ -2721,6 +2758,398 @@ std::vector<Node> Term::termVectorToNodes(const std::vector<Term>& terms)
...
+bool Term::isConstArray() const
+{
+  CVC5_API_TRY_CATCH_BEGIN;
+  CVC5_API_CHECK_NOT_NULL;
+  //////// all checks before this line
+  return d_node->getKind() == cvc5::Kind::STORE_ALL;
+  ////////
+  CVC5_API_TRY_CATCH_END;
+}
+Term Term::getConstArray() const
```
ajreynol, line 2835: I think this should be called "getConstArrayDefaultValue" or something.  We are not getting the constant array, but the value this is stored.

src/api/cpp/cvc5.cpp
```diff
@@ -2721,6 +2758,398 @@ std::vector<Node> Term::termVectorToNodes(const std::vector<Term>& terms)
...
+      << "Term should be a Set when calling getSet()";
+  //////// all checks before this line
+  std::set<Term> res;
+  collectSet(res, *d_node, d_solver);
+  return res;
+  ////////
+  CVC5_API_TRY_CATCH_END;
+}
+
+bool Term::isChar() const
```
ajreynol, line 3072: I'd omit isChar / getChar.

ajreynol (changes requested):
Also need unit tests, are those coming?

src/api/cpp/cvc5.cpp
nafur: Oh, actually we already have `getConstArrayBase` which does this. I'll remove this method.

src/api/cpp/cvc5.cpp
nafur: Hm, note that `getChar()` is different from `getString()[0]`. It mirrors the semantics of `mkChar()` that accepts a code point in hex representation.

nafur:
Unit tests have been added.

src/api/cpp/cvc5.cpp
```diff
@@ -2721,6 +2752,414 @@ std::vector<Node> Term::termVectorToNodes(const std::vector<Term>& terms)
...
+  for (; codePoint > 0; codePoint /= 16)
+  {
+    res += digits[codePoint % 16];
+  }
+  std::reverse(res.begin(), res.end());
+  return res;
+  ////////
+  CVC5_API_TRY_CATCH_END;
+}
+bool Term::isSequence() const
```
ajreynol, line 3095: As discussed offline, we should change this to `isConstSequence` which holds only when `getKind()==CONST_SEQUENCE`.

src/api/cpp/cvc5.cpp
```diff
@@ -2721,6 +2752,414 @@ std::vector<Node> Term::termVectorToNodes(const std::vector<Term>& terms)
...
+  return d_node->getKind() == cvc5::Kind::CONST_BITVECTOR;
+  ////////
+  CVC5_API_TRY_CATCH_END;
+}
+std::string Term::getBitVector(std::uint32_t base) const
+{
+  CVC5_API_TRY_CATCH_BEGIN;
+  CVC5_API_CHECK_NOT_NULL;
+  CVC5_API_CHECK(d_node->getKind() == cvc5::Kind::CONST_BITVECTOR)
+      << "Term should be a BitVector when calling getBitVector()";
```
aniemetz, line 2877: It would be nicer to use the `CVC5_ARG_CHECK_EXPECTED` macro for this kind of checks (all of them).
Also, `BitVector` is not a type known at API level, maybe use `bit-vector value`.

src/api/cpp/cvc5.cpp
```diff
@@ -2721,6 +2752,414 @@ std::vector<Node> Term::termVectorToNodes(const std::vector<Term>& terms)
...
+  return d_node->getKind() == cvc5::Kind::CONST_FLOATINGPOINT;
+  ////////
+  CVC5_API_TRY_CATCH_END;
+}
+std::tuple<std::uint32_t, std::uint32_t, Term> Term::getFloatingPoint() const
+{
+  CVC5_API_TRY_CATCH_BEGIN;
+  CVC5_API_CHECK_NOT_NULL;
+  CVC5_API_CHECK(d_node->getKind() == cvc5::Kind::CONST_FLOATINGPOINT)
+      << "Term should be a FloatingPoint when calling getFloatingPoint()";
```
aniemetz, line 3012: Same as with bit-vectors, `FloatingPoint` -> `floating-point value`.

src/api/cpp/cvc5.h
```diff
@@ -1266,6 +1259,146 @@ class CVC5_EXPORT Term
    */
   std::wstring getString() const;
 
+  bool isReal32() const;
```
aniemetz, line 1262: Document.

src/api/cpp/cvc5.h
```diff
@@ -1266,6 +1259,146 @@ class CVC5_EXPORT Term
...
+  /**
+   * Returns the stored rational as a pair of its numerator and denominator.
```
aniemetz, line 1264:
```suggestion
   * Get the representation of a rational value as a pair of its numerator and denominator.
```
I suggest to reword this accordingly in all of the functions below, `stored` is a bit weird (the term must be a value, that's the key point).

aniemetz (changes requested):
In the header documentation, `Returns` -> `Return`.

src/api/python/cvc5.pxi
```diff
@@ -735,23 +740,11 @@ cdef class Solver:
...
-                if not isinstance(u, int):
-                    raise ValueError("List should contain ints but got: {}"
-                                     .format(str_or_vec))
-                v.push_back(<unsigned> u)
-            term.cterm = self.csolver.mkString(<const vector[unsigned]&> v)
-        else:
-            raise ValueError("Expected string or vector of ASCII codes"
-                             " but got: {}".format(str_or_vec))
+        cdef Py_ssize_t size
+        cdef const wchar_t* tmp = PyUnicode_AsWideCharString(s, &size)
```
makaimann, line 746: This is a little out of my scope of understanding, but do you think you need to free this with `PyMem_Free`?  I'm looking at the documentation here: https://docs.python.org/3/c-api/unicode.html

nafur: As this is the first time I'm using this stuff: I have no idea :-)
I'll have a look...

nafur: Yeah, it seems we should do this.

ajreynol (commented):
One minor comment:
- Is `isConstInteger` better than `isInteger`?  Similar for others.

Then, I believe the only methods I disagree with in the PR are `isSet` / `getSet`.  This actually is not intuitive:
`isSet( (intersect A B) )` returns true but `getSet( (intersect A B) )` returns the empty set.

One could do something like `  return d_node->getType().isSet() && return d_node->isConst();` for `isSet`.  However, this is now exposing the detail of when we consider a set to be constant.

I do not think there is a good solution to this, thus I'd omit isSet / getSet.

This is analogous to why there are no methods to access "constant" datatypes or arrays (beyond the constant base arrays, which is fine).  For arrays, datatypes, and sets, constant values are terms with an AST, hence I don't think it makes sense to have a getter for any of them.

nafur:
As per offline discussion:
- to make clear these functions are about constant values, rename `isInteger()` to `isIntegerValue()` (and similar for the rest)
- handle sets in a sane way such that `isSetValue()` and `getSetValue()` are consistent

I hope we are done now...?

src/api/cpp/cvc5.cpp
```diff
@@ -2720,6 +2754,384 @@ std::vector<Node> Term::termVectorToNodes(const std::vector<Term>& terms)
...
+                              *d_node)
+      << "Term to be an abstract value when calling "
+         "getAbstractValue()";
+  //////// all checks before this line
+  return d_node->getConst<AbstractValue>().getIndex().toString();
+  ////////
+  CVC5_API_TRY_CATCH_END;
+}
+
+bool Term::isTupleValue() const
```
ajreynol, line 2912: This needs to check whether the underlying datatype is a tuple (`DType::isTuple`) and also check that d_node is a constant (`isConst`).

src/api/cpp/cvc5.cpp
```diff
@@ -2720,6 +2754,384 @@ std::vector<Node> Term::termVectorToNodes(const std::vector<Term>& terms)
...
+  for (size_t i = 0, n = d_node->getNumChildren(); i < n; ++i)
+  {
+    res.emplace_back(Term(d_solver, (*d_node)[i]));
+  }
+  return res;
+  ////////
+  CVC5_API_TRY_CATCH_END;
+}
+
+bool Term::isPosZero() const
```
ajreynol, line 2939: These should probably be prefixed with FloatingPoint, to distinguish from arithmetic.

ajreynol (changes requested):

src/api/cpp/cvc5.cpp
nafur: Are you fine with `isFPPosZero()`? `isFloatingPointPosZero()` is so long...

ajreynol: @aniemetz opinions?  In some ways I think `isFloatingPointPosZero` is more readable.

ajreynol: Actually, I think my preference is the longer name.  

Regardless, I think we should be consistent wrt `isFloatingPointValue`, which would change to `isFPValue`.

nafur: Done.

src/api/cpp/cvc5.h
```diff
@@ -1213,63 +1206,215 @@ class CVC5_EXPORT Term
...
    */
-  bool isString() const;
+  bool isStringValue() const;
   /**
-   * @return the stored string constant.
-   *
-   * Note: This method is not to be confused with toString() which returns the
-   *       term in some string representation, whatever data it may hold.
-   *       Asserts isString().
+   * Note: This method is not to be confused with toStringValue() which returns
```
ajreynol, line 1259:
```suggestion
   * Note: This method is not to be confused with toString() which returns
```

ajreynol (changes requested):
One minor fix, otherwise this looks good to me.

ajreynol (approved):
LGTM, thanks for the updates, its been a fun journey :)

aniemetz (approved):

PR #6499, opened 2021-05-06
Unify top-level substitutions and model substitutions

ajreynol:
This unifies top-level substitutions and theory model substitutions. Env is now passed to the TheoryModel, which has access to top-level substitutions.

The former was used for simplfying assertions in the preprocessor, the latter was used for evaluating terms in the model.  

There is no reason to have these two things be separate.


src/preprocessing/passes/non_clausal_simp.cpp
```diff
@@ -327,34 +327,33 @@ PreprocessingPassResult NonClausalSimp::applyInternal(
...
   {
-    Node lhs = (*pos).first;
-    TrustNode trhs = newSubstitutions->apply((*pos).second);
-    Node rhs = trhs.isNull() ? (*pos).second : trhs.getNode();
-    // If using incremental, we must check whether this variable has occurred
-    // before now. If it hasn't we can add this as a substitution.
-    if (!assertionsToPreprocess->storeSubstsInAsserts()
-        || d_preprocContext->getSymsInAssertions().find(lhs)
-               == d_preprocContext->getSymsInAssertions().end())
+    for (SubstitutionMap::iterator pos = nss.begin(); pos != nss.end(); ++pos)
```
nafur, line 340:
```suggestion
    for (const std::pair<const Node, Node>& pos: nss)
```

src/preprocessing/passes/non_clausal_simp.cpp
```diff
@@ -327,34 +327,33 @@ PreprocessingPassResult NonClausalSimp::applyInternal(
...
-          << std::endl;
-       trhs = newSubstitutions->apply((*pos).first);
-       Assert(!trhs.isNull());
-       assertionsToPreprocess->addSubstitutionNode(trhs.getProven(),
-       trhs.getGenerator());
+      Node lhs = (*pos).first;
+      // If using incremental, we must check whether this variable has occurred
+      // before now. If it has, we must add as an assertion.
+      if (d_preprocContext->getSymsInAssertions().find(lhs)
+          != d_preprocContext->getSymsInAssertions().end())
```
nafur, line 346:
```suggestion
      if (d_preprocContext->getSymsInAssertions().contains(lhs))
```

nafur (commented):
Only minor suggestions about making the code a bit easier to read.

src/preprocessing/passes/miplib_trick.cpp
```diff
@@ -550,7 +550,6 @@ PreprocessingPassResult MipLibTrick::applyInternal(
                   << "unexpected solution from arith's ppAssert()";
               Assert(nullMap.empty())
                   << "unexpected substitution from arith's ppAssert()";
-              d_preprocContext->addModelSubstitution(*ii, newVar.eqNode(one));
```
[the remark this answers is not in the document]
ajreynol: My justification for removing it is that we now have the invariant `top-level substitutions = model substitutions`.  Either:
(A) This preprocessing pass was already maintaining this invariant, in which case this call is spurious and can be removed, since after this PR, this is a consequence of its call to adding top-level substitutions (https://github.com/cvc5/cvc5/blob/master/src/preprocessing/passes/miplib_trick.cpp#L592).
(B) This preprocessing pass is not maintaining this invariant, in which case we shouldn't have been making this call in the first place, since this would imply that we weren't keeping our model consistent with our preprocessing.

I don't quite follow what this pass is doing, but regardless, in either case this line should be deleted.

test/regress/regress1/strings/issue5611-deq-norm-emp.smt2
```diff
@@ -7,4 +7,5 @@
 (assert (not (= str8 (str.replace_re_all (str.++ str9 str8 str7) (str.to_re "ULfQhzdcQSfqWSXyjuptqjqsazpyjzdDddlJPLDJhalmhBhlNBKZvoxoLOpfplkqhvIRHMOMDAGIdoRyiZmxmMvRijgpFnd") str9))))
 (push)
 (assert (and (str.in_re str8 (str.to_re str7)) (not (= str9 (str.replace_re_all (str.++ str8 str8) (str.to_re "ULfQhzdcQSfqWSXyjuptqjqsazpyjzdDddlJPLDJhalmhBhlNBKZvoxoLOpfplkqhvIRHMOMDAGIdoRyiZmxmMvRijgpFnd") str9)))))
+; note that (debug) model checking fails since this benchmark triggers a string variable (str7) to be solved after it appears in assertions
```
nafur, line 10: What does this mean?

nafur (commented):
It feels like this could change solver behavior. Have you done any benchmarks?
One question about the regression.
Otherwise LGTM.

ajreynol:
@nafur This should have very little impact on solving.  The performance impact of this change is to store top-level substitutions in a single place. This map is usually quite small.  Moreover it does not impact anything we are doing during solving, it mainly impacts how the model is stored.

test/regress/regress1/strings/issue5611-deq-norm-emp.smt2
ajreynol: --debug-check-models is an option which checks whether assertions are satisfied after each "sat".

In this benchmark a string variable `str7` is solved during non-clausal simplification after we already push assertions involving `str7` to the solver.  This causes debug-check-models to fail.

PR #6504, opened 2021-05-07
Loop over terms to reconstruct instead of obligations.

abdoo8080:
This PR modifies the rcons algorithm to loop over terms to reconstruct instead of obligations. It also modifies the Obs data structure to reflect this change. The rest of the PR is mostly updating documentation and refactoring the affected code.


src/theory/quantifiers/sygus/rcons_obligation.h
```diff
@@ -25,37 +25,46 @@ namespace theory {
...
- * class. Notice that the SyGuS type T of an obligation is not stored in this
- * class as it can be inferred from the type of the skolem k.
+ * of type T whose builtin analog is equivalent to t. This class encodes each
+ * obligation (T, t) as a skolem k of type T to embed obligations in candidate
+ * solutions (see d_candSols below). Notice that the SyGuS type T of an
+ * obligation is not stored in this class as it can be inferred from the type of
+ * the skolem k.
  */
-class RConsObligationInfo
+class Obligation
```
ajreynol, line 36: I think I'd prefer the name of this class was still specific to reconstruction, maybe RConsObligation, esp. since we don't enclose this in a namespace.

src/theory/quantifiers/sygus/sygus_reconstruct.cpp
```diff
@@ -169,165 +160,161 @@ Node SygusReconstruct::reconstructSolution(Node sol,
...
   if (Trace("sygus-rcons").isConnected())
   {
-    RConsObligationInfo::printCandSols(mainOb, d_obInfo);
+    Obligation::printCandSols(mainOb, d_obs);
     printPool(pool);
   }
 
   // if the main obligation is solved, return the solution
-  if (!d_sol[mainOb].isNull())
+  if (!d_sol[mainOb->getSkolem()].isNull())
```
ajreynol, line 177: Nit: may be clearer if mainOb->getSkolem() was put in a local variable.

src/theory/quantifiers/sygus/sygus_reconstruct.h
```diff
@@ -213,27 +216,28 @@ class SygusReconstruct : public expr::NotifyMatch
    * example, given:
    *
    * CandSols = {
-   *  mainOb -> {(+ z1 1)},
-   *  z1 -> {(* z2 x)},
-   *  z2 -> {2}
+   *   c_mainOb -> {(c_+ c_z1 c_1)},
```
ajreynol, line 219: `c_mainOb` is a skolem here?  I think it would be clearer if we used k_mainOb instead (I tend to use `k` to represent skolems).  

This also distinguishes this from c_+, which is a sygus constructor here I guess?

ajreynol (changes requested):
Some initial comments. Can you also add/enable a regression that is solved due to this PR? (I assume there is one that motivated this?)

src/theory/quantifiers/sygus/sygus_reconstruct.h
abdoo8080: Replaced `c_mainOb` with `k0`, which is the main skolem in the algorithm. Also, updated comments to be consistent with `z`s (represent free vars) and `k`s (represent skolems).

ajreynol (approved):

PR #6506, opened 2021-05-07
Fix and add missing REQUIRE labels for FP regression tests.

aniemetz:


ajreynol (commented):
Should this be `REQUIRES` not `REQUIRE`?  The other uses seem to be "REQUIRES".

aniemetz:
REQUIRES failed for some reason. Let me double check what is going on here.

aniemetz:
I'm not sure what happened. Maybe I misread `skipped` and `failed. You are, of course, right.

ajreynol (approved):

PR #6513, opened 2021-05-10
Remove read_only from options.

nafur:
This PR removes the possibility of declaring options `read_only`.
It's only effect is making an attempts to disallow changing the respective option from within our internal code (by not providing a setter method). However, a "read-only" option can still be set via the `setOption()` methods that is also used by the API, and by SMT-LIB's `set-option`.

### DIFF
--- a/src/options/mkoptions.py
+++ b/src/options/mkoptions.py
@@ -56,7 +56,7 @@ MODULE_ATTR_ALL = MODULE_ATTR_REQ + ['option']
 OPTION_ATTR_REQ = ['category', 'type']
 OPTION_ATTR_ALL = OPTION_ATTR_REQ + [
     'name', 'help', 'help_mode', 'smt_name', 'short', 'long', 'default',
-    'includes', 'handler', 'predicates', 'read_only',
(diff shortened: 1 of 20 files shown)


src/options/mkoptions.py
```diff
@@ -260,12 +251,11 @@ def __init__(self, d):
         self.__dict__ = dict((k, None) for k in OPTION_ATTR_ALL)
         self.includes = []
         self.predicates = []
-        self.read_only = False
         self.alternate = True    # add --no- alternative long option for bool
         self.filename = None
         for (attr, val) in d.items():
             assert attr in self.__dict__
-            if attr in ['read_only', 'alternate'] or val:
+            if attr in ['alternate'] or val:
```
mpreiner, line 258:
```suggestion
            if attr == 'alternate' or val:
```

mpreiner (approved):
LGTM, one minor suggestion. @ajreynol any objections to removing the `read_only` flag?

ajreynol:
Yes, I agree to remove it.  It's simpler this way, there is no real benefit to having it in the first place.

PR #6516, opened 2021-05-10
bv: Add support for --bitblast=eager.

mpreiner:
This PR adds support for handling --bitblast=eager in the new bitblast solver.

Note: This PR is based on #6515, for reviewing you can just look at the last commit https://github.com/cvc5/cvc5/pull/6516/commits/0ba8ec8a99ada01eafdbf52e2897bb6363c7d9c2.


src/smt/set_defaults.cpp
```diff
@@ -165,7 +165,11 @@ void setDefaults(LogicInfo& logic, bool isInternalSubsolver)
 
     // Force lazy solver since we don't handle EAGER_ATOMS in the
     // BVSolver::BITBLAST solver.
-    opts.set(options::bvSolver, options::BVSolver::LAZY);
+    if (options::bvSolver() != options::BVSolver::BITBLAST
```
ajreynol, line 168: Update the comment above.

ajreynol (approved):
One minor comment, else LGTM

src/smt/set_defaults.cpp
mpreiner: I removed the code block since all BV solver now support eager bitblasting.

PR #6521, opened 2021-05-11
Update to sygus standard output for check-synth responses

ajreynol:
This PR does two things:
(1) It eliminates the ad-hoc implementation of printSynthSolutions and removes it from the API.  Now, printing response to a check-synth command is done in a more standard way, using the API + symbol manager.  This is analogous to recent refactoring to get-model.
(2) It updates cvc5's output in response to check-synth to be compliant with the upcoming sygus 2.1 standard. The standard has changed slightly: responses to `check-synth` are now closed in parentheses, mirroring the smt2 response to `get-model`.

It also removes the unused command GetSynthSolutionCommand.


ajreynol:
Decided to eliminate printSynthSolutions, @abdoo8080 will update the examples in a separate PR first.

src/expr/symbol_manager.cpp
```diff
@@ -65,10 +66,14 @@ class SymbolManager::Implementation
   std::vector<api::Sort> getModelDeclareSorts() const;
   /** get model declare terms */
   std::vector<api::Term> getModelDeclareTerms() const;
+  /** get functions to synthesize */
+  std::vector<api::Term> getFunctionsToSynthesize() const;
   /** Add declared sort to the list of model declarations. */
   void addModelDeclarationSort(api::Sort s);
   /** Add declared term to the list of model declarations. */
   void addModelDeclarationTerm(api::Term t);
+  /** Add function to the list of functions-to-synthesize. */
```
HanielB, line 75:
```suggestion
  /** Add function to the list of functions to synthesize. */
```

src/expr/symbol_manager.cpp
```diff
@@ -192,6 +199,13 @@ std::vector<api::Term> SymbolManager::Implementation::getModelDeclareTerms()
   return declareTerms;
 }
 
+std::vector<api::Term> SymbolManager::Implementation::getFunctionsToSynthesize()
+    const
+{
+  std::vector<api::Term> sfuns(d_funToSynth.begin(), d_funToSynth.end());
+  return sfuns;
```
HanielB, line 206: Can you inline this?

src/expr/symbol_manager.h
```diff
@@ -120,6 +125,11 @@ class CVC5_EXPORT SymbolManager
    * Add declared term to the list of model declarations.
    */
   void addModelDeclarationTerm(api::Term t);
+  /**
+   * Add a function-to-synthesize. This ensures the solution for f is printed
```
HanielB, line 129:
```suggestion
   * Add a function to synthesize. This ensures the solution for f is printed
```

src/smt/command.cpp
```diff
@@ -699,6 +699,7 @@ const api::Grammar* SynthFunCommand::getGrammar() const { return d_grammar; }
 
 void SynthFunCommand::invoke(api::Solver* solver, SymbolManager* sm)
 {
+  sm->addFunctionToSynthesize(d_fun);
```
HanielB, line 702: Why is it called here rather than when the command is declared?

ajreynol: Creating a command should not do this.  It is only if/when the command is invoked that it should update the state of symbol manager.

HanielB: I see, cool.

HanielB (approved):

PR #6522, opened 2021-05-11
Preliminary draft of changes for SMT comp 2021 for quantifiers

ajreynol:
Covers improvements to quantifiers.

### DIFF
--- a/contrib/competitions/smt-comp/run-script-smtcomp-current
+++ b/contrib/competitions/smt-comp/run-script-smtcomp-current
@@ -75,10 +75,11 @@ QF_NRA)
   finishwith --nl-ext-tplanes --decision=justification
   ;;
 # all logics with UF + quantifiers should either fall under this or special cases below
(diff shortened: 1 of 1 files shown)


4tXJ7f (approved):
LGTM

ajreynol:
One issue: I've added 2 logics with FP to the "main" quantifiers block, which is the right place for them.  However, this doesnt enable `--fp-exp`.

Looking at the script, it looks like we always want `--fp-exp` enabled.  Perhaps we even want `--fp-exp` enabled by default in general at this point, as it would also simplify our testing.  What do you think @aniemetz ?

martin-cs:
( Not @aniemetz but `--fp-ext` will crash on some formats.  I not currently aware of any correctness issues but ... that could well be because I haven't looked hard enough. )

aniemetz:
Yes, what @martin-cs said.

aniemetz:
@ajreynol we should enable it in the SMT-COMP script for all configurations.

PR #6523, opened 2021-05-11
Add getId function to python API

alex-ozdemir:
(Z3 exposes it to facilitate custom hashes)


makaimann:
Thanks for the PR! It looks like `getId` is already defined here:
https://github.com/cvc5/cvc5/blob/87ddf639068e9bb917c6c00c6f2c73ec6844af9a/src/api/python/cvc5.pxi#L1487

And tested here:
https://github.com/cvc5/cvc5/blob/b016f603f2a3709faaf90008a09f2567d7283ff3/test/python/unit/api/test_term.py#L39

Do you still want to add this example though?

alex-ozdemir:
Oh, whoops! I seemed to have missed this PR: https://github.com/cvc5/cvc5/pull/6477.

Closing.

alex-ozdemir:
Whoops again, yeah, me might as well at the example.

examples/api/python/id.py
```diff
@@ -0,0 +1,39 @@
...
+#
+# This file is part of the cvc5 project.
+#
+# Copyright (c) 2009-2021 by the authors listed in the file AUTHORS
+# in the top-level source directory and their institutional affiliations.
+# All rights reserved.  See the file COPYING in the top-level source
+# directory for licensing information.
+# #############################################################################
+#
+# A simple demonstration of the solving capabilities of the cvc5 sets solver
```
makaimann, line 14: This description should be updated for this example.

examples/api/python/id.py
```diff
@@ -0,0 +1,39 @@
...
+# through the Python API. This is a direct translation of sets-new.cpp.
+##
+
+import pycvc5
+from pycvc5 import kinds
+
+if __name__ == "__main__":
+    slv = pycvc5.Solver()
+
+    # Optionally, set the logic. We need at least UF for equality predicate,
```
makaimann, line 24: Similarly, this comment is a little confusing because the logic for this example is arbitrary, right?

makaimann (changes requested):
Sounds good! Left a few easy-to-address change requests.

examples/api/python/id.py
alex-ozdemir: Good catch. Fixed in c7a99b06dd.

examples/api/python/id.py
alex-ozdemir: Good catch. Fixed in c7a99b06dd.

makaimann (approved):
LGTM, thanks for the changes!

PR #6524, opened 2021-05-11
Fix error message in toPythonObj

alex-ozdemir:


makaimann (approved):
LGTM, thanks for the fix!

PR #6527, opened 2021-05-11
Split options holder class

nafur:
This PR splits the `OptionsHolder` class into separate holder classes for every options module and makes them directly accessible via appropriate `Options::<module>()` methods. We forward declare these new holder classes and thereby retain that we only need to recompile when we change an option module that is explicitly included.
All (generated) methods that previously accessed the old holder object are changed to instead use the new holder objects.
This PR does the bare minimum to do this change, further PRs will eventually get rid of all template specializations that currently surround our options class.


src/options/options_template.h
```diff
@@ -36,19 +36,20 @@ class Solver;
 namespace options {
   struct OptionsHolder;
   class OptionsHandler;
+${holder_fwd_decls}$
```
mpreiner, line 39: Wrap with // clang-format off/on.

src/options/options_template.h
```diff
@@ -36,19 +36,20 @@ class Solver;
...
 class CVC5_EXPORT Options
 {
   friend api::Solver;
-  /** The struct that holds all option values. */
-  std::unique_ptr<options::OptionsHolder> d_holder;
 
   /** The handler for the options of the theory. */
   options::OptionsHandler* d_handler;
 
+${holder_mem_decls}$
```
mpreiner, line 51: Ditto

src/options/options_template.h
```diff
@@ -104,6 +105,8 @@ class CVC5_EXPORT Options
   Options(OptionsListener* ol = nullptr);
   ~Options();
 
+${holder_getter_decls}$
```
mpreiner, line 108: Ditto

src/options/module_template.h
```diff
@@ -26,13 +26,23 @@
...
 ${modes}$
 
+#if defined(CVC5_MUZZLED) || defined(CVC5_COMPETITION_MODE)
+#  define DO_SEMANTIC_CHECKS_BY_DEFAULT false
+#else /* CVC5_MUZZLED || CVC5_COMPETITION_MODE */
+#  define DO_SEMANTIC_CHECKS_BY_DEFAULT true
+#endif /* CVC5_MUZZLED || CVC5_COMPETITION_MODE */
+
+struct Holder${id_cap}$ {
+${holder_spec}$
```
mpreiner, line 41: Wrap with // clang-format off/on

mpreiner (approved):

PR #6530, opened 2021-05-11
Stop using the solver for printing sygus synthesis solutions.

abdoo8080:
This PR uses custom methods for printing the synthesis solutions instead of `Solver::printSynthSolution`, which is going to be removed by #6521.


examples/api/utils.cpp
```diff
@@ -0,0 +1,73 @@
...
+std::string defineFunToString(const cvc5::api::Term& f,
+                              const std::vector<cvc5::api::Term> params,
+                              const cvc5::api::Sort& sort,
+                              const cvc5::api::Term body)
+{
+  std::stringstream ss;
+  ss << "(define-fun " << f << " (";
+  std::vector<cvc5::api::Term>::const_iterator it = params.cbegin(),
+                                               cend = params.cend();
+  if (it != cend)
```
ajreynol, line 36: These 4 lines are spurious, if you remove them I think it is equivalent.

ajreynol (changes requested):

examples/api/utils.cpp
abdoo8080: Could you elaborate?

ajreynol: This is essentially just hard coding one run of the while loop below, if you delete lines 36-40, the code will be the same I believe.

abdoo8080: They're almost identical. Line 43 prints 1 more space character than line 38.

abdoo8080: Replaced with a for loop.

ajreynol (approved):

PR #6533, opened 2021-05-13
Decouple parser creation from input selection

4tXJ7f:
This commit decouples the creation of a `Parser` instance from creating
an `Input` and setting the `Input` on the parser. This is a first step
in refactoring the parser infrastructure. A future PR will split the
parser class into three classes: `Parser`, `ParserState`, and
`InputParser`. The `Parser` and `InputParser` classes will be the
public-facing classes. The new `Parser` class will have methods to
create `InputParser`s from files, streams, and strings. `InputParser`s
will have methods to get commands/exprs from a given input. The
`ParserState` class will keep track of the state of the parser and will
be the internal interface for the parsers. The current `Parser` class is
used both publicly and internally, which is messy.

Note: this PR includes #6532.


4tXJ7f:
[This](https://github.com/4tXJ7f/cvc5/commit/74e552f6cb60c3206b59277f2648cddbf67b6743) is what the `InputParser` class will look like.

test/unit/parser/parser_black.cpp
```diff
@@ -86,11 +86,11 @@ class TestParserBlackParser : public TestInternal
   void tryGoodInput(const std::string goodInput)
   {
     d_symman.reset(new SymbolManager(d_solver.get()));
-    Parser* parser = ParserBuilder(d_solver.get(), d_symman.get(), "test")
-                         .withStringInput(goodInput)
+    Parser* parser = ParserBuilder(d_solver.get(), d_symman.get())
```
nafur, line 89: Shouldn't we use `std::unique_ptr` here as well? (and a few times below...)

nafur (approved):
Almost LGTM, just a comment on the tests.

test/unit/parser/parser_black.cpp
4tXJ7f: Done, thanks!

PR #6534, opened 2021-05-13
Add std::hash overloads for Node, TNode and TypeNode.

mpreiner:
Eliminates NodeHashFunction, TNodeHashFunction and TypeNodeHashFunction.

Note to reviewers: Main changes are in `expr/node.{h,cpp}` and `expr/type_node.{h,cpp}`.


src/theory/substitutions.h
```diff
@@ -43,33 +43,33 @@ namespace theory {
...
 
-  typedef context::CDHashMap<Node, Node, NodeHashFunction> NodeMap;
-
-  typedef NodeMap::iterator iterator;
-  typedef NodeMap::const_iterator const_iterator;
+ typedef NodeMap::iterator iterator;
+ typedef NodeMap::const_iterator const_iterator;
 
 private:
+ typedef std::unordered_map<Node, Node> NodeCache;
```
4tXJ7f, line 52: Looks like the formatting is wrong because of `private`/`public` not being indented.

4tXJ7f (approved):
LGTM, just one minor formatting issue.

src/theory/substitutions.h
mpreiner: clang-formatted the whole class, should be fixed now.

nafur (approved):
TBH I've only looked at  node/type_node. But that LGTM.

PR #6540, opened 2021-05-14
Add non-templated method to set option defaults

nafur:
This PR replaces the templated `Options::setDefault()` methods by new non-templated free functions `options::{module}::setDefault{option}()`.
These methods should be used instead of the common `if (!set by user) { set option value }` pattern.


src/options/mkoptions.py
```diff
@@ -116,10 +116,18 @@
...
 
 TPL_HOLDER_MACRO_ATTR = "  {type} {name};\\\n"
-TPL_HOLDER_MACRO_ATTR += "  bool {name}__setByUser__ = false;"
+TPL_HOLDER_MACRO_ATTR += "  bool {name}__setByUser = false;"
 
 TPL_HOLDER_MACRO_ATTR_DEF = "  {type} {name} = {default};\\\n"
-TPL_HOLDER_MACRO_ATTR_DEF += "  bool {name}__setByUser__ = false;"
+TPL_HOLDER_MACRO_ATTR_DEF += "  bool {name}__setByUser = false;"
+
+TPL_DECL_SET_DEFAULT = 'void default_{name}(Options& opts, {type} value);'
```
ajreynol, line 124: Would `setDefault_` ... be a better name here?

nafur: using `setDefault{OptionName}` now

mpreiner (approved):

PR #6542, opened 2021-05-14
Improve integration of CAD with nl-Ext

nafur:
This PR improves the integration of the CAD solver with the nl-ext solver in a simple way: we simply use a few of the simple linearization lemmas in combination with CAD by default, significantly improving the performance on QF_NRA.


nafur:
Benchmarks on QF_NRA, timeout 10 minutes, master vs this PR:
```
$ ./analyze.sh -t slurmwd/05140432-cvc5*
                 status  total  solved   sat  unsat  best  timeout  memout  error  uniq  disagr  time_cpu   memory
config                                                                                                            
05140432-cvc5-pr     ok  11489   10634  5001   5633  9090      853       2      0   331       0  564120.6 390321.4
05140432-cvc5        ok  11489   10402  5023   5379  9513     1084       3      0    99       0  705114.5 436913.4

```

src/options/arith_options.toml
```diff
@@ -409,6 +409,14 @@ name   = "Arithmetic theory"
   default    = "true"
   help       = "incremental linearization approach to non-linear"
 
+[[option]]
```
ajreynol, line 412: Can we make nlExt a three valued mode?  Would avoid having a new flag and would ensure the mode was clearer.

ajreynol (changes requested):
Happy to have this behavior, one request.

src/options/arith_options.toml
nafur: For this change, I unfortunately needed to touch a bunch of regressions...

ajreynol (approved):

PR #6543, opened 2021-05-14
Use scikit-build CMake files for pycvc5

makaimann:
This PR removes copied CMake files for the pycvc5 Cython target, and instead adds a dependency on [scikit-build](https://github.com/scikit-build/scikit-build) (where those CMake files originated anyway). This keeps us up to date with fixes. Furthermore, the PR switches from `distutils` to the more modern `setuptools`. This does add another dependency but it's a fairly reasonable one. `setuptools` is not part of the base Python distribution, but my understanding is that it is now the de facto standard, and most package managers install it automatically with Python. The main motivation for switching is in preparation for building wheels.

This PR is a piece of this old one (https://github.com/cvc5/cvc5/pull/5657) which I am closing and splitting up.


src/api/python/CMakeLists.txt
```diff
@@ -19,6 +19,35 @@ if(POLICY CMP0057)
...
+check_python_module("setuptools")
+
+# Find cmake modules distributed with scikit-build
+# They are distributed under <scikit-build directory>/resources/cmake
+execute_process(
+  COMMAND
+    ${PYTHON_EXECUTABLE} -c "from __future__ import print_function;\
+                             import os; import skbuild;\
+                             print(os.path.dirname(skbuild.__file__) + \
+                             os.path.sep + 'resources' + os.path.sep + 'cmake', end='')"
```
mpreiner, line 36:
```suggestion
                             print(os.path.join(os.path.dirname(skbuild.__file__), 'resources', 'cmake'), end='')"
```
Probably needs formatting.

mpreiner: What exactly do you want to check with this? If `<scikit-build directory>/resources/cmake` exists? If yes, wouldn't it be better toi use `os.path.exists` instead?

src/api/python/CMakeLists.txt
```diff
@@ -19,6 +19,35 @@ if(POLICY CMP0057)
...
+  OUTPUT_VARIABLE
+    SKBUILD_CMAKE_MODULE_PATH
+  RESULT_VARIABLE
+    RET_SKBUILD_CMAKE_MODULE_PATH
+  )
+
+if (RET_SKBUILD_CMAKE_MODULE_PATH)
+  message(FATAL_ERROR "Failed to find CMake module path from scikit-build")
+endif()
```
mpreiner, line 45: If this happens what is the user supposed to do?

src/api/python/CMakeLists.txt
```diff
@@ -98,8 +127,12 @@ print('YES' if 'VIRTUAL_ENV' in os.environ else 'NO', end='')")
 set(INSTALL_CMD
     "${PYTHON_EXECUTABLE} ${CMAKE_CURRENT_BINARY_DIR}/setup.py install")
 # if we're in a virtualenv, we install it in the virtualenv lib location
+# otherwise install in site-packages
+# option --single-version-externally-managed prevents it from being
+# an .egg distribution and --record records the list of installed files
 if ("${IN_VIRTUALENV}" STREQUAL "NO")
-  set(INSTALL_CMD "${INSTALL_CMD} --prefix=${CMAKE_INSTALL_PREFIX}")
+  set(INSTALL_CMD "${INSTALL_CMD} --prefix=${CMAKE_INSTALL_PREFIX}
+--single-version-externally-managed --record=pycvc5-installed-files.txt")
```
mpreiner, line 135:
```suggestion
  set(INSTALL_CMD "${INSTALL_CMD}
          --prefix=${CMAKE_INSTALL_PREFIX}
          --single-version-externally-managed
          --record=pycvc5-installed-files.txt")
```

src/api/python/CMakeLists.txt
makaimann: Hmm.. probably complain to the developers haha. Should I add a message about that? This would happen if they change the directory structure.

mpreiner: The CMake modules *always* ship with scikit-build, the question is where you can find them?

src/api/python/CMakeLists.txt
makaimann: The point is only to get the path to the cmake modules. But it makes sense to check that the directory exists, especially given the error message below. Added with 4123e353d578a6e712396f8c808f7aba8b99b6b0.

src/api/python/CMakeLists.txt
makaimann: Right now it ships with `scikit-build` and is always in the same place. So this is actually just guarding against future changes.

src/api/python/CMakeLists.txt
```diff
@@ -19,6 +19,37 @@ if(POLICY CMP0057)
...
+# Find cmake modules distributed with scikit-build
+# They are distributed under <scikit-build directory>/resources/cmake
+execute_process(
+  COMMAND
+    ${PYTHON_EXECUTABLE} -c "from __future__ import print_function; \
+                             import os; import skbuild; \
+                             cmake_module_path=os.path.dirname(skbuild.__file__) + \
+                             os.path.sep + 'resources' + os.path.sep + 'cmake'; \
+                             assert os.path.isdir(cmake_module_path); \
+                             print(cmake_module_path, end='')"
```
mpreiner, line 38:
```suggestion
    ${PYTHON_EXECUTABLE} -c "from __future__ import print_function; \
                             import os; import skbuild; \
                             cmake_module_path=os.path.join(os.path.dirname(skbuild.__file__),
                             'resources', 'cmake'; print(cmake_module_path, end='')"
```
I'd check the path in cmake with `if(EXISTS ...)` since you can provide a better error message

makaimann: Good idea, addressed with 8764161ad7564fb7d80555c43c940ee3d7dd9a98

src/api/python/CMakeLists.txt
```diff
@@ -19,6 +19,36 @@ if(POLICY CMP0057)
...
+check_python_module("setuptools")
+
+# Find cmake modules distributed with scikit-build
+# They are distributed under <scikit-build directory>/resources/cmake
+execute_process(
+  COMMAND
+    ${PYTHON_EXECUTABLE} -c "from __future__ import print_function; \
+                             import os; import skbuild; \
+                             cmake_module_path=os.path.join(os.path.dirname(skbuild.__file__),
+                             'resources', 'cmake'); print(cmake_module_path, end='')"
```
mpreiner, line 36: Make sure that this is 80 chars wide, you can make a newline after -c, maybe something like this:
```suggestion
    ${PYTHON_EXECUTABLE} -c \
      "from __future__ import print_function; \
      import os; import skbuild; \
      cmake_module_path=os.path.join(os.path.dirname(skbuild.__file__),
      'resources', 'cmake'); print(cmake_module_path, end='')"
```

makaimann: Good catch and thanks for the suggestion!

mpreiner (commented):
LGTM, but we should also add the required dependencies for the Python bindings to https://github.com/cvc5/cvc5/blob/master/INSTALL.md

That is, for Python bindings we need:

- cython
- scikit-build
- pytest

makaimann:
Good idea! Done with d9431c2fad665cc9bcb146f71f8ebf3843871f38. I just listed them but let me know if you'd like me to include more information (e.g., why they're needed).

mpreiner (approved):
Looks good, thanks!

PR #6544, opened 2021-05-14
Add new interface for parsing inputs

4tXJ7f:
This commit introduces the `InputParser` class, which is now the main
interface for retrieving commands and expressions from parsers for a
given input. The idea is that a `Parser` can be used to parse multiple
inputs (e.g., in the interactive shell) and that the
commands/expressions in each input can be retrieved using an
`InputParser`.

### DIFF
--- a/src/parser/input_parser.cpp
+++ b/src/parser/input_parser.cpp
@@ -1,0 +1,91 @@
+/******************************************************************************
+ * Top contributors (to current version):
+ *   Andres Noetzli
+ *
+ * This file is part of the cvc5 project.
+ *
+ * Copyright (c) 2009-2021 by the authors listed in the file AUTHORS
+ * in the top-level source directory and their institutional affiliations.
+ * All rights reserved.  See the file COPYING in the top-level source
+ * directory for licensing information.
+ * ****************************************************************************
+ *
+ * The interface for parsing an input with a parser.
+ */
+
+#include "parser/input_parser.h"
+
+#include "base/output.h"
+#include "parser/input.h"
+#include "parser/parser.h"
+#include "smt/command.h"
+
+namespace cvc5 {
+namespace parser {
+
+Command* InputParser::nextCommand()
+{
+  Debug("parser") << "nextCommand()" << std::endl;
+  Command* cmd = nullptr;
+  if (d_state->hasCommand())
+  {
+    cmd = d_state->getNextCommand();
+    d_state->setDone(cmd == nullptr);
+  }
+  else
+  {
+    try
+    {
+      cmd = d_input->parseCommand();
+      d_state->preemptCommand(cmd);
+      cmd = d_state->getNextCommand();
+      d_state->setDone(cmd == nullptr);
+    }
+    catch (ParserException& e)
+    {
+      d_state->setDone();
+      throw;
+    }
+    catch (std::exception& e)
+    {
+      d_state->setDone();
+      d_input->parseError(e.what());
+    }
+  }
+  Debug("parser") << "nextCommand() => " << cmd << std::endl;
+  return cmd;
+}
+
+api::Term InputParser::nextExpression()
+{
+  Debug("parser") << "nextExpression()" << std::endl;
+  api::Term result;
+  if (!d_state->done())
+  {
+    try
+    {
+      result = d_input->parseExpr();
+      d_state->setDone(result.isNull());
+    }
+    catch (ParserException& e)
+    {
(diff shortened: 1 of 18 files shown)


src/parser/input_parser.h
```diff
@@ -0,0 +1,69 @@
...
+ * This class is the main interface for retrieving commands and expressions
+ * from an input using a parser.
+ */
+class CVC5_EXPORT InputParser
+{
+  friend Parser;
+
+ public:
+  /** Parse and return the next command. */
+  Command* nextCommand();
```
nafur, line 45: I suspect that the caller of `nextCommand()` owns the returned `Command`. If so, please make it a `std::unique_ptr`.

src/parser/input_parser.cpp
```diff
@@ -0,0 +1,91 @@
...
+namespace parser {
+
+Command* InputParser::nextCommand()
+{
+  Debug("parser") << "nextCommand()" << std::endl;
+  Command* cmd = nullptr;
+  if (d_state->hasCommand())
+  {
+    cmd = d_state->getNextCommand();
+    d_state->setDone(cmd == nullptr);
```
nafur, line 33: This feels like `d_state` should implement this logic itself. In particular, as the logic is repeated below.

src/parser/parser.h
```diff
@@ -105,132 +105,158 @@ inline std::ostream& operator<<(std::ostream& out, SymbolType type) {
...
+   * this parser.
+   */
+  SymbolManager* d_symman;
 
- /**
-  * This current symbol table used by this parser, from symbol manager.
-  */
- SymbolTable* d_symtab;
+  /** The language that we are parsing. */
+  InputLanguage d_lang;
```
nafur, line 120: Why do we suddenly need this information here? (i.e.: where did it live beforehand?)

src/parser/parser.h
```diff
@@ -105,132 +105,158 @@ inline std::ostream& operator<<(std::ostream& out, SymbolType type) {
...
+         bool parseOnly = false);
+
+ public:
   virtual ~Parser();
 
+  /**
+   * Parse a file with this parser.
+   *
+   * @param fname The name of the file to parse.
+   * @param useMmap Whether to map the file.
```
nafur, line 225:
```suggestion
   * @param useMmap Improve reading this file using `mmap`.
```

src/parser/parser.h
```diff
@@ -105,132 +105,158 @@ inline std::ostream& operator<<(std::ostream& out, SymbolType type) {
...
+   * @return An `InputParser` that can be used to retrieve commands/expressions.
+   */
+  std::unique_ptr<InputParser> parseFile(const std::string& fname,
```
nafur, line 228: Are we going to adopt `std::filesystem`? This may be a good place to start... though maybe not in this PR.

src/parser/parser.h
```diff
@@ -105,132 +105,158 @@ inline std::ostream& operator<<(std::ostream& out, SymbolType type) {
...
+                                         bool useMmap);
+
+  /**
+   * Parse a input stream with this parser.
+   *
+   * @param name The name of the stream (used for parser errors)
```
nafur, line 234:
```suggestion
   * @param name The name of the stream (used for parser errors).
```

src/parser/parser.h
```diff
@@ -105,132 +105,158 @@ inline std::ostream& operator<<(std::ostream& out, SymbolType type) {
...
+   * @param stream The stream to parse.
+   * @return An `InputParser` that can be used to retrieve commands/expressions.
+   */
+  std::unique_ptr<InputParser> parseStream(const std::string& name,
+                                           std::istream& stream);
+
+  /**
+   * Parse a string with this parser.
+   *
+   * @param name The name of the stream (used for parser errors)
```
nafur, line 244:
```suggestion
   * @param name The name of the stream (used for parser errors).
```

src/parser/parser.h
```diff
@@ -645,6 +674,9 @@ class CVC5_EXPORT Parser
    */
   void preemptCommand(Command* cmd);
 
+  /** Get the next command from the queue. */
+  Command* getNextCommand();
```
nafur, line 678: As for InputParser: should we return a unique_ptr here?

nafur (changes requested):
Some suggestions...

ajreynol:
An alternative implementation was introduced in https://github.com/cvc5/cvc5/pull/9276, closing this.

PR #6548, opened 2021-05-17
Include cinttypes instead of inttypes.h

4tXJ7f:
This commit changes the includes used by MiniSat. This commit changes
the includes from stdint.h/inttypes.h/limits.h to
cstdint/cinttypes/climits. This ensures that the macros in
cinttypes/inttypes.h, e.g., `PRIi64`, are actually defined. The C99
standard suggested that those macros are only defined for C++ code when
`__STDC_FORMAT_MACROS` is defined. This was never adopted by a C++
standard (https://en.cppreference.com/w/cpp/types/integer). However,
certain versions of mingw-w64 seem to require it with inttypes.h but not
cinttypes.

This fixes the nightly Windows build (tested in the Docker container
used by the nightlies).

### DIFF
--- a/src/prop/bvminisat/mtl/IntTypes.h
+++ b/src/prop/bvminisat/mtl/IntTypes.h
@@ -30,12 +30,20 @@ OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWA
 
 #else
 
-#   include <stdint.h>
-#   include <inttypes.h>
+// In contrast to the original MiniSat source code, we are including the
+// cstdint/cinttypes/climits headers instead of stdint.h/inttypes.h/limits.h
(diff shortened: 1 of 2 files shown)


nafur:
Can you check whether this would also fix the issues with libpoly? (that I workarounded here: https://github.com/cvc5/cvc5/blob/master/cmake/FindPoly.cmake#L59)

nafur (approved):
LGTM.

4tXJ7f:
> Can you check whether this would also fix the issues with libpoly? (that I workarounded here: https://github.com/cvc5/cvc5/blob/master/cmake/FindPoly.cmake#L59)

We have briefly discussed this offline. The code in question introduces `PRIu64` macros whereas this PR makes those kinds of macros actually work in the MiniSat source code, so the PRs are related but this PR unfortunately does not simplify the older one.

PR #6549, opened 2021-05-17
Improve handling of `:named` attributes

4tXJ7f:
Currently, when a `:named` attribute is used in a binder, the parser
complains about an illegal argument. This is because an argument check
in the `SymbolManager` fails. This is not very user friendly. This
commit makes the error message clearer for the user:

> Cannot name a term in a binder (e.g., quantifiers, definitions)

To do this, the commit changes the return type for
`SymbolManager::setExpressionName` to include more information that can
then be used by the parser to generate an appropriate error message.

The commit also changes `define-fun` to not push/pop the local scope
if it has zero arguments because it is semantically equivalent to a
`define-const`, which allows `:named` terms.


4tXJ7f:
I've updated the PR to allow named terms in nullary define-funs.

aniemetz (approved):
LGTM but I'd like @ajreynol to maybe have a quick look, too.

ajreynol (commented):
I'm a bit hesistant regarding whether `test/regress/regress0/parser/named-attr.smt2` should be allowed.  It relies on an optimization, made in this PR, that doesn't push scope when given an empty variable list.

While I guess this optimization is sensible, it makes the rule for when we allow `:named` more complex.  It also makes things ambiguous concerning syntax: if in the regression we asserted `f`, would it be a named assertion?  In other words, do define-fun carry attribute syntax?

ajreynol:
After offline discussion, I think this PR is fine.  The fact that `test/regress/regress0/parser/named-attr.smt2` succeeds is a bit confusing conceptually, but I think the optimization to not push/pop scopes for nullary define-fun makes sense, hence this benchmark must succeed.

PR #6551, opened 2021-05-17
Adding python API test part 2

Ying1123:
This PR (follow #6546) adds more unit tests for python API.
Subsequent PR will include additional missing functions and unit tests.

PR #6552, opened 2021-05-18
Adding python API test part 3

Ying1123:
This PR(follow #6551) adds more unit tests for python API.
Subsequent PR will include additional missing functions and unit tests.


yoni206:
Thanks I'll check this once #6553 and #6551 are merged (because then the diff will be updated).

yoni206 (approved):

PR #6555, opened 2021-05-18
Generate command line options for sphinx docs

nafur:
This PR adds documentation about the command line options to the sphinx documentation. It is mostly a reformatted version of what `--help` would print.


aniemetz (commented):
Can we have all headers in upper case? For example, `Arithmetic theory module` -> `Arithmetic Theory Module`? I didn't see where this is generated.

Also, what about documentation of modes?

nafur:
> Can we have all headers in upper case? For example, `Arithmetic theory module` -> `Arithmetic Theory Module`? I didn't see where this is generated.
> 
> Also, what about documentation of modes?

These names are generated from the toml file. It's the modules `name` + `" module"`. I can make them uppercase.

And I forgot about the modes...

aniemetz (approved):

PR #6556, opened 2021-05-18
Remove `Options::set()` method

nafur:
This PR gets rid of the `Options::set()` method, replacing it by direct access to the options data.
This method was only used internally and did nothing except for resolving the options data from the option tag type via template specializations (via `ref()`), which is no longer necessary.


ajreynol (commented):
I haven't followed all of the options refactor, and this may not be specific to this PR, but would it be possible to have the module accessors not be methods?  E.g. so I can write `d_options.bv.X` instead of `d_options.bv().X`?

nafur:
> I haven't followed all of the options refactor, and this may not be specific to this PR, but would it be possible to have the module accessors not be methods? E.g. so I can write `d_options.bv.X` instead of `d_options.bv().X`?

This would be possible, but also entail other things:
- as the members need to be pointers (to allow for forward declarations), it would actually be `d_options.bv->X`
- we would expose these pointers (or rather `std::unique_ptr`), and potentially allow users of the option class to reset them
- constness doesn't play with pointers in the way we would like it to. access to a `const Options& opts` would disallow changing `opts.bv`, but changing `opts.bv->X` would be fine.

Altogether, I found wrapper methods to be the better deal for our use case.

ajreynol:
Hm, I see, I guess the second point could be addressed by duplicating the pointers, e.g. storing both raw and unique. This only has the overhead of a few extra pointers overall.  I wonder if a similar trick could allow initialization of const reference while storing a unique ptr internally.

mpreiner (approved):

PR #6557, opened 2021-05-18
(proof-new) Miscellaneous updates to strings from proof-new

ajreynol:
This includes:
(1) The type rule for `re.range` no longer insists on constant arguments, or a non-empty range. This is required for LFSC proof conversion, where re.range terms take arguments that are not (cvc5 internal) constants. 
(2) Simplifications to reductions for indexof, which fixes proof checking errors in LFSC.


4tXJ7f (approved):
LGTM

Issue #6101, opened 2021-03-10
Solution soundness bug in string logic 


jiwonparc:
commit: 4c6e0a73

```
$ cvc4 -q --strings-exp --produce-models --check-models bug.smt2
Fatal failure within void CVC4::smt::CheckModels::checkModel(CVC4::smt::Model*, CVC4::context::CDList<CVC4::NodeTemplate<true> >*, bool) 
at CVC4/src/smt/check_models.cpp:131
Internal error detectedSmtEngine::checkModel(): ERRORS SATISFYING ASSERTIONS WITH MODEL:
assertion:     (let ((_let_1 (str.replace x "A" "B"))) (let ((_let_2 (not (not (= (= "A" _let_1) false))))) 
(= (= x (str.replace x (str.replace "B" (str.replace x _let_1 "B") "B") "B")) (not (= _let_2 _let_2)))))
simplifies to: false
expected `true'.
Run with `--check-models -v' for additional diagnostics.
Aborted
$ cvc4-1.8 -q --strings-exp --produce-models --check-models bug.smt2
unsat
$ z3release bug.smt2
unsat
$ cat bug.smt2
(declare-fun x () String)
(declare-fun y () String)
(declare-fun z () Int)
(assert (= (= x (str.replace x (str.replace "B" (str.replace x (str.replace x "A" "B") "B") "B") "B")) (not (= 
(not (not (= (= "A" (str.replace x "A" "B")) false))) (not (not (= (= "A" (str.replace x "A" "B")) false)))))))
(check-sat)
```

wintered:
Somewhat more reduced:

```
$cvc4 -q --strings-exp bug.smt2
sat
$z3release bug.smt2
unsat
$cat bug.smt2 
(declare-fun x () String)
(assert (not (= x (str.replace x (str.replace "B" (str.replace x (str.replace x "A" "B") "B") "B") "B"))))
(check-sat)
```

PR #6562, opened 2021-05-18
Fix `collectEmptyEqs()` in string utils

### LINKED ISSUE #6483: solution unsoundness on a string formula
Commit: https://github.com/CVC4/CVC4/commit/441c53b1a68cc16a345eb0dc8d9956c1301ed509
OS: Ubuntu 18.04 

It also affects cvc4-1.7 and cvc4-1.8, but both can detect the produced invalid model.

```
[619] % z3release small.smt2
unsat
(error "line 14 column 10: model is not available")
[620] % cvc5 -q --strings-exp --produce-models --check-models small.smt2
sat
(
(define-fun a () String "")
)
[621] % 
[621] % cat small.smt2
(declare-fun a () String)
(assert
 (xor
  (= (= (str.replace "B" (str.replace (str.++ (str.replace "B" a "B") a) "B" "") "B")
      (str.replace "B" (str.replace "B" a (str.++ a "B")) a))
   (not
    (= (str.replace "B" (str.replace a "B" "") "B")
     (str.replace "B" a (str.++ a "B")))))
  (= (str.replace "B" a "")
   (str.replace "B" a
    (ite (not (= (str.replace "" (str.replace a "" a) "") "")) ""
     (str.replace "" (str.replace a "B" "") "B"))))))
(check-sat)
(get-model)
[622] % 
```

4tXJ7f:
Fixes #6483. The benchmark in the issue was performing the following
incorrect rewrite:

```
 Rewrite (str.replace "B" (str.replace (str.++ (str.replace "B" a "B") a) "B" "") "B") to (str.replace "B" a "B") by RPL_X_Y_X_SIMP.
```

The rewrite `RPL_X_Y_X_SIMP` rewrites terms of the form `(str.replace x
y x)`, where `x` is of length one and `(= y "")` rewrites to a
conjunction of equalities of the form `(= y_i "")` where `y_i` is some
term. The function responsible for collecting the terms `y_i` from this
conjunction, `collectEmptyEqs()`, returns a `bool` and a vector of
`Node`s. The `bool` indicates whether all equalities in the conjunction
were of the form `(= y_i "")`. The rewrite `RPL_X_Y_X_SIMP` only applies
if this is true. However, `collectEmptyEqs()` had a bug where it would
not return false when all of the conjuncts were equalities but not all
of them were equalities with the empty string. This commit fixes
`collectEmptyEqs()` and adds tests.


ajreynol:
Does this PR also fix https://github.com/cvc5/cvc5/issues/6101 ?

ajreynol (approved):
LGTM, great find.

4tXJ7f:
> Does this PR also fix #6101 ?

Unfortunately, it does not look like it. The rewrite fixed in this PR does not apply very often.

PR #6563, opened 2021-05-18
Fix handling of non-standard re.range terms

### LINKED ISSUE #6561: (default) solution unsoundness and Fatal failure at src/util/string.cpp:241 on string formulas with re.range
Commit: 626a163
OS: Ubuntu 18.04

```
[571] % cvc5 -q small.smt2
Fatal failure within unsigned int cvc5::String::front() const at /local/suz-local/software/CVC4/src/util/string.cpp:241
Check failure

 !d_str.empty()

Aborted
[572] % 
[572] % cat small.smt2
(declare-fun a () String)
(assert (ite (str.in_re (str.++ a "c") (re.* (re.range "a" ""))) (not (= a "")) (str.in_re a (re.* (re.range a "")))))
(check-sat)
[573] % 
```

ajreynol:
Fixes #6561. That benchmark now gives an error:
`(error "expecting a single constant string term in regexp range")`.

This PR also makes `isConstRegExp` do a non-recursive dag traversal.


src/theory/strings/regexp_entail.cpp
```diff
@@ -328,21 +333,45 @@ Node RegExpEntail::simpleRegexpConsume(std::vector<Node>& mchildren,
...
-  for (unsigned i = 0; i < t.getNumChildren(); ++i)
-  {
-    if (!isConstRegExp(t[i]))
-    {
-      return false;
+  std::unordered_set<TNode> visited;
+  std::vector<TNode> visit;
+  TNode cur;
+  visit.push_back(t);
+  do {
```
4tXJ7f, line 340: Apply ClangFormat.

4tXJ7f (approved):
LGTM, please fix the formatting before merging.

PR #6564, opened 2021-05-18
Avoid using printSynthSolution in the python API and examples

yoni206:
The function `printSynthSolution` [declared here](https://github.com/cvc5/cvc5/blob/c214051068aefdf831bf67e6b7d72591e5a91ece/src/api/cpp/cvc5.h#L3988) is going to be removed in #6521.

This PR removes it from the python API.
Following #6530, this PR also replaces its usages in the examples by a utility function.
For this, we also add support for `getSynthSolutions` in the python API.


yoni206:
FYI @ajreynol

examples/api/python/utils.py
```diff
@@ -0,0 +1,55 @@
...
+#
+# Utility Methods, translated from examples/api/utils.h
+##
+
+import pycvc5
+from pycvc5 import kinds
+
+# Get the string version of define-fun command.
+# @param f the function to print
+# @param params the function parameters
```
makaimann, line 22: Missing a param for `sort`. Speaking of which, can't you get the return sort from the function `f`?

yoni206: > Missing a param for sort
Thanks! fixed here: 960fbb2

> can't you get the return sort from the function f?
I am not sure, I translated this from the cpp example. @ajreynol do you have any preference here? Either way I recommend keeping the cpp and python similar in this case.

ajreynol: I think its fine to remove the sort from both. Indeed it is redundant.

yoni206: done here: 8f1d614

makaimann (approved):
LGTM, thanks for the PR and the updates!

PR #6566, opened 2021-05-18
Fix positive contains indexof rewrites for empty string second argument

### LINKED ISSUE #6560: (default) solution unsoundness on string formula with str.indexof
Commit: https://github.com/cvc5/cvc5/commit/626a1634fc21aed6e1149e5579ce112c1efc72bc
OS: Ubuntu 18.04

```
[592] % z3release small.smt2
unsat
unsat
[593] % cvc5 -q --strings-exp --dump-models --check-models small.smt2
sat
(
(define-fun a () String "AA")
)
unsat
[594] % 
[594] % cat small.smt2
(declare-fun a () String)
(assert (> (str.indexof a "" (* 2 (str.len a))) 0))
(check-sat)

(reset)

(define-fun a () String "AA")
(assert (> (str.indexof a "" (* 2 (str.len a))) 0))
(check-sat)
[595] % 
```

ajreynol:
Fixes #6560.


4tXJ7f (approved):
LGTM, thanks for the fix.

PR #6570, opened 2021-05-19
Fix strings rewriter for non-standard re range

### LINKED ISSUE #6561: (default) solution unsoundness and Fatal failure at src/util/string.cpp:241 on string formulas with re.range

ajreynol:
Fixes #6561.

Previously we missed a case where a rewrite fired for (re.range x x) where x is not a character.


4tXJ7f (approved):
LGTM

PR #6572, opened 2021-05-19
Make output list of `mkoptions.py` more accurate

4tXJ7f:
After commit 6dc5b7469cee015a3bcf25a1543123da6c8317fe, cvc5 was always
being almost completely rebuilt, even if there hadn't been any changes
if cvc5 was configured _not_ to produce documentation (the default).
This was because `mkoptions.py` only produces the
`options_generated.rst` file when documentation is enabled. However, it
was unconditionally declared to be an output of the script in
`CMakeLists.txt`. As a result, the options (and thus most of the code
base) were rebuilt every time because the file was missing in builds
without documentation. This commit modifies the output list depending on
the configuration.


nafur (approved):
LGTM. I'm still somewhat amazed about this issue...

PR #6573, opened 2021-05-19
Fix echo printing.

aniemetz:
Previously, echo surpressed leading, trailing and escape quotes of the
string to print. However, the SMT-LIB standard states that the string is
to be printed as is, including those quote characters.


test/regress/regress0/echo.smt2
```diff
@@ -0,0 +1,16 @@
+; EXPECT: "str"""
+; EXPECT: "str""ing"
+; EXPECT: "hi"
+; EXPECT: "str""""ing"
+; EXPECT: "str""i""ng"
+; EXPECT: "str\ing"
+; EXPECT: "str\\ing"
+; EXPECT: "str\"
```
ajreynol, line 8: Can you add one more test here: `(echo "\u{65}")`, `\u{...}` is a string literal escape sequence, which I guess should not be processed and hence printed back verbatim.

ajreynol (approved):
LGTM, one request.

test/regress/regress0/echo.smt2
aniemetz: Done

PR #6575, opened 2021-05-19
Make strings emp inference an unhandled inference for proofs

ajreynol:
The strings inference id STRINGS_INFER_EMP is a rare inference that currently leads to a segfault when proof reconstruction is tried for it.  This PR makes it unhandled.

This was caught by a new regression on the soon-to-be-merged `jh-new` branch.


4tXJ7f (approved):
LGTM

PR #6577, opened 2021-05-19
Expand arith's farkas lemma rule as a macro

alex-ozdemir:
reflects that it is a macro, which we will eliminate

### DIFF
--- a/src/smt/proof_post_processor.cpp
+++ b/src/smt/proof_post_processor.cpp
@@ -22,7 +22,6 @@
 #include "preprocessing/assertion_pipeline.h"
 #include "smt/smt_engine.h"
 #include "smt/smt_statistics_registry.h"
-#include "theory/arith/arith_utilities.h"
 #include "theory/builtin/proof_checker.h"
 #include "theory/rewriter.h"
 #include "theory/theory.h"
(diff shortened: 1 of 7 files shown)


src/smt/proof_post_processor.cpp
```diff
@@ -1028,6 +1029,58 @@ Node ProofPostprocessCallback::expandMacros(PfRule id,
...
+      Node impl =
+          steps.tryStep(isPos ? PfRule::ARITH_MULT_POS : PfRule::ARITH_MULT_NEG,
+                        {},
+                        {scalar, child});
+      Assert(!impl.isNull());
+      // scaled
+      Node scaled =
+          steps.tryStep(PfRule::MODUS_PONENS, {scalarCmpAndRel, impl}, {});
+      Assert(!scaled.isNull());
+      scaledRels.emplace_back(std::move(scaled));
```
ajreynol, line 1075: Is this std::move necessary?

src/smt/proof_post_processor.cpp
```diff
@@ -22,6 +22,7 @@
 #include "preprocessing/assertion_pipeline.h"
 #include "smt/smt_engine.h"
 #include "smt/smt_statistics_registry.h"
+#include "theory/arith/arith_utilities.h"
```
ajreynol, line 25: What is this included for?

ajreynol (commented):
A few questions

src/smt/proof_post_processor.cpp
alex-ozdemir: Ah, good catch. While (in general) this is needed to move rather than copy, we don't implement move construction for Node, so it doesn't do anything here. Moving wouldn't help *that* much anyway for a node. It would just save a few refcount accesses.

I guess I'll trim this.

src/smt/proof_post_processor.cpp
alex-ozdemir: Hm, I guess it is unneeded. Good catch.

ajreynol (approved):
LGTM, thanks for the updates.

PR #6580, opened 2021-05-19
Support braced-init-lists with `mkNode()`

4tXJ7f:
This commit adds support for braced-init-lists in calls to `mkNode()`,
e.g., `mkNode(REGEXP_EMPTY, {})`. Previously, such a call would result
in a node of kind `REGEXP_EMPTY` with a single null node as a child
because the compiler chose the `mkNode(Kind kind, TNode child1)` variant
and converted `{}` to a node using the default constructor. This commit
adds an overload of `mkNode()` that takes an `initializer_list<TNode>`
to allow this use case. It also makes the default constructor of
`NodeTemplate` explicit to avoid such accidential conversions.


ajreynol (approved):

src/expr/node_manager.cpp
```diff
@@ -1189,4 +1189,11 @@ Kind NodeManager::getKindForFunction(TNode fun)
   return kind::UNDEFINED_KIND;
 }
 
+Node NodeManager::mkNode(Kind kind, std::initializer_list<TNode> children)
```
nafur, line 1192: Are you sure it is safe to take `TNode` here?

src/expr/node_manager.h
```diff
@@ -458,6 +458,17 @@ class NodeManager
...
   template <bool ref_count>
   Node mkNode(Kind kind, const std::vector<NodeTemplate<ref_count> >& children);
+
+  /** Create a node using an initializer list.
+   *
+   * This function serves two purposes:
+   * - We can avoid creating a temporary vector in some cases, e.g. when we
+   *   want to create a node with zero children
+   * - It makes sure that calls to `mkNode` that use braced-init-lists work as
+   *   expected, e.g., mkNode(REGEXP_EMPTY, {})
```
nafur, line 468: If the motivation is a node with zero children, why don't we have `mkNode(Kind)`?
I'd rather see this new overload as an alternative to `mkNode()` with 1,2,3,4,5 children...

nafur (commented):
Some comments...

src/expr/node_manager.h
aniemetz: I agree.

4tXJ7f: I see the point regarding `{}` and I am happy to add another overload for it. However, the idea was that this function adds a way of creating a node with an arbitrary number of children without the need of creating a temporary vector (C-style variadic functions are [not recommended](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-ellipses)). I have considered using C++17 [fold expressions](https://en.cppreference.com/w/cpp/language/fold) instead of defining all these different variants but it adds about 3 minutes of CPU time to recompiling cvc5.

src/expr/node_manager.cpp
4tXJ7f: Yes, it should be ok. The caller ensures that the nodes are alive. I have modified a unit test to pass a temporary node in an init list as a test.

4tXJ7f (commented):
I have added a nullary `mkNode` as suggested (and updated some of the callers accordingly). Additionally, I have removed the 4- and 5-node `mkNode` variants after an offline discussion with Gereon. Most uses of these variants were in tests. In the actual code, there were only three uses, which I have changed to use initializer lists instead.

nafur (approved):
LGTM.

PR #6581, opened 2021-05-20
Remove old unsat cores

HanielB:
This commit removes the remaining old proof code and the code to produce unsat cores based on it.


src/prop/minisat/minisat.h
```diff
@@ -22,6 +22,14 @@
 #include "util/statistics_registry.h"
 
 namespace cvc5 {
+
+template <class Solver>
```
ajreynol, line 26: Are these going to be deleted eventually as well?  Or why are they currently needed?

ajreynol (commented):
Just one question, else LGTM.

src/prop/minisat/minisat.h
HanielB: I had to add them here because they were previously defined in the deleted sat proof files. We currently need them because of the existence of BVMinisat. Since we want to remove that eventually, these could go at that time as well.

ajreynol: Ok cool, that's what I expected.

src/preprocessing/passes/non_clausal_simp.cpp
```diff
@@ -69,7 +69,7 @@ NonClausalSimp::NonClausalSimp(PreprocessingPassContext* preprocContext)
 PreprocessingPassResult NonClausalSimp::applyInternal(
     AssertionPipeline* assertionsToPreprocess)
 {
-  Assert(options::unsatCoresMode() != options::UnsatCoresMode::OLD_PROOF
+  Assert(options::unsatCoresMode() == options::UnsatCoresMode::OFF
```
ajreynol, line 72: I think this assertion should just be deleted now.

HanielB: Indeed. Done.

ajreynol (approved):

PR #6591, opened 2021-05-20
Formalize shared selectors as skolem functions

ajreynol:
This is work towards properly printing shared selectors in external proofs.


src/expr/skolem_manager.h
```diff
@@ -231,7 +233,8 @@ class SkolemManager
    */
   Node mkSkolemFunction(SkolemFunId id,
                         TypeNode tn,
-                        Node cacheVal = Node::null());
+                        Node cacheVal = Node::null(),
+                        int flags = NodeManager::SKOLEM_DEFAULT);
```
HanielB, line 237: use unsigned?

ajreynol: This is to match `NodeManager::` flags.  We shouldn't mix types here until those flags are fully refactored at once. (Note that `int flags` appears as arguments to other methods in this class for the same reason.

HanielB: I see, sounds good

HanielB (approved):
LGTM

PR #6593, opened 2021-05-20
Fix tests of unsat cores

ajreynol:
This updates all regressions that pass check-unsat-cores to enable check-unsat-cores.  This includes *any* incremental benchmark, which was disabled in run_regression.py previously.

It adds --no-check-unsat-cores to a few corner benchmarks that were previously disabled based on --incremental.

It also reverts a change to when proofs are disabled: options like `sygus-inference` should not permit proofs (or unsat cores).


HanielB (approved):
LGTM

PR #6597, opened 2021-05-21
Use references instead of getter functions for options

nafur:
This PR follows a suggestions of @ajreynol to use public references instead of getter functions to access the individual option modules.


ajreynol (commented):
This looks good to me, although I'd prefer @mpreiner review since I'm not an expert on the mkoptions.py code.

mpreiner (approved):

PR #6600, opened 2021-05-21
Move public wrapper functions out of options class

nafur:
This PR moves options wrapper functions out of the `Options` class. These wrapper functions are meant to be called by "external" code that should not access the options modules. This PR thereby significantly reduces the interface of the `Options` class.


nafur:
> What's the plan with `options_public.h`? Will this be exported via the public API or is this a semi public interface for the parser and main driver?

My plan is to have `options_public.h` as the semi public interface to stuff outside of `libcvc5`, i.e., the parser and the main driver. In the API, we have `Solver::getOption()` and `Solver::setOption()`.

ajreynol (approved):
LGTM

PR #6601, opened 2021-05-21
Reduce size of sphinx-gh output

nafur:
This PR reduces the disk size of the docs generated by `make sphinx-gh`.
Apart from reformatting the cmake source, we now not only remove the `_sources` folder, but also `.doctrees` (essentially the sphinx cache) and `_static/fonts/` (the fonts that are actually used live in `_static/css/fonts`).
In combination, this now reduces the disk size from about 20MB (`sphinx`) to less than 6MB (`sphinx-gh`).
Furthermore this PR only uploads the generated documentation if it differs from whatever we currently have for `master`.
This is relevant to make the `docs-ci` repository smaller (which already has more than 5GB...)


aniemetz:
@nafur I just built the docs for GH pages and it's only 9.8M for me. Why is it 20M for you?

nafur:
> @nafur I just built the docs for GH pages and it's only 9.8M for me. Why is it 20M for you?

It was also ~10M for me, until I upgraded to the latest sphinx version. With some recent sphinx version, the way fonts are handled has changed and way more files end up in the build folder (that this PR is getting rid of). The CI job actually produces about 20M...

aniemetz (approved):

PR #6609, opened 2021-05-24
Fix re-elim length requirement for symbolic RE memberships

### LINKED ISSUE #6604: (re-elim) solution unsoundness on string formula
Commit: https://github.com/cvc5/cvc5/commit/0e9fed387aaf5d0cb1844ab20551fbebfcb8893d
OS: Ubuntu 18.04 

```
[510] % cvc5 -q small.smt2 
unsat
[511] % cvc5 -q --strings-exp --re-elim small.smt2 
sat
[512] % cat small.smt2 
(declare-fun a () String)
(assert (str.in_re a (re.++ re.allchar (str.to_re a) (re.* re.allchar))))
(check-sat)
[513] % 
```

ajreynol:
Fixes #6604.

Previously, re-elim was solution unsound for cases where the LHS and a component of the RHS were both empty. This ensures a length requirement is given for the LHS to ensure proper containment.


4tXJ7f (approved):
Nice catch! LGTM

PR #6611, opened 2021-05-24
Move proof utilities to src/proof/

ajreynol:
This moves all generic proof utilites from src/expr/ and src/theory/ to src/proof/.

It also changes the include for term conversion proof generator to `conv_proof_generator` in preparation to rename this utility on a followup PR (to avoid confusion with the use of "Term").


HanielB (approved):
LGTM

PR #6614, opened 2021-05-24
Update proof namespaces

ajreynol:
This removes `namespace theory` from proof utilities, and moves `MethodId` to its own file in `src/proof/`.


HanielB (approved):
LGTM

PR #6617, opened 2021-05-24
 More precise includes of `Node` constants

4tXJ7f:
We store constants, e.g., `BitVector` and `Rational`, in our node infrastructure. As a result, we were indirectly including some headers in almost all files, e.g., the GMP headers. This commit changes that by forward-declaring the classes for the constants. As a result, we have to include headers like `util/rational.h` explicitly when we use `Rational` but it saves about 3 minutes in compile time (CPU time).

The commit changes `RoundingMode` from an `enum` to an `enum class` such that it can be forward declared.


src/preprocessing/passes/bv_to_int.h
```diff
@@ -73,6 +73,7 @@
 #include "context/cdhashset.h"
 #include "preprocessing/preprocessing_pass.h"
 #include "theory/arith/nl/iand_utils.h"
+#include "util/rational.h"
```
mpreiner, line 76: Rational doesn't seem to be used in this header, include shouldn't be needed here.

src/theory/bv/kinds
```diff
@@ -13,7 +13,8 @@ properties check propagate presolve ppStaticLearn
 rewriter ::cvc5::theory::bv::TheoryBVRewriter "theory/bv/theory_bv_rewriter.h"
 
 constant BITVECTOR_TYPE \
-	::cvc5::BitVectorSize \
+  struct \
+	BitVectorSize \
```
nafur, line 17:
```suggestion
      struct \
	BitVectorSize \
```

src/theory/bv/kinds
```diff
@@ -101,56 +103,64 @@ operator BITVECTOR_EAGER_ATOM 1 "formula to be treated as a bv atom via eager bi
 ### parameterized operator kinds ----------------------------------------------
 
 constant BITVECTOR_BITOF_OP \
-	::cvc5::BitVectorBitOf \
+  struct \
+	BitVectorBitOf \
```
nafur, line 107:
```suggestion
      struct \
	BitVectorBitOf \
```

nafur: and more below...

nafur (changes requested):
A few minor comments.

src/expr/type_node.cpp
```diff
@@ -684,6 +687,18 @@ TypeNode TypeNode::getBagElementType() const
...
+  return (getKind() == kind::BITVECTOR_TYPE
+          && getConst<BitVectorSize>() == size);
+}
+
+unsigned TypeNode::getBitVectorSize() const
+{
+  Assert(isBitVector());
+  return getConst<BitVectorSize>();
+}
+
```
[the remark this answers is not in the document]
aniemetz: I agree, it would be good to change this in this PR.

mpreiner (approved):

4tXJ7f:
Thanks for the comments. I have addressed them and updated my vim config to show me tabs as characters ;)

nafur (approved):
LGTM.

PR #6626, opened 2021-05-27
Add Lexicographic + Pareto Optimizations

Ouyancheng:
Lexicographic optimizations: Optimize the objectives one-by-one, in the order they are pushed. 
Pareto optimizations: Optimize the objectives to the extend that further optimization on any objective will worsen the other objective. 
Units tests are of course added. 

Lexicographic optimization is using iterative implementation, pretty similar to the naive box optimization.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -155,6 +155,8 @@ libcvc5_add_sources(
   proof/lazy_proof_chain.h
   proof/lazy_tree_proof_generator.cpp
   proof/lazy_tree_proof_generator.h
+  proof/method_id.cpp
+  proof/method_id.h
   proof/proof.cpp
   proof/proof.h
   proof/proof_checker.cpp
--- a/src/preprocessing/assertion_pipeline.cpp
+++ b/src/preprocessing/assertion_pipeline.cpp
@@ -82,9 +82,9 @@ void AssertionPipeline::push_back(Node n,
   }
 }
 
-void AssertionPipeline::pushBackTrusted(theory::TrustNode trn)
+void AssertionPipeline::pushBackTrusted(TrustNode trn)
 {
-  Assert(trn.getKind() == theory::TrustNodeKind::LEMMA);
+  Assert(trn.getKind() == TrustNodeKind::LEMMA);
   // push back what was proven
   push_back(trn.getProven(), false, false, trn.getGenerator());
 }
@@ -105,14 +105,14 @@ void AssertionPipeline::replace(size_t i, Node n, ProofGenerator* pgen)
   d_nodes[i] = n;
 }
 
-void AssertionPipeline::replaceTrusted(size_t i, theory::TrustNode trn)
+void AssertionPipeline::replaceTrusted(size_t i, TrustNode trn)
 {
   if (trn.isNull())
   {
     // null trust node denotes no change, nothing to do
     return;
   }
-  Assert(trn.getKind() == theory::TrustNodeKind::REWRITE);
+  Assert(trn.getKind() == TrustNodeKind::REWRITE);
   Assert(trn.getProven()[0] == d_nodes[i]);
   replace(i, trn.getNode(), trn.getGenerator());
 }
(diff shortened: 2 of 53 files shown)


src/smt/optimization_solver.cpp
```diff
@@ -31,18 +31,26 @@ OptimizationSolver::OptimizationSolver(SmtEngine* parent)
       d_optChecker(),
       d_objectives(),
       d_results(),
-      d_objectiveCombination(BOX)
+      d_objectiveCombination(LEXICOGRAPHIC)
```
yoni206, line 34: Why is `LEXICOGRAPHIC` the default mode? Please either document this fact in the `h` file, or add another mode `UNSET`, and then assert that the mode is not `UNSET` where you actually compute the optimized values. Otherwise the user may be unaware of this decision.

src/smt/optimization_solver.h
```diff
@@ -250,6 +250,32 @@ class OptimizationSolver
...
+   *   UNKNOWN if one of the objectives is UNKNOWN
+   *     and optimization will stop at that objective;
+   *   If the optimization is stopped at an objective,
+   *     all objectives following that objective will be UNKNOWN.
+   **/
+  OptimizationResult::ResultType optimizeLexicographicIterative();
+
+  /**
+   * Optimize multiple goals in Pareto order
+   * Using a variant of linear search called Guided Improvement Algorithm
```
yoni206, line 270: Where is this algorithm taken from? Please add a reference (e.g. paper citation).

src/smt/optimization_solver.h
```diff
@@ -250,6 +250,32 @@ class OptimizationSolver
...
+   * Could be called multiple times to iterate through all possible solutions
```
yoni206, line 271: Please clarify in the comment why does it make sense to call it multiple times, and why we won't get the same result each time.

src/smt/optimization_solver.h
```diff
@@ -250,6 +250,32 @@ class OptimizationSolver
...
+   * (Pareto front)
```
yoni206, line 272: It is not clear what is the Pareto front. Please clarify in the comment.

src/smt/optimization_solver.h
```diff
@@ -250,6 +250,32 @@ class OptimizationSolver
...
+   * @return if there's a result it will return OPTIMAL;
+   *   if it exhausts the results in the Pareto front it will return UNSAT;
+   *   if it doesn't know whether there's a result it will return UNKNOWN.
```
yoni206, line 275: Please specify when will this happen.

src/smt/optimization_solver.cpp
```diff
@@ -150,5 +158,152 @@ OptimizationResult::ResultType OptimizationSolver::optimizeBox()
...
+      }
+      default: Unreachable();
+    }
+  }
+
+  d_optChecker->pop();
+
+  // before we return!
+  // please do assert that some objective could be better
+  // in order to break the ties for the next run!!!
```
yoni206, line 302:
```suggestion
  // before we return:
  // assert that some objective could be better
  // in order to break the ties for the next run.
```

yoni206, line 302: Please clarify "break the ties" in the comment.

yoni206 (changes requested):
Thanks for the PR. Comments inline.

src/smt/optimization_solver.cpp
Ouyancheng: Lexico is more intuitive than Box. I will add some comment on it.

src/smt/optimization_solver.h
Ouyancheng: ? 
I deliberately make it to iterate... 
And it's natural to have multiple Pareto optimal solutions.

src/smt/optimization_solver.h
Ouyancheng: all possible solutions (Pareto front)

src/smt/optimization_solver.h
Ouyancheng: added

src/smt/optimization_solver.cpp
Ouyancheng: commented in checkOpt

yoni206 (approved):
LGTM, Thanks for the changes.

PR #6628, opened 2021-05-27
FP: Rename FLOATINGPOINT_PLUS to FLOATINGPOINT_ADD.

aniemetz:
This is to make it consistent with the name of the SMT-LIB operator
(fp.add).


mpreiner (approved):

martin-cs (commented):
Looks good, sorry for the confusion.  I fear the implementation (or at least that bit of it) pre-dates the standardisation of the name.  Anyway, thanks for fixing.

PR #6615, opened 2021-05-24
Fix `str.replace_re` and `str.replace_re_all`

4tXJ7f:
Fixes #6057. The reductions of `str.replace_re` and `str.replace_re_all`
were not correctly enforcing that the operations replace the _first_
occurrence of some regular expression in a string. This commit fixes the
issue by introducing a new operator `str.indexof_re(s, r, n)`, which,
analoguously to `str.indexof`, returns the index of the first match of
the regular expression `r` in `s`. The commit adds basic rewrites for
evaluating the operator as well as its reduction. Additionally, it
converts the reductions of `str.replace_re` and `str.replace_re_all` to
use that new operator. This simplifies the reductions of the two
operators and ensures that the semantics are consistent between the two.

PR #6631, opened 2021-05-28
Compute model values for nested sequences in order

### LINKED ISSUE #5665: Invalid model for the formula with Seq
```
[556] % cvc4 --strings-exp --check-models -q small.smt2
Fatal failure within void CVC4::smt::CheckModels::checkModel(CVC4::smt::Model*, CVC4::context::CDList<CVC4::NodeTemplate<true> >*, bool) at src/smt/check_models.cpp:249
assertion:   (distinct a b (str.++ (seq.unit d) c))
simplifies to: false
expected `true'.
Aborted
[557] % 
[557] % cat small.smt2
(declare-fun a () (Seq (Seq Int)))
(declare-fun b () (Seq (Seq Int)))
(declare-fun c () (Seq (Seq Int)))
(declare-fun d () (Seq Int))
(declare-fun e () Int)
(declare-fun f () Int)
(assert (= (seq.extract d e 1) (seq.unit f)))
(assert (distinct a b (seq.++ (seq.unit d) c)))
(check-sat)
[558] %
```

Commit: a49efc8

### LINKED ISSUE #6337: Invalid model for string formulas
cvc4 52b6d736220
~~~~
(declare-fun str3 () String)
(declare-fun str8 () String)
(declare-fun str12 () String)
(declare-fun str14 () String)
(declare-fun str15 () String)
(assert (distinct str15 (str.++ str14 str3 str8 str14) (str.replace_re_all str12 (re.comp (str.to_re str15)) (str.++ str14 str3 str8 str14)) str12))
(check-sat)
~~~~
~~~~
cvc4 -q -i --strings-exp --check-models delta.out.smt2 
Fatal failure within void cvc5::smt::CheckModels::checkModel(cvc5::smt::Model*, cvc5::context::CDList<cvc5::NodeTemplate<true> >*, bool) at /home/CVC4/src/smt/check_models.cpp:132
assertion:     (let ((_let_1 (str.++ str14 str3 str8 str14))) (distinct str15 _let_1 (str.replace_re_all str12 (re.comp (str.to_re str15)) _let_1) str12))
simplifies to: false
expected `true'.
Aborted (core dumped)
~~~~

4tXJ7f:
Fixes #6337 (the other benchmarks in this issue are either solved
correctly or time out after the changes in #6615) and fixes #5665.

While computing the model for a nested equivalence class containing
`seq.unit`, we were looking up the representative of the argument in
`(seq.unit (seq.unit j))` and the representative was simpliy `(seq.unit
j)`. However, we had assigned `(seq.unit 0)` to `(seq.unit j)` earlier.
A second equivalence class of type `(Seq (Seq Int))` and length 1 was
later assigned `(seq.unit (seq.unit 0))` and we didn't detect that
`(seq.unit (seq.unit j))` and `(seq.unit (seq.unit 0))` have the same
value. This was incorrect because we do not allow assigning the same
value to different equivalence classes. In this case, it led to one of
the assertions being false.

This commit fixes the issues in two ways: it ensures that types are
processed in ascending order of nesting (e.g., `(Seq Int)` terms are
processed before `(Seq (Seq Int))` terms) and it changes the procedure
to look up the representative in the model instead of the theory state
to take into account the model values assigned to the elements of
sequences.

cc @yoni206


src/theory/strings/theory_strings.cpp
```diff
@@ -198,6 +198,34 @@ void TheoryStrings::presolve() {
 // MODEL GENERATION
 /////////////////////////////////////////////////////////////////////////////
 
+/**
+ * Helper that allows ordering sequence types by how nested they are.
```
ajreynol, line 202: I'd prefer not adding this utility.

Can't we just have a simple policy in the code along the lines of:  to compute model values for a type T, first check if it is (Seq T1), if so, compute model values for T1 first?

ajreynol (changes requested):

ajreynol (approved):
LGTM, nice find.

PR #6634, opened 2021-05-28
(Optimization) remove popObjective, add resetObjectives, rename pushObjective => addObjective

Ouyancheng:
In order for OptimizationSolver to support pushing & popping, 
we could remove popObjective because it might be difficult to handle cases like: 
```
optSlv->pushObjective(...); 
optSlv->push();
optSlv->popObjective();
optSlv->pop();
```
In this case we need to add back the popped objective... 
If push/pop is supported, pop does not bring back objectives if you resetObjective
but it will revert the objs you add.
just like assertFormula and resetAssertions.


yoni206:
The changes look good to me, just the PR description (that will appear in the commit message) is not clear.
Please make it clearer.
For example:
```
This PR handles the addition of objectives internally in the optimization module, instead than relying on
the smt engine's `resetAssertions`.
```

yoni206:
The example I gave is actually sufficient, no need to include anything else in the PR description.

yoni206 (approved):
Thanks for the updated message.

PR #6640, opened 2021-05-30
Some additions to the datatypes python API

yoni206:
This PR makes the following additions, in order to sync the python API with the cpp API.

1. adding `getName` functions to datatypes related classes
2. allowing `mkDatatypeSorts` with 1 or 2 arguments (previously allowed only 2).
3. In case there is a second argument to `mkDatatypeSorts`, we make sure it is a set.
4. Corresponding changes to the tests.


Ying1123 (approved):

src/api/python/cvc5.pxi
```diff
@@ -122,6 +122,10 @@ cdef class Datatype:
         ds.cds = self.cd.getSelector(name.encode())
         return ds
 
+    def getName(self):
+        return self.cd.getName().decode()
+
+
```
4tXJ7f, line 128:
```suggestion
```

src/api/python/cvc5.pxi
```diff
@@ -501,9 +508,19 @@ cdef class Solver:
         sort.csort = self.csolver.mkDatatypeSort(dtypedecl.cdd)
         return sort
 
-    def mkDatatypeSorts(self, list dtypedecls, unresolvedSorts):
+    def mkDatatypeSorts(self, *args):
```
4tXJ7f, line 511: Since the only two options are to have one or two arguments, it would be better to use a default argument. Something like:

```
def mkDatatypeSorts(self, list dtypedecls, set unresolvedSorts = None)
```

and then you can check whether `unresolvedSorts is None`.

4tXJ7f (changes requested):
Thanks for doing this. Just a minor request.

src/api/python/cvc5.pxi
yoni206: Yes, that's better, thanks!
Fixed here: 723ecc0da26e5ecfa26d4e49a7b943acedd6aee5

While fixing this, I found the need to distinguish between cpp sets (now called `c_set`s) and python sets. Done here: https://github.com/cvc5/cvc5/pull/6640/commits/4c4b697e9564c454f58acec1c7de98da813feaf4

test/python/unit/api/test_datatype_api.py
```diff
@@ -209,6 +215,10 @@ def test_datatype_names(solver):
     with pytest.raises(RuntimeError):
         dt.getSelector("cons")
 
+    # possible to construct null datatype declarations if not using solver
```
4tXJ7f, line 218:
```suggestion
    # possible to construct null datatype declarations if not using mkDatatypeDecl
```

(I was confused at first because you are actually using `solver`)

src/api/python/cvc5.pxi
```diff
@@ -501,19 +507,28 @@ cdef class Solver:
...
         return sort
 
-    def mkDatatypeSorts(self, list dtypedecls, unresolvedSorts):
-        sorts = []
+    def mkDatatypeSorts(self, list dtypedecls, unresolvedSorts = None):
+        '''
+        Supports the following arguments:
+        std::vector<Sort> mkDatatypeSorts(const std::vector<DatatypeDecl>& dtypedecls)
+        std::vector<Sort> mkDatatypeSorts(const std::vector<DatatypeDecl>& dtypedecls, const std::c_set<Sort>& unresolvedSorts)
+        '''
```
4tXJ7f, line 515:
```suggestion
```

I think that we will need a different kind of comment here. Something similar to https://github.com/cvc5/cvc5/blob/320c48b1c82feb7a5d55af39d0cdbb7a1c2c6ff2/src/api/python/cvc5.pxi#L127 (can be done in a future PR).

4tXJ7f (approved):
LGTM. Thanks for the updates. Some minor suggestions for the comments.

test/python/unit/api/test_datatype_api.py
yoni206: done

src/api/python/cvc5.pxi
yoni206: done (please see new comment)

PR #6644, opened 2021-05-31
FP value support in python API

yoni206:
This PR adds new `is*` functions from the cpp API to the python API.
In particular, it adds `getFloatingPointValue()` function from the cpp API.
A test (translated from `term_black.cpp`) is added.

`getFloatingPointValue()` returns a tuple, and so this requires importing an instance of tuples into cython.

### DIFF
--- a/src/api/python/cvc5.pxd
+++ b/src/api/python/cvc5.pxd
@@ -27,6 +27,33 @@ cdef extern from "<string>" namespace "std":
         const wchar_t* data() except +
         size_t size() except +
 
+cdef extern from "<tuple>" namespace "std" nogil:
+    cdef cppclass tuple[T, U, S]:
+        ctypedef T first_type
+        ctypedef U second_type
+        ctypedef S third_type
+        T first
+        U second
+        S third
+        tuple() except +
+        tuple(tuple&) except +
+        tuple(T&, U&, S&) except +
+        bint operator==(tuple&, tuple&)
+        bint operator!=(tuple&, tuple&)
+        bint operator<(tuple&, tuple&)
+        bint operator>(tuple&, tuple&)
+        bint operator<=(tuple&, tuple&)
+        bint operator>=(tuple&, tuple&)
+
+ctypedef fused uint32_t_or_Term:
+    uint32_t
+    Term
+
+cdef extern from "<tuple>" namespace "std":
+    uint32_t get0 "std::get<0>"(tuple[uint32_t,uint32_t,Term]) except +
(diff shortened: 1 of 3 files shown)


src/api/python/cvc5.pxd
```diff
@@ -27,6 +27,29 @@ cdef extern from "<string>" namespace "std":
...
+        S third
+        tuple() except +
+        tuple(tuple&) except +
+        tuple(T&, U&, S&) except +
+        bint operator==(tuple&, tuple&)
+        bint operator!=(tuple&, tuple&)
+        bint operator<(tuple&, tuple&)
+        bint operator>(tuple&, tuple&)
+        bint operator<=(tuple&, tuple&)
+        bint operator>=(tuple&, tuple&)
```
nafur, line 46: I think we should only "import" what we need. I think we don't need any of this (but not sure)
```suggestion
    cdef cppclass tuple[T, U, S]:
        pass
```

src/api/python/cvc5.pxd
```diff
@@ -27,6 +27,29 @@ cdef extern from "<string>" namespace "std":
...
+
+cdef extern from "<tuple>" namespace "std":
+    uint32_t get0 "std::get<0>"(tuple[uint32_t,uint32_t,Term]) except +
+    uint32_t get1 "std::get<1>"(tuple[uint32_t,uint32_t,Term]) except +
+    Term get2 "std::get<2>"(tuple[uint32_t,uint32_t,Term]) except +
```
nafur, line 51: I think we can properly declare them a templated functions (not 100% sure about the syntax though, see https://cython.readthedocs.io/en/latest/src/userguide/wrapping_CPlusPlus.html#templates)
```suggestion
    T get0[T,U.S] "std::get<0>"(tuple[T,U,S]) except +
    U get1[T,U.S] "std::get<1>"(tuple[T,U,S]) except +
    S get2[T,U.S] "std::get<2>"(tuple[T,U,S]) except +
```

nafur (commented):
Two suggestions...

src/api/python/cvc5.pxd
yoni206: Right! Thanks, deleted here: f68979e6b4742f4210c2420af6fff6501061a15d

src/api/python/cvc5.pxd
nafur: Discussed offline. This does not work, and we don't know how to make it work...

nafur (approved):
LGTM.

PR #6650, opened 2021-06-01
Disable timeout regressions

ajreynol:
Disables two regressions that have been timing out causing nightlies to fail.


nafur (approved):
LGTM.

PR #6651, opened 2021-06-01
Use top-level substitutions in ITE simp

ajreynol:
With this PR, we use the central top-level substitutions instance in the ITE simplification preprocessing pass.  Previously the ITE simplification pass maintained its own copy of the substitutions.

Since the top-level substitutions now are the authority on models, this led to model failures after this change: https://github.com/cvc5/cvc5/commit/ac8cf53b07eb29687850f2ae64007f9f2688c9ad#diff-30677c5a1752b1d0e83ef25fd2cfb8949576ea42cf7821fe0ac00ebbd0122f8aL276.
This PR corrects the issue.

This is required for SMT-COMP.


4tXJ7f (approved):
LGTM, thanks for the quick fix!

PR #6652, opened 2021-06-01
Adding getters to the python API and testing them

yoni206:
This PR adds missing API functions from the cpp Term API to the python API.
Corresponding tests are translated from `term_black.cpp`.


src/api/python/cvc5.pxd
```diff
@@ -388,6 +389,8 @@ cdef extern from "api/cpp/cvc5.h" namespace "cvc5::api":
             Term operator*() except +
         const_iterator begin() except +
         const_iterator end() except +
+
+        bint isConstArray() except +
```
nafur, line 393: What about `getConstArrayBase()`?

src/api/python/cvc5.pxi
```diff
@@ -1698,6 +1712,49 @@ cdef class Term:
...
+    def isSetValue(self):
+        return self.cterm.isSetValue()
+
+    def getSetValue(self):
+        elems = []
+        for e in self.cterm.getSetValue():
+            term = Term(self.solver)
+            term.cterm = e
+            elems.append(term)
+        return set(elems)
```
nafur, line 1724: Is there a reason to first convert into a list and then convert to a set?
```suggestion
        elems = set()
        for e in self.cterm.getSetValue():
            term = Term(self.solver)
            term.cterm = e
            elems.add(term)
        return elems
```

src/api/python/cvc5.pxi
```diff
@@ -1698,6 +1712,49 @@ cdef class Term:
...
+
+    def isSequenceValue(self):
+        return self.cterm.isSequenceValue()
+
+    def getSequenceValue(self):
+        elems = []
+        for e in self.cterm.getSequenceValue():
+            term = Term(self.solver)
+            term.cterm = e
+            elems.append(term)
```
nafur, line 1734: I've seen this pattern over and over again. Can we add a constructor to term to allow for `elems.append(Term(self.solver, e))`?
And then have `return [Term(self.solver, e) for e in self.cterm.getSequenceValue()]`?

nafur (changes requested):
Some comments.

src/api/python/cvc5.pxd
yoni206: It was already implemented some place else in the file so wasn't visible in the diff. But now I moved it here.

nafur (approved):
LGTM.

Issue #6625, opened 2021-05-27
Fatal failure at src/smt/smt_engine.cpp:1453 on string formula (unsat cores)


zhendongsu:
Commit: https://github.com/cvc5/cvc5/commit/028d657dc41bbb908b7b9ad6ba707dc2216b15ac
OS: Ubuntu 18.04

```
[514] % cvc4-1.8 -q -i --check-unsat-cores small.smt2
unsat
unsat
[515] % cvc5 -q -i --check-unsat-cores small.smt2
unsat
Fatal failure within void cvc5::SmtEngine::checkUnsatCore() at /local/suz-local/software/CVC4/src/smt/smt_engine.cpp:1453
Internal error detectedSmtEngine::checkUnsatCore(): produced core was satisfiable.
Aborted
[516] % 
[516] % cat small.smt2
(declare-const a String)
(declare-const b String)
(declare-const c String)
(push)
(assert (= (str.++ a b) c))
(assert (= a "A"))
(assert (= c ""))
(check-sat)
(pop)
(assert (= (str.++ a b) (str.++ "A" c)))
(assert (= c (str.++ a b)))
(check-sat)
[517] % 
```

ajreynol:
Fixed by https://github.com/cvc5/cvc5/pull/6655.

PR #6655, opened 2021-06-01
Fix unsat core proofs

ajreynol:
Fixes cases of satisfiable unsat cores when proofs are enabled.

Unfortunately, this bug was also preventing us from doing the final scope check for all proof checking.  As this was not being tested, this PR uncovers that proof checking is now failing on 6 regressions.  I'm disabling proof checking here and will address these issues on later PRs.


mpreiner (approved):
LGTM, does this fix #6625 ?

4tXJ7f:
I briefly tested it and it looks like it does fix #6625.

HanielB:
Why ensuring that the proof is closed fixes the problem with that regression?

ajreynol:
The "ensureClosed" argument makes sure that the proof is processed in a way that connects its free assumptions to the arguments of the final scope. If this is not done, then (= x y) could be a free assumption where (= y x) is an input assumption, and hence (= y x) would not be included in the unsat core.

PR #6656, opened 2021-06-01
Move `toPythonObj` tests to the new API unit test directory

yoni206:
This is the last test file that we move from the old directory to the new one, and so the old directory is deleted.


makaimann (approved):
LGTM, thanks!

PR #6658, opened 2021-06-01
Adding unit tests for the datatypes python API

yoni206:
This PR adds unit tests that are translated from `datatype_api_black.cpp`. One API function is also added to the python API.
This is the last part of the python api unit tests for datatypes.

Co-authored-by: Andres Noetzli <andres.noetzli@gmail.com>


test/python/unit/api/test_datatype_api.py
```diff
@@ -294,3 +294,249 @@ def test_parametric_datatype(solver):
...
+    dtdecls.clear()
+    dtdecls.append(ns2)
+
+    dtsorts.clear()
+    # this is not well-founded due to non-simple recursion
+    dtsorts = solver.mkDatatypeSorts(dtdecls, unresTypes)
+    assert len(dtsorts) == 1
+    assert dtsorts[0].getDatatype()[0][0].getRangeSort().isArray()
+    assert dtsorts[0].getDatatype()[0][0].getRangeSort().getArrayElementSort(
+    ) == dtsorts[0]
```
Ying1123, line 379: Does this line been formatted?

test/python/unit/api/test_datatype_api.py
```diff
@@ -294,3 +294,249 @@ def test_parametric_datatype(solver):
...
+    elemArray = solver.mkDatatypeConstructorDecl("elemArray")
+    elemArray.addSelector("ndata", solver.mkArraySort(unresList, unresList))
+    ns.addConstructor(elemArray)
+
+    dtdecls = []
+    dtdecls.append(wlist)
+    dtdecls.append(llist)
+    dtdecls.append(ns)
+    # this is well-founded and has no nested recursion
+    dtsorts = []
```
4tXJ7f, line 343:
```suggestion
```

test/python/unit/api/test_datatype_api.py
```diff
@@ -294,3 +294,249 @@ def test_parametric_datatype(solver):
...
+    elem.addSelector("ndata", solver.mkSetSort(unresWList))
+    ns.addConstructor(elem)
+    elemArray = solver.mkDatatypeConstructorDecl("elemArray")
+    elemArray.addSelector("ndata", solver.mkArraySort(unresList, unresList))
+    ns.addConstructor(elemArray)
+
+    dtdecls = []
+    dtdecls.append(wlist)
+    dtdecls.append(llist)
+    dtdecls.append(ns)
```
4tXJ7f, line 341:
```suggestion
    dtdecls = [wlist, llist, ns]
```

(yeah, the corresponding C++ test pushes them separately but I don't think that it adds much here)

test/python/unit/api/test_datatype_api.py
```diff
@@ -294,3 +294,249 @@ def test_parametric_datatype(solver):
...
+    elem2.addSelector("ndata",
+                      solver.mkArraySort(solver.getIntegerSort(), unresNs2))
+    ns2.addConstructor(elem2)
+    nil2 = solver.mkDatatypeConstructorDecl("nil2")
+    ns2.addConstructor(nil2)
+
+    dtdecls.clear()
+    dtdecls.append(ns2)
+
+    dtsorts.clear()
```
4tXJ7f, line 373:
```suggestion
```

(the list gets replaced in the next line anyway)

test/python/unit/api/test_datatype_api.py
```diff
@@ -294,3 +294,249 @@ def test_parametric_datatype(solver):
...
+    ns3 = solver.mkDatatypeDecl("ns3")
+    elem3 = solver.mkDatatypeConstructorDecl("elem3")
+    elem3.addSelector("ndata", solver.mkSetSort(unresList3))
+    ns3.addConstructor(elem3)
+
+    dtdecls.clear()
+    dtdecls.append(list3)
+    dtdecls.append(ns3)
+
+    dtsorts.clear()
```
4tXJ7f, line 412:
```suggestion
```

test/python/unit/api/test_datatype_api.py
```diff
@@ -294,3 +294,249 @@ def test_parametric_datatype(solver):
...
+    ns4 = solver.mkDatatypeDecl("ns4")
+    elem4 = solver.mkDatatypeConstructorDecl("elem3")
+    elem4.addSelector("ndata", unresList4)
+    ns4.addConstructor(elem4)
+
+    dtdecls.clear()
+    dtdecls.append(list4)
+    dtdecls.append(ns4)
+
+    dtsorts.clear()
```
4tXJ7f, line 449:
```suggestion
```

test/python/unit/api/test_datatype_api.py
```diff
@@ -294,3 +294,249 @@ def test_parametric_datatype(solver):
...
+    unresList5 = solver.mkSortConstructorSort("list5", 1)
+    unresTypes.add(unresList5)
+
+    v = []
+    x = solver.mkParamSort("X")
+    v.append(x)
+    list5 = solver.mkDatatypeDecl("list5", v)
+
+    args = []
+    args.append(x)
```
4tXJ7f, line 472:
```suggestion
    args = [x]
```

test/python/unit/api/test_datatype_api.py
```diff
@@ -294,3 +294,249 @@ def test_parametric_datatype(solver):
...
+    unresList = solver.mkSortConstructorSort("plist", 1)
+    unresTypes.add(unresList)
+
+    v = []
+    x = solver.mkParamSort("X")
+    v.append(x)
+    plist = solver.mkDatatypeDecl("plist", v)
+
+    args = []
+    args.append(x)
```
4tXJ7f, line 511:
```suggestion
    args = [x]
```

test/python/unit/api/test_datatype_api.py
```diff
@@ -294,3 +294,249 @@ def test_parametric_datatype(solver):
...
+
+    pcons = solver.mkDatatypeConstructorDecl("pcons")
+    pcons.addSelector("car", x)
+    pcons.addSelector("cdr", urListX)
+    plist.addConstructor(pcons)
+    nil5 = solver.mkDatatypeConstructorDecl("pnil")
+    plist.addConstructor(nil5)
+
+    dtdecls = []
+    dtdecls.append(plist)
```
4tXJ7f, line 522:
```suggestion
    dtdecls = [plist]
```

test/python/unit/api/test_datatype_api.py
```diff
@@ -294,3 +294,249 @@ def test_parametric_datatype(solver):
...
+
+    dtsorts = []
```
4tXJ7f, line 524:
```suggestion
```

test/python/unit/api/test_datatype_api.py
```diff
@@ -294,3 +294,249 @@ def test_parametric_datatype(solver):
...
+    # make the datatype sorts
+    dtsorts = solver.mkDatatypeSorts(dtdecls, unresTypes)
+    assert len(dtsorts) == 1
+    d = dtsorts[0].getDatatype()
+    nilc = d[0]
+
+    isort = solver.getIntegerSort()
+    iargs = []
+    iargs.append(isort)
```
4tXJ7f, line 533:
```suggestion
    iargs = [isort]
```

4tXJ7f (changes requested):
Thank you for doing this. I have some minor suggestions about simplifying the use of lists. The general idea behind my suggestions:

- If we assign a list and immediately replace it in the next line, then we can remove the former.
- If we initialize a list and immediately append some elements, then we can combine those lines.
- If we clear a list and append elements, I left it as-is because there might be some subtle differences between clearing and appending to an existing list vs. replacing the list with a new one, so it may be a good idea to leave that in there (e.g., to check that the GC is working as expected).

Overall, the suggestions make the tests a bit less faithful to the C++ implementation but I don't think that keeping it 100% the same adds much to the tests.

test/python/unit/api/test_datatype_api.py
yoni206: Strangely, this is what `yapf` did to split it when is was in one line. I pushed a manual fix.

test/python/unit/api/test_datatype_api.py
yoni206: Of course... This is how I translated an assignmentless declaration (e.g., `std::vector<Sort> dtsorts;`) :-)

yoni206:
Thanks for the review and suggestions! Makes a lot of sense, and was overlooked while doing an aggressive semi-automatic translation :)

4tXJ7f (approved):
LGTM, thanks for the updates :)

PR #6663, opened 2021-06-02
Make `STRINGS_CTN_DECOMPOSE` an explicit conflict 

### LINKED ISSUE #6643: Assertion error at src/theory/strings/extf_solver.cpp:551
cvc5 6edc06b3fb6367e8366cab1
~~~~
(declare-fun y () String)
(declare-fun z () String)
(assert (not (= (str.contains y (str.replace "A" "" z)) (str.contains y "A"))))
(check-sat)
~~~~
~~~~
cvc5 -q  --strings-exp delta.out.smt2
Fatal failure within void cvc5::theory::strings::ExtfSolver::checkExtfInference(cvc5::Node, cvc5::Node, cvc5::theory::strings::ExtfInfoTmp&, int) at /home/cvc5/src/theory/strings/extf_solver.cpp:551
Check failure

 d_state.isInConflict()
~~~~

4tXJ7f:
Fixes #6643. The `STRINGS_CTN_DECOMPOSE` inference is always a conflict
but we sometimes sent it as an inference. To make sure that the
inference manager actually recognizes the inference as a conflict, this
commit ensures that the conclusion is always `false` and modifies the
explanation accordingly.


ajreynol (approved):
LGTM, good catch.

PR #6666, opened 2021-06-02
Update CVC4 URLs/macros

4tXJ7f:


src/api/cpp/cvc5_kind.h
```diff
@@ -1868,7 +1868,7 @@ enum CVC5_EXPORT Kind : int32_t
    *
    * Note: We currently support the creation of constant arrays, but under some
    * conditions when there is a chain of equalities connecting two constant
-   * arrays, the solver doesn't know what to do and aborts (Issue <a href="https://github.com/CVC4/CVC4/issues/1667">#1667</a>).
+   * arrays, the solver doesn't know what to do and aborts (Issue <a href="https://github.com/cvc5/cvc5/issues/1667">#1667</a>).
```
nafur, line 1871: I'd suggest to avoid HTML and use Sphinx here:
```suggestion
   * arrays, the solver doesn't know what to do and aborts (Issue `#1667 <https://github.com/cvc5/cvc5/issues/1667>`_).
```
(Please check that is is really correct...)

src/expr/dtype.cpp
```diff
@@ -260,7 +260,7 @@ void DType::setSygus(TypeNode st, Node bvl, bool allowConst, bool allowAll)
   // Notice we only want to do this for sygus datatypes that are user-provided.
   // At the moment, the condition !allow_all implies the grammar is
   // user-provided and hence may require a default constant.
-  // TODO (https://github.com/CVC4/cvc4-projects/issues/38):
+  // TODO (https://github.com/cvc5/cvc5-projects/issues/38):
```
nafur, line 263: This issue has been closed...

nafur (commented):
Two comments...

src/expr/dtype.cpp
4tXJ7f: Yeah, I noticed that but from a quick check, it looked like the comment is still relevant. I've already pinged @ajreynol about it. We'll see what he says :)

src/api/cpp/cvc5_kind.h
4tXJ7f: I compiled the docs and it seems to have worked.

src/expr/dtype.cpp
4tXJ7f: I have discussed it with Andy and removed the reference to the issue. The comment about the alternative design is not important enough to warrant an issue.

src/api/cpp/cvc5_kind.h
```diff
@@ -1868,7 +1868,7 @@ enum CVC5_EXPORT Kind : int32_t
    *
    * Note: We currently support the creation of constant arrays, but under some
    * conditions when there is a chain of equalities connecting two constant
-   * arrays, the solver doesn't know what to do and aborts (Issue <a href="https://github.com/CVC4/CVC4/issues/1667">#1667</a>).
+   * arrays, the solver doesn't know what to do and aborts (Issue `#1667 <https://github.com/cvc5/cvc5/issues/1667>`_).
```
aniemetz, line 1871: Break for 80 chars.

4tXJ7f:
For some reason, I couldn't get the RST-style link to work anymore. Either I mistakenly thought that it works and it actually didn't or some update broke it. Anyway, I've changed the comment to be a verbatim RST block now and changed the note to look like an actual note.

aniemetz:
What was wrong with using the html link? Sphinx rst syntax doesn't work here because this first goes through Doxygen (so you have to use Doxygen syntax to generate links), except how you did it just now. I'd prefer to not mix syntax here if possible.

4tXJ7f:
I ended up sticking with RST because it allows us to make a nice note box:

image

aniemetz (changes requested):
This note box you will also get when you use Doxygen's `@note`.
I still am against mixing syntax.

4tXJ7f:
Ah neat, TIL. I have changed it.

aniemetz (approved):

PR #6668, opened 2021-06-02
Fix issues when poly is disabled

nafur:
Recent changes introduced issues when libpoly is disabled.


4tXJ7f (approved):
LGTM. I was actually having issues with this :)

PR #6674, opened 2021-06-03
Fix handling of start index in `str.indexof_re`

### LINKED ISSUE #6636: (default) solution unsoundness on string formula with str.replace_re_all (fatal failure by cvc4-1.8)
Commit: https://github.com/cvc5/cvc5/commit/f62b46414cc47762857a4e3241318733ca8c973d
OS: Ubuntu 18.04
Note: (1) the formula should be unsat; (2) the bug doesn't reproduce if replacing `str.replace_re_all` with `str.replace_re` 

```
[549] % cvc5 -q --strings-exp --check-models small.smt2
Fatal failure within void cvc5::smt::CheckModels::checkModel(cvc5::smt::Model*, cvc5::context::CDList<cvc5::NodeTemplate<true> >*, bool) at /local/suz-local/software/CVC4/src/smt/check_models.cpp:125
assertion:     (let ((_let_1 (re.* re.allchar ))) (not (= a (str.replace_re_all a (re.++ _let_1 (str.to_re a) _let_1) a))))
simplifies to: false
expected `true'.
Aborted
[550] % 
[550] % cat small.smt2
(declare-fun a () String)
(assert (not (= a (str.replace_re_all a (re.++ (re.* re.allchar) (str.to_re a) (re.* re.allchar)) a))))
(check-sat)
[551] % 
```

### LINKED ISSUE #6637: (default) solution unsoundness on string formula with str.replace_re_all (regression from cvc4-1.8)
Commit: https://github.com/cvc5/cvc5/commit/0133367f9ed242aa01e42867364c7be74ffe5618
OS: Ubuntu 18.04
Note: (1) might be related to https://github.com/cvc5/cvc5/issues/6636, but cvc4-1.8 produces a fatal failure on the other instance for https://github.com/cvc5/cvc5/issues/6636; (2) model validation detects the produced invalid model 

```
[519] % cvc4-1.8 -q --strings-exp small.smt2
unsat
[520] % cvc5 -q --strings-exp small.smt2
sat
[521] % 
[521] % cat small.smt2
(declare-fun a () String)
(assert (= (str.len a) 2))
(assert (= (str.len (str.replace_re_all a (str.to_re "A") "B")) 3))
(check-sat)
[522] % 
```

4tXJ7f:
Fixes #6636, fixes #6637. When the start index was non-zero, the result of
`str.indexof_re` was not properly restricted to be greater or equal to
the start index. This commit fixes the issue by making the eager
reduction lemma more precise. Additionally, the commit fixes an issue
with the lower bound of the length of the match in `str.indexof_re`.


src/theory/strings/term_registry.cpp
```diff
@@ -94,9 +94,11 @@ Node TermRegistry::eagerReduce(Node t, SkolemCache* sc)
     //
     // where f in { str.indexof, str.indexof_re }
     Node l = nm->mkNode(STRING_LENGTH, t[0]);
-    lemma = nm->mkNode(AND,
-                       nm->mkNode(GEQ, t, nm->mkConst(Rational(-1))),
-                       nm->mkNode(LEQ, t, l));
+    lemma = nm->mkNode(
```
ajreynol, line 97: Can you update the comment above?

4tXJ7f: Ah yeah, sorry. Done.

ajreynol (approved):
LGTM

PR #6675, opened 2021-06-03
Renaming `pow2` to `p2` in regression tests

yoni206:
We plan to add a unary `pow2` operator to `cvc5`, that is obtained from the binary operator `pow` by fixing the first argument to `2`. 
An initial working branch is here: https://github.com/yoni206/cvc5/tree/pow2

This PR does the first step, which is to rename some uninterpreted symbols in regression tests from `pow2` to `p2`, to avoid clashing with the new operator.


ajreynol:
This is fine, although notice that when `pow2` is added as an operator, it should be guarded to be allowed only in non-strict parsing mode, see e.g. https://github.com/cvc5/cvc5/blob/master/src/parser/smt2/smt2.cpp#L136

ajreynol (approved):

PR #6678, opened 2021-06-03
Simplify automatic set-logic in smt2 parser

ajreynol:
This simplifies the smt2 parser so that automatically setting the logic is done directly instead of being buffered as a command.

This prevents spurious errors for features that require (A) checking the logic is set and (B) fully intialize the underlying SMT engine.  `declare-pool` is an example of an smt2 command where the user will get an error (instead of a warning) when set-logic is not used due to setting the logic, after fully initing SMT engine and then executing the buffered set-logic command.

Note this should also make dump=raw-benchmark more accurate (no set-logic is included when dumping benchmarks with no set-logic command).


4tXJ7f (approved):
LGTM

PR #6683, opened 2021-06-04
Properly consider aliases in option handlers

nafur:
This PR makes sure that option handlers have access to both the canonical option name and the option name that was actually used. It also updates the options README and gets rid of the `base_handlers.h` of which only a fraction was used.


src/options/README.md
```diff
@@ -0,0 +1,158 @@
...
+* `help_mode` (string): documentation for the mode enum (required if `mode` is given)
+
+
+Handler functions
+-----------------
+
+Custom handler functions are used to turn the option value from a `std::string` into the type specified by `type`.
+Standard handler functions are provided for basic types (`std::string`, `bool`, integer types and floating point types) as well as enums specified by `mode`.
+A custom handler function needs to be member function of `options::OptionsHandler` with signature `{type} {handler}(const std::string& option, const std::string& flag, const std::string& optionvalue)`, or alternatively `void {handler}(const std::string& option, const std::string& flag)` if the `type` is `void`.
+The two parameters `option` and `flag` hold the canonical and the actually used option names, respectively, and they may differ if an alternative name (from `alias`) was used. While `option` should be used to identify an option (e.g. by comparing against `*__name`), `flag` should be usually be used in user messages.
```
mpreiner, line 54:
```suggestion
The two parameters `option` and `flag` hold the canonical and the actually used option names, respectively, and they may differ if an alternative name (from `alias`) was used. While `option` should be used to identify an option (e.g. by comparing against `*__name`), `flag` should be usually used in user messages.
```

src/options/README.md
```diff
@@ -0,0 +1,158 @@
...
+
+Predicate functions are used to check whether an option value is valid after it has been parsed by a (standard or custom) handler function.
+Like a handler function, a predicate function needs to be a member function of `options::OptionsHandler` with signature `void {predicate}(const std::string& option, const std::string& flag, {type} value)`. If the check fails, the predicate should raise an `OptionException`.
+
+
+Mode options
+------------
+
+An option can be defined to hold one of a given set of values we call modes.
+Doing so automatically defines an `enum class` for the set of modes and makes the option accept on of the values from the enum.
```
mpreiner, line 68:
```suggestion
Doing so automatically defines an `enum class` for the set of modes and makes the option accept one of the values from the enum.
```

src/options/README.md
```diff
@@ -0,0 +1,158 @@
+Specifying Modules
```
mpreiner, line 1: Please format the file to 80 chars.

src/options/mkoptions.py
```diff
@@ -210,6 +210,34 @@ def get_holder_ref_decls(modules):
...
+        return []
+    optname = option.long_name if option.long else ""
+    if option.type == 'bool':
+        return ['opts.handler().{}("{}", option, value);'.format(x, optname) \
+                for x in option.predicates]
+    else:
+        assert option.type != 'void'
+        return ['opts.handler().{}("{}", option, value);'.format(x, optname) \
+                for x in option.predicates]
+
```
mpreiner, line 240: What's the difference of these two returns?

src/options/options_handler.cpp
```diff
@@ -283,18 +283,18 @@ void OptionsHandler::setStats(const std::string& option, bool value)
   }
 }
 
-void OptionsHandler::threadN(std::string option) {
-  throw OptionException(option + " is not a real option by itself.  Use e.g. --thread0=\"--random-seed=10 --random-freq=0.02\" --thread1=\"--random-seed=20 --random-freq=0.05\"");
+void OptionsHandler::threadN(const std::string& option, const std::string& flag) {
```
mpreiner, line 286: Doesn't look like the changes are clang-formatted.

src/options/options_handler.h
```diff
@@ -44,72 +44,73 @@ class OptionsHandler {
...
    */
-  void setProduceAssertions(std::string option, bool value);
+  void setProduceAssertions(const std::string& option, const std::string& flag, bool value);
 
-  void setStats(const std::string& option, bool value);
+  void setStats(const std::string& option, const std::string& flag, bool value);
 
-  unsigned long limitHandler(std::string option, std::string optarg);
-  void setResourceWeight(std::string option, std::string optarg);
+  unsigned long limitHandler(const std::string& option, const std::string& flag, std::string optarg);
```
mpreiner, line 92: Ditto (clang-format).

Can you also change the type to uint64_t since you are already touching the line?

mpreiner (changes requested):

src/options/mkoptions.py
nafur: I guess they used to be different at some point in the past. I'll merge them.

mpreiner (approved):

PR #6685, opened 2021-06-04
(proof-new) Fix missing connection in trust substitution proofs

ajreynol:
This PR fixes a missing connection in trust substitution proofs, which was the cause of open proofs when solved equalities from ppAssert were not justified by proofs.

Also distinguishes TRUST_SUBS_EQ from TRUST_SUBS_MAP for clarity.


src/proof/proof_rule.h
```diff
@@ -219,7 +219,7 @@ enum class PfRule : uint32_t
   THEORY_REWRITE,
   // The remaining rules in this section have the signature of a "trusted rule":
   //
-  // Children: none
+  // Children: ?
```
HanielB, line 222: clarify this. I think it'd be good to say that unless otherwise specified in the description below, the rule takes no children.

HanielB (approved):

PR #6687, opened 2021-06-04
Miscellaneous changes from central ee branch

ajreynol:
Minor reorganization to make calls to theory engine from combination engine / shared solver more organized.

### DIFF
--- a/src/theory/shared_solver.cpp
+++ b/src/theory/shared_solver.cpp
@@ -34,7 +34,8 @@ SharedSolver::SharedSolver(TheoryEngine& te, ProofNodeManager* pnm)
       d_logicInfo(te.getLogicInfo()),
       d_sharedTerms(&d_te, d_te.getSatContext(), d_te.getUserContext(), pnm),
       d_preRegistrationVisitor(&te, d_te.getSatContext()),
-      d_sharedTermsVisitor(&te, d_sharedTerms, d_te.getSatContext())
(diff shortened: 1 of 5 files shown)


src/theory/shared_solver.cpp
```diff
@@ -104,9 +105,13 @@ EqualityStatus SharedSolver::getEqualityStatus(TNode a, TNode b)
...
 
-void SharedSolver::sendLemma(TrustNode trn, TheoryId atomsTo)
+bool SharedSolver::propagateLit(TNode predicate, bool value)
 {
-  d_te.lemma(trn, LemmaProperty::NONE, atomsTo);
+  if (value)
+  {
+    return d_out.propagate(predicate);
+  }
+  return d_out.propagate(predicate.notNode());
```
4tXJ7f, line 114:
```suggestion
  return d_out.propagate(value ? predicate : predicate.notNode());
```

4tXJ7f (approved):
LGTM

src/theory/shared_solver.cpp
ajreynol: This is actually not possible due to Node vs TNode issues; notNode creates a Node, thus this saves ref counting as I understand it.  This pattern is used throughout the code in other places for propagations, e.g. https://github.com/cvc5/cvc5/blob/master/src/theory/theory_eq_notify.h#L41.

PR #6689, opened 2021-06-05
Remove unwanted side effects in `SPLIT_EQ_STRIP_L`

### LINKED ISSUE #6681: (ext-rew-prep) refutation unsoundness on string formula
Commit: https://github.com/cvc5/cvc5/commit/adf497af7a3fc8b06b875eaf9feca2568d0ba9d8
OS: Ubuntu 18.04
Note: (1) related to https://github.com/cvc5/cvc5/issues/6520; (2) also affects cvc4-1.8, but not cvc4-1.7 

```
[607] % cvc5 -q --dump-models small.smt2
sat
(
(define-fun a () String "")
(define-fun b () String "AC")
(define-fun c () String "A")
(define-fun d () String "A")
)
[608] % cvc5 -q --ext-rew-prep small.smt2
unsat
[609] % cat small.smt2
(declare-fun a () String)
(declare-fun b () String)
(declare-fun c () String)
(declare-fun d () String)
(assert (= (str.++ "A" a "CBA" b "BA" d) (str.++ b "BA" d a "CBA" c)))
(check-sat)
[610] % 
```

4tXJ7f:
Fixes #6681. When checking whether `SPLIT_EQ_STRIP_L` applies, we were
using `sripSymbolicLength`, which changes its inputs depending on how
many elements of the concatenation it could strip. However, one of the
arguments, `pfxv0`, was reused across multiple checks if the strip did
not result in a rewrite. Later invocations were wrong as a result. This
commit changes the call to `stripSymbolicLength()` to use a new copy of
the vector instead.

### DIFF
--- a/src/theory/strings/sequences_rewriter.cpp
+++ b/src/theory/strings/sequences_rewriter.cpp
@@ -512,7 +512,7 @@ Node SequencesRewriter::rewriteStrEqualityExt(Node node)
               // right-hand prefix).
               if (lenPfx1.isConst() && lenPfx1.getConst<Rational>().isZero())
               {
-                sfxv0.insert(pfxv0.end(), v0.begin() + i, v0.end());
+                sfxv0.insert(sfxv0.end(), v0.begin() + i, v0.end());
                 std::vector<Node> sfxv1(v1.begin() + j, v1.end());
                 Node ret =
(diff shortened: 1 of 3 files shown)


ajreynol (approved):
LGTM

PR #6691, opened 2021-06-05
Support public option modules

nafur:
This PR adds the possibility to make option modules public. As shown on the example of the main driver options, this allows to get rid of the wrappers from `options_public.h`. We plan to make only very few option modules public (i.e. main and parser).

### DIFF
--- a/src/main/driver_unified.cpp
+++ b/src/main/driver_unified.cpp
@@ -35,6 +35,7 @@
 #include "main/time_limit.h"
 #include "options/options.h"
 #include "options/options_public.h"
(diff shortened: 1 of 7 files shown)


ajreynol (approved):
LGTM.

It looks like options_public.h will eventually be deleted? I'm very much in favor of this.

nafur:
> LGTM.
> 
> It looks like options_public.h will eventually be deleted? I'm very much in favor of this.

Yes, this is the plan. It requires moving several options to other modules to avoid having many modules made public, which in turn requires further changes...

PR #6692, opened 2021-06-06
Change output of getRealValue to a fraction.

alex-ozdemir:
Our documentation for `toPythonObj` says that real values are represented as Fractions. However, getRealValue yields a float approximation thereof.

We should probably stick to Fractions, since they allow us to precisely capture values in LRA. Also, that's more aligned with the C++ API, which returns a string representation of the (unapproximated) Rational.

Also, document some (potential) weirdness with calling mkReal on floating-point values.

CC'ing Gereon **and** Yoni for review, since they touched this code last.


src/api/python/cvc5.pxi
```diff
@@ -740,6 +740,21 @@ cdef class Solver:
...
 
     def mkReal(self, val, den=None):
+        '''
+        Make a real number term.
+
+        Really, makes a rational term.
+
+        Can be used in various forms.
+        * Given a string "N/D" constructs the corresponding rational.
+        * Given a string "W.D" constructs the reduction of (N * P + D)/P, where
```
nafur, line 750:
```suggestion
        * Given a string "W.D" constructs the reduction of (W * P + D)/P, where
```

nafur (approved):
One comment, otherwise LGTM.

PR #6693, opened 2021-06-07
Remove `binary_name` option

nafur:
The `binary_name` is solely used as a temporary storage to pass the data from the options parser back to the `runCvc5` method where it is put in a static variable. This PR gets rid of the option and the public option getter in favor of directly storing the program name in the static variable using an additional argument to `parseOptions()`.


4tXJ7f (approved):
LGTM

PR #6695, opened 2021-06-07
Make TheoryUF compatible with central equality engine

ajreynol:
Work towards central equality engine.

This does some minor reorganization to TheoryUF, related to UF+cardinality constraints that makes it compatible with central equality engine. In particular, preNotifyFact is removed in favor of adding conflicts after cardinality constraints are added to the equality engine. This ensures that the central equality engine can explain conflicts that involve cardinality constraints (which will no longer be the responsibility of UF when --ee-mode=central).


4tXJ7f (approved):
LGTM

Issue #6653, opened 2021-06-01
(re-elim, re-elim-agg) invalid model for string formula with regexes 


zhendongsu:
Commit: https://github.com/cvc5/cvc5/commit/172573dba45f7d231ec06a3a3992f41cf794b75e
OS: Ubuntu 18.04

```
[537] % cvc5 -q --strings-exp --check-models small.smt2
sat
[538] % cvc5 -q --strings-exp --check-models --re-elim --re-elim-agg small.smt2
Fatal failure within void cvc5::smt::CheckModels::checkModel(cvc5::smt::Model*, cvc5::context::CDList<cvc5::NodeTemplate<true> >*, bool) at /local/suz-local/software/CVC4/src/smt/check_models.cpp:125
Internal error detectedSmtEngine::checkModel(): ERRORS SATISFYING ASSERTIONS WITH MODEL:
assertion:     (let ((_let_1 (str.++ b "a"))) (str.in_re (str.replace_re b (str.to_re (str.replace_all a _let_1 b)) _let_1) (re.+ (str.to_re "b"))))
simplifies to: false
expected `true'.
Run with `--check-models -v' for additional diagnostics.
Aborted
[539] % 
[539] % cat small.smt2
(declare-fun a () String)
(declare-fun b () String)
(assert (str.in_re (str.replace_re b (str.to_re (str.replace_all a (str.++ b "a") b)) (str.++ b "a")) (re.+ (str.to_re "b"))))
(assert (str.in_re a (re.* (re.range "a" "b"))))
(check-sat)
[540] % 
```

rainoftime:
Another invalid model issue at f2bbc8c1f6d8f3576937
~~~~
(set-option :strings-lazy-pp false)
(set-option :check-unsat-cores true)
(declare-fun s () String)
(assert (not (= "A" (str.substr (str.update "AAAAAA" 1 s) 5 1))))
(check-sat)
~~~~

~~~~
cvc5 -q -i --check-models xx.smt2
Fatal failure within void cvc5::smt::CheckModels::checkModel(cvc5::smt::Model *, context::CDList<Node> *, bool) at //cvc5/src/smt/check_models.cpp:125
Internal error detectedSmtEngine::checkModel(): ERRORS SATISFYING ASSERTIONS WITH MODEL:
assertion:     (not (= "A" (str.substr (str.update "AAAAAA" 1 s) 5 1)))
simplifies to: false
expected `true'.
[...]

PR #6696, opened 2021-06-07
Fix str.update reduction

ajreynol:
Fixes the 2nd benchmark from #6653.


4tXJ7f (approved):
LGTM. It's neat to encode the condition like that without branching.

PR #6697, opened 2021-06-07
Better support for HOL parsing and set up

HanielB:
This commit adds a new parser option, `--hol`, which marks that HOL is being used. This option has the effect of prepending the problem's logic with "HO_", which teels the solver that the logic is higher order. The parser builder, base parser, and SMT2 and TPTP parsers are all updated to work with this new setting, as is the logic info class.

For now this parser option is enabling the `--uf-ho` option for internal use, since for now higher-order solving relies it. In a future PR this dependency will be removed (since this information is already given to the SMT solver via the logic).

### DIFF
--- a/src/parser/parser.h
+++ b/src/parser/parser.h
@@ -149,9 +149,6 @@ private:
   */
  bool d_canIncludeFile;
 
- /** Whether the logic is higher order. */
- bool d_hol;
-
  /**
   * Whether the logic has been forced with --force-logic.
   */
@@ -194,6 +191,9 @@ private:
 protected:
  /** The API Solver object. */
  api::Solver* d_solver;
+
+ /** Whether the logic is higher order. */
+ Bool d_hol;
 
  /**
   * Create a parser state.
@@ -268,7 +268,7 @@ public:
   bool canIncludeFile() const { return d_canIncludeFile; }
 
   /** Set parser to accept higher-order logic. */
-  void setHOL();
+  virtual void setHOL();
 
   /** Expose the functionality from SMT/SMT2 parsers, while making
       implementation optional by returning false by default. */
--- a/src/parser/parser.h
+++ b/src/parser/parser.h
@@ -149,6 +149,9 @@ private:
   */
  bool d_canIncludeFile;
 
+ /** Whether the logic is higher order. */
+ bool d_hol;
+
  /**
   * Whether the logic has been forced with --force-logic.
   */
@@ -264,11 +267,17 @@ public:
   void disallowIncludeFile() { d_canIncludeFile = false; }
   bool canIncludeFile() const { return d_canIncludeFile; }
 
+  /** Set parser to accept higher-order logic. */
+  void setHOL();
+
   /** Expose the functionality from SMT/SMT2 parsers, while making
       implementation optional by returning false by default. */
   virtual bool logicIsSet() { return false; }
 
   virtual void forceLogic(const std::string& logic);
+
+  /** Whether higher-order logic. */
+  bool isHOL() const;
 
   const std::string& getForcedLogic() const { return d_forcedLogic; }
   bool logicIsForced() const { return d_logicIsForced; }
@@ -305,7 +314,7 @@ public:
   /**
    * Returns the expression that name should be interpreted as, based on the current binding.
    *
-   * This is the same as above but where the name has been type cast to t. 
+   * This is the same as above but where the name has been type cast to t.
    */
   virtual api::Term getExpressionForNameAndType(const std::string& name,
                                                 api::Sort t);
@@ -331,9 +340,9 @@ public:
    * This is a generalization of ExprManager::operatorToKind that also
    * handles variables whose types are "function-like", i.e. where
    * checkFunctionLike(fun) returns true.
-   * 
+   *
    * For examples of the latter, this function returns
-   *   APPLY_UF if fun has function type, 
+   *   APPLY_UF if fun has function type,
    *   APPLY_CONSTRUCTOR if fun has constructor type.
    */
   api::Kind getKindForFunction(api::Term fun);
@@ -379,7 +388,7 @@ public:
 
   /**
    * Checks whether the given expression is function-like, i.e.
-   * it expects arguments. This is checked by looking at the type 
+   * it expects arguments. This is checked by looking at the type
    * of fun. Examples of function types are function, constructor,
    * selector, tester.
    * @param fun the expression to check
@@ -443,7 +452,7 @@ public:
   std::vector<api::Term> bindBoundVars(const std::vector<std::string> names,
                                        const api::Sort& type);
 
-  /** Create a new variable definition (e.g., from a let binding). 
+  /** Create a new variable definition (e.g., from a let binding).
    * levelZero is set if the binding must be done at level 0.
    * If a symbol with name already exists,
    *  then if doOverload is true, we create overloaded operators.
@@ -648,7 +657,7 @@ public:
   /** Is the symbol bound to a boolean variable? */
   bool isBoolean(const std::string& name);
 
-  /** Is fun a function (or function-like thing)? 
+  /** Is fun a function (or function-like thing)?
   * Currently this means its type is either a function, constructor, tester, or selector.
   */
   bool isFunctionLike(api::Term fun);
--- a/src/parser/parser.h
+++ b/src/parser/parser.h
@@ -312,7 +312,8 @@ public:
   virtual api::Term getExpressionForName(const std::string& name);
 
   /**
-   * Returns the expression that name should be interpreted as, based on the current binding.
+   * Returns the expression that name should be interpreted as, based on the
+   * current binding.
    *
    * This is the same as above but where the name has been type cast to t.
    */
@@ -456,7 +457,8 @@ public:
    * levelZero is set if the binding must be done at level 0.
    * If a symbol with name already exists,
    *  then if doOverload is true, we create overloaded operators.
-   *  else if doOverload is false, the existing expression is shadowed by the new expression.
+   *  else if doOverload is false, the existing expression is shadowed by the
+   * new expression.
    */
   void defineVar(const std::string& name,
                  const api::Term& val,
@@ -658,8 +660,9 @@ public:
   bool isBoolean(const std::string& name);
 
   /** Is fun a function (or function-like thing)?
-  * Currently this means its type is either a function, constructor, tester, or selector.
-  */
+   * Currently this means its type is either a function, constructor, tester, or
+   * selector.
+   */
   bool isFunctionLike(api::Term fun);
 
   /** Is the symbol bound to a predicate? */
--- a/src/parser/parser_builder.cpp
+++ b/src/parser/parser_builder.cpp
@@ -56,6 +56,7 @@ void ParserBuilder::init(api::Solver* solver, SymbolManager* sm)
   d_strictMode = false;
   d_canIncludeFile = true;
   d_parseOnly = false;
+  d_hol = false;
   d_logicIsForced = false;
   d_forcedLogic = "";
 }
@@ -95,6 +96,11 @@ Parser* ParserBuilder::build()
     parser->disallowIncludeFile();
   }
 
+  if (d_hol)
+  {
+    parser->setHOL();
+  }
+
   if( d_logicIsForced ) {
     parser->forceLogic(d_forcedLogic);
   }
@@ -117,6 +123,12 @@ ParserBuilder& ParserBuilder::withParseOnly(bool flag) {
   return *this;
 }
 
+ParserBuilder& ParserBuilder::withHOL(bool flag)
+{
+  d_hol = flag;
+  return *this;
+}
+
 ParserBuilder& ParserBuilder::withOptions(const Options& opts)
 {
   ParserBuilder& retval = *this;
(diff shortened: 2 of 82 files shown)


src/expr/node_manager.cpp
```diff
@@ -531,7 +531,7 @@ TypeNode NodeManager::mkBagType(TypeNode elementType)
   CheckArgument(elementType.isFirstClass(),
                 elementType,
                 "cannot store types that are not first-class in bags. Try "
-                "option --uf-ho.");
+                "parsing option --hol.");
```
ajreynol, line 534: These actually are all outdated.  The NodeManager should never reference parsing options, I'd delete all these checks, which should be done at the API anyways.

src/expr/node_manager.cpp
```diff
@@ -543,7 +543,7 @@ TypeNode NodeManager::mkSequenceType(TypeNode elementType)
   CheckArgument(elementType.isFirstClass(),
                 elementType,
                 "cannot store types that are not first-class in sequences. Try "
-                "option --uf-ho.");
+                "parsing option --hol.");
```
ajreynol, line 546: Ditto.

src/expr/node_manager.h
```diff
@@ -1098,11 +1098,11 @@ inline TypeNode NodeManager::mkArrayType(TypeNode indexType,
   CheckArgument(indexType.isFirstClass(),
                 indexType,
                 "cannot index arrays by types that are not first-class. Try "
-                "option --uf-ho.");
+                "parsing option --hol.");
```
ajreynol, line 1101: Ditto and the 2 others in the h here.

src/parser/parser_builder.cpp
```diff
@@ -143,9 +156,19 @@ ParserBuilder& ParserBuilder::withIncludeFile(bool flag) {
   return *this;
 }
 
-ParserBuilder& ParserBuilder::withForcedLogic(const std::string& logic) {
+ParserBuilder& ParserBuilder::withForcedLogic(const std::string& logic)
+{
   d_logicIsForced = true;
-  d_forcedLogic = logic;
+  if (d_hol)
```
ajreynol, line 162: `d_hol` should impact this method, which talks about the `force-logic` option, the impact of `d_hol` should be to directly call a method LogicInfo::enableHigherOrder.

src/parser/smt2/smt2.cpp
```diff
@@ -503,6 +500,12 @@ Command* Smt2::setLogic(std::string name, bool fromCommand)
   }
 
   d_logicSet = true;
+  if (isHOL())
```
ajreynol, line 503: This has some issues.  One, `name` could already have `HO_` prefix.  

I would expect `isHOL` to force a call to `d_logicInfo.enableHigherOrder` immediately after `d_logic` is initialized, which I guess on the line 506 (master).

src/theory/logic_info.cpp
```diff
@@ -373,7 +373,14 @@ void LogicInfo::setLogicString(std::string logicString)
   enableTheory(THEORY_BUILTIN);
   enableTheory(THEORY_BOOL);
 
+  bool isHOL = false;
+
   const char* p = logicString.c_str();
+  if (!strncmp(p, "HO_", 3))
+  {
+    isHOL = true;
```
ajreynol, line 381: Is there a reason we can't just call enableHigherOrder here?

ajreynol (changes requested):

nafur (commented):
Is `--uf-ho` obsolete now? If so: can we remove it? (we could also make it an alias of the new `--hol`)

HanielB:
> Is `--uf-ho` obsolete now? If so: can we remove it? (we could also make it an alias of the new `--hol`)

Not yet. I'll make it obsolete in a follow-up PR and delete it.

src/parser/smt2/smt2.cpp
HanielB: The issue with that, and the source of the strange control flow it's bothering you here, is that `enableHigherOrder` resets the logic string. That is a common pattern in other `enable...` methods. Do you know why that is? Removing that things could be simplified.

src/expr/node_manager.cpp
```diff
@@ -531,7 +531,7 @@ TypeNode NodeManager::mkBagType(TypeNode elementType)
   CheckArgument(elementType.isFirstClass(),
                 elementType,
                 "cannot store types that are not first-class in bags. Try "
-                "parsing option --hol.");
+                "option --uf-ho.");
```
ajreynol: These checks should be deleted altogether.

src/parser/parser.h
```diff
@@ -192,6 +192,9 @@ class CVC5_EXPORT Parser
  /** The API Solver object. */
  api::Solver* d_solver;
 
+ /** Whether the logic is higher order. */
+ bool d_hol;
```
ajreynol, line 196: Should delete this.

src/parser/smt2/smt2.cpp
```diff
@@ -772,7 +769,7 @@ void Smt2::checkLogicAllowsFunctions()
     parseError(
         "Functions (of non-zero arity) cannot "
         "be declared in logic "
-        + d_logic.getLogicString() + " unless option --uf-ho is used");
+        + d_logic.getLogicString() + " unless logic is prefixed by HO_.");
```
ajreynol, line 772:
```suggestion
        + d_logic.getLogicString());
```

src/theory/logic_info.cpp
```diff
@@ -385,21 +390,45 @@ void LogicInfo::setLogicString(std::string logicString)
     p += 3;
   } else if(!strcmp(p, "QF_ALL")) {
     // the "all theories included" logic, no quantifiers
-    enableEverything();
+    if (d_higherOrder)
```
ajreynol, line 393: Can you explain why `d_higherOrder` is needed to be checked here?  Also, enableEverything should be pulled from each branch; then delete the else branch.

src/theory/logic_info.cpp
```diff
@@ -385,21 +390,45 @@ void LogicInfo::setLogicString(std::string logicString)
...
+    {
+      enableEverything();
+    }
     disableQuantifiers();
     arithNonLinear();
     p += 6;
   } else if(!strcmp(p, "ALL")) {
     // the "all theories included" logic, with quantifiers
-    enableEverything();
+    if (d_higherOrder)
```
ajreynol, line 407: Ditto.

src/theory/logic_info.cpp
```diff
@@ -385,21 +390,45 @@ void LogicInfo::setLogicString(std::string logicString)
...
+    }
     enableQuantifiers();
     arithNonLinear();
     p += 3;
   }
   else if (!strcmp(p, "HORN"))
   {
     // the HORN logic
-    enableEverything();
+    if (d_higherOrder)
```
ajreynol, line 423: Ditto.

ajreynol (changes requested):

src/parser/smt2/smt2.cpp
HanielB: Don't you think it's good to say how to solve the problem?

src/theory/logic_info.cpp
HanielB: The check is necessary because `enableEverything()` invokes the default constructor of `LogicInfo`, which sets `d_higherOrder` to false, since the "everything" case does not include higher-order. This is why the test is made s.t. if we are in the higher-order mode, after we enable everything (and thus set ho to false) we reset higher order to true.

src/parser/smt2/smt2.cpp
ajreynol: You could alternatively say "try adding HO_" to the logic.  As is, the statement is vacuous.

src/theory/logic_info.cpp
ajreynol: Ok, just refactor the branches then.

src/parser/smt2/smt2.cpp
HanielB: I see. I'll do that then.

src/theory/logic_info.cpp
HanielB: Also can't do that. I need to first check `d_higherOrder` before enabling everything and I need to do `enableHigherOrder` after `enableEverything`. I added some documentation to try to clarify that.

src/theory/logic_info.h
```diff
@@ -123,6 +123,9 @@ class CVC5_EXPORT LogicInfo
   /** Is this the all-inclusive logic? */
   bool hasEverything() const;
 
+  /** Is this the all-inclusive higher-order logic? */
```
ajreynol, line 126: I don't like having this method.  Can we just make `hasEverything` robust to whether HO is enabled? Then, this method is equivalent to `hasEverything() && isHigherOrder()`?

ajreynol (changes requested):

src/theory/logic_info.cpp
HanielB: That does simplify it, thanks!

src/theory/logic_info.h
ajreynol: Yes, I recommend changing how hasEverything is implemented.  It would check "ALL" or "HO_ALL" I suppose.

HanielB: OK, I've done that in the last commit.

src/theory/logic_info.cpp
```diff
@@ -275,14 +284,26 @@ std::string LogicInfo::getLogicString() const {
...
+    {
       d_logicString = "QF_ALL";
-    } else {
+    }
+    else
+    {
       size_t seen = 0; // make sure we support all the active theories
-
       stringstream ss;
+      if (isHigherOrder())
```
ajreynol, line 303: Can't we just move this to the beginning?

src/theory/logic_info.cpp
```diff
@@ -275,14 +284,26 @@ std::string LogicInfo::getLogicString() const {
     LogicInfo qf_all_supported;
     qf_all_supported.disableQuantifiers();
     qf_all_supported.lock();
-    if(hasEverything()) {
+    if (hasEverything())
+    {
       d_logicString = "ALL";
-    } else if(*this == qf_all_supported) {
+    }
+    else if (hasEverything(true))
```
ajreynol, line 291: I don't think we need this distinction.  Just print "HO_" if higher order, then "ALL" if hasEverything.

ajreynol (changes requested):

src/theory/logic_info.cpp
```diff
@@ -118,12 +118,21 @@ bool LogicInfo::isHigherOrder() const
 }
 
 /** Is this the all-inclusive logic? */
-bool LogicInfo::hasEverything() const {
-  PrettyCheckArgument(d_locked, *this,
+bool LogicInfo::hasEverything(bool andHol) const
```
ajreynol, line 121: I don't think we should have an argument here.  The use of hasEverything should always be agnostic to higher-order.

src/theory/logic_info.cpp
HanielB: Then I have to change the control flow of the whole method. The way it is seems harmless to me.

src/theory/logic_info.cpp
ajreynol: I don't see how this assymetry impacts anything.  It is much clearer to me to have `hasEverything` not take an argument.

src/theory/logic_info.cpp
ajreynol: The control flow is getting much more confusing in this PR. Let's keep it simple.

src/theory/logic_info.cpp
HanielB: It impacts what conceptually is going on. "Everything" means "ALL".

src/theory/logic_info.cpp
HanielB: Alright, I've changed it so that it's done like you suggested.

ajreynol (changes requested):

src/theory/logic_info.cpp
```diff
@@ -117,13 +117,25 @@ bool LogicInfo::isHigherOrder() const
...
-  PrettyCheckArgument(d_locked, *this,
+bool LogicInfo::hasEverything() const
+{
+  PrettyCheckArgument(d_locked,
+                      *this,
                       "This LogicInfo isn't locked yet, and cannot be queried");
-  LogicInfo everything;
-  everything.lock();
-  return *this == everything;
+  // A logic has everything if all theories are enabled as well as quantifiers
```
ajreynol, line 125: Can't we just do `everything.enableEverything(isHigherOrder())` and then compare equality to this?

ajreynol (changes requested):

src/theory/logic_info.cpp
HanielB: no because then a logic that does not have higher order enabled would fail the test. This would also not scale to the case of separation logic you mentioned.

ajreynol: it would pass, and would be agnostic to sl.

HanielB: It wouldn't. To do as you suggest I'd have to create a new LogicInfo guy, set HOL to it, and return true when one of the comparisons to `everything` and `everythingAndHol` worked. It would indeed be agnostic to separation logic since that's a theory. 

The issue is that the equality operator requires that everything coincides. Since this method is testing just whether a logic *contains* everything, testing like I did seems more natural.

ajreynol (approved):

PR #6698, opened 2021-06-07
docs: Migrate sets and relations theory reference.

aniemetz:
This migrates page https://cvc4.github.io/sets-and-relations.
It further adds the SMT2 version of examples/api/cpp/sets.cpp and adds
test/regress/regress0/rels/relations-ops.smt2 as smtlib example for
relations.


aniemetz:
https://cvc5.github.io/docs-ci/docs-pr6698/

docs/theories/sets-and-relations.rst
```diff
@@ -0,0 +1,201 @@
...
+The source code of these examples is available at:
+
+* `SMT-LIB 2 language example <https://github.com/cvc5/cvc5/blob/master/examples/api/smtlib/sets.smt2>`__
+* `C++ API example <https://github.com/cvc5/cvc5/blob/master/examples/api/cpp/sets.cpp>`__
+* `Python API example <https://github.com/cvc5/cvc5/blob/master/examples/api/python/sets.cpp>`__
+
+
+Below is a short summary of the sorts, constants, functions and
+predicates.  For more details, see
+`this paper at IJCAR 2016 <https://cvc4.github.io/publications/2016/BRBT16.pdf>`__.
```
nafur, line 24: Possibly use the references here?

docs/theories/sets-and-relations.rst
```diff
@@ -0,0 +1,201 @@
...
+
+For the C++ API examples in the table below, we assume that we have created
+a `cvc5::api::Solver solver` object.
+
++----------------------+----------------------------------------------+--------------------------------------------------------------------+
+|                      | SMTLIB language                              | C++ API                                                            |
++----------------------+----------------------------------------------+--------------------------------------------------------------------+
```
nafur, line 31: This table is too wide for our theme. It seems though, that sphinx directly translates the column width of the ASCII art into HTML column widths: they have a fixed width of 16%, 34% and 50%. Maybe it is sufficient to remove spacing from the second column?
Or, alternatively, it seems we can make sphinx leave the column widths calculations to the backend (not sure if this works, see https://docutils.sourceforge.io/docs/ref/rst/directives.html#table):
```suggestion
.. table::
   :widths: auto
+----------------------+----------------------------------------------+--------------------------------------------------------------------+
|                      | SMTLIB language                              | C++ API                                                            |
+----------------------+----------------------------------------------+--------------------------------------------------------------------+
```

docs/theories/sets-and-relations.rst
aniemetz: As discussed offline, I'll do this for all reference in a separate PR.

docs/theories/sets-and-relations.rst
aniemetz: The table can be scrolled, and afaict the whitespace in the source file makes no difference. I don't think it's an issue that you have to scroll for this, some commands are long and we might not get around this.

docs/theories/sets-and-relations.rst
```diff
@@ -0,0 +1,201 @@
...
+Finite Sets
+-----------
+
+cvc5 supports the theory of finite sets.
+The simplest way to get a sense of the syntax is to look at an example:
+
+.. api-examples::
+    ../../examples/api/cpp/sets.cpp
+    ../../examples/api/python/sets.py
+    ../../examples/api/smtlib/sets.smt2
```
nafur, line 13: Do we want to see SMT-LIB by default here?

nafur (approved):
Just one question, otherwise LGTM.

docs/theories/sets-and-relations.rst
aniemetz: I used the same order as for the others, and I'd rather be consistent over all examples in terms of order. I don't see a reason why we should prefer smt2 here.

PR #6709, opened 2021-06-08
Add learned literal manager utility

ajreynol:
This is a simple class to maintain a list of literals that we have learned that may be useful during preprocessing.

This is work towards a "learned rewrite" preprocessing pass / better support for Int/BV translations during preprocessing.

### DIFF
--- a/src/preprocessing/learned_literal_manager.cpp
+++ b/src/preprocessing/learned_literal_manager.cpp
@@ -1,0 +1,55 @@
+/******************************************************************************
+ * Top contributors (to current version):
+ *   Andrew Reynolds
+ *
+ * This file is part of the cvc5 project.
+ *
+ * Copyright (c) 2009-2021 by the authors listed in the file AUTHORS
+ * in the top-level source directory and their institutional affiliations.
+ * All rights reserved.  See the file COPYING in the top-level source
+ * directory for licensing information.
+ * ****************************************************************************
+ *
+ * Learned literal manager
+ */
+
+#include "preprocessing/learned_literal_manager.h"
+
+#include "theory/rewriter.h"
+
+namespace cvc5 {
+namespace preprocessing {
+
+LearnedLiteralManager::LearnedLiteralManager(theory::TrustSubstitutionMap& tls,
+                                             context::UserContext* u,
+                                             ProofNodeManager* pnm)
+    : d_topLevelSubs(tls), d_learnedLits(u)
+{
+}
+
+void LearnedLiteralManager::notifyLearnedLiteral(Node lit)
+{
+  d_learnedLits.insert(lit);
+  Trace("pp-llm") << "LLM:notifyLearnedLiteral: " << lit << std::endl;
+}
+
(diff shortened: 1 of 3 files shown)


src/preprocessing/learned_literal_manager.cpp
```diff
@@ -0,0 +1,55 @@
...
+                               itEnd = d_learnedLits.end();
+       it != itEnd;
+       ++it)
+  {
+    // update based on substitutions
+    Node tlsNode = d_topLevelSubs.get().apply(*it);
+    tlsNode = theory::Rewriter::rewrite(tlsNode);
+    currLearnedLits.push_back(tlsNode);
+    Trace("pp-llm") << "Learned literal : " << tlsNode << " from " << (*it)
+                    << std::endl;
```
nafur, line 49:
```suggestion
  for (const auto& lit: d_learnedLits)
  {
    // update based on substitutions
    Node tlsNode = d_topLevelSubs.get().apply(lit);
    tlsNode = theory::Rewriter::rewrite(tlsNode);
    currLearnedLits.push_back(tlsNode);
    Trace("pp-llm") << "Learned literal : " << tlsNode << " from " << lit
                    << std::endl;
```

src/preprocessing/learned_literal_manager.h
```diff
@@ -0,0 +1,71 @@
...
+ * this information. This class is concerned with the latter kind of literals.
+ */
+class LearnedLiteralManager
+{
+ public:
+  LearnedLiteralManager(theory::TrustSubstitutionMap& tls,
+                        context::UserContext* u,
+                        ProofNodeManager* pnm);
+  /**
+   * Notify learned literal. This method is called when a literal is
```
yoni206, line 45: I am not sure I understand the name -- who gets notified by it? Or is this internal and just triggered? In the latter case, maybe rename/

src/preprocessing/learned_literal_manager.h
```diff
@@ -0,0 +1,71 @@
...
+   * It should be rewritten, and such that top level substitutions have
+   * been applied to it.
+   */
+  void notifyLearnedLiteral(Node lit);
+  /**
+   * Get learned literals, which returns the current set of learned literals
+   * provided to this class. These literals are refreshed so that the current
+   * top-level substitutions are applied to them, and then are rewritten.
+   */
+  std::vector<Node> getLearnedLiterals();
```
yoni206, line 57: Maybe consider renaming to `getRewrittenLearnedLiterals` or something like that? only `get` sounds a bit like a getter. Also, looks like this can be a `const` method?

src/preprocessing/learned_literal_manager.h
ajreynol: We are notifying this class of a learned literal.

src/preprocessing/learned_literal_manager.h
ajreynol: It can be const.

ajreynol: I prefer the simpler name.

yoni206 (approved):

PR #6712, opened 2021-06-09
docs: Fix `Kind` description

4tXJ7f:
This commit changes the `Kind` description not to include C/C++
preprocessor statements and updates the kind of bit-vector addtion.
It also marks some of the information as internal to exclude it from the
public documentation.


nafur (approved):
Ah, nice. I didn't know about `\internal`.

PR #6715, opened 2021-06-09
Push complex check inside GetInstantiationsCommand

nafur:
This PR pushes a rather complex check from the `CommandExecutor` inside the `GetInstantiationsCommand`.
The aim is to only use the `instFormatMode` option in the library (`command.cpp`) but not the main driver (`command_executor.cpp`).


src/main/command_executor.cpp
```diff
@@ -172,16 +172,14 @@ bool CommandExecutor::doCommandSingleton(Command* cmd)
...
 
-    if (options::getDumpInstantiations(d_options)
-        && ((options::getInstFormatMode(d_options)
-                 != options::InstFormatMode::SZS
-             && (res.isSat()
-                 || (res.isSatUnknown()
-                     && res.getUnknownExplanation()
-                            == api::Result::INCOMPLETE)))
-            || isResultUnsat))
+    if (options::getDumpInstantiations(d_options))
```
ajreynol, line 175: I'd prefer not adding Result as an input to GetInstantiations command.

Can we just do a simple static method `static bool GetInstantiations::isEnabled(Result res)` ?

ajreynol (changes requested):

src/main/command_executor.cpp
nafur: This is a good idea, it feels way cleaner.

ajreynol (approved):

PR #6718, opened 2021-06-09
Integrate learned literal manager into preprocessing pass context

ajreynol:
Towards "learned rewrite" preprocessing pass / improvements to Int <-> BV.

### DIFF
--- a/src/preprocessing/preprocessing_pass_context.h
+++ b/src/preprocessing/preprocessing_pass_context.h
@@ -57,9 +57,6 @@ class PreprocessingPassContext
     return d_circuitPropagator;
   }
 
-  /** Get the learned literal manager */
-  LearnedLiteralManager* getLearnedLiteralManager();
-
   context::CDHashSet<Node>& getSymsInAssertions() { return d_symsInAssertions; }
 
   void spendResource(Resource r);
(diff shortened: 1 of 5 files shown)


src/preprocessing/preprocessing_pass_context.h
```diff
@@ -74,6 +77,18 @@ class PreprocessingPassContext
...
   void recordSymbolsInAssertions(const std::vector<Node>& assertions);
 
+  /**
+   * Notify learned literal. This method is called when a literal is
+   * entailed by the current set of assertions.
+   *
+   * It should be rewritten, and such that top level substitutions have
+   * been applied to it.
+   */
+  void notifyLearnedLiteral(Node lit);
```
nafur, line 87: Use `TNode` instead?

src/preprocessing/preprocessing_pass_context.h
```diff
@@ -57,6 +57,9 @@ class PreprocessingPassContext
     return d_circuitPropagator;
   }
 
+  /** Get the learned literal manager */
+  LearnedLiteralManager* getLearnedLiteralManager();
+
```
nafur, line 62: Do we need this? If so: there is no implementation, and why would we return it as a pointer?

nafur (changes requested):

yoni206 (commented):
LGTM, but will let @nafur approve as he had some comments.

src/preprocessing/preprocessing_pass_context.h
ajreynol: Good catch, this was redundant.

nafur (approved):
LGTM.

PR #6719, opened 2021-06-09
Fix type enumeration for non first-class sorts in FMF

### LINKED ISSUE #6690: (finite-model-finder) Assertion error at src/theory/type_enumerator.cpp:88
Hi, for the following formula, cvc5 7608747
~~~~
(set-option :finite-model-find true)
(assert (forall ((a String)) (forall ((b String)) (str.in_re a (re.++ re.allchar (str.to_re b))))))
(check-sat)

cvc5 -q -i --check-models delta.out.smt2
Fatal failure within static cvc5::theory::TypeEnumeratorInterface* cvc5::theory::TypeEnumerator::mkTypeEnumerator(cvc5::TypeNode, cvc5::theory::TypeEnumeratorProperties*) at /home//cvc5/build/src/theory/type_enumerator.cpp:88
Unhandled case encounteredNo type enumerator for type `RegLan'
Aborted (core dumped)
~~~~

ajreynol:
Fixes #6690.


HanielB (approved):
LGTM

PR #6722, opened 2021-06-09
Make `--solve-int-as-bv=X` robust to rewriting

4tXJ7f:
After commit 327a24508ed1d02a3fa233e680ffd0b30aa685a9, the int-to-bv
preprocessing pass is getting rewritten terms. As a result, the terms
can contain negative constants (instead of `(- c)`, i.e., `UMINUS` of a
positive constant) and `NONLINEAR_MULT`. The commit adds support for
those cases, does some minor cleanup, and adds regressions. The
regressions should allow us to detect if/when the preprocessing pass
breaks in the future.


src/preprocessing/passes/int_to_bv.cpp
```diff
@@ -226,7 +226,6 @@ Node intToBV(TNode n, NodeMap& cache)
           {
             Rational constant = current.getConst<Rational>();
             if (constant.isIntegral()) {
-              AlwaysAssert(constant >= 0);
```
yoni206, line 229: Don't other parts of this pp assume all constants are non-negative? For example, is the case of translating `<` and `<=` robust enough for negative constants?

src/preprocessing/passes/int_to_bv.cpp
```diff
@@ -69,7 +70,8 @@ Node intToBVMakeBinary(TNode n, NodeMap& cache)
     }
     else if (current.getNumChildren() > 2
              && (current.getKind() == kind::PLUS
-                 || current.getKind() == kind::MULT))
+                 || current.getKind() == kind::MULT
+                 || current.getKind() == kind::NONLINEAR_MULT))
```
yoni206, line 74: Could there be any other missing operators? If this list does not exhaust all smt-lib operators for integers, perhaps we should assert that `current.getKind()` is not any of the others?

src/preprocessing/passes/int_to_bv.cpp
4tXJ7f: Yes, it should be the same as `UMINUS` of a positive constant (without the widening by one bit because it is not necessary).

src/preprocessing/passes/int_to_bv.cpp
4tXJ7f: There are more operators (like division and mod) but we don't support them in this preprocessing pass. The check/error for unsupported operators is here: https://github.com/cvc5/cvc5/blob/2dc245a5922e660b6f44032d55d452ad27dfc20f/src/preprocessing/passes/int_to_bv.cpp#L169-L175

src/preprocessing/passes/int_to_bv.cpp
```diff
@@ -174,14 +177,14 @@ Node intToBV(TNode n, NodeMap& cache)
             }
             break;
         }
-        for (unsigned i = 0; i < children.size(); ++i)
+        for (size_t i = 0; i < children.size(); ++i)
```
ajreynol, line 179:
```suggestion
        for (size_t i = 0, csize = children.size(); i < csize; ++i)
```

ajreynol (changes requested):

ajreynol (approved):

yoni206 (approved):

PR #6723, opened 2021-06-09
Reorder ITE rewrites

### LINKED ISSUE #6717: Incompleteness because of lost rewrite?
Commit: cbb2fc4
Ubuntu: 18.04 

```
$ cvc4-1.7 -q f1.smt2                                                         
sat                                                                                
$ cvc4-1.8 -q f1.smt2                                                         
sat                                                                                
$ cvc5 -q f1.smt2                                                             
unknown                                                                            
$ cat f1.smt2                                                                 
(declare-fun T () Bool)                                                            
(declare-fun v () String)                                                          
(assert (ite T T true))                                                            
(assert (or T (and (str.prefixof v "") (exists ((x Int)) (= "t" (str.substr v 0 x))))))
(check-sat)                                                                        
```
Formula f1.smt2 is satisfiable by rewriting  `(ite T T true)` to `true` which cvc5 doesn't do. Is this expected behavior?                                                                                  
```                                                                  
$cvc4-1.7 --strings-exp -q f2.smt2                                                 
true                                                                               
$cvc4-1.8 --strings-exp -q f2.smt2                                                 
true                                                                               
$cvc5 --strings-exp -q f2.smt2                                                     
(or (not T) T)                                                                      
$cat f2.smt2                                                                      
(declare-fun T () Bool)                                                            
(simplify (ite T T true))  
```

4tXJ7f:
Fixes #6717. Commit 11c1fba70098cc72c59b2d335249332790287c20 added some
new rewrites for `ITE`. Due to the new rewrites taking precedence over
existing rewrites, it could happen that some of the previous rewrites
did not apply anymore even though they would have further simplified the
ITE. In the example from the issue, `(ite c c true)` was rewritten to
`(or (not T) T)` instead of `(ite T true true)` and then `true`. The
commit fixes the issue by moving rewrites resulting in
conjunctions/disjunctions to the end.


ajreynol (approved):

HanielB (approved):
LGTM, thanks for the fix!

PR #6725, opened 2021-06-10
Ensure bv2nat and int2bv are not rewritten when using solve-bv-as-int

ajreynol:
This PR ensures we do not eagerly rewrite bv2nat and int2bv when using solve-bv-as-int.  Instead they are rewritten during expandDefinitions (at the end of preprocessing).

It also updates regressions that relied on lazy extended function reductions in the lazy solver to use solve-bv-as-int, and adds a missing case (INT_TO_BITVECTOR) in the solve-int-as-bv preprocessing pass.

A followup PR will remove support for lazy extended function reductions for bv2nat / int2bv altogether.


mpreiner (approved):

test/regress/regress0/bv/bv-int-collapse1.smt2
```diff
@@ -1,4 +1,4 @@
-; COMMAND-LINE: --bv-solver=lazy
+; COMMAND-LINE: --solve-bv-as-int=sum
```
yoni206, line 1: how about adding this line while keeping the previous one instead of removing the previous one? Same for other tests here.

ajreynol: I'm planning to delete our prior support for lazy handling of bv2nat in the lazy solver, so this is intentional.

yoni206 (approved):

PR #6726, opened 2021-06-10
Minor simplification to equality engine notifications

ajreynol:
This removes the `d_performNotify` flag from equality engine which was supposedly used to guard a very old dependency concerning notifications during initialization.  We are now robust to this and hence the flag can be removed.


HanielB (approved):
LGTM

PR #6728, opened 2021-06-11
Remove support for lazy BV extended function reductions and inferences

ajreynol:
solve-int-as-bv is now the preferred method for solving these benchmarks.

Adds solve-int-as-bv to a regression that became slow in my previous commit.


mpreiner (approved):
LGTM

yoni206 (approved):

PR #6733, opened 2021-06-11
Add implementation for Lazard evaluation

nafur:
This PR adds the actual implementation for the Lazard evaluation based on CoCoALib. It is only used if CoCoALib is available and falls back to a default libpoly-based implementation otherwise.


ajreynol (approved):
Code style looks good to me; I can't speak to the technical details but I trust this is correct.  LGTM.

src/theory/arith/nl/cad/lazard_evaluation.cpp
```diff
@@ -2,6 +2,933 @@
...
+  }
+  catch (CoCoA::ErrorInfo& e)
+  {
+    e.myOutputSelf(std::cerr);
+    throw;
+  }
+  return {p};
+}
+
+/**
```
HanielB, line 822: shouldn't this documentation be in the header?

nafur: I've added some documentation here (in the source file) as well, as I wanted to say a bit about *how* this is implemented. This is irrelevant for the user (and thus not necessary in the header), but should be somewhere.

HanielB: I generally add this kind of high-level operational explanation to the header as well, but I guess this is fine.

HanielB (approved):
As Andy, in so far as I understand this looks good to me.

PR #6735, opened 2021-06-11
[proofs] [dot] Adding a counter for subproofs

HanielB:


src/proof/dot/dot_printer.cpp
```diff
@@ -46,23 +46,40 @@ std::string DotPrinter::sanitizeString(const std::string& s)
   return newS;
 }
 
+void DotPrinter::countSubproofs(
```
ajreynol, line 49: This is a tree traversal, are you sure you want to do this?

HanielB: For now, yes. It's like that by design. Eventually it'll be optional.

ajreynol: Should maybe make it non-recursive here at least.

HanielB: Changed to that. I already made it a DAG traversal as well, it was very weird otherwise, I think I was just confused before. :)

src/proof/dot/dot_printer.h
```diff
@@ -73,9 +75,17 @@ class DotPrinter
    * @return The string with the special characters escaped.
    */
   static std::string sanitizeString(const std::string& s);
+
+  /** Traverse proof node and populate map of subproofs to their quantity
```
ajreynol, line 79: This is counting the # subproofs underneath each subproof of pn? 

Can you state this?  I had initially thought this was mapping each subproofs to the number of times it occurs.

ajreynol (changes requested):

src/proof/dot/dot_printer.h
HanielB: Updated the doc, PTAL

ajreynol (approved):

Issue #6738, opened 2021-06-12
Incorrect answer with `--bv-assert-input`


4tXJ7f:
Input:
```
(set-logic QF_BV)
(declare-fun N () Bool)                                          
(assert (not (= (_ bv1 1) (bvnot (ite N (_ bv1 1) (_ bv0 1)))))) 
(push)                                                           
(assert (not N))                                                 
(check-sat)                                                      
(pop)                                                            
(check-sat)                                                      
```

With `--incremental`:
```
unsat
sat
```

With `--incremental --bv-assert-input`:
```
unsat
unsat
```

PR #6739, opened 2021-06-12
Final update to SMT-COMP 2021 options

4tXJ7f:
This commit:

- Disables `--tear-down-incremental=X` for the competition since
  it currently does not work correctly on master and a fixed version did
  not show significant benefits.
- Changes the occurrences of `--nl-ext` to `--nl-ext=full` because it
  is now a mode option.
- Removes the use of `--bv-assert-input` because the option currently
  has some issues in incremental mode (#6738)
- Removes the use of `--bitblast=eager` for the model validation track
  because it produces invalid models (#6741)

### DIFF
--- a/contrib/competitions/smt-comp/run-script-smtcomp-current-unsat-cores
+++ b/contrib/competitions/smt-comp/run-script-smtcomp-current-unsat-cores
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-cvc5=./cvc5
+cvc5=$(dirname "$(readlink -f "$0")")/cvc5
 bench="$1"
 
 logic=$(expr "$(grep -m1 '^[^;]*set-logic' "$bench")" : ' *(set-logic  *\([A-Z_]*\) *) *$')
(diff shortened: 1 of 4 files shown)


contrib/competitions/smt-comp/run-script-smtcomp-current-unsat-cores
```diff
@@ -20,10 +20,10 @@ QF_LIA)
   finishwith --miplib-trick --miplib-trick-subs=4 --use-approx --lemmas-on-replay-failure --replay-early-close-depth=4 --replay-lemma-reject-cut=128 --replay-reject-cut=512 --use-soi
   ;;
 QF_NIA)
-  finishwith --nl-ext --nl-ext-tplanes
+  finishwith --nl-ext=full --nl-ext-tplanes
   ;;
 QF_NRA)
-  finishwith --nl-ext --nl-ext-tplanes
+  finishwith --nl-ext=full --nl-ext-tplanes
```
nafur, line 26: Shouldn't we use the same config as for single-query here?

nafur (approved):
Otherwise LGTM.

PR #6740, opened 2021-06-12
[Optimization] Use Result in OptimizationResult

Ouyancheng:
OptimizationResult now contains: 
- cvc5::Result
- optimal value for objective
- whether the objective is unbounded

This gets benefit from cvc5::Result (e.g., we could get explanation for UNKNOWN) and it's slightly easier to integrate to the current API. 

Also refactors BV optimizer so that it uses switch statement (instead of if-then-else) to judge the checkSat results (I was planning to do this a long while ago)...


src/smt/optimization_solver.cpp
```diff
@@ -154,15 +144,21 @@ OptimizationResult::ResultType OptimizationSolver::optimizeBox()
...
   // resets the optChecker
   d_optChecker = createOptCheckerWithTimeout(d_parent);
-  OptimizationResult partialResult;
+  // partialResult defaults to SAT if no objective is present
+  // NOTE: the parenthesis around Result(Result::SAT) is required,
+  // otherwise compiler will report "parameter declarator cannot be qualified"
+  // check this out:
+  // https://stackoverflow.com/questions/44045257/c-compiler-error-c2751-what-exactly-causes-it
+  // and this:
+  // https://en.wikipedia.org/wiki/Most_vexing_parse
```
yoni206, line 160:
```suggestion
  // otherwise the compiler will report "parameter declarator cannot be qualified".
  // For more details:
  // https://stackoverflow.com/questions/44045257/c-compiler-error-c2751-what-exactly-causes-it
  // https://en.wikipedia.org/wiki/Most_vexing_parse
```

src/smt/optimization_solver.h
```diff
@@ -33,63 +33,70 @@ namespace smt {
...
  * The optimization result of an optimization objective
  * containing:
- * - whether it's optimal or not
- * - if so, the optimal value, otherwise the value might be empty node or
- *   something suboptimal
+ * - the optimization result: SAT/UNSAT/UNKNOWN
+ * - the optimal value if SAT and bounded,
+ *   or an empty node if SAT and unbounded,
+ *   otherwise the value might be empty node or something suboptimal
+ * - whether the objective is unbounded
```
yoni206, line 40: Please move this to where the fields are documented. Most of it is already there, except for the corner cases.

src/smt/optimization_solver.h
```diff
@@ -33,63 +33,70 @@ namespace smt {
 /**
  * The optimization result of an optimization objective
  * containing:
- * - whether it's optimal or not
- * - if so, the optimal value, otherwise the value might be empty node or
- *   something suboptimal
+ * - the optimization result: SAT/UNSAT/UNKNOWN
+ * - the optimal value if SAT and bounded,
+ *   or an empty node if SAT and unbounded,
```
yoni206, line 38: Please clarify in the comment what is SAT and (un)bounded.

src/smt/optimization_solver.h
```diff
@@ -33,63 +33,70 @@ namespace smt {
...
+  Result getResult() const { return d_result; }
+
   /**
    * Returns the optimal value.
    * @return Node containing the optimal value,
-   *   if getType() is not OPTIMAL, it might return an empty node or a node
-   *   containing non-optimal value
+   *   if result is unbounded, this will be an empty node,
+   *   if getResult() is not SAT, it might return an empty node or a node
+   *   containing sub-optimal value
```
yoni206, line 76: Please clarify in the comment when it will return an empty node and when a sub-optimal value.

src/smt/optimization_solver.h
```diff
@@ -33,63 +33,70 @@ namespace smt {
...
    **/
   Node getValue() const { return d_value; }
 
+  /**
+   * Checks whether the objective is unbouned
+   * @return whether the objective is unbounded
+   *   if the objective is unbounded, then it's:
+   *   +inf, if it's maximize;
+   *   -inf, if it's minimize
```
yoni206, line 85: This doesn't seem to belong here, because the return value and the field are both Boolean. Please remove.

yoni206 (changes requested):
Thanks, comments inline the review.

src/smt/optimization_solver.h
Ouyancheng: This is exactly the description for class OptimizationResult. So I think it's the appropriate location.

src/smt/optimization_solver.h
Ouyancheng: this is clarified

src/smt/optimization_solver.h
Ouyancheng: this is clarified.

src/smt/optimization_solver.h
Ouyancheng: I couldn't predict when it returns a sub-optimal value. it depends on when the solver returns SAT_UNKNOWN. 
but for UNSAT, it will be an empty node for sure (logically implied by the program).

src/smt/optimization_solver.h
```diff
@@ -33,63 +33,75 @@ namespace smt {
...
+  Result getResult() const { return d_result; }
+
   /**
    * Returns the optimal value.
    * @return Node containing the optimal value,
-   *   if getType() is not OPTIMAL, it might return an empty node or a node
-   *   containing non-optimal value
+   *   if result is unbounded, this will be an empty node,
+   *   if getResult() is UNSAT, it will return an empty node,
+   *   if getResult() is SAT_UNKNOWN, it will return something suboptimal
```
yoni206, line 79: Should this function ever be called when `getResult()` is `SAT_UNKNOWN`? If not, please add an assertion to prevent this.

yoni206 (commented):
Thanks for the changes and the answers. One more thing inline.

src/smt/optimization_solver.h
Ouyancheng: you could call it because for example in BOX optimization, one of the objectives gets UNKNOWN it's UNKNOWN, but the others are still good.

yoni206: I see, ok.

yoni206 (approved):

PR #6753, opened 2021-06-16
Fix model issues with --bitblast=eager.

### LINKED ISSUE #6741: Invalid models with `--bitblast=eager`
```
(set-logic QF_BV)                                        
(declare-fun x () (_ BitVec 1))                          
(declare-fun y () (_ BitVec 1))                          
(assert (= y (ite (= x (_ bv1 1)) (_ bv1 1) (_ bv0 1)))) 
(assert (= y (_ bv1 1)))                                 
(check-sat)                                              
```

`--bitblast=eager --check-models` complains:

```
assertion:     (= z$267 #b1)
simplifies to: false
expected `true'.
```

mpreiner:
For model construction we only compute model values for relevant terms. The set of relevant terms is computed in https://github.com/cvc5/cvc5/blob/master/src/theory/model_manager_distributed.cpp#L58, which skips equalities by default because equalities are usually handled by the equality engine.
When `--bitblast=eager` is enabled all assertions are wrapped into `BITVECTOR_EAGER_ATOM` nodes and passed to the BV solver, which results in equalities below BITVECTOR_EAGER_ATOM nodes not being handled by the equality engine but by the BV solver directly. These equalities, however, are skipped when computing the relevant terms and therefore the BV solver is not asked to compute model values for variables below these equalities.

If `--bitblast=eager` is enabled the BV solver now additionally adds the variables encountered during bit-blasting to the relevant terms via `computeRelevantTerms`.

Fixes #6741

Marking as do-not-merge until @ajreynol can verify that this is the intended use of `computeRelevantTerms`.


src/theory/bv/bitblast/simple_bitblaster.cpp
```diff
@@ -129,6 +129,17 @@ Node BBSimple::getModelFromSatSolver(TNode a, bool fullModel)
...
 }
 
+void BBSimple::computeRelevantTerms(std::set<Node>& termSet)
+{
+  for (const auto& var : d_variables)
+  {
+    if (termSet.find(var) == termSet.end())
+    {
+      termSet.insert(var);
+    }
```
4tXJ7f, line 139:
```suggestion
    termSet.insert(var);
```

Seems a bit redundant to check if `var` is in the set.

src/theory/bv/bv_solver.h
```diff
@@ -80,6 +80,9 @@ class BVSolver
     return TrustNode::null();
   }
 
+  /** Additionally collect terms relevant for collecting model values. */
+  virtual void computeRelevantTerms(std::set<Node>& termSet){};
```
4tXJ7f, line 84: ClangFormat.

src/theory/bv/bitblast/simple_bitblaster.cpp
```diff
@@ -129,6 +129,17 @@ Node BBSimple::getModelFromSatSolver(TNode a, bool fullModel)
   return utils::mkConst(bits.size(), value);
 }
 
+void BBSimple::computeRelevantTerms(std::set<Node>& termSet)
+{
+  for (const auto& var : d_variables)
```
4tXJ7f, line 134: Maybe assert that we are in eager bit-blast mode and add a comment on why it is only necessary in that case?

4tXJ7f (approved):
Makes sense to me but it would be good if @ajreynol could take a look as well. In the meantime, I have some suggestions.

src/theory/bv/bv_solver.h
mpreiner: This is what clang-format does :)

4tXJ7f: Oh interesting, I just tested it and it seems to be because of the trailing `;`. Without that, it adds the space between `)` and `{}` (which is what I expected).

src/theory/bv/bitblast/simple_bitblaster.h
```diff
@@ -53,6 +53,8 @@ class BBSimple : public TBitblaster<Node>
   /** Create 'bits' for variable 'var'. */
   void makeVariable(TNode var, Bits& bits) override;
 
+  /** Add d_variables to termSet. */
```
ajreynol, line 56: Can you give justification for why this is necessary here? I.e. that BV_EAGER_ATOM has terms that may appear beneath ITE.

PR #6758, opened 2021-06-17
Use SAT context level for --bv-assert-input instead of decision level.

### LINKED ISSUE #6738: Incorrect answer with `--bv-assert-input`

mpreiner:
The decision level as previously implemented was not accurate since it
did not consider the user context level. This resulted in facts being
incorrectly recognized as input assertions, which happened for
incremental benchmarks.

Fixes #6738

Marking do-not-merge until cluster runs are done.


4tXJ7f (approved):
LGTM, thanks for fixing this!

mpreiner:
This option doesn't work anymore right now due to the initial push we do for `(reset-assertions)`. Checking for level = 1 doesn't work either since this will break `(reset-assertions)` in the BV solver (since assertions are not removed from the SAT solver).

src/prop/minisat/minisat.cpp
```diff
@@ -258,9 +258,9 @@ bool MinisatSatSolver::isDecision(SatVariable decn) const {
   return d_minisat->isDecision( decn );
 }
 
-int32_t MinisatSatSolver::getDecisionLevel(SatVariable v) const
+int32_t MinisatSatSolver::getSatContextLevel(SatVariable v) const
```
ajreynol, line 261: I'm wondering if there is a better name for this, since this seems like it could be confused with the level of the cvc5::SatContext.  Would `getContextLevel` be a better name?

4tXJ7f (approved):
LGTM

PR #6759, opened 2021-06-17
Fix build without libpoly

4tXJ7f:
Commit f10087c3b347da6ef625a2ad92846551ad324d73 added new files that do
not compile without libpoly. This commit excludes those files when
building without libpoly. It also updates one of the regressions to
ignore a warning about approximate values in the model.


nafur (approved):
LGTM.

PR #6760, opened 2021-06-17
modular bv2int: Stubs and some implementations

yoni206:
This PR adds the file `int_blaster.cpp`. Most of the functions are not yet implemented. Two main functions are implemented.
Additionally, code that will no longer be needed is removed from the BV rewriter.


src/theory/bv/int_blaster.cpp
```diff
@@ -0,0 +1,269 @@
+/*********************                                                        */
```
ajreynol, line 1: I think this is the old header style, do update it.

src/theory/bv/int_blaster.cpp
```diff
@@ -0,0 +1,269 @@
...
+
+Node IntBlaster::mkRangeConstraint(Node newVar, uint64_t k) { return Node(); }
+
+Node IntBlaster::maxInt(uint64_t k) { return Node(); }
+
+Node IntBlaster::pow2(uint64_t k) { return Node(); }
+
+Node IntBlaster::modpow2(Node n, uint64_t exponent) { return Node(); }
+
+Node IntBlaster::makeBinary(Node n)
```
ajreynol, line 74: Does this method need to do a traversal?  Can't we just make this a "one step" method that is called as a first step of the intBlast traversal?  I.e. as intBlast traverses, we makeBinary the current node and then translate to int?

ajreynol (changes requested):

yoni206:
Thanks for the comments! Addressed both of them.

src/theory/bv/int_blaster.cpp
```diff
@@ -0,0 +1,241 @@
...
+    uint64_t currentNumChildren = current.getNumChildren();
+    if (d_intblastCache.find(current) == d_intblastCache.end())
+    {
+      // This is the first time we visit this node and it is not in the cache.
+      // We mark this node as visited but not translated by assiging
+      // a null node to it.
+      d_intblastCache[current] = Node();
+      // all the node's children are added to the stack to be visited
+      // before visiting this node again.
+      for (Node child : current)
```
ajreynol, line 124:
```suggestion
      for (const Node& child : current)
```

src/theory/bv/int_blaster.cpp
```diff
@@ -0,0 +1,241 @@
...
+          || k == kind::BITVECTOR_AND || k == kind::BITVECTOR_OR
+          || k == kind::BITVECTOR_XOR || k == kind::BITVECTOR_CONCAT))
+  {
+    result = n[0];
+    for (uint64_t i = 1; i < numChildren; i++)
+    {
+      result = d_nm->mkNode(n.getKind(), result, n[i]);
+    }
+  }
+    d_binarizeCache[n] = result;
```
ajreynol, line 92: Remember to clang-format.

ajreynol (changes requested):
Thanks for eliminating the traversal, looks a lot nicer now.  A few last minor comments, else LGTM.

ajreynol (approved):

PR #6761, opened 2021-06-17
Make CaDiCaL a required dependency.

mpreiner:
Since the new BV solver is enabled by default and uses CaDiCaL
(and optionally CryptoMiniSat) we make CaDiCaL a required dependency.

Marking do-not-merge until nightlies are fixed accordingly.


4tXJ7f (approved):
LGTM, can we also update the NEWS file?

4tXJ7f:
Since the nightlies have been updated, can we merge this?

PR #6764, opened 2021-06-18
Fix CaDiCaL build on Windows

4tXJ7f:
The Windows version of CaDiCaL does not require the `sys/resource.h`
file, so this commit removes that check. It does, however, require
linking against the Process Status API (psapi), because of a call to
`GetProcessMemoryInfo()`, so this commit adds that dependency for
Windows builds.


mpreiner (approved):
LGTM

PR #6769, opened 2021-06-19
Move cnfConversionTime statistic to CnfStream.

mpreiner:
The statistic in `smt_solver.cpp` was not accurate.


4tXJ7f (approved):
LGTM

4tXJ7f (commented):
Looks like there is some kind of issue, though. Maybe because the timer can be started while it is running?

PR #6772, opened 2021-06-19
[Attributes] Remove parameter `context_dependent`

4tXJ7f:
After commit d70a63324c95210f1d78c2efc46395d2369d2e2b, context-dependent
attributes have not been supported and, as a result, the template
parameter `context_dependent` of `Attribute` has not been used.
Context-dependent attributes also do not fit with our current design of
sharing attributes across different solvers, so it is unlikely that we
will add that feature back in the future. This commit removes the unused
template parameter.


ajreynol (approved):
LGTM, totally agree this should be removed.

PR #6782, opened 2021-06-21
docs: Add quickstart guide.

aniemetz:


aniemetz:
https://cvc5.github.io/docs-ci/docs-pr6782/api/cpp/quickstart.html

docs/api/cpp/quickstart.rst
```diff
@@ -0,0 +1,188 @@
+Quickstart Guide
+================
+
+First, create a cvc5 :cpp:class:`Solver <cvc5::api::Solver>` instance:
```
yoni206, line 4: Probably not for this PR, but I wonder what do you think about adding details on how to build this example, that is, starting one step before.

docs/api/cpp/quickstart.rst
```diff
@@ -0,0 +1,188 @@
...
+Alternatively, ``"QF_ALL"`` enables all logics without quantifiers.
+To optimize the solver's behavior for a more specific logic,
+use the logic name, e.g. ``"QF_BV"`` or ``"QF_AUFBV"``.
+
+.. literalinclude:: ../../../examples/api/cpp/quickstart.cpp
+     :language: cpp
+     :lines: 41
+
+In the following, we will define constraints of reals and integers.
+For this, we first query the solver for the corresponding sorts.
```
yoni206, line 29: Is `query` the right word? sounds a bit like a check-sat query or something.

docs/api/cpp/quickstart.rst
```diff
@@ -0,0 +1,188 @@
...
+
+.. literalinclude:: ../../../examples/api/cpp/quickstart.cpp
+     :language: cpp
+     :lines: 45-46
+
+Now, we create two constants ``x`` and ``y`` of sort ``Real``,
+and two constants ``a`` and ``b`` of sort ``Integer``.
```
yoni206, line 36: What do you think about adding here a clarification about the word `constant`?
Maybe something like:
```
Notice that these are *symbolic* constants, but not actual values.
```

docs/api/cpp/quickstart.rst
```diff
@@ -0,0 +1,188 @@
...
+  value for y: 1/6
+  value for x - y: 0
+
+We can convert these values to C++ types using standard conversion functions.
+
+.. literalinclude:: ../../../examples/api/cpp/quickstart.cpp
+     :language: cpp
+     :lines: 119-121
+
+Another way to independently compute the values of ``x`` and ``y`` would be
```
yoni206, line 115:
```suggestion
Another way to independently compute the value of ``x-y`` would be
```

Would you mind changing this also on the cpp?

docs/api/cpp/quickstart.rst
```diff
@@ -0,0 +1,188 @@
...
+using the C++ minus operator instead of asking the solver.
+However, for more complex terms, it is easier to let the solver do the
+evaluation.
+
+.. literalinclude:: ../../../examples/api/cpp/quickstart.cpp
+     :language: cpp
+     :lines: 119-121
```
yoni206, line 122:
```suggestion
     :lines: 127-135
```

docs/api/cpp/quickstart.rst
```diff
@@ -0,0 +1,188 @@
...
+This time the asserted formula is unsatisfiable:
+
+.. literalinclude:: ../../../examples/api/cpp/quickstart.cpp
+     :language: cpp
+     :lines: 156-157
+
+.. code:: text
+
+  expected: unsat
+  result: unsat
```
yoni206, line 161:
```suggestion
The output will be:
.. code:: text

  expected: unsat
  result: unsat
```

docs/api/cpp/quickstart.rst
```diff
@@ -0,0 +1,188 @@
...
+
+We can query the solver for an unsatisfiable core, that is, a subset
+of the assertions that is already unsatisfiable.
+
+.. literalinclude:: ../../../examples/api/cpp/quickstart.cpp
+     :language: cpp
+     :lines: 161-166
```
yoni206, line 168:
```suggestion
     :lines: 161-167
```

yoni206 (commented):
Looks very good, thanks for this! Some ideas and small fixes inline.

docs/api/cpp/quickstart.rst
aniemetz: Hm. This is the quickstart guide for the API, not sure if we should have build instructions there. I'm thinking this should probably rather be mentioned in https://cvc5.github.io/docs/examples/examples.html. @nafur what do you think?

docs/api/cpp/quickstart.rst
aniemetz: It is the right word, yes. You ask the solver for the sort.

docs/api/cpp/quickstart.rst
aniemetz: Oh, I thought I had copied that from the example. It seems that I didn't

yoni206 (approved):
Thanks for this, and for the changes.

PR #6783, opened 2021-06-21
Fix cases of ITE within regular expressions

### LINKED ISSUE #6776: Refutation soundness issue on a string formula
Commit: https://github.com/cvc5/cvc5/commit/5e6117cc183513bf676b36078e6507b31caa6ff0
OS: Ubuntu 18.04

```
[596] % cvc5 -q small.smt2
unsat
[597] % cvc4-1.8 -q small.smt2
unsat
[598] % z3 small.smt2
sat
[599] % cat small.smt2
(assert (str.in_re "A" (ite (= (div 0 0) 0) (str.to_re "A") (str.to_re "B"))))
(check-sat)
[600] % 
```

ajreynol:
Fixes #6776.

Our computation of when a regular expression was constant did not account for when ITE was embedded in an RE, leading to an unsound rewrite.

That benchmark now gives:
```
(error "Regular expression variables are not supported.")
```


src/theory/strings/regexp_entail.cpp
```diff
@@ -364,6 +364,10 @@ bool RegExpEntail::isConstRegExp(TNode t)
           }
         }
       }
+      else if (ck==ITE)
+      {
+        return false;
+      }
```
4tXJ7f, line 370: Apply ClangFormat.

4tXJ7f (approved):
Not what I expected :) LGTM

PR #6784, opened 2021-06-22
Fix linear arithmetic for duplicate lemmas in incremental

### LINKED ISSUE #6773: (default) Invalid model for NIA formula
cvc5 5e6117cc183513bf676b3607
~~~~
(set-logic NIA)
(declare-fun i3 () Int)
(declare-fun i9 () Int)
(push)
(assert (>= (- 1 1 i3 i9 1) i3))
(check-sat)
(pop)
(assert (or (= 0 i9) (< i3 0)))
(assert (= 1 (* i9 i9)))
(check-sat)
~~~~
~~~~
cvc5 -q -i --check-models xx.smt2
sat
Fatal failure within void cvc5::smt::CheckModels::checkModel(cvc5::smt::Model*, cvc5::context::CDList<cvc5::NodeTemplate<true> >*, bool) at /home/cvc5/src/smt/check_models.cpp:125
assertion:     (= 1 (* i9 i9))
simplifies to: false
expected `true'.
Aborted (core dumped)
~~~~

ajreynol:
The linear arithmetic solver was not robust to cases where a duplicate lemma is emitted. This leads to non-linear arithmetic not being called to check at full effort, leading to potential model soundness issues.

Fixes #6773.


nafur (approved):
LGTM.

PR #6786, opened 2021-06-22
FP: Remove sections guarded with undefined macro SYMFPUPROPISBOOL.

aniemetz:


martin-cs (approved):
This is fine technically.  The use of Boolean types rather than a BV(1) was an early idea and ... there was some reason why I didn't pursue it.  In addition I think the BV(1) approach performs better and there are better ways to do the splitting to cheaper and more expensive terms.

PR #6787, opened 2021-06-22
Avoid full normalization of lambdas in getValue

ajreynol:
This ensures that we don't apply lambda rewriting, which involves array value normalization, to lambda terms returned by `TheoryModel::getValue`. 

This can significantly speed up our time to return function terms for getValue.


4tXJ7f (approved):
LGTM, thanks for the fix.

PR #6789, opened 2021-06-23
Porting C++ API examples to SMT-LIB examples

HanielB:
SyGuS examples will come later.


examples/api/smtlib/quickstart.smt2
```diff
@@ -2,20 +2,38 @@
...
+;   (1)  0 < x
+;   (2)  0 < y
+;   (3)  x + y < 1
+;   (4)  x <= y
+
 (assert (< 0 x))
 (assert (< 0 y))
 (assert (< (+ x y) 1))
-(assert (<= x y))
+(assert (<= 0 x y))
```
aniemetz, line 22: I would rather keep this line as it was so that it matches (4).

HanielB: This was a mistake. I fixed now.

aniemetz (approved):

PR #6793, opened 2021-06-24
pow2: Adding monotonicity lemma 

yoni206:
We add the lemma x<=y --> pow2(x)<=pow2(y) to the pow2 solver.
Additionally, some renaming of variables is introduced for better clarity.

### DIFF
--- a/src/theory/arith/nl/pow2_solver.cpp
+++ b/src/theory/arith/nl/pow2_solver.cpp
@@ -91,7 +91,6 @@ void Pow2Solver::checkFullRefine()
   Trace("pow2-check") << "Pow2Solver::checkFullRefine";
   Trace("pow2-check") << "pow2 terms: " << std::endl;
   NodeManager* nm = NodeManager::currentNM();
-  std::vector<Node> pow2Args;
 
   struct
   {
--- a/src/theory/arith/nl/pow2_solver.cpp
+++ b/src/theory/arith/nl/pow2_solver.cpp
@@ -15,12 +15,16 @@
 
 #include "theory/arith/nl/pow2_solver.h"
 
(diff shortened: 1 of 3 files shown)


ajreynol (changes requested):
You are using a quadratic algorithm that sends redundant lemmas :)  Do write this as using a sort algorithm; then the lemmas are linear.  i.e. pow2(t1) .... pow2(tn), sort by t1^M .... tn^M then only add n-1 lemmas C => pow2(t_i1) < pow2( t_i2 ) .... C => pow2( t_i{n-1}) < pow2( t_in ).

src/theory/arith/nl/pow2_solver.cpp
```diff
@@ -90,31 +90,68 @@ void Pow2Solver::checkFullRefine()
 {
   Trace("pow2-check") << "Pow2Solver::checkFullRefine";
   Trace("pow2-check") << "pow2 terms: " << std::endl;
-  for (const Node& i : d_pow2s)
+  NodeManager* nm = NodeManager::currentNM();
```
ajreynol, line 93: Can you move this to its own method?

ajreynol (changes requested):

src/theory/arith/nl/pow2_solver.cpp
yoni206: done

ajreynol (changes requested):
Would be good to have a regression to test the code works already, or is this not possible yet?

yoni206:
Regressions are 1 PR away, if that's ok with you

ajreynol (approved):

PR #6800, opened 2021-06-25
pow2 -- final changes

yoni206:
This PR adds the remaining changes for a working and integrated `pow2` solver.

In particular, it adds a rewrite and a lemma that identify `pow2(x)` with `0` whenever `x<0`.
Regressions are added as well, including `pow2-native-0.smt2` that shows the semantics of `pow2` on negative values.

The next steps are new rewrites and and more lemma schemas.


src/theory/arith/nl/pow2_solver.cpp
```diff
@@ -148,6 +152,17 @@ void Pow2Solver::checkFullRefine()
         }
     }
 
+    // triviality lemmas: pow2(x) = 0 whenever
+    // x < 0
+    if (x < 0)
+    {
+      Node assumption = nm->mkNode(LT, n[0], d_zero);
```
ajreynol, line 159: Can you guard this so that it only applies when the premise is true in the model?

ajreynol (changes requested):
One minor request, else looks good.

src/theory/arith/nl/pow2_solver.cpp
yoni206: done

ajreynol (approved):
LGTM

PR #6806, opened 2021-06-28
Rewrite POW to POW2 when the base is 2

yoni206:
This PR introduces a rewrite from `(^ 2 t)` to `(pow2 t)` in order to make use of the specialized `pow2` solver.
One regression that expects an error when `t` is not a constant is changed accordingly.


ajreynol (approved):

yoni206:
FYI @HanielB the benchmarks with `^` can now stay as they are.

PR #6808, opened 2021-06-29
Fix initialization of `ConjectureGenerator`

4tXJ7f:
Commit 525fe1f4f33aca05f5e92b0cc0f3e0c7e6effa8b removed a flag in the
equality engine that prevented the `EqualityEngine` from notifying
others about the terms created during initialization. This was a problem
in `ConjectureGenerator` because
`ConjectureGenerator::eqNotifyNewClass()` pushes to `d_upendingAdds`,
which is created after the `EqualityEngine`. There are at least three
ways to solve this:

- Revert commit 525fe1f4f33aca05f5e92b0cc0f3e0c7e6effa8b, which would
  add back additional checks in the `EqualityEngine`.
- Move `d_upendingAdds` before `d_uequalityEngine`, which seems a bit
  brittle, especially because the constructor creates a class that calls
  back into the partially constructed class (AFAIK the C++ standard
  allows it, though).
- Add an `init()` method for `QuantifiersModule`, which allows it to
  do perform additional setup after the full class has been
  initialized.

This commit implements the third option since it seemed relatively
straightforward.


ajreynol:
Excellent catch. LGTM.

ajreynol (dismissed):

ajreynol:
Actually one alternate fix would be to initialize the notify class of equality engine after initialization? i.e. set it to `s_notifyNone` and then initialize the proper value of the notify class afterwards.  This would address the downside you mention I believe.  Then, the only thing that may be slightly different is that users are not notified when the classes `true` and `false` are created (which no one cares about anyways).

4tXJ7f:
I've implemented your suggestion. I've had to change the `d_notify` reference in the `EqualityEngine` to a pointer because references cannot be rebound in C++.

PR #6809, opened 2021-06-29
Add some printing functions for OptimizationObjective and OptimizationResult

Ouyancheng:
Implement the `operator <<` with `std::ostream` for `OptimizationObjective` and `OptimizationResult`. Currently only supports SMTLib2 or Sygus as output languages. 
Objective: (maximize [node] :signed) or (minimize [node]) or ... 
Result: (sat [expr]) or (unsat) or (unknown([explanation]) [expr/null])


src/smt/optimization_solver.cpp
```diff
@@ -28,6 +30,69 @@ using namespace cvc5::omt;
...
+  if ((!language::isOutputLang_smt2(lang))
+      && (!language::isOutputLangSygus(lang)))
+  {
+    Unimplemented()
+        << "Only SMTLib2 or Sygus languages support optimization right now";
+  }
+  out << "(" << result.getResult();
+  switch (result.getResult().isSat())
+  {
+    case Result::SAT: CVC5_FALLTHROUGH;
```
yoni206, line 46: [This](https://en.cppreference.com/w/cpp/language/attributes/fallthrough#:~:text=A%20fallthrough%20statement%20may%20only,same%20iteration%20of%20that%20loop.) suggests that the `fallthrough` attribute relates to a previous case, but this is the first case. Do you get a compiler warning if you remove this?

src/smt/optimization_solver.cpp
```diff
@@ -28,6 +30,69 @@ using namespace cvc5::omt;
 namespace cvc5 {
 namespace smt {
 
+std::ostream& operator<<(std::ostream& out, const OptimizationResult& result)
```
yoni206, line 33: I wonder if this code should go to the smt2 printer rather than to be implemented here. @ajreynol what do you think?

yoni206 (commented):
Looks good, thanks for this. A couple of questions inline.

src/smt/optimization_solver.cpp
Ouyancheng: I haven't tried. But I got a warning when I forgot to add a break at another switch statement.

src/smt/optimization_solver.cpp
Ouyancheng: smt::Result also has its individual << operator.

src/smt/optimization_solver.cpp
Ouyancheng: removing `[[fallthrough]]` seems to work. I got a warning because the case I was working on has some code there instead of direct falling through. but I would prefer keeping `CVC5_FALLTHROUGH`, in case you use an older standard like C++0x, you could just make the macro to be empty.

yoni206: Well as it is right now, it is in the first case, while the documentation of `fallthrough` only mentions how it is used when there are cases before it. Hence it is not clear to me how this is useful here. Please remove it. If you do find a good reason to keep it, which is consistent with `fallthrough` documentation, please document it in the code and we can iterate over it.

Ouyancheng: I think you misunderstood cppreference. It means: the case using [[fallthrough]] is the "previous case", while the case following it is the "next case". Your understanding might be: the case using [[fallthrough]] is "current case", the case before that is "previous case" and the following "current case" is the "next case". Well... I don't see any reason why it forbids using this in the first case. check proposal P0188R1 here: https://gcc.gnu.org/projects/cxx-status.html

Ouyancheng: OK, removing [[fallthrough]] looks a bit cleaner. I removed it purely because of aesthetic issue.

yoni206 (approved):

Ouyancheng:
@ajreynol @yoni206 What are your opinions?

yoni206:
The PR is approved.

yoni206:
@ajreynol , if you are ok with the PR, please merge when you can.

ajreynol:
For the SyGuS standard, a response to
`(optimize-synth (t1 ... tn))` 
is:
`infeasible | fail | ( (c1 ... cn) f1 ... fn)`
where f1 ... fn are solutions to the functions to synthesize and c1 ... cn are the evaluation of terms t1 ... tn in that solution.

Can you clarify what you expect the result for
`(optimize-sat (t1 ... tn))`
to be?

Ouyancheng:
> For the SyGuS standard, a response to
> `(optimize-synth (t1 ... tn))`
> is:
> `infeasible | fail | ( (c1 ... cn) f1 ... fn)`
> where f1 ... fn are solutions to the functions to synthesize and c1 ... cn are the evaluation of terms t1 ... tn in that solution.
> 
> Can you clarify what you expect the result for
> `(optimize-sat (t1 ... tn))`
> to be?

that would be something like
```
(
    (t1 \t (sat \t [expr]))
    (t2 \t (unknown([explanation]) \t [expr/null]))
)
or 
(unsat) 
```
and a single result is `(sat \t [expr])/(unsat)/(unknown[...], [expr])`. 
let me remove the Sygus thing for now.

PR #6811, opened 2021-06-29
int-to-bv: correct model values

yoni206:
the `int-to-bv` preprocessing pass produced wrong models.
This PR fixes this in a similar fashion to other preprocessing passes, by adding a substitution to the preprocessing pass context. This requires moving the main translation function to be a class method, rather than a helper method in an empty namespace.

Thanks to @alex-ozdemir for raising this issue and producing a triggering benchmark (added to regressions in this PR).

### DIFF
--- a/src/preprocessing/passes/int_to_bv.cpp
+++ b/src/preprocessing/passes/int_to_bv.cpp
@@ -29,6 +29,7 @@
 #include "options/base_options.h"
 #include "options/smt_options.h"
 #include "preprocessing/assertion_pipeline.h"
+#include "preprocessing/preprocessing_pass_context.h"
 #include "theory/rewriter.h"
 #include "theory/theory.h"
 #include "util/bitvector.h"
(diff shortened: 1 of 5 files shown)


ajreynol (approved):
LGTM.

Btw, if this preprocessing pass is used in incremental mode, you should handle BITVECTOR_TO_NAT as a case, i.e. (bv2nat x) ---> x.

yoni206:
Thanks!
I see that for now this pass does not support incremental solving:
https://github.com/cvc5/cvc5/blob/373b6e1e8e91874afab16416f7acc3839f0027af/src/smt/set_defaults.cpp#L177

Do you know why? I don't see, for example, any fields that would need to be changed to context dependent data structures. Should we just try to enable it and run on smtlib?

ajreynol:
Let's do this separately.  It's fine for now if incremental is disabled.

We may need to consider making some data structures context-dependent, and to ensure that bv2nat is handled, since variables x will be replaced by (bv2nat k) on future check-sat calls.

PR #6817, opened 2021-06-30
Do not notify during equality engine initialization

ajreynol:
This is an alternative fix to https://github.com/cvc5/cvc5/pull/6808.

This ensures that we do not notify the provided class of an equality engine during initialization.


4tXJ7f (approved):
LGTM, that's a more elegant solution.

PR #6820, opened 2021-06-30
Fix bv assert input reset assertions

mpreiner:
If reset-assertions was called it will now reset the SAT solver and the
CNF stream if clauses were permanently added to the SAT solver via
option --bv-assert-input.


src/theory/bv/bv_solver_bitblast.cpp
```diff
@@ -26,6 +26,46 @@ namespace cvc5 {
...
+      d_reset_assertions_notified = true;
+    }
+  }
+
+ private:
+  /** The user-context. */
+  context::Context* d_context;
+
+  /** Flag to notify whether reset assertions was called. */
+  bool d_reset_assertions_notified;
```
4tXJ7f, line 66: camelCase

src/theory/bv/bv_solver_bitblast.h
```diff
@@ -153,6 +157,9 @@ class BVSolverBitblast : public BVSolver
 
   /** Option to enable/disable bit-level propagation. */
   bool d_propagate;
+
+  /** Notifies when reset-assertion was called. */
+  std::unique_ptr<NotifyResetAssertions> d_reset_notify;
```
4tXJ7f, line 162: camelCase

src/theory/bv/bv_solver_bitblast.cpp
```diff
@@ -26,6 +26,46 @@ namespace cvc5 {
...
+{
+ public:
+  NotifyResetAssertions(context::Context* c)
+      : context::ContextNotifyObj(c, false),
+        d_context(c),
+        d_reset_assertions_notified(false)
+  {
+  }
+
+  bool notifiedResetAssertions() { return d_reset_assertions_notified; }
```
4tXJ7f, line 46: Suggestion: it sounds a bit like this method notifies reset-assertions. Maybe `doneAssertionReset`?

4tXJ7f (approved):
Great :) Please update the variable names before merging.

PR #6824, opened 2021-07-01
Add recursive function definitions to subsolver in sygus

ajreynol:
This passes recursive function definitions to the verification subsolver in sygus, with a default bounded limit of 3.  This required improving the interface for setting up subsolvers by allowing custom options; the sygus solver statically computes an instance of the options that it uses in all calls.

This allows us to solve non-PBE sygus problems with recursive function definitions.  The PR adds an example.


src/theory/smt_engine_subsolver.h
```diff
@@ -41,10 +41,13 @@ namespace theory {
  * if the current SMT engine has declared a separation logic heap.
  *
  * @param smte The smt engine pointer to initialize
+ * @param opts The options for the subsolver. If nullptr, then we copy the
+ * options from the current SmtEngine in scope.
  * @param needsTimeout Whether we would like to set a timeout
  * @param timeout The timeout (in milliseconds)
  */
 void initializeSubsolver(std::unique_ptr<SmtEngine>& smte,
+                         Options* opts = nullptr,
```
HanielB, line 50: Passing this kinda voids the need for the other two parameters. Maybe we want to simplify the interface? We could have a wrapper around this method which takes just the timeout for the case when one wants to change just the timeout.

src/theory/quantifiers/sygus/synth_conjecture.cpp
```diff
@@ -580,8 +593,30 @@ bool SynthConjecture::doCheck(std::vector<Node>& lems)
 
   if (!query.isConst() || query.getConst<bool>())
   {
+    // add recursive function definitions
+    FunDefEvaluator* feval = d_tds->getFunDefEvaluator();
+    const std::vector<Node>& fdefs = feval->getDefinitions();
+    if (!fdefs.empty())
+    {
+      // get the relevant definitions based on the symbols in the query
```
HanielB, line 601: How important is this? Like, if we were to pass formulas that ranged over symbols not in the query it'd be a non-issue? I ask because maybe this could be simplified then for us to always pass the recfuns relevant to all symbols in the grammar.

ajreynol: This is important. You only want to pass the recursive functions you need. In the regression added here, this actually may make the difference of decidability vs undecidability in the verfication step: candidates that don't involve the recursive function have a decidable verification step, and CEGIS gets a proper counterexample point.

src/theory/smt_engine_subsolver.h
ajreynol: I'd rather avoid having too many variants of the same method.  Indeed, the interface is slightly redundant but it still allows casual use with timeout (without having to deal with an explicit options object), this comes at the cost of passing an additional `nullptr` but I don't think this is too big of a deal.

HanielB: OK.

src/theory/quantifiers/sygus/synth_conjecture.cpp
HanielB: I see. I had not thought about making a difference that way, it'd be good to add a comment clarifying.

HanielB (approved):

PR #6826, opened 2021-07-02
Add output tags -o, --output.

mpreiner:
Output tags are similar to debug/trace tags, but are always enabled
(except for muzzled builds) to provide useful information for users.

Available output tags can be queried via -o help/--output help and are
specified in the base options module via enum values.

Note: Output/OutputC is defined and implemented in a separate header/cpp file `options/outputc.{h,cpp}` since we don't want to pull in `options/base_options.h` into `base/output.h` since this would trigger a lot of rebuilds when adding a new output tag or base option.
@nafur I had to do some workarounds in `mkoptions.py` (`mode_handler` flag) in order to allow mode options with handlers.


src/options/outputc.cpp
```diff
@@ -0,0 +1,26 @@
...
+  }
+  else
+  {
+    return Cvc5ostream();
+  }
+}
+
+bool OutputC::isOn(const options::OutputTag tag) const
+{
+  return options::outputTagHolder()[static_cast<size_t>(tag)] == 1;
```
nafur, line 23:
```suggestion
  return options::outputTagHolder()[static_cast<size_t>(tag)];
```

src/options/outputc.h
```diff
@@ -0,0 +1,38 @@
...
+
+ private:
+  std::ostream* d_os;
+
+}; /* class OutputC */
+
+extern OutputC OutputChannel CVC5_EXPORT;
+
+#ifdef CVC5_MUZZLE
+#define Output ::cvc5::__cvc5_true() ? ::cvc5::nullStream : ::cvc5::OutputChannel
```
nafur, line 31: Why can't we use 
```
#define Output true ? ::cvc5::nullStream : ::cvc5::OutputChannel
```
or even
```
#define Output ::cvc5::nullStream
```
here?

nafur (commented):
Two comments...

src/theory/quantifiers/instantiate.cpp
```diff
@@ -690,7 +692,7 @@ void Instantiate::notifyEndRound()
             << " * " << i.second << " for " << i.first << std::endl;
       }
     }
-    if (options::debugInst())
+    if (outputInst)
```
ajreynol, line 695: Update the use of `out` below to `Output(options::OutputTag::INST);`

ajreynol (changes requested):
Happy to have this. One minor comment.

src/options/outputc.h
mpreiner: Here's what the comment for `__cvc5_true` says:
```
// Disallow e.g. !Debug("foo").isOn() forms
// because the ! will apply before the ? .
// If a compiler error has directed you here,
// just parenthesize it e.g. !(Debug("foo").isOn())
```
Now I wonder why e.g. `Debug` is not defined as:

```
#define Debug (::cvc5::__cvc5_true() ? ::cvc5::nullStream : ::cvc5::DebugChannel)
```

and as you said, if we'd just just `#define Debug ::cvc5::nullStream`, there would be no need for cvc5_true at all...

mpreiner: Oh OK I think I know why...

```
Debug("foo") << "bar";
```

expands to

```
Debug ::cvc5::__cvc5_true() ? ::cvc5::nullStream : ::cvc5::DebugChannel("foo") << "bar"
```

for muzzled builds `::cvc5::DebugChannel("foo") << "bar"` will get optimized away.

src/theory/quantifiers/instantiate.cpp
mpreiner: I refactored printing in the whole function since `outputInst` and `debugInstTrace` don't have anything in common.

ajreynol (approved):

PR #6827, opened 2021-07-02
Fix rewriter for negative constant seq.nth

### LINKED ISSUE #6777: (strings-lazy-pp, check-unsat-cores) Invalid model for string formula
cvc5 5e6117cc183513bf676b36
~~~~
(set-option :check-unsat-cores true)
(declare-fun b () (Seq Int))
(declare-fun y () Int)
(declare-fun l () Bool)
(assert (distinct (seq.nth b y) (seq.nth b 1)))
(assert (= l (> y 2)))
(check-sat)
~~~~
~~~~
cvc5 -q -i --check-models xx.smt2
Fatal failure within void cvc5::smt::CheckModels::checkModel(cvc5::smt::Model *, context::CDList<Node> *, bool) at /home/cvc5/src/smt/check_models.cpp:125
assertion:     (distinct (seq.nth b y) (seq.nth b 1))
simplifies to: false
expected `true'.
Aborted
~~~~

ajreynol:
Fixes #6777.


4tXJ7f (approved):
Ah, so that was the issue :) Great catch!

PR #6831, opened 2021-07-02
Refactor setup of proof equality engine for central EE

ajreynol:
Users of an equality engine should each use the same proof equality engine that wraps it. This ensures that theory inference managers do so, by tracking the official proof equality engine for an equality engine in theory inference manager.

This is in preparation for (proper equality proofs for) central equality engine.

It also adds some debugging code that is highly useful for debugging issues related to when equalities are processed in theory inference manager.


src/theory/theory_inference_manager.cpp
```diff
@@ -388,6 +394,39 @@ bool TheoryInferenceManager::processInternalFact(TNode atom,
   Trace("infer-manager") << "TheoryInferenceManager::assertInternalFact: "
                          << (pol ? Node(atom) : atom.notNode()) << " from "
                          << expn << std::endl;
+#ifdef CVC5_ASSERTIONS
```
4tXJ7f, line 397: Could we use `Configuration::isAssertionBuild()` instead?

4tXJ7f (approved):
LGTM, one minor suggestion.

src/theory/theory_inference_manager.cpp
ajreynol: I've updated this.  I assume this will be 100% compiled away in non-assertion builds?  Asking since this is performance critical code.

PR #6833, opened 2021-07-02
Make buffered inference manager more robust to backtracking

ajreynol:
This makes TheoryEngine notify all theories when a theory sends a conflict.  This means that buffered inference managers always clear their buffers when any theory sends a conflict.

This is required for making theories robust to conflicts that may arise when using the central equality engine, where a different theory may raise a conflict during another theory's check.


nafur (approved):
LGTM.

PR #6839, opened 2021-07-05
Add script to build wheel for pycvc5

makaimann:
This PR adds a script for building a wheel distribution for pycvc5. It automatically reads the top-level `CMakeLists.txt` to obtain the current version number, and then runs the standard `setup` from `setuptools` with an extension command class that configures and builds cvc5 with Python bindings. This wheel file is sufficient for uploading to Pypi. Although, note that we need to build for different operating systems and different Python versions because of the extension modules.

Note: another option is to use the `scikit-build` instead of doing the configuring and building manually. However, I ran into some issues when the `setup.py` file was not at the top-level. Because we prefer to have it hidden in a sub-folder, I took the manual route for now. In the future, we could consider changing this.


src/api/python/wheels/build_wheel.py
```diff
@@ -0,0 +1,155 @@
...
+    pattern = re.compile(str_pattern)
+    with open(project_src_path + '/CMakeLists.txt', 'r') as f:
+        for line in f.read().split('\n'):
+            line = line.strip()
+            m = pattern.search(line)
+            if m:
+                gd = m.groupdict()
+                version[gd['component']] = gd['version']
+                if len(version) == 3:
+                    break
```
nafur, line 60: Instead of reading everything in a string and then splitting it, I'd suggest to either actually read line by line (`for line in f`) or read everything in a string and to `pattern.finditer()` to directly find all matches in the string.

makaimann: Good point! Fixed with 995324034246c5debadd6635f7fbfe5c4b8daaa1.

src/api/python/wheels/build_wheel.py
```diff
@@ -0,0 +1,155 @@
...
+
+
+def get_cvc5_version():
+    project_src_path = get_project_src_path()
+
+    # read CMakeLists.txt to get version number
+    version = dict()
+    str_pattern = 'set\(CVC5_(?P<component>MAJOR|MINOR|RELEASE)\s*(?P<version>\d+)\)'
+    pattern = re.compile(str_pattern)
+    with open(project_src_path + '/CMakeLists.txt', 'r') as f:
```
mpreiner, line 52:
```suggestion
    with open(os.path.join(project_src_path, '/CMakeLists.txt'), 'r') as f:
```

src/api/python/wheels/build_wheel.py
```diff
@@ -0,0 +1,155 @@
...
+        build_args = ['--config', cfg]
+
+        cpu_count = max(2, multiprocessing.cpu_count() // 2)
+        build_args += ['--', '-j{0}'.format(cpu_count)]
+
+        project_src_path = get_project_src_path()
+        build_dir = os.path.join(project_src_path, "build")
+
+        # to avoid multiple build, only call reconfigure if we couldn't find the makefile
+        # for python
```
mpreiner, line 114: Is this actually an issue? How do you make sure that the Makefile is up-to-date?

src/api/python/wheels/build_wheel.py
```diff
@@ -0,0 +1,155 @@
...
+
+    @staticmethod
+    def is_windows():
+        tag = platform.system().lower()
+        return tag == "windows"
+
+    @staticmethod
+    def is_linux():
+        tag = platform.system().lower()
+        return tag == "linux"
```
mpreiner, line 98: Seems to be unused.

src/api/python/wheels/build_wheel.py
makaimann: Yeah it's not that big of a deal, I only kept it in because it was in the one from smt-switch tbh. It is possible that the Makefile is stale and configured with python. Maybe a better alternative would be to use a special build directory name? Something like `python-wheel`. And if it already exists, we could just throw an exception? I'm not sure it'd be a good idea to automatically delete a whole build folder even if it's a dedicated name.

src/api/python/wheels/build_wheel.py
makaimann: Good idea, added in a1f6c9c540e12bdca91526ea28e94ed2eaf545d9.

src/api/python/wheels/build_wheel.py
makaimann: Good catch, addressed in a1f6c9c540e12bdca91526ea28e94ed2eaf545d9.

src/api/python/wheels/build_wheel.py
mpreiner: Why not just reconfigure and rebuild?

makaimann: We could do that. I just worried about messing up someone's existing build.

mpreiner: How about creating a `build-python-wheel` directory and use this as working directory?

src/api/python/wheels/build_wheel.py
```diff
@@ -0,0 +1,150 @@
...
+##
+
+import os
+import re
+import sys
+import platform
+import subprocess
+import multiprocessing
+import shutil
+import glob
```
mpreiner, line 29: Seems to be unused.

makaimann: Thanks, fixed with 5e08ae1f2f2ec4602e451fc9670ae7970cf031b1.

src/api/python/wheels/build_wheel.py
makaimann: Yeah sounds good. Updated with d4b75d15820298812c8acce8678fd614a726f7b9.

mpreiner (approved):
Thanks for the changes!

PR #6843, opened 2021-07-06
Add implementation of learned rewrite pass

ajreynol:


src/preprocessing/passes/learned_rewrite.cpp
```diff
@@ -155,14 +155,293 @@ Node LearnedRewrite::rewriteLearnedRec(Node n,
...
+  visit.push_back(n);
+  do
+  {
+    cur = visit.back();
+    visit.pop_back();
+    it = visited.find(cur);
+
+    if (it == visited.end())
+    {
+      visited[cur] = Node::null();
```
yoni206, line 171: Is this completely standard in your preprocessing passes? If not (and maybe even if yes), please consider adding a comment about the meaning of not being in the map vs. being assigned null node in the map vs. being assigned a non-null node in the map.

yoni206 (approved):
One minor comment.

src/preprocessing/passes/learned_rewrite.cpp
ajreynol: This is very common pattern in all "node conversion" methods, we store "null" to mark "pre-visited but not post-visited".

PR #6844, opened 2021-07-07
Clean up option usage in command executor

### LINKED ISSUE #2376: Check why --dump-instantiations and similar options are not accessible via input files.
`--dump-instantiations` via command-line works but `OPTION "dump-instantiations";`  via input file (e.g. CVC language) does not dump instantiations. Same issues with unsat cores etc.

nafur:
This PR makes the CommandExecutor class use the options object from its SmtEngine instead of the driver one. This makes sure that options that are set via `(set-option ...)` are in effect for the CommandExecutor. It still stores the driver options, though, as they are used for resets. 
The PR also does some minor cleanups along the way (remove unused `pOptions`, make things const).
Fixes #2376.


src/main/command_executor.cpp
```diff
@@ -66,23 +71,23 @@ CommandExecutor::~CommandExecutor()
 
 void CommandExecutor::printStatistics(std::ostream& out) const
 {
-  if (d_options.base.statistics)
+  if (options().base.statistics)
```
4tXJ7f, line 74: We could already switch these over to do `Solver::getOption()`.

src/main/command_executor.h
```diff
@@ -52,11 +52,20 @@ class CommandExecutor
...
+  /**
+   * A pointer to the original options from the driver. Contain options as
+   * parsed from the command line. Used when the solver is reset.
+   */
+  const Options* d_driverOptions;
+
+  /**
+   * Retrieve the current options from the smt engine.
+   */
+  const Options& options() const;
```
4tXJ7f, line 64: Methods go before data members of a class.

4tXJ7f (approved):
LGTM. Please fix the declaration order before merging.

PR #6847, opened 2021-07-07
Implement stop-only for new justification heuristic

ajreynol:
This also refactors decision engine so that we use inheritance instead of a dummy flag + members to determine which implementation to use.


4tXJ7f (approved):
LGTM

PR #6849, opened 2021-07-07
Rename operator pow2 to int.pow2.

aniemetz:
This name is the name of the symbol we parse. Internally, we still call
it POW2. This is a fix for the problem that `pow2` may clash (and
already does clash on SMT-LIB benchmarks) with user-defined symbols.


ajreynol (approved):

yoni206 (approved):
10x for noticing and fixing this

PR #6860, opened 2021-07-12
Add arithmetic preprocess rewrite equality module

ajreynol:
This is the first part of a refactoring which will ensure that the linear arithmetic solver does not violate integer type constraints in its model.

This PR moves the method https://github.com/ajreynol/CVC4/blob/master/src/theory/arith/theory_arith.cpp#L129 to its own module / file.


4tXJ7f (approved):
LGTM

PR #6861, opened 2021-07-12
Improvements to debug check model

ajreynol:
This makes it so that debug-check-models applies in production mode, not just in debug mode.  It also verifies that type constraints are met.


4tXJ7f (approved):
LGTM

PR #6862, opened 2021-07-12
Add trace for combination splits

ajreynol:
Followup PRs will unify these with the lemmas / stats infrastructure in theory inference manager.


4tXJ7f (approved):
LGTM

PR #6863, opened 2021-07-12
libpoly follows the debug/release of cvc5

Ouyancheng:
I would suggest using the debug build of libpoly in the debug build of cvc5. 
IMPORTANT NOTE: please make sure that ${CMAKE_BUILD_TYPE} referenced here is actually the CMAKE_BUILD_TYPE passed from ./configure.sh. I'm not 100% sure about this.

### DIFF
--- a/cmake/FindPoly.cmake
+++ b/cmake/FindPoly.cmake
@@ -79,7 +79,7 @@ if(NOT Poly_FOUND_SYSTEM)
       sed -i.orig
       "s,add_subdirectory(test/polyxx),add_subdirectory(test/polyxx EXCLUDE_FROM_ALL),g"
       <SOURCE_DIR>/CMakeLists.txt ${patchcmd}
-    CMAKE_ARGS -DCMAKE_BUILD_TYPE=Release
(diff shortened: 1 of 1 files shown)


Ouyancheng:
Oh ${CMAKE_BUILD_TYPE} is Debug/Production/Test/Competition, not Debug/Release! 
So this might not be a correct way of setting it. 

But as I experimented, it works (don't know why).

Ouyancheng:
This is incorrect. I'm getting a non-debugging build on CMAKE_BUILD_TYPE=Production only because it defaults not to include the debug symbol.

mpreiner:
Note, in general we don't want to compile debug builds for our dependencies if cvc5 is built in debug mode. Dependencies should be used in production mode independent from cvc5's build type.

Ouyancheng:
> Note, in general we don't want to compile debug builds for our dependencies if cvc5 is built in debug mode. Dependencies should be used in production mode independent from cvc5's build type.

OK. I don't mind manually switching it to debug mode if I encounter a problem.

PR #6864, opened 2021-07-12
Fix ANTLR build with CMake <3.11

4tXJ7f:
With CMake 3.11 and later, `<DOWNLOAD_DIR>` is substituted in
`ExternalProject_Add` but not in older versions [0]. To maintain
compatibility with older versions of CMake, this commit changes
`ExternalProject_Add` to use `<DOWNLOADED_FILE>` instead, which is both
nicer and substituted in older versions of CMake.

[0] https://cmake.org/cmake/help/latest/release/3.11.html#modules


4tXJ7f:
This should fix our nightly builds on the Ubuntu 18.04 containers, which use CMake 3.10.

mpreiner (approved):

PR #6865, opened 2021-07-12
Add branch and bound module

ajreynol:
This PR moves https://github.com/cvc5/cvc5/blob/master/src/theory/arith/theory_arith_private.cpp#L3451 to its own module.

The next PR will connect this module to theory_arith / theory_arith_private.

Towards ensuring type constraints are met for linear arithmetic models.

### DIFF
--- a/src/theory/arith/branch_and_bound.cpp
+++ b/src/theory/arith/branch_and_bound.cpp
@@ -1,0 +1,143 @@
+/******************************************************************************
+ * Top contributors (to current version):
+ *   Andrew Reynolds
+ *
+ * This file is part of the cvc5 project.
+ *
+ * Copyright (c) 2009-2021 by the authors listed in the file AUTHORS
+ * in the top-level source directory and their institutional affiliations.
+ * All rights reserved.  See the file COPYING in the top-level source
+ * directory for licensing information.
+ * ****************************************************************************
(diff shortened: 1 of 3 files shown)


src/theory/arith/branch_and_bound.cpp
```diff
@@ -0,0 +1,143 @@
...
+      lem =
+          d_pfGen->mkTrustNode(nm->mkNode(OR, ub, lb), PfRule::SPLIT, {}, {ub});
+    }
+    else
+    {
+      lem = TrustNode::mkTrustLemma(nm->mkNode(OR, ub, lb), nullptr);
+    }
+  }
+
+  Trace("integers") << "integers: branch & bound: " << lem << std::endl;
```
4tXJ7f, line 135: What about the additional debug output [here](https://github.com/cvc5/cvc5/blob/master/src/theory/arith/theory_arith_private.cpp#L3554-L3577)?

4tXJ7f (approved):
LGTM

src/theory/arith/branch_and_bound.cpp
ajreynol: This uses an internal method `getSatValue`; I'm not transferring this part of the debugging.

PR #6867, opened 2021-07-12
bv: Expand bitblast proof steps in the proof post processor.

mpreiner:
This commit changes BV proof logging to record coarse-grained bit-blast steps during solving and expanding these steps on-demand in the proof post processor.

PR #6869, opened 2021-07-13
bv: Do not rewrite below BV leafs in BBProof's TConvProofGenerator.

mpreiner:
Introduces a TermContext class that can be used to skip rewrites below theory leafs. Fixes issues related to bit-vector leafs begin incorrectly rewritten.

Note: This PR depends on #6867


ajreynol (approved):
LGTM, glad this worked out.

PR #6870, opened 2021-07-13
Fix models for sequences of infinite element type

ajreynol:
This fixes our model construction for sequences of infinite element type.

We were relying on `getModelValue` in our model construction which is incorrect since it assumes that the element type's theory can provide concrete values during model generation time. This makes the sequence model construction more robust by generalizing how model values are assigned: we use skeletons instead of concrete values when the element type is infinite. 

This fixes some open model generation issues with Facebook benchmarks.


Ying1123 (approved):

src/theory/strings/theory_strings.cpp
```diff
@@ -483,6 +499,38 @@ bool TheoryStrings::collectModelInfoType(
...
+              //   (seq.++ (seq.unit 0) (seq.unit 1) (seq.unit 2))
+              // becomes:
+              //   (seq.++ (seq.unit k_0) (seq.unit k_1) (seq.unit k_2))
+              // where k_0, k_1, k_2 are fresh integer variables. These
+              // variables will be assigned values in the standard way by the
+              // model. This construction is necessary since the strings solver
+              // must constrain the length of the model of an equivalence class
+              // (e.g. in this case to length 3); moreover we cannot assign a
+              // concrete value since it may conflict with other skeletons we
+              // have assigned, e.g. for the case of (seq.unit argVal) above.
```
4tXJ7f, line 517: Wouldn't it be better to directly do this in the `SeqEnumLen`?

4tXJ7f (approved):
LGTM

src/theory/strings/theory_strings.cpp
ajreynol: SeqEnumLen needs to return concrete values by default, as this is what is used in the default model construction and other places (e.g. exhaustive instantiation).  This variant for skeletons is specific to the sequences model construction here.

PR #6873, opened 2021-07-14
Fix for context on input proof generator

ajreynol:
The preprocess proof generator uses a dummy CDProof to mark input assertions that do not need justification. This CDProof should be user-context dependent to avoid rare cases where an input assertion was the symmetric equality of another, it was previously context independent.

This also refactors the debugging traces in this class.


HanielB (approved):
LGTM. Sounds like it was fun to debug this.

PR #6876, opened 2021-07-14
Arithmetic equality solver

ajreynol:
This is work towards the central equality engine. This adds a module of arithmetic that uses the equality engine in the default way.

This class will be incorporated into theory_arith.cpp.  It will be the replacement for CongruenceManager when we use the central equality engine architecture.


src/theory/arith/equality_solver.h
```diff
@@ -0,0 +1,115 @@
...
+   * Theory::needsEqualityEngine.
+   */
+  bool needsEqualityEngine(EeSetupInfo& esi);
+  /**
+   * Finish initialize
+   */
+  void finishInit();
+  //--------------------------------- end initialization
+  /**
+   * Pre-notify fact, return true if we are finished processed, false if
```
4tXJ7f, line 61:
```suggestion
   * Pre-notify fact, return true if we are finished processing, false if
```

src/theory/arith/equality_solver.cpp
```diff
@@ -0,0 +1,127 @@
...
+  return false;
+}
+
+TrustNode EqualitySolver::explain(TNode lit)
+{
+  Trace("arith-eq-solver-debug") << "explain " << lit << "?" << std::endl;
+  // check if we propagated it?
+  if (d_propLits.find(lit) == d_propLits.end())
+  {
+  Trace("arith-eq-solver-debug") << "...did not propagate" << std::endl;
```
4tXJ7f, line 73: ClangFormat

src/theory/arith/equality_solver.h
```diff
@@ -0,0 +1,115 @@
...
+      TNode atom, bool pol, TNode fact, bool isPrereg, bool isInternal);
+  /**
+   * Return an explanation for the literal lit (which was previously propagated
+   * by this solver).
+   */
+  TrustNode explain(TNode lit);
+
+ private:
+  /** Notification class from the equality engine */
+  class EqualitySolverNotify : public eq::EqualityEngineNotify
```
4tXJ7f, line 74: I know that this is kind of a common pattern that we use but why not simply have `EqualitySolver` inherit from `eq::EqualityEngineNotify` and get rid of this class?

(This is arguably a more general design decision and probably outside of the scope of this PR)

4tXJ7f (approved):
LGTM, just a typo and a formatting issue.

src/theory/arith/equality_solver.h
ajreynol: I tend not to like doing this, because it is misleading to what the purpose of the class is.  Although that is one of its responsibilities, it's main purpose is to setup / manage the equality engine.  One can also get into annoying cases where you are tempted to use multiple inheritence, say e.g. if EqualitySolver wanted to inherit from some kind of Subsolver class.

Nevertheless it's an interesting thought.

PR #6878, opened 2021-07-14
Add node converter utility

ajreynol:
Adds a utility for converting nodes.  This is the first step towards the LFSC proof conversion.


src/expr/node_converter.h
```diff
@@ -0,0 +1,97 @@
...
+namespace cvc5 {
+
+/**
+ * A node converter for terms and types. Implements term/type traversals,
+ * calling the provided implementations of conversion methods (pre/postConvert
+ * and pre/postConvertType) at pre-traversal and post-traversal.
+ *
+ * This class can be used as a generic method for converting terms/types.
+ */
+class NodeConverter
```
yoni206, line 36: This has similar goals and structure (yet still different) to https://github.com/cvc5/cvc5/blob/704127085e5ba2deb19e41337908a340e1b191dd/src/expr/node_traversal.h#L40

Will it make sense to use that class? Or perhaps at least to document here when is it better to use this class than that class, or the differences? I think it might be undesirable to have the same functionality implemented twice without some explanations in the code.

src/expr/node_converter.h
```diff
@@ -0,0 +1,97 @@
...
+  void addToTypeCache(TypeNode cur, TypeNode ret);
+  /** Node cache for preConvert */
+  std::unordered_map<Node, Node> d_preCache;
+  /** Node cache for postConvert */
+  std::unordered_map<Node, Node> d_cache;
+  /** TypeNode cache for preConvert */
+  std::unordered_map<TypeNode, TypeNode> d_preTCache;
+  /** TypeNode cache for postConvert */
+  std::unordered_map<TypeNode, TypeNode> d_tcache;
+  /** Whether this node converter is idempotent. */
```
yoni206, line 91: Can you expand on what this means (either here or in the constructor declaration)?

src/expr/node_converter.h
```diff
@@ -0,0 +1,97 @@
...
+  TypeNode convertType(TypeNode tn);
+
+ protected:
+  //------------------------- virtual interface
+  /** Should we traverse n? */
+  virtual bool shouldTraverse(Node n);
+  /** Run the conversion for n during pre-order traversal. */
+  virtual Node preConvert(Node n);
+  /**
+   * Run the conversion for at post-order traversal, where notice n is a term
```
yoni206, line 60:
```suggestion
   * Run the conversion for post-order traversal, where notice n is a term
```

src/expr/node_converter.h
```diff
@@ -0,0 +1,97 @@
...
+   * of the form:
+   *   (f i_1 ... i_m)
+   * where i_1, ..., i_m are terms that have been returned by previous calls
+   * to postConvert.
+   */
+  virtual Node postConvert(Node n);
+  /**
+   * Run the conversion, same as above, but for type nodes, which notice can
```
yoni206, line 68:
```suggestion
   * Run the conversion, same as `preConvert`, but for type nodes, which notice can
```

src/expr/node_converter.h
```diff
@@ -0,0 +1,97 @@
...
+   * be built from children similar to Node.
+   */
+  virtual TypeNode preConvertType(TypeNode n);
+  /**
+   * Run the conversion, same as above, but for type nodes, which notice can
```
yoni206, line 73:
```suggestion
   * Run the conversion, same as `postConvert`, but for type nodes, which notice can
```

src/expr/node_converter.h
```diff
@@ -0,0 +1,97 @@
...
+ *
+ * This class can be used as a generic method for converting terms/types.
+ */
+class NodeConverter
+{
+ public:
+  NodeConverter(bool forceIdem = true);
+  virtual ~NodeConverter() {}
+  /**
+   * This converts node n based on the runConvert method that can be overriden
```
yoni206, line 42: I don't see this method. Isn't it actually the two pre and post methods?

src/expr/node_converter.h
```diff
@@ -0,0 +1,97 @@
...
+   * by instances of this class.
+   */
+  Node convert(Node n);
+
+  /**
+   * This converts type node n based on the runConvertType method that can be
```
yoni206, line 48: Ditto.

src/expr/node_converter.cpp
```diff
@@ -0,0 +1,254 @@
...
+namespace cvc5 {
+
+NodeConverter::NodeConverter(bool forceIdem) : d_forceIdem(forceIdem) {}
+
+Node NodeConverter::convert(Node n)
+{
+  if (n.isNull())
+  {
+    return n;
+  }
```
yoni206, line 31: While standard in other places in the code, I suggest documenting this case in the `h` files, so that implementors of overriden conversion functions know that they don't have to worry about this case.

src/expr/node_converter.cpp
```diff
@@ -0,0 +1,254 @@
...
+    cur = visit.back();
+    visit.pop_back();
+    it = d_cache.find(cur);
+    Trace("nconv-debug2") << "convert " << cur << std::endl;
+    if (it == d_cache.end())
+    {
+      d_cache[cur] = Node::null();
+      Assert(d_preCache.find(cur) == d_preCache.end());
+      Node curp = preConvert(cur);
+      // if curp = cur, set null to avoid infinite loop
```
yoni206, line 49: I don't understand this comment - couldn't there be a legitimate conversion that would not change some nodes? I understand why you don't want to push `curp` to `visit` in such a case, but I don't see why the entry in the cache should point to null.

yoni206 (commented):
This looks cool! Some comments inline.

src/expr/node_converter.h
ajreynol: That is for node traversals, this is for node conversions.  The two are for different purposes, I'll mention this.

src/expr/node_converter.cpp
ajreynol: Actually I'll drop this.

src/expr/node_converter.cpp
ajreynol: I'll comment, preCache should be null when you don't convert, otherwise it should point to the node that you should follow to get the pre-rewritten form.

yoni206 (approved):
Thanks for the changes and answers!

PR #6880, opened 2021-07-14
Add print expression utility

ajreynol:
This will be used in the LFSC printer.  It may be of general use to other proof printers.


src/proof/print_expr.h
```diff
@@ -0,0 +1,84 @@
...
+ ** All rights reserved.  See the file COPYING in the top-level source
+ ** directory for licensing information.\endverbatim
+ **
+ ** \brief The module for printing Lfsc proof nodes
+ **/
+
+#include "cvc5_private.h"
+
+#ifndef CVC4__PROOF__PRINT_EXPR_H
+#define CVC4__PROOF__PRINT_EXPR_H
```
HanielB, line 18: CVC5

src/proof/print_expr.h
```diff
@@ -0,0 +1,84 @@
+/*********************                                                        */
+/*! \file print_expr.h
+ ** \verbatim
+ ** Top contributors (to current version):
+ **   Andrew Reynolds
+ ** This file is part of the CVC4 project.
+ ** Copyright (c) 2009-2020 by the authors listed in the file AUTHORS
```
HanielB, line 7: 2021

src/proof/print_expr.h
```diff
@@ -0,0 +1,84 @@
...
+  PExpr(Node n) : d_node(n), d_pnode(nullptr), d_typeNode() {}
+  PExpr(const ProofNode* pn) : d_node(), d_pnode(pn), d_typeNode() {}
+  PExpr(TypeNode tn) : d_node(), d_pnode(nullptr), d_typeNode(tn) {}
+  ~PExpr() {}
+  /** The node, if this p-exression is a node */
+  Node d_node;
+  /** The proof node, if this p-expression is a proof */
+  const ProofNode* d_pnode;
+  /** The type node, if this p-expression is a type */
+  TypeNode d_typeNode;
```
HanielB, line 46: maybe use a union?

src/proof/print_expr.h
```diff
@@ -0,0 +1,84 @@
...
+};
+
+/**
+ * A stream of p-expressions
```
HanielB, line 50: Could say more. It's not intuitive why the stream is a vector.

HanielB (changes requested):

src/proof/print_expr.h
ajreynol: This would require an identifier indicating Node|ProofNode|TypeNode. This isn't performance intensive code (only used in printers), so I'd prefer keeping it simpler.

HanielB: Sounds good.

HanielB (approved):

PR #6883, opened 2021-07-14
[proof] Generalize congruence handling for HO in eq proofs

HanielB:
Previously we were not considering proofs for HO equalities (i.e., between operators) that were transitivity chains. This commit generalizes the elaboration procedure in `eq_proof` to do so.


ajreynol (approved):
LGTM, nice find.

PR #6885, opened 2021-07-14
Simplify computation of relevant terms in datatypes

ajreynol:
This makes it so that we do not consider non-singleton equivalence classes of DT for non-datatype terms to be relevant.  This impacts how model construction is done for datatypes.

It's not clear why this method considered such terms to be relevant, I believe this was done to mask a previous bug; regressions now pass when this method is simplified. 

Doing this is definitely wrong when using central equality engine, hence the motivation for this simplification.


yoni206:
@barrettcw I was wondering if you could look at this small change. What worries me is that this at least looks like using only shared terms (Nelson Openn) instead of having the datatype theory responsible for terms that are not shared (Polite).

I am not sure this part of the code is really responsible for this aspect of combination, but it at least looks suspicious. Do you know/remember by any chance why was this function here to begin with?

yoni206 (approved):

PR #6886, opened 2021-07-14
bv: Disable bv-assert-input if proofs are enabled.

mpreiner:


src/smt/set_defaults.cpp
```diff
@@ -404,6 +404,13 @@ void setDefaults(LogicInfo& logic, bool isInternalSubsolver)
     opts.smt.produceAssertions = true;
   }
 
+  if (options::bvAssertInput() && options::produceProofs())
+  {
+    Notice() << "Disabling bv-assert-input it is incompatible with proofs."
```
HanielB, line 409:
```suggestion
    Notice() << "Disabling bv-assert-input since it is incompatible with proofs."
```

ajreynol (approved):

PR #6890, opened 2021-07-14
Fix context for proofs of instantiations

ajreynol:
Caught by a regression on proof-new, where an instantiation was the symmetric equality of an instantiation on a previous call to check-sat. Proofs of instantiation should be user-context dependent.


HanielB (approved):
LGTM

PR #6894, opened 2021-07-15
Connect the equality solver to theory arith

ajreynol:
`--arith-eq-solver` is a new option to enable the equality solver in arithmetic, disabled by default.

This PR connects the equality solver to TheoryArith when this option is enabled. 

This is in preparation for the central equality engine.

### DIFF
--- a/src/theory/arith/theory_arith.cpp
+++ b/src/theory/arith/theory_arith.cpp
@@ -20,6 +20,7 @@
 #include "proof/proof_rule.h"
 #include "smt/smt_statistics_registry.h"
 #include "theory/arith/arith_rewriter.h"
+#include "theory/arith/equality_solver.h"
 #include "theory/arith/infer_bounds.h"
 #include "theory/arith/nl/nonlinear_extension.h"
 #include "theory/arith/theory_arith_private.h"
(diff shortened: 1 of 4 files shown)


src/theory/arith/theory_arith.cpp
```diff
@@ -73,6 +80,10 @@ ProofRuleChecker* TheoryArith::getProofChecker()
 
 bool TheoryArith::needsEqualityEngine(EeSetupInfo& esi)
 {
+  if (d_eqSolver != nullptr)
+  {
+    return d_eqSolver->needsEqualityEngine(esi);
```
nafur, line 85: It looks a bit weird, that `d_internal` isn't even considered when `d_eqSolver` is used. It is correct, though, as `d_eqSolver` always returns `true`.
I'd still suggest to either formulate it like
`if (d_eqSolver != nullptr && d_eqSolver->needsEqualityEngine(esi)) return true` or just say `if (d_eqSolver != nullptr) return true;` to make this piece of code easier to understand.

nafur (commented):
Just one comment to make the code a bit easier to understand.

src/theory/arith/theory_arith.cpp
ajreynol: The purpose of this method is to initialize esi; I'll comment on why the control flow looks like this.

nafur (approved):
LGTM.

PR #6896, opened 2021-07-15
Add n-ary term utilities

ajreynol:
Initial work towards rewrite rule reconstruction

### DIFF
--- a/src/expr/nary_term_util.cpp
+++ b/src/expr/nary_term_util.cpp
@@ -1,0 +1,249 @@
+/******************************************************************************
+ * Top contributors (to current version):
+ *   Andrew Reynolds
+ *
+ * This file is part of the cvc5 project.
+ *
+ * Copyright (c) 2009-2021 by the authors listed in the file AUTHORS
+ * in the top-level source directory and their institutional affiliations.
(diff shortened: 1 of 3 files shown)


src/expr/nary_term_util.cpp
```diff
@@ -0,0 +1,249 @@
...
+    visit.pop_back();
+    if (it->second.isNull())
+    {
+      Node ret = cur;
+      bool childChanged = false;
+      std::vector<Node> children;
+      for (const Node& cn : cur)
+      {
+        // if it is a list variable, insert the corresponding list as children;
+        itv = std::find(vars.begin(), vars.end(), cn);
```
4tXJ7f, line 193: To get better performance, we could require that `vars` are `BOUND_VARIABLE`s and then only do this check if `cn` is a bound variable.

4tXJ7f (approved):
LGTM, just a minor suggestion.

src/expr/nary_term_util.cpp
ajreynol: This is true, although at the moment, I don't want to have this restriction (in risk of forgetting it).

PR #6899, opened 2021-07-16
Do not exhaustive instantiation when FMF is disabled

ajreynol:
This makes FMF not handle quantified formulas when FMF options are disabled, apart from those marked internal.

This is required for combinations of sequences + quantified formulas.


4tXJ7f (approved):
LGTM

PR #6901, opened 2021-07-16
Changes to datatypes in preparation for central equality engine

ajreynol:
This adds changes from the centralEeDev branch for datatypes.

The changes in behavior are that the inference manager clears its pending inferences when the state is in conflict, and preCheck processes pending facts immediately (which may have been enqueued prior to the call to check via ee merges).  These are required to make datatypes robust to the order in which check / merge / backtracks can occur.


yoni206 (approved):

yoni206:
Marked with do-not-merge to let @nafur also look.

yoni206 (approved):

nafur (approved):
LGTM.

PR #6902, opened 2021-07-16
Refactor equality engine setup for arithmetic congruence manager

ajreynol:
Allows congruence manager to use its own (non-official) equality engine if both it and the arithmetic equality solver are enabled.


nafur (approved):
LGTM.

PR #6903, opened 2021-07-16
[proofs] Set BV solver to better proof-producing one when proofs on

HanielB:
Since the internal bitblaster can be way slower, the regressions that would have slow runs when `--check-proofs` is passed now have the command line that forces the use of the default bitblaster.


HanielB:
I did a not-very-rigorous test of how long `make regress -j8` would take in my laptop (with 30s timeout per regression) and this branch was ~5% slower (141s vs 134s).

ajreynol:
@mpreiner I believe is supposed to enable the equality engine when the internal BV solver is used.  I think we should do this first, this may speed up some of the regressions.

mpreiner:
Marking do-no-merge until equality engine refactor for bitblast-internal is done.

src/smt/set_defaults.cpp
```diff
@@ -411,6 +411,19 @@ void setDefaults(LogicInfo& logic, bool isInternalSubsolver)
     opts.bv.bvAssertInput = false;
   }
 
+  // If proofs are required and neither the BV solver has not been set by the
+  // user, or the internal BV solver has not been set by the user to any other
+  // than Minisat, set the BV solver to the internal bitblaster
```
mpreiner, line 416:
```suggestion
  // If proofs are required and the user did not specify a specific BV solver,
  // we make sure to use the proof producing BITBLAST_INTERNAL solver.
```

mpreiner (approved):

PR #6919, opened 2021-07-22
Fix CoCoA build for newer compilers

4tXJ7f:
Newer compilers, such as GCC 11, default to C++17. CoCoA does not
compile with C++17 and its check for the C++ version used by the
compiler does not take into account newer C++ versions. As a result,
building CoCoA using `--auto-download` fails with GCC 11. This commit
adds a patch that makes the test in CoCoA take into account compilers
that default to C++17 or newer.


nafur (approved):
LGTM.

PR #6921, opened 2021-07-22
Configuration: Indicate dependencies being built

4tXJ7f:
This commit changes the output of our configuration to show which
dependencies are being built as part of the cvc5 build. For example:

```
ABC                       : off
CryptoMiniSat             : off
GLPK                      : off
Kissat                    : off
LibPoly                   : on (local)
CoCoALib                  : off

Build libcvc5 only        : off
MP library                : gmp (system)
Editline                  : off
```

Indicates that `LibPoly` was not found in the system and is thus built
as part of the cvc5 build and GMP was found in the system.


cmake/Helpers.cmake
```diff
@@ -131,17 +131,30 @@ macro(print_info msg)
...
     set(OPT_VAL_STR "on")
   elseif("${var}" STREQUAL "OFF" OR "${var}" STREQUAL "IGNORE")
     set(OPT_VAL_STR "off")
   else()
     set(OPT_VAL_STR "${var}")
   endif()
-  message("${Blue}${str}: ${Green}${OPT_VAL_STR}${ResetColor}")
+
+  if("${ARGS_FOUND_SYSTEM}" STREQUAL "FALSE")
+    set(OPT_FOUND_SYSTEM_STR " (building)")
```
aniemetz, line 152: Would it make sense to maybe switch this and print `(system)` when it's found in the system? I guess it depends what we consider the default case.

nafur: I'd suggest to do both: either `(system)` or `(building)` (or maybe `(local)`?)

4tXJ7f: Yeah, I was thinking about having both, too. I don't have a strong preference either way but maybe specifying both is a good idea because it allows us to distinguish between that information being present or not.

4tXJ7f:
I have updated the output to either print `(local)` or `(system)` (see above for an example).

nafur (approved):
LGTM.

PR #6922, opened 2021-07-22
Print link to docs preview

4tXJ7f:
This commit adds a message to the `docs` target with a link to a preview
of the documentation, e.g.:

```
Preview docs in browser: file://localhost/home/noetzli/repos/cvc5_win/build-docs/docs/sphinx/index.html
```


aniemetz:
Not that it hurts, but why? It's always in the same place.

nafur:
It's not a crucial feature, but I think it can be quite helpful.

4tXJ7f:
Yeah, I find it useful because when I work on the documentation and don't already have it open in my browser, I can just click the link to open it. Alternatively, I could define a separate target `make docs-preview` that opens the URL directly. Would that be better?

aniemetz:
No, it's fine.

aniemetz (approved):

PR #6926, opened 2021-07-26
Fix eq proof conversion for constant merged parameterized ops

ajreynol:
This issue arises when using the central equality engine on several regressions.


HanielB (approved):
LGTM

PR #6927, opened 2021-07-26
More updates to arithmetic in preparation for central equality engine

ajreynol:
Makes arithEqSolver more robust to propagations from multiple sources, changes the default relationship to congruence manager based on preliminary results on SMT-LIB.


nafur (approved):
LGTM.
Does this already affect solver behaviour in the default configuration?

ajreynol:
No, we don't use arith-eq-solver by default.  We will use it by default when the central equality engine is enabled (to be determined).

PR #6928, opened 2021-07-26
Minor fix for term database + central equality engine

ajreynol:
Previously we provided a bogus (self) explanation in a rare case of setting up the term database for higher order.  This is incompatible with cases where central equality engine = master equality engine.


src/theory/quantifiers/term_database.cpp
```diff
@@ -1034,7 +1034,9 @@ bool TermDb::reset( Theory::Effort effort ){
           eq = itpe->second;
         }
         Trace("quant-ho") << "- assert purify equality : " << eq << std::endl;
-        ee->assertEquality(eq, true, eq);
+        // Note that ee may be the central equality engine, in which case this
+        // equality is explained trivially with "true".
```
HanielB, line 1038: Add that the explanation does not matter since these are internal terms, which is why doing this is fine (IIRC).

HanielB (changes requested):

src/theory/quantifiers/term_database.cpp
ajreynol: Actually, for central equality engine, asserting this equality may trigger a real conflict. "true" is right explanation, since the equality is essentially like saying `purifySkolemFor(t) = t`.

HanielB (approved):

PR #6929, opened 2021-07-26
Minor updates to shared terms database for central equality engine

ajreynol:
Includes proper set up of the proof equality engine it uses (which may be shared over multiple theories). Also makes its assertEquality method robust to non-equality atoms.


4tXJ7f (approved):
LGTM

PR #6930, opened 2021-07-26
Make driver use options from the solver

nafur:
This PR removes explicit ownership of the options object from the main function and replaces it by an api::Solver object. To achieve this, it does a number of minor changes:
- `api::Solver` not takes a `unique_ptr<Options>` in its constructor
- `CommandExecutor` only holds a reference to (a unique ptr of) the `api::Solver`
- the main functions accesses options via the solver


mpreiner (approved):
LGTM

mpreiner (approved):

PR #6931, opened 2021-07-26
Enable default equality proofs for sets

ajreynol:
This enables default support for equality proofs in the theory of sets.

This is in preparation for proofs in the central equality engine, which requires that *all* active theories that use the central equality engine at least support the default interaction with the equality engine.


nafur (approved):
LGTM.

PR #6932, opened 2021-07-26
Default equality proofs for bags and separation logic

ajreynol:
This is the last 2 remaining theories not to use the default handling of equality proofs.

In preparation for proofs in central equality engine.


nafur (approved):
LGTM.

PR #6934, opened 2021-07-27
Refactor managed streams

nafur:
This PR introduces a new `ManagedStream` class that replaces the previous `ManagedOstream`. It allows to directly store the (wrapped) stream objects in the options. Handling the stream options is moved from the options manager to option predicates and the different options for input and output streams are combined into a single one.


src/options/managed_streams.h
```diff
@@ -0,0 +1,138 @@
...
+#define CVC5__OPTIONS__MANAGED_STREAMS_H
+
+#include <memory>
+#include <ostream>
+
+#include "options/options_public.h"
+
+namespace cvc5 {
+
+namespace detail {
```
mpreiner, line 31: Why `detail`? Can this be moved to the cpp file in an anonymous namespace?

src/options/managed_streams.h
```diff
@@ -0,0 +1,138 @@
...
+  }
+
+  Stream& operator*() const { return *getPtr(); }
+  Stream* operator->() const { return getPtr(); }
+  operator Stream&() const { return *getPtr(); }
+  operator Stream*() const { return getPtr(); }
+
+ protected:
+  std::shared_ptr<Stream> d_stream;
+};
```
mpreiner, line 98: Move `public:` to the top, then `protected:` and `private:` last.

src/options/managed_streams.h
```diff
@@ -0,0 +1,138 @@
...
+    if (specialCases(value)) return;
+    if constexpr (std::is_same<Stream, std::ostream>::value)
+    {
+      d_stream.reset(detail::openOStream(value));
+    }
+    else if constexpr (std::is_same<Stream, std::istream>::value)
+    {
+      d_stream.reset(detail::openIStream(value));
+    }
+  }
```
mpreiner, line 89: Move implementation to cpp file.

src/options/options_handler.h
```diff
@@ -133,6 +134,18 @@ class OptionsHandler {
...
+                 const ManagedOut& mo);
+  void applyErr(const std::string& option,
+                const std::string& flag,
+                const ManagedErr& me);
+  void applyIn(const std::string& option,
+               const std::string& flag,
+               const ManagedIn& mi);
+  void applyOut(const std::string& option,
+                const std::string& flag,
+                const ManagedOut& mo);
```
mpreiner, line 148: Wouldn't `enableXX` or `setXXStream` be better names?

src/options/managed_streams.h
nafur: They are used in `ManagedStream::open()` which uses `if constexpr` with the condition using the template argument of the class. It thus needs to be in the header, and `openOStream` be declared (though not defined) in the header as well.

src/options/managed_streams.h
nafur: See comment above.

mpreiner (approved):

PR #6937, opened 2021-07-27
Minor changes from proof-new

ajreynol:
Note the change to the unit test makes it so the test is not dependent on Node ID order.


4tXJ7f (approved):
LGTM

PR #6941, opened 2021-07-27
Move enum value generator to own file

ajreynol:
Work towards integrating the enumerator callback into the fast enumerator.


HanielB (approved):
LGTM

PR #6942, opened 2021-07-27
Refactor enumerator management in synth conjecture

ajreynol:
This moves the method `SynthConjecture::getEnumeratedValue` to its own class, EnumManager.

It also integrates the sygus enumerator callback into the fast enumerator.


src/theory/quantifiers/sygus/sygus_enumerator.cpp
```diff
@@ -51,6 +51,12 @@ void SygusEnumerator::initialize(Node e)
 {
   Trace("sygus-enum") << "SygusEnumerator::initialize " << e << std::endl;
   d_enum = e;
+  // allocate the default callback
+  if (d_sec == nullptr && options::sygusSymBreakDynamic())
+  {
+    d_secd.reset(new SygusEnumeratorCallbackDefault(e, d_stats));
+    d_sec = d_secd.get();
+  }
```
HanielB, line 59: Since this is the only place where `d_secd` is used couldn't you remove this class member and have this code directly make `d_sec` initialized with the default callback?

HanielB (changes requested):
This is very nice. :) A few comments:

src/theory/quantifiers/sygus/sygus_enumerator.cpp
ajreynol: There is a memory management issue: d_sec is a raw pointer, since it may be user-provided in the constructor. 

Moreover, d_secd cannot be initialized until this call is made, since we don't have the enumerator e in the constructor.

ajreynol:
FYI im going to change the name of "EnumManager" to "EnumValueManager"; the name of this class is not descriptive enough on second glance.

src/theory/quantifiers/sygus/sygus_enumerator.cpp
HanielB: I see. I also got confused that this was the constructor. :)

HanielB (approved):

PR #6944, opened 2021-07-27
Add LFSC node converter

ajreynol:
This utility converts from terms used in the internal calculus to terms that capture how they should be printed in LFSC.


src/proof/lfsc/lfsc_node_converter.h
```diff
@@ -0,0 +1,113 @@
...
+{
+ public:
+  LfscNodeConverter();
+  ~LfscNodeConverter() {}
+  /** convert to internal */
+  Node postConvert(Node n) override;
+  /** convert to internal */
+  TypeNode postConvertType(TypeNode tn) override;
+  /**
+   * Get the null terminator for kind k and type tn. The type tn can be
```
yoni206, line 40: Maybe clarify with an example of null terminators?

src/proof/lfsc/lfsc_node_converter.h
```diff
@@ -0,0 +1,113 @@
...
+  /**
+   * Return the properly named operator for n of the form (f t1 ... tn), where
+   * f could be interpreted or uninterpreted.  This method is used for cases
+   * where it is important to get the term corresponding to the operator for
+   * a term. An example is for the base REFL step of nested CONG.
+   */
+  Node getOperatorOfTerm(Node n, bool macroApply = false);
+  /** get closure operator */
+  Node getOperatorOfClosure(Node q, bool macroApply = false);
+  /** get closure operator, cop is return  */
```
yoni206, line 55: I am not sure what does ‘cop is return’ mean

src/proof/lfsc/lfsc_node_converter.h
```diff
@@ -0,0 +1,113 @@
...
+  Node getOperatorOfBoundVar(Node cop, Node v);
+  /** get or assign variable index for variable v */
```
yoni206, line 57: Maybe clarify when will it get and when it will assign?

src/proof/lfsc/lfsc_node_converter.h
```diff
@@ -0,0 +1,113 @@
...
+  Kind getBuiltinKindForInternalSymbol(Node op) const;
+
+  /** get name for user name */
+  static std::string getNameForUserName(const std::string& name);
+
+ private:
+  /** Should we traverse n? */
+  bool shouldTraverse(Node n) override;
+  /** Make skolem function */
+  Node maybeMkSkolemFun(Node k, bool macroApply = false);
```
yoni206, line 77: Can you clarify the maybe part?

src/proof/lfsc/lfsc_node_converter.cpp
```diff
@@ -0,0 +1,1013 @@
...
+{
+  NodeManager* nm = NodeManager::currentNM();
+  SkolemManager* sm = nm->getSkolemManager();
+  SkolemFunId sfi = SkolemFunId::NONE;
+  Node cacheVal;
+  TypeNode tn = k.getType();
+  if (sm->isSkolemFunction(k, sfi, cacheVal))
+  {
+    if (sfi == SkolemFunId::SHARED_SELECTOR)
+    {
```
yoni206, line 620: Can u add brief doc on the different cases?

src/proof/lfsc/lfsc_node_converter.cpp
```diff
@@ -0,0 +1,1013 @@
...
+  TypeNode tnc = nm->mkFunctionType(nm->integerType(), c.getType());
+  Node aconstf = getSymbolInternal(CONST_STRING, tnc, "char");
+  for (unsigned i = 0, size = vec.size(); i < size; i++)
+  {
+    Node cc = nm->mkNode(APPLY_UF, aconstf, nm->mkConst(Rational(vec[i])));
+    chars.push_back(cc);
+  }
+}
+
+bool LfscNodeConverter::isIndexedOperatorKind(Kind k)
```
yoni206, line 706: Maybe this can reside in a more general location? Looks like a generally useful utility.

yoni206 (commented):
Few specific comments inline. Also: is there a unit test you could add? Even simple checks? e.g. that an AlwaysAssert is not reached or that a returned node is as expected (not necessarily for each case but perhaps for a few)?

ajreynol:
@yoni206 I would prefer delaying adding tests until the LFSC printer is added to master; the LFSC node converter is a helper class that should not be used standalone.

src/proof/lfsc/lfsc_node_converter.h
ajreynol: The documentation was out of date, I've fixed.

src/proof/lfsc/lfsc_node_converter.h
ajreynol: Done

src/proof/lfsc/lfsc_node_converter.cpp
ajreynol: Thinking about this some more, I don't think we want this as a general utility.

I'll clarify that this method is "is k a kind that is printed as an indexed operator in the LFSC format".

yoni206 (approved):

PR #6945, opened 2021-07-27
Allow changing certain options while solving

nafur:
This PR allows changing some select options ever after the smt engine has been fully initialized, following the SMT-LIB standard (section 4.1.7).


ajreynol:
This looks fine to me, although I'll let another reviewer confirm.

4tXJ7f (changes requested):
It would be good to add a regression test for this. I'd also suggest to use a set of strings instead of individual conditions.

4tXJ7f (approved):
Thanks for the updates, LGTM.

PR #6947, opened 2021-07-28
Fixes for building python wheels on manylinux2014

makaimann:
This PR makes several fixes to the wheel building infrastructure for manylinux2014.
* don't statically link to Python (see: [this section](https://www.python.org/dev/peps/pep-0513/#libpythonx-y-so-1) of PEP-513)
* use standard 'build' directory. I noticed that the wheel infrastructure uses directory 'build' to prepare the wheel anyway, so it doesn't make a lot of sense for us to use a separate working directory. I tried changing the name of the build directory in setuptools by setting member variables in `initialize_options` but that didn't seem to work and I decided it wasn't worth the effort. The wheel should be built in a clean environment anyway, i.e., docker.
* find the correct include directory and python library paths for building on manylinux2014 using `skbuild`. manylinux2014 has several versions of python installed in `/opt/python`. The CMake infrastructure can't find the necessary information (plus the libraries are deleted from the docker image to save space -- the library is just set to a generic `libpython<version>.a`). Instead, this PR would manually set them via CMake options.
* to allow setting the python information, I extended the `configure.sh` script to allow directly setting CMake options. There are definitely other options here if you'd rather not change `configure.sh`. My reasoning was that it could be useful in other situations and I just marked it as an advanced feature. Another approach would be to use `skbuild` directly in the CMake infrastructure. I didn't like that solution because it would make the CMakeLists.txt more complicated and it wouldn't follow the standard CMake/Cython/Python approach which means we won't benefit from future improvements to that CMake infrastructure. Plus, this issue should only come up on systems with non-standard installations of Python, such as manylinux. This approach (setting CMake options) is basically what `scikit-build` does automatically if we had used that infrastructure directly.


src/api/python/CMakeLists.txt
```diff
@@ -92,7 +92,7 @@ target_include_directories(pycvc5
   ${CMAKE_BINARY_DIR}/src         # for cvc5_export.h
 )
 
-target_link_libraries(pycvc5 cvc5 ${PYTHON_LIBRARIES})
+target_link_libraries(pycvc5 cvc5)
```
mpreiner, line 95: How pycvc5 linked against the Python library if this is removed here?

makaimann: Good question, I'm just referring to [this section](https://www.python.org/dev/peps/pep-0513/#libpythonx-y-so-1) of PEP-513 on this. It sounds like it is almost never necessary and can actively be bad in some cases. Essentially, it's not needed because once the extension module is loaded into a running interpreter, it automatically has access to all the interpreter's libpython symbols.

mpreiner (approved):

PR #6950, opened 2021-07-29
Add optional debug information for dumping instantiations

ajreynol:
This complete the implementation of `dump-instantiations-debug`.

With this option, we can print the source of instantiations. For example:

```
$ cvc5 regress1/quantifiers/dump-inst-proof.smt2 --dump-instantiations-debug --produce-proofs
unsat
(instantiations (forall ((x Int)) (or (P x) (Q x)))
  (! ( 2 ) :source QUANTIFIERS_INST_E_MATCHING_SIMPLE ((not (= (P x) true))))
)
(instantiations (forall ((x Int)) (or (not (S x)) (not (Q x))))
  (! ( 2 ) :source QUANTIFIERS_INST_E_MATCHING_SIMPLE ((not (= (S x) false))))
)
````

### CODE REVIEW

File: src/smt/smt_engine.cpp [hunk 1]
```diff
@@ -1635,11 +1635,36 @@ void SmtEngine::printInstantiations( std::ostream& out ) {
...
+      // convert to instantiation list
+      Node q = i.first;
+      InstantiationList& ilq = rinsts[q];
+      ilq.initialize(q);
+      for (const std::vector<Node>& ii : i.second)
+      {
+        ilq.d_inst.push_back(InstantiationVec(ii));
+      }
+    }
+  }
```
mpreiner, line 1664:
```suggestion
    for (const auto& [q, vec] : insts)
    {
      // convert to instantiation list
      InstantiationList& ilq = rinsts[q];
      ilq.initialize(q);
      ilq.d_inst.insert(ilq.d_inst.end(), vec.begin(), vec.end());
    }
  }
```

ajreynol: I really don't like this syntax.

mpreiner: It was not correct anyways, I overlooked the `InstantiationVec` part in `push_back`.

mpreiner (approved):

PR #6951, opened 2021-07-29
Add IEEE-BV-to-FP to external-to-internal mapping in C++ API

alex-ozdemir:
Without this, you can't make terms with that kind.


martin-cs (approved):
Please be careful with the terminology here.  This is `to_fp` from an IEEE bit-vector, not FP -> IEEE BV.  The first one, as you note is vital.  The second one isn't in the standard.  If you could change the commit message / PR title I would be grateful.

alex-ozdemir:
Whoops. I know exactly what you mean, and totally messed up the terminology here. Thanks for catching it.

aniemetz (approved):

PR #6952, opened 2021-07-30
Perform statistics printing via the API

nafur:
This PR changes how the command executor prints the statistics by moving stuff around a bit, eventually using only proper API methods of `api::Solver`. This PR also removes the `smt_engine.h` include from `command_executor.cpp`.

### DIFF
--- a/src/api/cpp/cvc5.cpp
+++ b/src/api/cpp/cvc5.cpp
@@ -6643,6 +6643,7 @@ Result Solver::checkEntailed(const Term& term) const
   CVC5_API_SOLVER_CHECK_TERM(term);
   //////// all checks before this line
   cvc5::Result r = d_smtEngine->checkEntailed(*term.d_node);
+  d_smtEngine->printStatisticsDiff();
   return Result(r);
   ////////
   CVC5_API_TRY_CATCH_END;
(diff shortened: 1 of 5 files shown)


src/api/cpp/cvc5.cpp
```diff
@@ -6656,7 +6664,9 @@ Result Solver::checkEntailed(const std::vector<Term>& terms) const
          "(try --incremental)";
   CVC5_API_SOLVER_CHECK_TERMS(terms);
   //////// all checks before this line
-  return d_smtEngine->checkEntailed(Term::termVectorToNodes(terms));
+  cvc5::Result r = d_smtEngine->checkEntailed(Term::termVectorToNodes(terms));
+  d_smtEngine->printStatisticsDiff();
```
ajreynol, line 6668: Can we actually push this further into SmtEngine::checkEntailed / SmtEngine::checkSat?

ajreynol (changes requested):

src/api/cpp/cvc5.cpp
nafur: I've put it into `checkSatInternal`, so it only lives in a single place. (well, actually two when an exception is thrown...)

src/smt/smt_engine.cpp
```diff
@@ -1925,10 +1927,13 @@ void SmtEngine::printStatisticsSafe(int fd) const
   d_env->getStatisticsRegistry().printSafe(fd);
 }
 
-void SmtEngine::printStatisticsDiff(std::ostream& out) const
+void SmtEngine::printStatisticsDiff() const
 {
-  d_env->getStatisticsRegistry().printDiff(out);
-  d_env->getStatisticsRegistry().storeSnapshot();
+  if (d_env->getOptions().base.statisticsEveryQuery)
```
ajreynol, line 1932: Maybe move this to guard the call to this method, not guard within the method itself.

ajreynol (changes requested):
One minor request, otherwise looks good

ajreynol (approved):

PR #6954, opened 2021-07-30
Simplify smt2 printer

ajreynol:
The common case of printing function applications is to print the kind in smt2 format, this makes this the default case and removes spurious cases.  It also makes a few minor fixes.


4tXJ7f (approved):
Nice :)

PR #6955, opened 2021-07-30
Normalize val in BitVector(val_str, base) 

alex-ozdemir:
Fixes cpp API's mkBitVector(val_str, base) constructor.


src/api/cpp/cvc5.cpp
```diff
@@ -4862,6 +4864,8 @@ Term Solver::mkBVFromStrHelper(uint32_t size,
         << "Overflow in bitvector construction (specified bitvector size "
         << size << " too small to hold value " << s << ")";
   }
+  // Doesn't happen automatically.
+  val = val.modByPow2(size);
```
aniemetz, line 4868: This is done in the `BitVector` constructor, actually: https://github.com/cvc5/cvc5/blob/master/src/util/bitvector.h#L33

aniemetz (changes requested):
Can you also add a test to the solver_black tests?
https://github.com/cvc5/cvc5/blob/master/test/unit/api/solver_black.cpp#L307

src/api/cpp/cvc5.cpp
alex-ozdemir: Good point. Removed.

alex-ozdemir:
Ah---I didn't see that test file.

I've added tests for the current pattern of errors. You can amend them in your refactor, if we throw errors more liberally then.

aniemetz (approved):

PR #6958, opened 2021-08-02
Create grouping of tests that exercise '--use-approx'

andrewvaughanj:
Currently, it seems that `cvc5` has no tests that get anywhere near close to exercising the code in `approx_simplex`, which means that the GLPK integration is effectively untested in CI:

* https://cvc4.cs.stanford.edu/downloads/builds/coverage/cvc5-2021-08-02/index.src_theory_arith_approx_simplex.cpp.html

(which is at 2.8% at the time of writing!)

Indeed, if you run the whole of `ctest`, none of the code near GLPK is exercised; and, worse, there are *no* regression tests that use `--use-approx`!

This PR adds a selection of simple tests that have been derived (using `cvc5`'s existing regression suite and ddSMT) to reach code inside of `approx_simplex`; namely:

- `solveRelaxation`, and
- `solveMIP`

This is (hopefully) an initial start that will allow us to grow more regression tests exercising GLPK (e.g., following the same process of reduction, but using the SMTLIB benchmarks vs. only `cvc5`'s regression suite).

As a minor comment: I don't actually think that these tests need to be disabled if you don't have GLPK, but I created them *solely* to exercise GLPK, which is why they are segregated from the main regression tests in CMakeLists.txt.

The folder structure has been named such that I can do `ctest -R "use_approx"` to run only the tests that (should) reach GLPK.


andrewvaughanj:
Oh, it occurs to me that the GitHub Actions won't trigger these tests; it will probably only work when the nightlies are generated/run.

andrewvaughanj:
Oh, it looks like GLPK isn't even abled on CI (based on L547/L549 from https://cvc4.cs.stanford.edu/downloads/builds/coverage/cvc5-2021-08-02/index.src_theory_arith_approx_simplex.cpp.html) 😭

test/regress/CMakeLists.txt
```diff
@@ -2744,6 +2744,21 @@ set(regression_disabled_tests
...
+  regress0/use_approx/error0_approx.smt2
+  regress0/use_approx/issue2429_approx.smt2
+  regress0/use_approx/issue4714_approx.smt2
+  regress0/use_approx/siegel-nl-bases_approx.smt2
+  regress0/use_approx/specsharp-WindowsCard.15.RTE.Terminate_System.Int32_approx.smt2
+)
+
+if(USE_GLPK)
+  set(regress_0_tests ${regress_0_tests} ${use_appox_tests})
+endif()
```
4tXJ7f, line 2760: Instead of using CMake for this, please add a `; REQUIRES: glpk` at the top of the regression tests. That's the mechanism that we generally use to conditionally enable/disable tests depending on the features that cvc5 was compiled with.

4tXJ7f (changes requested):
Thank you for the tests! I just have a minor comment.

4tXJ7f:
> Oh, it looks like GLPK isn't even abled on CI (based on L547/L549 from https://cvc4.cs.stanford.edu/downloads/builds/coverage/cvc5-2021-08-02/index.src_theory_arith_approx_simplex.cpp.html) sob

Good point, we can definitely add a build with GLPK :) Would be good to have anyway because we use it for the competition.

test/regress/CMakeLists.txt
andrewvaughanj: Ah! Certainly! I didn't know about that!

However, these tests don't *need* GLPK, right? `cvc5` might use "do more" if GLPK is enabled, but I think it still does something if you have `--use-approx` and not a GLPK build.

Basically, I agree with you, I'm just making sure you still agree with me, after you're aware of the above.

andrewvaughanj:
> > Oh, it looks like GLPK isn't even abled on CI (based on L547/L549 from https://cvc4.cs.stanford.edu/downloads/builds/coverage/cvc5-2021-08-02/index.src_theory_arith_approx_simplex.cpp.html) sob
> 
> Good point, we can definitely add a build with GLPK :) Would be good to have anyway because we use it for the competition.

Am I able to do this? I thought the nightlies were configured in "some other way" (i.e., not GitHub Actions).

4tXJ7f:
> > > Oh, it looks like GLPK isn't even abled on CI (based on L547/L549 from https://cvc4.cs.stanford.edu/downloads/builds/coverage/cvc5-2021-08-02/index.src_theory_arith_approx_simplex.cpp.html) sob
> > 
> > 
> > Good point, we can definitely add a build with GLPK :) Would be good to have anyway because we use it for the competition.
> 
> Am I able to do this? I thought the nightlies were configured in "some other way" (i.e., not GitHub Actions).

Nope, we are not using GitHub actions for those, so we will take care of that internally :)

test/regress/CMakeLists.txt
4tXJ7f: It looks like is [always](https://github.com/cvc5/cvc5/blob/e4e19cd62b3eebed2de5b9b627509df0ffec23e1/src/theory/arith/theory_arith_private.cpp#L1884-L1885) [checked](https://github.com/cvc5/cvc5/blob/e4e19cd62b3eebed2de5b9b627509df0ffec23e1/src/theory/arith/theory_arith_private.cpp#L2969) in conjunction with [`ApproximateSimplex::enabled()`](https://github.com/cvc5/cvc5/blob/2283ee3b0327441c29caf26be977c1e4cd13c637/src/theory/arith/approx_simplex.cpp#L545-L551), so it won't have an effect if GLPK is disabled. Have you found a different location where this is not the case?

andrewvaughanj: Nope; not at all.

I guess that's my original point: `cvc5` does "something" if you use `--use-approx` **without** GLPK, but I am not entirely sure what that behaviour is ...

For example, if you *don't* have GLPK, is it possible we might get `unknown` where we get `sat`/`unsat` in the tests I added?

4tXJ7f (approved):
Sorry, forgot to approve this one after our last discussions. LGTM

PR #6959, opened 2021-08-02
Fix readme link

GUIpsp:


GUIpsp:
This is such a minor change that I didn't bother signing it. Hopefully that's ok.

4tXJ7f:
Thank you for the PR. This may seem unnecessary, but it would still be great if you could sign your commit despite this being a minor change. The issue is that we want to keep our rules simple and don't want to start deciding what is minor enough and what isn't. I hope that that makes sense.

4tXJ7f:
@GUIpsp, I somehow can't see the `Signed-off-by:` line in the commit message. Am I missing something? Thanks.

aniemetz:
Obsolete with #7222.

PR #6961, opened 2021-08-02
bv: Enable equality engine for bitblast-internal.

mpreiner:


src/theory/theory.cpp
```diff
@@ -620,11 +620,6 @@ bool Theory::usesCentralEqualityEngine(TheoryId id)
     // conditional on whether we are using the equality solver
     return options::arithEqSolver();
   }
-  if (id == THEORY_BV)
```
ajreynol, line 623: You can also remove the bv include from this file.

ajreynol (changes requested):
One minor request

ajreynol (approved):

PR #6963, opened 2021-08-02
Simplify how user-provided quantifier attributes are handled

ajreynol:
This makes INST_ATTRIBUTE (optionally) take multiple children, giving a way for the user to set attributes on quantified formulas, which does not require a new API command.

This is a special case of term annotations that occur as the body of a quantified formula.

If we need to extend our API to support further user-provided attributes, we should use a similar approach, e.g. a new kind ANNOTATE.


nafur (approved):
LGTM, but I'd really like @aniemetz have a look as well.

src/theory/quantifiers/quantifiers_attributes.cpp
```diff
@@ -36,7 +36,10 @@ bool QAttributes::isStandard() const
 
 QuantAttributes::QuantAttributes() {}
 
-void QuantAttributes::setUserAttribute( const std::string& attr, Node n, std::vector< Node >& node_values, std::string str_value ){
+void QuantAttributes::setUserAttribute(const std::string& attr,
+                                       Node n,
+                                       std::vector<Node>& node_values)
```
aniemetz, line 41: `node_values` should be const.

src/theory/quantifiers/quantifiers_attributes.cpp
```diff
@@ -200,7 +204,28 @@ void QuantAttributes::computeQuantAttributes( Node q, QAttributes& qa ){
...
+        // (INST_ATTRIBUTE v) where v has an internal attribute set on it.
+        // We distinguish these two cases by checking the kind of the first
+        // child.
+        if (q[2][i][0].getKind() == CONST_STRING)
+        {
+          // make a dummy variable to be used below
+          avar = nm->mkBoundVar(nm->booleanType());
+          std::vector<Node> nodeValues(q[2][i].begin() + 1, q[2][i].end());
+          // set user attribute on the dummy variable
+          setUserAttribute(
```
aniemetz, line 221: How deep may this recurse in the worst case?

aniemetz (changes requested):

src/theory/quantifiers/quantifiers_attributes.cpp
```diff
@@ -36,7 +36,10 @@ bool QAttributes::isStandard() const
 
 QuantAttributes::QuantAttributes() {}
 
-void QuantAttributes::setUserAttribute( const std::string& attr, Node n, std::vector< Node >& node_values, std::string str_value ){
+void QuantAttributes::setUserAttribute(const std::string& attr,
+                                       Node n,
```
aniemetz, line 40: Does this have to be `Node`? Why not a reference? Or a `TNode`?

src/theory/quantifiers/quantifiers_attributes.cpp
ajreynol: There is no recursion here, this is called from computeQuantAttributes.

aniemetz: Ah right, I missed that this was collapsed in. All good :)

aniemetz (approved):

PR #6964, opened 2021-08-02
Remove "inUnsatCore" flag throughout

ajreynol:
The internal solver no longer cares about what assertions are named / are in the unsat core.

This simplifies the code so that the (now unused) `inUnsatCore` flag is removed from all interfaces.

This eliminates another external use of `getSmtEngine`.


4tXJ7f (approved):
LGTM

PR #6965, opened 2021-08-02
Remove dependencies on smt engine in smt solver

ajreynol:
This is work towards eliminating circular dependencies on SmtEngine.

This simplifies several interfaces and makes it so that SmtSolver does not take a pointer to its parent SmtEngine.

It is also work towards eliminating the output manager, which is now subsumed by `Env`.


aniemetz (approved):
LGTM

PR #6974, opened 2021-08-04
[proof] Add getProof to API and use it in GetProofCommand

HanielB:
Also adds a call to get proof in a unit test.


ajreynol (approved):
Nice, LGTM

PR #6975, opened 2021-08-04
Proper printing of proofs in the internal calculus

ajreynol:
For example, before:
```
unsat
(proof
(SCOPE (EQ_RESOLVE (ASSUME |:args| ((not (= a a)))) (TRANS (CONG (THEORY_REWRITE |:args| ((= (= a a) true) 11 6)) |:args| (17)) (THEORY_REWRITE |:args| ((= (not true) false) 1 6)))) |:args| ((not (= a a))))
)
```
After:
```
unsat
(proof
(SCOPE (EQ_RESOLVE (ASSUME |:args| ((not (= a a)))) (TRANS (CONG (THEORY_REWRITE |:args| ((= (= a a) true) THEORY_STRINGS RW_REWRITE_THEORY_POST)) |:args| (NOT)) (THEORY_REWRITE |:args| ((= (not true) false) THEORY_BOOL RW_REWRITE_THEORY_POST)))) |:args| ((not (= a a))))
)
```


src/theory/inference_id.h
```diff
@@ -791,6 +791,9 @@ std::ostream& operator<<(std::ostream& out, InferenceId i);
 /** Make node from inference id */
 Node mkInferenceIdNode(InferenceId i);
 
+/** get a inference identifier from a node, return false if we fail */
```
HanielB, line 794:
```suggestion
/** get an inference identifier from a node, return false if we fail */
```

HanielB (approved):
LGTM, just a minor comment. Really nice to have this.

PR #6977, opened 2021-08-04
Add inference ids for remainder of sygus solver

ajreynol:
Now, all inferences throughout cvc5 in our regression are properly marked with an InferenceId.

This PR includes minor simplifications to the interface for sygus modules. In particular it changes the behavior to send inferences via the inference manager instead of passing them around as a vector.


nafur (approved):
LGTM.

PR #6980, opened 2021-08-04
Add API function to obtain information about a single option

nafur:
This PR adds `api::Solver::getOptionInfo()` that returns information about a single option, including its name, aliases, current and default value and its domain.


src/options/mkoptions.py
```diff
@@ -835,6 +835,33 @@ def codegen_all_modules(modules, build_dir, dst_dir, tpl_options_h,
...
+                        'maximum': option.maximum if option.maximum else '{}',
+                    }
+                    if option.type in ['bool', 'std::string']:
+                        constr = 'OptionInfo::ValueInfo<{type}>{{{default}, {value}}}'.format(**fmt)
+                    elif option.type == 'double' or is_numeric_cpp_type(option.type):
+                        constr = 'OptionInfo::NumberInfo<{type}>{{{default}, {value}, {minimum}, {maximum}}}'.format(**fmt)
+                    elif option.mode:
+                        values = ', '.join(map(lambda s: '"{}"'.format(s), option.mode.keys()))
+                        if not option.default:
+                            fmt['default'] = '???'
```
mpreiner, line 855: Not sure I understand this line :)

mpreiner (commented):
It'd be also good to add a unit test that retrieves some option infos.

src/options/mkoptions.py
nafur: This line sets the default values for mode options (i.e. enums) that have no explicit default specified.

mpreiner: When you set the default to `???`, when does it crash? Do we have mode options that do not define a default value? Do we even want to allow this?

nafur: It will not crash. This information is communicated via the `ModeInfo` struct that holds the default value as `std::string`. We have `outputTag` that does not define a default... maybe it's a better idea to enforce defaults for mode options.

mpreiner: Yeah `outputTag` is a bit special, but we could have a NONE tag for the default case and then enforce default values.

nafur: After some offline discussion, we decided on having `NONE` but hide it from the generated help.

src/api/cpp/cvc5.cpp
```diff
@@ -7026,6 +7026,122 @@ std::string Solver::getOption(const std::string& option) const
   CVC5_API_TRY_CATCH_END;
 }
 
+template<class... Ts> struct overloaded : Ts... { using Ts::operator()...; };
```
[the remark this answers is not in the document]
mpreiner: Ok, can you add the URL as documentation?

mpreiner (approved):

PR #6981, opened 2021-08-04
Fix policy for merging subproofs

ajreynol:
This makes it so that we replace subproof A with subproof B if A and B prove the same thing, and B contains no assumptions.  This is independent of the scope that A and B are in.

The previous policy replaced subproof A with subproof B if A and B prove the same thing and were in the same scope. This is not safe if A occurs in multiple scopes, where free assumptions can escape.

The new policy is more effective and safer than the previous one.

### DIFF
--- a/src/proof/proof_node_updater.cpp
+++ b/src/proof/proof_node_updater.cpp
@@ -77,6 +77,7 @@ void ProofNodeUpdater::processInternal(std::shared_ptr<ProofNode> pf,
 {
   // Note that processInternal uses a single scope; fa is updated based on
   // the current free assumptions of the proof nodes on the stack.
+
   // The list of proof nodes we are currently traversing beneath. This is used
   // for checking for cycles in the overall proof.
   std::vector<std::shared_ptr<ProofNode>> traversing;
@@ -85,6 +86,7 @@ void ProofNodeUpdater::processInternal(std::shared_ptr<ProofNode> pf,
   // Map from formulas to non-closed proof nodes that prove that fact. These
   // are replaced by proofs in the above map when applicable.
   std::map<Node, std::vector<std::shared_ptr<ProofNode>>> resCacheNcWaiting;
+  // Map from proof nodes to whether they contain assumptions
   std::unordered_map<const ProofNode*, bool> cfaMap;
   Trace("pf-process") << "ProofNodeUpdater::process" << std::endl;
   std::unordered_map<std::shared_ptr<ProofNode>, bool> visited;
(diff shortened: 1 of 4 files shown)


src/proof/proof_node_updater.h
```diff
@@ -141,11 +137,16 @@ class ProofNodeUpdater
   /**
    * Finalize the node cur. This is called at the moment that it is established
    * that cur will appear in the final proof. We do any final debug checking
-   * and add it to the results cache resCache if we are merging subproofs.
+   * and add it to the results cache resCache and resCacheNcWaiting
+   * if we are merging subproofs, where these map result formulas to proof
```
HanielB, line 141: phrasing

src/proof/proof_node_updater.cpp
```diff
@@ -69,26 +69,30 @@ void ProofNodeUpdater::process(std::shared_ptr<ProofNode> pf)
...
-void ProofNodeUpdater::processInternal(
-    std::shared_ptr<ProofNode> pf,
-    const std::vector<Node>& fa,
-    std::vector<std::shared_ptr<ProofNode>>& traversing)
+void ProofNodeUpdater::processInternal(std::shared_ptr<ProofNode> pf,
+                                       std::vector<Node>& fa)
 {
+  // Note that processInternal uses a single scope; fa is updated based on
+  // the current free assumptions of the proof nodes on the stack.
+  // The list of proof nodes we are currently traversing beneath. This is used
```
HanielB, line 80:
```suggestion

  // The list of proof nodes we are currently traversing beneath. This is used
```

src/proof/proof_node_updater.cpp
```diff
@@ -69,26 +69,30 @@ void ProofNodeUpdater::process(std::shared_ptr<ProofNode> pf)
...
+  // for checking for cycles in the overall proof.
+  std::vector<std::shared_ptr<ProofNode>> traversing;
+  // Map from formulas to (closed) proof nodes that prove that fact
+  std::map<Node, std::shared_ptr<ProofNode>> resCache;
+  // Map from formulas to non-closed proof nodes that prove that fact. These
+  // are replaced by proofs in the above map when applicable.
+  std::map<Node, std::vector<std::shared_ptr<ProofNode>>> resCacheNcWaiting;
+  std::unordered_map<const ProofNode*, bool> cfaMap;
```
HanielB, line 88: Explain this guy

HanielB (approved):
Looks great, just minor doc comments

PR #6983, opened 2021-08-04
Fix comments on dot printer

diegodrc:
This PR fixes the escaped characters in the dot printer. The output is now a valid DOT.

Signed-off-by: Diego Della Rocca de Camargos diegodellarocc@gmail.com


HanielB (approved):
LGTM

PR #6985, opened 2021-08-04
Avoid calling api::Solver constructor with Options

nafur:
This PR removes calls to the Solver constructor with an Options object, and instead uses the `setOption()` method.


4tXJ7f (approved):
LGTM

PR #6986, opened 2021-08-04
Consolidate solver resets

nafur:
This PR consolidates the two different reset implementations into a single function.


ajreynol (approved):
LGTM

PR #6989, opened 2021-08-05
Merge options cmake into general cmake file

nafur:
This PR removes `options/CMakeLists.txt` and merges its contents into `src/CMakeLists.txt` and does some generalization in the `mkoptions.py` script. This prepares the whole setup so that we can also generate options code in other directories than `options/` easily in the future: parts of the options code will be moved to the main folder.


mpreiner (approved):
LGTM! So the code and data for generating options code will still live in `options/`, but with this change it's now possible to generate code to be used in different modules of the solver such as the driver, right?

nafur:
> LGTM! So the code and data for generating options code will still live in `options/`, but with this change it's now possible to generate code to be used in different modules of the solver such as the driver, right?

Right, this change will simplify code generation in, e.g., the driver. The parsing code (https://github.com/cvc5/cvc5/blob/master/src/options/options_public_template.cpp#L345) should be moved to the driver at some point.

PR #6990, opened 2021-08-05
[Unit Tests] Add missing include

4tXJ7f:
For some configurations, the unit test `solver_black` was unable to find
the correct variant `std::find()`. This commit adds an explicit include
of `<algorithm>`, which declares the `std::find()` as required by the
unit test.


nafur (approved):
LGTM. Thanks!

PR #6992, opened 2021-08-06
Add CVC5ApiOptionException

nafur:
This PR adds the new `CVC5ApiOptionException`. While the driver does not (and can not) do anything special for the two existing api exceptions, it can (and should) properly inform the user about incorrect command line option usage.
The PR also removes the `UnrecognizedOptionException`. It is purely internal now, and immediately catched by the API wrapper. Having a separate exception for this is no longer useful.

The additional catch block in `main.cpp` is only temporary until option parsing has been migrated to the driver and setting the options is done properly via the API.


src/api/cpp/cvc5.h
```diff
@@ -118,6 +118,31 @@ class CVC5_EXPORT CVC5ApiRecoverableException : public CVC5ApiException
   }
 };
 
+/**
+ * A option-related API exception.
```
aniemetz, line 122:
```suggestion
 * An option-related API exception.
```

src/main/main.cpp
```diff
@@ -51,14 +51,27 @@ int main(int argc, char* argv[]) {
   Options opts;
   try {
     return runCvc5(argc, argv, opts);
-  } catch(OptionException& e) {
+  }
```
aniemetz, line 54: Can you maybe include this when formatting?

src/options/options_public_template.cpp
```diff
@@ -502,15 +502,14 @@ std::string get(const Options& options, const std::string& name)
 {
   Trace("options") << "Options::getOption(" << name << ")" << std::endl;
   ${getoption_handlers}$
-
-  throw UnrecognizedOptionException(name);
+  throw OptionException("Unrecognized informational or option key or setting: " + name);
```
aniemetz, line 505: "informational"?

aniemetz (changes requested):

src/options/options_public_template.cpp
nafur: I have no clue what I wanted the message to say there... I'll fix it.

src/options/options_public_template.cpp
```diff
@@ -502,15 +502,14 @@ std::string get(const Options& options, const std::string& name)
...
 }
 
 void setInternal(Options& opts, const std::string& name,
                                 const std::string& optionarg)
                                 {
 ${setoption_handlers}$
-  throw UnrecognizedOptionException(name);
+  throw OptionException("Unrecognized informational or option key or setting: " + name);
```
aniemetz, line 512: Ditto :)

aniemetz (approved):

PR #6993, opened 2021-08-06
Catch cases where recursive functions reference functions-to-synthesize

ajreynol:
We previously incorrectly allowed this, leading to problems that were unsolvable but where we would not warn the user this combination is not supported.

FYI @polgreen


HanielB (approved):
LGTM

PR #6994, opened 2021-08-06
Refactor setDefaults to use an options object

nafur:
This PR refactors the `setDefaults` method to accept an `Options` object as argument instead of using the current (static) `Options` object.


src/smt/set_defaults.cpp
```diff
@@ -46,35 +46,34 @@ using namespace cvc5::theory;
...
-  if (options::dumpUnsatCoresFull())
+  if (opts.driver.dumpUnsatCoresFull)
   {
     opts.driver.dumpUnsatCores = true;
   }
-  if (options::checkUnsatCores() || options::dumpUnsatCores()
-      || options::unsatAssumptions() || options::minimalUnsatCores()
-      || options::unsatCoresMode() != options::UnsatCoresMode::OFF)
+  if (opts.smt.checkUnsatCores || opts.driver.dumpUnsatCores
+      || opts.smt.unsatAssumptions || options::minimalUnsatCores()
```
4tXJ7f, line 70:
```suggestion
      || opts.smt.unsatAssumptions || opts.smt.minimalUnsatCores
```

4tXJ7f (changes requested):
Looks like one option wasn't updated yet :) Otherwise LGTM.

4tXJ7f (approved):
LGTM

PR #6995, opened 2021-08-06
Simplify generation of option module code.

nafur:
This PR simplifies both the code that generates the option modules and the generated code itself. It removes placeholders in the templates that are no longer used, gets rid of the option holder types (and replaces them by simple inline functions) and does some clean up on the related code in the `mkoptions.py` script.


4tXJ7f (approved):
LGTM, nice! :)

PR #7003, opened 2021-08-09
Support older CMake versions

### LINKED ISSUE #7001: Issues related to installation
I cannot successfully install cvc5 at the commit [c8bc488](https://github.com/cvc5/cvc5/commit/c8bc4881d4f3bf54258b0f01280fbda23a1dd651). 
The ERROR returned by `cmake version 3.10.2` is `list does not recognize sub-command PREPEND`. 

The installation works smoothly at the commit [e04248](https://github.com/cvc5/cvc5/commit/e04248ec858c97b8d716796f87ed42078a953967).
I would be very grateful if you could help me.

4tXJ7f:
Fixes #7001. Commit c8bc4881d4f3bf54258b0f01280fbda23a1dd651 introduced the use
of `list(PREPEND ...)` which was only introduced in version 3.15. We require
CMake 3.9 or later and this commit makes our build system compatible with older
CMake versions again. It also changes our CI to have two builds with Ubuntu
18.04 and two builds with Ubuntu 20.04 to better cover different versions of
build tools (including CMake 3.10).

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -1190,8 +1190,8 @@ set(options_toml_files
 )
 string(REPLACE "toml" "cpp;" options_gen_cpp_files ${options_toml_files})
 string(REPLACE "toml" "h;"   options_gen_h_files ${options_toml_files})
-list(PREPEND options_gen_cpp_files "options/options.cpp" "options/options_public.cpp")
-list(PREPEND options_gen_h_files "options/options.h")
+list(APPEND options_gen_cpp_files "options/options.cpp" "options/options_public.cpp")
+list(APPEND options_gen_h_files "options/options.h")
 
 libcvc5_add_sources(GENERATED ${options_gen_cpp_files} ${options_gen_h_files})
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,7 +26,7 @@ jobs:
             run_regression_args: --no-check-unsat-cores --no-check-proofs
 
           - name: ubuntu:production-clang
-            os: ubuntu-latest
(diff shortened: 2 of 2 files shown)


4tXJ7f:
This should also fix the nightlies.

src/CMakeLists.txt
```diff
@@ -1190,8 +1190,8 @@ set(options_toml_files
 )
 string(REPLACE "toml" "cpp;" options_gen_cpp_files ${options_toml_files})
 string(REPLACE "toml" "h;"   options_gen_h_files ${options_toml_files})
-list(PREPEND options_gen_cpp_files "options/options.cpp" "options/options_public.cpp")
-list(PREPEND options_gen_h_files "options/options.h")
+list(APPEND options_gen_cpp_files "options/options.cpp" "options/options_public.cpp")
+list(APPEND options_gen_h_files "options/options.h")
```
andrewvaughanj, line 1194: I don't know if the difference between `PREPEND` and `APPEND` matters here, but a few lines down, there's the use of a cvc5 CMake-based function called `list_prepend`, which implements `PREPEND` for older CMake.

nafur:
Thanks for the fix! (and sorry for breaking it in the first place...)

nafur (approved):

src/CMakeLists.txt
nafur: We only use the lists for fillings the dependencies for targets, so it should not matter.

4tXJ7f: Yeah, it doesn't matter in this context but thanks for the heads-up, I didn't know/remember about `list_prepend` :)

PR #7004, opened 2021-08-10
Disable external initialization of `thread_local` variables

4tXJ7f:
This commit adds `-fno-extern-tls-init` to the list of compiler flags to
tell the compiler that it can assume that there is no dynamic
initialization of thread-local variables in non-defining translation
units. This option should result in better performance and works around
crashing issues with our Windows build.


nafur (approved):
LGTM.

PR #7005, opened 2021-08-11
Replace `Maybe` with `std::optional`

4tXJ7f:
Because we are now using C++17, we can get rid of `Maybe` and instead
use `std::optional`, which offers the same functionality.


nafur (commented):
I would suggest using `*foo` instead of `foo.value()`.

src/theory/arith/simplex_update.h
```diff
@@ -329,13 +330,14 @@ class UpdateInfo {
...
-    else if (d_errorsChange.just() && d_errorsChange.value() < 0)
+    else if (d_errorsChange && d_errorsChange.value() < 0)
     {
       return ErrorDropped;
     }
-    else if (d_errorsChange.nothing() || d_errorsChange.value() == 0)
+    else if (d_errorsChange.value_or(0) == 0)
     {
-      if(d_focusDirection.just()){
+      if (d_focusDirection.has_value())
```
nafur, line 339:
```suggestion
      if (d_focusDirection)
```

nafur (approved):
LGTM.

PR #7007, opened 2021-08-12
[Strings] Make fact detection more robust

4tXJ7f:
Currently, our check for whether an inference is a fact or a lemma
involves checking whether the kind of the conclusion is a conjunction or
a disjunction. However, this does not take into account inferences of
other kinds such as `ite`s, which is a problem because they require a
decision from the SAT solver. This commit changes the condition to check
the theory of the conclusion. If the conclusion belongs to the theory of
strings, it considers it as a fact.


ajreynol (approved):
Good catch.

PR #7011, opened 2021-08-13
Make Theory class use Env

nafur:
This PR changes the Theory base class and the constructors of all theories to use the Env class to access contexts, logic information and the proof node manager.


ajreynol (approved):
LGTM, thanks for doing this.

PR #7012, opened 2021-08-14
Push Env class into TheoryState

nafur:
This PR is a follow-up to #7011, making the Env object available in the `TheoryState` base class.


4tXJ7f (approved):
LGTM

Issue #7010, opened 2021-08-13
`FindCryptoMiniSat.cmake` sets `-DSTATICCOMPILE=ON` without checking if statically linking is possible


andrewvaughanj:
**Describe the bug**

On my machine, I *do not* have the GCC static libraries (e.g., `pthread`, etc.); this is a problem if you try to build `cvc5` as follows:

```
./configure.sh --ninja --cryptominisat --auto-download
cd build
ninja
```

which gives:

```
FAILED: cryptominisat5_simple
: && /usr/bin/c++ -fno-stack-protector -fvisibility=hidden -O3 -O2 -Wl,--discard-all -Wl,--build-id=sha1 -static -Wl,--whole-archive -ldl -lpthread -Wl,--no-whole-archive -static cmsat5-src/CMakeFiles/cryptominisat5_simple-bin.dir/main_simple.cpp.o cmsat5-src/CMakeFiles/cryptominisat5_simple-bin.dir/main_common.cpp.o -o cryptominisat5_simple  lib/libcryptominisat5.a  -pthread && :
/usr/lib64/gcc/x86_64-suse-linux/11/../../../../x86_64-suse-linux/bin/ld: cannot find -ldl
/usr/lib64/gcc/x86_64-suse-linux/11/../../../../x86_64-suse-linux/bin/ld: cannot find -lpthread
/usr/lib64/gcc/x86_64-suse-linux/11/../../../../x86_64-suse-linux/bin/ld: cannot find -lm
/usr/lib64/gcc/x86_64-suse-linux/11/../../../../x86_64-suse-linux/bin/ld: cannot find -lpthread
/usr/lib64/gcc/x86_64-suse-linux/11/../../../../x86_64-suse-linux/bin/ld: cannot find -lc
collect2: error: ld returned 1 exit status
```

The "best" fix here is probably to get CMS to add support for something like `BUILD_EXECUTABLES`, which means that CMS's CMake does not build *either* of `cryptominisat5_simple` or `cryptominisat5` (I guess this would match cvc5's `BUILD_LIB_ONLY`). However, this would require an update of the CMS version that cvc5 uses (which might not be desirable).

Given this, maybe the fix is for `cmake/FindCryptoMiniSat.cmake` to check if it can statically link something with (at least) ` -ldl -lpthread` before doing `ExternalProject_Add` with `-DSTATICCOMPILE=ON`.

andrewvaughanj:
Oh, or this (which seems a lot nicer):

[...]

PR #7014, opened 2021-08-15
Add check for static libraries when compiling CryptoMiniSat #7010

andrewvaughanj:
This PR adds a check for CryptoMiniSat's static dependencies when configuring CryptoMiniSat. It changes a build-time failure into a configure-time failure.

`find_library` in CMake > 3.18 supports `REQUIRED`; as `cvc5` targets 3.10, I've implemented a check for what `find_library` returns.


4tXJ7f (approved):
LGTM, configure-time failure is definitely the better option here.

PR #7018, opened 2021-08-16
Move collectAssertedTerms to the Theory class

nafur:
This PR moves the `collectAssertedTerms` utility function from the model manager to the theory class. This allows to use the utility also when we are not currently building the model.


ajreynol (approved):
LGTM, you can test if you want, but this looks fine.

PR #7021, opened 2021-08-17
Initial refactoring of set defaults

ajreynol:
This commit starts to carve out better control flow structure in setDefaults.

It makes setDefaults contained in a class, and moves a few blocks of code to their own functions.

This class also makes options manager obsolete, it is deleted in this PR.

There should be no behavior change in this PR.


src/smt/smt_engine.cpp
```diff
@@ -195,10 +192,10 @@ void SmtEngine::finishInit()
   // set the random seed
   Random::getRandom().setSeed(d_env->getOptions().driver.seed);
 
-  // Call finish init on the options manager. This inializes the resource
-  // manager based on the options, and sets up the best default options
-  // based on our heuristics.
-  d_optm->finishInit(d_env->d_logic, d_isInternalSubsolver);
+  // Call finish init on the options manager. This inializes the logic
+  // and the best default options based on our heuristics.
```
nafur, line 196: We no longer have an options manager

nafur (approved):
Just one comment about a comment. Otherwise LGTM.

PR #7027, opened 2021-08-17
Adding parameters to Datatype python API documentation

yoni206:


src/api/python/cvc5.pxi
```diff
@@ -111,28 +111,42 @@ cdef class Datatype:
         return dc
 
     def getConstructor(self, str name):
-        """Return a constructor by name."""
+        """
```
yoni206, line 114: Some of the links to cpp did not work. Check that they work now.

nafur (approved):
LGTM, but the indentation is inconsistent.

PR #7029, opened 2021-08-17
Refactor proof output for TPTP

ajreynol:
This eliminates the old option "inst format mode" and makes proof output for TPTP organized as an ordinary proof output format.


HanielB (approved):
LGTM

PR #7030, opened 2021-08-17
Generalize inequality elimination in quantifiers rewriter

### LINKED ISSUE #6999: quantifier incompleteness compared to z3
Another minimized example where z3 says unsat and cvc5 says unknown:

(set-logic ALL)
(declare-sort T 0)
(declare-fun S (T Int) Int)
(declare-fun S (T Int Int) T)
(assert (forall ((? T) (x Int)) (= x (S (S ? 0 0) 0))))
(check-sat)

ajreynol:
This generalizes our inequality elimination technique to handle disequalities as well as inequalities.

#6999 is an example where a variable can be eliminated: if a variable x occurs only in an equality with negative required polarity, then the variable and that literal can be eliminated.

Fixes #6999.


HanielB (approved):
This is great. :)

PR #7031, opened 2021-08-18
Make TheoryProxy use Env, simplify initialization of PropEngine

ajreynol:
This simplifies our management of how/when proofs are enabled in the PropEngine.

### DIFF
--- a/src/smt/smt_solver.cpp
+++ b/src/smt/smt_solver.cpp
@@ -40,7 +40,6 @@ SmtSolver::SmtSolver(Env& env,
       d_state(state),
       d_pp(pp),
       d_stats(stats),
-      d_pnm(nullptr),
       d_theoryEngine(nullptr),
       d_propEngine(nullptr)
 {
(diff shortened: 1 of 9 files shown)


src/smt/smt_solver.cpp
```diff
@@ -61,16 +60,17 @@ void SmtSolver::finishInit(const LogicInfo& logicInfo)
     theory::TheoryConstructor::addTheory(d_theoryEngine.get(), id);
   }
   // Add the proof checkers for each theory
-  if (d_pnm)
+  ProofNodeManager* pnm = d_env.getProofNodeManager();
+  if (pnm)
   {
-    d_theoryEngine->initializeProofChecker(d_pnm->getChecker());
+    d_theoryEngine->initializeProofChecker(pnm->getChecker());
```
nafur, line 66: Shouldn't we check `env.isTheoryProofProducing()` instead?

nafur (approved):
Mostly LGTM, just one question.

src/smt/smt_solver.cpp
ajreynol: No, this is actually a key realization we had: the theory engine contains ways of checking e.g. Boolean rules that need to be enabled for preprocessing even if the TheoryEngine isnt producing proofs.

PR #7032, opened 2021-08-18
Start using Options via Env in arithmetic

nafur:
This PR refactors part of the arithmetic theory to use the options via the Env object.


src/theory/arith/nl/nonlinear_extension.cpp
```diff
@@ -37,27 +37,26 @@ namespace arith {
...
-      d_extTheoryCb(ee),
+      d_extTheoryCb(state.getEqualityEngine()),
       d_extTheory(d_extTheoryCb,
                   containing.getSatContext(),
                   containing.getUserContext(),
                   d_im),
       d_model(containing.getSatContext()),
-      d_trSlv(d_im, d_model, pnm, containing.getUserContext()),
-      d_extState(d_im, d_model, pnm, containing.getUserContext()),
+      d_trSlv(d_im, d_model, d_astate.getEnv().getProofNodeManager(), containing.getUserContext()),
```
ajreynol, line 52: I assume this will be cleaned up to just pass d_astate.getEnv() later?

src/theory/arith/theory_arith_private.cpp
```diff
@@ -138,7 +138,7 @@ TheoryArithPrivate::TheoryArithPrivate(TheoryArith& containing,
                           d_partialModel,
                           RaiseEqualityEngineConflict(*this),
                           d_pnm),
-      d_cmEnabled(c, options::arithCongMan()),
+      d_cmEnabled(c, options().arith.arithCongMan),
```
ajreynol, line 141: It would probably be cleaner to have the method options getter return `options().arith`, is this possible?

src/theory/arith/theory_arith_private.h
```diff
@@ -89,6 +89,11 @@ class TheoryArithPrivate {
 
   TheoryArith& d_containing;
 
+  const Options& options() const
+  {
+    return d_containing.d_theoryState->getEnv().getOptions();
```
ajreynol, line 94: I'd prefer if we already passed env / stored d_env here.

ajreynol (changes requested):

src/theory/arith/nl/nonlinear_extension.cpp
nafur: Yes, I've done that now.

ajreynol (approved):

PR #7017, opened 2021-08-16
Fix policy for eliminating quantified formulas

ajreynol:
This also consolidates the option `strictTriggers` into `userPatMode`.

Fixes #6996.

PR #7033, opened 2021-08-18
Minor fixes of policy for eliminating quantifiers

4tXJ7f:
PR #7017 fixed the policy for eliminating quantifiers but introduced
some minor issues, which this commit fixes:

- the `InstantiationEngine::checkOwnership()` method was changed to look
  for patterns in the wrong node.
- the PR changed the modes of the `--user-pat=MODE` method but reused
  one of the names. This commit fixes the name and adds a check in the
  options script.
- fixing the policy caused cvc5 to answer `unsat` instead of the
  expected `unknown` for `regress0/use_approx/bug812_approx.smt2`. The
  commit updates the expected result.

### DIFF
--- a/src/options/mkoptions.py
+++ b/src/options/mkoptions.py
@@ -610,9 +610,17 @@ def codegen_module(module, dst_dir, tpls):
                     cases=''.join(cases)))
 
             # Generate str-to-enum handler
+            names = set()
             cases = []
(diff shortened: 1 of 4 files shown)


ajreynol (approved):
LGTM, thanks for the fixes.

PR #7034, opened 2021-08-19
[unsat cores] [proofs] Revert test about when to explain propagations

HanielB:
Reverts a change from #7031, which changed the contract for when the proof CNF stream is handling propagations. When doing unsat cores with sat proofs (but not full proofs), theory engine will not be proof producing but the proof cnf stream still needs to connect theory lemmas with their clausified equivalents in the SAT solver.

### DIFF
--- a/src/prop/theory_proxy.cpp
+++ b/src/prop/theory_proxy.cpp
@@ -98,7 +98,8 @@ void TheoryProxy::explainPropagation(SatLiteral l, SatClause& explanation) {
 
   TrustNode tte = d_theoryEngine->getExplanation(lNode);
   Node theoryExplanation = tte.getNode();
-  if (d_env.isTheoryProofProducing())
+  if (options::produceProofs()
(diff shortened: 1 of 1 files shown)


src/prop/theory_proxy.cpp
```diff
@@ -98,7 +98,8 @@ void TheoryProxy::explainPropagation(SatLiteral l, SatClause& explanation) {
 
   TrustNode tte = d_theoryEngine->getExplanation(lNode);
   Node theoryExplanation = tte.getNode();
-  if (d_env.isTheoryProofProducing())
+  if (options::produceProofs()
```
ajreynol, line 101: I think this is the same as d_env.isSatProofProducing right?

ajreynol (changes requested):

src/prop/theory_proxy.cpp
HanielB: Indeed. I'll change.

ajreynol (approved):

PR #7039, opened 2021-08-19
Use Env class in nonlinear extension

nafur:
This PR refactors the nonlinear extension(s) to access options only via the environment class.


src/theory/arith/nl/ext/ext_state.cpp
```diff
@@ -33,11 +33,11 @@ namespace nl {
 ExtState::ExtState(InferenceManager& im, NlModel& model, Env& env)
     : d_im(im), d_model(model), d_env(env)
 {
-  d_false = NodeManager::currentNM()->mkConst(false);
-  d_true = NodeManager::currentNM()->mkConst(true);
-  d_zero = NodeManager::currentNM()->mkConst(Rational(0));
-  d_one = NodeManager::currentNM()->mkConst(Rational(1));
-  d_neg_one = NodeManager::currentNM()->mkConst(Rational(-1));
+  d_false = d_env.getNodeManager()->mkConst(false);
```
ajreynol, line 36: In general I'm slightly uneasy changing instances of NodeManager::currentNM() since I think its still an open question to me whether NodeManager should be a singleton.

Changing this would be under the assumption that it is not a singleton. I suppose this is progress to change to access via `d_env` since it makes it configurable based on what d_env stores, but future refactoring may become more complicated if we have multiple ways of doing this throughout the code.

ajreynol (changes requested):

src/theory/arith/nl/ext/ext_state.cpp
nafur: I see. I've removed these changes.

ajreynol (approved):

PR #7043, opened 2021-08-20
More refactoring of set defaults

ajreynol:
A few minor changes to which options are enabled for sygus, otherwise no intended changes.


nafur (approved):
LGTM.

PR #7044, opened 2021-08-20
Uniform treatment of trusted theory inferences in proofs

ajreynol:
Makes it so that all theory-specific proof rules for this purpose are replaced by the generic THEORY_INFERENCE.


HanielB (approved):
LGTM

PR #7046, opened 2021-08-20
Fix single invocation partition for higher-order

ajreynol:
It was not robust to cases where a function-to-synthesize occurred in a higher-order context.

Also does general clean up of the single invocation utility.


src/theory/quantifiers/single_inv_partition.cpp
```diff
@@ -233,55 +233,47 @@ bool SingleInvocationPartition::init(std::vector<Node>& funcs,
...
         for (unsigned j = 0; j < args.size(); j++)
         {
           Trace("si-prt") << args[j] << " ";
           if (args[j].getKind() == BOUND_VARIABLE
-              && std::find(terms.begin(), terms.end(), args[j]) == terms.end())
+              && !sb.contains(args[j]))
           {
-            terms.push_back(args[j]);
-            subs.push_back(d_si_vars[j]);
+            sb.add(args[j],d_si_vars[j]);
```
HanielB, line 267: clang-format

src/theory/quantifiers/single_inv_partition.cpp
```diff
@@ -445,7 +436,17 @@ bool SingleInvocationPartition::processConjunct(Node n,
         if (std::find(d_input_funcs.begin(), d_input_funcs.end(), f)
             != d_input_funcs.end())
         {
-          success = true;
+          // n must be fully applied. This catches cases where n is
+          // a function-to-synthesize that occurs in a higher-order context.
```
HanielB, line 440: Clarify that `n` is not fully applied since its sort is functional.

HanielB (changes requested):

HanielB:
Sorry, I only realized after commenting what had first confused me: who is not fully applied is `f`, which is why `n` has a functional type. Could you mention `f` in the comment?

HanielB (approved):

PR #7047, opened 2021-08-20
Top-level structure for set defaults

ajreynol:
This breaks setDefaults into three functions that summarize the high-level behavior of setDefaults.


src/smt/set_defaults.cpp
```diff
@@ -627,12 +635,12 @@ void SetDefaults::setDefaults(LogicInfo& logic, Options& opts)
     opts.smt.produceModels = true;
   }
 
-  /////////////////////////////////////////////////////////////////////////////
-  // Widen logic
-  /////////////////////////////////////////////////////////////////////////////
+  // widen the logic
   widenLogic(logic, opts);
-  /////////////////////////////////////////////////////////////////////////////
+}
```
4tXJ7f, line 640: Suggestion: move the `widenLogic` function to the top level?

ajreynol: I think this will eventually be inlined into finalizeLogic.  I'm leaving it here for now.

nafur (approved):
LGTM.

PR #7052, opened 2021-08-23
Add missing methods to Solver API for models

ajreynol:
This adds the last two remaining API methods required for implementing the text output of `get-model` on the API side.

A followup PR will implement `GetModelCommand` on the API side and eliminate the (last remaining) call to getSmtEngine() from the command layer.

This PR does some minor reorganization of the model cores code to account for the new interface.  It also removes a deprecated interface from TheoryModel.


src/theory/theory_model.h
```diff
@@ -318,6 +316,11 @@ class TheoryModel
   RepSet* getRepSetPtr() { return &d_rep_set; }
 
   //---------------------------- model cores
+  /**
+   * Is using model core (has a model core already been computed for this
```
aniemetz, line 320: That's a bit of a weird way to document this IMO. There's no added value to just repeating the name of the function. Maybe be more explicit and write
```suggestion
   * True if a model core has already been computed for this for this model.
```

src/theory/theory_model.h
```diff
@@ -318,6 +316,11 @@ class TheoryModel
...
+   * model)?
```
aniemetz, line 321:
```suggestion
```

test/unit/api/solver_black.cpp
```diff
@@ -1473,6 +1473,40 @@ TEST_F(TestApiBlackSolver, getValue3)
...
+  Sort uSort = d_solver.mkUninterpretedSort("u");
+  Sort intSort = d_solver.getIntegerSort();
+  Term x = d_solver.mkConst(uSort, "x");
+  Term y = d_solver.mkConst(uSort, "y");
+  Term z = d_solver.mkConst(uSort, "z");
+  Term f = d_solver.mkTerm(DISTINCT, x, y, z);
+  d_solver.assertFormula(f);
+  d_solver.checkSat();
+  ASSERT_NO_THROW(d_solver.getModelDomainElements(uSort));
+  ASSERT_TRUE(d_solver.getModelDomainElements(uSort).size() >= 3);
```
aniemetz, line 1488: Maybe also add a test where this is false?

test/unit/api/solver_black.cpp
```diff
@@ -1473,6 +1473,40 @@ TEST_F(TestApiBlackSolver, getValue3)
...
+  Term x = d_solver.mkConst(uSort, "x");
+  Term y = d_solver.mkConst(uSort, "y");
+  Term z = d_solver.mkConst(uSort, "z");
+  Term zero = d_solver.mkInteger(0);
+  Term f = d_solver.mkTerm(NOT, d_solver.mkTerm(EQUAL, x, y));
+  d_solver.assertFormula(f);
+  d_solver.checkSat();
+  ASSERT_TRUE(d_solver.isModelCoreSymbol(x));
+  ASSERT_TRUE(d_solver.isModelCoreSymbol(y));
+  ASSERT_TRUE(!d_solver.isModelCoreSymbol(z));
```
aniemetz, line 1506: Ditto

aniemetz (changes requested):

test/unit/api/solver_black.cpp
ajreynol: This is testing that x,y are model core symbols but z is not, so this is distinguish true/false returns.

test/unit/api/solver_black.cpp
ajreynol: This is saying there is a lower bound on the sort's interpretation: the sort must have 3 (or more) elements.

I've added another test that is an *upper* bound on a sort's interpretation, which necessarily requires the unit test to assert formulas with quantifiers and use finite model finding, but it should work.

PR #7053, opened 2021-08-23
Use options correctly in competition mode

nafur:
Fix competition mode by accessing options in the correct way in main functions.


4tXJ7f (approved):
LGTM

PR #7055, opened 2021-08-23
Avoiding duplicate search in maps.

MikolasJanota:
This commit identifies a couple of scenarios where an operation required 2 searches into a map/hashmap and replaces them with a single search. This also makes the code  shorter.


src/theory/quantifiers/instantiate.cpp
```diff
@@ -591,19 +591,13 @@ bool Instantiate::recordInstantiationInternal(Node q, std::vector<Node>& terms)
...
     Trace("inst-add-debug")
         << "Adding into context-dependent inst trie" << std::endl;
-    CDInstMatchTrie* imt;
-    std::map<Node, CDInstMatchTrie*>::iterator it = d_c_inst_match_trie.find(q);
-    if (it != d_c_inst_match_trie.end())
-    {
-      imt = it->second;
-    }
-    else
+    const auto [it, was_inserted] = d_c_inst_match_trie.insert({q, nullptr});
```
4tXJ7f, line 594: Nit: We generally use camelCase for variable names.

4tXJ7f (changes requested):
Thanks for your contribution. LGTM, I just have a minor comment. Could you also sign your commit?

src/theory/quantifiers/instantiate.cpp
MikolasJanota: Right, done

src/theory/quantifiers/instantiate.cpp
```diff
@@ -591,19 +591,13 @@ bool Instantiate::recordInstantiationInternal(Node q, std::vector<Node>& terms)
...
     Trace("inst-add-debug")
         << "Adding into context-dependent inst trie" << std::endl;
-    CDInstMatchTrie* imt;
-    std::map<Node, CDInstMatchTrie*>::iterator it = d_c_inst_match_trie.find(q);
-    if (it != d_c_inst_match_trie.end())
-    {
-      imt = it->second;
-    }
-    else
+    const auto [it, wasInserted] = d_c_inst_match_trie.insert({q, nullptr});
```
ajreynol, line 594: I'd prefer not using this syntax, can we just use `.first` / `.second` as needed?

ajreynol (changes requested):

src/theory/quantifiers/instantiate.cpp
MikolasJanota: I would say that this finally gives you a way of returning multiple values in C++,  so we might be missing out. Should this may be part of the guidelines? (The use of modern C++17 features in general)

nafur: We had a lengthy discussion internally about this point in particular. We decided that these "very new" features can be used, but are not mandated, and every developer can decide whether he wants to have/use them in the code he feels responsible for.
As for this PR, @ajreynol feels responsible for this part of the code and thus there will be no structured bindings for the time being :-).

MikolasJanota: Structured binding removed from this request.

4tXJ7f (commented):
It looks like our debug build is currently failing with the new changes. Could you please take another look? Thanks!

ajreynol (approved):

ajreynol:
Thanks for the improvements.

4tXJ7f (approved):
LGTM

PR #7059, opened 2021-08-24
Add `isNull` to cpp api tests, python api, and python api tests

yoni206:
While working on API documentation for python, I noticed that `isNull` is not wrapped by the python API. It is also not tested in the cpp API tests. This PR fixes both issues, and also updates the python api tests accordingly.

### DIFF
--- a/test/unit/api/datatype_api_black.cpp
+++ b/test/unit/api/datatype_api_black.cpp
@@ -37,6 +37,10 @@ TEST_F(TestApiBlackDatatype, mkDatatypeSort)
   Datatype d = listSort.getDatatype();
   DatatypeConstructor consConstr = d[0];
   DatatypeConstructor nilConstr = d[1];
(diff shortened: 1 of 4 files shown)


test/unit/api/datatype_api_black.cpp
```diff
@@ -37,6 +37,10 @@ TEST_F(TestApiBlackDatatype, mkDatatypeSort)
   Datatype d = listSort.getDatatype();
   DatatypeConstructor consConstr = d[0];
   DatatypeConstructor nilConstr = d[1];
+  ASSERT_FALSE(d.isNull());
```
aniemetz, line 40: I would prefer to have a dedicated `isNull` test, similarly to https://github.com/cvc5/cvc5/blob/master/test/unit/api/term_black.cpp#L215

aniemetz (changes requested):

PR #7062, opened 2021-08-24
Minor changes to E-matching utilities

ajreynol:


src/theory/quantifiers/ematching/pattern_term_selector.cpp
```diff
@@ -82,7 +81,7 @@ Node PatternTermSelector::getIsUsableEq(Node q, Node n)
       if (i == 1 && n.getKind() == EQUAL
           && !quantifiers::TermUtil::hasInstConstAttr(n[0]))
       {
-        return NodeManager::currentNM()->mkNode(n.getKind(), n[1], n[0]);
+        return NodeManager::currentNM()->mkNode(EQUAL, n[1], n[0]);
```
HanielB, line 84: Maybe you should be asserting hat `n`'s kind is `EQUAL`? The test above that it is a relational trigger does not entail this, AFAIK.

ajreynol: This is entailed by the condition directly above this line.

HanielB: Missed that, sorry. :)

HanielB (approved):

PR #7063, opened 2021-08-24
Reimplement support for relational triggers

ajreynol:
This makes relational triggers its own kind of instantiation match generator.  Future PRs will delete the code interleaved in the core InstMatchGenerator class for handling relational triggers.

This PR also fixes two issues related to how triggers are combined:
(1) instantiation match generators that are generated as part of multi-triggers now are properly specialized (using getInstMatchGenerator),
(2) there was a bug with auto-generated triggers: when the first auto-generated trigger was generated that was already a user trigger, then we would ignore all other auto-generated triggers.

This is work towards finding a solution for the choice.move.hard Facebook benchmark, where `relational-triggers` appear to be a possible solution.

### DIFF
--- a/src/theory/quantifiers/ematching/relational_match_generator.h
+++ b/src/theory/quantifiers/ematching/relational_match_generator.h
@@ -40,10 +40,11 @@ class RelationalMatchGenerator : public InstMatchGenerator
   bool reset(Node eqc) override;
   /** Get the next match. */
   int getNextMatch(Node q, InstMatch& m) override;
-
  private:
   /** the variable */
   Node d_var;
+  /** The index of the variable */
+  int64_t d_vindex;
   /** the relation kind */
   Kind d_rel;
   /** the right hand side */
@@ -55,6 +56,8 @@ class RelationalMatchGenerator : public InstMatchGenerator
   /** the current number of terms we have generated since the last call to reset
    */
   size_t d_counter;
+  /** stores whether we have written a value for d_var in the current match. */
+  bool d_rmPrev;
 };
 
 }  // namespace inst
--- a/src/theory/quantifiers/ematching/relational_match_generator.h
+++ b/src/theory/quantifiers/ematching/relational_match_generator.h
@@ -39,8 +39,12 @@ class RelationalMatchGenerator : public InstMatchGenerator
   int getNextMatch(Node q, InstMatch& m) override;
 
  private:
-  /** the relational trigger */
-  Node d_relTrigger;
+  /** the variable */
+  Node d_var;
+  /** the relation kind */
+  Kind d_rel;
+  /** the right hand side */
+  Node d_rhs;
   /** whether we have a required polarity */
   bool d_hasPol;
   /** the required polarity, if it exists */
--- a/src/theory/quantifiers/ematching/relational_match_generator.h
+++ b/src/theory/quantifiers/ematching/relational_match_generator.h
@@ -31,7 +31,10 @@ namespace inst {
 class RelationalMatchGenerator : public InstMatchGenerator
 {
  public:
-  RelationalMatchGenerator(Trigger* tparent, Node rtrigger, bool hasPol, bool pol);
+  RelationalMatchGenerator(Trigger* tparent,
+                           Node rtrigger,
+                           bool hasPol,
+                           bool pol);
 
   /** Reset */
   bool reset(Node eqc) override;
@@ -49,7 +52,8 @@ class RelationalMatchGenerator : public InstMatchGenerator
   bool d_hasPol;
   /** the required polarity, if it exists */
   bool d_pol;
-  /** the current number of terms we have generated since the last call to reset */
+  /** the current number of terms we have generated since the last call to reset
+   */
   size_t d_counter;
 };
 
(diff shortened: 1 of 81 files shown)


ajreynol:
+268-3275 on quantified logics on SMT-LIB with `--relational-triggers`, thus it is still not viable as a default option.

src/theory/quantifiers/ematching/trigger_term_info.h
```diff
@@ -108,6 +108,19 @@ class TriggerTermInfo
   static bool isRelationalTrigger(Node n);
   /** Is k a relational trigger kind? */
   static bool isRelationalTriggerKind(Kind k);
+  /**
+   * Is n a usable relational trigger, which is true if RelationalMatchGen
```
HanielB, line 112:
```suggestion
   * Is n a usable relational trigger, which is true if RelationalMatchGenerator
```

src/theory/quantifiers/ematching/trigger_term_info.h
```diff
@@ -108,6 +108,19 @@ class TriggerTermInfo
...
+   * can process n.
+   */
+  static bool isUsableRelationTrigger(Node n);
+  /**
+   * Same as above, but lit / hasPol / pol are updated to the required
+   * constructor arguments for RelationalMatchGen.
```
HanielB, line 118:
```suggestion
   * Same as above, but lit / hasPol / pol are updated to the required
   * constructor arguments for RelationalMatchGenerator.
```
Moreover, what is `lit` here seems to be `rtrigger` in that class, so you should sync them.

src/theory/quantifiers/ematching/trigger_term_info.cpp
```diff
@@ -70,6 +70,46 @@ bool TriggerTermInfo::isRelationalTriggerKind(Kind k)
...
+  bool hasPol, pol;
+  Node lit;
+  return isUsableRelationTrigger(n, hasPol, pol, lit);
+}
+bool TriggerTermInfo::isUsableRelationTrigger(Node n,
+                                              bool& hasPol,
+                                              bool& pol,
+                                              Node& lit)
+{
+  // relational triggers (not) (= (~ x t) true|false)
```
HanielB, line 84: define `~`

src/theory/quantifiers/ematching/relational_match_generator.h
```diff
@@ -0,0 +1,85 @@
...
+namespace inst {
+
+/**
+ * Match generator for relational triggers x ~ t where t is a ground term.
+ * This match generator tries a small fixed set of terms based on the kind of
+ * relation and the required polarity of the trigger in the quantified formula.
+ *
+ * For example, for quantified formula (forall ((x Int)) (=> (> x n) (P x))),
+ * we have that (> x n) is a relational trigger with required polarity "true".
+ * This generator will try the match `x -> n+1` only, where notice that n+1 is
```
HanielB, line 36: Clarify whether this is done modulo rewriting

src/theory/quantifiers/ematching/relational_match_generator.h
```diff
@@ -0,0 +1,85 @@
...
+ * `x -> n-1` will be generated.
+ *
+ * Currently this class handles only equality between real or integer valued
+ * terms, or inequalities (kind GEQ). It furthermore only considers ground terms
+ * t for the right hand side of relations.
+ */
+class RelationalMatchGenerator : public InstMatchGenerator
+{
+ public:
+  RelationalMatchGenerator(Trigger* tparent,
```
HanielB, line 51: document arguments

src/theory/quantifiers/ematching/relational_match_generator.h
```diff
@@ -0,0 +1,85 @@
...
+
+/**
+ * Match generator for relational triggers x ~ t where t is a ground term.
+ * This match generator tries a small fixed set of terms based on the kind of
+ * relation and the required polarity of the trigger in the quantified formula.
+ *
+ * For example, for quantified formula (forall ((x Int)) (=> (> x n) (P x))),
+ * we have that (> x n) is a relational trigger with required polarity "true".
+ * This generator will try the match `x -> n+1` only, where notice that n+1 is
+ * the canonical term chosen to satisfy x>n.
```
HanielB, line 37: say more about canonical terms, like how they're chosen and what properties are they supposed to be satisfy to be canonical.

HanielB (changes requested):

src/theory/quantifiers/ematching/relational_match_generator.h
ajreynol: What do you mean?  Whether we rewrite n+1?

HanielB: IIRC what I was trying to better understand here is whether `x` is instantiated only to a term syntactically equal to `n+1` or to a term that rewrites to the same as `n+1`. Does that make sense?

HanielB: So it's not worthwhile to try even `1+n`?

HanielB: I see now that I was confused. Since `n` is a constant, then `n+1` is generated automatically, not taken from the ground context, right?

ajreynol: `n` is a term actually

PR #7065, opened 2021-08-24
Consolidate language types

nafur:
This PR combines the two enums `InputLanguage` and `OutputLanguage` into a single `Language` type. It makes sure that `AST` is not used as input language using a predicate whenever the option is set.

### DIFF
--- a/src/expr/node_value.cpp
+++ b/src/expr/node_value.cpp
@@ -37,7 +37,7 @@ namespace expr {
 string NodeValue::toString() const {
   stringstream ss;
 
-  OutputLanguage outlang = (this == &null()) ? language::output::LANG_AUTO
+  Language outlang = (this == &null()) ? Language::LANG_AUTO
                                              : options::outputLanguage();
   toStream(ss, -1, false, outlang);
   return ss.str();
@@ -46,7 +46,7 @@ string NodeValue::toString() const {
 void NodeValue::toStream(std::ostream& out,
                          int toDepth,
                          size_t dag,
-                         OutputLanguage language) const
+                         Language language) const
 {
   // Ensure that this node value is live for the length of this call.
   // It really breaks things badly if we don't have a nonzero ref
--- a/src/expr/node_value.cpp
+++ b/src/expr/node_value.cpp
@@ -37,8 +37,8 @@ namespace expr {
 string NodeValue::toString() const {
   stringstream ss;
 
-  Language outlang = (this == &null()) ? Language::LANG_AUTO
-                                             : options::outputLanguage();
+  Language outlang =
+      (this == &null()) ? Language::LANG_AUTO : options::outputLanguage();
   toStream(ss, -1, false, outlang);
   return ss.str();
 }
--- a/src/options/language.cpp
+++ b/src/options/language.cpp
@@ -15,167 +15,67 @@
 
 #include "options/language.h"
 
-#include <sstream>
-
-#include "base/exception.h"
 #include "options/option_exception.h"
 
 namespace cvc5 {
+
+std::ostream& operator<<(std::ostream& out, Language lang) {
+  switch(lang) {
+  case Language::LANG_AUTO:
+    out << "LANG_AUTO";
+    break;
+  case Language::LANG_SMTLIB_V2_6:
+    out << "LANG_SMTLIB_V2_6";
+    break;
+  case Language::LANG_TPTP:
+    out << "LANG_TPTP";
+    break;
+  case Language::LANG_CVC: out << "LANG_CVC"; break;
+  case Language::LANG_SYGUS_V2: out << "LANG_SYGUS_V2"; break;
+  default:
+    out << "undefined_input_language";
+  }
+  return out;
+}
+
 namespace language {
 
-/** define the end points of smt2 languages */
-namespace input {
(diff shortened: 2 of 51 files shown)


src/options/language.cpp
```diff
@@ -15,167 +15,60 @@
...
-       << "' to an input language.";
-    throw cvc5::Exception(ss.str());
+  switch (lang)
+  {
+    case Language::LANG_AUTO: out << "LANG_AUTO"; break;
+    case Language::LANG_SMTLIB_V2_6: out << "LANG_SMTLIB_V2_6"; break;
+    case Language::LANG_TPTP: out << "LANG_TPTP"; break;
+    case Language::LANG_CVC: out << "LANG_CVC"; break;
+    case Language::LANG_SYGUS_V2: out << "LANG_SYGUS_V2"; break;
+    default: out << "undefined_input_language";
```
aniemetz, line 31: Shouldn't this be `undefined_language`?

src/options/options_handler.h
```diff
@@ -149,12 +149,12 @@ class OptionsHandler {
                     int value);
   void increaseVerbosity(const std::string& option, const std::string& flag);
   void decreaseVerbosity(const std::string& option, const std::string& flag);
-  OutputLanguage stringToOutputLanguage(const std::string& option,
-                                        const std::string& flag,
-                                        const std::string& optarg);
-  InputLanguage stringToInputLanguage(const std::string& option,
-                                      const std::string& flag,
-                                      const std::string& optarg);
+  Language stringToLanguage(const std::string& option,
```
aniemetz, line 152: Document

src/options/options_handler.h
```diff
@@ -149,12 +149,12 @@ class OptionsHandler {
...
+                            const std::string& flag,
+                            const std::string& optarg);
+  void languageIsNotAST(const std::string& option,
```
aniemetz, line 155: Ditto

src/options/set_language.cpp
```diff
@@ -37,37 +36,37 @@ SetLanguage::Scope::~Scope(){
...
-OutputLanguage SetLanguage::getLanguage(std::ostream& out) {
+Language SetLanguage::getLanguage(std::ostream& out)
+{
   long& l = out.iword(s_iosIndex);
   if(l == 0) {
     // set the default language on this ostream
     // (offset by one to detect whether default has been set yet)
     if(not Options::isCurrentNull()) {
-      l = options::outputLanguage() + 1;
+      l = static_cast<long>(options::outputLanguage()) + 1;
```
aniemetz, line 53: Can we use a fixed width type here?

src/options/set_language.cpp
```diff
@@ -37,37 +36,37 @@ SetLanguage::Scope::~Scope(){
...
     }
-    if(l <= 0 || l > language::output::LANG_MAX) {
+    if (l <= 0 || l > static_cast<long>(Language::LANG_MAX))
```
aniemetz, line 55: Ditto

src/options/language.h
```diff
@@ -24,147 +24,47 @@
...
-bool isInputLang_smt2(InputLanguage lang) CVC5_EXPORT;
-bool isOutputLang_smt2(OutputLanguage lang) CVC5_EXPORT;
-
-/**
-  * Is the language smtlib 2.6 or above? If exact=true, then this method returns
-  * false if the input language is not exactly SMT-LIB 2.6.
-  */
-bool isInputLang_smt2_6(InputLanguage lang, bool exact = false) CVC5_EXPORT;
-bool isOutputLang_smt2_6(OutputLanguage lang, bool exact = false) CVC5_EXPORT;
+inline bool isLang_smt2(Language lang)
```
aniemetz, line 56: Shouldn't this be CamelCase?

aniemetz (changes requested):
Shouldn't we also make sure that TPTP is not used as output language?

src/expr/node_value.cpp
```diff
@@ -37,16 +37,16 @@ namespace expr {
 string NodeValue::toString() const {
   stringstream ss;
 
-  OutputLanguage outlang = (this == &null()) ? language::output::LANG_AUTO
-                                             : options::outputLanguage();
+  Language outlang =
+      (this == &null()) ? Language::LANG_AUTO : options::outputLanguage();
```
aniemetz, line 41: Not for this PR, but maybe we should introduce `NodeValue::isNull()`. This is just ugly.

src/options/set_language.cpp
nafur: Not really: `std::ios_base::iword()` returns a `long`, and I'd rather stay with this type.

src/options/options_handler.h
```diff
@@ -149,12 +149,14 @@ class OptionsHandler {
...
-                                        const std::string& flag,
-                                        const std::string& optarg);
-  InputLanguage stringToInputLanguage(const std::string& option,
-                                      const std::string& flag,
-                                      const std::string& optarg);
+  /** Convert optarg to Language enum */
+  Language stringToLanguage(const std::string& option,
+                            const std::string& flag,
+                            const std::string& optarg);
+  /** Check that lang is not LANG_AST (which is not allowed for input) */
```
aniemetz, line 156:
```suggestion
  /** Check that lang is not LANG_AST (which is not allowed as input language). */
```

PR #7066, opened 2021-08-25
Implement alpha equivalence proofs

ajreynol:
This is a modified version of https://github.com/cvc5/cvc5/pull/6137 which accounts for extended rewriting between quantified formulas that are considered alpha equivalent.

It also generalizes the proof rule ALPHA_EQUIV.  Notice that if we were to make this rule more pedantic, we would check for variable shadowing during substitution, although this is not currently done.


src/proof/proof_rule.h
```diff
@@ -835,14 +835,21 @@ enum class PfRule : uint32_t
   // has prefix "QUANTIFIERS_INST_E_MATCHING", then t is the trigger that
   // generated the instantiation.
   INSTANTIATE,
+  // ======== Alpha equivalence
+  // Children: none
+  // Arguments: (F, y1 = z1, ..., yn = zn)
```
HanielB, line 840:
```suggestion
  // Arguments: (F, (= y1 z1), ..., (= yn zn)
```

src/theory/quantifiers/proof_checker.cpp
```diff
@@ -120,6 +121,41 @@ Node QuantifiersProofRuleChecker::checkInternal(
...
+    {
+      if (args[i].getKind() != kind::EQUAL)
+      {
+        return Node::null();
+      }
+      for (size_t j = 0; j < 2; j++)
+      {
+        Node v = args[i][j];
+        if (v.getKind() != kind::BOUND_VARIABLE
+            || allVars[j].find(v) != allVars[j].end())
```
HanielB, line 146: Feels a bit odd to use sets when you are enforcing uniqueness. Maybe just use vector?

HanielB (changes requested):
Maybe add a regression, like that of #6137?

src/theory/quantifiers/proof_checker.cpp
ajreynol: I think set is cleaner in this case.

HanielB: ok

HanielB (approved):

PR #7073, opened 2021-08-26
Refactoring of proof manager initialization

ajreynol:
No longer takes a backwards reference to SmtEngine.

Also takes minor changes to proof post-processor from proof-new.


src/smt/env.h
```diff
@@ -56,6 +57,7 @@ class TrustSubstitutionMap;
 class Env
 {
   friend class SmtEngine;
+  friend class smt::PfManager;
```
aniemetz, line 60: I don't think this is necessary?

ajreynol: We need this to call setProofNodeManager actually.

The reason for this: (1) Env has to be constructed first, (2) we don't know if proofs are enabled until after setDefaults runs, (3) Env needs pointer to the ProofNodeManager if/when proofs are enabled.

At the moment, this means we need a call to private initialization of setProofNodeManager in the constructor of PfManager.  I'm not sure if there is a cleaner way to do this.

aniemetz: I think there's no really good reason to have `Env::setProofNodeManager` private, is there?

aniemetz: After discussing this offline, we decided that this is fine. We only want `PfManager` to be able to call `Env::setProofNodeManager`, that's why it's private.

aniemetz (approved):

PR #7080, opened 2021-08-27
Refactor options handlers

nafur:
This PR does a major refactoring to the default option handlers (i.e. the piece of code that converts strings to the respective option types). It gets rid of the current multi-layer template specialization in favor of a simple overloaded template function.


src/options/options_public_template.cpp
```diff
@@ -30,200 +30,166 @@
...
+   * unsigned, checks that `optionarg` contains no minus. Then `conv` is called
+   * and the error conditions are handled: `conv` may throw an exception or
+   * `pos` may be smaller than the size of `optionarg`, indicating that not the
+   * entirety of `optionarg` was parsed.
+   */
+  template <typename T, typename FF>
+  T parseNumber(const std::string& flag,
+                const std::string& optionarg,
+                FF&& conv,
+                const std::string& type)
```
mpreiner, line 51: Is it possible to derive the name of the type as a string from T?

nafur: We could do `typeid(T).name()`, but for example `double` would be printed as `d`...

mpreiner: Ok, then let's keep it the way it is.

mpreiner (approved):

PR #7083, opened 2021-08-30
Fix issues with cyclic proofs due to double SYMM applications

ajreynol:
Our way of constructing proofs from the equality engine in very rare cases may cause cyclic proofs due to constructing double applications of SYMM, which are not recognized as assumptions in CDProof. This is due to an interplay between LazyProofChain using an underlying CDProof as its default proof generator, where the proof chain would use the proofs from the CDProof to form a cyclic proof.

This was encountered in 9 SMT-LIB benchmarks in QF_SLIA.

This PR makes us safer in several places related to double SYMM steps.


4tXJ7f (approved):
LGTM

PR #7086, opened 2021-08-30
bv-to-int-module: implementations and stubs for translating operators

yoni206:
This PR introduces most of the code for the translation of terms with operators. Some methods are left as stubs for future PRs.
A unit test is added with sanity checks for all implemented operators.


src/theory/bv/int_blaster.cpp
```diff
@@ -194,8 +215,352 @@ Node IntBlaster::translateWithChildren(
...
+  switch (oldKind)
+  {
+    case kind::BITVECTOR_ADD:
+    {
+      Assert(original.getNumChildren() == 2);
+      uint64_t bvsize = original[0].getType().getBitVectorSize();
+      std::cout << "panda before" << std::endl;
+      returnNode = createBVAddNode(
+          translated_children[0], translated_children[1], bvsize);
+      std::cout << "panda after" << std::endl;
```
ajreynol, line 256: Remove std::cout

src/theory/bv/int_blaster.cpp
```diff
@@ -194,8 +215,352 @@ Node IntBlaster::translateWithChildren(
...
+  Node returnNode;
+
+  // Translate according to the kind of the original node.
+  switch (oldKind)
+  {
+    case kind::BITVECTOR_ADD:
+    {
+      Assert(original.getNumChildren() == 2);
+      uint64_t bvsize = original[0].getType().getBitVectorSize();
+      std::cout << "panda before" << std::endl;
```
ajreynol, line 253: Ditto, remove.

ajreynol (changes requested):

src/theory/bv/int_blaster.cpp
yoni206: sorry for overlooking these debug messages, removed. in 14ac299.

ajreynol (approved):

PR #7087, opened 2021-08-30
Remove more instances of ufHo

ajreynol:
Towards replacing that option with a logic check throughout.


src/theory/theory_model_builder.cpp
```diff
@@ -21,6 +21,7 @@
 #include "options/smt_options.h"
```
aniemetz, line 21: Can we remove this?

aniemetz (approved):
One minor thing, else LGTM.

src/theory/theory_model_builder.cpp
ajreynol: Its still required for options::debugCheckModels() unfortunately.

PR #7088, opened 2021-08-30
Refactor filename handling

nafur:
This PR simplifies how we store the current input file name and handle setting and getting it.
It is now an option, that can also be set and get via `setInfo()` and `getInfo()`.

### DIFF
--- a/src/smt/smt_engine.cpp
+++ b/src/smt/smt_engine.cpp
@@ -712,7 +712,7 @@ void SmtEngine::defineFunctionRec(Node func,
 Result SmtEngine::quickCheck() {
   Assert(d_state->isFullyInited());
   Trace("smt") << "SMT quickCheck()" << endl;
-  const std::string& filename = d_env->getFilename();
+  const std::string& filename = d_env->getOptions().driver.filename;
   return Result(
       Result::ENTAILMENT_UNKNOWN, Result::REQUIRES_FULL_CHECK, filename);
 }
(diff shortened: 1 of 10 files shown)


src/util/statistics_public.cpp
```diff
@@ -30,7 +30,6 @@ void registerPublicStatistics(StatisticsRegistry& reg)
   reg.registerHistogram<api::Kind>("api::TERM", false);
 
   reg.registerValue<std::string>("driver::filename", false);
-  reg.registerValue<std::string>("driver::sat/unsat", false);
```
4tXJ7f, line 33: This change looks unrelated.

4tXJ7f (changes requested):
LGTM except for one minor comment.

src/smt/smt_engine.cpp
```diff
@@ -432,7 +413,10 @@ void SmtEngine::setInfo(const std::string& key, const std::string& value)
 
   if (key == "filename")
   {
-    d_env->setFilename(value);
+    d_env->d_options.driver.filename = value;
```
4tXJ7f, line 416: Can we remove the filename and the setters/getters from `Env` now?

src/util/statistics_public.cpp
4tXJ7f: Discussed offline. The change is on purpose and this is fine with me.

4tXJ7f (approved):
LGTM, thanks for the updates.

src/util/statistics_public.cpp
nafur: It technically is, but fits well with this PR

PR #7089, opened 2021-08-30
Make driver::totalTime a TimerStat

nafur:
This PR makes the `driver::totalTime` statistic a `TimerStat`. This eliminates the need to obtain the overall runtime in the driver and write it to the statistics towards the end of the runtime. This not only eliminates some code in the driver, but also gets rid of the call to `getSmtEngine()` that goes around the API.
One disclaimer: The statistics output changes from seconds as a double (i.e. `1.234`) to milliseconds (i.e. `1234ms`).


src/smt/env.cpp
```diff
@@ -51,6 +51,7 @@ Env::Env(NodeManager* nm, const Options* opts)
   {
     d_options.copyValues(*opts);
   }
+  d_statisticsRegistry->registerTimer("driver::totalTime").start();
```
4tXJ7f, line 54: Maybe we shouldn't call it `driver::totalTime` anymore since the timer does not live in the driver anymore? How about `global::totalTime`?

4tXJ7f (changes requested):
Just a request for a name change.

src/smt/env.cpp
nafur: Done.

4tXJ7f (approved):
LGTM, thanks!

PR #7090, opened 2021-08-30
Fix normal forms context-dependent simplification for strings

ajreynol:
This corrects an issue in our context-dependent simplification which limited the cases it was applied.

Previously, we were using a best content heuristic when it was applicable to compute the substitution, even if we were in an effort where normal forms had been computed. The latter should always be used if possible, since normal forms always have at least as much or more concrete components.


src/theory/strings/base_solver.cpp
```diff
@@ -368,13 +368,11 @@ void BaseSolver::checkConstantEquivalenceClasses(TermIndex* ti,
       {
         // Add explanations for the empty children
         Node emps;
-        while (count < n.getNumChildren()
-               && d_state.isEqualEmptyWord(n[count], emps))
+        if (d_state.isEqualEmptyWord(n[count], emps))
```
4tXJ7f, line 371: Can we turn the `while` loop above into a `for` loop and then turn this into:

```
if (...)
{
  ...
  continue;
}

...
```

4tXJ7f (approved):
LGTM. Wow, can't believe that we didn't discover this earlier :)

ajreynol:
+18-1 unsat +42-21 sat on SMT-LIB QF_S + QF_SLIA.

PR #7093, opened 2021-08-31
Add API check whether option in getOptionInfo() exists

nafur:
This PR adds a missing check in the API for `getOptionInfo()`.


test/unit/api/solver_black.cpp
```diff
@@ -1332,10 +1332,7 @@ TEST_F(TestApiBlackSolver, getOptionNames)
 TEST_F(TestApiBlackSolver, getOptionInfo)
 {
   {
-    auto info = d_solver.getOptionInfo("verbose");
-    ASSERT_EQ(info.name, "verbose");
-    ASSERT_EQ(info.aliases, std::vector<std::string>{});
-    ASSERT_TRUE(std::holds_alternative<api::OptionInfo::VoidInfo>(info.valueInfo));
+    EXPECT_THROW(d_solver.getOptionInfo("asdf-invalid"), CVC5ApiException);
```
4tXJ7f, line 1335: Why not keep both tests?

4tXJ7f (changes requested):
LGTM, except for a minor comment.

test/unit/api/solver_black.cpp
nafur: Changed the behaviour after some offline discussion, now both tests are kept.

mpreiner (approved):

4tXJ7f (approved):
LGTM, thanks for the update.

PR #7094, opened 2021-08-31
No longer use direct access to options in driver

nafur:
This PR refactors the driver to no longer directly access the Options object, but instead use `Solver::getOption()` or `Solver::getOptionInfo()`.


mpreiner (approved):
LGTM

PR #7096, opened 2021-08-31
Lazy proof reconstruction for strings theory solver

ajreynol:
This makes it so that we call the strings::InferProofCons::convert lazily, instead deferring to a temporary, trusted STRING_INFERENCE rule.

This decreases our average proof overhead on commonly solved instances from 64% to 58% on QF_S + QF_SLIA.  It also reduces timeouts significantly, from 2010 to 1460.  (Note these numbers were with a slightly extended version of this branch that includes other pending PRs for proofs).

Also fixes one subtle issue where proof reconstruction was misusing sequential substitutions (leading to warnings during proof post-processing).


4tXJ7f (approved):
LGTM

PR #7097, opened 2021-08-31
Make sure modes are sorted in ModeInfo

nafur:
This PR ensures that the possible modes returned in `getOptionInfo()` are always sorted. Their order would depend on the python dictionary ordering, which changed with a somewhat recent python version and thereby breaks our tests.


4tXJ7f (approved):
LGTM

Issue #3468, opened 2019-11-16
Make rewriter an instance in a central object


4tXJ7f:
The rewriter should not be a singleton.

PR #7100, opened 2021-08-31
rewriter: Make clearCaches non-static.

aniemetz:
This works towards getting rid of SmtEngine::currentSmtEngine and
closing #3468.


4tXJ7f (approved):
LGTM

PR #6396, opened 2021-04-20
Finish the Java Api

mudathirmahgoub:
This PR finishes the implementation of the Java API.
It also includes all java files in the build along with their unit tests.

PR #7105, opened 2021-09-01
Updated .clang_format for java

mudathirmahgoub:
This PR updates .clang_format file to format java files, in particular, the java test files for easy comparison with the corresponding c++ test files. 
Other than .clang_format, the only change in the java files are their format.


mudathirmahgoub:
Integrated in PR https://github.com/cvc5/cvc5/pull/6396, because I needed to format more files there. 
So no need for this PR.

PR #7106, opened 2021-09-01
Implement lazy proof checking modes

ajreynol:
This implements several variants of lazy proof checking in the core proof checker.

Note this extends the ProofNode class with an additional Boolean `d_provenChecked` indicating whether the `d_proven` field was checked by the underlying proof checker.

This PR updates the default proof checking mode to `lazy`.  The previous default can now be enabled by `--proof-check=eager-simple`.


4tXJ7f (approved):
LGTM

PR #7107, opened 2021-09-01
Clean up TheoryEngine header according to code style guidelines.

aniemetz:
This only moves code in order for the class declaration to conform to code style guidelines.


ajreynol (approved):
Nit: I think we can remove all occurrences of `inline` as I believe these have no effect.

PR #7109, opened 2021-09-01
Driver & Options cleanup

nafur:
This PR does some cleanup in the driver and the options. It removes the now obsolete `public` attribute that allowed using the options in the driver, and removes a bunch of includes from the driver that are no longer necessary.


mpreiner (approved):

aniemetz (approved):

4tXJ7f (approved):
LGTM

PR #7112, opened 2021-09-01
pp: Derive PreprocessingPass from EnvObj.

aniemetz:
This is work towards making Rewriter::rewrite non-static.
It further reorders class declarations for ITECompressor
and ITESimplifier to comply to code style guidelines.

PR #7110, opened 2021-09-01
rewriter: Make rewriteEqualityExt non-static.

aniemetz:
More work towards getting rid of SmtEngine::currentSmtEngine and
closing #3468.

PR #7113, opened 2021-09-01
Add class EnvObj.

aniemetz:
This class will serve as base class for classes that need access to the
environment. This does not yet have classes derive from this base class.

Will update #7110 and #7112 to use this after this is in.


src/smt/envobj.h
```diff
@@ -0,0 +1,53 @@
+/******************************************************************************
```
ajreynol, line 1: Nit: should this file be `env_obj.h`?

ajreynol (changes requested):

ajreynol (approved):

PR #7114, opened 2021-09-01
[Unit Tests] Fix bags rewrite test

4tXJ7f:
PR #7103 did not quite fix the unit test: The types of the lambda and
the expected bags were still off. This fixes the unit test.


mudathirmahgoub (approved):

mudathirmahgoub:
Nice catch Andres

PR #7117, opened 2021-09-02
[Unit Tests] Fix shell test for Editline

4tXJ7f:
When using `--editline`, our `interactive_shell_black` unit test was not
working because the unit test was redirecting `std::cin` and `std::cout`
to `std::stringstream`s, which does not work with Editline. This commit
refactors our `InteractiveShell` class to explicitly take the input and
output streams as arguments, which fixes the issue because we do not use
Editline for input streams that are not `std::cin`. Additionally, the
commit updates the unit test to use SMT-LIB syntax instead of the
deprecated CVC language.


nafur (approved):
LGTM.

PR #7120, opened 2021-09-02
Refactoring and fixes of set defaults for quantifiers

ajreynol:
This PR further refactors set defaults for incompatibility issues related to quantifiers.

It adds a new restriction that was discovered recently: --nl-rlv should not be used in quantified logics. Some regressions are modified that are impacted by this restriction.

Also does minor rearrangements to the order in which default options are set.


test/regress/regress0/quantifiers/cegqi-needs-justify.smt2
```diff
@@ -1,4 +1,4 @@
-; COMMAND-LINE: --nl-rlv=always --no-sygus-inst
+; COMMAND-LINE: --no-sygus-inst
```
aniemetz, line 1: I think `--no-sygus-inst` is not necessary for this one anymore.

aniemetz (approved):

PR #7121, opened 2021-09-02
EnvObj: Restrict access.

aniemetz:
This makes all members of EnvObj protected in order to prevent creating
EnvObj instances and call member functions. It further uses protected
inheritance (instead of public) for derived classes in order to disallow
`EnvObj* a = new DerivedObj();`.


ajreynol (approved):
I'm in favor of this, LGTM.

4tXJ7f (approved):
🤘

ajreynol:
Actually there is a different reason I had in mind why this PR is a good idea: it would technically be possible for one EnvObj to call another EnvObj to rewrite things for one another, like a preprocessing pass could ask TheoryEngine to rewrite something for it.  This PR makes it so that these unintuitive interactions are prohibited.

PR #7123, opened 2021-09-02
Add Lfsc print channel utilities

ajreynol:
These are utilities used for two-pass printing of LFSC proofs.


src/proof/lfsc/lfsc_print_channel.h
```diff
@@ -0,0 +1,128 @@
...
+
+namespace cvc5 {
+namespace proof {
+
+/**
+ * LFSC proofs are printed in two phases, the first phase computes the
+ * letification of terms in the proof as well as other information that is
+ * required for printing the preamble of the proof, the second phase prints the
+ * proof to an output stream. This is the base class for these two phases.
+ */
```
yoni206, line 37:
```suggestion
/**
 * LFSC proofs are printed in two phases: the first phase computes the
 * letification of terms in the proof as well as other information that is
 * required for printing the preamble of the proof. The second phase prints the
 * proof to an output stream. This is the base class for these two phases.
 */
```

src/proof/lfsc/lfsc_print_channel.cpp
```diff
@@ -0,0 +1,161 @@
...
+  std::stringstream ss;
+  n.toStream(ss, -1, 0, Language::LANG_SMTLIB_V2_6);
+  std::string s = ss.str();
+  cleanSymbols(s);
+  out << s;
+}
+
+void LfscPrintChannelOut::printTypeNodeInternal(std::ostream& out, TypeNode tn)
+{
+  // must clean indexed symbols
```
yoni206, line 90: Maybe explain why? And is this the right line for this? Same thing in the previous function.

yoni206 (approved):
LGTM, minor comments inline for your consideration.

PR #7128, opened 2021-09-03
theory: Have Theory and TheoryArith* derive from EnvObj.

aniemetz:
Note: the diff in theory.h is only huge because of small changes that caused huge reformat. The effective change is only that it now derives from `EnvObj`, does not have a member `d_env` (because it inherits it) and does not need `getLogicInfo` (because that's in `EnvObj`).


ajreynol:
I think the formating in theory.h is still off due to a weird indentation all the way up on line 62 of theory.h.  The `Theory` class should have no indentation, can you fix this? I'd guess removing the indentation on line 62 would fix this.

aniemetz:
Good catch! Have fixed.

ajreynol (approved):

PR #7132, opened 2021-09-03
Make quantifier module classes derive from EnvObj

ajreynol:


src/theory/quantifiers/sygus/sygus_module.h
```diff
@@ -49,7 +50,7 @@ class QuantifiersInferenceManager;
  * Modules implement an initialize function, which determines whether the module
  * will take responsibility for the given conjecture.
  */
-class SygusModule
+class SygusModule : protected EnvObj
```
aniemetz, line 53: It looks like this doesn't actually use anything from `Env`. Do we have to derive here?

src/theory/quantifiers/quant_module.h
```diff
@@ -40,7 +41,7 @@ class TermDb;
  * This is the virtual class for defining subsolvers of the quantifiers theory.
  * It has a similar interface to a Theory object.
  */
-class QuantifiersModule
+class QuantifiersModule : protected EnvObj
```
aniemetz, line 44: Ditto, this doesn't use anything from `Env`?

src/theory/quantifiers/sygus/sygus_module.h
ajreynol: This inheritance in part allows `Rewriter::rewrite` to be replaced by `rewrite` throughout the majority of the quantifiers code.

src/theory/quantifiers/quant_module.h
ajreynol: Ditto.

aniemetz (approved):

PR #7138, opened 2021-09-03
Refactor options::set()

nafur:
This PR refactors the code generation and the generate code for `options::set()`.

### DIFF
--- a/src/options/options_public_template.cpp
+++ b/src/options/options_public_template.cpp
@@ -220,7 +220,10 @@ namespace cvc5::options
   ${set_impl}$
     // clang-format on
   }
-  else { throw OptionException("Unrecognized option key or setting: " + name); }
+  else
+  {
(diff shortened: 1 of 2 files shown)


src/options/options_public_template.cpp
```diff
@@ -211,27 +211,16 @@ namespace cvc5::options
...
-}
-
-void set(Options& opts, const std::string& name, const std::string& optionarg)
-{
-
-  Trace("options") << "setOption(" << name << ", " << optionarg << ")"
-                   << std::endl;
-  // first update this object
-  setInternal(opts, name, optionarg);
+  else { throw OptionException("Unrecognized option key or setting: " + name); }
```
mpreiner, line 223:
```suggestion
  else
  {
    throw OptionException("Unrecognized option key or setting: " + name);
  }
```

mpreiner (approved):
LGTM, one minor suggestion.

PR #7142, opened 2021-09-03
sygus: Eliminate calls to Rewriter::rewrite.

aniemetz:
This derives sygus unification utility objects from EnvObj where
necessary. There's one remaining occurrence of Rewriter::rewrite in
sygus_unif_rl.cpp, which is a bit tricky to address and thus subject to
a future PR.

### DIFF
--- a/src/theory/quantifiers/sygus/sygus_unif_io.cpp
+++ b/src/theory/quantifiers/sygus/sygus_unif_io.cpp
@@ -509,8 +509,9 @@ void SubsumeTrie::getLeaves(const std::vector<Node>& vals,
   getLeavesInternal(vals, pol, v, 0, -2);
 }
 
-SygusUnifIo::SygusUnifIo(SynthConjecture* p)
-    : d_parent(p),
+SygusUnifIo::SygusUnifIo(Env& env, SynthConjecture* p)
+    : SygusUnif(env),
+      d_parent(p),
       d_check_sol(false),
       d_cond_count(0),
       d_sol_term_size(0),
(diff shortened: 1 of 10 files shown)


src/theory/quantifiers/sygus/sygus_unif_io.cpp
```diff
@@ -560,7 +561,7 @@ void SygusUnifIo::notifyEnumeration(Node e, Node v, std::vector<Node>& lemmas)
   Assert(!d_examples.empty());
   Assert(d_examples.size() == d_examples_out.size());
 
-  EnumInfo& ei = d_strategy[c].getEnumInfo(e);
+  EnumInfo& ei = d_strategy.at(c).getEnumInfo(e);
```
ajreynol, line 564: Whats the motivation for this change?

aniemetz: You can't use `[c]` because `SygusUnifStrategy` doesn't have a default constructor -- it takes `Env`. So you have to make sure that it is already inserted, you can't default insert if the an entry with they key `c` doesn't exist. From what I understand from the code, that's actually the invariant here anyways, so it would be better practice to always use `at` if we assert that it exists and no new entry is to be inserted. It would better document what is going on here (for cases where there is a default constructor).

ajreynol: Ah, I see, make sense.

ajreynol (approved):

PR #7144, opened 2021-09-05
Use `EnvObj` methods instead of `Theory` methods

4tXJ7f:
This removes the methods `getEnv()`, `options()`, `getSatContext()`, and
`getUserContext()` from the `Theory` class because they are now part of
`EnvObj`. Additionally, this commit converts the inference managers to
`EnvObj` because of there were some calls to retrieve the contexts from
`Theory` in those classes.

### DIFF
--- a/src/theory/arith/inference_manager.cpp
+++ b/src/theory/arith/inference_manager.cpp
@@ -24,13 +24,14 @@ namespace cvc5 {
 namespace theory {
 namespace arith {
 
-InferenceManager::InferenceManager(TheoryArith& ta,
+InferenceManager::InferenceManager(Env& env,
+                                   TheoryArith& ta,
                                    ArithState& astate,
                                    ProofNodeManager* pnm)
-    : InferenceManagerBuffered(ta, astate, pnm, "theory::arith::"),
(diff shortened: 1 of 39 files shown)


ajreynol (changes requested):
I'm in favor of most of these changes.

However, there is one point that I'd suggest to change: the constructor of TheoryInferenceManager (and its derived classes) already takes TheoryState, which has Env.  I'd prefer if we just extract the Env& from the passed TheoryState instead of taking it as another argument.  This would make things cleaner, since it e.g. would be nonsensical to have an inference manager and theory state using different Env&.

ajreynol:
This also brings up another question: maybe TheoryState should not be EnvObj, but rather seen as a further wrapper on Env?  I notice it has `TheoryState::getEnv()` which is a bit funny.

aniemetz:
I think we need to make a basic design decision here. Either states never own an `Env` but only non-`Env` stuff, and the classes that need an `Env` are an `EnvObj`. Or States are wrappers around `Env` (but then it should be an `EnvObj`) or are an `EnvObj`.

ajreynol (approved):
LGTM.

We decided during the meeting that it is ok to pass Env& as an additional argumemt; TheoryState::getEnv should be removed.

PR #7145, opened 2021-09-07
Towards standard usage of ExtendedRewriter

ajreynol:
This PR:
- Adds `extendedRewrite` to EnvObj and Rewriter.
- Eliminates static calls to Rewriter::rewrite from *within* the extended rewriter. Instead, the use of extended rewriter is always through Rewriter, which passes itself to the ExtendedRewriter.
- Make most uses of extended rewriter non-static.  I've added a placeholder method `Rewriter::callExtendedRewrite` for places in the code that call the extended rewriter are currently difficult to eliminate.

### DIFF
--- a/src/theory/quantifiers/sygus/enum_value_manager.cpp
+++ b/src/theory/quantifiers/sygus/enum_value_manager.cpp
@@ -34,7 +34,7 @@ namespace theory {
 namespace quantifiers {
 
 EnumValueManager::EnumValueManager(Node e,
-                   Env& env,
+                                   Env& env,
                                    QuantifiersState& qs,
                                    QuantifiersInferenceManager& qim,
                                    TermRegistry& tr,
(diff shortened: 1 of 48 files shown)


src/theory/quantifiers/sygus/enum_value_manager.cpp
```diff
@@ -34,12 +34,14 @@ namespace theory {
 namespace quantifiers {
 
 EnumValueManager::EnumValueManager(Node e,
+                                   Env& env,
```
4tXJ7f, line 37: I have a slight preference for always passing `Env` as the first argument. What do you think about that?

4tXJ7f (approved):
LGTM, just one minor suggestion.

src/theory/quantifiers/sygus/enum_value_manager.cpp
ajreynol: I think it could go either way, I did it this way so that `env` directly proceeds `qs`. Perhaps `e` should be all the way towards the end.

PR #7147, opened 2021-09-07
proofs: Properly track pre- and post-rewrites in bbAtom().

mpreiner:
This commit refactors the proof bit-blaster to properly track the pre- and post-rewrites in bbAtom(). The individual bit-blast steps are recorded in a term conversion proof generator. The overall bit-blast proof is stored in a BitblastProofGenerator, which tracks the pre- and post-rewrite and includes the bit-blast proof of the TCPG.


src/theory/bv/bitblast/proof_bitblaster.h
```diff
@@ -27,6 +27,40 @@ class TConvProofGenerator;
 namespace theory {
 namespace bv {
 
+/** Proof generator fot bit-blast proofs. */
```
ajreynol, line 30: I'd prefer if this was moved to its own file.

src/theory/bv/bitblast/proof_bitblaster.h
```diff
@@ -27,6 +27,40 @@ class TConvProofGenerator;
...
+
+  /**
+   * Get proof for, which expects an equality of the form t = bb(t).
+   * This returns a proof based on the term conversion proof generator utility.
+   */
+  std::shared_ptr<ProofNode> getProofFor(Node eq) override;
+
+  std::string identify() const override { return "BitblastStepProofGenerator"; }
+
+  /** Record bit-blast step. */
```
ajreynol, line 45: Can you add some more documentation here about the arguments?

ajreynol (changes requested):
A few minor requests.

src/theory/bv/bitblast/proof_bitblaster.cpp
```diff
@@ -23,6 +23,69 @@ namespace cvc5 {
...
+  }
+  else
+  {
+    /* Fine-grained bit-blast step for bit-blasting bit-vector atoms (bbAtom()).
+     * Bit-blasting atoms involves three steps:
+     * 1) pre-rewrite: rewrite atom
+     * 2) bit-blast rewritten atom
+     * 3) post-rewrite: rewrite bit-blasted atom
+     * bit-blasts and post-rewrites.
+     */
```
HanielB, line 50: add here a proof tree combining these three things

src/theory/bv/bitblast/proof_bitblaster.cpp
```diff
@@ -23,6 +23,69 @@ namespace cvc5 {
...
+
+    // Record pre-rewrite of bit-blasted term.
+    if (t != rwt)
+    {
+      cdp.addStep(t.eqNode(rwt), PfRule::REWRITE, {}, {t});
+      transSteps.push_back(t.eqNode(rwt));
+    }
+
+    // Record bit-blast step.
+    cdp.addProof(d_tcpg->getProofFor(rwt.eqNode(bbt)));
```
HanielB, line 64: when `rwt` is the same as `t` this is the statement that allows the return of this method to work properly. This is not very clear currently. Add to the beginning of this block, where you describe the proof being built, how it relates to what is stored in `cdp` and the final call in the method.

src/theory/bv/bitblast/proof_bitblaster.cpp
```diff
@@ -23,6 +23,69 @@ namespace cvc5 {
...
+
+    // Record post-rewrite of bit-blasted term.
+    Node rwbbt = Rewriter::rewrite(bbt);
+    if (bbt != rwbbt)
+    {
+      cdp.addStep(bbt.eqNode(rwbbt), PfRule::REWRITE, {}, {bbt});
+      transSteps.push_back(bbt.eqNode(rwbbt));
+    }
+
+    if (transSteps.size() > 1)
```
HanielB, line 75: reforce here that the `else` of this case is a no-op because `eq` will already have been justified above

HanielB (changes requested):

HanielB (approved):
LGTM

ajreynol (approved):

PR #7148, opened 2021-09-08
Remove `TheoryState::options()`

4tXJ7f:
This commit removes `TheoryState::options()` by changing more classes to
`EnvObj`s.

### DIFF
--- a/src/theory/quantifiers/quantifiers_modules.cpp
+++ b/src/theory/quantifiers/quantifiers_modules.cpp
@@ -24,9 +24,8 @@ namespace cvc5 {
 namespace theory {
 namespace quantifiers {
 
-QuantifiersModules::QuantifiersModules(Env& env)
-    : EnvObj(env),
-      d_rel_dom(nullptr),
+QuantifiersModules::QuantifiersModules()
+    : d_rel_dom(nullptr),
       d_alpha_equiv(nullptr),
       d_inst_engine(nullptr),
       d_model_engine(nullptr),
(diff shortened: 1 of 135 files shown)


src/theory/quantifiers/quantifiers_modules.cpp
```diff
@@ -24,8 +24,9 @@ namespace cvc5 {
 namespace theory {
 namespace quantifiers {
 
-QuantifiersModules::QuantifiersModules()
-    : d_rel_dom(nullptr),
+QuantifiersModules::QuantifiersModules(Env& env)
```
ajreynol, line 27: This class is purely for initialization of the quantifiers strategy, it doesnt have to be an EnvObj (in another PR, I've added Env& to the initialize function).

src/theory/quantifiers/sygus/synth_conjecture.cpp
```diff
@@ -45,18 +45,20 @@ namespace cvc5 {
 namespace theory {
 namespace quantifiers {
 
-SynthConjecture::SynthConjecture(QuantifiersState& qs,
+SynthConjecture::SynthConjecture(Env& env,
```
ajreynol, line 48: FYI this is also converted to EnvObj in https://github.com/cvc5/cvc5/pull/7145.

ajreynol (changes requested):
There is one class that doesn't have to be EnvObj.

Otherwise LGTM, note that some of these classes were converted in https://github.com/cvc5/cvc5/pull/7145; I don't think anything will be hard to resolve (depending on which PR goes in first).

src/theory/quantifiers/quantifiers_modules.cpp
4tXJ7f: Makes sense. I've updated this.

ajreynol (approved):

ajreynol:
Will need to resolve conflicts, else LGTM

PR #7150, opened 2021-09-08
Add difficulty post-processor

ajreynol:
This is work towards supporting a `(get-difficulty)` command.

This is a helper class for transforming the difficulty of preprocessed assertions to difficulty of the input assertions.


src/smt/difficulty_post_processor.h
```diff
@@ -0,0 +1,77 @@
...
+   * the source of an assertion.
+   */
+  bool shouldUpdate(std::shared_ptr<ProofNode> pn,
+                    const std::vector<Node>& fa,
+                    bool& continueUpdate) override;
+  /** Get the (acculumated) difficulty map for the last processed proof node */
+  void getDifficultyMap(std::map<Node, Node>& dmap) const;
+
+ private:
+  /** Reference to the env class */
```
nafur, line 63: You don't seem to use `d_env` (yet?). Also, how about making it an `EnvObj`?

src/smt/difficulty_post_processor.h
```diff
@@ -0,0 +1,77 @@
...
+
+/**
+ * A postprocess callback that computes the difficulty based on the structure
+ * of the proof. In particular, this class assesses what the source of an
+ * assertion was by considering the shape of the proof. For instance, if
+ * assertion A entails x=t, and this was used to derive a substitution
+ * { x -> t } to assertion B, then B is the source of B*{ x -> t }. The
+ * difficulty of this assertion is carried to B and not A.
+ */
+class DifficultyPostprocessCallback : public ProofNodeUpdaterCallback
```
nafur, line 42: Can you write a bit about how this is supposed to be used?

src/smt/difficulty_post_processor.h
```diff
@@ -0,0 +1,77 @@
...
+{
+ public:
+  DifficultyPostprocessCallback(Env& env);
+  ~DifficultyPostprocessCallback() {}
+  /**
+   * Set current difficulty of the next proof to process to the (integer)
+   * value stored in Node d.
+   */
+  bool setCurrentDifficulty(Node d);
```
nafur, line 51: Is this only called with nodes that hold integers? If so, just accept a `uint64_t` directly.

src/smt/difficulty_post_processor.h
```diff
@@ -0,0 +1,77 @@
...
+  /**
+   * Should proof pn be updated? This is used to selectively traverse to e.g.
+   * the source of an assertion.
+   */
+  bool shouldUpdate(std::shared_ptr<ProofNode> pn,
+                    const std::vector<Node>& fa,
+                    bool& continueUpdate) override;
+  /** Get the (acculumated) difficulty map for the last processed proof node */
+  void getDifficultyMap(std::map<Node, Node>& dmap) const;
```
nafur, line 60: Why not simply return a reference to `d_accMap`? Why would we convert it to a `Node -> Node` map?

src/smt/difficulty_post_processor.h
```diff
@@ -0,0 +1,77 @@
...
+ */
+
+#include "cvc5_private.h"
+
+#ifndef CVC5__SMT__DIFFICULTY_POST_PROCESSOR_H
+#define CVC5__SMT__DIFFICULTY_POST_PROCESSOR_H
+
+#include <map>
+#include <sstream>
+#include <unordered_set>
```
nafur, line 24: You don't seem to use an `unordered_set` or an `stringstream`

src/smt/difficulty_post_processor.h
```diff
@@ -0,0 +1,77 @@
...
+#include "proof/proof_node_updater.h"
+
+namespace cvc5 {
+
+class Env;
+
+namespace smt {
+
+/**
+ * A postprocess callback that computes the difficulty based on the structure
```
HanielB, line 35:
```suggestion
 * A postprocess callback that computes difficulty based on the structure
```

src/smt/difficulty_post_processor.h
```diff
@@ -0,0 +1,77 @@
...
+ * of the proof. In particular, this class assesses what the source of an
+ * assertion was by considering the shape of the proof. For instance, if
+ * assertion A entails x=t, and this was used to derive a substitution
+ * { x -> t } to assertion B, then B is the source of B*{ x -> t }. The
+ * difficulty of this assertion is carried to B and not A.
```
HanielB, line 40: I'm confused by what the "source of an assertion" is supposed to be. Reading the explanation I was surprised that A is not the source. I think I see what you're getting at, but the documentation could be improved to be more clear.

src/smt/difficulty_post_processor.h
HanielB, line 51: Document the return. Refer explicitly to `d_currDifficulty`.

src/smt/difficulty_post_processor.h
```diff
@@ -0,0 +1,77 @@
...
+                    bool& continueUpdate) override;
+  /** Get the (acculumated) difficulty map for the last processed proof node */
+  void getDifficultyMap(std::map<Node, Node>& dmap) const;
+
+ private:
+  /** Reference to the env class */
+  Env& d_env;
+  /**
+   * The current difficulty of the assertion whose proof of preprocessing
+   * we are considering.
```
HanielB, line 67: This alludes to a control flow that is not clear.

src/smt/difficulty_post_processor.cpp
```diff
@@ -0,0 +1,77 @@
...
+bool DifficultyPostprocessCallback::shouldUpdate(std::shared_ptr<ProofNode> pn,
+                                                 const std::vector<Node>& fa,
+                                                 bool& continueUpdate)
+{
+  PfRule r = pn->getRule();
+  if (r == PfRule::ASSUME)
+  {
+    Trace("difficulty-debug")
+        << "  found assume: " << pn->getResult() << std::endl;
+    d_accMap[pn->getResult()] += d_currDifficulty;
```
HanielB, line 54: Seems really ugly to do this when checking whether to update. Couldn't you just move this block to an update method that does nothing else?

HanielB (changes requested):

src/smt/difficulty_post_processor.h
ajreynol: I'd like to keep this flexible.  This is part of a pipeline that goes out to the API, where you'd need to be able to convert these Node -> Term.

src/smt/difficulty_post_processor.h
ajreynol: Because this is passed back to the Solver as part of the (locally public) interface of SmtEngine, which shouldn't have to worry about CD data structures.

src/smt/difficulty_post_processor.h
ajreynol: This class assigns a constant difficulty to all free assumptions in the given proof, since it processes the proof of each preprocessed assertion independently. It is the proof manager that ties this together. I'll make this clear in the doc.

src/smt/difficulty_post_processor.h
ajreynol: I've removed it.

src/smt/difficulty_post_processor.h
HanielB: Cool, thanks.

src/smt/difficulty_post_processor.cpp
ajreynol: This is also unclear since there is no updates.
Note the final proof callback also has this style: https://github.com/cvc5/cvc5/blob/master/src/smt/proof_final_callback.cpp#L56
Maybe `shouldUpdate` needs a more general name.

HanielB: Indeed. When I did something like that for letifying a proof node I had a similar situation but with `update` being a no-op for the proof node that updated internal information. 

I like `shouldUpdate` as is, but what we want is for updating only the internal state of the callback, which feels like a new method to be called in the pipeline between `shouldUpdate` and `update`.  What do you think?

ajreynol: I'm actually just wondering if we should have a ProofVisitor class, which would be a simpler interface.

HanielB: Yes! That's the best solution. We already have three use cases, thus it's more than justified.

ajreynol: Ok, let's do this in a later PR.

HanielB (approved):

PR #7152, opened 2021-09-08
Add option for using bound inference for relevant assertions

ajreynol:
The method for discarding assertions based on bound inference is quite expensive (40% of runtime) for some benchmarks with quantified formulas from Facebook.  This adds an option to disable this, which is done by default for quantified logics.

### DIFF
--- a/src/theory/arith/nl/nonlinear_extension.cpp
+++ b/src/theory/arith/nl/nonlinear_extension.cpp
@@ -140,7 +140,8 @@ void NonlinearExtension::getAssertions(std::vector<Node>& assertions)
       // not relevant, skip
       continue;
     }
-    if (bounds.add(lit, false))
+    // if using the bound inference utility
+    if (options().arith.nlRlvAssertBounds && bounds.add(lit, false))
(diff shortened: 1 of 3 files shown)


src/theory/arith/nl/nonlinear_extension.cpp
```diff
@@ -140,7 +140,8 @@ void NonlinearExtension::getAssertions(std::vector<Node>& assertions)
       // not relevant, skip
       continue;
     }
-    if (bounds.add(lit, false))
+    // if using the bound inference utility
+    if (options().arith.nlRlvAssertBounds && bounds.add(lit, false))
```
nafur, line 144: Just to make sure: this already uses the new EnvObj infrastructure?

nafur (approved):
One minor comment, otherwise LGTM.

src/theory/arith/nl/nonlinear_extension.cpp
ajreynol: Yes

PR #7153, opened 2021-09-08
Improve pre-skolemization, move quantifiers preprocess to own file

ajreynol:
This also heavily refactors the preskolemization method (now in QuantifiersPreprocess), in preparation for it being enabled by default.  This method previously was doing a tree traversal, it now maintains a visited cache.

It makes minor cleanup to the dependencies of this method.


src/theory/quantifiers/quant_util.h
```diff
@@ -78,8 +78,18 @@ class QuantPhaseReq
   std::map< Node, bool > d_phase_reqs_equality;
   std::map< Node, Node > d_phase_reqs_equality_term;
 
-  static void getPolarity( Node n, int child, bool hasPol, bool pol, bool& newHasPol, bool& newPol );
-  static void getEntailPolarity( Node n, int child, bool hasPol, bool pol, bool& newHasPol, bool& newPol );
+  static void getPolarity(Node n,
```
HanielB, line 81: document

src/theory/quantifiers/quant_util.h
```diff
@@ -78,8 +78,18 @@ class QuantPhaseReq
...
+                          size_t child,
+                          bool hasPol,
+                          bool pol,
+                          bool& newHasPol,
+                          bool& newPol);
+  static void getEntailPolarity(Node n,
```
HanielB, line 87: document

src/theory/quantifiers/quantifiers_preprocess.h
```diff
@@ -0,0 +1,77 @@
...
+ */
+class QuantifiersPreprocess : protected EnvObj
+{
+ public:
+  QuantifiersPreprocess(Env& env);
+  /** preprocess
+   *
+   * This returns the result of applying simple quantifiers-specific
+   * preprocessing to n, including but not limited to:
+   * - pre-skolemization,
```
HanielB, line 41: be consistent with this term with/without the hyphen (above is without)

src/theory/quantifiers/quantifiers_preprocess.h
```diff
@@ -0,0 +1,77 @@
...
+ * depends on knowing the polarity of the position in which quantifiers occur.
+ */
+class QuantifiersPreprocess : protected EnvObj
+{
+ public:
+  QuantifiersPreprocess(Env& env);
+  /** preprocess
+   *
+   * This returns the result of applying simple quantifiers-specific
+   * preprocessing to n, including but not limited to:
```
HanielB, line 40: I think it'd be helpful to mention in the documentation all the transformations that are applied to the formula

src/theory/quantifiers/quantifiers_preprocess.h
```diff
@@ -0,0 +1,77 @@
...
+   * @param n The formula to preskolemize.
+   * @param polarity The polarity of n in the input.
+   * @param fvs The free variables
+   */
+  Node preSkolemizeQuantifiers(
+      Node n,
+      bool polarity,
+      std::vector<TNode>& fvs,
+      std::unordered_map<std::pair<Node, bool>, Node, NodePolPairHashFunction>&
+          visited) const;
```
HanielB, line 66: add `visited` to the doc as well

src/theory/quantifiers/quantifiers_preprocess.h
```diff
@@ -0,0 +1,77 @@
...
+  /**
+   * Apply prenexing aggressively. Returns the prenex normal form of n.
+   */
+  Node computePrenexAgg(Node n, std::map<Node, Node>& visited) const;
+}; /* class QuantifiersRewriter */
```
HanielB, line 71: update

src/theory/quantifiers/quantifiers_preprocess.cpp
```diff
@@ -0,0 +1,267 @@
...
+      Node sub;
+      std::vector<unsigned> sub_vars;
+      // return skolemized body
+      ret = Skolemize::mkSkolemizedBody(n, nn, fvs, sk, sub, sub_vars);
+    }
+    visited[key] = ret;
+    return ret;
+  }
+  // check if it contains a quantifier as a subterm
+  // if so, we will write this node
```
HanielB, line 184: write?

HanielB (changes requested):

src/theory/quantifiers/quantifiers_preprocess.h
ajreynol: Isnt that what the document is already listing?

HanielB: I read "not limited to" as the listed ones being a strict subset of all the transformation applied.

ajreynol: "not limited to" is so that this documentation is more maintainable: the description of this method should not be responsible for describing all of the methods it contains, those methods should be described more locally, e.g. in options help.

HanielB: Do we do that? The place where I'd look for what preprocessing is doing, apart from the implementation of this method itself, is the documentation in the method's signature

ajreynol: Yes, for example: 
https://github.com/cvc5/cvc5/blob/master/src/theory/bags/bags_rewriter.h#L73

HanielB: I meant adding the list to options help or something. :) But regardless, This is a more general issue that should not block this PR. I'll approve.

HanielB (approved):

src/theory/quantifiers/quantifiers_preprocess.h
ajreynol: Oh, yeah, there are a lot of quantifiers options that should be better categorized, this is a larger TODO.

PR #7154, opened 2021-09-08
Use C++17 attributes

4tXJ7f:
Currently, we are using non-standard attributes. With C++17, all the
attributes that we use regularly are now standardized. This commit
replaces the compiler-specific attributes with standard ones.

### DIFF
--- a/src/base/check.h
+++ b/src/base/check.h
@@ -54,14 +54,7 @@
 #define CVC5_PREDICT_TRUE(x) x
 #endif
 
-#ifdef __has_cpp_attribute
-#if __has_cpp_attribute(fallthrough)
 #define CVC5_FALLTHROUGH [[fallthrough]]
-#endif // __has_cpp_attribute(fallthrough)
-#endif // __has_cpp_attribute
-#ifndef CVC5_FALLTHROUGH
-#define CVC5_FALLTHROUGH
-#endif
 
 namespace cvc5 {
 
--- a/src/base/check.h
+++ b/src/base/check.h
@@ -39,23 +39,6 @@
 #include "base/exception.h"
 #include "cvc5_export.h"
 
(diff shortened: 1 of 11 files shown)


src/base/check.h
```diff
@@ -54,14 +54,7 @@
 #define CVC5_PREDICT_TRUE(x) x
 #endif
 
-#ifdef __has_cpp_attribute
-#if __has_cpp_attribute(fallthrough)
 #define CVC5_FALLTHROUGH [[fallthrough]]
```
nafur, line 57: Why is this in `check.h`?

src/theory/theory.h
```diff
@@ -440,18 +440,12 @@ class Theory : protected EnvObj
     EFFORT_LAST_CALL = 200
   }; /* enum Effort */
 
-  static inline bool standardEffortOrMore(Effort e) CVC5_CONST_FUNCTION
-  {
-    return e >= EFFORT_STANDARD;
-  }
-  static inline bool standardEffortOnly(Effort e) CVC5_CONST_FUNCTION
+  static bool standardEffortOrMore(Effort e) { return e >= EFFORT_STANDARD; }
+  static bool standardEffortOnly(Effort e)
```
nafur, line 444: Maybe this should be `constexpr` now?

src/base/check.h
4tXJ7f: Yeah, I was wondering about that, too. Moved.

nafur (approved):
LGTM.

src/theory/theory.h
4tXJ7f: Quick note: We have discussed this offline and came to the conclusion that `constexpr` does not apply here (and that the two methods `standardEffortOrMore()` and `standardEffortOnly()` are not really used).

PR #7155, opened 2021-09-08
Remove deprecated SyGuS method evaluateWithUnfolding

ajreynol:
This was a predecessor to the evaluator and TermDbSygus::rewriteNode that is no longer necessary.

This refactors the code so that evaluateWithUnfolding is replaced by rewriteNode as necessary throughout the SyGuS solver.

Note that in a few places, the extended rewriter was being used where TermDbSygus::rewriteNode (which also evaluates recursive function definitions) should have been used.


4tXJ7f (approved):
LGTM

PR #7160, opened 2021-09-09
More refactoring of set defaults

ajreynol:
This moves a large portion of the `finalizeLogic` method to more appropriate places.

It also fixes an issue : `opts.datatypes.dtSharedSelectorsWasSetByUser` was checked with wrong polarity, making a previous commit not effective.


4tXJ7f (approved):
LGTM

PR #7161, opened 2021-09-09
Connect difficulty manager to TheoryEngine

ajreynol:
This also introduces the `produceDifficulty` option which is analogous to `produceUnsatCores`.

It requires another unsat cores mode `PP_ONLY`, indicating that we are only tracking proofs of preprocessing.  This option should perhaps be renamed to `proofMode` instead of `unsatCoresMode`, since its use is more general than for unsat cores.  This will be addressed in a future refactoring.


src/theory/relevance_manager.h
```diff
@@ -73,7 +79,7 @@ class RelevanceManager
   typedef context::CDList<Node> NodeList;
 
  public:
-  RelevanceManager(context::UserContext* userContext, Valuation val);
+  RelevanceManager(Env& env, Valuation val);
```
nafur, line 82: Should we make `RelevanceManager` an `EnvObj`?

ajreynol: Actually this is unecessary for now, I'll revert this.

HanielB:
Regarding the option, an issue with that change is that there would be unsat-core-specific stuff in it, like whether we are using assumptions or not in the SAT solver. Maybe we should have two different options, one about how proofs are being generated and another about how unsat cores are being generated. A disadvantage with that is that it'd add more checks to `set_defaults`.

src/theory/relevance_manager.h
```diff
@@ -157,6 +169,17 @@ class RelevanceManager
    * aborts and indicates that all literals are relevant.
    */
   bool d_success;
+  /** Are we tracking the sources of why a literal is relevant */
+  bool d_trackRSetExp;
+  /** Miniscope top-level AND */
```
HanielB, line 174: clarify

src/theory/relevance_manager.cpp
```diff
@@ -64,8 +78,10 @@ void RelevanceManager::addAssertionsInternal(std::vector<Node>& toProcess)
   while (i < toProcess.size())
   {
     Node a = toProcess[i];
-    if (a.getKind() == AND)
+    if (d_miniscopeTopLevel && a.getKind() == AND)
     {
+      // difficulty tracking requires splitting top-level AND earlier
```
HanielB, line 83: clarify

HanielB (changes requested):

HanielB (approved):

PR #7166, opened 2021-09-10
Use StatisticsRegistry from Env

nafur:
This PR better integrates the StatisticsRegistry with the environment. It makes the registry an `EnvObj` itself and adds a getter to `EnvObj` to get the registry. It also refactors parts of cvc5 to use this new mechanism to obtain the registry instead of using the (global, static) `smtStatisticsRegistry()` function.


4tXJ7f (approved):
LGTM

PR #7172, opened 2021-09-10
bv: Use EnvObj::rewrite() and EnvObj::options() in NodeBitblaster.

mpreiner:
This PR is based on #7171

### DIFF
--- a/src/theory/bv/bitblast/node_bitblaster.h
+++ b/src/theory/bv/bitblast/node_bitblaster.h
@@ -28,12 +28,12 @@ namespace bv {
  *
  * Implements the bare minimum to bit-blast bit-vector atoms/terms.
  */
-class NodeBitblaster : public TBitblaster<Node>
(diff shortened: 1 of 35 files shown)


src/theory/bv/bitblast/node_bitblaster.h
```diff
@@ -28,12 +28,12 @@ namespace bv {
  *
  * Implements the bare minimum to bit-blast bit-vector atoms/terms.
  */
-class NodeBitblaster : public TBitblaster<Node>
+class NodeBitblaster : public TBitblaster<Node>, protected EnvObj
```
aniemetz, line 31: I would prefer that `TBitblaster` derives from `EnvObj`, but we discussed this offline and the conclusion is that `TBitblaster` will become obsolete soon, so this is fine.

aniemetz (approved):

nafur (approved):
LGTM

PR #7173, opened 2021-09-10
Add getNodeManager() to EnvObj.

mpreiner:
This also replaces NodeManager::currentNM() occurrences in TheoryEngine
to getNodeManager().

### DIFF
--- a/src/theory/theory_engine.cpp
+++ b/src/theory/theory_engine.cpp
@@ -84,24 +84,6 @@ namespace theory {
 
 /* -------------------------------------------------------------------------- */
 
-inline void flattenAnd(Node n, std::vector<TNode>& out){
-  Assert(n.getKind() == kind::AND);
-  for(Node::iterator i=n.begin(), i_end=n.end(); i != i_end; ++i){
-    Node curr = *i;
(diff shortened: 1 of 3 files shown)


src/theory/theory_engine.cpp
```diff
@@ -1916,7 +1899,9 @@ std::pair<bool, Node> TheoryEngine::entailmentCheck(options::TheoryOfMode mode,
         }
         std::pair<bool, Node> chres2 = entailmentCheck(mode, ch2);
         if( chres2.first ){
-          return std::pair<bool, Node>(true, NodeManager::currentNM()->mkNode(kind::AND, chres.second, chres2.second));
+          return std::pair<bool, Node>(
+              true,
+              getNodeManager()->mkNode(kind::AND, chres.second, chres2.second));
```
aniemetz, line 1904: Maybe retrieve the node manager before the for loop.

ajreynol:
As discussed on Zulip, I'd like to have a larger discussion about the long term plans for NodeManager first.

mpreiner:
Per offline-discussion we decided that NodeManager will be a thread_local utility that doesn't need to be associated with an Env object.

PR #7174, opened 2021-09-10
Remove context getters from `TheoryState`

4tXJ7f:
This removes `TheoryState::getSatContext()` and
`TheoryState::getUserContext()`.


src/theory/quantifiers/inst_match_trie.h
```diff
@@ -131,7 +132,8 @@ class CDInstMatchTrie
    * equalities in the equality engine of qs.
    * It additionally takes a context c, for which the entry is valid in.
    */
-  bool existsInstMatch(QuantifiersState& qs,
+  bool existsInstMatch(Env& env,
```
ajreynol, line 135: Can you pass user context instead of Env here?  I'm actually planning to simplify this method so that is furthermore does not depend on `qs` either.

ajreynol (changes requested):
One minor request, otherwise LGTM.

Thanks a lot for doing this.

src/theory/quantifiers/inst_match_trie.h
4tXJ7f: Makes sense, done!

PR #7177, opened 2021-09-10
checkModel: Extend documentation.

aniemetz:
Add comment that partial operators are one reason that the current
assertion cannot be checked.


4tXJ7f (approved):
LGTM

PR #7179, opened 2021-09-10
bv: Move IsPowerOfTwo rule to preprocessing pass and use EnvObj.

mpreiner:
IsPowerOfTwo rule is specific to the BvIntroPow2 preprocessing pass and should not be a general bit-vector rewrite rule.


src/preprocessing/passes/bv_intro_pow2.cpp
```diff
@@ -33,11 +32,85 @@ namespace passes {
...
+BvIntroPow2::BvIntroPow2(PreprocessingPassContext* preprocContext)
+    : PreprocessingPass(preprocContext, "bv-intro-pow2"){};
 
-Node pow2Rewrite(Node node, NodeMap& cache)
+PreprocessingPassResult BvIntroPow2::applyInternal(
+    AssertionPipeline* assertionsToPreprocess)
 {
-  NodeMap::const_iterator ci = cache.find(node);
+  std::unordered_map<Node, Node> cache;
+  for (unsigned i = 0, size = assertionsToPreprocess->size(); i < size; ++i)
```
aniemetz, line 42:
```suggestion
  for (size_t i = 0, size = assertionsToPreprocess->size(); i < size; ++i)
```

src/preprocessing/passes/bv_intro_pow2.cpp
```diff
@@ -33,11 +32,85 @@ namespace passes {
...
+/**
+ * x & (x-1) = 0 => 1 << sk
+ *
+ * WARNING: this is an **EQUISATISFIABLE** transformation!
+ * Only to be called on top level assertions.
+ *
+ * @param node
+ *
+ * @return
+ */
```
aniemetz, line 64: Move to header.

aniemetz (approved):

PR #7186, opened 2021-09-13
Alethe: Added Callback Function to alethe_proof_processor

Lachnitt:
Added alethe_proof_processor header file and introduced callback class.

Should be merged after PR #7185.

### DIFF
--- a/src/proof/alethe/alethe_post_processor.h
+++ b/src/proof/alethe/alethe_post_processor.h
@@ -16,8 +16,8 @@
 #ifndef CVC4__PROOF__ALETHE_PROOF_PROCESSOR_H
 #define CVC4__PROOF__ALETHE_PROOF_PROCESSOR_H
 
-#include "proof/proof_node_updater.h"
 #include "proof/alethe/alethe_proof_rule.h"
+#include "proof/proof_node_updater.h"
 
 namespace cvc5 {
 
@@ -97,10 +97,10 @@ class AletheProofPostprocessCallback : public ProofNodeUpdaterCallback
    * @return True if the step could be added, or false if not.
    */
   bool addAletheStepFromOr(Node res,
-                          AletheRule rule,
-                          const std::vector<Node>& children,
-                          const std::vector<Node>& args,
-                          CDProof& cdp);
+                           AletheRule rule,
+                           const std::vector<Node>& children,
+                           const std::vector<Node>& args,
+                           CDProof& cdp);
 };
 
 }  // namespace proof
--- a/src/proof/alethe/alethe_post_processor.h
+++ b/src/proof/alethe/alethe_post_processor.h
@@ -59,8 +59,10 @@ class AletheProofPostprocessCallback : public ProofNodeUpdaterCallback
    **/
   Node d_cl;
   /**
-   * This method adds a new step to the proof applying rule but adds a
-   * conclusion different from the result as the third argument.
+   * This method adds a new ALETHE_RULE step to the proof, with `rule` as the
+   * first argument, the original conclusion `res` as the second and
+   * `conclusion`, the result to be printed (which may or may not differ from
+   * `res`), as the third.
    *
    * @param rule The id of the Alethe rule,
    * @param res The expected result of the application,
--- a/src/proof/alethe/alethe_post_processor.h
+++ b/src/proof/alethe/alethe_post_processor.h
@@ -79,10 +79,11 @@ class AletheProofPostprocessCallback : public ProofNodeUpdaterCallback
                      const std::vector<Node>& args,
                      CDProof& cdp);
   /**
-   * This method adds a new step (or F1 ... Fn) to the proof applying rule while
-   * printing a version where the outermost or is replaced by cl, i.e. (cl F1
-   * ... Fn). For this it internally calls addAletheStep. The kind of the given
-   * Node has to be OR.
+   * As above, but for proof nodes with original conclusions of the form `(or F1
+   * ... Fn)` whose conclusion-to-be-printed must be `(cl F1 ... Fn)`.
+   *
+   * This method internally calls addAletheStep. The kind of the given Node has
+   * to be OR.
    *
    * @param res The expected result of the application in form (or F1 ... Fn),
    * @param rule The id of the Alethe rule,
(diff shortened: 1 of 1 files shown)


src/proof/alethe/alethe_post_processor.h
```diff
@@ -0,0 +1,129 @@
...
+ public:
+  AletheProofPostprocessCallback(ProofNodeManager* pnm);
+  ~AletheProofPostprocessCallback() {}
+  /** Should proof pn be updated? Only if its top-level proof rule is not an
+   *  Alethe proof rule.
+   *
+   * @param pn the proof node that maybe should be updated
+   * @param continueUpdate indicates whether we should continue recursively
+   * updating pn
+   * @return whether we should run the update method on pn
```
HanielB, line 41: you can remove this since this is the same as in the extended class.

src/proof/alethe/alethe_post_processor.h
```diff
@@ -0,0 +1,129 @@
...
+   */
+  bool shouldUpdate(std::shared_ptr<ProofNode> pn,
+                    const std::vector<Node>& fa,
+                    bool& continueUpdate) override;
+  /**
+   * This method updates the proof rule application by splitting on the given
```
HanielB, line 47: Clarify "splitting"

src/proof/alethe/alethe_post_processor.h
```diff
@@ -0,0 +1,129 @@
...
+   * rule and translating it into a proof node in terms of the Alethe rules.
+   *
+   * @param res The expected result of the application,
+   * @param rule The id of the Alethe rule,
+   * @param children The children of the application,
+   * @param args The arguments of the application,
+   * @param cdp The proof to add to,
+   * @return True if the step could be added, or null if not.
```
HanielB, line 55: as above, remove

src/proof/alethe/alethe_post_processor.h
```diff
@@ -0,0 +1,129 @@
...
+              const std::vector<Node>& children,
+              const std::vector<Node>& args,
+              CDProof* cdp,
+              bool& continueUpdate) override;
+
+ private:
+  /** The proof node manager */
+  ProofNodeManager* d_pnm;
+  /** The node manager */
+  NodeManager* d_nm;
```
HanielB, line 68: Delete this argument. Instead have, at every place where it would be used, the pointer to the current node manager retrieved.

src/proof/alethe/alethe_post_processor.h
```diff
@@ -0,0 +1,129 @@
...
+  /** The variable cl **/
```
HanielB, line 69: Give more details about it.

src/proof/alethe/alethe_post_processor.h
```diff
@@ -0,0 +1,129 @@
...
+   *
+   * @param res The expected result of the application,
+   * @param rule The id of the Alethe rule,
+   * @param conclusion The conclusion of the application as the Alethe printer
+   * @param children The children of the application,
+   * @param args The arguments of the application
+   * @param cdp The proof to add to
+   * @return True if the step could be added, or false if not.
+   */
+  bool addAletheStep(Node res,
```
HanielB, line 100: You can delete the version above if everytime you use it you just pass the same node as the first and third arguments. Also it'd be clear if had `rule` as the first argument, then `res` then `conclusion`.

src/proof/alethe/alethe_post_processor.h
```diff
@@ -0,0 +1,129 @@
...
+                    const std::vector<Node>& children,
+                    const std::vector<Node>& args,
+                    CDProof& cdp);
+  /**
+   * This method adds a new step to the proof applying the ALETHE_RULE but adds
+   * a conclusion different from the result as the third argument.
+   *
+   * @param res The expected result of the application,
+   * @param rule The id of the Alethe rule,
+   * @param conclusion The conclusion of the application as the Alethe printer
```
HanielB, line 94:
```suggestion
   * @param conclusion The conclusion to be printed for the step
```

src/proof/alethe/alethe_post_processor.h
```diff
@@ -0,0 +1,129 @@
...
+   * given Node has to be OR.
+   *
+   * @param res The expected result of the application in form (or F1 ... Fn),
+   * @param rule The id of the Alethe rule,
+   * @param children The children of the application,
+   * @param args The arguments of the application
+   * @param cdp The proof to add to
+   * @return True if the step could be added, or false if not.
+   */
+  bool addAletheStepFromOr(Node res,
```
HanielB, line 118: Do I understand correctly that this method is taking `(or F1 ... Fn)` and effectively calling the version above with `(or F1 ... Fn)` as the first argument and `(cl F1 ... Fn)` as the third? 

I think it's fine to keep this method, but explain it in terms of the above method to have things more clear.

HanielB (changes requested):

src/proof/alethe/alethe_post_processor.h
Lachnitt: Done

src/proof/alethe/alethe_post_processor.h
Lachnitt: Replaced by depending.

src/proof/alethe/alethe_post_processor.h
Lachnitt: This function is not present in the extended class? Should the comment still be removed?

src/proof/alethe/alethe_post_processor.h
Lachnitt: Added more extensive comment. 
For every step the conclusion is a clause. But since the or operator requires at least two arguments it is extended by the cl operator. In case of more than one argument it corresponds to or otherwise it is the identity.

src/proof/alethe/alethe_post_processor.h
Lachnitt: Yes, it is overloaded for more convenient use but it can be substituted without any problems. Done.

src/proof/alethe/alethe_post_processor.h
Lachnitt: Yes, exactly. I will make this clearer in the comments.

src/proof/alethe/alethe_post_processor.h
Lachnitt: Done.

src/proof/alethe/alethe_post_processor.h
HanielB: Sorry, I meant to remove just from lines (now) 44 to 50. The method should remain.

HanielB (changes requested):
Just a few more comments. Also you should clang-format the whole file, as there are still a few places not formatted correctly.

HanielB (approved):

PR #7189, opened 2021-09-13
Towards standard usage of evaluator

ajreynol:
This makes the evaluator accessible via EnvObj through the Rewriter.  It furthermore removes `Rewriter::rewrite` from inside the evaluator itself.

Construction of Evaluator utilities is now discouraged. 

The include dependencies were cleaned slightly in this PR, leading to more precise includes throughout.

This is work towards having a configurable cardinality for strings, as well as eliminating SmtEngineScope.


src/theory/datatypes/sygus_extension.cpp
```diff
@@ -1101,16 +1102,20 @@ Node SygusExtension::registerSearchValue(Node a,
...
-          if (its == d_sampler[a].end())
+          std::map<TypeNode, std::unique_ptr<quantifiers::SygusSampler>>& smap =
+              d_sampler[a];
+          std::map<TypeNode,
+                   std::unique_ptr<quantifiers::SygusSampler>>::iterator its =
+              smap.find(tn);
+          if (its == smap.end())
           {
-            d_sampler[a][tn].initializeSygus(
+            smap[tn].reset(new quantifiers::SygusSampler(d_env.getRewriter()));
```
4tXJ7f, line 1112: It is not necessary to change this to a `std::unique_ptr`, I think. We can potentially do:

```
smap[tn].emplace(tn, d_env.getRewriter());
```

src/theory/builtin/proof_checker.cpp
```diff
@@ -349,8 +349,8 @@ Node BuiltinProofRuleChecker::checkInternal(PfRule id,
     if (res1 != res2)
     {
       // can rewrite the witness forms
-      res1 = Rewriter::rewrite(SkolemManager::getOriginalForm(res1));
-      res2 = Rewriter::rewrite(SkolemManager::getOriginalForm(res2));
+      res1 = d_env.getRewriter()->rewrite(SkolemManager::getOriginalForm(res1));
+      res2 = d_env.getRewriter()->rewrite(SkolemManager::getOriginalForm(res2));
```
4tXJ7f, line 353: Why not make this proof checker an `EnvObj`?

4tXJ7f (changes requested):
Looks good, I just have some minor questions around the design.

src/theory/builtin/proof_checker.cpp
ajreynol: Proof checkers, by their nature, should have as few dependencies as possible, thus we don't want to make them EnvObj.

src/theory/rewriter.h
```diff
@@ -70,6 +73,27 @@ class Rewriter {
...
+  Node evaluate(TNode n,
+                const std::vector<Node>& args,
+                const std::vector<Node>& vals,
+                bool useRewriter) const;
+  /** Same as above, with a visited cache. */
+  Node evaluate(TNode n,
+                const std::vector<Node>& args,
+                const std::vector<Node>& vals,
+                const std::unordered_map<Node, Node>& visited,
+                bool useRewriter = true) const;
```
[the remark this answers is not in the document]
ajreynol: I think this is a pretty good suggestion.  I believe the right place for `rewriteViaMethod` would be `Env`.

src/theory/datatypes/sygus_extension.cpp
ajreynol: Actually, i believe this does not work.  The second argument to emplace is the value of the range, but I don't think it can determine that this is the argument to the constructor of the range. Think if the sampler had multiple arguments in its constructor.

PR #7190, opened 2021-09-13
Utilities in preparation for print benchmark utility

ajreynol:
This adds a few utilities in preparation for the print benchmark utility, which will replace our own dumping infrastructure.

### DIFF
--- a/src/expr/dtype.h
+++ b/src/expr/dtype.h
@@ -425,6 +425,12 @@ class DType
   const std::vector<std::shared_ptr<DTypeConstructor> >& getConstructors()
       const;
 
+  /**
+   * Return the subfield types of this datatype. This is the set of all types T
+   * for which there exists an argument to a constructor of type T.
+   */
+  std::unordered_set<TypeNode> getSubfieldTypes() const;
+
   /** prints this datatype to stream */
   void toStream(std::ostream& out) const;
 
--- a/src/printer/printer.h
+++ b/src/printer/printer.h
@@ -87,6 +87,8 @@ class Printer
   virtual void toStreamCmdDeclareFunction(std::ostream& out,
                                           const std::string& id,
                                           TypeNode type) const;
+  /** Variant of above for a pre-existing variable */
+  void toStreamCmdDeclareFunction(std::ostream& out, const Node& v) const;
   /** Print declare-pool command */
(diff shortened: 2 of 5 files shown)


src/expr/dtype.h
```diff
@@ -425,6 +425,12 @@ class DType
   const std::vector<std::shared_ptr<DTypeConstructor> >& getConstructors()
       const;
 
+  /**
+   * Return the subfield types of this datatype. This is the set of all types T
+   * for which there exists an argument to a constructor of type T.
+   */
+  std::unordered_set<TypeNode> getSubfieldTypes() const;
```
abdoo8080, line 432: With this function, do we still need `cvc5::theory::quantifiers::SygusTypeInfo::getSubfieldTypes`?

src/printer/printer.h
```diff
@@ -87,6 +87,8 @@ class Printer
   virtual void toStreamCmdDeclareFunction(std::ostream& out,
                                           const std::string& id,
                                           TypeNode type) const;
+  /** Variant of above for a pre-existing variable */
+  void toStreamCmdDeclareFunction(std::ostream& out, const Node& v) const;
```
abdoo8080, line 91: I wonder if there is any need for the original methods if these variants are more convenient.

abdoo8080 (commented):
Looks good! Just 2 questions.

src/expr/dtype.h
ajreynol: There are 2 reasons:  `cvc5::theory::quantifiers::SygusTypeInfo::getSubfieldTypes` is (1) cached for efficiency, and (2) is slightly different: it is recursive and only follows "sygus" datatypes.

src/printer/printer.h
ajreynol: It's possible, I'll check.

ajreynol: There is one instance where this command is called now (via Command::toStream).

However, the reason why we want the original is because it is easier to implement: the derived printer does not need to know how to extract the name from the variable.

Regardless, I've updated Command to use the new version.

src/expr/dtype.h
abdoo8080: Fair enough!

abdoo8080 (approved):
LGTM!

PR #7191, opened 2021-09-13
Reimplement `--dump=raw-benchmark` as `-o raw-benchmark`

ajreynol:
Printing the original benchmark is simple, as it is exactly the commands we execute.

This removes the previous code from SmtEngine, which is currently broken.


src/smt/smt_engine.cpp
```diff
@@ -680,8 +668,7 @@ void SmtEngine::defineFunctionsRec(
       lem = nm->mkNode(kind::FORALL, boundVars, lem, aexpr);
     }
     // assert the quantified formula
-    //   notice we don't call assertFormula directly, since this would
-    //   duplicate the output on raw-benchmark.
+    //   notice we don't call assertFormula directly.
```
nafur, line 671: Is there another reason to not call `assertFormula` directly now?

nafur (commented):
`--dump raw-benchmark` is also used in `datatype-dump.cvc` and `run_regression.py`

src/smt/smt_engine.cpp
ajreynol: Haha I thought you'd ask, the general reason is that we shouldn't use the SmtEngine interface from itself.

nafur (approved):
LGTM.

PR #7192, opened 2021-09-14
Eliminate global access to options:: from quantifiers rewriter

ajreynol:


src/theory/quantifiers/quantifiers_rewriter.h
```diff
@@ -267,27 +273,18 @@ class QuantifiersRewriter : public TheoryRewriter
...
+                     bool pol,
+                     bool prenexAgg) const;
+  Node computeSplit(std::vector<Node>& args, Node body, QAttributes& qa) const;
 
-private:
-  /** options */
- static bool doOperation(Node f, RewriteStep computeOption, QAttributes& qa);
+ public:
+  RewriteResponse preRewrite(TNode in) override;
+  RewriteResponse postRewrite(TNode in) override;
```
mpreiner, line 286: Can you make sure that there is only one `public:` and one `private:` block?

ajreynol: Yes, although it will be a big diff, I'll do this now.

mpreiner (approved):

PR #7194, opened 2021-09-14
Add get-difficulty to the API

ajreynol:
Adds smt2 parsing, printing and API support for `get-difficulty`.  Adds unit tests.


src/main/command_executor.cpp
```diff
@@ -174,6 +174,12 @@ bool CommandExecutor::doCommandSingleton(Command* cmd)
       getterCommands.emplace_back(new GetUnsatCoreCommand());
     }
 
+    if (d_solver->getOptionInfo("dump-difficulty").boolValue()
+        && (isResultUnsat || res.isSat() || res.isSatUnknown()))
```
HanielB, line 178:
```suggestion
        && (isResultUnsat || isResultSat || res.isSatUnknown()))
```

HanielB (changes requested):

src/main/command_executor.cpp
ajreynol: Good catch

HanielB (approved):

PR #7195, opened 2021-09-14
Make `-o raw-benchmark` work with `--parse-only`

4tXJ7f:
Before, cvc5 was returning with `--parse-only` before the code could
reach the code responsible for dumping the raw benchmark. This moves the
check for `--parse-only` to the appropriate place and updates the
`run_regression.py` script to use `--parse-only`.

### DIFF
--- a/test/regress/run_regression.py
+++ b/test/regress/run_regression.py
@@ -141,7 +141,7 @@ def run_benchmark(dump, wrapper, scrubber, error_scrubber, cvc5_binary,
     exit_status = None
     if dump:
         dump_args = [
-            '--parse-only', '--output-lang=smt2'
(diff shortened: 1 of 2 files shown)


test/regress/run_regression.py
```diff
@@ -141,8 +141,7 @@ def run_benchmark(dump, wrapper, scrubber, error_scrubber, cvc5_binary,
     exit_status = None
     if dump:
         dump_args = [
-            '--preprocess-only', '-o', 'raw-benchmark',
-            '--output-lang=smt2', '-qq'
```
mpreiner, line 145: Isn't this missing `-o raw-benchmark` now?

mpreiner (approved):

PR #7197, opened 2021-09-15
Remove options that are no longer used

nafur:
This PR removes a handful of options that are no longer used anywhere.


ajreynol:
Awesome, thanks for doing this.

ajreynol (approved):

PR #7198, opened 2021-09-15
Fix relevant domain for parametric operators

### LINKED ISSUE #6531: (full-saturate-quant) Assertion error at src/theory/quantifiers/term_tuple_enumerator.cpp:295
cvc5 b016f603f2a3709faaf
~~~~
(declare-datatypes ((a 0)) (((b))))
(assert (forall ((e (Seq a))) (seq.nth (seq.extract (seq.++ (seq.unit false) (seq.unit false)) 1 (seq.len (ite (= 0 (seq.len e)) seq.empty (seq.unit false)))) 0)))
(check-sat)
~~~~
~~~~
cvc5 -q -i --check-models delta.out.smt2
Fatal failure within virtual void cvc5::theory::quantifiers::TermTupleEnumeratorBase::next(std::vector<cvc5::NodeTemplate<true> >&) at /home/cvc5/src/theory/quantifiers/term_tuple_enumerator.cpp:295
Check failure

 terms[variableIx].isNull() || terms[variableIx].getType().isComparableTo( d_quantifier[0][variableIx].getType())

Aborted
~~~~

ajreynol:
Fixes #6531.

This issue also occurs when using `--full-saturate-quant` on facebook benchmarks that combine multiple sequence types.

It does some cleanup on relevant domain in the process.

### DIFF
--- a/src/theory/quantifiers/relevant_domain.cpp
+++ b/src/theory/quantifiers/relevant_domain.cpp
@@ -85,8 +85,15 @@ RelevantDomain::RelevantDomain(Env& env,
 }
 
 RelevantDomain::~RelevantDomain() {
-  for( std::map< Node, std::map< size_t, RDomain * > >::iterator itr = d_rel_doms.begin(); itr != d_rel_doms.end(); ++itr ){
-    for( std::map< size_t, RDomain * >::iterator itr2 = itr->second.begin(); itr2 != itr->second.end(); ++itr2 ){
+  for (std::map<Node, std::map<size_t, RDomain*> >::iterator itr =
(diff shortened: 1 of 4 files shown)


4tXJ7f (changes requested):
LGTM, but can we add a regression test before merging?

ajreynol:
I don't know any benchmark that is solved that had this error.  Those that have this error are "unknown".  I think it'd be hard to craft a benchmark that would be solved that otherwise would have had this error.

4tXJ7f (approved):
Sounds good, let's get this in then.

PR #7199, opened 2021-09-15
Refactor tag suggestion mechanism

nafur:
This PR refactors the internal tag suggestion mechanism to no longer rely on C-style char arrays, but use a C++ vector of strings instead.

### DIFF
--- a/src/base/configuration.cpp
+++ b/src/base/configuration.cpp
@@ -255,11 +255,8 @@ const std::vector<std::string>& Configuration::getDebugTags()
 
 bool Configuration::isDebugTag(const std::string& tag)
 {
-#if defined(CVC5_DEBUG) && defined(CVC5_TRACING)
   return std::find(Debug_tags.begin(), Debug_tags.end(), tag)
          != Debug_tags.end();
-#endif /* CVC5_DEBUG && CVC5_TRACING */
-  return false;
 }
 
 const std::vector<std::string>& Configuration::getTraceTags()
@@ -269,11 +266,8 @@ const std::vector<std::string>& Configuration::getTraceTags()
 
 bool Configuration::isTraceTag(const std::string& tag)
 {
-#if CVC5_TRACING
   return std::find(Trace_tags.begin(), Trace_tags.end(), tag)
          != Trace_tags.end();
-#endif /* CVC5_TRACING */
-  return false;
 }
 
 bool Configuration::isGitBuild() {
--- a/src/base/configuration.cpp
+++ b/src/base/configuration.cpp
@@ -25,13 +25,8 @@
 #include "base/configuration_private.h"
 #include "base/cvc5config.h"
 
-#if defined(CVC5_DEBUG) && defined(CVC5_TRACING)
-#  include "base/Debug_tags.h"
-#endif /* CVC5_DEBUG && CVC5_TRACING */
-
-#ifdef CVC5_TRACING
-#  include "base/Trace_tags.h"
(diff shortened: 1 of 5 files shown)


src/base/configuration.cpp
```diff
@@ -253,68 +254,38 @@ bool Configuration::isBuiltWithPoly()
...
-#if defined(CVC5_DEBUG) && defined(CVC5_TRACING)
-  /* -1 because a NULL pointer is inserted as the last value */
-  return (sizeof(Debug_tags) / sizeof(Debug_tags[0])) - 1;
-#else  /* CVC5_DEBUG && CVC5_TRACING */
-  return 0;
-#endif /* CVC5_DEBUG && CVC5_TRACING */
-}
-
-char const* const* Configuration::getDebugTags() {
+std::vector<std::string> Configuration::getDebugTags()
```
4tXJ7f, line 257: Can we return a `const std::vector<std::string>&` instead and get rid of the `#if` below? Isn't the default an empty vector?

src/base/configuration.cpp
```diff
@@ -253,68 +254,38 @@ bool Configuration::isBuiltWithPoly()
...
-#if CVC5_TRACING
-  /* -1 because a NULL pointer is inserted as the last value */
-  return sizeof(Trace_tags) / sizeof(Trace_tags[0]) - 1;
-#else  /* CVC5_TRACING */
-  return 0;
-#endif /* CVC5_TRACING */
-}
-
-char const* const* Configuration::getTraceTags() {
+std::vector<std::string> Configuration::getTraceTags()
```
4tXJ7f, line 275: Ditto (same as for debug tags).

src/base/configuration.cpp
```diff
@@ -253,68 +254,38 @@ bool Configuration::isBuiltWithPoly()
...
-  unsigned ntags = getNumDebugTags();
-  char const* const* tags = getDebugTags();
-  for (unsigned i = 0; i < ntags; ++ i) {
-    if (strcmp(tag, tags[i]) == 0) {
-      return true;
-    }
-  }
+  return std::find(Debug_tags.begin(), Debug_tags.end(), tag)
+         != Debug_tags.end();
 #endif /* CVC5_DEBUG && CVC5_TRACING */
```
4tXJ7f, line 271: No need for the `#if` either, I think.

src/base/configuration.cpp
```diff
@@ -253,68 +254,38 @@ bool Configuration::isBuiltWithPoly()
...
-  char const* const* tags = getTraceTags();
-  for (unsigned i = 0; i < ntags; ++ i) {
-    if (strcmp(tag, tags[i]) == 0) {
-      return true;
-    }
-  }
+  return std::find(Trace_tags.begin(), Trace_tags.end(), tag)
+         != Trace_tags.end();
 #endif /* CVC5_TRACING */
   return false;
```
4tXJ7f, line 290: Ditto (same as for debug tags).

4tXJ7f (changes requested):
Just some suggestions for further simplifications/better performance.

src/base/configuration.cpp
nafur: Right now, we always generate the full vector in `collect_tags.py` and the whole `#ifdef` stuff is done so that this vector does not end up in the binary if it is not necessary. We could push the `#ifdef` stuff into the generated headers (`Debug_tags.h`)...

4tXJ7f (approved):
LGTM, thanks for the updates.

PR #7201, opened 2021-09-15
Clean options handlers

nafur:
Some cleanup on the option handlers, starting with handlers for base and main options.


src/options/options_handler.cpp
```diff
@@ -393,114 +400,199 @@ void OptionsHandler::showDebugTags(const std::string& option,
...
+  {
+    throw OptionException("trace tags not available in non-tracing build");
+  }
+  printTags(Configuration::getNumTraceTags(), Configuration::getTraceTags());
+  std::exit(0);
+}
+
+// theory/quantifiers/options_handlers.h
+
+void OptionsHandler::checkInstWhenMode(const std::string& option,
```
ajreynol, line 420: You can delete this, delete the enum value PRE_FULL.

ajreynol (approved):

PR #7204, opened 2021-09-16
Use a single `NodeManager` per thread

4tXJ7f:
This changes cvc5 to use a single `NodeManager` per thread (using
`thread_local`). We have decided that this is more convenient because
nodes between solvers in the same thread could be exchanged and that
there isn't really an advantage of having multiple `NodeManager`s per
thread.

One wrinkle of this change is that `NodeManager::init()` must be called
explicitly before the `NodeManager` can be used. This code can currently
not be moved to the constructor of `NodeManager` because the code
indirectly calls `NodeManager::currentNM()`, which leads to a loop
because the `NodeManager` is created in `NodeManager::currentNM()`.
Further refactoring is required to get rid of this restriction.

### DIFF
--- a/src/expr/node_manager.cpp
+++ b/src/expr/node_manager.cpp
@@ -41,8 +41,6 @@ using namespace cvc5::expr;
 
 namespace cvc5 {
 
-thread_local NodeManager* NodeManager::s_current = NULL;
-
 namespace {
 
 /**
@@ -97,6 +95,7 @@ typedef expr::Attribute<attr::LambdaBoundVarListTag, Node> LambdaBoundVarListAtt
 NodeManager::NodeManager()
     : d_skManager(new SkolemManager),
       d_bvManager(new BoundVarManager),
+      d_initialized(false),
       next_id(0),
       d_attrManager(new expr::attr::AttributeManager()),
       d_nodeUnderDeletion(nullptr),
@@ -104,7 +103,12 @@ NodeManager::NodeManager()
       d_abstractValueCount(0),
       d_skolemCounter(0)
 {
-  init();
+}
+
+NodeManager* NodeManager::currentNM()
+{
+  thread_local static NodeManager nm;
+  return &nm;
 }
 
 bool NodeManager::isNAryKind(Kind k)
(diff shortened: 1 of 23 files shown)


src/expr/node_manager.cpp
```diff
@@ -97,14 +95,20 @@ typedef expr::Attribute<attr::LambdaBoundVarListTag, Node> LambdaBoundVarListAtt
...
       d_inReclaimZombies(false),
       d_abstractValueCount(0),
       d_skolemCounter(0)
 {
-  init();
+}
+
+NodeManager* NodeManager::currentNM()
+{
+  thread_local static NodeManager nm;
```
ajreynol, line 110: Just a minor question since I'm not 100% sure how this thread_local stuff works: the change of moving this to the cpp should have no difference from before?

ajreynol (approved):
LGTM, this is great.

src/expr/node_manager.cpp
mpreiner: That's exactly what we want, each thread will have its own instance of `nm`. Whether this is in the header or the cpp file doesn't make a difference.

ajreynol: Sure, I was just wondering the motivation for the move.  This is better style I assume?

4tXJ7f: Yep, the motivation for moving it to the source file was purely a style decision and doesn't have any effect on how `thread_local` works.

4tXJ7f:
I ran on 20k randomly chosen benchmarks from SMT-LIB and the overall results show no performance difference (120s timeout, master = master after this PR, nms = master before this PR):

```
                   status  total  solved   sat  unsat   best  timeout  memout  error  uniq  disagr  time_cpu    memory 
config                                                                                                                 
2021-09-17-nms         ee  20000   17215  7484   9731  15807     2588       0     54     4       0  351854.3 1207718.0 
2021-09-17-master/     ee  20000   17215  7482   9733  15642     2587       0     54     4       0  351882.1 1207707.8 
```

PR #7205, opened 2021-09-16
Make cegqi subsolvers EnvObj

ajreynol:
Makes a few places do multiplication for Rational directly instead of invoking the rewriter.


src/theory/quantifiers/cegqi/inst_strategy_cegqi.cpp
```diff
@@ -505,7 +505,7 @@ CegInstantiator * InstStrategyCegqi::getInstantiator( Node q ) {
   std::map<Node, std::unique_ptr<CegInstantiator>>::iterator it =
       d_cinst.find(q);
   if( it==d_cinst.end() ){
-    d_cinst[q].reset(new CegInstantiator(q, d_qstate, d_treg, this));
+    d_cinst[q].reset(new CegInstantiator(d_env, q, d_qstate, d_treg, this));
```
nafur, line 508: A more general comment: can we have a constructor `EnvObj(EnvObj&)`?
Here, we could then say `CegInstantiator(*this)`, but we could also have utilities that construct some objects that inherit from `EnvObj` given another `EnvObj`, but without access to the `env` themselves...

mpreiner: Sounds reasonable to me.

mpreiner (approved):

src/theory/quantifiers/cegqi/inst_strategy_cegqi.cpp
ajreynol: We could do that, however in general I'd prefer getting rid of all circular references.  The general principle I think we should follow is that the `this` pointer should never be passed to a constructor of a submodule, which is what is happening here.

PR #7211, opened 2021-09-18
Alethe: Add helper function to remove attributes

Lachnitt:
This function is later used to remove nodes with the kind INST_PATTERN and INST_PATTERN_LIST from a node.

Should this be renamed?

### DIFF
--- a/src/proof/alethe/alethe_post_processor.cpp
+++ b/src/proof/alethe/alethe_post_processor.cpp
@@ -24,8 +24,8 @@ namespace cvc5 {
 
 namespace proof {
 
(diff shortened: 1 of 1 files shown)


HanielB (changes requested):
This function applies a tree-traversal, which means that it is visiting a node as a tree rather than as a DAG. In the latter, repeated nodes would be visited only once (you'd need a cache to tell you which nodes had already been visited). Instead of defining such a method yourself, it's best to use a `NodeConverter`, which transforms a node into another according to a transformation you can define.

HanielB:
Closing since this PR has been preempted by #7220

PR #7212, opened 2021-09-18
Alethe: Implementation of AletheProofPostprocessCallback

Lachnitt:
Implementation of addAletheStep and addAletheStepFromOr. Added stub for AletheProofPostprocessCallback update function that will be populated by subsequent PRs. 

Should be merged after PR #7211.

Co-authored-by: Haniel Barbosa <hanielbbarbosa@gmail.com>

PR #7213, opened 2021-09-18
Alethe: Translate ASSUME rule

Lachnitt:
Implementation of the translation of ASSUME rules into the Alethe calculus.

Should be merged after PR #7212.

### DIFF
--- a/src/proof/alethe/alethe_post_processor.cpp
+++ b/src/proof/alethe/alethe_post_processor.cpp
@@ -24,8 +24,8 @@ namespace cvc5 {
 
 namespace proof {
 
-// This function removes all attributes contained in the list of attributes from
-// a Node res while only recursively updating the node further if
+// This function removes all attributes contained in a given list of attributes
+// from a Node res while only recursively updating the node further if
 // continueRemoval is true.
 static Node removeAttributes(Node res,
                              const std::vector<Kind>& attributes,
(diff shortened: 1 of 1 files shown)


src/proof/alethe/alethe_post_processor.cpp
```diff
@@ -0,0 +1,160 @@
...
+    // Children: none
+    // Arguments: (F)
+    // --------------
+    // Conclusion: F
+    //
+    // proof rule: assume
+    // proof node: (VP:F)
+    // proof term: F
+    // premises: ()
+    // args: ()
```
HanielB, line 97: I'd prefer to delete this. I understand it's helpful during development, but now that `addAletheStep` is well documented this conversion is pretty straightforward without the above comment.

HanielB (changes requested):

src/proof/alethe/alethe_post_processor.cpp
Lachnitt: I would vote to keep these comments since, e.g. in the example of SCOPE, it might be confusing to have an artificially constructed proof node that is created to print the same proof node differently. Also sometimes the proof node and term are hard to read in the code.

I will still delete the comments.

HanielB: I agree that there are cases (like for #7224) where it's certainly helpful to have such comments, but they can be written in a way where it's clearer what's going on, as judgments, like here: https://github.com/cvc5/cvc5/blob/c07d80a0ce4dc7144bad7146b0dc96574dd250c9/src/theory/uf/eq_proof.h#L107

HanielB:
You should merge master and fix the conflicts. With that I think the PR is good to go.

HanielB (approved):

PR #7217, opened 2021-09-20
arrays: Use EnvObj::rewrite and EnvObj::options.

aniemetz:
This does not yet clean up the usage of Rewriter::rewrite in the arrays
type enumerator yet. Will be cleaned up in a follow-up PR.


ajreynol (approved):
LGTM, we should eventually try to see if we can avoid passing Rewriter* to TheoryArrayRewriter, but this is a good step for now.

PR #7219, opened 2021-09-20
Remove CVC language support

mpreiner:
This commit removes the support for the CVC language and converts all *.cvc regression tests to SMT-LIBv2.


src/expr/dtype.cpp
```diff
@@ -897,8 +897,6 @@ const std::vector<std::shared_ptr<DTypeConstructor> >& DType::getConstructors()
 
 std::ostream& operator<<(std::ostream& os, const DType& dt)
 {
-  // can only output datatypes in the cvc5 native language
-  language::SetLanguage::Scope ls(os, Language::LANG_CVC);
```
nafur, line 901: I assume the comment was incorrect already?

mpreiner: We can print datatypes in SMT-LIBv2. Not sure when this comment got obsolete. @ajreynol?

nafur (commented):
I didn't look at the individual regressions, but the rest LGTM.
One general question: `checkEntailed` is apparently only used on tptp input now. We should consider getting rid of it entirely, or have some regressions that exercise it...

src/expr/dtype.cpp
ajreynol: This comment is extremely obsolete.

ajreynol (approved):
LGTM, this is great.

PR #7221, opened 2021-09-20
Remove usage of static options in arithmetic theory

nafur:
This PR removes the usage of static accesses to options from the arithmetic theory, mostly by making more classes inherit from `EnvObj`. The single exception is the `proof_macros` header that will be addressed in a later PR.


src/theory/arith/fc_simplex.cpp
```diff
@@ -340,7 +341,7 @@ UpdateInfo FCSimplexDecisionProcedure::selectPrimalUpdate(ArithVar basic, Linear
     }
   }
 
-  CompPenaltyColLength colCmp(&d_linEq);
+  CompPenaltyColLength colCmp(&d_linEq, options::havePenalties());
```
ajreynol, line 344: This still needs to be eliminated.

src/theory/arith/pp_rewrite_eq.cpp
```diff
@@ -49,13 +50,16 @@ TrustNode PreprocessRewriteEq::ppRewriteEq(TNode atom)
...
         rewritten,
-        d_pnm->mkNode(
+        d_env.getProofNodeManager()->mkNode(
             PfRule::THEORY_INFERENCE, {}, {atom.eqNode(rewritten), t}));
   }
   return TrustNode::mkTrustRewrite(atom, rewritten, nullptr);
 }
 
-bool PreprocessRewriteEq::proofsEnabled() const { return d_pnm != nullptr; }
+bool PreprocessRewriteEq::proofsEnabled() const
```
ajreynol, line 59: Let's inline/eliminate this method.

ajreynol (changes requested):
A few minor requests.

ajreynol (approved):

PR #7223, opened 2021-09-21
Eliminate Output macro in favor of simple Env functions

nafur:
This PR eliminates the `Output` macro together with the associated output stream (stored in a static variable). It is replaces by a set of simple utility functions of the `Env` class, and we instead use the output stream from `options.base.out`.
To achieve this, a couple of quantifier classes are now derived from `EnvObj`.


ajreynol:
Don't we have the same issues with `Trace` and `Debug`?  I'm wondering if there is a better long term solution that doesn't make the use of `Output(tag)` verbose.  Something along the lines of "Output/Debug/Trace is not specific to the solver instance".

I think this needs more discussion.

nafur:
> Don't we have the same issues with `Trace` and `Debug`? I'm wondering if there is a better long term solution that doesn't make the use of `Output(tag)` verbose. Something along the lines of "Output/Debug/Trace is not specific to the solver instance".
> 
> I think this needs more discussion.

So there are multiple things here:
- to my understanding, these macros are necessarily tied to the solver instance, as there behaviour depends on the currently set options. As long as we can enable/disable trace, debug and output tags via the command line, we need to somehow move this knowledge into the macro in one way or another.
- right now this is not an issue for `Trace` and `Debug`, because they store the set of enabled tags statically (in some sense alongside the macros) whereas `Output` stores them in the options. While we should also do something about `Trace` and `Debug`, they currently work very different from `Output`.
- finally, I think we could legitimately argue that `Trace` and `Debug` work in a somewhat weird way in that they behave "incorrectly" when multiple solvers are present: after all, they are debugging tools for the developers. `Output` on the other hand is only used in a few places, and thus a slightly more verbose calling convention should not be that bad.

With all that being said: yes, we should discuss this

mpreiner:
I don't think that making Output more verbose is an issue here since this won't be used in many places compared to Trace/Debug. I'd be find with having a global thread_local storage for enabled trace/debug tags.

ajreynol:
I've thought about this a bit more, and I think we are in agreement.  One thing to note: I think it is important that the semantics of `Output` is indeed that it is specific to the solver instance.  This means that if a user set `-o trigger`, they would be notified about triggers that were selected in the current solver instance and *not* in any subsolvers that were spawned.  This also suggests that our calls to subsolvers e.g. for checking unsat cores should disable all the Output tags.  Otherwise, we are probably double printing.

I agree that making Output more verbose is ok, but Debug/Trace is not.  I also agree with @mpreiner that some thread local storage for Debug/Trace may be the way to go.

ajreynol:
One further thought: we should probably have a utility method in Env for getting the `ostream` from the tag that does the pattern which is being repeated.

src/theory/quantifiers/sygus/synth_conjecture.cpp
```diff
@@ -359,7 +359,7 @@ bool SynthConjecture::doCheck()
     }
   }
 
-  bool printDebug = Output.isOn(options::OutputTag::SYGUS);
+  bool printDebug = Output.isOn(options(), options::OutputTag::SYGUS);
```
ajreynol, line 362: I think this one can also be cleaned up?

ajreynol (changes requested):
This is starting to look really nice!

ajreynol (approved):
Awesome, LGTM.

PR #7224, opened 2021-09-21
Alethe: Translate SCOPE rule

Lachnitt:
Implementation of the translation of SCOPE rules into the Alethe calculus.

Should be merged after PR #7213.


src/proof/alethe/alethe_post_processor.cpp
```diff
@@ -60,6 +60,177 @@ bool AletheProofPostprocessCallback::update(Node res,
...
+    // remains the same.
+    //
+    // Let (not (and F1 ... Fn))^i denote the repetition of (not (and F1 ...
+    // Fn)) for i times.
+    //
+    // T1:
+    //
+    //   P
+    // ----- ANCHOR    ------- ... ------- AND_POS
+    //  VP1             VP2_1  ...  VP2_n
```
HanielB, line 76: A question: how are you handling the fact that you'd need to introduce an anchor step marking the beginning of the subproof? Is this handled in the printer?

src/proof/alethe/alethe_post_processor.cpp
```diff
@@ -60,6 +60,177 @@ bool AletheProofPostprocessCallback::update(Node res,
...
+    //
+    //
+    // *  the corresponding proof node is (not (and F1 ... Fn))
+    // ** the corresponding proof node is (=> (and F1 ... Fn) F)
+    case PfRule::SCOPE:
+    {
+      bool success = true;
+
+      // Build vp1
+      std::vector<Node> negNode;
```
HanielB, line 129:
```suggestion
      std::vector<Node> negNode{d_cl};
```

src/proof/alethe/alethe_post_processor.cpp
```diff
@@ -60,6 +60,177 @@ bool AletheProofPostprocessCallback::update(Node res,
...
+      std::vector<Node> sanitized_args;
+      for (Node arg : args)
+      {
+        negNode.push_back(arg.notNode());  // (not F1) ... (not Fn)
+        sanitized_args.push_back(d_anc.convert(arg));
+      }
+      negNode.push_back(children[0]);         // (not F1) ... (not Fn) F
+      negNode.insert(negNode.begin(), d_cl);  // (cl (not F1) ... (not F) F)
```
HanielB, line 137:
```suggestion
```

src/proof/alethe/alethe_post_processor.cpp
```diff
@@ -60,6 +60,177 @@ bool AletheProofPostprocessCallback::update(Node res,
...
+
+      // Build vp1
+      std::vector<Node> negNode;
+      std::vector<Node> sanitized_args;
+      for (Node arg : args)
+      {
+        negNode.push_back(arg.notNode());  // (not F1) ... (not Fn)
+        sanitized_args.push_back(d_anc.convert(arg));
+      }
+      negNode.push_back(children[0]);         // (not F1) ... (not Fn) F
```
HanielB, line 136:
```suggestion
      negNode.push_back(children[0]);         // (cl (not F1) ... (not Fn) F)
```

src/proof/alethe/alethe_post_processor.cpp
```diff
@@ -60,6 +60,177 @@ bool AletheProofPostprocessCallback::update(Node res,
     {
       return addAletheStep(AletheRule::ASSUME, res, res, children, {}, *cdp);
     }
+    // See proof_rule.h for documentation on the SCOPE rule. This comment uses
+    // variable names as introduced there. Instead of (not (and F1 ... Fn))
+    // (cl (not (and F1 ... Fn))) is printed and instead of (=> (and F1 ... Fn)
+    // F) the term (cl (=> (and F1 ... Fn))) is printed while the proof node
+    // remains the same.
```
HanielB, line 67:
```suggestion
    // variable names as introduced there. If the SCOPE conclusion is (not (and
    // F1 ... Fn)), the transformation below generates (cl (not (and F1
    // ... Fn))) to be printed, while for the conclusion (=> (and F1 ... Fn) F)
    // it generates (cl (=> (and F1 ... Fn))).
```

src/proof/alethe/alethe_post_processor.cpp
```diff
@@ -60,6 +60,177 @@ bool AletheProofPostprocessCallback::update(Node res,
...
+    // VP2a: (cl F (not (and F1 ... Fn))^n)
+    // VP2b: (cl (not (and F1 ... Fn))^n F)
+    // VP3: (cl (not (and F1 ... Fn)) F)
+    // VP4: (cl (=> (and F1 ... Fn) F) (and F1 ... Fn)))
+    // VP5: (cl (=> (and F1 ... Fn) F) F)
+    // VP6: (cl (=> (and F1 ... Fn) F) (not F))
+    // VP7: (cl (=> (and F1 ... Fn) F) (=> (and F1 ... Fn) F))
+    //
+    // The reorder step is not necessary if n = 1, because in that case VP2a is
+    // (cl (not F1 F) which is the same as VP2b.
```
HanielB, line 99:
```suggestion
    // Note that if n = 1, then the ANCHOR step yields (cl (not F1) F), which is
    // the same as VP3. Since VP1 = VP3, the steps for that transformation are
    // not generated.
```

HanielB (changes requested):
Made some changes to the comments, PTAL

src/proof/alethe/alethe_post_processor.cpp
Lachnitt: Yes, in src/proof/alethe/alethe_proof_rule.h the ANCHOR_SUBPROOF rule is added as a special rule. This indicates to the printer that it should print a subproof.

src/proof/alethe/alethe_post_processor.cpp
```diff
@@ -60,6 +60,177 @@ bool AletheProofPostprocessCallback::update(Node res,
...
+        andNode = nm->mkNode(kind::AND, args);  // (and F1 ... Fn)
+      }
+      else
+      {
+        andNode = args[0];  // F1
+      }
+      std::vector<Node> premisesVP2 = {vp1};
+      std::vector<Node> notAnd = {d_cl, children[0]};  // cl F
+      Node vp2_i;
+      for (long unsigned int i = 0; i < args.size(); i++)
```
HanielB, line 159:
```suggestion
      for (size_t i = 0, size = args.size(); i < size; i++)
```

src/proof/alethe/alethe_post_processor.cpp
```diff
@@ -60,6 +60,177 @@ bool AletheProofPostprocessCallback::update(Node res,
     {
       return addAletheStep(AletheRule::ASSUME, res, res, children, {}, *cdp);
     }
+    // See proof_rule.h for documentation on the SCOPE rule. This comment uses
+    // variable names as introduced there. If the SCOPE conclusion is (not (and
+    // F1 ... Fn)), the transformation below generates (cl (not (and F1
+    // ... Fn))) to be printed, while for the conclusion (=> (and F1 ... Fn) F)
+    // it generates (cl (=> (and F1 ... Fn))).
```
HanielB, line 67:
```suggestion
    // See proof_rule.h for documentation on the SCOPE rule. This comment uses
    // variable names as introduced there. Since the SCOPE rule originally concludes
    // (=> (and F1 ... Fn) F) or (not (and F1 ... Fn)) but the ANCHOR rule
    // concludes (cl (not F1) ... (not Fn) F), to keep the original shape of the
    // proof node it is necessary to rederive the original conclusion. The
    // transformation is described below, depending on the form of SCOPE's
    // conclusion.
    //
    // Note that after the original conclusion is rederived the new proof node
    // will actually have to printed, respectively, (cl (=> (and F1 ... Fn) F))
    // or (cl (not (and F1 ... Fn))).
```

I reread the comment and I noticed that it might still be a bit confusing, so I tried to clarify. PTAL

src/proof/alethe/alethe_post_processor.cpp
```diff
@@ -60,6 +60,177 @@ bool AletheProofPostprocessCallback::update(Node res,
...
+
+      notAnd.erase(notAnd.begin() + 1);  //(cl (not (and F1 ... Fn))^n)
+      notAnd.push_back(children[0]);     //(cl (not (and F1 ... Fn))^n F)
+      Node vp2b = nm->mkNode(kind::SEXPR, notAnd);
+      success &=
+          addAletheStep(AletheRule::REORDER, vp2b, vp2b, {vp2a}, {}, *cdp);
+
+      Node vp3 = nm->mkNode(kind::SEXPR, d_cl, andNode.notNode(), children[0]);
+      success &= addAletheStep(
+          AletheRule::DUPLICATED_LITERALS, vp3, vp3, {vp2b}, {}, *cdp);
```
HanielB, line 180: All of this should be guarded by whether `n = 1`, no? Since in that case you can directly say that `vp3` is the same as `vp1`, and otherwise you do the computation that is happening here.

HanielB (changes requested):

src/proof/alethe/alethe_post_processor.cpp
Lachnitt: Sounds good! The first part of course applies to a lot of proof rules, so I will copy it to wherever it is needed.

src/proof/alethe/alethe_post_processor.cpp
Lachnitt: I did that once and it seemed to make no difference, because the proof was not updated in that case.

Lachnitt: Done

HanielB (approved):

PR #7225, opened 2021-09-21
Update java examples using the new Java API

mudathirmahgoub:
This PRs updates java examples using the new Java API, by converting C++ examples to Java. 
Examples CVC4Streams.java and PipedInput.java are removed since they are not longer supported by the API. 
All examples are not included in the build which would be added in a future PR.


mudathirmahgoub:
This PR depends on PR  https://github.com/cvc5/cvc5/pull/6396

examples/api/java/Combination.java
```diff
@@ -10,90 +10,119 @@
...
+    System.out.println("slv.getValue(" + p_f_y + "): " + slv.getValue(p_f_y) + "\n");
+
+    System.out.println("Alternatively, iterate over assertions and call slv.getValue(...) "
+        + "on all terms.");
+    prefixPrintGetValue(slv, assertions);
+
+    System.out.println();
+    System.out.println("You can also use nested loops to iterate over terms.");
+    Iterator<Term> it1 = assertions.iterator();
+    while (it1.hasNext())
```
nafur, line 107: Can we do `for (Term it1: assertions)` here?

examples/api/java/StatisticsExample.java
```diff
@@ -0,0 +1,231 @@
+/******************************************************************************
+ * Top contributors (to current version):
+ *   Andres Noetzli
```
nafur, line 3: Please call it `Statistics.java`, no other example has `Example` in the filename.

examples/api/java/Combination.java
mudathirmahgoub: Yes, which is written few lines later. 
https://github.com/cvc5/cvc5/pull/7225/files#diff-84065c47299f1c133669b2a4c677efbf630b733dc7ec4071368d778256ba517dR119

examples/api/java/StatisticsExample.java
mudathirmahgoub: Done. I renamed it to StatisticsExample.java since there is a class already named Statistics. 
The package name is used to avoid this conflict.

nafur (approved):
LGTM.

PR #7226, opened 2021-09-21
Eliminate arithmetic proof macros

nafur:
This PR eliminates some macros for arithmetic proofs, that only slightly simplified access to the `produce-proofs` option. Now that static access is deprecated, these checks needed to be implemented differently anyway.


alex-ozdemir (approved):
Looks good to me!

PR #7227, opened 2021-09-21
Fix stod issue with  quickstart example

mudathirmahgoub:
stod function does not fail with strings like "1/6", and only returns the numerator (see the output below). 
The condition `if (xMinusYComputed == xMinusYDouble)` does not capture the error because both xDouble and yDouble 
have the same value 1. 
This PR fixes the issue by using `getReal64Value` which returns a pair , and then convert them to double. 

```
value for x: 1/6
value for y: 1/6
value for x - y: 0.0
value for x: 1
value for y: 1
value for x - y: 0
computed correctly
```

### DIFF
--- a/examples/api/cpp/quickstart.cpp
+++ b/examples/api/cpp/quickstart.cpp
@@ -114,11 +114,18 @@ int main()
   std::cout << "value for y: " << yStr << std::endl;
   std::cout << "value for x - y: " << xMinusYStr << std::endl;
 
-  // Further, we can convert the values to cpp types,
-  // using standard cpp conversion functions.
-  double xDouble = std::stod(xStr);
-  double yDouble = std::stod(yStr);
-  double xMinusYDouble = std::stod(xMinusYStr);
+  // Further, we can convert the values to cpp types
+  std::pair<int64_t, uint64_t> xPair = xVal.getReal64Value();
+  std::pair<int64_t, uint64_t> yPair = yVal.getReal64Value();
+  std::pair<int64_t, uint64_t> xMinusYPair = xMinusYVal.getReal64Value();
+  double xDouble = xPair.first / static_cast<double>(xPair.second);
+  double yDouble = yPair.first / static_cast<double>(yPair.second);
+  double xMinusYDouble =
+      xMinusYPair.first / static_cast<double>(xMinusYPair.second);
+
+  std::cout << "value for x: " << xDouble << std::endl;
+  std::cout << "value for y: " << yDouble << std::endl;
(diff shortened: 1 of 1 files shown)


yoni206 (changes requested):
Looks good, thanks for finding this!
Could you please also update the line numbers in the docs accordingly?
This would be [here](https://github.com/cvc5/cvc5/blob/master/docs/api/cpp/quickstart.rst), starting from the part about this conversion, and then updating the subsequent changed line numbers.

mudathirmahgoub:
> Looks good, thanks for finding this!
> Could you please also update the line numbers in the docs accordingly?
> This would be [here](https://github.com/cvc5/cvc5/blob/master/docs/api/cpp/quickstart.rst), starting from the part about this conversion, and then updating the subsequent changed line numbers.

Done. Using line numbers in the documentation is just not maintainable. In the future it is just easier to copy the code.

yoni206:
Thanks! I checked how it looks [here](https://cvc5.github.io/docs-ci/docs-pr7227/api/cpp/quickstart.html) and it looks right.

Regarding line numbers, I agree it is hard to maintain, but hopefully these docs should not change too often, as the API will be stable. Copying the code would be equally difficult, I think.

One last thing I am not sure about is whether this is indeed the intended way to get real values from the API. I'd be happy to know what @nafur thinks of this, in particular:
is this the way you would get doubles our of strings using the API?

block starts here:
https://github.com/cvc5/cvc5/blob/9708f1b361494deb0e777545e5c5240dc9a81182/examples/api/cpp/quickstart.cpp#L117

block ends here: https://github.com/cvc5/cvc5/blob/9708f1b361494deb0e777545e5c5240dc9a81182/examples/api/cpp/quickstart.cpp#L128

nafur:
Okay, we should go a step back here. We should generally not convert to double. The code may print `computed correctly`, but there is really no way to tell. `double` variables should never be compared to equality.

My suggestion would be to print all values as fractions (after all, they are rational values) and perform a proper rational comparison:
```
std::pair<int64_t, uint64_t> xMinusYComputed = { xPair.first * yPair.second - yPair.first * xPair.second, xPair.second * yPair.second };
if (xMinusYComputed == xMinusYDouble) ...
```

mudathirmahgoub:
I agree with @nafur although the user should be aware of precision loss with conversion from rational to double. 
I suggest removing `getRealValue` from the example and only use `getReal64Value` to print rationals:  
```cpp
 std::cout << "value for x: " << xPair.second << "/" xPair.second << std::endl;
 ``` 
 @yoni206  can you take over this PR?

PR #7218, opened 2021-09-20
Improve implementation of parser for tuple project

ajreynol:
This PR refactors how special cases of indexed operators are parsed.

PR #7228, opened 2021-09-22
Minimal fixing version for tuple update parsing

ajreynol:
This takes the essential changes from https://github.com/cvc5/cvc5/pull/7218 so that the current ANTLR issues are avoided.


mudathirmahgoub (approved):
LGTM

PR #7229, opened 2021-09-22
Add extensionality option for strings disequalities

ajreynol:
Towards the strings array solver. 

Extensionality for disequalities is disabled by default, but will be used when the strings array solver is enabled.


4tXJ7f (approved):
LGTM, we should have regressions that use this option at some point.

PR #7231, opened 2021-09-22
Alethe: Add Alethe Files to be Compiled

Lachnitt:
Adds Alethe proof rule and option. Adds alethe_post_processor and alethe_proof_rule files to list of files to be compiled.

Should be merged after PR #7224.


HanielB (approved):
LGTM

PR #7233, opened 2021-09-22
a quick fix on solver_black unit test

Ouyancheng:
In file `test/unit/api/solver_black.cpp` line 1499, 
`for (const std::pair<Term, Term>& t : dmap)` is not the correct way of iterating through the element pairs, 
it should be `for (const std::pair<const Term, Term>& t : dmap)` as the keys are immutable. 
This triggers a warning on LLVM clang 12.0.1 (not AppleClang) on macOS.


4tXJ7f (approved):
LGTM, thanks for the fix.

Ouyancheng:
> LGTM, thanks for the fix.

No problem!

PR #7236, opened 2021-09-23
Alethe: Translate THEORY_REWRITE

Lachnitt:
Implementation of the translation of THEORY_REWRITE rules into the Alethe calculus.

Should be merged after PR #7231.


src/proof/alethe/alethe_post_processor.cpp
```diff
@@ -60,6 +60,335 @@ bool AletheProofPostprocessCallback::update(Node res,
...
+                                 *cdp);
+      }
+
+      return success;
+    }
+    // The rule is translated according to the theory id tid and the outermost
+    // connective of the conclusion F. This is not an exact translation but
+    // should work in most cases.
+    //
+    // E.g. if the F: (= (* 0 d) 0) and tid = THEORY_ARITH, then prod_simplify
```
HanielB, line 238:
```suggestion
    // E.g. if F is (= (* 0 d) 0) and tid = THEORY_ARITH, then prod_simplify
```

src/proof/alethe/alethe_post_processor.cpp
```diff
@@ -60,6 +60,335 @@ bool AletheProofPostprocessCallback::update(Node res,
...
+                                 {vp8, vp10},
+                                 {},
+                                 *cdp);
+      }
+
+      return success;
+    }
+    // The rule is translated according to the theory id tid and the outermost
+    // connective of the conclusion F. This is not an exact translation but
+    // should work in most cases.
```
HanielB, line 236:
```suggestion
    // The rule is translated according to the theory id tid and the outermost
    // connective of the first term in the conclusion F, since F always has the
    // form (= t1 t2) where t1 is the term being rewritten. This is not an exact
    // translation but should work in most cases.
```

src/proof/alethe/alethe_post_processor.cpp
```diff
@@ -60,6 +60,335 @@ bool AletheProofPostprocessCallback::update(Node res,
...
+    //
+    // E.g. if the F: (= (* 0 d) 0) and tid = THEORY_ARITH, then prod_simplify
+    // is correctly guessed as the rule.
+    case PfRule::THEORY_REWRITE:
+    {
+      AletheRule vrule = AletheRule::UNDEFINED;
+      Node t = res[0];
+
+      switch (static_cast<theory::TheoryId>(std::stoul(args[1].toString())))
```
HanielB, line 245: Instead of this you should use the standard method to retrieve TheoryIds, `theory::builtin::BuiltinProofRuleChecker::getTheoryId`. You could do something like 

```suggestion
      theory::TheoryId tid;
      if (!theory::builtin::BuiltinProofRuleChecker::getTheoryId(args[1], tid))
      {
        return addAletheStep(
            vrule, res, nm->mkNode(kind::SEXPR, d_cl, res), children, {}, *cdp);
      }
      switch (tid)
```

HanielB (changes requested):

HanielB (approved):

PR #7239, opened 2021-09-23
Update large-model.smt2 to use regular sed syntax

dddejan:
On some older systems sed does not accept the -E option. This uses the non-extended syntax to match the digits.


4tXJ7f (changes requested):
Thanks for the fix. Could you please sign your commit? See [contribution guidelines](https://github.com/cvc5/cvc5/blob/master/CONTRIBUTING.md) for details.

PR #7240, opened 2021-09-23
Use `|` to print quoted strings in `set-info` command.

abdoo8080:
This PR is a step towards enabling `raw-benchmark` for `cvc5` regressions. `cvc5` fails to reparse 57 regressions (in `regress0`) printed be `raw-benchmark` because they contain multi-line strings in `set-info` that we don't print between vertical bars right now. This PR fixes that bug and removes 2 commands (derived from `set-info` commands) that are not used by the parsers anymore.


ajreynol (approved):
LGTM

PR #7243, opened 2021-09-23
Add Statistics and Stat to the Java API

mudathirmahgoub:
This PR adds Statistics and Stat to the Java API


src/api/java/cvc5/Stat.java
```diff
@@ -0,0 +1,169 @@
...
+   }
+
+   // endregion
+
+  /**
+   * @return a string representation of this Stat
+   */
+  protected String toString(long pointer)
+  {
+    throw new UnsupportedOperationException("Not implemented in the cpp api");
```
nafur, line 57: Can we wrap `operator<<()` here?

src/api/java/cvc5/Stat.java
```diff
@@ -0,0 +1,169 @@
...
+ */
+
+package cvc5;
+
+import java.util.Map;
+
+/**
+ * Represents a snapshot of a single statistic value.
+ * A value can be of type `int64_t`, `double`, `std::string` or a histogram
+ * (`std::map<std::string, uint64_t>`).
```
nafur, line 23: I'm not sure how you've handled this so far, but I think we should have the comment in terms of java types, not cpp types.

src/api/java/cvc5/Stat.java
```diff
@@ -0,0 +1,169 @@
...
+    return isHistogram(pointer);
+  }
+
+  private native boolean isHistogram(long pointer);
+
+  /**
+   * Return the histogram value.
+   * @return The histogram value.
+   */
+  public Map<String, Long> getHistogram()
```
nafur, line 163: It is probably not a big deal here, but technically the types don't match as the C++ api uses `uint64_t` whereas `Long` corresponds to `int64_t`. How did you handle this in other places?
Maybe, we should explicitly safeguard this cast in the JNI code?

src/api/java/cvc5/Stat.java
mudathirmahgoub: Done by using << operator in the jni. 
I just assumed it is not supported because other classes like Term, Sort have explicit toString function.

src/api/java/cvc5/Stat.java
mudathirmahgoub: Fixed. Let us say, the cpp documentation did not account for copying to Java :)

src/api/java/cvc5/Stat.java
mudathirmahgoub: I thought of using BigInteger here, but since this is a histogram, there is no precision loss here. The user can use functions like 
https://docs.oracle.com/javase/8/docs/api/java/lang/Long.html#toUnsignedString-long-

nafur (commented):
LGTM, please have a look @4tXJ7f as I don't know a lot about the whole java stuff...

4tXJ7f (approved):
LGTM

PR #7249, opened 2021-09-24
Refactor regressions script

4tXJ7f:
This makes the regression script more modular by refactoring all the
different checks into separate classes, which makes it easier to add
additional tests and to run only a subset of the tests.

### DIFF
--- a/test/regress/run_regression.py
+++ b/test/regress/run_regression.py
@@ -11,12 +11,12 @@
 # directory for licensing information.
 # #############################################################################
 ##
-
 """
 Runs benchmark and checks for correct exit status and output.
 """
 
 import argparse
+import collections
 import difflib
 import os
 import re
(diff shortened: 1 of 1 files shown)


test/regress/run_regression.py
```diff
@@ -194,38 +493,37 @@ def run_regression(check_unsat_cores, check_proofs, dump, use_skip_return_code,
...
-    elif benchmark_ext == '.p':
-        status_regex = r'% Status\s*:\s*(Theorem|Unsatisfiable|CounterSatisfiable|Satisfiable)'
-        status_to_output = lambda s: '% SZS status {} for {}'.format(
-            s, benchmark_filename)
-    elif benchmark_ext == '.sy':
-        comment_char = ';'
-        # Do not check proofs/unsat-cores with .sy files
-        check_unsat_cores = False
-        check_proofs = False
+    if benchmark_ext == ".smt":
```
abdoo8080, line 499: Do we still support old `.smt` formats?

test/regress/run_regression.py
```diff
@@ -11,36 +11,330 @@
...
+
+
+class ModelTester(Tester):
+    def __init__(self):
+        pass
+
+    def applies(self, benchmark_info):
+        expected_output_lines = benchmark_info.expected_output.split()
+        return (
+            ("sat" in expected_output_lines or "unknown" in expected_output_lines)
```
abdoo8080, line 187: I expected to see `benchmark_info.benchmark_ext != ".sy"` here and in `AbductTester.applies`. Does it make sense to run `ModelTester` and `AbductTester` on sygus benchmarks?

test/regress/run_regression.py
```diff
@@ -236,188 +534,109 @@ def run_regression(check_unsat_cores, check_proofs, dump, use_skip_return_code,
...
+            benchmark_ext=benchmark_ext,
+            benchmark_content=benchmark_content,
+            expected_output=expected_output,
+            expected_error=expected_error,
+            expected_exit_status=expected_exit_status,
+            command_line_args=all_args,
+        )
+        for tester_name, tester in g_testers.items():
+            if tester_name in testers and tester.applies(benchmark_info):
+                tests.append((tester, benchmark_info))
```
abdoo8080, line 623: 😍

abdoo8080 (changes requested):
This PR looks awesome!!! Just a few nitpicks.

test/regress/run_regression.py
```diff
@@ -11,36 +11,330 @@
...
+
+class AbductTester(Tester):
+    def __init__(self):
+        pass
+
+    def applies(self, benchmark_info):
+        return (
+            "--no-check-abducts" not in benchmark_info.command_line_args
+            and "--check-abducts" not in benchmark_info.command_line_args
+            and "get-abduct" in benchmark_info.benchmark_content
```
abdoo8080, line 231:
```suggestion
            and "--get-abduct" in benchmark_info.benchmark_content
```

test/regress/run_regression.py
4tXJ7f: Heh, good point. We don't, I'll remove.

test/regress/run_regression.py
4tXJ7f: Yeah, that's a good point. I've used the checks of the original code but I agree that it would be better to not run those on SyGuS benchmarks.

PR #7251, opened 2021-09-24
Refactor our static builds

nafur:
This PR does a major refactoring on how we organize our builds to allow both shared and static builds. We now build the libraries using object libraries to allow building the libraries both dynamically and statically in the same build folder, though the static library is optional (`ENABLE_STATIC_LIBRARY`). The binary is linked either dynamically or statically (depending on `ENABLE_STATIC_BINARY`).


src/main/CMakeLists.txt
```diff
@@ -36,16 +36,18 @@ set_source_files_properties(${libmain_gen_src_files} PROPERTIES GENERATED TRUE)
...
+if(USE_CLN)
+  target_link_libraries(main PUBLIC CLN)
 endif()
 
-# We can't use target_link_libraries(main ...) here since this is only
-# supported for cmake version >= 3.12. Hence, we have to manually add the
-# library dependencies for main. As a consequence, include directories from
-# dependencies are not propagated and we need to manually add the include
-# directories of libcvc5 to main.
-add_dependencies(main cvc5 cvc5parser gen-tokens)
```
mpreiner, line 48: Do we need to bump the cmake requirement to 3.12 now?

CMakeLists.txt
```diff
@@ -301,27 +288,12 @@ if(ENABLE_SHARED)
   # More information on RPATH in CMake:
   # https://gitlab.kitware.com/cmake/community/wikis/doc/cmake/RPATH-handling
   set(CMAKE_INSTALL_RPATH "${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_LIBDIR};${PROJECT_SOURCE_DIR}/deps/install/lib")
-else()
-  # When building statically, we *only* want static archives/libraries
-  if (WIN32)
-      set(CMAKE_FIND_LIBRARY_SUFFIXES .lib .a)
-  else()
-      set(CMAKE_FIND_LIBRARY_SUFFIXES .a)
-  endif()
```
mpreiner, line 310: We don't need this anymore?

CMakeLists.txt
```diff
@@ -301,27 +288,12 @@ if(ENABLE_SHARED)
...
-  set(BUILD_SHARED_LIBS OFF)
-  cvc5_set_option(ENABLE_STATIC_BINARY ON)
-
-  # Never build unit tests as static binaries, otherwise we'll end up with
-  # ~300MB per unit test.
-  if(ENABLE_UNIT_TESTING)
-    message(STATUS "Disabling unit tests since static build is enabled.")
-    set(ENABLE_UNIT_TESTING OFF)
-  endif()
```
mpreiner, line 319: Is it now guaranteed that we always link units tests against the shared library?

nafur: Yes. `main-test` always links against `cvc5-shared`.

CMakeLists.txt
nafur: This is a bit more complex now, as "some libraries" (gmp) need to modify it: as the static build of GMP is not build with `-fPIC` it is not suitable for dynamic linking and we thus need both `libgmp.a` and `libgmp.so`. It has been replaced by a set of macros now, that some `FindX.cmake` scripts may use.

src/main/CMakeLists.txt
nafur: To be honest, I lost track of which cmake version can build this now. My suggestion would be to simply try with a number of cmake versions to figure it out.

nafur: The current state apparently works from cmake 3.15 upwards. This would support:
- ubuntu 20.04 (not 18.04)
- debian buster and stretch (i.e. oldstable and oldoldstable) with backports enabled
- debian bullseye (stable)
- homebrew

4tXJ7f: Actually, it looks like it builds on Ubuntu 18.04 with CMake 3.10 (that's part of the CI).

nafur: the CI image uses a newer cmake: https://github.com/actions/virtual-environments/blob/ubuntu18/20210919.1/images/linux/Ubuntu1804-README.md
It actually uses cmake 3.21.

CMakeLists.txt
mpreiner: What do we do if the system provides a lib.a that is not compiled with `-fPIC`?

nafur: This is now the responsibility of the `FindX` script. Most libraries can simply export the target `X` that holds the static library and works for both static and dynamic builds (because it's build with `-fPIC`). If it is not, the `FindX` script is expected to create targets `X-shared` and `X-static`, as `FindGMP` does. Of course our usage of these targets (when linking them) needs to be aware of what type of find-script it is...

src/CMakeLists.txt
```diff
@@ -1295,63 +1295,92 @@ add_subdirectory(util)
...
+if(ENABLE_STATIC_LIBRARY)
+  add_library(cvc5-static STATIC)
+  target_link_libraries(cvc5-static PRIVATE cvc5-obj)
+  set_target_properties(cvc5-static PROPERTIES OUTPUT_NAME cvc5)
+  target_include_directories(cvc5-static
+    PUBLIC
+      $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}>
+      $<INSTALL_INTERFACE:include>
+  )
+  install(TARGETS cvc5-obj cvc5-static
```
mpreiner, line 1335: Why is `cvc5-obj` an install target?

src/main/CMakeLists.txt
```diff
@@ -36,16 +36,18 @@ set_source_files_properties(${libmain_gen_src_files} PROPERTIES GENERATED TRUE)
...
+set_target_properties(main PROPERTIES POSITION_INDEPENDENT_CODE ON)
+
+if(ENABLE_STATIC_BINARY)
+  target_link_libraries(main PUBLIC cvc5-static cvc5parser-static)
+else()
+  target_link_libraries(main PUBLIC cvc5-shared cvc5parser-shared)
+endif()
+if(USE_CLN)
+  target_link_libraries(main PUBLIC CLN)
 endif()
```
mpreiner, line 48: Do we still have to link against CLN separately here? If so, why?

cmake/FindPoly.cmake
```diff
@@ -66,8 +66,8 @@ if(NOT Poly_FOUND_SYSTEM)
     unset(patchcmd)
   endif()
 
-  get_target_property(GMP_INCLUDE_DIR GMP INTERFACE_INCLUDE_DIRECTORIES)
-  get_target_property(GMP_LIBRARY GMP IMPORTED_LOCATION)
+  get_target_property(GMP_INCLUDE_DIR GMP_SHARED INTERFACE_INCLUDE_DIRECTORIES)
+  get_target_property(GMP_LIBRARY GMP_SHARED IMPORTED_LOCATION)
```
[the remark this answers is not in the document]
nafur: This (implicitly) assumes that both GMP_STATIC and GMP_SHARED are in the same directory. Note that `FindPoly` only uses the directory, not the full filename.

src/main/CMakeLists.txt
nafur: No, and we don't need GMP either.

src/CMakeLists.txt
nafur: This is an excellent question, that I have no really good answer to.
If we remove it, cmake will complain:
```
CMake Error: install(EXPORT "cvc5-targets" ...) includes target "cvc5-static" which requires target "cvc5-obj" that is not in any export set.
```
even though, `cvc5-static` only links against `cvc5-obj` with `PRIVATE`. This is one of the quirks I had in mind when I said something to the effect of "we support static builds, but they may behave in weird ways when someone tries to install them".

What I *think* is going on here: when linking statically, you need to explicitly link against all dependencies, and their dependencies, etc. To properly reflect this in the export, all these targets must be in the export set. What I don't understand, though, is why cmake does not enforce the same for the recursive dependencies. `cvc5Targets.cmake` also contains references of `CaDiCaL` or `SymFPU`, and thus actually using `cvc5-static` would most likely fail anyway...

mpreiner: This link might help to resolve the issue: https://discourse.cmake.org/t/install-ing-a-static-library-also-requires-installing-an-internal-object-library/683/3

cmake/FindPoly.cmake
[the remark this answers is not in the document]
mpreiner: I guess we can assume this.

src/CMakeLists.txt
mpreiner: Passing cvc5-obj is expected behavior.

cmake/FindCLN.cmake
```diff
@@ -67,7 +67,7 @@ if(NOT CLN_FOUND_SYSTEM)
     BUILD_BYPRODUCTS <INSTALL_DIR>/lib/libcln.a
   )
 
-  add_dependencies(CLN-EP GMP)
+  add_dependencies(CLN-EP GMP_SHARED)
```
mpreiner, line 70: Are GMP_SHARED and GMP_STATIC always built? How about moving add_dependencies down to the `if(ENABLE_STATIC_LIBRARY)` block and either add GMP_STATIC or GMP_SHARED as dependency?

mpreiner (approved):

PR #7253, opened 2021-09-27
Add new versioning scheme

nafur:
This PR introduces a new versioning mechanism that allows for better automation.


src/options/options_handler.cpp
```diff
@@ -320,27 +320,19 @@ void OptionsHandler::showConfiguration(const std::string& option,
...
-    print_config("scm", ss.str());
-  } else {
+  if (Configuration::isGitBuild())
+  {
     print_config_cond("scm", false);
   }
+  else
+  {
+    print_config("scm", Configuration::getGitInfo());
+  }
```
4tXJ7f, line 330: Maybe I am mistaken but it looks like the two branches should be swapped.

cmake/version-base.cmake
```diff
@@ -0,0 +1,14 @@
+# These are updated when making a release
+set(CVC5_LAST_RELEASE "0.0.0")
+set(CVC5_IS_RELEASE "false")
+
+# These are used in other places in cmake
+# If possible, they are updated by version-get-info.cmake
```
4tXJ7f, line 6:
```suggestion
# If possible, they are updated by version.cmake
```

contrib/make-release.py
```diff
@@ -0,0 +1,120 @@
...
+    exec(['git', 'add', 'cmake/version-base.cmake'])
+    exec(['git', 'commit', '-m', 'Start post-release for {}'.format(version)])
+    return exec(['git', 'rev-parse', 'HEAD'])
+
+
+if __name__ == '__main__':
+    parse_options()
+
+    # Compute next version
+    version = identify_next_version()
```
4tXJ7f, line 100: Shouldn't we bump the version in the post-release commit only?

4tXJ7f (changes requested):
Looking good, I just have some questions.

contrib/make-release.py
nafur: In the context of a release, we have three commits to consider:
- pre-release (the last commit before you run this script): `LAST_RELEASE` is whatever the last release was, and `IS_RELEASE` is false.
- release (the tagged commit created by this script): `LAST_RELEASE` is the new release, and `IS_RELEASE` is true
- post-release (the second commit from this script): `LAST_RELEASE` is still the same, `IS_RELEASE` is false

I see that `LAST_RELEASE` my be slightly misleading for the release commit itself, though.

src/options/options_handler.cpp
nafur: Thanks!

contrib/make-release.py
4tXJ7f: Ah right, I see, makes sense.

nafur: As per offline discussion: we will have the last tag in `CVC5_LAST_RELEASE`, but have `version.cmake` print `CVC5_LAST_RELEASE + 0.0.1` when the current commit is not a tag.

PR #7254, opened 2021-09-27
Add CI workflow to test different cmake versions

nafur:
This PR adds a new CI workflow that allows to test the current master branch against many different cmake versions. It is mostly useful after modifying our cmake setup to check compatibility with older cmake versions.
The workflow is not triggered automatically, but can be started manually.

I found it useful, do we want this in our repository?

### DIFF
--- a/.github/workflows/cmake-version.yml
+++ b/.github/workflows/cmake-version.yml
@@ -1,0 +1,144 @@
+on: [workflow_dispatch]
+name: cmake-version
+
+jobs:
+  build:
+    continue-on-error: true
+    name: ubuntu:production-dbg
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        cmake_version: [
+          "3.12", "3.13", "3.14", "3.15", "3.16",
+          "3.17", "3.18", "3.19", "3.20", "3.21"
+        ]
+
+    steps:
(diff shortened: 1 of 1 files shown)


4tXJ7f (changes requested):
I am in favor of having this workflow (and maybe run it nightly?). If it turns out that we don't use it often, we can always remove it. However, I'd be in favor of making the compilation/testing part reusable and share it with our main `ci.yml` workflow.

nafur:
After we finally realized that there are composite actions and we can have them locally within `.github/`, I've refactored our CI "a bit" to make it more maintainable and reusable. I'm open for suggestions :)

.github/workflows/cmake-version.yml
```diff
@@ -0,0 +1,62 @@
...
+      run: ccache -s
+
+    - name: Run tests
+      uses: ./.github/actions/run-tests
+      with:
+        check-examples: true
+        check-python-bindings: true
+        check-unit-tests: true
+
+    - name: Build Documentation
```
4tXJ7f, line 55: We could call the build-documentation action instead, no?

4tXJ7f: Makes sense

PR #7255, opened 2021-09-27
First round of refactoring on NlModel

nafur:
This PR performs a first refactoring on the NlModel class. It improves model value computation, comparison and stores the model substitutions in a map (instead of two vectors).


ajreynol (changes requested):
Generally looks good although I question the use of map for substitutions, as discussed offline.

src/theory/arith/arith_utilities.h
```diff
@@ -313,13 +313,13 @@ void printRationalApprox(const char* c, Node cr, unsigned prec = 5);
...
  *
- * This computes the substitution n { vars -> subs }, but with the caveat
+ * This computes the substitution n { subs }, but with the caveat
  * that subterms of n that belong to a theory other than arithmetic are
  * not traversed. In other words, terms that belong to other theories are
  * treated as atomic variables. For example:
  *   (5*f(x) + 7*x ){ x -> 3 } returns 5*f(x) + 7*3.
  */
-Node arithSubstitute(Node n, std::vector<Node>& vars, std::vector<Node>& subs);
+Node arithSubstitute(Node n, const std::map<Node, Node>& sub);
```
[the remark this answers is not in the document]
nafur: We should have some serious discussion about this pattern. I've changed it to use `std::vector<std::pair<>>` now: my main concern was to unify the (technically completely separate) vectors into a single data structure.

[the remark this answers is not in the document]
ajreynol: The `Subs` class (https://github.com/cvc5/cvc5/blob/master/src/expr/subs.h) was one intended way to mitigate this, e.g. you have a trusted interface to adding to parallel vectors.

src/theory/arith/nl/nl_model.cpp
```diff
@@ -63,46 +57,43 @@ void NlModel::resetCheck()
...
+  Trace("nl-ext-mv-debug") << "computeModelValue " << n
+                           << ", isConcrete=" << isConcrete << std::endl;
   Node ret;
-  Kind nk = n.getKind();
   if (n.isConst())
   {
     ret = n;
   }
-  else if (!isConcrete && hasTerm(n))
+  else if (auto it = d_arithVal.find(n); !isConcrete && it != d_arithVal.end())
```
ajreynol, line 87: This is a bit ugly, can we split this up so that we aren't declaring variables within a conditional?

ajreynol (changes requested):
One last minor comment.

src/theory/arith/nl/nl_model.cpp
nafur: There are two reasons I did it like this:
1. this makes `it` a variable that is local to this `if`.
2. it is actually pretty hard to rewrite this code so that the overall structure (the `if-then-else`) makes sense and at the same time only have the `d_arithVal.find()` executed when necessary.

ajreynol: I would prefer a refactoring where e.g. a new method is used, like `bool hasTerm(TNode n, TNode& rep)`.  When this returns true, it sets `rep` to the value of the lookup.

nafur: I don't know. The usage of such output parameters is usually discouraged (IMHO for good reason). This is standard C++ syntax meant for exactly such cases.

I see that this syntax is rather new (C++17) and as such not widely used in our code base (yet). However, I think it is not hard to understand (consider the resemblance in both syntax and semantics to a `for` loop), and we should really use it when appropriate.

PR #7257, opened 2021-09-28
Remove linking against RT

nafur:
This PR removes long obsolete cmake code that is only required when using a pre-2013 glibc.


4tXJ7f (approved):
How can you not love PRs that only remove code?

PR #7258, opened 2021-09-29
Remove BUILD_LIB_ONLY flag

nafur:
This PR removes the `BUILD_LIB_ONLY` cmake option, and the associated `--lib-only` configure script option.
If you only want the library, just run `make cvc5` instead of `make`.


mpreiner (approved):

makaimann (approved):
LGTM!

PR #7259, opened 2021-09-29
contrib: Fix check for get-script-header.sh.

mpreiner:
PR #7219 removed CVC language support and therefore also the file `src/parser/cvc/Cvc.g`. This commit fixes the check (and the nightlies).


4tXJ7f (approved):
LGTM

PR #7261, opened 2021-09-29
Execute `set-option` commands in parse-only mode

4tXJ7f:
Under some circumstances, the parser relies on options to determine
which symbols are defined. In parse-only mode, we were not executing
_any_ commands, which lead to spurious parse failures in `--parse-only`
mode (see regression in this commit). This fixes the issue by always
executing commands that set options.

### DIFF
--- a/src/main/command_executor.cpp
+++ b/src/main/command_executor.cpp
@@ -205,7 +205,11 @@ bool solverInvoke(api::Solver* solver,
     cmd->toStream(ss);
   }
 
-  if (solver->getOptionInfo("parse-only").boolValue())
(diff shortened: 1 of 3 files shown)


mpreiner (approved):

ajreynol (changes requested):
I'm not sure I agree with the philosophy of this change.

I think the proper fix for this issue is to make it so that options do not impact what symbols are defined.  In particular, this means that `eqarray` should be defined as a symbol when in non-strict compliance mode.  This is analogous to some extended operators in strings.

ajreynol:
FYI, Abdal and I have discussed this issue yesterday, he was planning to submit a fix.

PR #7263, opened 2021-09-29
[API] Update comments w.r.t. SymFPU, fix typos

4tXJ7f:
Previously, SymFPU was an optional dependency but it is now required.
The comments in the API were not updated to reflect that. This commit
fixes the comments.


martin-cs (approved):
Thanks for correcting my spelling mistakes!

PR #7266, opened 2021-09-29
Use the proper evaluator for optimized SyGuS datatype rewriting

ajreynol:
This updates the datatypes rewriter to use the evaluator from Env instead of creating local copies of Evaluator.  This makes all uses of the Evaluator dependent on the proper options (e.g. which will be based later on the cardinality of the alphabet for strings).

This PR moves one utility method (sygusToBuiltinEval) to the datatypes rewriter, as it uses an Evaluator that will be dependent on options.

Notice that this is another instance where it is important for us to make the cache for the rewriter local.  The same issue occurs for other places where rewriting is dependent on options.  This issue will be revisited when the option for strings alphabet cardinality is added.


4tXJ7f (approved):
LGTM

PR #7268, opened 2021-09-29
Simplify the syntax and representation of the separation logic empty heap constraint

ajreynol:
Since ~1 year ago, we insist that the location and data types of the separation logic heap are set upfront, analogous to the `set-logic` command. This means that the separation logic empty heap constraint does not need to be annotated with its types.

This makes `SEP_EMP` a nullary Boolean operator instead of binary predicate (taking dummy arguments whose types specify the heap type).  It changes the smt2 extended syntax from `(_ emp T1 T2)` to `sep.emp`.


src/api/cpp/cvc5.cpp
```diff
@@ -5080,20 +5080,24 @@ Sort Solver::mkTupleSortHelper(const std::vector<Sort>& sorts) const
 Term Solver::mkTermFromKind(Kind kind) const
 {
   CVC5_API_KIND_CHECK_EXPECTED(
-      kind == PI || kind == REGEXP_EMPTY || kind == REGEXP_SIGMA, kind)
+      kind == PI || kind == REGEXP_EMPTY || kind == REGEXP_SIGMA || kind==SEP_EMP, kind)
```
abdoo8080, line 5083: Format?

abdoo8080 (commented):
LGTM!

abdoo8080 (approved):

PR #7269, opened 2021-09-29
Properly restrict PBE symmetry breaking for abduction queries

ajreynol:
This ensures we infer when a conjecture is PBE based on the conjecture *plus* the side condition for abduction.  This fixes issues where the sygus solver was over-pruning solutions for abduction queries.

It also changes the names of internal symbols used for abduction/interpolation queries. These names are used when the experimental `sygus-stream` is used.  These symbols are changed (from `"A"`) to avoid confusion with user symbols.


abdoo8080 (approved):
LGTM

PR #7270, opened 2021-09-29
Remove support for extended `(check-sat <term>)` command.

abdoo8080:
This PR removes support for the extended `(check-sat <term>)` command which overlaps in functionality with the standard `(check-sat-assuming (<prop_literal>*))` command.


4tXJ7f (approved):
LGTM, this is great :)

PR #7242, opened 2021-09-23
Properly cache assertions in static learning preprocessing pass.

mpreiner:
The cache is user-context dependent and guarantees that `ppStaticLearn` is only called once per assertion in a given context - if assertions are popped from the context and added again `ppStaticLearn` is called again on that assertion. Further, the preprocessing pass now also handles the splitting of top-level AND terms. Before that each theory was responsible for splitting and caching. Refactoring the corresponding `ppStaticLearn` theory methods will be part of future PRs.

Marking do-not-merge until cluster runs finished.

PR #7271, opened 2021-09-29
bv: Refactor ppRewrite and move to TheoryBV.

mpreiner:
This commit moves and refactors the ppRewrite() and ppStaticLearn() code from the layered solver to TheoryBV  in order to apply a default set of preprocessing rules for each solver. BV solver can still implement additional rules.

Note: Some of the rewrite rules in ppRewrite() have been removed since cluster runs showed that they don't improve anything, but actually makes the solver a lot worse.

Note2: This PR should only be merged after #7242 was merged.


ajreynol:
Are you waiting on cluster runs for this? Otherwise, this LGTM.

mpreiner:
No cluster runs are done and based on them I removed some of the rewrite rules.

ajreynol (approved):

PR #7277, opened 2021-09-30
Make theory engine modules use Env

ajreynol:
This updates several core modules of TheoryEngine to use Env and eliminates some getters from TheoryEngine.


4tXJ7f (approved):
LGTM, I still have a slight preference to use multiple inheritance over using an `Env&` in classes that already extend a different class, but we should discuss that in a Tuesday meeting.

ajreynol:
Yep, I noticed that, actually, another solution may be to eliminate the inheritance on ContextObj there via a callback contained in the class and then go with EnvObj.

PR #7281, opened 2021-09-30
Enable static builds in CI

nafur:
This PR modifies our CI builds to have two static production builds. These binaries will be used as release artifacts later.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,20 +6,20 @@ jobs:
     strategy:
       matrix:
         include:
-          - name: ubuntu:production
+          - name: ubuntu:production-static
             os: ubuntu-latest
-            config: production --auto-download --all-bindings --editline --docs
-            cache-key: production
+            config: production --auto-download --all-bindings --editline --docs --static
+            cache-key: production-static
             python-bindings: true
             build-documentation: true
             check-examples: true
             exclude_regress: 3-4
             run_regression_args: --no-check-unsat-cores --no-check-proofs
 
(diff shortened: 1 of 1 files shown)


src/main/CMakeLists.txt
```diff
@@ -78,9 +78,11 @@ endif()
...
 #   https://cmake.org/cmake/help/v3.0/prop_tgt/LINK_SEARCH_END_STATIC.html
 if(ENABLE_STATIC_BINARY)
-  set_target_properties(cvc5-bin PROPERTIES LINK_FLAGS -static)
-  set_target_properties(cvc5-bin PROPERTIES LINK_SEARCH_START_STATIC ON)
-  set_target_properties(cvc5-bin PROPERTIES LINK_SEARCH_END_STATIC ON)
+  if(NOT CMAKE_SYSTEM_NAME STREQUAL "Darwin")
+    set_target_properties(cvc5-bin PROPERTIES LINK_FLAGS -static)
+    set_target_properties(cvc5-bin PROPERTIES LINK_SEARCH_START_STATIC ON)
+    set_target_properties(cvc5-bin PROPERTIES LINK_SEARCH_END_STATIC ON)
+  endif()
```
4tXJ7f, line 85: Our current solution for this problem is that on macOS, we always compile with `--static --no-static-binary` if we want almost-static binaries. I think that that is the better solution because it does not silently change what the user asked for.

4tXJ7f (changes requested):
Cool, just a request regarding the macOS build.

PR #7282, opened 2021-09-30
Rename SmtEngine to SolverEngine.

aniemetz:


ajreynol (approved):
LGTM, I think SmtEngineState still needs to be renamed to SolverEngineState, we can do this after this PR.

aniemetz:
> LGTM, I think SmtEngineState still needs to be renamed to SolverEngineState, we can do this after this PR.

Yes, same with SmtScope. I'm trying to keep it small :-)

PR #7283, opened 2021-09-30
Various improvements to documentation

nafur:
This PR adds a couple of improvements to our documentation setup
- add an explicit dependency on the extension so that sphinx rebuilds pages appropriately when an extension is modified
- sphinx now emits warnings for examples that are not implemented in all languages (smt2, cpp, java and python)
- add `-v` to make the sphinx output more log-file friendly
- avoid a warning when the java bindings (and thus java docs) are not build
- replace `file(WRITE` by `configure_file` to avoid rather common erroneous rebuilds of python docs
- fix a reference to a label in the python docs
- suppress timestamps in javadoc output to avoid docs-ci being triggered for every PR
- improve placeholder message when java bindings are disabled


docs/api/java/index.rst
```diff
@@ -1,4 +1,4 @@
 Java API Documentation
 ======================
 
-This file is just a placeholder that is replaced by javadocs index.html
\ No newline at end of file
+This documentation was built while java bindings were disabled. This part of the documentation is empty. Please enable :code:`BUILD_BINDINGS_JAVA` in :code:`cmake` and build the documentation again.
```
4tXJ7f, line 4:
```suggestion
This documentation was built while Java bindings were disabled. This part of the documentation is empty. Please enable :code:`BUILD_BINDINGS_JAVA` in :code:`cmake` and build the documentation again.
```

4tXJ7f (approved):
LGTM

PR #7284, opened 2021-10-01
Rename SmtScope to SolverEngineScope.

aniemetz:


src/theory/bv/abstraction.cpp
```diff
@@ -19,9 +19,9 @@
 #include "options/bv_options.h"
 #include "printer/printer.h"
 #include "smt/dump.h"
-#include "smt/smt_engine_scope.h"
 #include "smt/smt_statistics_registry.h"
 #include "smt/solver_engine.h"
+#include "smt/solver_engine_scope.h"
```
ajreynol, line 24: Nit: it might be nice to already delete these includes, since they all refer to places in the code that used to rely on SmtEngine in scope but now should not (deleting these would confirm this).

ajreynol (changes requested):

PR #7285, opened 2021-10-01
Finish the LFSC printer

ajreynol:
This adds the remaining implementation of the LFSC printer.

It also corrects a bug introduced during merging the last PR.

The printer will be connected to the proof manager in a follow up PR.  It will also be extended with support for DSL proof rule printing when the infrastructure is added.


src/proof/lfsc/lfsc_printer.cpp
```diff
@@ -44,7 +44,219 @@ void LfscPrinter::print(std::ostream& out,
                         const std::vector<Node>& assertions,
                         const ProofNode* pn)
 {
-  // to be added
+  Trace("lfsc-print-debug") << "; ORIGINAL PROOF: " << *pn << std::endl;
+  // closing parentheses
+  std::stringstream cparen;
+  const ProofNode* pnBody = pn->getChildren()[0].get();
```
yoni206, line 50: How about asserting that `pn` has children right before this line?

src/proof/lfsc/lfsc_printer.cpp
```diff
@@ -44,7 +44,219 @@ void LfscPrinter::print(std::ostream& out,
...
+  // Print the body of the outermost scope
+  printProofInternal(&lpcp, pnBody, emptyLetBind, pletMap, passumeMap);
+
+  // [3] print warnings
+  for (PfRule r : d_trustWarned)
+  {
+    out << "; WARNING: adding trust step for " << r << std::endl;
+  }
+
+  // [4] print the DSL rewrite rule declarations
```
yoni206, line 223: This feels like a TODO, no?
I suggest removing this for now, or adding a TODO (and a corresponding issue).

src/proof/lfsc/lfsc_printer.cpp
```diff
@@ -44,7 +44,219 @@ void LfscPrinter::print(std::ostream& out,
...
+            for (size_t j = 0, nargs = cons.getNumArgs(); j < nargs; j++)
+            {
+              preamble << "(! s" << j << " sort ";
+              tcparen << ")";
+            }
+            preamble << "sort" << tcparen.str() << ")";
+          }
+          preamble << std::endl;
+        }
+        else
```
yoni206, line 124: I think it would help to mention in a comment what are the remaining cases other than sorts and datatypes.

PR #7286, opened 2021-10-01
Add cardinality constraint utilities

ajreynol:
This class will be used to replace the `fmf.card` operator with a 0-ary indexed parameterized nullary predicate for representing cardinality constraints for UF.

This is in preparation for eliminating hacks in the core of theory combination for e.g. not considering the dummy arguments of cardinality constraints to be shared terms.


src/expr/cardinality_constraint.cpp
```diff
@@ -0,0 +1,108 @@
...
+{
+  return out << "fmf.card(" << cc.getType() << ", " << cc.getUpperBound()
+             << ')';
+}
+
+size_t CardinalityConstraintHashFunction::operator()(
+    const CardinalityConstraint& cc) const
+{
+  return std::hash<TypeNode>()(cc.getType())
+         * IntegerHashFunction()(cc.getUpperBound());
```
4tXJ7f, line 63: It'd be slightly better to use our [`PairHashFunction`](https://github.com/cvc5/cvc5/blob/master/src/util/hash.h#L60-L66) util to combine the hashes.

4tXJ7f (changes requested):
Just a minor suggestion regarding the hash function.

src/expr/cardinality_constraint.cpp
ajreynol: What is the code I'd use?

4tXJ7f: It should be possible to do something like:

```
using CardinalityConstraintHashFunction = PairHashFunction<TypeNode, Integer, std::hash<TypeNode>, IntegerHashFunction>;
```

instead of declaring and implementing the class.

4tXJ7f (approved):
LGTM, thanks for the update.

PR #7287, opened 2021-10-01
Move isFiniteType from theory engine to Env

ajreynol:
In preparation for breaking more dependencies on TheoryEngine.


4tXJ7f (approved):
Great, I think that this move makes a lot of sense.

PR #7293, opened 2021-10-01
Fix some python docstrings which lead to sphinx warnings

nafur:
This PR fixes some docstrings that are not properly indented for sphinx.


alex-ozdemir (approved):
Ah, thank you for doing this!

PR #7294, opened 2021-10-01
Remove (apparently obsolete) checks in cms config

nafur:
This PR remove checks from the CryptoMiniSat config for the existence of a couple of static libraries. Apparently, these are not necessary any more.


4tXJ7f (approved):
LGTM

PR #7295, opened 2021-10-01
Finish refactoring on option handlers

nafur:
This PR finishes refactoring on the option handlers.


src/options/options_handler.cpp
```diff
@@ -541,40 +428,125 @@ void OptionsHandler::setDefaultDagThresh(const std::string& option,
   Dump.getStream() << expr::ExprDag(dag);
 }
 
-// main/options_handlers.h
+static void print_config (const char * str, std::string config) {
```
ajreynol, line 431: Format?

src/options/options_handler.h
```diff
@@ -128,65 +129,58 @@ class OptionsHandler {
...
+  /** Check that abc is enabled */
   void abcEnabledBuild(const std::string& option,
                        const std::string& flag,
                        bool value);
+  /** Check that abc is enabled */
   void abcEnabledBuild(const std::string& option,
                        const std::string& flag,
                        const std::string& value);
-
+  /** Check that abc is enabled */
```
ajreynol, line 142: Fix this comment?

src/options/smt_options.toml
```diff
@@ -254,20 +254,12 @@ name   = "SMT Layer"
...
-  predicates = ["setProduceAssertions"]
-  help       = "deprecated name for produce-assertions"
-
 [[option]]
   name       = "produceAssertions"
   category   = "common"
   long       = "produce-assertions"
   type       = "bool"
-  predicates = ["setProduceAssertions"]
+  alias      = ["interactive-mode"]
```
ajreynol, line 262: What exactly is this doing?  Is this saying that produce assertions is true when interactive-mode is true?  Can this be refactored not to rely on the toml?

ajreynol (commented):
A few comments...

src/options/options_handler.h
nafur: What is wrong with the comment? (the library is actually called abc)

src/options/smt_options.toml
nafur: This makes `--produce-assertions` and `--interactive-mode` are different names for the same thing. This actually preserves current behaviour, as they have, in fact, always the same value; this is, however, achieved in a weird and very convoluted way right now...

ajreynol: Really? I would perhaps expect that produce-assertions is *implied* by interactive-mode but not equivalent to it?

Regardless this is indeed very convoluted.

nafur: This handler was called by both options: https://github.com/cvc5/cvc5/pull/7295/files#diff-831a3e7302809583cb776be88353361408845885dc7d393ca45cad5e069d7e18L509 ...

nafur: What do you want it to do? Have `--interactive-mode` imply `--produce-assertions`?
Note that the the value behind `--interactive-mode` is not used in all of cvc5, from all that I can tell...

ajreynol: If its not used, then can we delete `interactive-mode`?

Otherwise maybe it'd make sense to have it imply produce-assertions but not vice versa.

nafur: The current change was my defensive way of essentially deleting `--interactive-mode`, keeping it around as an alias in case someone uses it. Fwiw, you can simply use `--produce-assertions` to do the very same thing.

ajreynol: I think I'd outright delete it if no one is using it.

ajreynol (approved):

PR #7297, opened 2021-10-03
Add missing functions in Term.java

mudathirmahgoub:
This PR adds recent API functions that were added to terms. 
It also uses BigInteger now for integer terms.

### DIFF
--- a/src/api/java/cvc5/Utils.java
+++ b/src/api/java/cvc5/Utils.java
@@ -151,7 +151,25 @@ public class Utils
       String[] pair = rational.split("/");
       return new Pair<>(new BigInteger(pair[0]), new BigInteger(pair[1]));
     }
+    if (rational.contains("."))
+    {
+      String[] pair = rational.split("\\.");
+      if ((new BigInteger(pair[1])).equals(BigInteger.ZERO))
+      {
+        return new Pair<>(new BigInteger(pair[0]), new BigInteger("1"));
+      }
+      BigInteger denominator = BigInteger.TEN.pow(pair[1].length());
+      return new Pair<>(new BigInteger(rational.replace(".", "")), denominator);
+    }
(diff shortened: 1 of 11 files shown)


src/api/java/cvc5/Utils.java
```diff
@@ -151,9 +151,27 @@ public static void validateUnsigned(long[] integers, String name) throws CVC5Api
...
+    if (rational.contains("."))
+    {
+      String[] pair = rational.split("\\.");
+      if ((new BigInteger(pair[1])).equals(BigInteger.ZERO))
+      {
+        return new Pair<>(new BigInteger(pair[0]), new BigInteger("1"));
+      }
+      BigInteger denominator = BigInteger.TEN.pow(pair[1].length());
+      return new Pair<>(new BigInteger(rational.replace(".", "")), denominator);
+    }
```
nafur, line 163: Is this necessary? I don't think we ever return a "real" value in decimal, but always use a fractional representation.

mudathirmahgoub: Yes, I encountered a case where there is a decimal point like "1.0"

nafur: Ah, I've just seen why this happens: https://github.com/cvc5/cvc5/blob/master/src/api/cpp/cvc5.cpp#L3012

mudathirmahgoub: Ok. By the way, I prefer to have a pair of strings <"1", "2"> than a single string  "1/2". 
I guess most languages does not handle "1/2" properly, so we have to do splitting to convert it to a pair of integers. 
Alternative is to use `std::pair<int64_t, uint64_t>` with the risk of precision loss.

nafur:
```suggestion
```

4tXJ7f (approved):
LGTM

PR #7299, opened 2021-10-04
Add missing methods to Solver.java

mudathirmahgoub:
A life of solver object was simple before summer. Then it got complicated with `getDifficulty`, optional proofs `getDifficulty`, more assumptions `addSygusAssume`, and finally different options to choose from `OptionInfo`. 
This PR attempts to prepare solver objects for this new life.

### DIFF
--- a/src/api/java/cvc5/DriverOptions.java
+++ b/src/api/java/cvc5/DriverOptions.java
@@ -1,0 +1,63 @@
+/******************************************************************************
+ * Top contributors (to current version):
+ *   Gereon Kremer, Mudathir Mohamed
+ *
+ * This file is part of the cvc5 project.
+ *
+ * Copyright (c) 2009-2021 by the authors listed in the file AUTHORS
+ * in the top-level source directory and their institutional affiliations.
+ * All rights reserved.  See the file COPYING in the top-level source
+ * directory for licensing information.
+ * ****************************************************************************
+ *
+ * The cvc5 java API.
+ */
+
+package cvc5;
+
+import java.util.Optional;
+
+/**
+ * Holds some description about a particular option, including its name, its
+ * aliases, whether the option was explicitly set by the user, and information
+ * concerning its value. The `valueInfo` member holds any of the following
+ * alternatives:
+ * - VoidInfo if the option holds no value (or the value has no native type)
+ * - ValueInfo<T> if the option is of type boolean or String, holds the
+ *   current value and the default value.
+ * - NumberInfo<T> if the option is of type BigInteger or double, holds
+ *   the current and default value, as well as the minimum and maximum.
+ * - ModeInfo if the option is a mode option, holds the current and default
+ *   values, as well as a list of valid modes.
+ * Additionally, this class provides convenience functions to obtain the
+ * current value of an option in a type-safe manner using boolValue(),
+ * stringValue(), intValue(), and doubleValue(). They assert that
+ * the option has the respective type and return the current value.
+ */
+public class DriverOptions extends AbstractPointer
+{
+  // region construction and destruction
+  DriverOptions(Solver solver, long pointer)
+  {
+    super(solver, pointer);
+  }
+
+  protected static native void deletePointer(long pointer);
+
+  public long getPointer()
+  {
+    return pointer;
+  }
+
+  @Override public void finalize()
+  {
+    deletePointer(pointer);
+  }
+
+  /**
+   * @return a string representation of this optionInfo.
+   */
+  protected native String toString(long pointer);
+
+  // endregion
+};
--- a/src/api/java/cvc5/DriverOptions.java
+++ b/src/api/java/cvc5/DriverOptions.java
@@ -15,24 +15,11 @@
 
 package cvc5;
 
-import java.util.Optional;
-
 /**
- * Holds some description about a particular option, including its name, its
- * aliases, whether the option was explicitly set by the user, and information
- * concerning its value. The `valueInfo` member holds any of the following
- * alternatives:
- * - VoidInfo if the option holds no value (or the value has no native type)
- * - ValueInfo<T> if the option is of type boolean or String, holds the
- *   current value and the default value.
- * - NumberInfo<T> if the option is of type BigInteger or double, holds
- *   the current and default value, as well as the minimum and maximum.
- * - ModeInfo if the option is a mode option, holds the current and default
- *   values, as well as a list of valid modes.
- * Additionally, this class provides convenience functions to obtain the
- * current value of an option in a type-safe manner using boolValue(),
- * stringValue(), intValue(), and doubleValue(). They assert that
- * the option has the respective type and return the current value.
+ * Provides access to options that can not be communicated via the regular
+ * getOption() or getOptionInfo() methods. This class does not store the options
+ * itself, but only acts as a wrapper to the solver object. It can thus no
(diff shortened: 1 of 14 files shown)


src/api/java/cvc5/OptionInfo.java
```diff
@@ -0,0 +1,209 @@
...
+ * - NumberInfo if the option is of type BigInteger or double, holds
+ *   the current and default value, as well as the minimum and maximum.
+ * - ModeInfo if the option is a mode option, holds the current and default
+ *   values, as well as a list of valid modes.
+ * Additionally, this class provides convenience functions to obtain the
+ * current value of an option in a type-safe manner using boolValue(),
+ * stringValue(), intValue(), and doubleValue(). They assert that
+ * the option has the respective type and return the current value.
+ */
+public class OptionInfo extends AbstractPointer
```
nafur, line 37: All (?) other classes only hold a pointer to the corresponding cpp class (and sometimes a reference to the solver). Here, we immediately copy the members as well... why is that?

src/api/java/cvc5/OptionInfo.java
```diff
@@ -0,0 +1,209 @@
...
+  private native String getName(long pointer);
+
+  private native String[] getAliases(long pointer);
+
+  private native boolean getSetByUser(long pointer);
+
+  private native ValueInfo getValueInfo(long pointer);
+
+  /** Has no value information */
+  public class VoidInfo extends ValueInfo<Object>
```
nafur, line 77: Having VoidInfo inherit from ValueInfo seems strange: VoidInfo does not hold a value.
I see that there is no direct equivalent for a variant in Java, so I'm not sure how to properly implement this...
@4tXJ7f any ideas?

src/api/java/cvc5/Solver.java
```diff
@@ -1828,7 +1866,47 @@ public String getOption(String option)
...
+    for (Map.Entry<Long, Long> entry : map.entrySet())
+    {
+      Term key = new Term(this, entry.getKey());
+      Term value = new Term(this, entry.getValue());
+      ret.put(key, value);
+    }
+    return ret;
+  }
+
+  private native Map<Long, Long> getDifficulty(long pointer);
```
nafur, line 1889: The cpp API method already returns `Term` objects. why convert them to `Long` and then convert them back to `Term`?

src/api/java/cvc5/DriverOptions.java
```diff
@@ -0,0 +1,53 @@
...
+
+package cvc5;
+
+/**
+ * Provides access to options that can not be communicated via the regular
+ * getOption() or getOptionInfo() methods. This class does not store the options
+ * itself, but only acts as a wrapper to the solver object. It can thus no
+ * longer be used after the solver object has been destroyed.
+ */
+public class DriverOptions extends AbstractPointer
```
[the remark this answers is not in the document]
mudathirmahgoub: I thought the same, thanks for the confirmation. However, it is possible to redirect the streams from C++ to Java streams and vice versa, but it is a hassle. None of the streams is implemented in the Java API now. So this may be a place holder for when streams are supported. Do you still want it to be removed?

src/api/java/cvc5/OptionInfo.java
mudathirmahgoub: In general, calls to jni should be minimized for performance reasons. In this class, it seems we only have getters and no setters. So it makes sense to me to visit jni only once during object construction. This also mimics the c++ side where OptionInfo objects which are just DTOs unlike classes like terms and sorts. I think the implementation for other classes could be optimized as well.

src/api/java/cvc5/DriverOptions.java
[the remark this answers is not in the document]
nafur: I had a short discussion with @4tXJ7f about that. We figured that it may be possible to wrap the streams, but that it would probably not that easy. In particular, note that the `DriverOptions` allows to **set** streams as well (they are non-const references), which is intentional.
If you have a clear idea on how to achieve this, it would be awesome to have it. But we are fine with just not supporting this in the java API, and then not having the DriverOptions object.

src/api/java/cvc5/Solver.java
mudathirmahgoub: I guess the main reason is to avoid the map destructor deleting the terms allocated in the stack when the jni function is out of scope, by using the copy constructor [here](https://github.com/cvc5/cvc5/pull/7299/files#diff-829d558eb6fd73952144aeb593e9d50bf9f237f45d6403bf84a4eae7e59350ceR1985) and [here](https://github.com/cvc5/cvc5/pull/7299/files#diff-829d558eb6fd73952144aeb593e9d50bf9f237f45d6403bf84a4eae7e59350ceR1986) to get a copy of these terms in the heap.
The second reason is to minimize the annoying part of looking for java classes, methodIds to construct java objects for the map.

src/api/java/cvc5/OptionInfo.java
mudathirmahgoub: I also don't know. I tried using a static class with singleton instance but I could not get it to work for inner classes. 
So I am also curious how to get a singleton instance of this class while maintaining it as an inner class of `OptionInfo`.

src/api/java/cvc5/DriverOptions.java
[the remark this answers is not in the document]
mudathirmahgoub: It seems there is a way to support streams with buffers as described in this article, 
https://codeforfun.wordpress.com/2010/04/22/sharing-output-streams-through-a-jni-interface/
but we can support this in the future.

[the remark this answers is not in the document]
nafur: Okay, it seems doable, but I think there are enough technical details to warrant a separate PR. Also, we should figure out whether we actually want to support this... @4tXJ7f

src/api/java/cvc5/Solver.java
nafur: `getDifficulty()` explicitly returns `Term` objects to allow for values beyond machine-size integers (be it rationals or big integers).
While I have full sympathies for the pain caused by lifetime issues, please let us avoid this conversion. It is almost guaranteed to become a problem in the future...

mudathirmahgoub: The long Map here is used for pointers to terms (not machine-size integers).
See the code starting from this line
https://github.com/cvc5/cvc5/pull/7299/files#diff-829d558eb6fd73952144aeb593e9d50bf9f237f45d6403bf84a4eae7e59350ceR1985

mudathirmahgoub:
> After talking to @nafur offline, I am also thinking that it is not worth keeping `VoidInfo` as a singleton for now. There isn't really any benefit and it is more complicated. @mudathirmahgoub Could we change it to not be a singleton?

Done, although it is not my preference.

src/api/java/cvc5/OptionInfo.java
```diff
@@ -0,0 +1,205 @@
...
+      this.modes = modes;
+    }
+    public String[] getModes()
+    {
+      return modes;
+    }
+  }
+
+  /** Has no value information */
+  public static class VoidInfo extends BaseInfo
```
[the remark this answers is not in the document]
mudathirmahgoub: Removed static which is a reminiscent of the singleton instance

PR #7300, opened 2021-10-04
Make decision engine use env

ajreynol:


src/decision/decision_engine.cpp
```diff
@@ -38,15 +36,11 @@ void DecisionEngine::finishInit(CDCLTSatSolverInterface* ss, CnfStream* cs)
 
 prop::SatLiteral DecisionEngine::getNext(bool& stopSearch)
 {
-  d_resourceManager->spendResource(Resource::DecisionStep);
+  d_env.getResourceManager()->spendResource(Resource::DecisionStep);
```
4tXJ7f, line 39: Should we add a `resourceManager()` method to `EnvObj`?

4tXJ7f (approved):
LGTM, thanks for the update

PR #7302, opened 2021-10-04
Eliminating static calls to rewriter from strings

ajreynol:


src/theory/strings/regexp_operation.cpp
```diff
@@ -994,7 +996,6 @@ Node RegExpOpr::reduceRegExpNegConcatFixed(Node mem, Node reLen, size_t index)
     }
   }
   Node r2 = nvec.size() == 1 ? nvec[0] : nm->mkNode(REGEXP_CONCAT, nvec);
-  r2 = Rewriter::rewrite(r2);
```
4tXJ7f, line 997: Why is this rewrite removed?

4tXJ7f (changes requested):
Just a minor question.

src/theory/strings/regexp_operation.cpp
ajreynol: It wasn't necessary and actually will prevent us from doing proofs.

PR #7307, opened 2021-10-05
Avoid calling `quoteSymbol` multiple times.

abdoo8080:
Calling `quoteSymbol(...)` twice on an empty string `""` returns `"|__|"`.


4tXJ7f (approved):
Nice catch! This looks good to me.

PR #7309, opened 2021-10-05
Eliminate more hard coded uses of user context

ajreynol:
This is in preparation to make the "lemma context" configurable.


4tXJ7f (approved):
LGTM

src/theory/shared_terms_database.cpp
```diff
@@ -55,8 +53,8 @@ void SharedTermsDatabase::setEqualityEngine(eq::EqualityEngine* ee)
     if (d_pfee == nullptr)
     {
       ProofNodeManager* pnm = d_env.getProofNodeManager();
-      d_pfeeAlloc.reset(
-          new eq::ProofEqEngine(d_satContext, d_userContext, *ee, pnm));
+      d_pfeeAlloc.reset(new eq::ProofEqEngine(
+          env.getContext(), env.getUserContext(), *ee, pnm));
```
4tXJ7f, line 57: The following change should fix the PR:

```suggestion
          d_env.getContext(), d_env.getUserContext(), *ee, pnm));
```

ajreynol: Thanks :)

PR #7311, opened 2021-10-06
Eliminate more circular dependencies on solver engine

ajreynol:
This is work towards replacing our old dump infrastructure.

### DIFF
--- a/src/smt/assertions.cpp
+++ b/src/smt/assertions.cpp
@@ -35,10 +35,11 @@ namespace cvc5 {
 namespace smt {
 
 Assertions::Assertions(Env& env, AbstractValues& absv)
-    : d_env(env),
+    : EnvObj(env),
       d_absValues(absv),
       d_produceAssertions(false),
-      d_assertionList(env.getUserContext()),
+      d_assertionList(userContext()),
+      d_assertionListDefs(userContext()),
       d_globalNegation(false),
       d_assertions()
 {
@@ -104,7 +105,7 @@ void Assertions::initializeCheckSat(const std::vector<Node>& assumptions,
     Node n = d_absValues.substituteAbstractValues(e);
     // Ensure expr is type-checked at this point.
     ensureBoolean(n);
-    addFormula(n, true, true, false, false);
+    addFormula(n, true, false, false);
   }
   if (d_globalDefineFunLemmas != nullptr)
   {
@@ -113,7 +114,7 @@ void Assertions::initializeCheckSat(const std::vector<Node>& assumptions,
     // zero assertions)
     for (const Node& lemma : *d_globalDefineFunLemmas)
     {
-      addFormula(lemma, true, false, true, false);
+      addFormula(lemma, false, true, false);
     }
   }
 }
@@ -122,7 +123,7 @@ void Assertions::assertFormula(const Node& n)
 {
   ensureBoolean(n);
   bool maybeHasFv = language::isLangSygus(options::inputLanguage());
-  addFormula(n, true, false, false, maybeHasFv);
+  addFormula(n, false, false, maybeHasFv);
 }
 
 std::vector<Node>& Assertions::getAssumptions() { return d_assumptions; }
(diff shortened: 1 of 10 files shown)


src/smt/assertions.cpp
```diff
@@ -138,9 +139,12 @@ context::CDList<Node>* Assertions::getAssertionList()
 {
   return d_produceAssertions ? &d_assertionList : nullptr;
 }
+context::CDList<Node>* Assertions::getAssertionListDefinitions()
+{
+  return d_produceAssertions ? &d_assertionListDefs : nullptr;
+}
```
nafur, line 145: Can we please change this pattern?

Apart from the interesting semantics of the return value (that is not even mentioned in the function documentation), it is completely unnecessary: there is only a single place where `getAssertionList()` is called and we do not immediately assert the result to be `!= nullptr` (in `sygus_solver.cpp`); furthermore, if `!d_produceAssertions` the list is empty anyway.
Simply return the list (as a `const &`), get rid of all the pointer stuff and convert the `nullptr` check in the sygus solver by a `.empty()`.

src/smt/preprocessor.h
```diff
@@ -43,18 +46,15 @@ class PreprocessProofGenerator;
...
-               Env& env,
-               AbstractValues& abs,
-               SmtEngineStatistics& stats);
+  Preprocessor(Env& env, AbstractValues& abs, SmtEngineStatistics& stats);
   ~Preprocessor();
   /**
    * Finish initialization
    */
-  void finishInit();
+  void finishInit(SolverEngine* slv);
```
nafur, line 57: Is `nullptr` a valid argument here? If not, please use `SolverEngine&` instead.

src/smt/process_assertions.cpp
```diff
@@ -256,31 +251,28 @@ bool ProcessAssertions::apply(Assertions& as)
   }
 
   // rephrasing normal inputs as sygus problems
-  if (!d_slv.isInternalSubsolver())
+  if (options::sygusInference())
```
nafur, line 254:
```suggestion
  if (options().quantifiers.sygusInference)
```

src/smt/process_assertions.cpp
```diff
@@ -256,31 +251,28 @@ bool ProcessAssertions::apply(Assertions& as)
...
-      d_passes["sygus-infer"]->apply(&assertions);
-    }
-    else if (options::sygusRewSynthInput())
-    {
-      // do candidate rewrite rule synthesis
-      d_passes["synth-rr"]->apply(&assertions);
-    }
+    d_passes["sygus-infer"]->apply(&assertions);
+  }
+  else if (options::sygusRewSynthInput())
```
nafur, line 258:
```suggestion
  else if (options().quantifiers.sygusRewSynthInput)
```
Maybe also convert other calls to `options::foo()` when modifying this file anyway...

src/smt/solver_engine.cpp
```diff
@@ -124,8 +124,7 @@ SolverEngine::SolverEngine(NodeManager* nm, const Options* optr)
   // make statistics
   d_stats.reset(new SmtEngineStatistics());
   // reset the preprocessor
-  d_pp.reset(
-      new smt::Preprocessor(*this, *d_env.get(), *d_absValues.get(), *d_stats));
+  d_pp.reset(new smt::Preprocessor(*d_env.get(), *d_absValues.get(), *d_stats));
```
nafur, line 127:
```suggestion
  d_pp.reset(new smt::Preprocessor(*d_env, *d_absValues, *d_stats));
```

src/smt/solver_engine.cpp
```diff
@@ -124,8 +124,7 @@ SolverEngine::SolverEngine(NodeManager* nm, const Options* optr)
...
   // make the SMT solver
   d_smtSolver.reset(new SmtSolver(*d_env.get(), *d_state, *d_pp, *d_stats));
```
nafur, line 129:
```suggestion
  d_smtSolver.reset(new SmtSolver(*d_env, *d_state, *d_pp, *d_stats));
```

src/smt/solver_engine.cpp
```diff
@@ -258,7 +257,7 @@ void SolverEngine::finishInit()
     d_interpolSolver.reset(new InterpolationSolver(*d_env.get()));
```
nafur, line 257:
```suggestion
    d_interpolSolver.reset(new InterpolationSolver(*d_env));
```

src/smt/process_assertions.cpp
ajreynol: I'll do this separately.

nafur (approved):
LGTM.

PR #7315, opened 2021-10-06
Add a binary / SMT-LIB quickstart

nafur:
This PR adds a binary/SMT-LIBv2 quickstart example, based on the cpp quickstart example.


docs/binary/quickstart.rst
```diff
@@ -0,0 +1,125 @@
+Quickstart Guide
+================
+
+The primary input language for cvc5 is
+`SMT-LIB v2 <http://smtlib.cs.uiowa.edu/language.shtml>`_.
+We give a short explanation of the SMT-LIB v2 equivalent of the quickstart
```
yoni206, line 6:
```suggestion
We give a short explanation of the SMT-LIB v2 quickstart
```

I think that the smtlib quickstart example stands on its own. I expect many of the readers of this guide will not be aware of the corresponding cpp example.

docs/binary/quickstart.rst
```diff
@@ -0,0 +1,125 @@
...
+example here.
+
+First, we set the logic..
```
yoni206, line 9:
```suggestion
First, we set the logic.
```

examples/api/smtlib/quickstart.smt2
```diff
@@ -5,7 +5,7 @@
 ; necessary to print in the unsat core assertions that do not have names
```
yoni206, line 5:
```suggestion
; the following option is makes it possible to print the unsat core without naming the assertions.
```

Also, I would put the two options that are related to unsat cores together.

docs/binary/quickstart.rst
```diff
@@ -0,0 +1,125 @@
...
+     :language: smtlib
+     :lines: 8-10
+
+We define the following constraints regarding ``x`` and ``y``:
+
+.. math::
+
+  (0 < x) \wedge (0 < y) \wedge (x + y < 1) \wedge (x \leq y)
+
+We construct the required terms and assert them as follows:
```
yoni206, line 42: I think this line would be more appropriate if we would have used some `define-fun`s, but we don't. So how about:

```suggestion
We assert them as follows. Notice that in SMT-LIB v2, terms are written in prefix notation, e.g., we write `(+ x y)` instead of `(x + y)`.
```

docs/binary/quickstart.rst
```diff
@@ -0,0 +1,125 @@
...
+Now we check if the asserted formula is satisfiable, that is, we check if
+there exist values of sort ``Real`` for ``x`` and ``y`` that satisfy all
+the constraints.
+
+.. literalinclude:: ../../examples/api/smtlib/quickstart.smt2
+     :language: smtlib
+     :lines: 24
+
+The result we get from this satisfiability check is either ``sat``, ``unsat``
+or ``unknown``, and it is printed to standard output
```
yoni206, line 57:
```suggestion
or ``unknown``, and it is printed to standard output.
```

docs/binary/quickstart.rst
```diff
@@ -0,0 +1,125 @@
...
+Now, we query the solver for the values for ``x`` and ``y`` that satisfy
+the constraints.
+It is also possible to get values for terms that do not appear in the original
+formula.
+
+.. literalinclude:: ../../examples/api/smtlib/quickstart.smt2
+     :language: smtlib
+     :lines: 25-26
+
+This will print the following:
```
yoni206, line 69:
```suggestion
This will print the values of `x`, `y`, and `x-y`, in a key-value format `(<term> <value>)` one after the other:
```

docs/binary/quickstart.rst
```diff
@@ -0,0 +1,125 @@
...
+     :language: smtlib
+     :lines: 28-32
+
+Next, we assert the same assertions as above, but with integers.
+
+.. literalinclude:: ../../examples/api/smtlib/quickstart.smt2
+     :language: smtlib
+     :lines: 33-36
+
+Now, we check whether the revised assertion is satisfiable.
```
yoni206, line 90:
```suggestion
Now, we check whether the revised query is satisfiable.
```

`assertion` is a bit confusing here, I think, because there are several of them.

yoni206 (commented):
Thanks a lot for this! Some cosmetic suggestions.

yoni206 (approved):

PR #7319, opened 2021-10-07
Fix linking ANTLR3 on some systems

### LINKED ISSUE #7318: Build fails on Amazon Linux 2
Commit 65e5a909b10855d2e6b3844f7bc4efb7fbe4fe2f broke the build on my linux machine.
The linux distribution is `Amazon Linux release 2 (Karoo)`. 
CMake is version 3.13.1.

Build steps:
```
configure.sh --auto-download
cd build
make
```
No issues reported by configure. Antlr is built and installed successfully as `build/deps/lib64/libantlr3c.a`.
But `make` fails at the last step:
```
[100%] Linking CXX executable ../../bin/cvc5
../parser/libcvc5parser.so.1: error: undefined reference to 'antlr3FileStreamNew'
../parser/libcvc5parser.so.1: error: undefined reference to 'antlr3StringStreamNew'
../parser/libcvc5parser.so.1: error: undefined reference to 'antlr3BitsetLoad'
../parser/libcvc5parser.so.1: error: undefined reference to 'antlr38BitSetupStream'
../parser/libcvc5parser.so.1: error: undefined reference to 'antlr3GenericSetupStream'
../parser/libcvc5parser.so.1: error: undefined reference to 'antlr3CommonTokenStreamSourceNew'
../parser/libcvc5parser.so.1: error: undefined reference to 'antlr3SetTokenAPI'
../parser/libcvc5parser.so.1: error: undefined reference to 'antlr3LexerNewStream'
../parser/libcvc5parser.so.1: error: undefined reference to 'antlr3dfaspecialStateTransition'
../parser/libcvc5parser.so.1: error: undefined reference to 'antlr3dfaspecialTransition'
../parser/libcvc5parser.so.1: error: undefined reference to 'antlr3dfapredict'
../parser/libcvc5parser.so.1: error: undefined reference to 'antlr3ParserNewStream'
```

This appears to be caused by changes in `src/parser/CMakeFile.txt`. 
Library `libcvc5parser.so.1` is not linked against `libantlr3c.a`  as it was in
the commit just before 65e5a909b10855d2e6b3844f7bc4efb7fbe4fe2f .

Trying to build statically:
```
./configure.sh --auto-download --static
```
does not help.

4tXJ7f:
Fixes #7318. This changes how we link ANTLR3 to more closely resemble
what we do for, e.g., GMP. This fixes the issue on Amazon Linux 2
(tested using a Docker image) with CMake 3.13.3.

### DIFF
--- a/src/parser/CMakeLists.txt
+++ b/src/parser/CMakeLists.txt
@@ -101,14 +101,14 @@ endforeach()
 add_library(cvc5parser-objs OBJECT ${libcvc5parser_src_files})
 set_target_properties(cvc5parser-objs PROPERTIES POSITION_INDEPENDENT_CODE ON)
 target_compile_definitions(cvc5parser-objs PUBLIC -D__BUILDING_CVC5PARSERLIB)
-target_link_libraries(cvc5parser-objs PRIVATE ANTLR3)
-
+target_include_directories(cvc5parser-objs PRIVATE ${ANTLR3_INCLUDE_DIR})
 
 add_library(cvc5parser-shared SHARED)
 set_target_properties(cvc5parser-shared PROPERTIES SOVERSION ${CVC5_SOVERSION})
 set_target_properties(cvc5parser-shared PROPERTIES OUTPUT_NAME cvc5parser)
 target_link_libraries(cvc5parser-shared PRIVATE cvc5-shared)
 target_link_libraries(cvc5parser-shared PRIVATE cvc5parser-objs)
+target_link_libraries(cvc5parser-shared PRIVATE ANTLR3)
 
 install(TARGETS cvc5parser-shared
   EXPORT cvc5-targets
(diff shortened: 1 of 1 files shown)


4tXJ7f:
I think that it would be good to at least support the CMake versions of the Ubuntu LTS releases that still have standard support (currently 18.04 and 20.04), which means that the minimum version would be CMake 3.10.

BrunoDutertre:
Are you telling me that CMake is not backward compatible?

BrunoDutertre:
3.10 is fine for me.

4tXJ7f:
> Are you telling me that CMake is not backward compatible?

Oh no, I was just suggesting that we should support all versions >= 3.10 (instead of >= 3.15).

nafur:
These changes are included in #7329.

PR #7326, opened 2021-10-08
Add argument to distinguish lemmas and input assertions

ajreynol:
Work towards virtual clause deletion, where lemmas will be SAT-context dependent.

This adds an argument to the decision engine so it can distinguish lemmas from input assertions.


nafur (approved):
LGTM.

PR #7328, opened 2021-10-08
Provide a non-traversal interface to term formula removal

ajreynol:
Towards making theory preprocessing a single pass.


src/smt/term_formula_removal.cpp
```diff
@@ -237,17 +242,22 @@ Node RemoveTermFormulas::runInternal(TNode assertion,
...
-  Debug("ite") << "removeITEs(" << node << ")"
-               << " " << inQuant << " " << inTerm << std::endl;
-  Assert(!inQuant);
+  Node k = runCurrentInternal(node, inTerm, newLem);
+  if (!k.isNull())
+  {
+    return TrustNode::mkTrustRewrite(node, k, d_tpg.get());
+  }
+  return TrustNode::null();
+}
```
nafur, line 255: Given that `runInternal` no longer calls `runCurrent` (but only `runCurrentInternal`), this function is never called, right?

ajreynol: The purpose of these PRs is to make this method available, so its not called yet, but it is called on my branch.

PR #7329, opened 2021-10-08
Restore compatibility with cmake 3.9

nafur:
This PR goes back to the new cmake setup and makes it compatible again with cmake 3.9.
It mostly means we can not link object libraries and also not link other libraries to object libraries.

I've tested these changes within docker on `ubuntu:18-04` with a manually installed `cmake-3.9.6`. Can you check on your machines? @BrunoDutertre @dddejan


BrunoDutertre:
Doesn't work for me. 

``./configure --auto-download``

gives me this error:

```
CMake Error at cmake/version.cmake:102 (list):
  list does not recognize sub-command POP_BACK
Call Stack (most recent call first):
  CMakeLists.txt:27 (include)
```

nafur:
Okay, this part of the code should also work now.

4tXJ7f (approved):
LGTM

PR #7332, opened 2021-10-10
fix deprecation of std::iterator

Ouyancheng:
When compiling cvc5 with clang-13, it will emit lots of warnings of usages of `std::iterator` as it's deprecated since C++17. 
The recommended implementation of iterators is to manually define the following five types: 
```
template< class Iter >
struct iterator_traits;

difference_type	        Iter::difference_type
value_type	        Iter::value_type
pointer	                Iter::pointer
reference	        Iter::reference
iterator_category	Iter::iterator_category

```
And the iterator-related types could be accessed by for example `typename std::iterator_traits<Iter>::value_type value`. 

This pull request performs the fix, and the program is semantically equivalent before and after the fix. 

References: 
https://en.cppreference.com/w/cpp/iterator/iterator_traits 
https://en.cppreference.com/w/cpp/iterator/iterator

### DIFF
--- a/.github/actions/add-to-release/action.yml
+++ b/.github/actions/add-to-release/action.yml
@@ -7,6 +7,7 @@ runs:
   using: composite
   steps:
     - name: Rename binaries for release
+      shell: bash
       run: |
         cp build/bin/cvc5 cvc5-${{ runner.os }}
 
--- a/cmake/version-base.cmake
+++ b/cmake/version-base.cmake
@@ -1,5 +1,5 @@
 # These are updated when making a release
-set(CVC5_LAST_RELEASE "0.0.0")
+set(CVC5_LAST_RELEASE "0.0.2")
 set(CVC5_IS_RELEASE "false")
 
 # These are used in other places in cmake
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -353,8 +353,8 @@ libcvc5_add_sources(
   smt/solver_engine_scope.h
   smt/smt_engine_state.cpp
   smt/smt_engine_state.h
-  smt/smt_engine_stats.cpp
-  smt/smt_engine_stats.h
+  smt/solver_engine_stats.cpp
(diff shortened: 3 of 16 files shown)


Ouyancheng:
@nafur

src/api/cpp/cvc5.h
```diff
@@ -1167,11 +1167,30 @@ class CVC5_EXPORT Term
...
-  class const_iterator : public std::iterator<std::input_iterator_tag, Term>
+  class const_iterator
   {
     friend class Term;
 
    public:
+    /* The following types are required by trait std::iterator_traits */
+
+    /** Iterator tag */
+    using iterator_category = std::input_iterator_tag;
```
nafur, line 1178: Disclaimer: I'm not an expert on this whole iterators business.
After reading https://en.cppreference.com/w/cpp/iterator/iterator_tags (and related), I think we should rather use `forward_iterator_tag`.
This is a general thing: input iterator means that the thing that is pointed to is (or may be) invalidated after you increment the iterator (I guess the analogy would be reading from `stdin` or so). We can however do things like `auto it = it2 = ...; ++it; auto x = *it2;`.
This comment most probably holds for all iterators in this PR.

Ouyancheng: Hmm... I will think about it. actually I'm not familiar with this either...

src/context/cdlist.h
```diff
@@ -384,7 +384,7 @@ class CDList : public ContextObj {
 
     // FIXME we support operator--, but do we satisfy all the
     // requirements of a bidirectional iterator ?
-    typedef std::input_iterator_tag iterator_category;
+    typedef std::forward_iterator_tag iterator_category;
```
nafur, line 387: This is actually bidirectional.
```suggestion
    typedef std::bidirectional_iterator_tag iterator_category;
```

nafur (approved):
One comment, otherwise LGTM

PR #7334, opened 2021-10-11
Minor cleaning of instantiation utilities

ajreynol:
Also fixes a bug in the quantifiers macro utility which did not compute the instantiation constant body of a quantified formula properly.

This is work towards a major refactoring of conflict-based instantiation / entailment checks.

### DIFF
--- a/src/theory/quantifiers/quantifiers_registry.cpp
+++ b/src/theory/quantifiers/quantifiers_registry.cpp
@@ -38,7 +38,7 @@ void QuantifiersRegistry::registerQuantifier(Node q)
   {
     return;
   }
-  Assert (q.getKind()==kind::FORALL);
+  Assert(q.getKind() == kind::FORALL);
   NodeManager* nm = NodeManager::currentNM();
   Debug("quantifiers-engine")
       << "Instantiation constants for " << q << " : " << std::endl;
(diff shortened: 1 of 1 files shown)


src/theory/quantifiers/quantifiers_registry.cpp
```diff
@@ -144,42 +144,53 @@ Node QuantifiersRegistry::substituteBoundVariablesToInstConstants(Node n,
...
-  return n.substitute(d_inst_constants[q].begin(),
-                      d_inst_constants[q].end(),
-                      terms.begin(),
-                      terms.end());
+  std::map<Node, std::vector<Node> >::const_iterator itc =
+      d_inst_constants.find(q);
+  Assert(itc != d_inst_constants.end());
+  Assert(itc->second.size() == terms.size());
+  return n.substitute(
+      itc->second.begin(), itc->second.end(), terms.begin(), terms.end());
```
4tXJ7f, line 193: Using the following pattern for this method and the methods above would be a bit clearer, IMHO:

```suggestion
  std::vector<Node>& consts = d_inst_constants.at(q);
  Assert(consts.size() == terms.size());
  return n.substitute(consts.begin(), consts.end(), terms.begin(), terms.end());
```

4tXJ7f (approved):
LGTM, just a suggestion.

PR #7336, opened 2021-10-11
Make (proof) equality engine use Env

ajreynol:


src/theory/arith/congruence_manager.cpp
```diff
@@ -81,14 +81,13 @@ void ArithCongruenceManager::finishInit(eq::EqualityEngine* ee)
   if (options().arith.arithEqSolver)
   {
     // use our own copy
-    d_allocEe.reset(
-        new eq::EqualityEngine(d_notify, context(), "arithCong::ee", true));
+    d_allocEe.reset(new eq::EqualityEngine(
+        d_env, context(), d_notify, "arithCong::ee", true));
```
nafur, line 85: Have you considered using the following?
```suggestion
    d_allocEe = std::make_unique<eq::EqualityEngine>(
        d_env, context(), d_notify, "arithCong::ee", true);
```

I'm aware we have this pattern all over our code base, but imho we should consider modernizing this.

src/theory/ee_manager.cpp
```diff
@@ -41,10 +41,11 @@ eq::EqualityEngine* EqEngineManager::allocateEqualityEngine(EeSetupInfo& esi,
   if (esi.d_notify != nullptr)
   {
     return new eq::EqualityEngine(
-        *esi.d_notify, c, esi.d_name, esi.d_constantsAreTriggers);
+        d_env, c, *esi.d_notify, esi.d_name, esi.d_constantsAreTriggers);
   }
   // the theory doesn't care about explicit notifications
-  return new eq::EqualityEngine(c, esi.d_name, esi.d_constantsAreTriggers);
+  return new eq::EqualityEngine(
+      d_env, c, esi.d_name, esi.d_constantsAreTriggers);
```
nafur, line 48: I realize that this is probably out of scope for this PR, but we should refactor it to return a `std::unique_ptr<>`.

src/theory/quantifiers/candidate_rewrite_filter.cpp
```diff
@@ -54,7 +55,7 @@ void CandidateRewriteFilter::initialize(SygusSampler* ss,
   ssn << "_dyn_rewriter_" << drewrite_counter;
   drewrite_counter++;
   d_drewrite = std::unique_ptr<DynamicRewriter>(
-      new DynamicRewriter(ssn.str(), &d_fakeContext));
+      new DynamicRewriter(d_env, &d_fakeContext, ssn.str()));
```
nafur, line 58: This should also be changed to `d_drewrite = std::make_unique<DynamicRewriter>(d_env, &d_fakeContext, ssn.str());`

src/theory/quantifiers/dynamic_rewrite.h
```diff
@@ -55,7 +58,7 @@ class DynamicRewriter
   typedef context::CDList<Node> NodeList;
 
  public:
-  DynamicRewriter(const std::string& name, context::Context* c);
+  DynamicRewriter(Env& env, context::Context* c, const std::string& name);
```
nafur, line 61: Is the context ever different from the env context?

src/theory/uf/equality_engine.h
```diff
@@ -70,19 +73,25 @@ class EqualityEngine : public context::ContextNotifyObj {
...
+   * @param c The context which this equality engine depends, which is typically
+   * although not necessarily same as the SAT context of env.
+   * @param name The name of this equality engine, for statistics
    * @param constantTriggers Whether we treat constants as trigger terms
    * @param anyTermTriggers Whether we use any terms as triggers
    */
-  EqualityEngine(EqualityEngineNotify& notify,
-                 context::Context* context,
+  EqualityEngine(Env& env,
+                 context::Context* c,
```
nafur, line 84: Ditto: `c == env.conext()`?

nafur (approved):
LGTM.

Issue #6973, opened 2021-08-04
(re-elim, re-elim-agg) Fatal failure in src/proof/conv_proof_generator.cpp:229 


zhendongsu:
Commit: https://github.com/cvc5/cvc5/commit/81544bb6de5b21db7ed5e41d56277105bbea103d
OS: Ubuntu 18.04

It is a regression from cvc4-1.8:
```
[522] % cvc4-1.8 -q --re-elim --re-elim-agg --check-unsat-cores small.smt2
unsat
[523] % cvc5 -q --re-elim --re-elim-agg --check-unsat-cores small.smt2
Fatal failure within virtual std::shared_ptr<cvc5::ProofNode> cvc5::TConvProofGenerator::getProofFor(cvc5::Node) at /local/suz-local/software/CVC4/src/proof/conv_proof_generator.cpp:229
Unhandled case encountered TConvProofGenerator::getProofFor: TheoryPreprocessor::preprocess_rewrite (policy=FIXPOINT, cache policy=NEVER, term-context-sensitive): failed, mismatch (see -t tconv-pf-gen-debug for details)
                   source: (let ((_let_1 (re.comp (re.* re.allchar )))) (let ((_let_2 (str.to_re a))) (let ((_let_3 (re.++ _let_2 _let_1))) (str.in_re "" (re.++ (re.inter _let_1 (re.comp _let_3)) (str.to_re (ite (str.in_re "" (re.++ (str.to_re (ite (str.in_re "" _let_3) a "")) (re.inter _let_2 _let_3))) a "")))))))
     requested conclusion: false
conclusion from generator: (let ((_let_1 (re.comp (re.* re.allchar )))) (str.in_re "" (re.++ (re.inter _let_1 (re.comp (re.++ (str.to_re a) _let_1))) (str.to_re (ite (and (= a "") false) a "")))))

Aborted
[524] % 
[524] % cat small.smt2
(declare-fun a () String)
(assert
 (str.in_re ""
  (re.++ (re.diff (re.comp re.all) (re.++ (str.to_re a) (re.comp re.all)))
   (str.to_re
    (ite
     (str.in_re ""
      (re.++ (str.to_re (ite (str.in_re "" (re.++ (str.to_re a) (re.comp re.all))) a ""))
       (re.inter (str.to_re a)
        (re.++ (str.to_re a)
         (re.comp (re.union (re.++ (str.to_re a) (re.comp re.all)) re.all))))))
     a "")))))
(check-sat)
[...]

PR #7345, opened 2021-10-12
Eliminate a level of nesting of traversals in theory preprocessing

ajreynol:
This simplifies the theory preprocessor so that it does not rely on the term formula removal to do a nested traversal.  Instead, it only relies on its "runCurrent" method.

Notice that this PR essentially replaces TheoryPreprocessor's `theoryPreprocess` method with TermFormulaRemoval's `runInternal` method, while adding the required call to `rewriteWithProof` and `preprocessWithProof` as post-rewrites.  This is far simpler than the previous method, which invoked nested traversals using `TermFormulaRemoval::run`.

There are two things that will be improved in follow up PRs:
- The initial full rewrite step in the theory preprocessor can be merged into the main traversal of theory preprocess (I believe this is why we are slower on nec benchmark than we were previously),
- We should eliminate the traversal methods from TermFormulaRemoval altogether, as they are now subsumed by the theory preprocessors main traversal.  This will require refactoring how "early ITE removal" works, as this is the only user now of TermFormulaRemoval::run.

Note we should probably performance test this change.

This incidentally fixes https://github.com/cvc5/cvc5/issues/6973, as the previous theory preprocessing code was to blame for that issue.


nafur (approved):
LGTM.

PR #7348, opened 2021-10-12
Refactor NlModel::solveEquality

nafur:
This PR refactors `NlModel::solveEquality`. While preserving it's general approach, it introduces a more uniform way to collect variable information and handle linear equalities. Some of the computations that were executed on Nodes that were known to hold rational constants are now done on Rationals.
Experiments show no discernible impact.


nafur:
Benchmark results show small improvements on `QF_NIA` (as we no longer bail out early for integer variables) and no changes anywhere else.

src/theory/arith/nl/nl_model.cpp
```diff
@@ -1311,6 +1020,263 @@ bool NlModel::hasLinearModelValue(TNode v, Node& val) const
   return false;
 }
 
+namespace
+{
+  void categorize_term(TNode mon, const Rational& coeff, Rational& constant, std::map<Node, std::tuple<Rational, Rational, bool>>& map)
```
ajreynol, line 1025: Document this method.

PR #7354, opened 2021-10-13
Eliminate uses of rewrite from datatypes theory

ajreynol:
Also simplifies slightly how rewritten forms of operators in sygus grammars are obtained.


4tXJ7f (approved):
LGTM

PR #7355, opened 2021-10-13
Eliminate last static calls to rewriter from smt layer

ajreynol:


src/proof/proof_node_manager.h
```diff
@@ -184,6 +188,8 @@ class ProofNodeManager
   static ProofNode* cancelDoubleSymm(ProofNode* pn);
 
  private:
+  /** The rewriter */
+  theory::Rewriter* d_rr;
```
aniemetz, line 192: Can we maybe use `d_rewriter` instead?

src/smt/witness_form.h
```diff
@@ -85,6 +92,8 @@ class WitnessFormGenerator : public ProofGenerator
    * Return a proof generator that can prove the given axiom exists.
    */
   ProofGenerator* convertExistsInternal(Node exists);
+  /** The rewriter we are using */
+  theory::Rewriter* d_rr;
```
aniemetz, line 96: Ditto

src/proof/proof_node_manager.h
ajreynol: Done.

PR #7357, opened 2021-10-13
Fix GLPK linking -- second try

nafur:
While #7341 added cmake-3.9 compatible linking for GLPK, it did not actually remove the line that triggers the error on older cmake versions. This PR does.


4tXJ7f (approved):
I just fixed the same issue locally in the same way and it worked on our nightly docker image.

PR #7358, opened 2021-10-13
Fix quantifiers variable elimination for parametric datatypes

### LINKED ISSUE #7353: Unknown error "Datatype type not fully instantiated"
Hi, I've encountered an undocumented error, which would be an internal error?

cvc5 069fde2

```
% cat input.smt2
(set-logic ALL)
(declare-datatype Option (par (T) ((none) (some (extractSome T)))))
(assert
 (forall ((v0010 (Option Int)))
         (=> (and ((_ is some) v0010)
                  (= (extractSome v0010) 0))
             (= v0010 (some 0)))))
(check-sat)
% cvc5 --lang smtlib2.6 input.smt2
(error "Datatype type not fully instantiated")
```

ajreynol:
Fixes #7353.


test/regress/regress0/quantifiers/issue7353-var-elim-par-dt.smt2
```diff
@@ -0,0 +1,9 @@
+(set-logic ALL)
+(set-info :status sat)
+(declare-datatype Option (par (T) ((none) (some (extractSome T)))))
+(assert
+ (forall ((x (Option Int)))
+         (=> (and ((_ is some) x)
+                  (= (extractSome x) 0))
+             (= x (some 0)))))
+(check-sat)
```
4tXJ7f, line 9: Please update CMakeLists.txt to include this test.

PR #7360, opened 2021-10-14
Add LFSC signature for n-ary programs

ajreynol:
This is the last utility signature. Theory signatures will be added afterwards.


yoni206 (commented):
Looks great. Some minor comments inline.

Another question:
Can you add some "unit tests" for these programs? e.g. for each program, calling it once with input that should fail and once with input that shouldn't. Similar to:

https://github.com/CVC4/CVC4-archived/blob/master/proofs/signatures/drat_test.plf

yoni206 (approved):
Great, thanks for the changes and for the creation of the project.

Issue #7252, opened 2021-09-24
Fatal failure at src/theory/arith/theory_arith.cpp:386


zhendongsu:
Commit: https://github.com/cvc5/cvc5/commit/7179be03b049d3046140316c4c5987efbdbd09b8
OS: Ubuntu 18.04 

```
[614] % cvc5 -q small.smt2
sat
[615] % cvc5 -q --strings-exp small.smt2
sat
[616] % cvc5 -q --force-logic=QF_NIA --strings-exp small.smt2
Fatal failure within bool cvc5::theory::arith::TheoryArith::sanityCheckIntegerModel() at /local/suz-local/software/CVC4/src/theory/arith/theory_arith.cpp:386
Check failure

 false
TheoryArithPrivate generated a bad model value for integer variable c : (/ (- 1) 2)
Aborted
[617] % cat small.smt2
(declare-fun a () Int)
(declare-fun b () Int)
(declare-fun c () Int)
(declare-fun d () Int)
(declare-fun e () Int)
(declare-fun f () Int)
(declare-fun g () Int)
(assert (> 0 (* a (mod 0 b))))
(assert (or (and (> (* b d) (* 2 (+ g c))) (= g (- c)) (> (+ e c) 0)) (> f 0)))
(check-sat)
[618] % 
```

Issue #6988, opened 2021-08-05
(produce-unsat-cores) Assertion error at src/theory/arith/theory_arith.cpp:268


rainoftime:
Hi, for the following instance,
cvc5 183808b608071890b4d
~~~~
(set-logic ANIA)
(declare-const x Bool)
(set-option :produce-unsat-cores true)
(declare-fun i () Int)
(declare-fun i5 () Int)
(declare-fun i8 () Int)
(assert (or x (< i5 0)))
(push)
(assert (and (= i8 1) (= i5 (+ 1 i)) (= i8 (+ 1 i))))
(push)
(check-sat)
(pop)
(pop)
(assert (= i8 (* i8 3 (+ i8 1))))
(check-sat)
~~~~
~~~~
sat
Fatal failure within virtual bool cvc5::theory::arith::TheoryArith::collectModelValues(cvc5::theory::TheoryModel*, const std::set<cvc5::NodeTemplate<true> >&) at /home/cvc5/src/theory/arith/theory_arith.cpp:268
Check failure

 false
TheoryArithPrivate generated a bad model value for integer variable i8 : (/ 3 2)
Aborted (core dumped)
~~~~

ajreynol:
This assertion failure demonstrates the patch from SMT COMP 2021 is necessary.  We should perhaps downgrade this to a warning.

I also might be helpful to bisect the error on this one to see what commit introduced these issues with real values for integer variables.

ajreynol:
This assertion failure has been downgraded to a warning in this PR: https://github.com/cvc5/cvc5/pull/7363.

PR #7363, opened 2021-10-14
Fix issue related to sanity checking integer models

### LINKED ISSUE #6774: (default) Invalid model for ALIRA formula
~~~~
(set-logic ALIRA)
(declare-const x Real)
(declare-fun i () Int)
(declare-fun i1 () Int)
(push)
(assert (< 1 (- i)))
(check-sat)
(pop)
(push)
(assert (or (>= i1 (* 5 (- i)))))
(check-sat)
(pop)
(assert (or (> i1 1) (= x (to_real i))))
(check-sat)
(assert (not (is_int x)))
(check-sat)
~~~~
~~~~
sat
sat
sat
Fatal failure within void cvc5::smt::CheckModels::checkModel(cvc5::smt::Model*, cvc5::context::CDList<cvc5::NodeTemplate<true> >*, bool) at /home/cvc5/src/smt/check_models.cpp:125
assertion:     (not (is_int x))
simplifies to: false
expected `true'.
Aborted (core dumped)
~~~~

ajreynol:
We now insist that we sanity check integer models in linear logics.  Previously, we could trigger an assertion failure in Minisat if a user asked for a model in a state where the linear solver had assigned a real value to an integer variable, as we would be sending a lemma during collectModelValues in a state where we had already terminated with "sat".

This also changes an assertion to warning, to allow the regression to pass.

Fixes #6774 (This PR fixes a followup issue after the original bad model reported there was fixed by sanity checks).

As the assertion failure is downgraded to a warning, fixes https://github.com/cvc5/cvc5/issues/6988, fixes https://github.com/cvc5/cvc5/issues/7252.

### DIFF
--- a/src/options/mkoptions.py
+++ b/src/options/mkoptions.py
@@ -480,17 +480,6 @@ def generate_module_option_names(module):
         'static constexpr const char* {name}__name = "{long_name}";', relevant)
 
 
-def generate_module_setdefaults_decl(module):
-    res = []
-    for option in module.options:
-        if option.name is None:
-            continue
-        funcname = option.name[0].capitalize() + option.name[1:]
-        res.append('void setDefault{}(Options& opts, {} value);'.format(
-            funcname, option.type))
-    return '\n'.join(res)
-
-
 ################################################################################
 # for options/<module>.cpp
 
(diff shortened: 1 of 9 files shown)


ajreynol:
No, since we still are delaying this check until model construction time if we need a last call effort check.

On second thought, we could either guard it by only running it if we haven't already (e.g. if needs last call effort is true).  Alternatively, we could have the policy that we always sanity check at full effort.

What do you think?

ajreynol:
Yeah, I do too, I'll change.

src/theory/arith/theory_arith.cpp
```diff
@@ -383,9 +386,9 @@ bool TheoryArith::sanityCheckIntegerModel()
       Trace("arith-check") << p.first << " -> " << p.second << std::endl;
       if (p.first.getType().isInteger() && !p.second.getType().isInteger())
       {
-        Assert(false) << "TheoryArithPrivate generated a bad model value for "
-                        "integer variable "
-                      << p.first << " : " << p.second;
+        Warning() << "TheoryArithPrivate generated a bad model value for "
+                     "integer variable "
+                  << p.first << " : " << p.second;
```
nafur, line 391: Why is this only a warning now?

nafur (approved):

PR #7369, opened 2021-10-14
Add regression for fixed issue

### LINKED ISSUE #4393: Invalid model for formula with codatatypes
[599] % cvc4 -q small.smt2
sat
[600] % cvc4 -q --check-unsat-cores --check-models small.smt2
Fatal failure within void CVC4::TheoryEngine::checkTheoryAssertionsWithModel(bool) at src/theory/theory_engine.cpp:2283
Internal error detectedTHEORY_DATATYPES has an asserted fact that the model doesn't satisfy.
The fact: (not (= f (b 0 f)))
Model value: false
Aborted
[601] % 
[601] % cat small.smt2
(declare-codatatypes ((a 0)) (((b (c Int) (d a)))))
(declare-fun e () a)
(declare-fun f () a)
(assert (distinct f (b 0 f)))
(assert (= e f))
(check-sat)
[602] %
OS: Ubuntu 18.04
Commit: 1cacd32

ajreynol:
Fixes #4393.

The issue does not occur on current master.

### DIFF
--- a/test/regress/CMakeLists.txt
+++ b/test/regress/CMakeLists.txt
@@ -498,6 +498,7 @@ set(regress_0_tests
   regress0/datatypes/is_test.smt2
   regress0/datatypes/issue1433.smt2
   regress0/datatypes/issue2838.cvc.smt2
+  regress0/datatypes/issue4393-cdt-model.smt2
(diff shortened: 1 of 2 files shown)


nafur (approved):
LGTM.

PR #7380, opened 2021-10-15
Initial refactoring of conflict-based instantiation

ajreynol:
This done an initial pass at refactoring conflict-based instantiation (the code has been largely unchanged since it was written in early 2014).  This code is a bottleneck on some Facebook benchmarks and needs revisiting.

This changes the style so the parent modules are made data members instead of passed to methods.  Otherwise, it changes `int` -> `size_t` wherever applicable, although `int` is often used to denote possibly valid indices where -1 means invalid.

It updates to use range-based for loops and mostly organizes the order of data members in classes to meet style guidelines.

Followup PRs will make more significant refactoring, e.g. to split the classes into multiple utilities.

### DIFF
--- a/cmake/version-base.cmake
+++ b/cmake/version-base.cmake
@@ -1,5 +1,5 @@
 # These are updated when making a release
-set(CVC5_LAST_RELEASE "0.0.5")
+set(CVC5_LAST_RELEASE "0.0.6")
 set(CVC5_IS_RELEASE "false")
 
 # These are used in other places in cmake
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -549,6 +549,8 @@ libcvc5_add_sources(
   theory/bags/bag_reduction.h
   theory/bags/bags_statistics.cpp
   theory/bags/bags_statistics.h
+  theory/bags/card_solver.cpp
+  theory/bags/card_solver.h
   theory/bags/infer_info.cpp
   theory/bags/infer_info.h
   theory/bags/inference_generator.cpp
--- a/src/options/base_options.toml
+++ b/src/options/base_options.toml
@@ -198,6 +198,11 @@ name   = "Base"
   help = "print top-level substitutions learned during preprocessing"
   description = "With ``-o subs``, cvc5 prints top-level substitutions learned during preprocessing."
   example-file = "regress0/printer/subs-output.smt2"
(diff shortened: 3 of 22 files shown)


src/theory/quantifiers/quant_conflict_find.cpp
```diff
@@ -595,15 +666,18 @@ bool QuantInfo::isTConstraintSpurious(QuantConflictFind* p,
...
+      inst = rewrite(inst);
       Node inst_eval =
           echeck->evaluateTerm(inst, options::qcfTConstraint(), true);
       if( Trace.isOn("qcf-instance-check") ){
         Trace("qcf-instance-check") << "Possible propagating instance for " << d_q << " : " << std::endl;
-        for( unsigned i=0; i<terms.size(); i++ ){
+        for (size_t i = 0, nterms = terms.size(); i < nterms; i++)
+        {
           Trace("qcf-instance-check") << "  " << terms[i] << std::endl;
         }
```
4tXJ7f, line 682:
```suggestion
        Trace("qcf-instance-check") << "  " << terms << std::endl;
```

src/theory/quantifiers/quant_conflict_find.h
```diff
@@ -309,6 +248,63 @@ class QuantConflictFind : public QuantifiersModule
    * algorithm performed by this class.
    */
   bool isPropagatingInstance(Node n) const;
+
+  enum Effort : unsigned
```
4tXJ7f, line 252: The indentation is off because of the indentation of `public:` above. Can you try indenting `public:` by one space and reapplying ClangFormat?

src/theory/quantifiers/quant_conflict_find.h
```diff
@@ -80,14 +53,12 @@ class MatchGen {
...
-  MatchGen( QuantInfo * qi, Node n, bool isVar = false );
+
   bool d_tgt;
   bool d_tgt_orig;
   bool d_wasSet;
   Node d_n;
-  std::vector< MatchGen > d_children;
+  std::vector<std::unique_ptr<MatchGen> > d_children;
   short d_type;
   bool d_type_not;
```
4tXJ7f, line 63: Data members go after method declarations.

ajreynol:
I finally managed to test this on SMT-LIB on Starexec, and the performance appears to be preserved.

PR #7381, opened 2021-10-15
Working on windows builds

nafur:
This PR aims to make our windows builds work again.
It's not done yet.


4tXJ7f (approved):
LGTM, I am still a little unclear about whether `-Dcvc5_obj_EXPORTS` is the right thing to do at the library level instead of doing it in a more fine-grained way (e.g., having a separate macro per library). Anyway, let's merge this for now and then see how things work out.

Issue #7374, opened 2021-10-14
(get-info :zzz) differs between CVC4 and CVC5


davidcok:
(get-info :zzz) produces

unsupported   on CVC4
(error "Unrecognized flag for getInfo.")   on CVC5

Because solver-specific info keywords are permitted, it seems most appropriate to return 'unsupported' for any unrecognized info label

**Command line arguments**: 
**cvc5 version/commit**: .0.0.2  1.18
**Operating system**: MacOS

PR #7384, opened 2021-10-17
Print `unsupported` for unrecognized flags.

abdoo8080:
This PR addresses issue #7374.


src/api/cpp/cvc5.cpp
```diff
@@ -6929,6 +6950,11 @@ std::string Solver::getInfo(const std::string& flag) const
 std::string Solver::getOption(const std::string& option) const
 {
   CVC5_API_TRY_CATCH_BEGIN;
+  std::vector<std::string> options = options::getNames();
+  CVC5_API_UNSUPPORTED_CHECK(
+      option.find("command-verbosity") != std::string::npos
```
aniemetz, line 6955: Why this extra handling?

aniemetz: @nafur do you know why `command-verbosity` is separate and not in `options`?

nafur: Hm, I've never changed it. It has a special syntax (`(<command-name> verbosity)`), but we can do that with a option handler. I'll put in on my todo list...

aniemetz (approved):

PR #7388, opened 2021-10-18
Fix error management of `ErrorInformation`

4tXJ7f:
Fixes
https://scan6.coverity.com/reports.htm#v37053/p11644/fileInstanceId=125548448&defectInstanceId=32441274&mergedDefectId=1453884.


4tXJ7f:
Makes sense, I've addressed it.

PR #7390, opened 2021-10-18
Do not construct instantiation for checking propagating instantiations spurious

ajreynol:
This makes the check for when an instantiation is "propagating" faster by not constructing the substitution + rewriting of the entire formula, and instead threading the substitution through the entailment check utility's evaluateTerm utility.

On a handful of challenge Facebook benchmarks, we go 35 seconds -> 18 seconds with this change.

This also eliminates the argument `exp` to the evaluateTerm method, which is no longer used, and eliminates `hasSubs` from several methods, which is redundant.


src/theory/quantifiers/entailment_check.h
```diff
@@ -44,32 +44,41 @@ class EntailmentCheck : protected EnvObj
...
-   * Returns a term n' such that n = n' is entailed based on the equality
-   * information ee.  This function may generate new terms. In particular,
-   * we typically rewrite subterms of n of maximal size to terms that exist in
-   * the equality engine specified by ee.
+   * Returns a term n' such that n * subs = n' is entailed based on the current
+   * set of equalities, where ( n * subs ) is term n under the substitution
+   * subs.
+   *
+   * This function may generate new terms. In particular, we typically rewrite
+   * subterms of n of maximal size to terms that exist in the equality engine.
```
HanielB, line 52: Clarify "subterms of maximal size"

src/theory/quantifiers/quant_conflict_find.cpp
```diff
@@ -578,29 +572,31 @@ bool QuantInfo::isTConstraintSpurious(QuantConflictFind* p,
...
         return true;
       }
     }else{
-      Node inst =
-          getInferenceManager().getInstantiate()->getInstantiation(d_q, terms);
-      inst = Rewriter::rewrite(inst);
-      Node inst_eval =
-          echeck->evaluateTerm(inst, options::qcfTConstraint(), true);
+      // see if the body of the quantified formula evaluates to a Boolean
+      // combination of known terms under the current substitution.
```
HanielB, line 597: It'd be good to clarify how this is computed. The check of line 608 requires some thinking to understand.

HanielB (changes requested):

PR #7391, opened 2021-10-18
Move check for experimental arrays features to `theory_arrays.cpp`.

abdoo8080:
Check for experimental array features was done at the parser module, which meant that users of the API could use them without calling `Solver::setOption("arrays-exp")`. This PR fixes that by moving the check to the internal theory module.

### DIFF
--- a/src/theory/arrays/theory_arrays.cpp
+++ b/src/theory/arrays/theory_arrays.cpp
@@ -298,7 +298,7 @@ TrustNode TheoryArrays::ppRewrite(TNode term, std::vector<SkolemLemma>& lems)
 {
   // first, check for logic exceptions
   Kind k = term.getKind();
-  if (!options::arraysExp())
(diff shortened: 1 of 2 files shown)


src/theory/arrays/theory_arrays.cpp
```diff
@@ -296,7 +296,19 @@ Node TheoryArrays::solveWrite(TNode term, bool solve1, bool solve2, bool ppCheck
 
 TrustNode TheoryArrays::ppRewrite(TNode term, std::vector<SkolemLemma>& lems)
 {
-  // first, see if we need to expand definitions
+  // first, check for logic exceptions
+  Kind k = term.getKind();
+  if (!options::arraysExp())
```
ajreynol, line 301:
```suggestion
  if (!options().arrays.arraysExp)
```

PR #7393, opened 2021-10-18
Correctly parse uninterpreted constant values in get-value

### LINKED ISSUE #6908: Cannot reference abstract value in get-value
**Describe the bug**
When an abstract value is produced as part of the model, that value cannot be referenced inside a `(get-value ...)` command.
Page 49 of the smt-lib 2.6 spec states that this ought to be doable.

Sample input:
```smt2
(declare-sort Foo 0)
(declare-fun f (Foo) Int)
(assert (exists ((x Foo)) (= (f x) 3)))
(check-sat)
(get-model)
(get-value (f @uc_Foo_0))
```

Output:
```
test.smt:2.13: No set-logic command was given before this point.
test.smt:2.13: cvc5 will make all theories available.
test.smt:2.13: Consider setting a stricter logic for (likely) better performance.
test.smt:2.13: To suppress this warning in the future use (set-logic ALL).
sat
(
; cardinality of Foo is 1
; rep: (as @uc_Foo_0 Foo)
(define-fun f ((_arg_1 Foo)) Int 3)
)
(error "Parse Error: test.smt:7.24: Symbol @uc_Foo_0 is not declared.

  (get-value ((f @uc_Foo_0)))
                 ^
")
```

**Command line arguments**: `--lang smt2`
**CVC5 version/commit**: Current master: 17fc4c975cefba7082e5557cb9c9515f16b44096
**Operating system**:  Archlinux

ajreynol:
SMT-LIB indicates that abstract values can appear in terms as arguments to `get-value`.  This is not currently the case, as pointed out by #6908.

This updates our parser so that bindings are added to the symbol table for all uninterpreted constants in the model whenever we parse a `get-value` term.

Fixes #6908.

### DIFF
--- a/src/parser/parser.cpp
+++ b/src/parser/parser.cpp
@@ -744,20 +744,22 @@ void Parser::pushGetValueScope()
   // the set of uninterpreted constants that are printed in the definition
   // of a model.
   std::vector<api::Sort> declareSorts = d_symman->getModelDeclareSorts();
-  Trace("parser") << "Push get value scope, with " << declareSorts.size() << " declared sorts" << std::endl;
+  Trace("parser") << "Push get value scope, with " << declareSorts.size()
+                  << " declared sorts" << std::endl;
   for (const api::Sort& s : declareSorts)
   {
     std::stringstream uprefix;
     uprefix << "@uc_" << s.toString() << "_";
     std::vector<api::Term> elements = d_solver->getModelDomainElements(s);
-    for (size_t i = 0, nelements = elements.size(); i<nelements; i++)
+    for (size_t i = 0, nelements = elements.size(); i < nelements; i++)
     {
       std::stringstream en;
       en << uprefix.str() << i;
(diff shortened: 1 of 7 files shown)


src/parser/parser.cpp
```diff
@@ -737,6 +737,36 @@ void Parser::pushScope(bool isUserContext)
...
+      // Uninterpreted constants are abstract values, which by SMT-LIB are
+      // required to be annotated with their type, e.g. (as @uc_Foo_0 Foo).
+      // Thus, the element is not printed simply as its name.
+      std::string en = e.toString();
+      size_t index = en.find("(as ");
+      if (index == 0)
+      {
+        index = en.find(" ", 4);
+        en = en.substr(4, index - 4);
+      }
```
4tXJ7f, line 763: Can't we get the name by looking into the term? I think that it would be better to add a method to retrieve the name otherwise. Parsing the string seems error prone, e.g. what if the user chooses a different output language?

PR #7409, opened 2021-10-19
Avoid escaping `double-quotes` twice.

abdoo8080:
With `-o raw-benchmark`, The command:
```smt2
(echo """")
```
was printed as
```smt2
(echo """""""")
```
because we run the quote-escaping procedure twice on it. This PR fixes the issue by ensuring that we only run it once.


ajreynol (approved):
LGTM

PR #7411, opened 2021-10-19
Make proofs of rewrites robust to use in internal subsolvers

ajreynol:
We are currently using an attribute to mark "rewritten with proofs".  This is incorrect as attributes are global, but proofs of rewrites are local to the solver instance.

This enables applications of proofs within internal subsolvers, and is required for a new internal fuzzing technique.


4tXJ7f (approved):
LGTM

Issue #6605, opened 2021-05-22
(produce-abducts) Fatal failure at src/theory/eager_proof_generator.cpp:35 


zhendongsu:
Commit: 0e9fed3
OS: Ubuntu 18.04

```
[532] % cvc4-1.8 -q --check-unsat-cores small.smt2
(define-fun d () Bool (= b a))
[533] % cvc5 -q --check-unsat-cores small.smt2
Fatal failure within void cvc5::theory::EagerProofGenerator::setProofFor(cvc5::Node, std::shared_ptr<cvc5::ProofNode>) at /local/suz-local/software/CVC4/src/theory/eager_proof_generator.cpp:35
Check failure

 pf->getResult() == f
EagerProofGenerator::setProofFor: unexpected result
Expected: (>= rsk_11 1)
Actual: (not (not (>= rsk_11 1)))

Aborted
[534] % 
[534] % cat small.smt2
(set-option :produce-abducts true)
(declare-fun a () Int)
(declare-fun b () Int)
(declare-fun c () Int)
(assert (= (>= b c) (>= 0 a)))
(assert (= c a))
(get-abduct d (<= a b))
[535] % 
```

rainoftime:
**src/smt/abduction_solver.cpp:108**
~~~~
(get-abduct A true)
~~~~
cvc5  -q -i --produce-abducts delta.out.smt2 
Fatal failure within bool cvc5::smt::AbductionSolver::getAbductInternal(cvc5::Node&) at /home/cvc5/src/smt/abduction_solver.cpp:108
Check failure

 !agdtbv.isNull()

Aborted
~~~~

rainoftime:
**src/theory/uf/equality_engine.cpp:585**
~~~~
(declare-sort S0 0)
(declare-sort S2 0)
(declare-fun arr0 () (Array Bool S2))
(declare-fun arr1 () (Array S2 (Array S0 (Array S0 (Array S0 Bool)))))
(declare-fun arr2 () (Array Bool (Array Bool (Array S0 (Array S0 (Array S0 Bool))))))
(get-abduct A (= (select (select arr2 false) true) (select arr1 (select arr0 false))))
~~~~
~~~~
cvc5 -q --produce-abducts x.smt2 
Fatal failure within cvc5::TNode cvc5::theory::eq::EqualityEngine::getRepresentative(cvc5::TNode) const at /home/cvc5/src/theory/uf/equality_engine.cpp:585
Check failure

 hasTerm(t)

[...]

PR #7417, opened 2021-10-19
Make SyGuS solver robust to non-closed enumerable sorts

ajreynol:
This makes the SyGuS solver robust to variables that are not closed enumerable, e.g. arrays of uninterpreted sorts.

It corrects an issue in array's mkGroundTerm issue which would allow constants to enter into constraints for SyGuS problems with arrays.  This method does not cause further issues currently since quantifiers is guarded in several places to ensure array constants are not constructed via this method.

It also makes it so that we don't add explicit CEGIS refinement lemmas unless evaluation unfolding is enabled *and* the counterexamples are from closed enumerable types; there is no reason to add these unless we are combining with evaluation unfolding.

This addresses several of the issues raised in #6605.

### DIFF
--- a/src/api/cpp/cvc5.cpp
+++ b/src/api/cpp/cvc5.cpp
@@ -7205,7 +7205,7 @@ std::string Solver::getProof(void) const
 {
   CVC5_API_TRY_CATCH_BEGIN;
   CVC5_API_CHECK(d_slv->getOptions().smt.produceProofs)
-      << "Cannot get proof explicitly enabled (try --produce-proofs)";
+      << "Cannot get proof unless proofs are enabled (try --produce-proofs)";
   CVC5_API_RECOVERABLE_CHECK(d_slv->getSmtMode() == SmtMode::UNSAT)
       << "Cannot get proof unless in unsat mode.";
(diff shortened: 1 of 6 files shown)


src/theory/arrays/theory_arrays_type_rules.cpp
```diff
@@ -249,7 +250,24 @@ bool ArraysProperties::isWellFounded(TypeNode type)
...
+  {
+    return NodeManager::currentNM()->mkConst(ArrayStoreAll(type, elem));
+  }
+  // Note the distinction between mkGroundTerm and mkGroundValue. While
+  // an arbitrary value can be obtained by calling the type enumerator here,
+  // that is wrong for types that are not closed enumerable since it may
+  // return a term containing values that should not appear in e.g. assertions.
+  // For example, arrays whose element type is an uninterpreted sort will
+  // incorrectly introduce uninterpreted sort values if this is done.
+  // It is current infeasible to construct an ArrayStoreAll with the element
```
HanielB, line 266:
```suggestion
  // It is currently infeasible to construct an ArrayStoreAll with the element
```

PR #7420, opened 2021-10-19
Fix inadvertent failure of workflow step

nafur:
This PR fixes an issue introduced with #7352: jobs fail immediately if any command has a non-zero exit code. The way we obtained the exit code of the `diff` command was incompatible with this policy, which is why we currently never upload any docs to `docs-ci`.


4tXJ7f (approved):
LGTM

PR #7424, opened 2021-10-20
Unify abstract values and uninterpreted constants

4tXJ7f:
This commit unifies abstract values and "uninterpreted constants" into a
single kind. Note that "uninterpreted constants" is a bit of a misnomer
in the context of the new API, since they do not correspond to the
equivalent of a `declare-const` command, but instead are values for
symbols of an uninterpreted sort (and thus special cases of abstract
values). Instead of treating "uninterpreted constants" as a separate
kind, this commit extends abstract values to hold a type (instead of
marking their type via attribute in `NodeManager::mkAbstractValue()`)
and uses the type of the abstract values to determine whether they are a
value for a constant of uninterpreted sort or not. Unifying these
representations simplifies code and brings the terminology more in line
with the SMT-LIB standard.

This commit also updates the APIs to remove support for creating
abstract values and "uninterpreted constants". Users should never create
those. They can only be returned as a value for a term after a
satisfiability check.

Finally, the commit removes code in the parser for parsing abstract
values and updates the code for getting values involving abstract
values. Since the parser does not allow the declaration/definition of
abstract values, and determining whether a symbol is an abstract value
was broken (not all symbols starting with an `@` are abstract values),
the code was effectively dead. Getting values involving "uninterpreted
constants" now no longer requires parsing the string of the values, but
instead, we can use existing API functionality.


src/api/cpp/cvc5.cpp
```diff
@@ -3309,31 +3308,6 @@ std::vector<Term> Term::getSequenceValue() const
   CVC5_API_TRY_CATCH_END;
 }
 
-bool Term::isUninterpretedValue() const
```
ajreynol, line 3312: I think we should keep these 2 methods.  They are analogous to other model getters for values, regardless of how values for uninterpreted sorts are represented.

src/api/cpp/cvc5_kind.h
```diff
@@ -64,18 +64,7 @@ enum CVC5_EXPORT Kind : int32_t
...
-   * Uninterpreted constant.
-   *
-   * Parameters:
-   *   - 1: Sort of the constant
-   *   - 2: Index of the constant
-   *
-   * Create with:
-   *   - `Solver::mkUninterpretedConst(const Sort& sort, int32_t index) const`
-   */
-  UNINTERPRETED_CONSTANT,
```
ajreynol, line 76: I assume what used to be uninterpreted constants now has kind ABSTRACT_VALUE?

src/theory/evaluator.cpp
```diff
@@ -44,10 +44,7 @@ EvalResult::EvalResult(const EvalResult& other)
       new (&d_str) String;
       d_str = other.d_str;
       break;
-    case UCONST:
-      new (&d_uc)
-          UninterpretedConstant(other.d_uc.getType(), other.d_uc.getIndex());
-      break;
+    case ABSTRACT: new (&d_av) UninterpretedSortValue(other.d_av); break;
```
ajreynol, line 47: I don't like the name `ABSTRACT` here, how about `UVALUE`?

src/theory/quantifiers/sygus/sygus_grammar_cons.cpp
```diff
@@ -439,8 +439,6 @@ void CegGrammarConstructor::mkSygusConstantsForType(TypeNode type,
   {
     // generate constant array over the first element of the constituent type
     Node c = nm->mkGroundTerm(type);
-    // note that c should never contain an uninterpreted constant
-    Assert(!expr::hasSubtermKind(UNINTERPRETED_CONSTANT, c));
```
ajreynol, line 443: Can this assertion be preserved?

src/theory/uf/theory_uf.cpp
```diff
@@ -294,17 +294,17 @@ void TheoryUF::preRegisterTerm(TNode node)
...
-    // back in constraints, hence this check is necessary.
+    // Uninterpreted sort values should only appear in models, and should never
+    // appear in constraints. They are unallowed to ever appear in constraints
+    // since the cardinality of an uninterpreted sort may have an upper bound,
+    // e.g. if (forall ((x U) (y U)) (= x y)) holds, then @uc_U_2 is a
+    // ill-formed term, as its existence cannot be assumed.  The parser
+    // prevents the user from ever constructing abstract values. However, they
+    // may be exported via models to API users. It is thus possible that these
+    // abstract values are asserted back in constraints, hence this check is
+    // necessary.
```
ajreynol, line 307: Update word "constant" in the line below.

ajreynol (changes requested):
I like the change to make getUninterpretedSortValue return a string.

A few minor requests.

ajreynol (approved):

PR #7425, opened 2021-10-20
Use codatatype bound variables for codatatype values

ajreynol:
Previously, we had conflated codatatype bound variables and uninterpreted constants (as they have the same functionality). However, we should separate these concepts so we can ensure that uninterpreted constants are only used for uninterpreted sorts.

### DIFF
--- a/src/expr/codatatype_bound_variable.h
+++ b/src/expr/codatatype_bound_variable.h
@@ -1,0 +1,91 @@
+/******************************************************************************
+ * Top contributors (to current version):
+ *   Andrew Reynolds
+ *
+ * This file is part of the cvc5 project.
+ *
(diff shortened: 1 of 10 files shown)


src/expr/codatatype_bound_variable.h
```diff
@@ -0,0 +1,91 @@
...
+
+class TypeNode;
+
+/**
+ * In theory, codatatype values are represented in using a MU-notation form,
+ * where recursive values may contain free constants indexed by their de Bruijn
+ * indices. This is sometimes written:
+ *    MU x. (cons 0 x).
+ * where the MU binder is label for a term position, which x references.
+ * Instead of construct an explicit MU binder (which is problematic for
```
4tXJ7f, line 36:
```suggestion
 * Instead of constructing an explicit MU binder (which is problematic for
```

PR #7426, opened 2021-10-20
Make expression mining use configurable options and logic

ajreynol:
Required for doing options-specific internal fuzzing using SyGuS.


src/theory/quantifiers/expr_miner.cpp
```diff
@@ -52,18 +52,24 @@ Node ExprMiner::convertToSkolem(Node n)
...
 {
   Assert (!query.isNull());
   if (Options::current().quantifiers.sygusExprMinerCheckTimeoutWasSetByUser)
   {
-    initializeSubsolver(checker,
-                        d_env,
-                        true,
-                        options::sygusExprMinerCheckTimeout());
+    initializeSubsolver(
+        checker, opts, logicInfo, true, options::sygusExprMinerCheckTimeout());
```
4tXJ7f, line 68: It feels a bit dangerous that if the expression miner check timeout is set, the passed options are ignored.

4tXJ7f (approved):
LGTM, I just have a concern that we can address in a future PR if needed.

src/theory/quantifiers/expr_miner.cpp
ajreynol: What do you mean? We do carry `opts`, `logicInfo` here?

src/theory/quantifiers/expr_miner.cpp
```diff
@@ -52,18 +52,24 @@ Node ExprMiner::convertToSkolem(Node n)
...
+  initializeChecker(checker, query, options(), logicInfo());
+}
+
+void ExprMiner::initializeChecker(std::unique_ptr<SolverEngine>& checker,
+                                  Node query,
+                                  const Options& opts,
+                                  const LogicInfo& logicInfo)
 {
   Assert (!query.isNull());
   if (Options::current().quantifiers.sygusExprMinerCheckTimeoutWasSetByUser)
```
nafur, line 65: Can we use the `opts` object here instead of using `options::` and `Options::current()`.

ajreynol: Yeah, I'll clean this on a separate PR, this is old.

PR #7427, opened 2021-10-20
Fix `(set-info <sexpr>)` parsing and printing bugs.

abdoo8080:
Given
```smt2
(set-info :source (0 1 True False x ""))
```
`cvc5` currently prints the command below with `-o raw-benchmark`
```smt2
(set-info : |(0 1 True False x )|)
```
This PR ensures that `cvc5` correctly prints the command by
- Parsing and storing keywords eagerly before parsing their values.
- Removing pre-processing steps done to symbols and string literals.


src/parser/smt2/Smt2.g
```diff
@@ -728,10 +728,9 @@ setInfoInternal[std::unique_ptr<cvc5::Command>* cmd]
   std::string name;
   api::Term sexpr;
 }
-  : KEYWORD symbolicExpr[sexpr]
-    { name = AntlrInput::tokenText($KEYWORD);
-      cmd->reset(new SetInfoCommand(name.c_str() + 1, sexprToString(sexpr)));
-    }
+  : KEYWORD { name = AntlrInput::tokenText($KEYWORD); }
+    symbolicExpr[sexpr]
+    { cmd->reset(new SetInfoCommand(name.c_str() + 1, sexprToString(sexpr))); }
```
4tXJ7f, line 733: It is not quite clear to me why this change is needed. Could you explain?

abdoo8080: I am not too sure either. But it looks like the value stored in `KEYWORD` gets overridden somewhere in `symbolicExpr[sexpr]`, hence, the missing `source` attribute in the printed command. Andy suggested that I immediately store its value `name` and that seems to solve the problem.

PR #7436, opened 2021-10-20
Enable CI for Junit tests

mudathirmahgoub:
This PR enables CI for java tests by adding  --java-bindings to ci.yml.
It also replaces the unreliable `finalize` method and instead uses `AutoCloseable` and explicit `close` method to clean up dynamic memory allocated by `java native interface`. 
The PR fixes compile errors for `SolverTest.java` and runtime errors for `Solver.defineFun`.


.github/workflows/ci.yml
```diff
@@ -58,6 +59,10 @@ jobs:
     steps:
 
     - uses: actions/checkout@v2
+    - uses: actions/setup-java@v2
+      with:
+        distribution: 'temurin'
+        java-version: '11'
```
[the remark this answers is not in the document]
mudathirmahgoub: I thought we needed this for junit5. It turned out we don't, since last commit passed.

src/api/java/jni/solver.cpp
```diff
@@ -39,7 +39,7 @@ JNIEXPORT jlong JNICALL Java_io_github_cvc5_api_Solver_newSolver(JNIEnv*,
 JNIEXPORT void JNICALL
 Java_io_github_cvc5_api_Solver_deletePointer(JNIEnv*, jclass, jlong pointer)
 {
-  delete (reinterpret_cast<Solver*>(pointer));
+  // delete (reinterpret_cast<Solver*>(pointer));
```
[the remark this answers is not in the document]
mudathirmahgoub: Fixed  by `Solver.close()` method.

.github/workflows/ci.yml
```diff
@@ -38,7 +38,7 @@ jobs:
 
           - name: ubuntu:production-dbg
             os: ubuntu-18.04
-            config: production --auto-download --assertions --tracing --unit-testing --editline
+            config: production --auto-download --assertions --tracing --unit-testing --java-bindings --editline
```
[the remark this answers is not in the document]
mudathirmahgoub: I think python tests are already see this [log](https://github.com/cvc5/cvc5/runs/4075609468?check_suite_focus=true#step:7:5411).
Java tests are enabled in this [log](https://github.com/cvc5/cvc5/runs/4075609532?check_suite_focus=true#step:7:8734).

src/api/java/io/github/cvc5/api/Solver.java
```diff
@@ -31,14 +31,32 @@ public long getPointer()
...
+  // store pointers for terms, sorts, etc
+  List<AbstractPointer> abstractPointers = new ArrayList<>();
+
+  @Override public void close()
+  {
+    // delete heap memory for terms, sorts, etc
+    for (AbstractPointer abstractPointer : abstractPointers)
+    {
+      abstractPointer.deletePointer();
+    }
```
nafur, line 52: It struck me that this loop makes us free objects in the same order as they were created, while one would usually use the opposite order. We might be able to argue that the objects have no inter-dependencies and thus the destruction order does not matter.
I'd suggest to simply delete in reverse order, this should always be safe, even if we introduce inter-dependencies.

examples/api/java/BitVectorsAndArrays.java
```diff
@@ -95,5 +95,6 @@ public static void main(String[] args) throws CVC5ApiException
     System.out.println("The satisfying model is: ");
     System.out.println("  current_array = " + slv.getValue(current_array));
     System.out.println("  current_array[0] = " + slv.getValue(current_array0));
+    slv.close();
```
nafur, line 98: I would suggest (here, and in all other examples) to use the "try-with-resources" idiom instead of calling `close()` manually.

src/api/java/io/github/cvc5/api/Solver.java
mudathirmahgoub: Done.

examples/api/java/BitVectorsAndArrays.java
mudathirmahgoub: Done.

src/api/java/io/github/cvc5/api/Datatype.java
```diff
@@ -24,20 +24,16 @@
   Datatype(Solver solver, long pointer)
   {
     super(solver, pointer);
+    solver.addAbstractPointer(this);
```
nafur, line 27: Can we move call into the `AbstractPointer` base class?

test/unit/api/java/SolverTest.java
```diff
@@ -102,6 +107,7 @@
 
     Solver slv = new Solver();
     assertThrows(CVC5ApiException.class, () -> slv.mkArraySort(boolSort, boolSort));
+    slv.close();
```
nafur, line 110: Use try-with-resource here.

test/unit/api/java/SolverTest.java
```diff
@@ -132,6 +138,7 @@
 
     DatatypeDecl throwsDtypeSpec = d_solver.mkDatatypeDecl("list");
     assertThrows(CVC5ApiException.class, () -> d_solver.mkDatatypeSort(throwsDtypeSpec));
+    slv.close();
```
nafur, line 141: Use try-with-resource here.

test/unit/api/java/SolverTest.java
```diff
@@ -175,6 +182,7 @@
 
     assertThrows(
         CVC5ApiException.class, () -> slv.mkDatatypeSorts(Arrays.asList(udecls), unresSorts));
+    slv.close();
```
nafur, line 185: Ditto, and some more below.

test/unit/api/java/SolverTest.java
nafur: Why are we using a separate solver object here?
It seems that this test mostly checks whether we can use a sort object from solver object a and use it on solver object b... (this is the case for a lot of the tests here). Is this the intention?

src/api/java/io/github/cvc5/api/Datatype.java
mudathirmahgoub: Done.

test/unit/api/java/SolverTest.java
mudathirmahgoub: I think so.  I am just following the corresponding test in `solver_black.cpp`. 
However, I am not a fan of using  `try-with-resource` here. Unlike examples which is written for users, these are unit tests and I don't like the extra indentation that come with `try-with-resources`. 
 I am already using `solver.close` [here](https://github.com/cvc5/cvc5/pull/7436/files#diff-0f2f1183d810770d1917a0457a53e85ddf5cb52c593c50798d8b63ac1ef6cc66R40)
, so I prefer to be consistent here.

test/unit/api/java/SolverTest.java
mudathirmahgoub: same reason above.

test/unit/api/java/SolverTest.java
mudathirmahgoub: same reason above.

aniemetz (approved):

PR #7449, opened 2021-10-21
Start refactoring of `-o` and `-v`

nafur:
This PR does the first step in consolidating our various output mechanisms. The goal is to have only three major ways to output information:
- `verbose(level)` prints log-like information to the user via stderr, `-v` and `-q` increase and decrease the verbosity, respectively.
- `output(tag)` prints structured information (as s-expressions) to the user via stdout, `-o` enables individual tags.
- `Trace(tag)` prints log-like information to the developer via stdout, `-t` enables individual tags.
While `Debug` and `Trace` (and eventually `Dump`) will be combined within `Trace`, all of `Warning`, `Message`, `Notice` and `Chat` will be combined into `verbose`.


src/smt/sygus_solver.cpp
```diff
@@ -257,8 +257,11 @@ void SygusSolver::checkSynthSolution(Assertions& as)
 {
   NodeManager* nm = NodeManager::currentNM();
   SkolemManager* sm = nm->getSkolemManager();
-  Notice() << "SygusSolver::checkSynthSolution(): checking synthesis solution"
-           << std::endl;
+  if (isOutputOn(OutputTag::SYGUS))
```
ajreynol, line 260: This should be a different (new) output tag.  Something like `-o solver`, which summarizes what the (submodules) of SolverEngine are doing.

ajreynol (changes requested):
One request regarding the specifics of the output tag, otherwise this looks good.

src/smt/sygus_solver.cpp
nafur: Done.

src/theory/datatypes/theory_datatypes.cpp
```diff
@@ -382,8 +382,9 @@ void TheoryDatatypes::postCheck(Effort level)
   }
 
   Trace("datatypes-check") << "Finished check effort " << level << std::endl;
-  if( Debug.isOn("datatypes") || Debug.isOn("datatypes-split") ) {
-    Notice() << "TheoryDatatypes::check(): done" << endl;
+  if (isVerboseOn(1))
```
ajreynol, line 385: Let's just change this to Debug("datatypes"), im not sure why Notice() was used here.

src/theory/theory_engine.cpp
```diff
@@ -1892,7 +1892,7 @@ void TheoryEngine::checkTheoryAssertionsWithModel(bool hardFailure) {
               // assertions with unevaluable operators, e.g. transcendental
               // functions. It also may happen for separation logic, where
               // check-model support is limited.
-              Warning() << ss.str();
+              verbose(0) << ss.str();
```
ajreynol, line 1895: Can we make `warning()` short for `verbose(0)` in EnvObj?

ajreynol (changes requested):

ajreynol (approved):

PR #7512, opened 2021-10-27
Initial steps towards a 'native' Windows build

aytey:
## Overview

This PR adds some initial steps towards getting a `cvc5` build working natively on Windows. It follows the approach for: https://github.com/Boolector/boolector/blob/master/COMPILING_WINDOWS.md

This has been tested with:

```
$ gcc --version
gcc.exe (Rev9, Built by MSYS2 project) 10.2.0
Copyright (C) 2020 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

and

```
$ cmake --version
cmake version 3.21.4

CMake suite maintained and supported by Kitware (kitware.com/cmake).
```

on:

```
> ver

Microsoft Windows [Version 10.0.19042.1288]
```

and it allows the following to successfully build and test:

```
./configure.sh  --auto-download --win64-native
cd build
make -j$(nproc)
find . -iname "*.dll" -exec cp {} bin \;    # this is necessary on Windows because there's no notion of `LD_LIBRARY_PATH`
make test
```

## Things I don't like

- [ ] Do we have a better way of finding/setting `SHELL` before calling `configure` for ANTLR?
- [ ] Do want to set `SHELL` to be `""` on non-Windows or leave it as an undefined variable (so it expands to nothing)?
- [ ] Can we add automated steps to copy the DLLs into the `bin` directory on Windows?
- [ ] The changes to `cvc5`'s `configure.sh` are 🤮 but I couldn't find any way to add `MSYS Makefiles` to `cmake_opts` and for it to still work

## Things I haven't done

- [ ] Run a full `make test` and see what else does/doesn't work

### DIFF
--- a/docs/api/python/python.rst
+++ b/docs/api/python/python.rst
@@ -41,7 +41,16 @@ The base and pythonic Python API can be installed via `pip` as follows:
 Installation (ARM64 variants of Linux and macOS)
 ------------
 
-For ARM64-based machines (including Apple computers with M1 and M2 chips), the base and the pythonic Python API can be installed from source as follows:
+For ARM64-based machines (including Apple computers with M1 and M2 chips), the base and the pythonic Python API can be installed from source.
+Before building and installing, the following dependencies should be installed, using `brew` and `pip`:
+
+.. code:: bash
+
+  brew install cmake python cython gmp java
+  pip3 install toml scikit-build
+
+
+Then `cvc5` can be installed from source as follows:
 
 .. code:: bash
 
--- a/src/theory/bv/theory_bv_rewrite_rules_simplification.h
+++ b/src/theory/bv/theory_bv_rewrite_rules_simplification.h
@@ -2167,6 +2167,8 @@ inline Node RewriteRule<UltAddOne>::apply(TNode node)
 /* -------------------------------------------------------------------------- */
 
 /**
+ * MultSltMult
+ *
  * Rewrite
  *   sign_extend(x+t,n) * sign_extend(a,m) < sign_extend(x,n) * sign_extend(a,m)
  * to
@@ -2191,8 +2193,9 @@ inline Node RewriteRule<UltAddOne>::apply(TNode node)
  * where the BV engine struggles due to the high bit widths of the
  * multiplication's operands.
  */
-static std::tuple<Node, Node, bool>
-extract_ext_tuple(TNode node)
+
+namespace {
+std::tuple<Node, Node, bool> extract_ext_tuple(TNode node)
 {
   TNode a = node[0];
   TNode b = node[1];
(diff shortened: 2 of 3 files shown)


test/regress/CMakeLists.txt
```diff
@@ -2859,9 +2859,10 @@ add_custom_target(regress
 
 macro(cvc5_add_regression_test level file)
   add_test(${file}
+    ${PYTHON_EXECUTABLE}
     ${run_regress_script}
     ${RUN_REGRESSION_ARGS}
-    ${path_to_cvc5}/cvc5 ${CMAKE_CURRENT_LIST_DIR}/${file})
+    ${path_to_cvc5}/cvc5${CMAKE_EXECUTABLE_SUFFIX} ${CMAKE_CURRENT_LIST_DIR}/${file})
```
nafur, line 2865: Can we instead get the output filename of the `cvc5-bin` target? I guess we can get it via `get_target_property(... cvc5-bin OUTPUT_NAME)`

cmake/FindANTLR3.cmake
```diff
@@ -103,6 +103,13 @@ if(NOT ANTLR3_FOUND_SYSTEM)
         unset(64bit)
     endif()
 
+    # On Windows, we need to have a shell interpreter to call 'configure'
+    if(CMAKE_SYSTEM_NAME STREQUAL "Windows")
+        find_program (SHELL "sh")
```
nafur, line 108: Should we make this `REQUIRED`?

nafur:
@4tXJ7f Is there a reasonably elegant way to make the regressions work for windows line endings as well?

cmake/FindANTLR3.cmake
andrewvaughanj: Done!

test/regress/CMakeLists.txt
andrewvaughanj: It seems that, experimentally, `OUTPUT_NAME` gives `cvc5` and not `cvc5.exe` 🤷 

I tided-up the logic a bit to make the intent more explicit.

andrewvaughanj:
With the "these shouldn't be merged" changes to deal with the line-endings, I'm at:

image

after merging with master.

nafur (approved):
LGTM. Thank you so much for making progress on this front.

nafur:
Could you squash your changes and sign them off?

4tXJ7f:
FYI, #8652 fixes the issue with the line endings to make the regressions work with Wine.

RyanGlScott:
Is there anything keeping this from being merged? I'd think it would be fantastic if we could build CVC5 natively on Windows, as having to cross-compile it is cumbersome.

PR #7515, opened 2021-10-27
Fix upwards closure for relations

ajreynol:
This PR fixes an issue where we did compute upwards closure (for join, product, etc.) on equivalence classes whose membership cache had already been initialized (perhaps recursively from the computation of upwards/downwards closures on other terms).

It also generalizes the fix from #7511. Instead of doing explicit splitting, we mark shared terms and let theory combination (when necessary) do splitting.  This is required to fix a more general version of the benchmark from the previous PR, where instead now a term `c+1` is used in a position that is relevant to a join.

It disables a regression that times out after these fixes, and does further cleanup.


src/theory/sets/theory_sets_rels.cpp
```diff
@@ -229,24 +246,7 @@ void TheorySetsRels::check(Theory::Effort level)
...
-              KIND_TERM_IT kind_term_it = terms_it->second.find(eqc_node.getKind());
-
-              if( kind_term_it == terms_it->second.end() ) {
-                terms.push_back(eqc_node);
-                d_terms_cache[eqc_rep][eqc_node.getKind()] = terms;
-              } else {
-                kind_term_it->second.push_back(eqc_node);
-              }
-            }
+            d_terms_cache[eqc_rep][eqc_node.getKind()].push_back(eqc_node);
```
mudathirmahgoub, line 249: 16 lines shorted to just a single line. Nice

src/theory/sets/theory_sets_rels.cpp
```diff
@@ -125,44 +125,61 @@ void TheorySetsRels::check(Theory::Effort level)
 
     TERM_IT t_it = d_terms_cache.begin();
     while( t_it != d_terms_cache.end() ) {
```
mudathirmahgoub, line 127: As we discussed internally, this loop can be merged with the previous loop in this [line](https://github.com/cvc5/cvc5/blob/master/src/theory/sets/theory_sets_rels.cpp#L76). However, this can be done in a future PR. 
Also the relational solver needs to be refactored to avoid unnecessary cashing for membership constraints which are already cashed in SolverState.

test/regress/CMakeLists.txt
```diff
@@ -2759,6 +2760,8 @@ set(regression_disabled_tests
   regress0/auflia/fuzz01.smtv1.smt2
   ###
   regress0/bv/test00.smtv1.smt2
+  # timeout after fixing upwards closure for relations
```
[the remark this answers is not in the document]
mudathirmahgoub: For this particular case, rewriting can help as follows: 
```smt2
(= (join (tclosure R) R) (tclosure R) 
(= (join R (tclosure R)) (tclosure R)
```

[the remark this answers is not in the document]
mudathirmahgoub: This would be added in a separate PR

src/theory/sets/theory_sets_rels.cpp
```diff
@@ -1178,15 +1176,25 @@ void TheorySetsRels::check(Theory::Effort level)
...
+    TypeNode atn = a.getType();
+    if (atn.isTuple())
+    {
+      size_t tlen = atn.getTupleLength();
+      for (size_t i = 0; i < tlen; i++)
+      {
+        if (!areEqual(RelsUtils::nthElementOfTuple(a, i),
+                      RelsUtils::nthElementOfTuple(b, i)))
+        {
+          return true;
```
mudathirmahgoub, line 1189:
```suggestion
          return false;
```
I think you wanted to return false here

src/theory/sets/theory_sets_rels.cpp
```diff
@@ -1178,15 +1176,25 @@ void TheorySetsRels::check(Theory::Effort level)
...
+        }
       }
-      return equal;
-    } else if(!a.getType().isBoolean()){
+      return false;
```
mudathirmahgoub, line 1192:
```suggestion
      return true;
```

PR #7521, opened 2021-10-28
Distinguish purification types for strings proof reconstruction

ajreynol:
This improves the success rate for strings proof reconstruction for context-dependent simplifications.

It also makes a minor modification to how EXTF inferences work.  As a special case, `A => B` becomes `A ^ ~B => false` when B is already equal to false.  This PR removes this behavior for the sake of consistency.  (This could be explored as a more general strategy for turning facts to conflicts from any source if necessary).

The strings solver applies context-dependent simplification in various ways. This ensures that the proof reconstruction is faithful for cases of *congruence* + rewriting instead of substitution+rewriting, which the inferences STRINGS_EXTF and STRINGS_EXTF_N rely on.

This fixes 11 cases where proof reconstruction fails for string theory lemmas on Amazon benchmarks.


src/theory/strings/infer_proof_cons.h
```diff
@@ -198,29 +219,48 @@ class InferProofCons : public ProofGenerator
...
-                                     bool concludeTgtNew = false);
+                                     std::unordered_set<Node>& termsToPurify);
   /**
    * Return the purified form of the predicate lit with respect to a set of
    * terms to purify, call the returned literal lit'.
    * If concludeNew is true, then we add a proof of lit' from lit in psb;
-   * otherwise we add a proof of lit from lit'.
-   * Note that string predicates that require purification are string
-   * (dis)equalities only.
+   * otherwise we add a proof of lit from lit'. The position which are purified
```
HanielB, line 239:
```suggestion
   * otherwise we add a proof of lit from lit'. The positions which are purified
```

src/theory/strings/infer_proof_cons.h
```diff
@@ -188,6 +197,18 @@ class InferProofCons : public ProofGenerator
...
+   * The argument pt determines which arguments are relevant for purification.
+   * For core calculus (CORE_EQ) rules, examples of relevant positions are
+   * non-concatenation terms in positions like:
+   *   (= * (str.++ * (str.++ * *) * *))
+   *   (not (= (str.++ * *) *))
+   * For extended function simplification, examples of relevant positions are:
+   *   (= (str.replace * * *) "")
+   *   (str.contains * *)
+   * If we are purifying a substitution with equality, the LHS is a relevant
+   * position to purify, and the RHS is treated like CORE_EQ.
```
HanielB, line 209: maybe you could better contextualize why the different policies are necessary? An example would be helpful.

HanielB (approved):
Thanks for the example!

PR #7531, opened 2021-10-29
Replace more static options accesses

nafur:
This PR replaces a bunch of static accesses to options (`options::foo()`) by using the `EnvObj::options()` method.


4tXJ7f (approved):
LGTM

PR #7532, opened 2021-10-29
Alethe: Translate CNF rules

Lachnitt:
Implementation of the translation of various CNF rules into the Alethe calculus.


src/proof/alethe/alethe_post_processor.cpp
```diff
@@ -1077,6 +1077,85 @@ bool AletheProofPostprocessCallback::update(Node res,
...
+    }
+    // The following rules are all translated according to the clause pattern.
+    case PfRule::CNF_ITE_NEG1:
+    {
+      return addAletheStepFromOr(AletheRule::ITE_NEG2, res, children, {}, *cdp);
+    }
+    case PfRule::CNF_ITE_NEG2:
+    {
+      return addAletheStepFromOr(AletheRule::ITE_NEG1, res, children, {}, *cdp);
+    }
```
HanielB, line 1123: Move these to above `CNF_ITE_POS3` so that they fall in the same "block" of cnf rules following the clausal pattern.

HanielB (changes requested):

src/proof/alethe/alethe_post_processor.cpp
Lachnitt: I tried to preserve the order in proof_rule.h but I am fine with this.

HanielB (approved):

PR #7535, opened 2021-10-30
Refactor DidYouMean

nafur:
This PR refactors the `DidYouMean` class, which is used to make helpful suggestions for misspelled tags and options. It uses `std::vector` instead of `std::set`, moves it from `options` to `util` (so that we can keep using it in both libcvc5 and the driver) and generally modernizes the code a bit. Also, it replaces the stand-alone and never executed test by a regular unit test.


4tXJ7f (approved):
LGTM

PR #7536, opened 2021-10-30
api: Fix documentation for kind IAND.

aniemetz:


ajreynol:
I'm not sure what the ideal format should be for documenting indexed operators.

It seems that our documentation of all of the indexed operators follow a pattern where we say how to construct the operator, then how to construct an application of that operator, see e.g. https://github.com/cvc5/cvc5/blob/master/src/api/cpp/cvc5_kind.h#L1236

We should keep this consistent.  This would imply that if we remove the documentation for how to construct IAND applications, then we should remove the documentation for all the other indexed operators as well.

Alternatively, we could keep the documentation for how to construct the application.

ajreynol (approved):
LGTM

PR #7539, opened 2021-10-31
Fix soundess issue for bags with negative multiplicity

mudathirmahgoub:
This PR fixes soundness issues found by cvc5 fuzzy sygus when it answers unsat for sat problems. 
One error was with the rewrite rule  `(bag.count x (bag x c) = c` which did not account for cases when c is negative. 
This conflicts with fact that all multiplicities are non-negative. 
Another error was that the  `difference_remove` inference rule didn't consider the negative case for (count e B)
```smt2
(= (count e (difference_remove A B)) (ite (=  (count e B) 0) (count e A) 0))) ; not enough
(= (count e (difference_remove A B)) (ite (<= (count e B) 0) (count e A) 0))) ; the right rule
```


src/theory/bags/inference_generator.h
```diff
@@ -50,15 +50,14 @@ class InferenceGenerator
...
-   * (=>
-   *   true
-   *   (= (bag.count e skolem) (ite (= e x) c 0)))
+   * @return an inference that represents the following cases:
+   * 1- e, x are in the same equivalent class, then we infer:
+   *    (= (bag.count e skolem) (ite (>= c 1) c 0)))
+   * 2- e, x are known to be disequal, then we infer:
+   *    (= (bag.count e skolem) 0))
+   * 3- if neither holds, we infer:
+   *    (= (bag.count e skolem) (ite (and (= e x) (>= c 1)) c 0)))
```
mudathirmahgoub, line 59: This reminds me of the splitting lemma for the join operator.

ajreynol: I think this version is fine.

The challenge is that the theory of bags will not be informed of the direction that is split, since c >= 1 belongs to arithmetic.

ajreynol (approved):

PR #7541, opened 2021-11-01
Add join tclosure rewrite rule for sets

mudathirmahgoub:
This PR adds the following two rewrite rules for sets:
```smt2
(join (tclosure r) r) = (tclosure r)
(join s (tclosure s)) = (tclosure s)
```


ajreynol (approved):
LGTM, thanks for adding this.

mudathirmahgoub:
There is a counter example for this rule when `r` is not empty, but the join is empty 
```
r                     = {(0,1)}
tclosure r            = {(0,1)}
(join (tclosure r) r) = empty 
```

PR #7546, opened 2021-11-01
Fix exclusion criteria for codatatype model values

### LINKED ISSUE #4851: (--check-unsat-cores) Fatal failure at smt/smt_engine.cpp:2602 (codatatypes)
```
[549] % cvc4 -q small.smt2
sat
[550] % cvc4 --check-models -q small.smt2
sat
[551] % cvc4 --check-models --check-unsat-cores -q small.smt2
Fatal failure within void CVC4::SmtEngine::checkModel(bool) at smt/smt_engine.cpp:2602
assertion:   (distinct f (b 0 f))
simplifies to: false
expected `true'.
Aborted
[552] % 
[552] % cat small.smt2
(declare-codatatypes ((a 0)) (((b (c Int) (d a)))))
(declare-fun e () a)
(declare-fun f () a)
(assert (distinct f (b 0 f)))
(assert (= e f))
(check-sat)
[553] %
```
OS: Ubuntu 18.04
Commit: c324bb21

ajreynol:
This fixes codatatype model value construction.

Model construction for codatatypes is non-standard since it requires analyzing whether (possibly recursively defined) terms are alpha equivalent to others during model construction.  This is currently handled as a special case within the theory model builder.  (This should eventually be moved somewhere more appropriate via a new callback to theory).

This fixes the criteria which was too permissive about which values can be used in models.  We now exclude values that match the skeleton of representative codatatype terms. Previously we additionally required that the terms were bisimilar.

Fixes #4851.


HanielB (approved):
LGTM

PR #7550, opened 2021-11-01
Eliminate calls to Rewriter::rewrite and options:: in sets and bags

ajreynol:
There is 1 remaining call to Rewriter::rewrite in the bags type enumerator which is not straightforward to eliminate; it should perhaps call an intermediate call to a normal form utility instead of the full rewriter.


mudathirmahgoub (approved):
LGTM

Issue #4813, opened 2020-07-30
(default) Fatal failure at quant_elim_solver.cpp:83(quantifier elimination in the presence of asserted quantifiers)


rainoftime:
~~~~
(set-logic LIA)
(declare-const v8 Bool)
(assert (not (exists ((q16 Bool)) (xor true q16 v8 q16))))
(declare-const v24 Bool)
(get-qe (exists ((q17 Bool) (q18 Int) (q19 Int) (q20 Int) (q21 Int) (q22 Bool)) v24))
~~~~

cvc4 f5eb204df
~~~~
Fatal failure within CVC4::Expr CVC4::SmtEngine::doQuantifierElimination(const CVC4::Expr&, bool, bool) at /home/peisen/test/tofuzz/CVC4/src/smt/smt_engine.cpp:2891
Check failure

 inst_qs.size() <= 1

Aborted
~~~~

nafur:
A somewhat smaller example:
```
(declare-const v24 Bool)
(assert (forall ((q16 Bool)) (not (xor q16 v24 q16))))
(get-qe (exists ((q18 Int)) v24))
```
For some reason, `d_theoryEngine->getInstantiatedQuantifiedFormulas( inst_qs );` collects both quantifiers (from the `assert` and from the `get-qe`) where the code only expects one of them. Is it possible that `get-qe` does not expect to have quantified formulae in the preceding asserts?

ajreynol:
`get-qe` doesn't expect there to be other quantified formulas besides the one in the command itself.  The problem is we are missing an assertion that checks this.  I will look into this.

PR #7551, opened 2021-11-01
Make quant elimination robust to presence of other quantified formulas

ajreynol:
Fixes https://github.com/cvc5/cvc5/issues/4813


nafur (approved):
LGTM.

PR #7553, opened 2021-11-02
Improve defaults for sygus default grammars

ajreynol:
We now add constants from the conjecture to default grammars by default.  We also ensure that integer constants are used as real constants when applicable.


HanielB:
Didn't we have performance issues with doing this by default and that was why it was not done? Or maybe I'm misremembering.

ajreynol:
I also have that recollection.

Regardless, I think this is the right default value even if it is worse on the SyGuS library, as otherwise we have no chance to solve problems that require large constants by default.  I've had several users who I've suggested to enable this option for.

HanielB:
I see, sounds good to me then.

HanielB (approved):

PR #7554, opened 2021-11-02
Formalize more string skolems

ajreynol:
This properly formalizes all string skolems used in reduction and preprocessing. This avoids proof checking failures due to non-deterministism when checking steps for these modules.

Fixes https://github.com/cvc5/cvc5-projects/issues/334. Fixes https://github.com/cvc5/cvc5-projects/issues/331.


test/api/proj-issue334.cpp
```diff
@@ -0,0 +1,20 @@
+#include "api/cpp/cvc5.h"
```
4tXJ7f, line 1: Add boilerplate.

src/theory/strings/skolem_cache.h
```diff
@@ -58,6 +58,10 @@ class SkolemCache
    * preconditions below, e.g. where we are considering a' ++ a = b' ++ b.
    *
    * All skolems assume a and b are strings unless otherwise stated.
+   *
+   * Notice that these identifiers are each syntax sugar for constructing a
+   * purification skolem. It is required for the purposes of proof checking
+   * that this only result in calls to SkolemManager::mkPurifySkolem.
```
4tXJ7f, line 64:
```suggestion
   * that this only results in calls to SkolemManager::mkPurifySkolem.
```

4tXJ7f (changes requested):
Just some minor requests.

test/api/proj-issue334.cpp
ajreynol: What do you mean?

4tXJ7f: Our usual copyright header.

4tXJ7f (approved):
LGTM, thanks for the update.

PR #7555, opened 2021-11-02
Refactor IO stream manipulators

nafur:
This PR consolidates `SetLanguage`, `ExprSetDepth` and `ExprDag` into a single consistent utility. Also, it makes it play more nicely with users setting these options and removes some obsolete code.


src/options/io_utils.cpp
```diff
@@ -0,0 +1,105 @@
...
+#include "options/io_utils.h"
+
+#include <iomanip>
+#include <iostream>
+
+namespace cvc5::options::ioutils {
+namespace {
+
+template <typename T>
+void setData(std::ios_base& ios, int iosIndex, T value)
```
mpreiner, line 25: Use fixed-size types.

nafur: The `iword` methods actually uses `int`. Given that the `iosIndex` is completely internal to this, I'd rather keep it as `int`. However, the actual values for `dagThresh` and `nodeDepth` should be changed...

mpreiner (approved):

PR #7556, opened 2021-11-02
Improve syntax for fmf cardinality constraints

ajreynol:
This is an experimental extension of smt2.


abdoo8080 (approved):
LGTM! I will open a PR to fix the printing issue later.

PR #7558, opened 2021-11-02
Fix preregistration for floating point theory

ajreynol:
Preregistering terms must always add them to the equality engine of TheoryFp, otherwise there may be preregistered terms that do not occur in the equality engine of TheoryFp, thus leading to crashes during theory combination.

Fixes https://github.com/cvc5/cvc5-projects/issues/329.


4tXJ7f (approved):
LGTM

PR #7561, opened 2021-11-03
Refactor skolem construction

ajreynol:
This makes all methods for constructing skolems local to SkolemManager.

It removes infrastructure for node manager listeners being notified when skolems are constructed.  This was used for 2 things:
(1) The old dumping infrastructure, where skolems could contribute in part to traces to print benchmarks. This code will be deleted soon.
(2) The miplib preprocessing pass, which used this functionality to listen to what skolems were constructed, and subsequently assumed these skolems coincided with what Boolean variables appeared in assertions.  This is highly error prone, and has been replaced by a simple traversal to collect Boolean variables in assertions.

This is in preparation for a lazy lambda lifting technique, which will require more infrastructure in SkolemManager.


4tXJ7f (approved):
LGTM

PR #7562, opened 2021-11-03
Add support for special tag collectors

nafur:
This PR makes the collection of trace and debug tags more flexible by allowing for "special tag collectors". They can be used to inject tags that are not used as constant strings, but only assembled dynamically at runtime.
One example is the set of `assertions::pre-X` and `assertions::post-X` tags. They are currently handled by `Dump`, and will be migrated to `Trace`.


src/base/collect_tags.py
```diff
@@ -28,16 +28,32 @@ def parse_args():
...
+        'collector' : lambda basedir:
+            re.findall('registerPassInfo\("([a-z-]+)"',
+                open(os.path.join(basedir, 'preprocessing', 'preprocessing_pass_registry.cpp')).read()
+            )
+    }
+]
 
 def collect_tags(basedir):
-    """Collect all tags used in filed within the given base directory.
+    """Collect all tags used in filed within the given base directory and then
```
4tXJ7f, line 44:
```suggestion
    """Collect all tags used in files within the given base directory and then
```

4tXJ7f (approved):
LGTM, just a minor typo.

PR #7566, opened 2021-11-03
Minor cleanup in SolverEngine::setInfo()

nafur:
This PR does a bit of cleanup by properly calling the option handler instead of trying to manually replicating it. This is way more robust in the face of modifications to the option handlers.


4tXJ7f (approved):
LGTM

PR #7568, opened 2021-11-04
Fix links in README.md

### LINKED ISSUE #7016: Link to prebuilt binaries in README 404s
**Describe the bug**
The README contains a link to http://cvc5.github.io/downloads, which 404s. It seems like this was a result of #6497 changing links to the cvc5 site before that site was actually ready.

**Command line arguments**: N/A
**cvc5 version/commit**: N/A
**Operating system**: N/A

**`configure.sh` options**  
N/A

**`configure.sh` output**  
N/A

nafur:
This PR fixes links to releases and nightly builds. Fixes #7016.


4tXJ7f (commented):
I think that it would be better to fix the website instead of reverting our links back to https://cvc4.github.io/downloads.

nafur:
> I think that it would be better to fix the website instead of reverting our links back to https://cvc4.github.io/downloads (that was my original intention for resolving this issue).

I generally agree. I figured we would probably want to directly put the new buildbot-generated nightlies on the cvc5 website...

mpreiner (approved):

PR #7570, opened 2021-11-04
Move functions and lambdas from builtin to uf

ajreynol:
This is in preparation for adding better native support for handling lambdas in the higher-order extension of the UF theory.

We require that LAMBDA and function types belong to theory UF so that the theory solver is properly notified.

This also splits the utility methods for computing whether a function is "constant" to its own file.

This PR is code move only.


src/theory/uf/function_const.cpp
```diff
@@ -0,0 +1,412 @@
...
+namespace cvc5 {
+namespace theory {
+namespace uf {
+
+TypeNode FunctionConst::getFunctionTypeForArrayType(TypeNode atn, Node bvl)
+{
+  std::vector<TypeNode> children;
+  for (unsigned i = 0; i < bvl.getNumChildren(); i++)
+  {
+    Assert(atn.isArray());
```
mudathirmahgoub, line 30: This assertion dose not depend on index `i`

src/theory/uf/theory_uf_rewriter.cpp
```diff
@@ -204,6 +210,55 @@ Node TheoryUfRewriter::decomposeHoApply(TNode n,
 }
 bool TheoryUfRewriter::canUseAsApplyUfOperator(TNode n) { return n.isVar(); }
 
+Node TheoryUfRewriter::rewriteLambda(Node node)
+{
+  // The following code ensures that if node is equivalent to a constant
```
mudathirmahgoub, line 215: To prevent calling this function with the wrong kind,
the following assertion can be added here `Assert(node.getKind() == kind::LAMBDA);`.

src/theory/uf/theory_uf_type_rules.cpp
```diff
@@ -160,6 +162,112 @@ TypeNode HoApplyTypeRule::computeType(NodeManager* nodeManager,
...
+    Trace("lambda-const") << "Non-constant lambda : " << n
+                          << " since it has no array representation."
+                          << std::endl;
+  }
+  return false;
+}
+
+Cardinality FunctionProperties::computeCardinality(TypeNode type)
+{
+  // Don't assert this; allow other theories to use this cardinality
```
mudathirmahgoub, line 234: Is this comment needed?

ajreynol: Yeah, it's saying the possible assertion doesn't hold, so I'd keep it.

mudathirmahgoub (approved):
LGTM. Thanks for the refactoring and new documentation.

PR #7571, opened 2021-11-04
Enable CDCAC solver for selected quantified logics

nafur:
This PR enables the CDCAC solver for quantified logics with reals, but no integers. Also, it disables SyGuS if no integers are used. The regression is a benchmark that is only solved with this new configuration.


nafur:
This change only affects the `NRA` logic on SMT-LIB:
```
NRA         11031531-cvc5-master      ok   4058    3847    64   3783   3745      117       0      0     4       0   36794.4   21187.9
            11031531-cvc5-nosygus     ok   4058    4021   213   3808   4021       28       0      0   178       0    8541.4    7949.6
```

ajreynol (approved):

PR #7572, opened 2021-11-04
Replace the old dump infrastructure

ajreynol:
This deletes the old dumping infrastructure, which due to changes in our usage of SolverEngine does not print valid benchmarks.

This eliminates the concept of a "NodeManagerListener" which has been problematic to maintain.

This PR reimplements the `--dump=assertions:pre-X` and `--dump=assertions:post-X` as `-t assertions::pre-X` and `-t assertions::post-X`.  This is done in a self contained utility method in ProcessAssertions using smt::PrintBenchmark and does not require keeping the rest of the code in sync.

The other dumping tags are deleted for now, and will be reimplemented as needed.


4tXJ7f (approved):
LGTM

PR #7575, opened 2021-11-04
Eliminate `Warning` macro in favor of `EnvObj::warning`

nafur:
This PR eliminates almost all usages of the `Warning` macro, replacing it mostly by calling `EnvObj::warning`.

### DIFF
--- a/src/preprocessing/passes/miplib_trick.cpp
+++ b/src/preprocessing/passes/miplib_trick.cpp
@@ -586,10 +586,6 @@ PreprocessingPassResult MipLibTrick::applyInternal(
           Node newAssertion = var.eqNode(rewrite(sum));
           if (top_level_substs.hasSubstitution(newAssertion[0]))
           {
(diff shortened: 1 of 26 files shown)


ajreynol:
I had been holding off using multiple inheritance combined with EnvObj, which this PR now is doing.

After some thought, I believe I'm okay with doing this, if no one else objects.  I believe @4tXJ7f is also in favor of this style.

4tXJ7f (approved):
LGTM, but @ajreynol should probably take a look too, since we've had some discussions related to `EnvObj` and multiple inheritance and I seem to remember that he was against that.

4tXJ7f:
> I had been holding off using multiple inheritance combined with EnvObj, which this PR now is doing.
> 
> After some thought, I believe I'm okay with doing this, if no one else objects. I believe @4tXJ7f is also in favor of this style.

Saw your comment after posting my review :) Yeah, I am in favor of this style. I'd like to keep multiple inheritance down to a minimum but I am ok with it if the second class is an `EnvObj`.

PR #7578, opened 2021-11-05
Add unit test and assertion to test and catch cvc5-projects/issues/#337.

aniemetz:
Note: This is currently failing and needs to be merged after cvc5/cvc5-projects/issues/337 is fixed.


src/api/cpp/cvc5.cpp
```diff
@@ -6564,7 +6564,9 @@ Term Solver::simplify(const Term& term)
   CVC5_API_TRY_CATCH_BEGIN;
   CVC5_API_SOLVER_CHECK_TERM(term);
   //////// all checks before this line
-  return Term(this, d_slv->simplify(*term.d_node));
+  Term res = Term(this, d_slv->simplify(*term.d_node));
+  Assert(res.getSort() == term.getSort());
```
ajreynol, line 6568: Actually, until subtyping is eliminated, this needs to be relaxed since we allow Real to be simplified to Int.

ajreynol, line 6568:
```suggestion
  Assert(res.getSort().d_type->isSubtypeOf(*term.getSort().d_type));
```

ajreynol (changes requested):

ajreynol:
This should be fixed when https://github.com/cvc5/cvc5/pull/8209 is merged.

ajreynol (approved):

PR #7580, opened 2021-11-05
Remove many static calls to rewrite

ajreynol:
This is the result of a global replace `Rewriter::rewrite(` -> `rewrite(`  + patching the results.

It makes several classes into EnvObj.  Many calls to Rewriter::rewrite remain (that are embedded in classes that should not be EnvObj).


src/theory/fp/fp_word_blaster.h
```diff
@@ -106,7 +106,7 @@ class nodeWrapper : public Node
  * allowing them to be traced via GDB.
  */
 #ifdef CVC5_SYM_SYMBOLIC_EVAL
-  nodeWrapper(const Node& n) : Node(theory::Rewriter::rewrite(n)) {}
+  nodeWrapper(const Node& n) : Node(theory::rewrite(n)) {}
```
nafur, line 109: I think this should be different...

nafur (commented):
Just one thing...

src/theory/fp/fp_word_blaster.h
ajreynol: Hm, good catch, apparently this code is disabled?

nafur (approved):
LGTM

PR #7582, opened 2021-11-05
Remove more static option accesses

nafur:
This PR removes more than half of the remaining static option accesses `options::foo()` from the theory solvers.


ajreynol:
I assume this depends on https://github.com/cvc5/cvc5/pull/7581.

ajreynol (approved):

PR #7583, opened 2021-11-05
Fix some issues with the java api

nafur:
This PR fixes two issues with the java api:
- the `JNI_HEADERS` variable was set to a non-existent file, which caused the `generate-jni-headers` target to always rebuilt.
- the directory structure was unnecessarily nested (probably because we use `CMAKE_CURRENT_BINARY_DIR` instead of `CMAKE_BINARY_DIR`).


src/api/java/CMakeLists.txt
```diff
@@ -76,7 +76,7 @@ set(JAVA_FILES
 
 # specify generated jni headers
 set(JNI_HEADERS
-  ${JNI_DIR}/cvc5_Solver.h
+  ${JNI_DIR}/io_github_cvc5_api_Solver.h
```
mudathirmahgoub, line 79: I forgot why this was added. Is this variable really needed?

nafur: This variable is used as the `OUTPUT` of the custom command behind the `generate-jni-headers` target and the `DEPENDS` of said target.

mudathirmahgoub (approved):
LGTM

PR #7584, opened 2021-11-05
Minor changes to circuit propagator

HanielB:
Previously in a given part of the code a proof would be retrieved only so that it'd be printed. This commit guards that part of the code with whether the trace is on and gives more information in what is printed.

Also changes the style of a call.


nafur (approved):
LGTM.

PR #7586, opened 2021-11-05
Remove `Chat()` in favor of new `verbose()`

nafur:
This PR completely removes the first of several logging macros. Instead of `Chat()`, we now use `verbose(2)`.


ajreynol (commented):
I'm wondering if it would make sense to simply change all of this to Trace/Debug instead.  That way we can design what we want `verbose` to print from scratch.

nafur:
My idea was to first consolidate how we can output stuff without changing current behaviour (if possible), and then in a second step reorganize what is printed when. I think this redesign of what is printed when needs a global look an all these outputs, and should better not be mixed with refactoring how the printing is done.

ajreynol (approved):

PR #7587, opened 2021-11-05
[proofs] Adding NoSubtype node converter to Alethe

HanielB:
Initial version. It'll be improved on demand according to new cases that may lead to issues.


src/proof/alethe/alethe_nosubtype_node_converter.h
```diff
@@ -0,0 +1,43 @@
...
+/**
+ * This is a helper class for the Alethe post-processor that converts nodes into
+ * form that Alethe expects.
+ */
+class AletheNoSubtypeNodeConverter : public NodeConverter
+{
+ public:
+  AletheNoSubtypeNodeConverter() {}
+  ~AletheNoSubtypeNodeConverter() {}
+  /** Convert by removing attributes of quantifiers. */
```
Lachnitt, line 36: This comment might be a bit unclear, you are not removing them just transforming them.

HanielB: good catch, it's a wrong copy-past. I'll fix

ajreynol (approved):

PR #7593, opened 2021-11-06
Integrate java unit tests into ctest

nafur:
This PR properly integrates the java api unit tests into ctest.


mudathirmahgoub:
I tried your branch locally and made a runtime error, but it was not captured by ctest. 
Can you make a commit with runtime error to see if CI would capture it?

nafur:
> I tried your branch locally and made a runtime error, but it was not captured by ctest. Can you make a commit with runtime error to see if CI would capture it? `ctest -R unit/api/java/SolverTest`

I'm not entirely sure what you mean. I removed the `d_solver = new Solver();` line from the `SolverTest` and get the following:
```
$ ctest -R SolverTest
Test project /home/gereon/cvc5-master/build_debug
    Start 3023: unit/api/java/SolverTest
1/1 Test #3023: unit/api/java/SolverTest .........***Failed    0.93 sec

0% tests passed, 1 tests failed out of 1

Label Time Summary:
unit java    =   0.93 sec*proc (1 test)

Total Test time (real) =   1.09 sec

The following tests FAILED:
	3023 - unit/api/java/SolverTest (Failed)
Errors while running CTest
```

mudathirmahgoub:
From the trace you posted, it only says there is a failure in SolverTest, and it seems I have to use another run to see which unit test is failing, and the stack trace of the error. 
From the CI, I want to see exactly which unit is failing and the stack trace of the error message. 
Would this be preserved by using ctest here?
Also what's wrong with the current setup that just runs junit directly?

nafur:
This is true; you get the information you want with `--output-on-failure`. This is the same for all other unit tests as well, and the argument is passed in the CI runs.
In the current setup, the junit tests are run with the `ALL` target (unlike any other test), and are completely separate from all other tests. We want all tests bundled in one place, see also the comment in `test/CMakeLists.txt` above the `check` target.

mudathirmahgoub (approved):
LGTM. Thanks for the clarification and changes.

PR #7601, opened 2021-11-08
Add lambda lift utility

ajreynol:
Towards a lazy approach for handling lambdas in the higher-order extension.


src/theory/uf/lambda_lift.h
```diff
@@ -0,0 +1,78 @@
...
+class LambdaLift : protected EnvObj
+{
+  typedef context::CDHashSet<Node> NodeSet;
+  typedef context::CDHashMap<Node, Node> NodeNodeMap;
+
+ public:
+  LambdaLift(Env& env);
+
+  /** process, return the trust node corresponding to the lemma */
+  TrustNode lift(Node node);
```
mudathirmahgoub, line 45: Can you add to the documentation that if the node (or lemma) is already lifted, this function would return null?

src/theory/uf/lambda_lift.cpp
```diff
@@ -0,0 +1,215 @@
...
+      // Make the skolem to represent the lambda
+      NodeManager* nm = NodeManager::currentNM();
+      SkolemManager* sm = nm->getSkolemManager();
+      skolem = sm->mkPurifySkolem(
+          node,
+          "lambdaF",
+          "a function introduced due to term-level lambda removal");
+    }
+  }
+  else if (k == WITNESS)
```
mudathirmahgoub, line 161: How come `witness` is related to this class `LambdaLift`?

ajreynol: Didn't mean to include this, i removed now.

src/theory/uf/lambda_lift.cpp
```diff
@@ -0,0 +1,187 @@
...
+    {
+      std::vector<Node> args(node.begin(), node.end());
+      Node app = betaReduce(opl, args);
+      Trace("uf-lazy-ll") << "Beta reduce: " << node << " -> " << app
+                          << std::endl;
+      return TrustNode::mkTrustRewrite(node, app, nullptr);
+    }
+  }
+  // otherwise, unchanged
+  return TrustNode::null();
```
mudathirmahgoub, line 170: Did you mean to return `node` here or `null`? I guess it depends on how you call this function. Better to document the expected behavior in the header file.

mudathirmahgoub (approved):
LGTM.

src/theory/uf/lambda_lift.h
```diff
@@ -0,0 +1,88 @@
...
+{
+  typedef context::CDHashSet<Node> NodeSet;
+  typedef context::CDHashMap<Node, Node> NodeNodeMap;
+
+ public:
+  LambdaLift(Env& env);
+
+  /**
+   * process, return the trust node corresponding to the lemma for the lambda
+   * lifting of (lambda) term node, or null if it is not a lambda, or if
```
HanielB, line 49:
```suggestion
   * lifting of (lambda) term node, or null if it is not a lambda or if
```

src/theory/uf/lambda_lift.h
```diff
@@ -0,0 +1,88 @@
...
+  Node getLambdaFor(TNode skolem) const;
+
+  /**
+   * Beta-reduce node. If node is APPLY_UF and its operator is a lambda
+   * function known by this class, then this method returns the beta
+   * reduced version of node.
+   */
+  TrustNode betaReduce(TNode node) const;
+  /** Beta-reduce the given lambda on the given arguments. */
+  Node betaReduce(TNode lam, const std::vector<Node>& args) const;
```
HanielB, line 69: clarify the behavior when the input does not respect the pre-conditions

src/theory/uf/lambda_lift.cpp
```diff
@@ -0,0 +1,187 @@
...
+}
+
+Node LambdaLift::getSkolemFor(TNode node)
+{
+  Node skolem;
+  Kind k = node.getKind();
+  if (k == LAMBDA)
+  {
+    // if a lambda, do lambda-lifting
+    if (!expr::hasFreeVar(node))
```
HanielB, line 137: clarify

HanielB (changes requested):

src/theory/uf/lambda_lift.cpp
```diff
@@ -0,0 +1,215 @@
...
+  }
+  Kind k = node.getKind();
+  Node assertion;
+  if (k == LAMBDA)
+  {
+    NodeManager* nm = NodeManager::currentNM();
+    // The new assertion
+    std::vector<Node> children;
+    // bound variable list
+    children.push_back(node[0]);
```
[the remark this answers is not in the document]
ajreynol: its directly under a condition for this on line 113.

HanielB (approved):

PR #7603, opened 2021-11-08
Only eliminate lambdas in higher-order elimination if ho-elim is enabled

ajreynol:
Required for lazy lambdas in HO.

Also properly guards the case when the preprocessing pass is a no-op.


test/regress/CMakeLists.txt
```diff
@@ -2800,6 +2799,8 @@ set(regression_disabled_tests
   ###
   regress1/bug472.smt2
   regress1/datatypes/non-simple-rec-set.smt2
+  # disabled temporarily until lazy lambda handling is fuzz_1_to_52_merged
```
HanielB, line 2802: Something funny with the comment

test/regress/regress1/ho/bug_freevar_PHI004^4-delta.smt2
```diff
@@ -1,4 +1,4 @@
-; COMMAND-LINE: --finite-model-find -q --decision=justification-old
+; COMMAND-LINE: --finite-model-find --decision=justification-old -q --ho-elim
```
HanielB, line 1: Why add --ho-elim? If there is something here to be tested then you could have the two cases:
```
; COMMAND-LINE: --finite-model-find --decision=justification-old -q
; COMMAND-LINE: --finite-model-find --decision=justification-old -q --ho-elim
```

ajreynol: I was trying to patch a fix to this regression (it is disabled by this PR), I'll revert this change.

HanielB (approved):

PR #7604, opened 2021-11-08
Preparation for non-constant lambda models

ajreynol:
This makes our core model construction amenable to (non-constant) lambdas as assignments to function equivalence classes for higher-order.

We currently only generate almost constant functions for values of functions. After this PR, we allow explicitly provided lambdas as representatives provided by a theory, which will be used by the higher-order extension.

It also makes some improvements to debug check models regarding checking when representatives are equal to their model values.


src/theory/theory_model.cpp
```diff
@@ -715,7 +717,8 @@ void TheoryModel::assignFunctionDefinition( Node f, Node f_def ) {
     while( !eqc_i.isFinished() ) {
       Node n = *eqc_i;
       // if an unassigned variable function
-      if( n.isVar() && d_uf_terms.find( n )!=d_uf_terms.end() && !hasAssignedFunctionDefinition( n ) ){
+      if (n.isVar() && !hasAssignedFunctionDefinition(n))
```
mudathirmahgoub, line 720: Why did we need to check `n` is a key in the map `d_uf_terms` before? why is it not needed now?

src/theory/theory_model_builder.cpp
```diff
@@ -1121,18 +1128,24 @@ void TheoryEngineModelBuilder::debugCheckModel(TheoryModel* tm)
...
-        if (tm->getValue(*eqc_i) != rep)
+        std::stringstream err;
+        err << "Failed representative check:" << std::endl
+            << "( " << repCheckInstance << ") "
+            << "n: " << n << endl
+            << "getValue(n): " << tm->getValue(n) << std::endl
+            << "rep: " << rep << std::endl;
+        if (val.isConst() && rep.isConst())
+        {
+          AlwaysAssert(val == rep) << err.str();
```
mudathirmahgoub, line 1142: You can use `AlwaysAssert(false)` based on the if condition above.

mudathirmahgoub (approved):
LGTM

src/theory/theory_model_builder.cpp
ajreynol: Yeah, I saw, I think this was intentional since the error message is slightly more informative (the user sees `val==rep`, which indicates some information about what the error was).

PR #7533, opened 2021-10-29
Alethe: Translate REORDERING rule

Lachnitt:
Implementation of the translation of REORDERING rules into the Alethe calculus.

Should be merged after PR #7532.

Co-authored-by: Haniel Barbosa hanielbbarbosa@gmail.com

PR #7605, opened 2021-11-08
Alethe: Translate Equality rules

Lachnitt:
Implementation of the translation of SYMM, TRANS and CONG rules into the Alethe calculus.

Should be merged after PR #7533.


src/proof/alethe/alethe_post_processor.cpp
```diff
@@ -1162,6 +1162,105 @@ bool AletheProofPostprocessCallback::update(Node res,
...
+                             children,
+                             {},
+                             *cdp);
+      }
+      return addAletheStep(AletheRule::SYMM,
+                           res,
+                           nm->mkNode(kind::SEXPR, d_cl, res),
+                           children,
+                           {},
+                           *cdp);
```
HanielB, line 1193:
```suggestion
      return addAletheStep(
          res.getKind() == kind::NOT ? AletheRule::NOT_SYMM : AletheRule::SYMM,
          res,
          nm->mkNode(kind::SEXPR, d_cl, res),
          children,
          {},
          *cdp);
```

src/proof/alethe/alethe_post_processor.cpp
```diff
@@ -1162,6 +1162,105 @@ bool AletheProofPostprocessCallback::update(Node res,
...
+      return addAletheStep(AletheRule::TRANS,
+                           res,
+                           nm->mkNode(kind::SEXPR, d_cl, res),
+                           children,
+                           {},
+                           *cdp);
+    }
+      // ======== Congruence
+      // In the case that the kind of the function symbol ?f is forall or
+      // exists, the cong rule needs to be converted into a bind rule. The first
```
HanielB, line 1206:
```suggestion
      // In the case that the kind of the function symbol f? is FORALL or
      // EXISTS, the cong rule needs to be converted into a bind rule. The first
```

src/proof/alethe/alethe_post_processor.cpp
```diff
@@ -1162,6 +1162,105 @@ bool AletheProofPostprocessCallback::update(Node res,
...
+      // n children will be refl rules, e.g. (= (v0 Int) (w0 Int)).
```
HanielB, line 1207: isn't it the case that `v0` must be equal to `w0` in the case of a CONG rule on FORALL/EXISTS?

src/proof/alethe/alethe_post_processor.cpp
```diff
@@ -1162,6 +1162,105 @@ bool AletheProofPostprocessCallback::update(Node res,
...
+      //    P1 ... Pn
+      //  -------------------------------------------------------- cong
+      //   (cl (= (<kind> f? t1 ... tn) (<kind> f? s1 ... sn)))**
+      //
+      // ** the corresponding proof node is (= (<kind> f? t1 ... tn) (<kind> f?
+      // s1 ... sn))
+    case PfRule::CONG:
+    {
+      if (args[0] == ProofRuleChecker::mkKindNode(kind::FORALL)
+          || args[0] == ProofRuleChecker::mkKindNode(kind::EXISTS))
```
HanielB, line 1235:
```suggestion
      if (res[0].isClosure())
```

src/proof/alethe/alethe_post_processor.cpp
```diff
@@ -1162,6 +1162,105 @@ bool AletheProofPostprocessCallback::update(Node res,
...
+      {
+        std::vector<Node> vpis;
+        bool success = true;
+        for (size_t i = 0, size = children[0][0].getNumChildren(); i < size;
+             i++)
+        {
+          Node vpi =
+              nm->mkNode(kind::EQUAL, children[0][0][i], children[0][1][i]);
```
HanielB, line 1243:
```suggestion
          Node vpi = children[0][0][i].eqNode(children[0][1][i]);
```

src/proof/alethe/alethe_post_processor.cpp
```diff
@@ -1162,6 +1162,105 @@ bool AletheProofPostprocessCallback::update(Node res,
...
+          new_args.push_back(vpi);
+          vpis.push_back(nm->mkNode(kind::SEXPR, d_cl, vpi));
+          success&& addAletheStep(AletheRule::REFL, vpi, vpi, {}, {}, *cdp);
```
HanielB, line 1246:
```suggestion
          success &= addAletheStep(AletheRule::REFL, vpi, vpi, {}, {}, *cdp);
```

HanielB (changes requested):

src/proof/alethe/alethe_post_processor.cpp
Lachnitt: Yes it is, otherwise it cannot be proven by a reflexivity rule. However, I changed this from v0 to w0 last minute because I thought that when this rule is in a context that maps w0 to v0 it might be easier to understand this way. Does this make any sense?

HanielB: If I understand correctly I think what you are referring to would only happen to `PfRule::ALPHA_EQUIV` steps, no? It should be the case that for FORALL/EXISTS nodes the first proof child should always be `REFL` (you could add an assertion).

Lachnitt: Makes sense, let me change this back.

src/proof/alethe/alethe_post_processor.cpp
```diff
@@ -1162,6 +1162,95 @@ bool AletheProofPostprocessCallback::update(Node res,
...
+      //  * the corresponding proof node is (or (= vi vi))
+      //
+      // Otherwise, the rule follows the singleton pattern, i.e.:
+      //
+      //    P1 ... Pn
+      //  -------------------------------------------------------- cong
+      //   (cl (= (<kind> f? t1 ... tn) (<kind> f? s1 ... sn)))**
+      //
+      // ** the corresponding proof node is (= (<kind> f? t1 ... tn) (<kind> f?
+      // s1 ... sn))
```
HanielB, line 1223: the formatting is off by two

Lachnitt: Thanks for catching that, I fixed it.

HanielB (approved):

PR #7608, opened 2021-11-09
Remove `antlr_tracing.h`

nafur:
This PR removes the long obsolete `antlr_tracing.h`. The file contains an ugly hack to make lexer tracing work for ANTLR 3.2. It has been obsolete since ANTLR 3.3 (April 2011). Note that the code is only enabled with `-DCVC5_TRACE_ANTLR`, a flag that is never used.


4tXJ7f (approved):
Yikes!

PR #7597, opened 2021-11-07
Translating API tests to Python — part 1

yoni206:
This PR translates some of the API tests from [here](https://github.com/cvc5/cvc5/tree/master/test/api) to Python. The other tests are translated in a [private branch](https://github.com/yoni206/cvc5/tree/python_dir_struct/test/python/api) and will be added in a separate PR.

PR #7612, opened 2021-11-09
Reorganize test/unit/api directory.

aniemetz:
This moves .cpp files to directory test/unit/api/cpp and python files in
test/python/unit/api to test/unit/api/python.

@yoni206 this will affect #7597.


test/unit/api/cpp/CMakeLists.txt
```diff
@@ -0,0 +1,24 @@
...
+#
+# Copyright (c) 2009-2021 by the authors listed in the file AUTHORS
+# in the top-level source directory and their institutional affiliations.
+# All rights reserved.  See the file COPYING in the top-level source
+# directory for licensing information.
+# #############################################################################
+#
+# The build system configuration.
+##
+cvc5_add_unit_test_black(datatype_api_black api)
```
nafur, line 15: Why not move the definition of these macros to this file as well?

test/unit/api/python/CMakeLists.txt
```diff
@@ -17,21 +17,22 @@
 check_python_module("pytest")
 
 # Add Python bindings API tests.
-macro(cvc5_add_python_api_test name filename)
+macro(cvc5_add_python_api_unit_test name filename)
 # We create test target 'python/unit/api/myapitest' and run it with
 # 'ctest -R "python/unit/api/myapitest"'.
-add_test (NAME python/unit/api/${name}
+add_test (NAME ${name}
```
nafur, line 23: Can you remove the prefix from the Java tests as well for consistency?

aniemetz: We decided offline to keep the prefix and use a prefix for all unit tests.

nafur (approved):
LGTM.

PR #7614, opened 2021-11-09
api: Add Solver::mkRegexpAll().

aniemetz:


mudathirmahgoub:
Please don't forget to add the corresponding methods to python and java APIs

ajreynol:
Note: we could consider adding the kind REGEXP_ALL.

aniemetz:
> Note: we could consider adding the kind REGEXP_ALL.

Yes, I think we should.

ajreynol (approved):

PR #7615, opened 2021-11-09
Fix soundness issue of missing premises for count bag lemmas

mudathirmahgoub:
This PR fixes soundness issues found by cvc5 fuzzy sygus when it answers unsat for sat problems.
The error is related to a previous fix which did not include the required antecedent  in the two lemmas:
```smt2
(=>  (= e x) 
        (= 
              (bag.count e (bag x c)) 
             (ite (>= c 1) c 0)))
``` 
 and 
```smt2
(=>  (distinct x e))  (= (bag.count e (bag x c)) 0))
```


src/theory/bags/inference_generator.h
```diff
@@ -52,9 +52,13 @@ class InferenceGenerator
...
    * @return an inference that represents the following cases:
    * 1- e, x are in the same equivalent class, then we infer:
-   *    (= (bag.count e skolem) (ite (>= c 1) c 0)))
+   *    (=>
+   *        (= e x)
+   *        (bag.count e skolem) (ite (>= c 1) c 0)))
    * 2- e, x are known to be disequal, then we infer:
-   *    (= (bag.count e skolem) 0))
+   *    (=>
+   *      (distinct x e))
```
ajreynol, line 60:
```suggestion
   *      (distinct x e)
```

src/theory/bags/inference_generator.h
```diff
@@ -52,9 +52,13 @@ class InferenceGenerator
    * @param e is a node of type E
    * @return an inference that represents the following cases:
    * 1- e, x are in the same equivalent class, then we infer:
-   *    (= (bag.count e skolem) (ite (>= c 1) c 0)))
+   *    (=>
```
ajreynol, line 55: I'm not sure why this method takes `e` in the first place.  I would actually expect these inferences not to have premises, but use `x` instead of `e` throughout.  Also, in this case, only the first inference is necessary.

ajreynol (changes requested):

src/theory/bags/inference_generator.h
ajreynol: Ah I see, ok, so then actually, I would just send a single inference, corresponding to the 3rd case.

`(ite (and (= e x) (>= c 1)) (= (bag.count e skolem) c) (= (bag.count e skolem) 0)`

mudathirmahgoub: Done

ajreynol (approved):

PR #7621, opened 2021-11-09
[proofs] Further handling of int/real subtyping for Alethe backend

HanielB:
Extends the conversion to prevent subtyping in applications of arithmetic operators when one of the arguments is a "true real" and some is an integer. While the conversion is restricted to constants (since the whole procedure for now is based on wrapping integers with `CAST_TO_REAL`), it supports terms that are not constants but that amount to applications of arithmetic operators to constants, such as `(* (+ 1 1) (- 1))`.


ajreynol:
The policy on subtypes have changed, I'm wondering if this code is necessary.

HanielB:
Closing this since we will no longer be eliminating int/real subtyping from Alethe proofs.

PR #7628, opened 2021-11-10
[Strings] Minor refactor of eager solver

4tXJ7f:
This moves code that is not strictly related to the eager solver out of
the eager solver and into `TheoryStrings`. This is cleaner and makes it
easier to enable/disable the eager solver.

### DIFF
--- a/src/theory/strings/theory_strings.cpp
+++ b/src/theory/strings/theory_strings.cpp
@@ -749,8 +749,60 @@ void TheoryStrings::eqNotifyNewClass(TNode t){
     Trace("strings-debug") << "New length eqc : " << t << std::endl;
     //we care about the length of this string
     d_termReg.registerTerm(t[0], 1);
+
+    eq::EqualityEngine* ee = d_state.getEqualityEngine();
+    Node r = ee->getRepresentative(t[0]);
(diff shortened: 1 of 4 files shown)


src/theory/strings/theory_strings.cpp
```diff
@@ -749,10 +749,62 @@ void TheoryStrings::eqNotifyNewClass(TNode t){
...
+    {
+      ei->d_codeTerm = t[0];
+    }
   }
   d_eagerSolver.eqNotifyNewClass(t);
 }
 
+void TheoryStrings::eqNotifyMerge(TNode t1, TNode t2)
+{
+  d_eagerSolver.eqNotifyMerge(t1, t2);
```
ajreynol, line 770: Can you move this to line 777 (or maybe to line 779 even)?  Also it might be worth considering passing e1, e2 to this method to avoid extra lookups.

ajreynol (changes requested):

src/theory/strings/theory_strings.cpp
4tXJ7f: Yes, that makes sense. I have added `e1` and `e2` as arguments to `eqNotifyMerge`.

ajreynol (approved):

PR #7629, opened 2021-11-10
Add some basic rewrites for regular expression intersection

ajreynol:
Adds some basic rewrites for re.inter and re.union involving children of the form `(str.to_re c)` for constants `c`.

This is towards addressing bottlenecks for Zelkova benchmarks.


src/theory/strings/sequences_rewriter.cpp
```diff
@@ -1010,13 +1016,100 @@ Node SequencesRewriter::rewriteAndOrRegExp(TNode node)
...
+            return returnRewrite(
+                node, ret, Rewrite::RE_INTER_CONST_RE_CONFLICT);
+          }
+        }
+      }
+    }
+    if (!toRemove.empty())
+    {
+      std::vector<Node> nodeVecTmp = node_vec;
+      node_vec.clear();
```
4tXJ7f, line 1103: To avoid an unnecessary copy, we can do:

```suggestion
      std::vector<Node> nodeVecTmp;
      node_vec.swap(nodeVecTmp);
```

4tXJ7f (approved):
LGTM, just minor suggestions to be considered before merging.

PR #7630, opened 2021-11-10
Cancel SYMM in CDProof, throw an error for cyclic proofs during double SYMM elimination

ajreynol:
This fixes the potential for cyclic proofs in CDProof when P and (SYMM P) are added as proofs to a CDProof with automatic managing of symmetry.

This manifested as non-termination on the seqArray branch with proofs enabled, instead, if this were to occur, we should throw an error, which is easy to catch.


HanielB (approved):
LGTM

PR #7631, opened 2021-11-10
Fixes for update/nth over constant sequences

ajreynol:
This includes fixes related to how we process seq.update / seq.nth applied to sequence constants.


yoni206 (changes requested):
Can you include the benchmark that failed before this fix?

ajreynol:
No, since this requires the remainder of `seqArray` to be merged. I have however added the regression that failed on that branch, the regression with be merged when the rest of that branch is merged to master.

PR #7633, opened 2021-11-11
Generalize eager length bound conflicts for regular expression memberships

ajreynol:
This generalizes eager length bound conflicts to take into account regular expression memberships.

For example:

If `(str.in_re x (re.++ (re.* re.allchar) (str.to_re "abc") (re.* re.allchar))` is asserted, then we know `(str.len x) >= 3`.

If e.g. equivalence class of `x` is merged with `(str.substr y 0 2)`, we get the conflict

`(and (str.in_re x (re.++ (re.* re.allchar) (str.to_re "abc") (re.* re.allchar)) (= x (str.substr y 0 2))`

since `(str.len (str.substr y 0 2)) <= 2`.

This PR also does some minor refactoring to eager prefix conflicts to make it more analogous to our implementation of length conflicts.


src/theory/strings/regexp_entail.cpp
```diff
@@ -696,6 +703,70 @@ Node RegExpEntail::getFixedLengthForRegexp(Node n)
...
+        rr += r;
+      }
+      else if (firstTime)
+      {
+        rr = r;
+        firstTime = false;
+      }
+      else if ((k == REGEXP_UNION) == isLower)
+      {
+        rr = r < rr ? r : rr;
```
4tXJ7f, line 750:
```suggestion
        rr = std::min(r, rr);
```

src/theory/strings/regexp_entail.cpp
```diff
@@ -696,6 +703,70 @@ Node RegExpEntail::getFixedLengthForRegexp(Node n)
...
+      }
+      else
+      {
+        rr = r > rr ? r : rr;
```
4tXJ7f, line 754:
```suggestion
        rr = std::max(r, rr);
```

src/theory/strings/regexp_entail.cpp
```diff
@@ -696,6 +703,70 @@ Node RegExpEntail::getFixedLengthForRegexp(Node n)
...
+  }
+  else if (k == REGEXP_UNION || k == REGEXP_INTER || k == REGEXP_CONCAT)
+  {
+    bool success = true;
+    bool firstTime = true;
+    Rational rr(0);
+    for (const Node& nc : n)
+    {
+      Node bc = getConstantBoundLengthForRegexp(nc, isLower);
+      if (bc.isNull())
```
4tXJ7f, line 732: That seems overly conservative for lower bounds on concats or bounds on intersections.

I think that it should be something like `bc.isNull() && (k == REGEXP_UNION || (k == REGEXP_CONCAT && !isLower))` (and `success` would have to be set differently).

4tXJ7f (changes requested):
Just a minor request for making the bounds procedure a bit more aggressive.

src/theory/strings/regexp_entail.cpp
ajreynol: Yeah you're right, I'll improve.

ajreynol:
Made this optional, disabled by default, as it makes a key regression time out.  We can proceed with this PR now.

src/theory/strings/arith_entail.cpp
```diff
@@ -885,12 +892,12 @@ Node ArithEntail::getConstantBoundLength(TNode s, bool isLower) const
...
-    if (success)
+    if (success && (!isLower || sum.sgn() != 0))
     {
       ret = nm->mkConst(CONST_RATIONAL, sum);
     }
   }
-  else if (isLower)
+  if (ret.isNull() && isLower)
   {
     ret = d_zero;
```
4tXJ7f, line 902: Shouldn't we remove this case? It doesn't match the comment in the header file.

4tXJ7f (changes requested):
Changes look good to me. There is just one minor question regarding a case in `getConstantBoundLength()`.

src/theory/strings/arith_entail.cpp
```diff
@@ -852,18 +854,23 @@ Node ArithEntail::getConstantBound(Node a, bool isLower)
...
   {
     return ret;
   }
   NodeManager* nm = NodeManager::currentNM();
   if (s.isConst())
   {
-    ret = nm->mkConst(CONST_RATIONAL, Rational(Word::getLength(s)));
+    size_t len = Word::getLength(s);
+    // return null for lower bound on length
+    if (!isLower || len != 0)
```
4tXJ7f, line 870: I think that we can remove this case, too, now.

ajreynol: Good catch.

4tXJ7f (approved):
LGTM, thanks for the updates.

PR #7634, opened 2021-11-11
Generalize front-end checks to check for shadowed variables

ajreynol:
When using the API, we currently check for free variables in assertions and in terms passed to get-value. 

We also require checking for variable shadowing.

This generalizes the utility method so that both free variables and shadowed variables are checked and reported to the user.

### DIFF
--- a/src/expr/node_algorithm.cpp
+++ b/src/expr/node_algorithm.cpp
@@ -282,12 +282,18 @@ bool hasBoundVar(TNode n)
 }
 
 /**
- * Get the free variables in n, that is, the subterms of n of kind
+ * Check variables internal, which is used as a helper to implement many of the
+ * methods in this file.
+ *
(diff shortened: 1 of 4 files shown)


src/expr/node_algorithm.cpp
```diff
@@ -406,6 +387,95 @@ bool getFreeVariablesScope(TNode n,
...
+  if (!n.getAttribute(HasClosureComputedAttr()))
+  {
+    bool hasC = false;
+    if (n.isClosure())
+    {
+      hasC = true;
+    }
+    else
+    {
+      for (auto i = n.begin(); i != n.end() && !hasC; ++i)
```
aniemetz, line 445:
```suggestion
      for (const auto& i = n.begin(); i != n.end() && !hasC; ++i)
```

aniemetz (approved):
One minor thing, else LGTM.

PR #7637, opened 2021-11-12
Preparations for eliminating arithmetic subtyping

ajreynol:
Adds TypeNode::isArithmetic, NodeManager::mkConstReal and NodeManager::mkConstInt.

The first method (for now) is equivalent to TypeNode::isReal, and the latter 2 are equivalent to NodeManager::mkConst(CONST_RATIONAL, ...).

This furthermore distinguishes all uses of `isReal`: all that intend to be `isArithmetic` are changed in this PR.  Redundant uses of `isReal() || isInteger()` are merged to `isArithmetic()`.

Due to the above, there are no behavior changes in this PR.


src/theory/quantifiers/term_util.cpp
```diff
@@ -557,7 +557,7 @@ Node TermUtil::isSingularArg(Node n, Kind ik, unsigned arg)
   }
   else
   {
-    if (n.getType().isReal() && n.getConst<Rational>().sgn() < 0)
+    if (n.getType().isArithmetic() && n.getConst<Rational>().sgn() < 0)
```
4tXJ7f, line 560: This should eventually become `n.getType().isInteger()`, no?

src/theory/quantifiers/sygus_sampler.cpp
```diff
@@ -617,7 +617,7 @@ Node SygusSampler::getRandomValue(TypeNode tn)
       return ret;
     }
   }
-  else if (tn.isReal())
+  else if (tn.isArithmetic())
```
4tXJ7f, line 620: Shouldn't this remain `tn.isReal()`? The integer case is handled above.

src/theory/quantifiers/cegqi/ceg_arith_instantiator.cpp
```diff
@@ -191,7 +191,7 @@ bool ArithInstantiator::processAssertion(CegInstantiator* ci,
         }
         else
         {
-          Assert(d_type.isReal());
+          Assert(d_type.isArithmetic());
```
4tXJ7f, line 194:
```suggestion
          Assert(d_type.isReal());
```
The integer case is handled above.

src/theory/quantifiers/cegqi/ceg_arith_instantiator.cpp
```diff
@@ -262,7 +262,7 @@ bool ArithInstantiator::processAssertion(CegInstantiator* ci,
       }
       else
       {
-        Assert(d_type.isReal());
+        Assert(d_type.isArithmetic());
```
4tXJ7f, line 265:
```suggestion
        Assert(d_type.isReal());
```

The integer case is handled above.

4tXJ7f (changes requested):
Some minor comments and we should consider changing the name of `isArithmetic()` as discussed during the meeting today.

src/theory/quantifiers/term_util.cpp
ajreynol: True, I think we can change it now.

4tXJ7f (approved):
LGTM, thanks for the updates.

PR #7638, opened 2021-11-12
[proofs] Make sure --proof-check=... is no-op when not checking proofs

HanielB:
Fixes https://github.com/cvc5/cvc5-projects/issues/342


src/smt/solver_engine.cpp
```diff
@@ -799,20 +799,10 @@ Result SolverEngine::checkSatInternal(const std::vector<Node>& assumptions,
...
     }
     // Check that UNSAT results generate a proof correctly.
-    if (d_env->getOptions().smt.checkProofs
-        || d_env->getOptions().proof.proofCheck
-               == options::ProofCheckMode::EAGER)
+    if (d_env->getOptions().smt.checkProofs)
     {
       if (r.asSatisfiabilityResult().isSat() == Result::UNSAT)
       {
-        if ((d_env->getOptions().smt.checkProofs
```
ajreynol, line 808: This should simplify to the 2nd disjunct.

ajreynol (changes requested):

src/smt/solver_engine.cpp
HanielB: Right, but I thought of reproducing the behavior of checkModels above, where this is not checked. In setDefaults checkProofs cannot be set to true if produceProofs is not

ajreynol (approved):

PR #7643, opened 2021-11-13
Add rewrite rule for bag.card operator using bag.map and lambda

mudathirmahgoub:
Add rewrite rule for bag.card operator using bag.map and lambda


ajreynol:
I think we need to be careful about this.  I'm not sure I would add this to the rewriter.  Instead, can we add it as an (optional) rewrite in `ppRewrite`?

The problem is that introducing lambdas further constrains the logic: it makes it so that you need HO_ logic to reason about bag.card.

ajreynol (approved):

PR #7644, opened 2021-11-13
Skip `str.code` inferences for sequence eqcs

4tXJ7f:
Fixes https://github.com/cvc5/cvc5-projects/issues/340. Type checking
failed because cvc5 was trying to construct a term `(str.to_code
(seq.unit false))`. We do not allow the construction of terms
`(str.to_code t)` where `t` is not of type string. This commit fixes the
issue by skipping sequence equivalence classes when doing inferences
related to `str.to_code`.

Note that the regression test is slightly different from the original
unit test. It asserts that the index passed to `seq.nth` is
non-negative, which ensures that we can check the resulting model. I
have checked that the modified regression test triggers the issue before
the change in this commit.


ajreynol (approved):
LGTM, thanks for the fix.

PR #7600, opened 2021-11-08
Minor optimizations for term database

ajreynol:
A few minor optimizations for term database.

Also collects private/public blocks in quantifiers engine.

PR #7645, opened 2021-11-15
Revert change and clean datatypes cons candidate generator

ajreynol:
PR https://github.com/cvc5/cvc5/pull/7600/ refactored and optimized candidate generator.  This PR actually corrected an issue where the datatypes constructor expansion candidate generator did *not* generate instances for triggers whose top symbol is a constructor.  However, fixing it to generate instances led to a performance regression on Facebook benchmarks.  

This PR reverts the behavior change from that PR, and moreover makes it explicit that we do not generate instances in this case.

FYI @barrettcw


HanielB:
If I understand correctly, this PR reverts to not generating instances for triggers that have constructors as toy symbols? Shouldn't this maybe be controlled via an option? It sounds like generating instances for such triggers should be on by default (and could be disabled for the FB benchmarks via the option).

ajreynol:
I've added this option, but it should be disabled by default.  This is used as a liberal heuristic for creating constructor terms to match constructors as triggers.  E.g. y ---> cons(head(y), tail(y)) to match with cons(p1, p2).

HanielB (approved):
I see. LGTM then.

PR #7647, opened 2021-11-16
Update SimpleVC.java with try with resource

mudathirmahgoub:
Update SimpleVC.java with try with resource missed in previous  PR


4tXJ7f (approved):
LGTM

PR #7649, opened 2021-11-16
Clean up relationship of metakind and node_manager

4tXJ7f:
Before this commit, we were including `metakind.h` twice in
`node_manager.h`, once without and once with defining
`CVC5__NODE_MANAGER_NEEDS_CONSTANT_MAP`. Additionally, `mkmetakind`
generated structs that mapped types to kinds. This commit makes all of
this obsolete by directly generating instantiations of `mkConst`, which
allows us to get rid of the double include and the `ConstantMap`.


src/expr/CMakeLists.txt
```diff
@@ -168,6 +168,16 @@ add_custom_command(
   DEPENDS mkmetakind metakind_template.cpp metakind.h ${KINDS_FILES}
 )
 
+add_custom_command(
+  OUTPUT node_manager.cpp
+  COMMAND
+    ${mkmetakind_script}
+    ${CMAKE_CURRENT_LIST_DIR}/node_manager_template.cpp
+    ${KINDS_FILES}
+    > ${CMAKE_CURRENT_BINARY_DIR}/node_manager.cpp
```
nafur, line 177: to my knowledge, using such shell built-ins is discouraged in cmake commands. Instead, the script should directly write to that file...

4tXJ7f: Hmm, this just copies what we do for the other commands. How about I clean that up in a future PR?

nafur (approved):
LGTM.

PR #7652, opened 2021-11-16
Add documentation for z3py compatibility API

nafur:
This PR includes documentation for the z3py compatibility API into our general API.
It does so by adding the z3py compatibility API as an external project to download it and then have sphinx document it via the `autodoc` extension that we already use for our regular python API.
Right now we simply show everything on one page, which should be refactored in the future.


alex-ozdemir:
Looks good to me, but someone else should take a look too.

alex-ozdemir (approved):

docs/api/python/python.rst
```diff
@@ -1,25 +1,25 @@
...
-    datatypedecl
-    datatypeselector
-    grammar
-    op
-    result
-    roundingmode
-    solver
-    sort
-    unknownexplanation
+Most of the time, use the :doc:`z3py compatibility API <z3compat/z3compat>`.
```
yoni206, line 25: How about saying something more specific, and having arguments for both? For example:
```
If you are a new user, or already have an application that uses Z3's python API, use the :doc:`z3py compatibility API <z3compat/z3compat>`.

If you would like a more feature-complete python API, with the ability to do almost everything that the cpp API allows, use the :doc:`regular Python API <regular/python>`.
```

yoni206 (commented):
LGTM, one comment.

docs/api/python/python.rst
nafur: Thanks!

PR #7654, opened 2021-11-17
Update Python packaging infrastructure

alex-ozdemir:
Changes:
* Stop passing `--lib-only` (it no longer exists)
* Fix the regex search for the cvc5 version
* Take a version suffix as an (optional) environmental variable


src/api/python/wheels/build_wheel.py
```diff
@@ -49,20 +58,21 @@ def get_cvc5_version():
 
     # read CMakeLists.txt to get version number
     version = dict()
-    str_pattern = 'set\(CVC5_(?P<component>MAJOR|MINOR|RELEASE)\s*(?P<version>\d+)\)'
+    str_pattern = 'set\(CVC5_LAST_RELEASE\s*"([^"]+)"\)'
     pattern = re.compile(str_pattern)
-    with open(os.path.join(project_src_path, 'CMakeLists.txt'), 'r') as f:
+    with open(os.path.join(project_src_path, 'cmake', 'version-base.cmake'), 'r') as f:
```
nafur, line 63: I'd suggest to read the whole file in one go (`f.read()`) and apply the regex once to the whole thing. The file will always be small (so there is no memory issue) and the code will be way easier.

alex-ozdemir: Good point. I was going for the "minimal diff" approach, but there's no harm in doing some cleanup.

nafur (approved):
LGTM.

PR #7655, opened 2021-11-17
Fix binding of quoted symbols in `define-fun`

4tXJ7f:
Our `DefineFunctionCommand` was binding the string representation of the
function symbol including the `|` quotes instead of the symbol without
the quotes. This commit fixes the issue.


nafur (approved):
LGTM.

Issue #349, opened 2017-08-19
stack<TNode::const_iterator> is not getting allocated. (Bugzilla #129)


mpreiner:
**Imported from Bugzilla**
Reporter: *Tim King*
Status: *RESOLVED*
Severity: *normal*
Assigned to: *Morgan Deters*
Component: *{Internals} Expression package*
Milestone: *---*
Version: *master*
Platform: *PC*
OS: *Linux*

Original attachment names and IDs:
- _[patch](http://cvc4.cs.stanford.edu/bugzilla_attachments/10_patch)_ (patch to r490's arith_rewriter.cpp, ID 10)

On 2010-05-24 12:08:36 -0700, Tim King wrote:
> I am having a error allocating an stl container for storing TNode::const\_iterators. It seems to be on initial allocation.
> 
> The offending command line:
> gdb \-\-args cvc4 \-\-debug arithrewrite \~/smt/QF\_LRA/clock\_synchro/clocksynchro\_10clocks.main\_invar.induct.smt
> 
> The offending code tarbar:
> http://www.cs.nyu.edu/~taking/cvc4.stack_tnode.tar.bz2
> 
> The offending gdb trace:
> \# 1  0x00007ffff699f5c0 in \*\_\_GI\_abort () at abort.c:92
> 92	abort.c: No such file or directory.
> 	in abort.c
> (gdb) 
> \# 2  0x00007ffff69d54fb in \_\_libc\_message (do\_abort\=<value optimized out>, 
>     fmt\=<value optimized out>) at ../sysdeps/unix/sysv/linux/libc\_fatal.c:189
> 189	../sysdeps/unix/sysv/linux/libc\_fatal.c: No such file or directory.
> 	in ../sysdeps/unix/sysv/linux/libc\_fatal.c
> (gdb) 
> \# 3  0x00007ffff69df5b6 in malloc\_printerr (action\=3, 
>     str\=0x7ffff6ab17d0 "malloc(): memory corruption (fast)", ptr\=<value optimized out>)
>     at malloc.c:6264
> 6264	malloc.c: No such file or directory.
> 	in malloc.c
> (gdb) 
> \# 4  0x00007ffff69e38e9 in \_int\_malloc (av\=0x7ffff6ce5e40, bytes\=64) at malloc.c:4306
> 4306	in malloc.c
> (gdb) 
> \# 5  0x00007ffff69e458e in \*\_\_GI\_\_\_libc\_malloc (bytes\=64) at malloc.c:3660
[...]

PR #7658, opened 2021-11-17
Fix cvc5-projects/issues/349

mudathirmahgoub:
This PR fixes [#349](https://github.com/cvc5/cvc5-projects/issues/349).


src/theory/bags/inference_generator.cpp
```diff
@@ -47,9 +47,12 @@ InferInfo InferenceGenerator::nonNegativeCount(Node n, Node e)
 
   InferInfo inferInfo(d_im, InferenceId::BAGS_NON_NEGATIVE_COUNT);
   Node count = d_nm->mkNode(BAG_COUNT, e, n);
-
-  Node gte = d_nm->mkNode(GEQ, count, d_zero);
+  Node skolem = d_sm->mkPurifySkolem(count, "bag.count");
+  Node equal = d_nm->mkNode(EQUAL, count, skolem);
+  d_state->getEqualityEngine()->assertEquality(equal, true, d_true);
```
mudathirmahgoub, line 52: This line is necessary here although the same lemma is sent  when this InferInfo is processed later because of `inferInfo.d_skolems[skolem] = count;`

ajreynol:
As discussed offline, `getPreprocessedTerm` is the right method to use to ensure we have the right term, although we should not call this method during `getModelValues`.

PR #7660, opened 2021-11-17
Improve naming in term canonization when handling HO variables

HanielB:
Previously all higher-order variables would me named with "-" followed by the index, since the method did not account for functional types (which are printed as with `(-> ...)`). This commit changes it so that it takes the return type, which will always be atomic, thus giving more meaningful names to the canonical variables.

### DIFF
--- a/src/expr/term_canonize.cpp
+++ b/src/expr/term_canonize.cpp
@@ -104,15 +104,22 @@ Node TermCanonize::getCanonicalFreeVar(TypeNode tn, unsigned i, uint32_t tc)
   std::vector<Node>& tvars = d_cn_free_var[key];
   while (tvars.size() <= i)
   {
-    std::stringstream oss;
-    oss << (tn.isFunction() ? tn.getRangeType() : tn);
-    std::string typ_name = oss.str();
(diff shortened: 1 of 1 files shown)


src/expr/term_canonize.cpp
```diff
@@ -105,7 +105,7 @@ Node TermCanonize::getCanonicalFreeVar(TypeNode tn, unsigned i, uint32_t tc)
   while (tvars.size() <= i)
   {
     std::stringstream oss;
-    oss << tn;
```
ajreynol, line 108: Can you perhaps add "fun_" or something to the name here?

ajreynol: Better, can you make the name simply `f`?

HanielB: You mean for all functional-type vars to have the names as `f` ++ `i`?

ajreynol: Yes

HanielB: OK, done

ajreynol (approved):

PR #7671, opened 2021-11-19
Remove n-ary builder utility

ajreynol:
Adds the only usage of this file into ite_simp.cpp, where it is specialized for AND.

This is work towards eliminating arithmetic subtyping (that utility had an unused and ambiguous use of mkConst(CONST_RATIONAL, ...)).


4tXJ7f (approved):
LGTM

PR #7673, opened 2021-11-20
Alethe: Printer Specification

Lachnitt:
Header of the Alethe printer


src/proof/alethe/alethe_printer.h
```diff
@@ -0,0 +1,63 @@
...
+#ifndef CVC4__PROOF__ALETHE_PROOF_PRINTER_H
+#define CVC4__PROOF__ALETHE_PROOF_PRINTER_H
+
+#include "proof/proof_node.h"
+
+namespace cvc5 {
+
+namespace proof {
+
+class AletheProofPrinter
```
HanielB, line 27:
```suggestion
/**
 * The Alethe printer, which prints proof nodes in a Alethe proof, according to
 * the proof rules defined in alethe_proof_rule.h.
 *
 * It expects to print proof nodes that have processed by the Alethe proof post
 * processor.
 */
class AletheProofPrinter
```

src/proof/alethe/alethe_printer.h
```diff
@@ -0,0 +1,63 @@
...
+   * This method prints a proof node that has been transformed into the Alethe
+   * proof format
+   *
+   * @param out The stream to write to
+   * @param pfn The proof node to be printed
+   */
+  void alethePrinter(std::ostream& out, std::shared_ptr<ProofNode> pfn);
+
+ private:
+  /** Used for printing the node after the initial anchor has been printed */
```
HanielB, line 42:
```suggestion
  /** Used for printing the node after the initial Alethe anchor has been
   * printed
   *
   * The initial anchor introduces the initial assumptions of the problem, which
   * correspond to the problem assertions.
   */
```

src/proof/alethe/alethe_printer.h
```diff
@@ -0,0 +1,63 @@
...
+  std::string alethePrinterInternal(std::ostream& out,
+                                    std::shared_ptr<ProofNode> pfn);
+  /** The current level of nesting, which increases if a subproof is entered */
+  int nested_level;
```
HanielB, line 46: You should use `size_t` for numerical values to be used to access indices of standard C++ containers.

Moreover all attributes of a class should have a "d_" at its prefixes and be in caml case. So the above should be `d_nestedLevel`.

src/proof/alethe/alethe_printer.h
```diff
@@ -0,0 +1,63 @@
...
+  /** Current step id */
+  int step_id;
```
HanielB, line 48: ditto

src/proof/alethe/alethe_printer.h
```diff
@@ -0,0 +1,63 @@
...
+  /** The current prefix which is updated whenever a subproof is encountered
+   * E.g. prefix = "t19.t2." */
```
HanielB, line 50:
```suggestion
   *
   * E.g. the prefix "t19.t2." is used when we are under a subproof started at
   * step "t19" and another at "t2" without leaving the first subproof. */
```

src/proof/alethe/alethe_printer.h
```diff
@@ -0,0 +1,63 @@
...
+  std::string prefix;
+  /** A list of assumption lists, one for every level of the nested proof node
+   */
+  std::vector<std::unordered_map<Node, int>> assumptions;
+  /** A list of step lists, one for every level of the nested proof node */
+  std::vector<std::unordered_map<Node, int>> steps;
```
HanielB, line 56: also use `size_t` rather than `int`.

HanielB (changes requested):

src/proof/alethe/alethe_printer.h
```diff
@@ -0,0 +1,63 @@
...
+  AletheProofPrinter();
+  ~AletheProofPrinter() {}
+  /**
+   * This method prints a proof node that has been transformed into the Alethe
+   * proof format
+   *
+   * @param out The stream to write to
+   * @param pfn The proof node to be printed
+   */
+  void alethePrinter(std::ostream& out, std::shared_ptr<ProofNode> pfn);
```
HanielB, line 39:
```suggestion
  void print(std::ostream& out, std::shared_ptr<ProofNode> pfn);
```

src/proof/alethe/alethe_printer.h
```diff
@@ -0,0 +1,63 @@
...
+
+ private:
+  /** Used for printing the node after the initial anchor has been printed */
+  std::string alethePrinterInternal(std::ostream& out,
```
HanielB, line 43:
```suggestion
  std::string printIntenal(std::ostream& out,
```

HanielB (changes requested):

Lachnitt:
@HanielB There have been extensive changes to this file to accommodate to the comments from PR #7675. A completely new look on it might be required.

src/proof/alethe/alethe_printer.h
```diff
@@ -0,0 +1,81 @@
...
+   *
+   * The initial anchor introduces the initial assumptions of the problem, which
+   * correspond to the problem assertions.
+   *
+   * @param out The stream to write to
+   * @param pfn The proof node to be printed
+   * @param assumptions The list of assumptions made before the current step,
+   * that are visible as premises to that step
+   * @param steps The list of steps occuring before the current step, that can
+   * that are visible as premises to that step
```
HanielB, line 60:
```suggestion
   * @param steps The list of steps occurring before the current step, that are
   * visible as premises to that step
```

src/proof/alethe/alethe_printer.h
```diff
@@ -0,0 +1,81 @@
...
+   * under a subproof started at step "t19" and another at "t2" without leaving
+   * the first subproof.
+   * @param current_step_id The id of a step within a subproof (without the
+   * prefix).
+   * @return The full id (including the prefix) of the last step of pfn.
+   */
+  std::string printInternal(std::ostream& out,
+                            std::shared_ptr<ProofNode> pfn,
+                            std::unordered_map<Node, std::string> assumptions,
+                            std::unordered_map<Node, std::string> steps,
```
HanielB, line 72: I think these maps should be passed as constant references?

src/proof/alethe/alethe_printer.h
```diff
@@ -0,0 +1,81 @@
...
+                            std::string current_prefix,
+                            int& current_step_id);
```
HanielB, line 74: Since these are unsigned, use `uint32_t` (we've been avoiding using platform-dependent-types).

src/proof/alethe/alethe_printer.h
Lachnitt: Done. I think it might not make a lot of difference since they are copied anyway.

HanielB (approved):

PR #7674, opened 2021-11-20
Alethe: Implementation of Printer

Lachnitt:
Adds an implementation file of the Alethe printer with function stubs to be filled in in further PRs. 

Should be merged after PR #7673.


src/proof/alethe/alethe_printer.cpp
```diff
@@ -0,0 +1,67 @@
...
+  for (unsigned long int i = 3, size = pfn->getArguments().size(); i < size;
+       i++)
+  {
+    Trace("alethe-printer")
+        << "... print assumption " << pfn->getArguments()[i] << std::endl;
+    out << "(assume a" << std::to_string(i - 3) << " " << pfn->getArguments()[i]
+        << ")\n";
+    assumptions[0][pfn->getArguments()[i]] = i - 3;
+  }
+
```
HanielB, line 52:
```suggestion
  const std::vector<Node>& args = pfn->getArguments();
  // Special handling for the first scope
  // Print assumptions and add them to the list but do not print anchor.
  for (size_t i = 3, size = args.size(); i < size; i++)
  {
    Trace("alethe-printer")
        << "... print assumption " << args[i] << std::endl;
    out << "(assume a" << i - 3 << " " << args[i] << ")\n";
    assumptions[0][args[i]] = i - 3;
  }

```

HanielB (changes requested):

HanielB (approved):

PR #7675, opened 2021-11-20
Alethe: Further Printer Implementation

Lachnitt:
Adds implementation of PrintInternal a function to print a proof node after the first step has been printed.

Should be merged after PR #7674.


src/proof/alethe/alethe_printer.cpp
```diff
@@ -0,0 +1,321 @@
...
+    out << "(assume a" << std::to_string(i - 3) << " " << pfn->getArguments()[i]
+        << ")\n";
+    assumptions[0][pfn->getArguments()[i]] = i - 3;
+  }
+
+  // Then, print the rest of the proof node
+  alethePrinterInternal(out, pfn->getChildren()[0]);
+}
+
+std::string AletheProofPrinter::alethePrinterInternal(
```
HanielB, line 57: this should be deleted, I imagine?

src/proof/alethe/alethe_printer.cpp
```diff
@@ -0,0 +1,321 @@
...
+    std::ostream& out, std::shared_ptr<ProofNode> pfn)
+{
+  // Store current id in case a subproof overwrites step_id
+  int current_step_id = step_id;
+
+  // If the proof node is untranslated a problem might have occured during
+  // postprocessing
+  if (pfn->getArguments().size() < 3 || pfn->getRule() != PfRule::ALETHE_RULE)
```
HanielB, line 86: cache `pfn->getArguments()` as I suggested in the previous PR

src/proof/alethe/alethe_printer.cpp
```diff
@@ -0,0 +1,321 @@
...
+    Trace("alethe-printer")
+        << "... printing failed! Encountered untranslated Node. "
+        << pfn->getResult() << " " << toString(pfn->getRule()) << " "
+        << " / " << pfn->getArguments() << std::endl;
+    return "";
+  }
+
+  // Get the alethe proof rule
+  AletheRule vrule =
+      static_cast<AletheRule>(std::stoul(pfn->getArguments()[0].toString()));
```
HanielB, line 97:
```suggestion
  AletheRule vrule = getAletheRule(args[0]);
```

Also use "arule", as the "v" is coming from the previous name.

src/proof/alethe/alethe_printer.cpp
```diff
@@ -0,0 +1,321 @@
...
+
+    // Store step_id until children are printed to resume counter at current
+    // position
+    current_step_id = step_id;
+    step_id = 1;
+  }
+
+  // Assumptions are printed at the anchor and therefore have to be in the list
+  // of assumptions when an assume is reached.
+  if (vrule == AletheRule::ASSUME)
```
HanielB, line 172:
```suggestion
  // Assumptions are printed at the anchor and therefore have to be in the list
  // of assumptions when an assume is reached.
  else if (vrule == AletheRule::ASSUME)
```

src/proof/alethe/alethe_printer.cpp
```diff
@@ -0,0 +1,321 @@
...
+    // assignments, i.e. args=[(= v0 v1)] is printed as (:= (v0 Int) v1).
+    if (vrule == AletheRule::ANCHOR_BIND)
+    {
+      out << " :args (";
+      for (unsigned long int j = 3, size = pfn->getArguments().size(); j < size;
+           j++)
+      {
+        out << "(:= (" << pfn->getArguments()[j][0].toString() << " "
+            << pfn->getArguments()[j][0].getType().toString() << ") "
+            << pfn->getArguments()[j][1].toString() << ")";
```
HanielB, line 139: You can remove the `toString()` methods here, they'll be called automatically.

src/proof/alethe/alethe_printer.cpp
```diff
@@ -0,0 +1,321 @@
...
+
+    // If the subproof is a genuine subproof the arguments are printed as
+    // assumptions
+    if (vrule == AletheRule::ANCHOR_SUBPROOF)
+    {
+      for (size_t i = 3, size = pfn->getArguments().size(); i < size; i++)
+      {
+        Trace("alethe-printer")
+            << "... print assumption " << pfn->getArguments()[i] << std::endl;
+        out << "(assume " << prefix << "a" << std::to_string(i - 3) << " "
```
HanielB, line 158:
```suggestion
        out << "(assume " << prefix << "a" << i - 3 << " "
```

src/proof/alethe/alethe_printer.cpp
```diff
@@ -0,0 +1,321 @@
...
+  {
+    Trace("alethe-printer")
+        << "... reached assumption " << pfn->getResult() << " " << vrule << " "
+        << " / " << pfn->getArguments() << " " << nested_level << std::endl;
+
+    // While in most cases the assumption is printed at the same level than the
+    // step whose premise it is, it is possible that it is from a different
+    // level. Thus, the whole list needs to be traversed. Since this case is
+    // rare adapting the prefix should be rarely necessary.
+    for (int i = nested_level; i >= 0; i--)
```
HanielB, line 182:
```suggestion
    for (size_t i = d_nestedLevel + 1; i > 0; i--)
```
And change all occurrences of `i` below to `i-1`.

HanielB (changes requested):
A few comments, but more importantly I think this design needs to change, as it
can be simplified if you apply the following principle: rather than keep "global
variables" for controlling the prefix, the number of nesting etc have
`printInternal` take these as arguments. For each new subproof, you do a new
call of the method.

This method will take the current prefix and *copies* of the current maps from
assumptions to ids and step conclusions to ids. Since you have the complication
that some assumptions (but this could also apply to other steps, yes? It was not
clear to me why you only special-cased assumptions) may have been introduced in
an outer scope, the mapped ids are better kept as strings. So when you create a
new step you map its node conclusion to the string you printed as its id. I
believe this will preempt the computation you were doing to remove parts of the
prefix when printing assumptions.

As a consequence of the above design, you also don't need to bother managing the
levels and stacks, as they'll be handled naturally by the recursion.

barrettcw:
Sorry to jump in here, but I get nervous when someone recommends replacing a non-recursive procedure with a recursive one.  In general, we try to avoid recursion in cvc5.  It should only be used when (1) it greatly simplifies things AND (2) we can essentially prove that it will never lead to a stack overflow.  Are you confident about (2) in this situation?

Lachnitt:
> (but this could also apply to other steps, yes? It was not clear to me why you only special-cased assumptions) 

I only observed this behavior for assumptions never for regular steps. Originally, the assumption we operated upon was even that this is not happening at all but I found examples where it does.

Lachnitt:
> A few comments, but more importantly I think this design needs to change, as it can be simplified if you apply the following principle: rather than keep "global variables" for controlling the prefix, the number of nesting etc have `printInternal` take these as arguments. For each new subproof, you do a new call of the method.
> 
> This method will take the current prefix and _copies_ of the current maps from assumptions to ids and step conclusions to ids. Since you have the complication that some assumptions (but this could also apply to other steps, yes? It was not clear to me why you only special-cased assumptions) may have been introduced in an outer scope, the mapped ids are better kept as strings. So when you create a new step you map its node conclusion to the string you printed as its id. I believe this will preempt the computation you were doing to remove parts of the prefix when printing assumptions.
> 
> As a consequence of the above design, you also don't need to bother managing the levels and stacks, as they'll be handled naturally by the recursion.

My reason for this approach with nested global lists is that I don't want to store all assumptions and steps all the time. Rather once a subproof is finished I want to forget about all assumptions and steps in it to keep these maps small.  This does seem more elegant with the proposed approach.

The printing of the ids would indeed be simpler if the whole id is stored in the map. I surmised that the case that an assumption from outside the subproof is used as premise is particularly seldom since the original assumptions was that it never happens at all. Thus, I did not want to store the full id every time but was still not overly happy with the current solution. However, if this is the only reason to change the printer to a recursive function it might be interesting to find out how often this occurs. Because otherwise, I can also integrate this change (store full ids) in the current version and also add the prefix to be recursively given to the function while having assumptions and steps remain the way they are.

HanielB:
> In general, we try to avoid recursion in cvc5. It should only be used when (1) it greatly simplifies things AND (2) we can essentially prove that it will never lead to a stack overflow. Are you confident about (2) in this situation?

@barrettcw The recursion I proposed would only be applied to "subproofs", not to the overall traversal of proof nodes. It's similar if we were, when traversing nodes, only start a new recursive call when we see a quantifier. For this reason the recursion should not be a problem. A similar design is used in the LFSC printer. (Actually currently the whole traversal in the Alethe printer is recursive, but this can be easily changed to iterative, module what I proposed, afterwards, which is what I'm currently planning after we get this first version in.)

> My reason for this approach with nested global lists is that I don't want to store all assumptions and steps all the time. Rather once a subproof is finished I want to forget about all assumptions and steps in it to keep these maps small. This does seem more elegant with the proposed approach.
>
> The printing of the ids would indeed be simpler if the whole id is stored in the map. I surmised that the case that an assumption from outside the subproof is used as premise is particularly seldom since the original assumptions was that it never happens at all. Thus, I did not want to store the full id every time but was still not overly happy with the current solution. However, if this is the only reason to change the printer to a recursive function it might be interesting to find out how often this occurs. Because otherwise, I can also integrate this change (store full ids) in the current version and also add the prefix to be recursively given to the function while having assumptions and steps remain the way they are.

@Lachnitt The motivation to add the recursion I suggested was for simplicity of the code overall. Storing the full id would allow removing the code you put in place to compute when something from outside needs a different id, so it'd simplify the printing code at no real cost added.

src/proof/alethe/alethe_printer.cpp
```diff
@@ -0,0 +1,265 @@
...
+  }
+
+  // Print children
+  std::vector<std::string> child_prefixes;
+  std::string prefix = current_prefix;
+  if (current_prefix != "" && current_prefix.back() != '.')
+  {
+    prefix.append(".");
+  }
+  for (const std::shared_ptr<ProofNode> child : pfn->getChildren())
```
HanielB, line 168:
```suggestion
  const std::vector<std::shared_ptr<ProofNode>>& children = pn->getChildren();
  for (const std::shared_ptr<ProofNode>& child : children)
```

src/proof/alethe/alethe_printer.cpp
```diff
@@ -0,0 +1,265 @@
...
+  {
+    // Look up if subproof has already been printed
+    auto it = steps.find(args[2]);
+    if (it != steps.end())
+    {
+      Trace("alethe-printer")
+          << "... subproof is already printed " << pfn->getResult() << " "
+          << arule << " / " << args << std::endl;
+      return it->second;
+    }
```
HanielB, line 86: you're checking this but I'm not sure where you are adding to the map?

src/proof/alethe/alethe_printer.cpp
```diff
@@ -0,0 +1,265 @@
...
+    return "";
+  }
+
+  // Print children
+  std::vector<std::string> child_prefixes;
+  std::string prefix = current_prefix;
+  if (current_prefix != "" && current_prefix.back() != '.')
+  {
+    prefix.append(".");
+  }
```
HanielB, line 167: I'm not sure if I missed something but couldn't you add the "." already when you update the current prefix?

src/proof/alethe/alethe_printer.cpp
```diff
@@ -0,0 +1,265 @@
...
+
+  // Print current step
+  Trace("alethe-printer") << "... print node " << pfn->getResult() << " "
+                          << arule << " / " << args << std::endl;
+  std::string current_t =
+      current_prefix + "t" + std::to_string(current_step_id);
+  out << "(step " << current_t << " ";
+  out << args[2] << " :rule " << arule;
+  current_step_id = current_step_id + 1;
```
HanielB, line 222:
```suggestion
  out << "(step " << current_t << " " << args[2] << " :rule " << arule;
  current_step_id++;
```

src/proof/alethe/alethe_printer.cpp
```diff
@@ -0,0 +1,265 @@
...
+    if (arule == AletheRule::ANCHOR_SUBPROOF)
+    {
+      out << " :discharge (";
+      for (unsigned long int j = 0; j < current_assumptions.size(); j++)
+      {
+        out << current_assumptions[j];
+        if (j != current_assumptions.size() - 1)
+        {
+          out << " ";
+        }
```
HanielB, line 196:
```suggestion
      for (size j = 0, size = current_assumptions.size(); j < size; j++)
      {
        out << current_assumptions[j]
            << (j != current_assumptions.size() - 1 ? " " : "");
```

src/proof/alethe/alethe_printer.cpp
```diff
@@ -0,0 +1,265 @@
...
+      }
+      else
+      {
+        out << args[i];
+      }
+      if (i != args.size() - 1)
+      {
+        out << " ";
+      }
+    }
```
HanielB, line 240:
```suggestion
    for (size_t i = 3, size = args.size(); i < size; i++)
    {
      if (arule == AletheRule::FORALL_INST)
      {
        out << "(:= " << args[i][0] << " " << args[i][1] << ")";
      }
      else
      {
        out << args[i];
      }
      if (i != size - 1)
      {
        out << " ";
      }
    }
```

But for the `AletheRule::FORALL_INST` case distinction there there is a better way to handle this, which will allow the printing to be uniform: in the post processor, when processing FORALL_INST, you can transform its list of arguments into a list of arguments where each element is of the form `(SEXPR ":=" args[i][0] args[i][1])`. This way you can simplify the printer.

src/proof/alethe/alethe_printer.cpp
```diff
@@ -0,0 +1,265 @@
...
+  {
+    out << " :premises (";
+    for (unsigned long int i = 0, size = child_prefixes.size(); i < size; i++)
+    {
+      out << child_prefixes[i];
+      if (i != child_prefixes.size() - 1)
+      {
+        out << " ";
+      }
+    }
```
HanielB, line 253:
```suggestion
    for (size_t i = 0, size = child_prefixes.size(); i < size; i++)
    {
      out << child_prefixes[i] << (i != size - 1? " " : "");
    }

**Review by HanielB — CHANGES_REQUESTED** [review]

File: src/proof/alethe/alethe_printer.cpp (hunk 9 above)
Comment (Lachnitt) [reply src/proof/alethe/alethe_printer.cpp:86]: That's because I did not add it :sweat_smile: It is fixed now.

File: src/proof/alethe/alethe_printer.cpp (hunk 10 above)
Comment (Lachnitt) [reply src/proof/alethe/alethe_printer.cpp]: You can but you have to delete the last "." then if you end the subproof. I implemented this now.

File: src/proof/alethe/alethe_printer.cpp (hunk 1 above)
Comment (Lachnitt) [reply src/proof/alethe/alethe_printer.cpp]: Why that?

File: src/proof/alethe/alethe_printer.cpp (hunk 13 above)
Comment (Lachnitt) [reply src/proof/alethe/alethe_printer.cpp:239]: After discussion this with @HanielB it came out that there is some work elsewhere needed to print ":=" as := and not as |:=|.  Thus, we decided to keep this handling for now and change it to the more convenient way in a separate PR.

File: src/proof/alethe/alethe_printer.cpp [hunk 15]
```diff
@@ -0,0 +1,266 @@
...
+  Trace("alethe-printer") << "... print node " << pfn->getResult() << " "
+                          << arule << " / " << args << std::endl;
+  std::string current_t =
+      current_prefix + "t" + std::to_string(current_step_id);
+  out << "(step " << current_t << " " << args[2] << " :rule " << arule;
+  current_step_id++;
+  if (args.size() > 3)
+  {
+    out << " :args (";
+    for (unsigned long int i = 3, size = args.size(); i < size; i++)
```
HanielB, line 231:
```suggestion
    for (size_t i = 3, size = args.size(); i < size; i++)
```

HanielB (approved):

src/proof/alethe/alethe_printer.cpp
HanielB: This was on something you already changed. There was some code duplication before.

PR #7676, opened 2021-11-20
Make `node_value.h` not depend on `node_manager.h`

4tXJ7f:
This commit breaks a circular dependency by making `node_value.h` not
depend on `node_manager.h`. As a result, we can remove the hack-y
include of `node_manager.h` in the middle of the `node_value.h` file.


src/expr/node_value.cpp
```diff
@@ -94,5 +94,68 @@ NodeValue::iterator<NodeTemplate<false> > operator+(
...
+  Warning().flush();
+}
+
+void __attribute__((used)) debugPrintRawNodeValue(const expr::NodeValue* nv)
+{
+  nv->printAst(Warning(), 0);
+  Warning().flush();
+}
+
+#endif /* CVC5_DEBUG */
```
nafur, line 159: This stuff has been removed in #7555.

nafur (approved):
LGTM, except for the merge conflicts...

PR #7679, opened 2021-11-20
Improve error for check theory assertions with model

ajreynol:
Makes it so that we report *all* unsatisfied assertions, not just the first one.


4tXJ7f (approved):
LGTM

PR #7680, opened 2021-11-22
Enable model-based reduction technique for strings

ajreynol:
This PR makes change our default strategy for deferring (some) reductions until after the model is constructed. It introduces the option `--strings-mbr` (model-based reduction) which is enabled by default.  

When using model-based reductions, only *negatIve* contains and negative memberships are deferred for reduction/unfolding until LAST_CALL effort, where a candidate model is available. These steps are performed only if the constraints are not already satisfied in the model.  The intuition is that negative contains/membership are the *most* expensive constraints to process and are moreover the *least* likely to be false in the model.

It makes a few fixes to the extended/RE solvers:
- 2 kinds of inferences in `checkExtfEval` should not be performed for substitutions based on models
- The regular expression solver should not use inclusion tests to justify reduction of memberships when the basis for the reduction is an unfolded membership, due to the reasoning being potentially cyclic.  This led to a bogus model on a regression when the new strategy was enabled.

It also does minor refactoring of those solvers that was required for implementing the new policy.

This branch is +446-1 on SMT-LIB, with all gains coming on "sat" benchmarks, mostly from pyex.  It also solves 2 previously unsolved challenge Amazon benchmarks quickly.


src/theory/strings/extf_solver.cpp
```diff
@@ -93,77 +94,72 @@ bool ExtfSolver::doReduction(int effort, Node n)
...
-          return true;
-        }
-        else
+        Trace("strings-extf-debug")
+            << "  resolve extf : " << n
+            << " based on equal lengths disequality." << std::endl;
+        // We can reduce negative contains to a disequality when lengths are
+        // equal. In other words, len( x ) = len( s ) implies
+        //   ~contains( x, s ) reduces to x != s.
+        if (!d_state.areDisequal(x, s))
```
4tXJ7f, line 132: We could consider making this inference regardless of `stringModelBasedReduction`, I think.

4tXJ7f (commented):
LGTM, I have to think some more about the regexp part and have a suggestion.

src/theory/strings/extf_solver.cpp
ajreynol: True, but I don't think its necessary.

Actually, I am considering making disequality handling also deferred to model effort.

Notice that `x != s` would only be useful if `x = s`, but in that case we should get `contains(x,s) ---> true`.

4tXJ7f (approved):
LGTM

PR #7684, opened 2021-11-22
bags: register count terms instead of elements

mudathirmahgoub:
bags: register count terms instead of bag elements


src/theory/bags/solver_state.cpp
```diff
@@ -43,9 +43,12 @@ void SolverState::registerBag(TNode n)
 void SolverState::registerCountTerm(TNode n)
 {
   Assert(n.getKind() == BAG_COUNT);
-  Node element = getRepresentative(n[0]);
   Node bag = getRepresentative(n[1]);
-  d_bagElements[bag].insert(element);
+  Node element = getRepresentative(n[0]);
```
mudathirmahgoub, line 47:
```suggestion
  Node pCount = d_valuation.getPreProcessedTerm(n);
```

mudathirmahgoub:
Replaced by #7743

PR #7686, opened 2021-11-23
Turn kinds in python API into a proper Enum

nafur:
This PR does multiple things:
- the kinds are changed from custom objects to a proper `enum.Enum` class (including according changes to the cython code and the kind generation scripts)
- all examples and tests are modified to account for the change how to use kinds (`Kind` instead of `kinds`)
- add docstrings to the kind enum values
- add a custom documenter that properly renders enums via autodoc
- extend doxygen setup so that we can write comments as `rst` (allowing us to copy the documentation for kinds from the cpp api to the other apis)


docs/ext/autoenum.py
```diff
@@ -0,0 +1,41 @@
...
+
+        # overriding this flag prints __doc__ just as we want to.
+        self.doc_as_attr = False
+        super().add_content(more_content, no_docstring)
+        self.doc_as_attr = True
+
+
+def setup(app: Sphinx) -> None:
+    app.setup_extension('sphinx.ext.autodoc')
+    app.add_autodocumenter(EnumDocumenter)
```
yoni206, line 41:
```suggestion
    app.add_autodocumenter(EnumDocumenter)
    
```

examples/api/python/id.py
```diff
@@ -16,7 +16,7 @@
 ##
 
 import pycvc5
-from pycvc5 import kinds
+from pycvc5 import Kind
```
yoni206, line 19: Maybe this can even be deleted?

yoni206 (approved):
LGTM, modulo minor comments inline.

yoni206:
added do-not-merge to allow a review from @aniemetz before merging.

docs/api/python/regular/python.rst
```diff
@@ -1,4 +1,4 @@
-Python API Documentation
+regular Python API
```
aniemetz, line 1:
```suggestion
Regular Python API Documentation
```

docs/api/python/python.rst
```diff
@@ -1,4 +1,4 @@
-Python API Documentation
+Python API
```
aniemetz, line 1: Should the visibility of the individual api toctrees really be hidden?
Also: `compatibility API` -> `Compatibility Python API` and `regular Python API` -> `Regular Python API`
Can you make the former change also for the header of the compat api page?

aniemetz (changes requested):

aniemetz (approved):

PR #7687, opened 2021-11-23
Fix potential for cycles in trust substitutions

ajreynol:
This ensures we use only the prefix of substitutions for the *first* time a formula is proven in a substitution map.

This avoids the possibility for cycles in proof generators during non-clausal simplification, where we may reprove a formula F later at a point where later substitutions depend on F.


4tXJ7f (approved):
LGTM

PR #7695, opened 2021-11-24
cmake: Add option --[no]-static-binary.

mpreiner:
Allows to disable linking against static system libraries explicitly.


nafur (approved):
LGTM

PR #7698, opened 2021-11-25
Guard regression tests that require libpoly to pass.

mpreiner:
Regression tests either timeout or return unknown if cvc5 is configured with --no-poly.


nafur (approved):
LGTM.

PR #7702, opened 2021-11-29
Enable Java examples

mudathirmahgoub:


examples/CMakeLists.txt
```diff
@@ -85,23 +85,39 @@ add_subdirectory(api/cpp)
 
 if(TARGET cvc5::cvc5jar)
   find_package(Java REQUIRED)
+  find_package(JNI REQUIRED)
```
4tXJ7f, line 88: Why do we need JNI for the examples?

4tXJ7f (approved):
LGTM

examples/CMakeLists.txt
mudathirmahgoub: We need them to run ctest java examples. Without adding the 2 directories that contain libjawt.so and libjvm.so, ctest fails with unsatisfied linking errors. 
It seems this was not needed in cvc4 because SWIG. 
```bash
/usr/lib/jvm/default-java/lib/libjawt.so;
/usr/lib/jvm/default-java/lib/server/libjvm.so
```

examples/CMakeLists.txt
```diff
@@ -85,23 +85,39 @@ add_subdirectory(api/cpp)
...
+  set(JNI_LIBRARIES_PATHS "")
+  foreach(LIB ${JNI_LIBRARIES})
+    get_filename_component(LIB_PATH ${LIB} DIRECTORY)
+    set(JNI_LIBRARIES_PATHS "${JNI_LIBRARIES_PATHS}${LIB_PATH}:")
+  endforeach()
+
+  message(STATUS "JNI_LIBRARIES_PATHS: ${JNI_LIBRARIES_PATHS}")
+  # get directory build/install/lib where libcvc5jni.so is installed
+  get_target_property(CVC5_LIB_FILE cvc5::cvc5 LOCATION)
+  get_filename_component(CVC5_JNI_PATH ${CVC5_LIB_FILE} DIRECTORY)
```
mudathirmahgoub, line 102: This works for now, because libcvcjni is installed in the same directory of libcvc5.so. 
It would fail if there are installed in different directories. 
Is there a simple way to get a path to `build/install/lib` for libcvcjni5 without adding exporting it in a cmake file?

PR #7707, opened 2021-11-30
Integrate the offline opt solver into SolverEngine and implement the API function in SolverEngine

Ouyancheng:
This PR integrates the offline opt solver into SolverEngine. 
And the API functions, though not exposed yet, are implemented in SolverEngine. 
For re-using get-value for retrieving optimal values, we need inline optimization, which needs to integrate the opt solver into a layer somewhere lower than SmtSolver (or even lower than PropEngine...).


Ouyancheng:
@ajreynol @yoni206 (this could be a demo on how the API looks like, it's fine not merging it, and it seems like it's possible to lower the solver to PropEngine and I am planning to work on it.

PR #7708, opened 2021-11-30
Add unit tests for api::Solver::setOption()

nafur:
This PR adds a generic unit test for `api::Solver::setOption()`. It iterates over all options and calls `setOption()` with somewhat cleverly chosen values according to `getOptionInfo()`.
While building this, it occurred to me that calling `std::exit()` in an option handler is weird (for example for `version`) and we should do this in the driver instead. This PR also refactors this issue, moving all calls to `std::exit()` from the option handlers into the driver.


test/unit/options/CMakeLists.txt
```diff
@@ -0,0 +1,17 @@
+###############################################################################
+# Top contributors (to current version):
+#   Aina Niemetz
```
aniemetz, line 3:
```suggestion
#   Gereon Kremer
```

test/unit/options/options_black.cpp
```diff
@@ -0,0 +1,168 @@
+/******************************************************************************
+ * Top contributors (to current version):
+ *   Aina Niemetz, Mudathir Mohamed, Andrew Reynolds
```
aniemetz, line 3:
```suggestion
 *   Gereon Kremer
```

test/unit/options/options_black.cpp
```diff
@@ -0,0 +1,168 @@
...
+    {
+      testing::internal::CaptureStdout();
+    }
+    auto info = d_solver.getOptionInfo(name);
+
+    try
+    {
+      std::visit(
+          overloaded{
+              [this, &name](const OptionInfo::VoidInfo& v) {
```
aniemetz, line 61: Do we still actually use void options after this PR?

test/unit/options/options_black.cpp
```diff
@@ -0,0 +1,168 @@
...
+                             CVC5ApiOptionException);
+                for (const auto& m : v.modes)
+                {
+                  d_solver.setOption(name, m);
+                }
+              },
+          },
+          info.valueInfo);
+    }
+    catch (const CVC5ApiOptionException&)
```
aniemetz, line 157: None of these tests should throw and not be wrapped in `ASSERT_THROW` or `EXPECT_THROW`, right?

src/options/options_handler.cpp
```diff
@@ -96,8 +97,7 @@ Languages currently supported as arguments to the --output-lang option:
   tptp                           TPTP format
   ast                            internal format (simple syntax trees)
 )FOOBAR" << std::endl;
-    std::exit(1);
-    return Language::LANG_AUTO;
+    throw OptionException("help is not a valid language");
```
aniemetz, line 100: Why `help`?

test/unit/options/options_black.cpp
nafur: Indeed we have two `void` options left: `verbose` and `quiet` (which are really just triggers to modify `verbosity`).
We may want to trade having two additional bools (that are never accessed) in favor of removing the possibility of `void`, I don't have a strong opinion on this.

src/options/options_handler.cpp
nafur: We need some way to abort the solver here for `--lang help`. I want to avoid actually aborting the binary right here, but instead communicate this to the API via an exception. It may very well not the proper exception, though.

test/unit/options/options_black.cpp
aniemetz: Fewer special cases is probably better. I also don't have a strong opinion on this, but maybe we should get rid of void options entirely.

src/options/options_handler.cpp
aniemetz: I missed that this is the `help` case. `OptionException` looks like the right one to me.

test/unit/options/options_black.cpp
nafur: I think I agree, but I'd like to do that in a separate PR.

aniemetz (approved):

PR #7710, opened 2021-11-30
Alethe: Add function that adds final steps to proof

Lachnitt:
There are some cases when the result has to be transformed, e.g. when it was printed as (cl false) which is correct everywhere except in the last step. This function then transforms the result into (cl). 

Should be merged after PR #7675.


src/proof/alethe/alethe_post_processor.h
```diff
@@ -50,6 +50,24 @@ class AletheProofPostprocessCallback : public ProofNodeUpdaterCallback
               const std::vector<Node>& args,
               CDProof* cdp,
               bool& continueUpdate) override;
+  /**
+   * This method is used to add some last steps to a
+   * proof when this is necessary. The final step should always be printed as
+   * (cl).
```
HanielB, line 56:
```suggestion
   * This method is used to add some last steps to a proof when this is
   * necessary. The final step should always be printed as (cl). However:
```

src/proof/alethe/alethe_post_processor.h
```diff
@@ -50,6 +50,24 @@ class AletheProofPostprocessCallback : public ProofNodeUpdaterCallback
...
+   *
+   * 1. If the last step of a proof is reached (which is false) it is printed as
+   * (cl false).
+   * 2. If one of the assumptions is false it is printed as false.
+   *
+   * Thus, an additional resolution step with (cl (not true)) has to be added to
+   * transfer (cl false) or false into (cl).
```
HanielB, line 63:
```suggestion
   * transform (cl false) or false into (cl).
```

src/proof/alethe/alethe_post_processor.cpp
```diff
@@ -1488,6 +1488,102 @@ bool AletheProofPostprocessCallback::update(Node res,
...
+{
+  NodeManager* nm = NodeManager::currentNM();
+
+  if (
+      // If the last proof rule was not translated yet
+      (id == PfRule::ALETHE_RULE) &&
+      // This case can only occur if the last step is an assumption
+      ((args[2].end() - args[2].begin()) > 1) &&
+      // If the proof node has result (false) additional steps have to be added.
+      (args[2][1].toString() != nm->mkConst(false).toString()))
```
HanielB, line 1523: Since you're using the node for `false` more than once, create it in the beginning of the method and use that.

Also when you caompare nodes you don't need to convert them to strings.

src/proof/alethe/alethe_post_processor.cpp
```diff
@@ -1488,6 +1488,102 @@ bool AletheProofPostprocessCallback::update(Node res,
...
+    const std::vector<Node>& args,
+    CDProof* cdp)
+{
+  NodeManager* nm = NodeManager::currentNM();
+
+  if (
+      // If the last proof rule was not translated yet
+      (id == PfRule::ALETHE_RULE) &&
+      // This case can only occur if the last step is an assumption
+      ((args[2].end() - args[2].begin()) > 1) &&
```
HanielB, line 1521: here you're trying to compare that the number of children of `args[2]` is bigger than 1?

src/proof/alethe/alethe_post_processor.cpp
```diff
@@ -1488,6 +1488,102 @@ bool AletheProofPostprocessCallback::update(Node res,
...
+
+  // remove attribute for outermost scope
+  if (id != PfRule::ALETHE_RULE)
+  {
+    std::vector<Node> sanitized_args{
+        res,
+        res,
+        nm->mkConst<Rational>(CONST_RATIONAL,
+                              static_cast<unsigned>(AletheRule::ASSUME))};
+    for (auto arg : args)
```
HanielB, line 1536:
```suggestion
    for (const Node& arg : args)
```

src/proof/alethe/alethe_post_processor.cpp
```diff
@@ -1488,6 +1488,102 @@ bool AletheProofPostprocessCallback::update(Node res,
...
+    {
+      sanitized_args.push_back(d_anc.convert(arg));
+    }
+    return cdp->addStep(res,
+                        PfRule::ALETHE_RULE,
+                        children,
+                        sanitized_args,
+                        true,
+                        CDPOverwrite::ALWAYS);
```
HanielB, line 1545: remove these last two arguments, IIRC you don't need them

HanielB (changes requested):

HanielB (approved):

PR #7711, opened 2021-11-30
Add rewrite for is_int pi

ajreynol:
Fixes https://github.com/cvc5/cvc5-projects/issues/365.

Fixes https://github.com/cvc5/cvc5-projects/issues/356.


src/theory/arith/arith_rewriter.cpp
```diff
@@ -847,6 +819,37 @@ RewriteResponse ArithRewriter::rewriteIntsDivMod(TNode t, bool pre)
   return RewriteResponse(REWRITE_DONE, t);
 }
 
+RewriteResponse ArithRewriter::rewriteExtIntegerOp(TNode t, bool pre)
```
nafur, line 822: What is the `pre` argument for?

ajreynol: Its not really necessary, I was following style of other methods but I've deleted the argument now.

nafur (approved):
LGTM.

PR #7712, opened 2021-11-30
Proper check for first-class types in datatype subfields

ajreynol:
Fixes two issues with our check for datatype subfields (isFunctionLike -> isFirstClass): functions should be allowed, RegLan should not.

Fixes https://github.com/cvc5/cvc5-projects/issues/368.  That issue now throws an error:
```
(error "Parse Error: proj-368.smt2:3.39: cannot use fields in datatypes that are not first class types

  (declare-datatype x ((_x (x8 RegLan))))
                                         ^
")
```
This adds a regression showing we allow functions as subfields.


HanielB (approved):
LGTM

PR #7716, opened 2021-11-30
Fixes for sygus-rr-synth-input

ajreynol:
Includes aborting when no non-constant subterms exist, a spurious assertion failure, and bad variable names.

Fixes https://github.com/cvc5/cvc5-projects/issues/351

Fixes https://github.com/cvc5/cvc5-projects/issues/353


src/preprocessing/passes/synth_rew_rules.cpp
```diff
@@ -173,7 +173,7 @@ PreprocessingPassResult SynthRewRulesPass::applyInternal(
       std::stringstream ssv;
       if (varCounter < 26)
       {
-        ssv << static_cast<char>(varCounter + 61);
+        ssv << static_cast<char>(varCounter + 65);
```
mpreiner, line 176: Just for documentation reasons it would be probably better to use `'A'` here.

mpreiner (approved):

PR #7717, opened 2021-11-30
Extend docs example extension

nafur:
This PR extends the custom sphinx extension for examples. It now allows for simple patterns in the file names and matches the file types using arbitrary regular expressions instead of just looking at the file extensions. This is necessary to integrate examples from the z3pycompat API: the examples live at a nontrivial place (in the build folder), which we inject via the file name patterns; we will have two separate examples which both end in `.py` but can be distinguished via the pattern used in the beginning.


alex-ozdemir (approved):
Looks good. Thanks!

PR #7718, opened 2021-11-30
add bag.fold operator

mudathirmahgoub:


src/theory/bags/theory_bags.cpp
```diff
@@ -152,6 +156,114 @@ TrustNode TheoryBags::expandCardOperator(TNode n,
...
+};
+typedef expr::Attribute<IndexVarAttributeId, Node> IndexVarAttribute;
+
+TrustNode TheoryBags::expandFoldOperator(TNode node,
+                                         std::vector<SkolemLemma>& lems)
+{
+  Assert(node.getKind() == BAG_FOLD);
+  if (d_env.getLogicInfo().isHigherOrder())
+  {
+    NodeManager* nm = NodeManager::currentNM();
```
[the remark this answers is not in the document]
mudathirmahgoub: Added file `bag_reduction.h,cpp` for class `BagReduction`. For now it only reduces the fold operator. Other operators (bag.choose, bag.card, bag.map) would be refactored in future PR.

src/theory/bags/bag_reduction.h
```diff
@@ -0,0 +1,78 @@
...
+   *   n: Int is the cardinality of bag A
+   *   uf:Int -> T1 is an uninterpreted function that represents elements of A
+   *   combine: Int -> T2 is an uninterpreted function
+   *   unionDisjoint: Int -> (Bag T1) is an uninterpreted function
+   */
+  Node reduceFoldOperator(Node node, std::vector<Node>& asserts);
+
+ private:
+  /** The associated environment. */
+  Env& d_env;
```
ajreynol, line 71: As a style point here, you should make this class inherit from EnvObj, and not have an explicit reference d_env.

ajreynol (changes requested):
One minor request.

src/theory/bags/bag_reduction.h
mudathirmahgoub: Done.

ajreynol (approved):

PR #7719, opened 2021-11-30
Always run update-pr

nafur:
This PR removes the event condition on the `update-pr` ci job. Right now, it is not run although it should.
It also fixes the cancel job (see https://github.com/styfle/cancel-workflow-action#advanced-pull-requests-from-forks)


4tXJ7f (approved):
LGTM

PR #7721, opened 2021-11-30
Add bitwise option to IntBlaster

makaimann:
This PR is part of a series of PRs by @yoni206  to add full int-blasting support to the master branch of `cvc5`. This adds support for the bitwise encoding to `IntBlaster`. Future PRs would unify the `bv2int` and `IntBlaster` helper functions.


src/options/smt_options.toml
```diff
@@ -395,6 +395,9 @@ name   = "SMT Layer"
 [[option.mode.BV]]
   name = "bv"
   help = "Translate bvand back to bit-vectors"
+[[option.mode.BITWISE]]
```
yoni206, line 398: On the one hand, we need this option to merge the new code to `int_blaster.cpp`. On the other hand, what is actually running right now is `bv_to_int.cpp`, which does not acknowledge this option. I suggest, for now, adding a check to `bv_to_int.cpp` that throws a meaningful error to the user (`bitwise option is not supported currently`, or something), in case someone (or some testing tool) tries this option.

test/unit/theory/theory_bv_int_blaster_white.cpp
```diff
@@ -280,5 +280,42 @@ TEST_F(TestTheoryWhiteBvIntblaster, intblaster_with_children)
...
+  // if original is BV, result should be Int.
+  // Otherwise, they should have the same type.
+  Node original;
+  Node result;
+
+  // bvand
+  original = d_nodeManager->mkNode(BITVECTOR_AND, v1, v2);
+  result = intBlaster.translateWithChildren(original, {i1, i2}, lemmas);
+  // should have kind skolem, would use bitwise comparisons to refine
+  ASSERT_TRUE(result.getKind() == kind::SKOLEM);
```
yoni206, line 317: I wonder whether it makes sense in this case to check that something was added to `skolems` and `lemmas`. What do you think?

makaimann: Great idea! Done with: 93f5a8eb604e7e3173178d37728594f56b00c598

src/options/smt_options.toml
makaimann: Good point! Added with 31c246889261b9fc6c862e1131acc248b523c641. This made me realize that there are now two `BITWISE` options: one for `solve-bv-as-int` and one for `iand-mode`. Did I do this correctly? In this PR do I actually need the `BITWISE` option for `solve-bv-as-int` (which I just added), or should I just be checking the `wand-mode` (which already has a [bitwise option](https://github.com/cvc5/cvc5/blob/1339d73750e15422a5c2085c4819c15b77bededd/src/options/smt_options.toml#L421)). Sorry for the dumb questions! Still paging this back in.

yoni206: 31c2468 looks great, thanks.

You are correct in everything you pointed out:
1. `solve-bv-as-int=bitwise` === eager-bitwise
2. `solve-bv-as-int=iand --iand-mode=bitwise` === lazy-bitwise
3. The code is relatively modular, and so the lazy bitwise option (bitwise in iand mode) is already tested: https://github.com/cvc5/cvc5/blob/70997d0e3ebf2027279373d9594c66119f3fa656/test/regress/regress3/nl/iand-native-1.smt2#L3 (notice that the test was added by you :) https://github.com/cvc5/cvc5/pull/5328). True, we still don't have a unit test for the iand solver, but I think that this is out of scope for the current PR.

So what you implemented and tested in this PR is eager-bitwise.

makaimann:
Good ideas! I added both in 5819b839dada5e50756c1ef3f3dea86367ca154e. In fact, number 2 revealed a bug -- I wasn't adding to `skolems`! So I just passed it through the functions as needed and added to it. It would be good if you could double check that for me. Thanks for the suggestions!

yoni206:
Looking more deeply into the purpose of `skolems`, I think that my original suggestion was flawed. Sorry about that. `skolems` is used to store the connection between a symbol the original formula to a term in the new formula. For example, a BV variable `x` is mapped to `(nat2bv k xx)` where `k` is the bit-width of `x` and `xx` is the integer variable that we introduced for `x`.

In the particular case of this PR, `original` is a `bvand` term. This does not need to be tracked. If we really wanted to, we could maybe map `(bvand x y)` to something that involves `iand` or the new skolem that we introduce. But this is not needed because the semantics of `iand` is fixed in the solver, and the semantics of the new skolem is fixed from the lemmas that we add. 

So my suggestion is to revert the addition of a skolem in this case, and in the test, just check that a lemma is tested. Please let me know if you think otherwise.

makaimann:
Got it, makes sense to me! I reverted that last commit and just checked that lemmas was added to.

yoni206 (approved):
Thanks for all the changes!

PR #7724, opened 2021-12-01
Alethe: Add finalize function to insert missing OR steps

Lachnitt:
A post visit that adds an OR step is needed in case a premise is printed as a singleton (cl (or F1 ... Fn)) but used as a clause (cl F1 ... Fn).


src/proof/alethe/alethe_post_processor.cpp
```diff
@@ -86,6 +87,14 @@ bool AletheProofPostprocessCallback::update(Node res,
     // node by using an extra pair of parenthesis, i.e. ((or b c)) is the proof
     // node printed as (cl (or b c)).
     //
+    // Adding an OR node to a premises will take place in the finalize function
+    // where in the case that a step is printed as (cl (or F1 ... Fn)) but used
+    // as (cl F1 ... Fn) an OR step is added to transform it to this very thing.
+    // This is necessary for rules that work on clauses, i.e. RESOLUTION,
+    // CHAIN_RESOLUTION, REORDERING and FACTORING. Thus, for all these rules the
+    // reader can assume that the premise was printed correctly since if not it
+    // is after the pass of the proof node in finalize.
```
HanielB, line 96:
```suggestion
    // CHAIN_RESOLUTION, REORDERING and FACTORING.
```
I think it's more clear if we omit the last sentence.

src/proof/alethe/alethe_post_processor.cpp
```diff
@@ -1488,6 +1397,180 @@ bool AletheProofPostprocessCallback::update(Node res,
...
+  {
+    // In the case of a resolution rule that step might originally have been a
+    // cvc5 RESOLUTION or CHAIN_RESOLUTION step. In these cases it is possible
+    // that one of the children was printed as (cl (or F1 ... Fn)) but used as
+    // (cl F1 ... Fn). However, since the information about the pivot of the
+    // resolution step the child is used in is provided, it is always possible
+    // to figure out if an additional OR step is necessary.
+    case AletheRule::RESOLUTION:
+    {
+      if (args.size() < 4)
```
HanielB, line 1426:
```suggestion
      // This means it's a resolution step generated during the processing of
      // other rules, so the confusion we are looking for does not happen.
      if (args.size() < 4)
```

HanielB (changes requested):
Looks good, just some requests about documentation

HanielB (approved):

PR #7726, opened 2021-12-01
[proofs] Make LazyCDProofChain extend CDProof

HanielB:
This commit makes LazyCDProofChain behave more similarly to LazyCDProof, in that it is an object that not only builds a proof to a fact based on lazy steps but also on regular steps that may already have for it.

The motivation for this commit is the upcoming handling of context-dependent proofs in incremental mode, which benefits from a more uniform behavior between CDProof, LazyCDProof and LazyCDProofChain (with the latter two both extending CDProof).


ajreynol (changes requested):
I'm wondering how facts that are given lazy steps in LazyCDProofChain *and* concrete steps in CDProof.  It seems that the steps in the LazyCDProofChain take priority.  Is this the same policy as LazyCDProof?  This should be documented.

ajreynol:
I think this is fine for now.

The behavior of this class can still be somewhat strange, where you'd have recursive alternations of concrete proof skeletons / proofs from generators.

Regardless, it seems that this PR doesn't make the potential behaviors of this class that much more complex.

ajreynol (approved):

PR #7728, opened 2021-12-01
Add explicit 64bit getters for Integer class

nafur:
This PR addresses a few issues on our 32bit builds. Most importantly, it adds `int64_t Integer::getSigned64()` and `uint64_t Integer::getUnsigned64()`.


src/util/integer_gmp_imp.cpp
```diff
@@ -402,28 +430,61 @@ unsigned int Integer::getUnsignedInt() const
...
+    return getLong();
+  }
+  else
+  {
+    if (mpz_fits_slong_p(d_value.get_mpz_t()) != 0)
+    {
+      return getLong();
+    }
+    // ensure there isn't overflow
+    CheckArgument(mpz_sizeinbase(d_value.get_mpz_t(), 2) < 64,
```
mpreiner, line 464: Add a comment here why it is < 64 in the signed case (since `mpz_sizeinbase` ignores the sign bit).

mpreiner (approved):

PR #7729, opened 2021-12-01
Check docs for broken links before uploading

nafur:
We currently have a few broken links in our documentation. This PR make our CI check for broken links before uploading it. Note that it will not fail any job, so we still need to check the output occasionally for now.


nafur:
See https://github.com/nafur/cvc5/runs/4388868282?check_suite_focus=true#step:6:26

mpreiner (approved):

PR #7730, opened 2021-12-01
Add documentation for QuickStart.java

mudathirmahgoub:


nafur (commented):
Please follow the style of the other quickstarts and use `literal-include` instead of `code-block`.

mudathirmahgoub:
> Please follow the style of the other quickstarts and use `literal-include` instead of `code-block`.

I have a strong opinion against this.  `literal-include` is only reasonable if line numbers do not change, which is rare. 
It does not save any effort if they change, and is even worse when the change happens at the beginning of the file, where you need to update all blocks, compared to only one place with `code-block`. 
It is not flexible if you want to diverge from the original code as I did in this [line](https://github.com/cvc5/cvc5/pull/7730/files#diff-6e9db0b83ba3947096189ca5e6c4e5eff8bc0df1fb13abf2fe329296c00b490fR10).

nafur:
> > Please follow the style of the other quickstarts and use `literal-include` instead of `code-block`.
> 
> I have a strong opinion against this. `literal-include` is only reasonable if line numbers do not change, which is rare. It does not save any effort if they change, and is even worse when the change happens at the beginning of the file, where you need to update all blocks, compared to only one place with `code-block`. It is not flexible if you want to diverge from the original code as I did in this [line](https://github.com/cvc5/cvc5/pull/7730/files#diff-6e9db0b83ba3947096189ca5e6c4e5eff8bc0df1fb13abf2fe329296c00b490fR10).

I acknowledge your opinion, and you may very well include **additional** code blocks (like the one you mentioned). I agree that keeping the line numbers in sync can be an issue. However, keeping the code itself in sync is an issue as well.
I find both variants bad for different reasons.

At the end, though, we concluded that using line numbers is safer: outdated line numbers usually result in code that does obviously not match the description, while out-of-date code that was copied may have more subtle issues due to refactorings (which will never show up anywhere, until some user tries to copy code from the quickstart). I just did such a refactoring concerning the Kinds in the python API. Also, such an error is more likely to be "transparent" to a user (i.e. the user recognizes it as a simple error that he can workaround himself).

aniemetz:
I fully agree with @nafur. We have discussed this in the past and concluded that including code via line numbers is the way to go for reasons @nafur mentioned.

mudathirmahgoub:
Since it has already been decided, I updated the file to use line numbers. 
I guess we have to live with some aesthetics issues like the first block in the image below which results from the indented lines in the original file, unlike the second code block which is displayed as expected. 

image

4tXJ7f:
For the quickstart examples, maybe we could also use a [literate programming extension](https://sphinx-litprog.readthedocs.io/en/stable/)?

docs/api/java/quickstart.rst
```diff
@@ -0,0 +1,176 @@
...
+
+We will ask the solver to produce models and unsat cores in the following,
+and for this we have to enable the following options.
+
+.. literalinclude:: ../../../examples/api/java/QuickStart.java
+     :language: java
+     :lines: 32-33
+
+Next we set the logic.
+The simplest way to set a logic for the solver is to choose "ALL".
```
4tXJ7f, line 21:
```suggestion
The simplest way to set a logic for the solver is to choose ``"ALL"``.
```

docs/api/java/quickstart.rst
```diff
@@ -0,0 +1,176 @@
+Quickstart Guide
+================
+
+First, create a cvc5 `Solver <io/github/cvc5/api/Solver.html>`_ instance using try with resources:
```
4tXJ7f, line 4: Nit: wrap line

docs/api/java/quickstart.rst
```diff
@@ -0,0 +1,176 @@
...
+
+.. code-block:: java
+
+     try (Solver solver = new Solver())
+     {
+       /** write your code here */
+     }
+
+We will ask the solver to produce models and unsat cores in the following,
+and for this we have to enable the following options.
```
4tXJ7f, line 14:
```suggestion
To produce models and unsat cores, we have to enable the following options.
```

(I know that this is just copied from the other guides, but the wording is a bit verbose)

docs/api/java/quickstart.rst
```diff
@@ -0,0 +1,176 @@
...
+This enables all logics in the solver.
+Alternatively, ``"QF_ALL"`` enables all logics without quantifiers.
+To optimize the solver's behavior for a more specific logic,
+use the logic name, e.g. ``"QF_BV"`` or ``"QF_AUFBV"``.
+
+.. literalinclude:: ../../../examples/api/java/QuickStart.java
+     :language: java
+     :lines: 42
+
+In the following, we will define constraints of reals and integers.
```
4tXJ7f, line 31:
```suggestion
In the following, we will define real and integer constraints.
```

docs/api/java/quickstart.rst
```diff
@@ -0,0 +1,176 @@
...
+For this, we first query the solver for the corresponding sorts.
+
+.. literalinclude:: ../../../examples/api/java/QuickStart.java
+     :language: java
+     :lines: 46-47
+
+Now, we create two constants ``x`` and ``y`` of sort ``Real``,
+and two constants ``a`` and ``b`` of sort ``Integer``.
+Notice that these are *symbolic* constants, but not actual values.
```
4tXJ7f, line 40:
```suggestion
Notice that these are *symbolic* constants, not actual values.
```

docs/api/java/quickstart.rst
```diff
@@ -0,0 +1,176 @@
...
+only this time over integer variables ``a`` and ``b``.
+For this, we first reset the assertions added to the solver.
+
+.. literalinclude:: ../../../examples/api/java/QuickStart.java
+     :language: java
+     :lines: 140
+
+Next, we assert the same assertions as above, but with integers.
+This time, we inline the construction of terms
+to the assertion command.
```
4tXJ7f, line 130:
```suggestion
in the assertion command.
```

4tXJ7f (changes requested):
Some minor requests and some suggestions.

docs/api/java/quickstart.rst
mudathirmahgoub: Done

docs/api/java/java.rst
```diff
@@ -0,0 +1,78 @@
+Java API Documentation
+======================
+
+The Java API for cvc5 mostly mirrors the :doc:`C++ API <../cpp/cpp>` and supports operator
+overloading, iterators, and exceptions.
+There are a few differences from the C++ API, such as using arbitrary-precision integer pairs,
+specifically, pairs of Java BigInteger objects, to represent rational numbers.
```
4tXJ7f, line 7:
```suggestion
specifically, pairs of Java `BigInteger` objects, to represent rational numbers.
```

docs/api/java/java.rst
```diff
@@ -0,0 +1,78 @@
...
+       value for x: 1/6
+       value for y: 1/6
+       expected: unsat
+       result: unsat
+       unsat core size: 3
+       unsat core:
+       (< 0 a)
+       (< 0 b)
+       (< (+ a b) 1)
+
```
nafur, line 51: The quickstart does not show up in the menu. I'd suggest to add a hidden toc here:
```suggestion

.. toctree::
    :hidden:

    quickstart
    API documentation <index>
```

nafur:
> I guess we have to live with some aesthetics issues like the first block in the image below which results from the indented lines in the original file, unlike the second code block which is displayed as expected.

The `literalinclude` directive should support the `:dedent:` option (https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-option-code-block-dedent). I see that this is more of an issue for java, as all relevant code is indented three times (class, method, try) while both python and c++ code is only indented once.

mudathirmahgoub:
> > I guess we have to live with some aesthetics issues like the first block in the image below which results from the indented lines in the original file, unlike the second code block which is displayed as expected.
> 
> The `literalinclude` directive should support the `:dedent:` option (https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-option-code-block-dedent). I see that this is more of an issue for java, as all relevant code is indented three times (class, method, try) while both python and c++ code is only indented once.

I already tried this option yesterday, it did not work with `literalinclude`.

nafur:
> I already tried this option yesterday, it did not work with `literalinclude`.

Which sphinx version do you use? I found that the sphinx that comes from the ubuntu repositories is pretty old (in my case 3.5.4 vs. 4.3.0 from pip). It works on my machine with 4.3.0 and the following syntax (I tested on the python quickstart):
```
.. literalinclude:: ../../../../examples/api/python/quickstart.py
     :language: python
     :lines: 22
     :dedent: 2
```

mudathirmahgoub:
> 

Mine is installed with pip, 
```bash
pip install sphinx
Requirement already satisfied: sphinx in /home/mudathir/.local/lib/python3.8/site-packages (4.2.0)
```
This change does not work
```rst
.. literalinclude:: ../../../examples/api/java/QuickStart.java
     :language: java
     :lines: 160-166
     :dedent: 1
     ```

**Comment by mudathirmahgoub** [comment]
I used the wrong value `:dedent:`.  6 seems to fix it. 
So not just you need to worry about line numbers, but you also need to figure out how many characters of the indentation. 
I am not a fan of `.. literalinclude::`.

File: docs/api/java/java.rst [hunk 9]
```diff
@@ -0,0 +1,83 @@
...
+There are a few differences from the C++ API, such as using arbitrary-precision integer pairs,
+specifically, pairs of Java `BigInteger` objects, to represent rational numbers.
+The :doc:`quickstart guide <quickstart>` gives a short introduction,
+and more examples can be found `here <../../examples/examples.html>`_.
+
+
+For most applications, the `Solver <io/github/cvc5/api/Solver.html>`_ class is the main entry point to cvc5.
+The class hierarchy of `cvc5 package <io/github/cvc5/api/package-summary.html>`_
+provides more details on the individual classes.
+
```
nafur, line 15:
```suggestion

.. toctree::
    :hidden:

    quickstart
    API documentation <index>

```

docs/api/java/java.rst
```diff
@@ -0,0 +1,83 @@
...
+       (< 0 a)
+       (< 0 b)
+       (< (+ a b) 1)
+
+.. toctree::
+    :hidden:
+
+    quickstart
+    API documentation <index>
+
```
nafur, line 57:
```suggestion
```

nafur (commented):
Ah I forgot about how sphinx integrates tocs into the overall tree. This moves it out of the "Building cvc5 Java API" section...

docs/api/java/quickstart.rst
```diff
@@ -0,0 +1,191 @@
...
+
+Example
+-------
+
+| The SMT-LIB input for this example can be found at `examples/api/smtlib/quickstart.smt2 <https://github.com/cvc5/cvc5/blob/master/examples/api/smtlib/quickstart.smt2>`_.
+| The source code for this example can be found at `examples/api/java/QuickStart.java <https://github.com/cvc5/cvc5/blob/master/examples/api/java/QuickStart.java>`_.
+
+.. api-examples::
+    <examples>/api/cpp/quickstart.cpp
+    <examples>/api/java/QuickStart.java
```
nafur, line 189: Show the java code by default.
```suggestion
    <examples>/api/java/QuickStart.java
    <examples>/api/cpp/quickstart.cpp
```

nafur (approved):
Just a final suggestion. Thanks for all the changes!

4tXJ7f (approved):
LGTM, thanks for the updates.

PR #7731, opened 2021-12-02
Remove void as possible option type

nafur:
This PR removes the possibility to have `void` options. While there are (very) few options that may legitimately be considered `void` (they do not actually hold a value, but merely trigger a handler or predicate), the type being `void` introduces a number of special cases in the options setup. Their only real benefit was to avoid having data members that would never be used.
As it turns out, though, the same can be achieved by not specifying a `name`, which already supported as well.


src/options/base_options.toml
```diff
@@ -63,16 +63,18 @@ name   = "Base"
   category   = "common"
   short      = "v"
   long       = "verbose"
-  type       = "void"
-  handler    = "increaseVerbosity"
+  type       = "bool"
```
4tXJ7f, line 66: If we declare those options to be of `type` `bool`, don't we also start accepting `--no-verbose`?

nafur: note the `alternate = false` below. This prevents exactly that.

4tXJ7f: Ah, great!

4tXJ7f (approved):
LGTM, thanks for the explanation.

PR #7735, opened 2021-12-02
More robust fix for 32bit issues.

nafur:
This PR implements a more robust fix for the 32bit issues with GMP. Relying on `mpz_sizeinbase` proves to be very tricky for the corner cases. This PR instead does a simple comparison against the min and max values of the respective types.

### DIFF
--- a/src/util/integer_gmp_imp.cpp
+++ b/src/util/integer_gmp_imp.cpp
@@ -486,11 +486,8 @@ uint64_t Integer::getUnsigned64() const
     }
     try
     {
-      if (sgn() < 0)
-      {
-        CheckArgument(
-            false, this, "Overflow detected in Integer::getUnsigned64().");
-      }
+      CheckArgument(
+          sgn() >= 0, this, "Overflow detected in Integer::getUnsigned64().");
       return std::stoull(toString());
     }
     catch (const std::exception& e)
--- a/src/util/integer_gmp_imp.cpp
+++ b/src/util/integer_gmp_imp.cpp
@@ -461,8 +461,10 @@ int64_t Integer::getSigned64() const
       return getLong();
     }
     // ensure there is no overflow.
-    // mpz_sizeinbase ignores the sign bit, thus at most 63 bits.
-    CheckArgument(mpz_sizeinbase(d_value.get_mpz_t(), 2) < 64,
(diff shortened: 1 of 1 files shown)


src/util/integer_gmp_imp.cpp
```diff
@@ -461,8 +461,10 @@ int64_t Integer::getSigned64() const
...
     // ensure there is no overflow.
-    // mpz_sizeinbase ignores the sign bit, thus at most 63 bits.
-    CheckArgument(mpz_sizeinbase(d_value.get_mpz_t(), 2) < 64,
+    CheckArgument(mpz_class(std::to_string(std::numeric_limits<int64_t>::min()))
+                          <= d_value
+                      && d_value <= mpz_class(std::to_string(
+                             std::numeric_limits<int64_t>::max())),
                   this,
                   "Overflow detected in Integer::getSigned64().");
     return std::stoll(toString());
```
mpreiner, line 470: Can't we rely on `std::stoll` here to throw an `std::out_of_range` exception instead of using this check?

src/util/integer_gmp_imp.cpp
```diff
@@ -480,10 +482,14 @@ uint64_t Integer::getUnsigned64() const
...
-                  this,
-                  "Overflow detected in Integer::getUnsigned64().");
+    // ensure there is no overflow
+    CheckArgument(
+        mpz_class(std::to_string(std::numeric_limits<uint64_t>::min()))
+                <= d_value
+            && d_value <= mpz_class(
+                   std::to_string(std::numeric_limits<uint64_t>::max())),
+        this,
+        "Overflow detected in Integer::getUnsigned64().");
```
mpreiner, line 492: ditto

src/util/integer_gmp_imp.cpp
```diff
@@ -480,12 +484,22 @@ uint64_t Integer::getUnsigned64() const
...
-                  this,
-                  "Overflow detected in Integer::getUnsigned64().");
-    return std::stoull(toString());
+    try
+    {
+      if (sgn() < 0)
+      {
+        CheckArgument(
+            false, this, "Overflow detected in Integer::getUnsigned64().");
+      }
```
mpreiner, line 493:
```suggestion
      CheckArgument(
          sgn() >= 0, this, "Overflow detected in Integer::getUnsigned64().");
```

mpreiner (approved):

PR #7736, opened 2021-12-03
Towards support for incremental sygus

ajreynol:
This makes several key changes to smt::SygusSolver in preparation for incremental mode.

The key idea is to use a subsolver that:
- may take multiple `check-sat` commands for an encoded synthesis conjecture
- does *not* push/pop
- is reconstructed when the SyGuS conjecture becomes stale.

This is motivated by 2 use cases of incremental SyGuS:
- Where constraints are push/pop, in which case we should start from scratch, since the SyGuS solver uses symmetry breaking techniques that become unsound when new constraints are added.
- Where multiple solutions are needed for the same set of constraints, in which the state of the subsolver should be preserved.

This functionality is still guarded by an exception.

A follow up PR will make several further changes to allow for the above use cases.


src/smt/sygus_solver.h
```diff
@@ -37,13 +38,27 @@ class SmtSolver;
...
+ * check satisfiability using a separate SolverEngine "subsolver".
  *
- * It also maintains a reference to a preprocessor for implementing
- * checkSynthSolution.
+ * This solver operates in two modes.
+ *
+ * If in incremental mode, then the "main" SolverEngine for SyGuS inputs only
+ * maintains a (user-context) dependent state of SyGuS assertions, as well as
+ * assertions corresponding to (recursive) function definitions. The subsolver
+ * that solves SyGuS conjectures may be called to checkSat multiple times,
```
HanielB, line 48: I'm a bit confused by this. The goal of multiple check-sats would be to allow getting multiple solutions? If so clarify this, because currently it's not clear what is the motivation to have this since any change in the conjecture triggers a destruction of the solver.

src/smt/sygus_solver.h
```diff
@@ -170,25 +182,63 @@ class SygusSolver : protected EnvObj
...
   /**
    * Whether we need to reconstruct the sygus conjecture.
+   *
+   * The sygus conjecture is stale if either:
+   * (1) no sygus conjecture has been added as an assertion to this SMT engine,
+   * (2) there is a sygus conjecture that has been added as an assertion
+   * internally to this SMT engine, and there have been further calls such that
+   * the asserted conjecture is no longer up-to-date.
+   *
+   * This method should be called when new sygus constraints are asserted and
```
HanielB, line 223: no longer a method

src/smt/sygus_solver.cpp
```diff
@@ -185,45 +192,76 @@ Result SygusSolver::checkSynth(Assertions& as)
...
+      d_subsolverCd = d_subsolver.get();
 
+      // also assert the internal SyGuS conjecture
+      d_subsolver->assertFormula(d_conj);
+    }
+  }
+  else
+  {
+    Assert(d_subsolver != nullptr);
+  }
```
HanielB, line 253: You could delete this and add `Assert(d_sygusConjectureStale || d_subsolver != nullptr);` before the `if`

src/smt/sygus_solver.cpp
```diff
@@ -233,39 +271,47 @@ Result SygusSolver::checkSynth(Assertions& as)
...
-bool SygusSolver::getSynthSolutions(std::map<Node, Node>& sol_map)
+bool SygusSolver::getSynthSolutions(std::map<Node, Node>& solMap)
 {
   Trace("smt") << "SygusSolver::getSynthSolutions" << std::endl;
-  std::map<Node, std::map<Node, Node>> sol_mapn;
+  if (usingSygusSubsolver())
+  {
+    // use the call to get the synth solutions from the subsolver
+    return d_subsolver->getSubsolverSynthSolutions(solMap);
+  }
```
HanielB, line 281: if you have this why all the upstream methods to get subsolver solutions?

src/smt/sygus_solver.cpp
```diff
@@ -279,100 +325,39 @@ void SygusSolver::checkSynthSolution(Assertions& as)
...
-    {
-      conjBody = conj[1].substitute(
-          fvars.begin(), fvars.end(), fsols.begin(), fsols.end());
-    }
-    else
-    {
-      conjBody = conj[1][1].substitute(
-          fvars.begin(), fvars.end(), fsols.begin(), fsols.end());
-
-      /* Skolemize property */
```
HanielB, line 362: why is this no longer necessary?

HanielB (dismissed):
A few requests, but mostly I was wondering if the overall setup wouldn't be simpler if we were to always use a subsolver, even when not in incremental?

src/smt/sygus_solver.cpp
ajreynol: It was never necessary to begin with, the subsolver can skolemize itself.

src/smt/sygus_solver.cpp
ajreynol: There will be code added to this block (e.g. to block the current solution).

ajreynol:
Actually, I originally did have it so that the subsolver is *always* called not just in incremental.  There is one major downside to this: it is not currently possible to access stats.  You will only get the (trivial) stats from the original solver, which is why (for now) I made it possible to have both, in case e.g. we wanted to analyze stats on non-incremental problems.

src/smt/sygus_solver.cpp
ajreynol: do you mean why are other methods using the `SolverEngine::getSubsolverSynthSolutions` interface?

HanielB: right, like the one in SMT engine

HanielB:
I see, had not anticipated that. :)

abdoo8080 (approved):
LGTM! Will the conflict with `--sygus-stream` be handled in a separate PR?

ajreynol:
Yes, there is still nothing enabled in this PR, the relationship with `sygus-stream` will be addressed in a followup PR.

src/smt/sygus_solver.cpp
ajreynol: `usingSygusSubsolver` is true if we use a sygus subsolver if we received `check-synth` calls.  The differences is that a subsolver may receive a synthesis conjecture and must return synth solutions after `check-sat` calls.

ajreynol: The documentation https://github.com/cvc5/cvc5/pull/7736/files#diff-f7beb6dbdfd17c4aa0317171df9c685a78d8e393e5a4fc0137ba87d1c3f890e5R567 makes this distinction.

PR #7737, opened 2021-12-03
Check constructor is used in APPLY_CONSTRUCTOR

ajreynol:
Fixes https://github.com/cvc5/cvc5-projects/issues/373.


aniemetz (commented):
Can you add the unit test (but changing it so that we expect this to throw a Cvc5ApiException)?

aniemetz (approved):

PR #7740, opened 2021-12-03
Eliminate more static calls to Rewriter::rewrite

ajreynol:


nafur (approved):
LGTM.
I assume that you did some runs to make sure that everything is still rewritten when necessary?

ajreynol:
The cases where Rewriter::rewrite was removed were places that should not impact anything. E.g. there were a few places where quantifiers rewriter was making a call to Rewriter::rewrite, whereas it should just simply return the term to be rewritten again.

Issue #375, opened 2017-08-19
Draft CC module (Bugzilla #155)


mpreiner:
**Imported from Bugzilla**
Reporter: *Morgan Deters*
Status: *RESOLVED*
Severity: *enhancement*
Assigned to: *Morgan Deters*
Component: *{Internals} other*
Milestone: *---*
Version: *master*
Platform: *All*
OS: *All*

On 2010-06-15 11:31:22 -0700, Morgan Deters wrote nothing.
On 2010-09-01 17:09:37 -0700, Morgan Deters wrote:
> Drafted and merged to trunk.  Closing this bug.

Issue #358, opened 2017-08-19
Failing regress1 test (Bugzilla #138)


mpreiner:
**Imported from Bugzilla**
Reporter: *Tim King*
Status: *RESOLVED*
Severity: *normal*
Assigned to: *Morgan Deters*
Component: *{Internals} Expression package*
Milestone: *---*
Version: *master*
Platform: *PC*
OS: *Linux*

On 2010-05-29 18:50:39 -0700, Tim King wrote:
> The regression test test/regress/regress1/fuzz\_2.smt is failing as of revision 520.
> 
> The reason is appears to be TNode deallocation. Nothing more is currently known.

On 2010-06-01 17:35:07 -0700, Tim King wrote:
> This was caused by something quite different than TNode deletion. Appearances can be deceiving.
> 
> The problem was that a variable was not getting registered to d\_arith. It was instead registered to d\_uf. The node that caused this problem has the form
>  (\= skolem (<\= v1 v1))
> This formula made it to Theory::get() as it is an equality.  Because of how the trap doors in theoryOf() work, the theory this was asserted to was d\_uf.  During the get() call, RegisteredAttr() was getting set, and subsequently get() did not register the variable with d\_arith.  This violated assumptions in d\_arith causing the crash that was observed.
> 
> The skolem was introduced during ite rewriting inappropriately because the check that the then branch was a boolean had a bug.
> 
> This should now be working as of revision 527.

PR #7743, opened 2021-12-03
Fix cvc5-projects issues  #358 and #375

mudathirmahgoub:


ajreynol (approved):
As discussed offline, this is not an ideal solution, but let's go ahead with it since it is the only solution that seems to work at the moment.

PR #7744, opened 2021-12-03
Use attribute for mkEmptyWord

ajreynol:
Word::mkEmptyWord still takes 2% runtime on average for some Amazon benchmarks, this adds an attribute which makes it neglibile.


ajreynol:
I'm closing this for now, as this only marginally improves runtime, I think the better way to go is to reduce our calls to this method @4tXJ7f

PR #7745, opened 2021-12-03
Alethe: Implementation of Process Function

Lachnitt:
Add call to updater and finalizer to process function of the AlethePostProcessor.


HanielB (approved):
LGTM

PR #7749, opened 2021-12-06
Fix memory leak in stream options

nafur:
This PR fixes a memory leak when invalid values are supplied to the stream options `err`, `in` or `out`. Whenever opening the stream would fail, we throw an exception and the created (though not properly opened) stream is leaked.
We now properly wrap the streams in unqiue_ptrs directly on creation.

This fixes the memory leak observed in our asan-enabled nightlies.


mpreiner (approved):

4tXJ7f (approved):
LGTM

PR #7751, opened 2021-12-06
Disable option regression for competition build

nafur:
This PR disables a regression that checks the didyoumean output for an invalid output on competition builds. On competition builds, we simply print "unknown" instead of shown any exceptions.


4tXJ7f (approved):
LGTM

PR #7752, opened 2021-12-06
Fix 32bit issue in sep_log_api test

nafur:
This PR uses the proper `t.getInt64Value()` getter instead of `std::stoll(t.toString())`, fixing an issue on 32bit platforms.


4tXJ7f (approved):
LGTM

PR #7754, opened 2021-12-07
Alethe: Removed Steps that are Double-Printed

Lachnitt:
This PR fixes that some steps are printed more than once. To this end, it deletes that the reference maps in which the steps are stored are constant and optimizes where the steps are added to the maps. It also stores ProofNodes instead of Nodes in the maps. 


src/proof/alethe/alethe_printer.cpp
```diff
@@ -74,20 +72,49 @@ std::string AletheProofPrinter::printInternal(
...
+  // If the current step is already printed return its id
+  auto it = steps.find(pfn);
+  if (it != steps.end())
+  {
+    Trace("alethe-printer")
+        << "... step is already printed " << it->second << " "
+        << pfn->getResult() << " " << arule << " / " << args << std::endl;
+    return it->second;
+  }
+
```
HanielB, line 113:
```suggestion

  // If the current step is already printed return its id
  auto it = steps.find(pfn);
  if (it != steps.end())
  {
    Trace("alethe-printer")
        << "... step is already printed " << it->second << " "
        << pfn->getResult() << " " << arule << " / " << args << std::endl;
    return it->second;
  }
  std::vector<std::string> current_assumptions;
  std::unordered_map<Node, std::string> assumptions_before_subproof =
      assumptions;
  std::unordered_map<std::shared_ptr<ProofNode>, std::string>
      steps_before_subproof = steps;

```

HanielB (changes requested):
Looks good, just minor comments

HanielB (approved):

PR #7755, opened 2021-12-07
Eliminate more static calls to Rewriter::rewrite

ajreynol:


src/theory/theory_model_builder.cpp
```diff
@@ -1278,7 +1278,8 @@ void TheoryEngineModelBuilder::assignFunction(TheoryModel* m, Node f)
   }
   std::stringstream ss;
   ss << "_arg_";
-  Node val = ufmt.getFunctionValue(ss.str().c_str(), condenseFuncValues);
+  Rewriter* r = condenseFuncValues ? d_env.getRewriter() : nullptr;
+  Node val = ufmt.getFunctionValue(ss.str().c_str(), r);
```
nafur, line 1282:
```suggestion
  Node val = ufmt.getFunctionValue(ss.str(), r);
```

nafur (approved):
LGTM.

PR #7758, opened 2021-12-07
Simpler versioning if release flag is set

nafur:
Following some discussion, this PR introduces a simpler versioning scheme if the `CVC5_IS_RELEASE` flag is set.
In our regular versioning scheme, the first commit after release `x.y.z` has the version `x.y.(z+1)-dev.1.abcdef0`. Some users however may want to apply some patches, and still have cvc5 show the release version.
With this PR, the version printed (if any commits are present since the tag) is `x.y.z-modified` if the `CVC5_IS_RELEASE` flag is still true. We always set it to false immediately after the tagged commit, so it does not change the versions printed by mainline cvc5.


4tXJ7f (approved):
LGTM

4tXJ7f (changes requested):
In the case where we have a release, we shouldn't increment the patch part, right?

nafur:
> In the case where we have a release, we shouldn't increment the patch part, right?

I admit that the code is a bit weird... note that we simply ignore `GIT_LAST_TAG` for releases, because it was easier to structure the code.

4tXJ7f (approved):
Oh right, missed that. LGTM then :)

PR #7759, opened 2021-12-07
Do not make SyGuS subsolver in unsat state after solving

ajreynol:
This makes subsolvers answer "unknown" instead of "unsat" when solving SyGuS inputs, by setting unknown when a SyGuS input is solved instead of asserting the negation of the input.

This will allow for multiple solutions for the same conjecture.


src/theory/quantifiers/sygus/synth_engine.h
```diff
@@ -52,6 +52,8 @@ class SynthEngine : public QuantifiersModule
   QEffort needsModel(Theory::Effort e) override;
   /* Call during quantifier engine's check */
   void check(Theory::Effort e, QEffort quant_e) override;
+  /** check complete for quantified formula */
```
HanielB, line 55: Clarify what that means

src/smt/sygus_solver.h
```diff
@@ -169,8 +169,13 @@ class SygusSolver : protected EnvObj
...
    * synthesized solutions, which is a quantifier-free formula, is
    * unsatisfiable. If not, then the found solutions are wrong.
+   *
+   * @param as The background assertions, which may include define-fun and
+   * define-fun-rec,
+   * @param sol_map Map from functions-to-synthesize to their solution (of the
+   * same type) for the asserted synthesis conjecture.
    */
-  void checkSynthSolution(Assertions& as);
+  void checkSynthSolution(Assertions& as, const std::map<Node, Node>& sol_map);
```
HanielB, line 178:
```suggestion
  void checkSynthSolution(Assertions& as, const std::map<Node, Node>& solMap);
```

src/smt/sygus_solver.cpp
```diff
@@ -261,11 +261,46 @@ Result SygusSolver::checkSynth(Assertions& as)
...
     r = d_smtSolver.checkSatisfiability(as, query, false);
   }
-  // Check that synthesis solutions satisfy the conjecture
-  if (options().smt.checkSynthSol
-      && r.asSatisfiabilityResult().isSat() == Result::UNSAT)
+  // The result returned by the above call is typically "unknown", which may
+  // or may not correspond to a state in which we solved the conjecture
+  // successfully. Instead we call getSynthSolutions below. If this returns
+  // true, then we were successful. In this case, we set the result to "unsat",
+  // since the synthesis conjecture was negated when assered to the subsolver.
```
HanielB, line 268:
```suggestion
  // since the synthesis conjecture was negated when asserted to the subsolver.
```

src/smt/sygus_solver.cpp
```diff
@@ -261,11 +261,46 @@ Result SygusSolver::checkSynth(Assertions& as)
...
+  // is *T-valid* (in the presence of defined recursive functions). In other
+  // words, a SyGuS query asks to prove that the conjecture is valid when
+  // witnessed by the given solution.
+  // (2) we do not want the solver to explicitly answer "unsat" by giving an
+  // unsatisfiable set of formulas to the underlying PropEngine, or otherwise
+  // we will not be able to ask for further solutions. This is critical for
+  // incremental solving where multiple solutions are returned for the same
+  // set of constraints. Thus, the internal SyGuS solver will mark unknown
+  // with IncompleteId::QUANTIFIERS_SYGUS_SOLVED. Furthermore, this id may be
+  // overwritten by other means of incompleteness, so we cannot rely on this
```
HanielB, line 284: So what is the ID for?

src/theory/quantifiers/sygus/synth_conjecture.cpp
```diff
@@ -558,7 +559,7 @@ bool SynthConjecture::doCheck()
     // We should set incomplete, since a "sat" answer should not be
     // interpreted as "infeasible", which would make a difference in the rare
     // case where e.g. we had a finite grammar and exhausted the grammar.
-    d_qim.setIncomplete(IncompleteId::QUANTIFIERS_SYGUS_NO_VERIFY);
+    // d_qim.setIncomplete(IncompleteId::QUANTIFIERS_SYGUS_NO_VERIFY);
```
HanielB, line 562: ?

HanielB (changes requested):

src/smt/sygus_solver.cpp
ajreynol: All forms of incompleteness must be marked with an id, its just that we can't rely on this being the only source of incompleteness.

abdoo8080: Why not maintain a list of ids then?

HanielB (approved):

src/smt/sygus_solver.cpp
ajreynol: We could consider doing this in a later PR, although it is still more reliable to just call `getSynthSolutions`.

PR #7768, opened 2021-12-08
Consider polarity in relevance manager

ajreynol:
This ensures we only consider things relevant if their polarity is compatible with current asserted value.

As a consequence, this makes our computation of `get-difficulty` more accurate, as well as making `Valuation::isRelevant` more precise (e.g. for non-linear).

Notice that this fixes a bug in PolarityTermContext as well.


src/expr/term_context.cpp
```diff
@@ -131,7 +131,7 @@ uint32_t PolarityTermContext::getValue(bool hasPol, bool pol)
 
 void PolarityTermContext::getFlags(uint32_t val, bool& hasPol, bool& pol)
 {
-  hasPol = val == 0;
+  hasPol = val != 0;
```
nafur, line 134: What consequences does this change have?

src/theory/relevance_manager.cpp
```diff
@@ -165,22 +166,24 @@ bool RelevanceManager::isBooleanConnective(TNode cur)
          || (k == EQUAL && cur[0].getType().isBoolean());
 }
 
-bool RelevanceManager::updateJustifyLastChild(TNode cur,
+bool RelevanceManager::updateJustifyLastChild(RlvPair cur,
```
nafur, line 169:
```suggestion
bool RelevanceManager::updateJustifyLastChild(const RlvPair& cur,
```

src/theory/relevance_manager.cpp
```diff
@@ -165,22 +166,24 @@ bool RelevanceManager::isBooleanConnective(TNode cur)
...
-  Assert(d_jcache.find(cur[index]) != d_jcache.end());
-  Kind k = cur.getKind();
+  Kind k = cur.first.getKind();
   // Lookup the last child's value in the overall cache, we may choose to
   // add this to childrenJustify if we return true.
-  int lastChildJustify = d_jcache[cur[index]];
+  RlvPair cp(cur.first[index],
+             d_ptctx.computeValue(cur.first, cur.second, index));
+  Assert(d_jcache.find(cp) != d_jcache.end());
+  int lastChildJustify = d_jcache[cp];
```
nafur, line 186:
```suggestion
  uint64_t lastChildJustify = d_jcache[cp];
```

src/expr/term_context.cpp
ajreynol: This corrects the value of `hasPol` for this method, RelevantDomain in quantifiers was the only current user of this class, which was used as a heuristic for enumerative instantiation. In summary, this means that enumerative instantiation (might) be slightly better after this commit. (Regardless of whether it is, the code should be corrected).

nafur (approved):
LGTM

PR #7770, opened 2021-12-08
Refactor bag solver

mudathirmahgoub:


src/theory/bags/solver_state.cpp
```diff
@@ -132,6 +132,20 @@ void SolverState::collectDisequalBagTerms()
...
 }
 
+void SolverState::registerCountSkolem(Node n, Node e, Node skolem)
+{
+  Node nRep = getRepresentative(n);
+  Node eRep = getRepresentative(e);
+  d_countSkolems[n][e] = skolem;
+}
+
+Node SolverState::getCountSkolem(Node n, Node e)
```
ajreynol, line 142: Can you make this a `const` method and ensure that the entry in the map exists?

mudathirmahgoub: replaced with `getElementCountPairs`

src/theory/bags/solver_state.h
```diff
@@ -58,20 +59,25 @@ class SolverState : public TheoryState
    * (assert (= 0 (bag.count x B)))
    * element x is associated with bag B, albeit x is definitely not in B.
    */
-  const std::set<Node>& getElements(Node B);
+  std::set<Node> getElements(Node B);
   /** initialize bag and count terms */
-  void initialize();
+  std::vector<Node> initialize();
```
ajreynol, line 64: Document the return value here.

src/theory/bags/solver_state.h
```diff
@@ -84,7 +90,7 @@ class SolverState : public TheoryState
   /** collection of bag representatives */
   std::set<Node> d_bags;
   /** bag -> associated elements */
-  std::map<Node, std::set<Node>> d_bagElements;
+  std::map<Node, std::vector<std::pair<Node, Node>>> d_bagElements;
```
ajreynol, line 93: Can you document the pair here?

ajreynol: Also, this seems to be a fairly important data structure, can you document a bit more the lifetime of this?  Is this cleared on full effort? When is it valid?

ajreynol (changes requested):

src/theory/bags/solver_state.h
mudathirmahgoub: Done.

ajreynol (approved):

PR #7777, opened 2021-12-09
test: Remove CDList memory limit test.

mpreiner:
This test is of little usefulness and has issues with different platforms (mac) and architectures (arm64).


nafur (approved):
LGTM.

PR #7779, opened 2021-12-09
[proofs] Alethe: fixing formatting and adding missing comment

HanielB:


src/proof/alethe/alethe_post_processor.cpp
```diff
@@ -204,7 +204,7 @@ bool AletheProofPostprocessCallback::update(Node res,
     // VP8: (cl (=> (and F1 ... Fn))) (cl (not (=> (and F1 ... Fn) false))
```
Lachnitt, line 204: This should be VP8: (cl (not (=> (and F1 ... Fn) false))

HanielB: Actually I think it's `// VP8: (cl (=> (and F1 ... Fn) false))`?

Lachnitt: Ah yes of course!

Lachnitt:
LGTM

ajreynol (approved):

PR #7782, opened 2021-12-09
Add a random Sygus enumerator.

abdoo8080:
This PR adds a Sygus enumerator that generates random terms from a grammar. The size of the terms generated by the enumerator approximates a geometric distribution and can be configured with an option.


src/theory/quantifiers/sygus/sygus_random_enumerator.h
```diff
@@ -0,0 +1,140 @@
...
+   * the given sygus term.
+   */
+  Node getMin(Node n);
+
+  /** Pointer to term database sygus. */
+  TermDbSygus* d_tds;
+  /** The type to enumerate. */
+  TypeNode d_tn;
+  /** Random number generator. */
+  std::mt19937_64 d_gen;
```
ajreynol, line 116: I think we should use standard Random utilities for this.  https://github.com/cvc5/cvc5/blob/master/src/util/random.h

src/theory/quantifiers/sygus/sygus_random_enumerator.h
```diff
@@ -0,0 +1,140 @@
...
+  /** Bernoulli distribution. */
+  std::bernoulli_distribution d_bernoulli;
```
ajreynol, line 118: Not sure if this and the utility below this are available in the standard utilites.  Maybe @aniemetz can comment.

ajreynol (changes requested):

src/theory/quantifiers/sygus/sygus_random_enumerator.cpp
```diff
@@ -0,0 +1,199 @@
...
+  d_cache.insert(bn);
+  d_currTerm = n;
+  return true;
+}
+
+Node SygusRandomEnumerator::incrementH()
+{
+  NodeManager* nm = NodeManager::currentNM();
+  SkolemManager* sm = NodeManager::currentNM()->getSkolemManager();
+  Random& rand = Random::getRandom();
```
ajreynol, line 77: This is a pretty confusing name, `rand()` is builtin C++ method, you are also using `rand()` below, which is not clear which is getting called.

abdoo8080: I named it `rand` at first because I thought no one cares about implementation details. Though, it does bring some maintenance issues now that I think about it. Changed to `rnd`.

ajreynol (approved):
LGTM

PR #7784, opened 2021-12-09
Add switches to toggle eager and inclusion solvers

4tXJ7f:


src/theory/strings/sequences_rewriter.h
```diff
@@ -309,6 +311,8 @@ class SequencesRewriter : public TheoryRewriter
   ArithEntail d_arithEntail;
   /** Instance of the entailment checker for strings. */
   StringsEntail d_stringsEntail;
+  /** Whether to use regular expression inclusion tests. */
+  bool d_useRegexpInclusion;
```
ajreynol, line 315: I'd prefer if this wasn't an option on the rewriter, and instead guards the portion of the RegExpSolver that does the inclusion tests, isn't this easier to do?

ajreynol (changes requested):

src/theory/strings/sequences_rewriter.h
4tXJ7f: Discussed offline and addressed.

ajreynol (approved):

PR #7785, opened 2021-12-09
Allow SyGuS subsolver to be reused in incremental mode

ajreynol:
This allows SyGuS subsolvers to be called multiple times to produce solutions for a given set of SyGuS constraints using a new command `check-synth-next`.

By default, the SyGuS subsolver will generate a *new* solution for each successive check-synth.

This simplifies the internal SyGuS solver in several ways for this purpose.  It also ensures that solutions are cached; current master may recompute solutions in the same state more than once if asked, which is no longer the case.

This completes our support for incremental SyGuS.


src/theory/quantifiers/sygus/synth_conjecture.cpp
```diff
@@ -93,8 +94,22 @@ SynthConjecture::~SynthConjecture() {}
 
 void SynthConjecture::presolve()
 {
-  // we don't have a solution yet
-  d_hasSolution = false;
+  // If we previously had a solution, block it. Notice that
+  // excludeCurrentSolution in the block below ensures we implement a
+  // policy where a *new* solution is generated for check-synth in the set of
```
HanielB, line 99:
```suggestion
  // policy where a *new* solution is generated for check-synth if the set of
```

HanielB (commented):
Is there any way we can guard for being running a new `check-synth` command but no different solution existing? So that we could mark that there is no *new* solution rather than no solution?

ajreynol:
Hm maybe we should just do `check-synth-next`....

src/smt/solver_engine.cpp
```diff
@@ -934,11 +934,17 @@ void SolverEngine::assertSygusInvConstraint(Node inv,
...
-Result SolverEngine::checkSynth()
+Result SolverEngine::checkSynth(bool isNext)
 {
   SolverEngineScope smts(this);
   finishInit();
-  return d_sygusSolver->checkSynth(*d_asserts);
+  if (isNext)
+  {
+    
+  }
```
HanielB, line 943: ?

HanielB: Ah, sorry, I looked at the wrong version

HanielB (approved):

ajreynol:
Marking do not merge for now, will need to add a unit test.

test/regress/regress1/sygus/incremental-stream-ex.sy
```diff
@@ -0,0 +1,28 @@
+; COMMAND-LINE: -i --sygus-out=status
+;EXPECT: unsat
+;EXPECT: unsat
+;EXPECT: unsat
+;EXPECT: unsat
```
abdoo8080, line 5: This test does not ensure that the results of `check-synth-next` are different. How about adding this one?
```
; COMMAND-LINE: -i
; SCRUBBER: grep "define-fun" | sort
; EXPECT: (define-fun f ((x Int) (y Int)) Int 0)
; EXPECT: (define-fun f ((x Int) (y Int)) Int 1)
; EXPECT: (define-fun f ((x Int) (y Int)) Int x)
; EXPECT: (define-fun f ((x Int) (y Int)) Int y)

(set-logic LIA)

(synth-fun f ((x Int) (y Int)) Int
  ((Start Int))
  ((Start Int (0 1 x y))))

(check-synth)
(check-synth-next)
(check-synth-next)
(check-synth-next)
```
It requires #7819 though...

ajreynol: Sounds good.

ajreynol: I'll add this regression after that PR is merged.

PR #7786, opened 2021-12-09
Eliminate more static rewrites

nafur:
This PR eliminates almost all remaining static rewrites from the arithmetic theory.


src/theory/arith/theory_arith_private.cpp
```diff
@@ -4782,8 +4782,11 @@ std::pair<bool, Node> TheoryArithPrivate::entailmentCheck(TNode lit, const Arith
   return make_pair(false, Node::null());
 }
 
-bool TheoryArithPrivate::decomposeTerm(Node term, Rational& m, Node& p, Rational& c){
-  Node t = Rewriter::rewrite(term);
+bool TheoryArithPrivate::decomposeTerm(Node t,
```
ajreynol, line 4785: I assume the rewrite was removed here since it always is already rewritten? Maybe change to assertion?

src/theory/arith/theory_arith_private.h
```diff
@@ -162,20 +162,31 @@ class TheoryArithPrivate : protected EnvObj
...
+                       Rational& rm,
+                       Node& rp,
+                       Rational& dm,
+                       Node& dp,
+                       DeltaRational& sep);
+ static void setToMin(int sgn,
+                      std::pair<Node, DeltaRational>& min,
+                      const std::pair<Node, DeltaRational>& e);
+
+ /**
```
ajreynol, line 180: Delete.

ajreynol (changes requested):
A few minor requests

src/theory/arith/theory_arith_private.cpp
nafur: This is true. For an assertion like `Assert(t == rewrite(t))` we'd still need access to the rewriter, but `decomposeTerm` is still static. I didn't want to make it non-static just to add an assertion...

ajreynol: Ah I didnt realize it was static.

ajreynol (approved):

PR #7790, opened 2021-12-10
Remove static options from sat solver.

nafur:
This PR removes `options::foo()` from minisat.


src/prop/minisat/core/Solver.cc
```diff
@@ -51,10 +51,10 @@ namespace {
...
  * keeps using them.
  */
-bool assertionLevelOnly()
+bool assertionLevelOnly(const Options& opts)
 {
-  return (options::produceProofs() || options::unsatCores())
-         && options::incrementalSolving();
+  return (opts.smt.produceProofs || opts.smt.unsatCores)
+         && opts.base.incrementalSolving;
 }
```
mpreiner, line 58: This looks like this should be a flag that is set once instead of always checking the same options over again.

mpreiner (approved):

PR #7791, opened 2021-12-10
Allow for wildcards in `-t`

nafur:
This PR extends `-t` to allow for wildcards: `-t nl-ext-*`. Internally, these wildcards are replaced by `.*` and fed into `std::regex`.


4tXJ7f (approved):
Neat!

PR #7794, opened 2021-12-10
api: Use 'note' constructs for API documentation.

aniemetz:
This uses `@note` for notes in the C++ API documentation, `.. note::`
for Python, and `@apiNote` for Java.


src/api/cpp/cvc5.h
```diff
@@ -987,9 +991,10 @@ class CVC5_EXPORT Op
 
   /**
    * The internal node wrapped by this operator.
-   * Note: This is a shared_ptr rather than a unique_ptr to avoid overhead due
-   *       to memory allocation (cvc5::Node is already ref counted, so this
-   *       could be a unique_ptr instead).
+   *
+   * @note This is a ``std::shared_ptr`` rather than a ``std::unique_ptr`` to
+   *       avoid overhead due to memory allocation (``cvc5::Node`` is already
+   *       ref counted, so this could be a ``std::unique_ptr`` instead).
```
nafur, line 997: ... so it should be changed to a unique_ptr?

aniemetz: No, because overhead due to memory allocation, as the note says.

src/api/python/cvc5.pxi
```diff
@@ -3060,19 +3062,23 @@ cdef class Term:
...
+        .. code::
+
+            (union
+                (singleton c1) ... (union (singleton c_{n-1}) (singleton c_n))))
+
+        where ``c1 ... cn`` are values ordered by id such that
+        ``c1 > ... > cn`` (see also :cpp:func:`cvc5::api::Term::operator>()`).
+
+        .. note::
+            A universe set term `(kind SET_UNIVERSE)` is not considered to be
```
nafur, line 3077:
```suggestion
            A universe set term ``(kind SET_UNIVERSE)`` is not considered to be
```

nafur (approved):
LGTM. I've not checked every single place in the generated docs, though.

aniemetz:
> LGTM. I've not checked every single place in the generated docs, though.

I have and unless I missed anything, it should be generated correctly.

PR #7795, opened 2021-12-11
api: Fix smt-lib code blocks and math in C++ docs.

aniemetz:
@ajreynol please have a careful look at the changes made for `getQuantifierElimination(Disjunct)`. Best to check the generated html at https://cvc5.github.io/docs-ci/docs-pr7795/api/cpp/solver.html#_CPPv4NK4cvc53api6Solver24getQuantifierEliminationERK4Term and https://cvc5.github.io/docs-ci/docs-pr7795/api/cpp/solver.html#_CPPv4NK4cvc53api6Solver32getQuantifierEliminationDisjunctERK4Term.

### DIFF
--- a/src/api/cpp/cvc5.h
+++ b/src/api/cpp/cvc5.h
@@ -1510,9 +1510,9 @@ class CVC5_EXPORT Term
    *
    * A term is a set value if it is considered to be a (canonical) constant set
    * value.  A canonical set value is one whose AST is:
-   * ```
-   *   (union (singleton c1) ... (union (singleton c_{n-1}) (singleton c_n))))
-   * ```
+   *
+   *     (union (singleton c1) ... (union (singleton c_{n-1}) (singleton c_n))))
+   *
    * where `c1 ... cn` are values ordered by id such that `c1 > ... > cn` (see
    * also @ref Term::operator>(const Term&) const).
    *
@@ -1532,7 +1532,7 @@ class CVC5_EXPORT Term
   bool isSequenceValue() const;
   /**
    * Asserts isSequenceValue().
-   * @note It is usually necessary for sequences to call `Solver::simplify()`
+   * @note It is usually necessary for sequences to call Solver::simplify()
    *       to turn a sequence that is constructed by, e.g., concatenation of
    *       unit sequences, into a sequence value.
    * @return the representation of a sequence value as a vector of terms.
@@ -1983,13 +1983,18 @@ class CVC5_EXPORT DatatypeConstructor
    *
    * This method is required for constructors of parametric datatypes whose
    * return type cannot be determined by type inference. For example, given:
-   *   (declare-datatype List (par (T) ((nil) (cons (head T) (tail (List T))))))
+   *
+   *     (declare-datatype List
+   *         (par (T) ((nil) (cons (head T) (tail (List T))))))
+   *
    * The type of nil terms need to be provided by the user. In SMT version 2.6,
    * this is done via the syntax for qualified identifiers:
-   *   (as nil (List Int))
+   *
+   *     (as nil (List Int))
+   *
    * This method is equivalent of applying the above, where this
    * DatatypeConstructor is the one corresponding to nil, and retSort is
-   * (List Int).
+   * ``(List Int)``.
    *
    * @note the returned constructor term ``t`` is an operator, while
    *       ``Solver::mkTerm(APPLY_CONSTRUCTOR, t)`` is used to construct the
@@ -3691,11 +3696,11 @@ class CVC5_EXPORT Solver
 
   /**
    * Create (first-order) constant (0-arity function symbol).
+   *
    * SMT-LIB:
-   * \verbatim
-   *   ( declare-const <symbol> <sort> )
-   *   ( declare-fun <symbol> ( ) <sort> )
-   * \endverbatim
+   *
+   *     ( declare-const <symbol> <sort> )
+   *     ( declare-fun <symbol> ( ) <sort> )
    *
    * @param sort the sort of the constant
    * @param symbol the name of the constant
@@ -3779,30 +3784,33 @@ class CVC5_EXPORT Solver
 
   /**
    * Assert a formula.
+   *
    * SMT-LIB:
-   * \verbatim
-   *   ( assert <term> )
-   * \endverbatim
+   *
+   *     ( assert <term> )
+   *
    * @param term the formula to assert
    */
   void assertFormula(const Term& term) const;
 
   /**
    * Check satisfiability.
+   *
    * SMT-LIB:
-   * \verbatim
-   *   ( check-sat )
-   * \endverbatim
+   *
+   *     ( check-sat )
+   *
    * @return the result of the satisfiability check.
    */
   Result checkSat() const;
 
   /**
    * Check satisfiability assuming the given formula.
+   *
    * SMT-LIB:
-   * \verbatim
-   *   ( check-sat-assuming ( <prop_literal> ) )
-   * \endverbatim
+   *
+   *     ( check-sat-assuming ( <prop_literal> ) )
+   *
    * @param assumption the formula to assume
    * @return the result of the satisfiability check.
    */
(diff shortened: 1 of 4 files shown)


src/api/cpp/cvc5.h
```diff
@@ -1505,9 +1505,9 @@ class CVC5_EXPORT Term
    *
    * A term is a set value if it is considered to be a (canonical) constant set
    * value.  A canonical set value is one whose AST is:
-   * ```
-   *   (union (singleton c1) ... (union (singleton c_{n-1}) (singleton c_n))))
-   * ```
+   *
+   *     (union (singleton c1) ... (union (singleton c_{n-1}) (singleton c_n))))
+   *
```
nafur, line 1510: A few lines below, we should add backticks to `(kind SET_UNIVERSE)`.

src/api/cpp/cvc5.h
```diff
@@ -4043,21 +4069,23 @@ class CVC5_EXPORT Solver
...
 
   /**
    * Get the values of the given terms in the current model.
+   *
    * SMT-LIB:
-   * \verbatim
-   * ( get-value ( <term>+ ) )
-   * \endverbatim
+   *
+   *     ( get-value ( <term>+ ) )
```
nafur, line 4087: Technically, this should be `<term>*`...

src/api/cpp/cvc5.h
```diff
@@ -4102,49 +4131,59 @@ class CVC5_EXPORT Solver
...
-   * @param q a quantified formula of the form:
-   *   Q x1...xn. P( x1...xn, y1...yn )
-   * where P( x1...xn, y1...yn ) is a quantifier-free formula
-   * @return a formula ret such that, given the current set of formulas A
-   * asserted to this solver:
-   *   - ( A ^ q ) and ( A ^ ret ) are equivalent
-   *   - ret is quantifier-free formula containing only free variables in
-   *     y1...yn.
+   * @param q a quantified formula of the form
+   *          @f$Q x_1...x_n. P( x_1...x_n, y_1...y_n )@f$ where
```
nafur, line 4142: Can we handle quantifier alternations here? @ajreynol

src/api/cpp/cvc5.h
```diff
@@ -4223,42 +4267,47 @@ class CVC5_EXPORT Solver
...
+   *               and @f$A \wedge \neg B \wedge C@f$ is unsatisfiable, where
+   *               @f$A@f$ is the current set of assertions and @f$B@f$ is
+   *               given in the input by ``conj``
+   * @return true if it gets abduct @f$C@f$ successfully, false otherwise
    */
   bool getAbduct(const Term& conj, Term& output) const;
 
   /**
    * Get an abduct.
+   *
```
nafur, line 4287: From here on, the remaining functions are not in the generated documentation!?!

nafur (commented):
A general thing: should we add proper highlighting to the smtlib code blocks with `..code:: smtlib`? If so, I'd suggest to do the same for the AST "code" blocks; although they are not smtlib, they are sufficiently alike that the smtlib highlighter should be good enough. Note that right now there is some syntax-highlighting, catching `assert` and `as`, but not the remaining smtlib stuff).

Also: why do we have spaces in some code blocks? `( assert <term> )`

src/api/cpp/cvc5.h
ajreynol: Yes.

FYI, the documentation should be updated to convey that LRA and LIA are the only ones where QE is complete, although I believe we should support any theory in an incomplete way here.

src/api/cpp/cvc5.h
```diff
@@ -4223,42 +4267,47 @@ class CVC5_EXPORT Solver
...
    * Requires enabling option 'produce-abducts'
    * @param conj the conjecture term
-   * @param grammar the grammar for the abduct C
-   * @param output a term C such that A^C is satisfiable, and A^~B^C is
-   *        unsatisfiable, where A is the current set of assertions and B is
-   *        given in the input by conj
-   * @return true if it gets C successfully, false otherwise
+   * @param grammar the grammar for the abduct @f$C@f$
+   * @param output a term C such that @f$A \wedge @f$C@f$ is satisfiable, and
```
nafur, line 4295: I suspect that we have an odd number of `@f$` here...
```suggestion
   * @param output a term C such that @f$A \wedge C@f$ is satisfiable, and
```

src/api/cpp/cvc5.h
aniemetz: According to the SMT-LIB standard,it is required that `n > 0` for terms `t1...tn`. We do not guard against this though, we simply return an empty vector in that case. So yeah, I guess we should update this to `*`.

src/api/cpp/cvc5.h
aniemetz: I added a sentence.

nafur: If we support quantifier alternations, we should update to something like
`@f$ Q_1 x_1...x_n Q_2 x_{n+1}...x_{m}. P( x_1...x_m, y_1...y_n)@f$`. Also, we should not use `n` for both x and y, they do not need to be the same amount of `x`s and `y`s.

src/api/cpp/cvc5.h
```diff
@@ -4184,20 +4227,22 @@ class CVC5_EXPORT Solver
...
    * Get an interpolant
+   *
    * SMT-LIB:
-   * \verbatim
-   * ( get-interpol <conj> )
-   * \endverbatim
+   *
+   *     ( get-interpol <conj> )
+   *
    * Requires to enable option 'produce-interpols'.
```
nafur, line 4246: It would be awesome if we could reference the option here. In rst, it works like this:
```
:ref:`--output <lbl-option-output>`
```

src/api/cpp/cvc5.h
```diff
@@ -4284,50 +4338,55 @@ class CVC5_EXPORT Solver
...
 
   /**
    * Set option.
+   *
    * SMT-LIB:
-   * \verbatim
-   *   ( set-option <option> )
-   * \endverbatim
+   *
+   *     ( set-option <option> )
```
nafur, line 4388:
```suggestion
   *     ( set-option :<option> <value> )
```

src/api/cpp/cvc5.h
```diff
@@ -4385,10 +4446,11 @@ class CVC5_EXPORT Solver
 
   /**
    * Synthesize n-ary function following specified syntactic constraints.
+   *
    * SyGuS v2:
-   * \verbatim
-   *   ( synth-fun <symbol> ( <boundVars>* ) <sort> <g> )
-   * \endverbatim
+   *
+   *     ( synth-fun <symbol> ( <boundVars>* ) <sort> <g> )
```
nafur, line 4452:
```suggestion
   *     ( synth-fun <symbol> ( <boundVars>* ) <sort> <grammar> )
```

src/api/cpp/cvc5.h
```diff
@@ -4415,10 +4478,11 @@ class CVC5_EXPORT Solver
 
   /**
    * Synthesize invariant following specified syntactic constraints.
+   *
    * SyGuS v2:
-   * \verbatim
-   *   ( synth-inv <symbol> ( <boundVars>* ) <g> )
-   * \endverbatim
+   *
+   *     ( synth-inv <symbol> ( <boundVars>* ) <g> )
```
nafur, line 4484:
```suggestion
   *     ( synth-inv <symbol> ( <boundVars>* ) <grammar> )
```

src/api/cpp/cvc5.h
```diff
@@ -4590,10 +4658,11 @@ class CVC5_EXPORT Solver
 
   /**
    * Synthesize n-ary function following specified syntactic constraints.
+   *
    * SMT-LIB:
-   * \verbatim
-   *   ( synth-fun <symbol> ( <boundVars>* ) <sort> <g>? )
-   * \endverbatim
+   *
+   *     ( synth-fun <symbol> ( <boundVars>* ) <sort> <g>? )
```
nafur, line 4664:
```suggestion
   *     ( synth-fun <symbol> ( <boundVars>* ) <sort> <grammar>? )
```

nafur:
LGTM. Great!

nafur (approved):
LGTM. Great!

PR #7801, opened 2021-12-13
Add relations.cpp, relations.py examples

mudathirmahgoub:


examples/api/cpp/relations.cpp
```diff
@@ -0,0 +1,148 @@
...
+{
+  Solver solver;
+
+  // Set the logic
+  solver.setLogic("ALL");
+
+  // options
+  solver.setOption("produce-models", "true");
+  solver.setOption("finite-model-find", "true");
+  solver.setOption("sets-ext", "true");
```
aniemetz, line 32: Can you maybe comment on why we need those 2? (And add these comments to .py and .java, too)

examples/api/cpp/relations.cpp
```diff
@@ -0,0 +1,148 @@
...
+
+  // (assert (= people (as set.universe (Set (Tuple Person)))))
+  Term peopleAreTheUniverse = solver.mkTerm(EQUAL, people, universeSet);
+  // (assert (not (= males (as set.empty (Set (Tuple Person))))))
+  Term maleSetIsNotEmpty = solver.mkTerm(NOT, isEmpty1);
+  // (assert (not (= females (as set.empty (Set (Tuple Person))))))
+  Term femaleSetIsNotEmpty = solver.mkTerm(NOT, isEmpty2);
+
+  // (assert (= (set.inter males females) (as set.empty (Set (Tuple
+  // Person)))))
```
aniemetz, line 78:
```suggestion
  // (assert
  //   (= (set.inter males females) (as set.empty (Set (Tuple Person)))))
```

examples/api/cpp/relations.cpp
mudathirmahgoub: Done.

examples/api/cpp/relations.cpp
mudathirmahgoub: Done.

nafur (commented):
When you are working on this anyway, could you add a short description to `relations.rst` on what this example is supposed to demonstrate? I'd really like to have that for the examples, much like for the linear arithmetic one. It should probably include a reference to the theory reference on sets and relations.

Otherwise this looks good. Thanks!

mudathirmahgoub:
> When you are working on this anyway, could you add a short description to `relations.rst` on what this example is supposed to demonstrate? I'd really like to have that for the examples, much like for the linear arithmetic one. It should probably include a reference to the theory reference on sets and relations.
> 
> Otherwise this looks good. Thanks!

Added some documentation to the example. Please check if it makes sense

mudathirmahgoub:
It is hard to read the rst source file, so this image may be better for review
image

docs/examples/relations.rst
```diff
@@ -1,8 +1,31 @@
...
 ===================
 
+This simple family example demonstrates theory of finite relations extension to the theory of finite sets.
+Relations are defined as sets of tuples with arity :math:`\geq 1`.
+We have unary relations :math:`people, males,` and :math:`females` of arity 1.
+We also have binary relations :math:`father, mother, parent, ancestor`, and :math:`descendant` of arity 2.
+Here we follow a prefix convention (i.e., :math:`R(x,y)` or
+:math:`\langle x, y \rangle \in R` if and only if
+person :math:`x` has relation `R` with person :math:`y`,
+where :math:`R \in \{father, mother, parent, ancestor, descendant\}`.
```
nafur, line 11:
```suggestion
This simple example demonstrates the combined :ref:`theory of finite sets and finite relations <../theories/sets-and-relations>` using a family tree.
Relations are defined as sets of tuples with arity :math:`\geq 1`.
The example includes the unary relations :math:`people, males,` and :math:`females` and the binary relations :math:`father, mother, parent, ancestor`, and :math:`descendant`.
```

nafur (commented):
Thanks! I'd suggest to remove the mathematical definitions (and possibly adding them to the theory reference page instead).

docs/examples/relations.rst
```diff
@@ -1,8 +1,31 @@
...
+
+- All relations are nonempty.
+- People is the universe set.
+- Males and females are disjoint sets (i.e., :math:`males \cap females = \phi`).
+- Fathers are males (i.e., :math:`father \bowtie people \subseteq males`).
+- Mothers are females (i.e., :math:`mother \bowtie people \subseteq females`).
+- A parent is a father or a mother (i.e., :math:`parent = father \cup mother`).
+- Ancestor relation is the transitive closure of parent (i.e., :math:`ancestor = parent^{+}`).
+- Descendant relation is the transpose of ancestor (i.e., :math:`descendant = ancestor^{-1}`).
+- No self ancestor (i.e., :math:`\forall x: Person. \langle x, x \rangle \not\in ancestor`).
```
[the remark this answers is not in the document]
mudathirmahgoub: I agree both syntax and semantics of these operators need to be defined in the theory page, but 
the mathematical formulation here is related to the example, and I think it is important to show the mathematical notation. 
Without them, one would express the constraint `fathers are males` using quantifiers instead of the appropriate relational join operator.

docs/examples/relations.rst
mudathirmahgoub: Done

docs/examples/relations.rst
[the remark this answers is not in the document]
mudathirmahgoub: I added a footnote for the bowtie symbol with the citation paper.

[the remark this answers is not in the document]
mudathirmahgoub: I prefer to address the definitions of relational operators in a separate PR.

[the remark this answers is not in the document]
nafur: Okay, that's fine.

nafur (approved):
Okay, thanks!

PR #7804, opened 2021-12-13
Fix cvc5-projects issue 358

mudathirmahgoub:
This PR fixes issue https://github.com/cvc5/cvc5-projects/issues/358 which is caused by unregistered terms in the equality engine of bags


src/theory/bags/theory_bags.cpp
```diff
@@ -299,7 +299,7 @@ Node TheoryBags::getModelValue(TNode node) { return Node::null(); }
 
 void TheoryBags::preRegisterTerm(TNode n)
 {
-  Trace("bags::TheoryBags::preRegisterTerm") << n << std::endl;
+  Trace("bags") << "TheoryBags::preRegisterTerm(" << n << ")" << std::endl;
```
ajreynol, line 302: Can you also register trigger predicate in n.getKind()==EQUAL here?

ajreynol (changes requested):

ajreynol (approved):

PR #7805, opened 2021-12-13
Make some undocumented options regular/expert

nafur:
We have some options that are currently `"undocumented"` for no good reason. This PR makes them regular or expert options.

### DIFF
--- a/src/options/README.md
+++ b/src/options/README.md
@@ -26,7 +26,7 @@ Specifying Options
 Options can be defined within a module file with the `[[option]]` tag, the
 required attributes for an option are:
 
-* `category` (string): one of `common`, `expert`, `regular`, or `undocumented`
+* `category` (string): one of `common`, `regular`, `expert`, or `undocumented`
 * `type` (string): the C++ type of the option value, see below for more details.
 
 Optional attributes are:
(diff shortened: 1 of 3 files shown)


src/options/README.md
```diff
@@ -59,6 +59,16 @@ Optional attributes are:
   given)
 
 
+Option categories
+-----------------
+
+Every option has one of the following categories that influences where and how an option is visible:
+
+* `common`: used for the most common options. All `common` options are shown at the very top in both the online documentation and the output of `--help` on the command line.
```
aniemetz, line 67:
```suggestion
* `common`: Used for the most common options. All `common` options are shown at the very top in both the online documentation and the output of `--help` on the command line.
```

src/options/README.md
```diff
@@ -59,6 +59,16 @@ Optional attributes are:
...
+* `regular`: this should be used for most options.
+* `expert`: used for options that should be used with care only. A warning is shown in both the online documentation and the command line help.
```
aniemetz, line 69:
```suggestion
* `expert`: This is for options that should be used with care only. A warning is shown in both the online documentation and the command line help.
```

aniemetz: Maybe clarify what kind of warning? Without context this might be a bit weird for users.

src/options/README.md
```diff
@@ -59,6 +59,16 @@ Optional attributes are:
...
+* `undocumented`: such an option is skipped entirely in both the online documentation and the command line help. This should only be used when no user has a (reasonable) use case for this option (e.g. because it stores data that is added via another option like for `output` and `outputTagHolder`).
```
aniemetz, line 70:
```suggestion
* `undocumented`: Such an option is skipped entirely in both the online documentation and the command line help. This should only be used when users don't have a (reasonable) use case for this option (e.g., because it stores data that is added via another option like for `output` and `outputTagHolder`).
```

aniemetz (approved):

PR #7808, opened 2021-12-13
Eliminate use of rewrite, CONST_RATIONAL in ArithMSum

ajreynol:


src/theory/arith/arith_msum.cpp
```diff
@@ -83,6 +83,7 @@ bool ArithMSum::getMonomialSumLit(Node lit, std::map<Node, Node>& msum)
 {
   if (lit.getKind() == GEQ || lit.getKind() == EQUAL)
   {
+    TypeNode tn = lit[0].getType();
```
nafur, line 86: Please move this further down, probably toward line 100.

src/theory/arith/arith_msum.cpp
```diff
@@ -81,7 +81,8 @@ bool ArithMSum::getMonomialSum(Node n, std::map<Node, Node>& msum)
 
 bool ArithMSum::getMonomialSumLit(Node lit, std::map<Node, Node>& msum)
 {
-  if (lit.getKind() == GEQ || lit.getKind() == EQUAL)
+  if (lit.getKind() == GEQ
+      || (lit.getKind() == EQUAL && lit[0].getType().isRealOrInt()))
```
nafur, line 85: This guards against being called with `lit[0]` being Boolean, right? Shouldn't this rather be checked by the caller?

nafur (approved):
One comment, otherwise LGTM.

src/theory/arith/arith_msum.cpp
ajreynol: Well its already guarding for the kind of literal here (GEQ, EQUAL), so this check should go here.

PR #7793, opened 2021-12-10
Replace `Debug` by `Trace`

nafur:
This PR replaces all usages of `Debug()` by `Trace()`, the same for similar methods like `Debug.isOn()`.
Given that `Debug` is no longer used then, it is removed.

PR #7809, opened 2021-12-14
Fix issues with tracing builds

nafur:
This PR fixes two issues that were revealed when analyzing decreased performance on trace-enabled builds.
`hasIntegerModel()` turns out to be a rather expensive method and should thus not be called in a `Trace` output.
Furthermore, the implementation of `Rational::isIntegral()` was generally bad because it used `getDenominator()` (which returns a copy of the denominator as an `Integer`) instead of directly checking the underlying denominator with equality for zero.


nafur:
On a (random) QF_IDL benchmark, these changes moved the tracing build from >3s to ~900ms, compared to ~700ms for a non-tracing build. I hope these changes are sufficient to make #7793 acceptable.

src/util/rational_gmp_imp.h
```diff
@@ -286,7 +286,7 @@ class CVC5_EXPORT Rational
     return (*this);
   }
 
-  bool isIntegral() const { return getDenominator() == 1; }
+  bool isIntegral() const { return mpz_cmp_ui(d_value.get_den_mpz_t(), 1) == 0; }
```
ajreynol, line 289: I believe the same thing be done for the CLN rational?

ajreynol (changes requested):

ajreynol (approved):

PR #7810, opened 2021-12-14
int-to-bv: fail if one of the arguments has type real

yoni206:
This PR generates a failure in int-to-bv in case a real argument is detected as a child of an arithmetic operation.
fixes https://github.com/cvc5/cvc5-projects/issues/348 .

Co-authored-by: Andrew Reynolds <andrew.j.reynolds@gmail.com>


ajreynol (changes requested):
This is not the right way to check this, what if another operator is added that has real type?

Isn't it the case that anytime any subterm of the input has real type, the pass should fail?

yoni206:
Should this fail?
```
(set-logic ALL)
(set-option :solve-int-as-bv 1)
(declare-const x Real)
(declare-const y Real)
(assert (= y x))
(check-sat)
```

On current master it doesn’t, but we can make it fail with this PR (I guess the pass is not aimed at benchmarks like this one).
What do you think?

ajreynol:
It is fine if the above benchmark fails.  Although notice that it may be simplified to true eagerly if e.g. non-clausal simplification is run first.

yoni206:
Is this better?  https://github.com/cvc5/cvc5/pull/7810/commits/c163c7803b351e38bd6627ebb15b96fbd408e4b3

ajreynol (changes requested):
Can't you just check whether the type of current is real?  Also, you need to use `isReal() && !isInteger()` to check whether something has type `Real` currently due to arithmetic subtyping.

ajreynol (approved):

PR #7812, opened 2021-12-14
bv-to-int: use pow2 operator

yoni206:
Currently, the int-blaster translates shifts using an `ite` whose size is the bit-width.
This PR adds an option to use the internal `pow2` operator instead.


yoni206:
done here: https://github.com/cvc5/cvc5/pull/7810/commits/bb4b52d7d4e90bbc0748d88d54ca88c6cdbdc3d7

ajreynol (approved):

PR #7813, opened 2021-12-14
Add trace to see inferences in final proof

ajreynol:
Adds `-t im-pf` to see which inferences occur in the final proof.  Must be used with proofs and `--proof-annotate`.


4tXJ7f (approved):
Awesome :)

PR #7816, opened 2021-12-15
Eliminate more uses of CONST_RATIONAL

ajreynol:


src/parser/smt2/smt2.cpp
```diff
@@ -1019,15 +1019,14 @@ api::Term Smt2::applyParseOp(ParseOp& p, std::vector<api::Term>& args)
...
-        && constVal[1].getKind() == api::CONST_RATIONAL)
+    if (constVal.getKind() == api::DIVISION && isConstIntOrReal(constVal[0])
+        && isConstIntOrReal(constVal[1]))
     {
       std::stringstream sdiv;
       sdiv << (isNeg ? "-" : "") << constVal[0] << "/" << constVal[1];
       constVal = d_solver->mkReal(sdiv.str());
     }
-    else if (constVal.getKind() == api::CONST_RATIONAL && isNeg)
+    else if (isConstIntOrReal(constVal) && isNeg)
```
nafur, line 1029: Firstly: I think this only works if these values are integer.For rational values, the strings will not be properly formatted.
But secondly: why do we convert to string anyway???

src/theory/quantifiers/cegqi/ceg_instantiator.cpp
```diff
@@ -125,6 +125,26 @@ std::ostream& operator<<(std::ostream& os, CegHandledStatus status)
   return os;
 }
 
+/**
+ * Return the result of multiplying constant integer or real nodes c1 and c2.
+ * The returned type is real if either have type real.
+ */
+Node multConstants(const Node& c1, const Node& c2)
```
nafur, line 132: Maybe this could be a more general utility?

src/parser/smt2/smt2.cpp
ajreynol: Converting to string is the only overflow safe method for constructing integers in the API.

src/theory/quantifiers/cegqi/ceg_instantiator.cpp
ajreynol: Sure, I can move this to `arith_utilities.h`

src/parser/smt2/smt2.cpp
ajreynol: I've guarded for constant integers now.

There is a more general fix that I will apply in a later PR.

nafur (approved):
LGTM.

PR #7819, opened 2021-12-16
[Regressions] Support more complex scrubbers

4tXJ7f:
Currently, we only support single commands in our `SCRUBBER` directives
for regression tests. This commit adds support for executing more
complex commands, including pipes. It also updates the regression script
to use the more modern, higher-level `subprocess.run()` instead of
`subprocess.Popen()`.


test/regress/run_regression.py
```diff
@@ -370,33 +370,31 @@ def print_diff(actual, expected):
...
     output and the error output are empty and the exit code is 124."""
 
-    proc = subprocess.Popen(
-        args,
-        cwd=cwd,
-        stdin=subprocess.PIPE,
-        stdout=subprocess.PIPE,
-        stderr=subprocess.PIPE,
-    )
+    cmd = shlex.join(args) if isinstance(args, list) else args
```
nafur, line 379: `join` only comes with 3.8. I guess we should use something like `' '.join([shlex.quote(a) for a in args])` instead...

test/regress/run_regression.py
```diff
@@ -370,33 +370,31 @@ def print_diff(actual, expected):
...
-    finally:
-        if timeout:
-            # The timer killed the process and is not active anymore.
-            if exit_status == -9 and not timer.is_alive():
-                exit_status = STATUS_TIMEOUT
-            timer.cancel()
+        res = subprocess.run(
+            cmd,
+            cwd=cwd,
+            stdin=subprocess.PIPE,
```
nafur, line 388:
```suggestion
            stdin=subprocess.PIPE,
            input=s_input,
```

test/regress/run_regression.py
4tXJ7f: Argh, good catch

test/regress/run_regression.py
4tXJ7f: Great point

abdoo8080 (changes requested):

test/regress/regress0/parser/shadow_fun_symbol_all.smt2
```diff
@@ -1,5 +1,5 @@
 ; EXPECT: Symbol `sin' is shadowing a theory function symbol
-; SCRUBBER: grep -o "Symbol `sin' is shadowing a theory function symbol"
+; SCRUBBER: grep -o "Symbol \`sin' is shadowing a theory function symbol"
```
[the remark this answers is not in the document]
nafur: This is because we added `shell=True`. Note that the backtick needs to be escaped when executed in a regular shell as well. I'd thus suggest to leave it like it is right now (in the PR).

test/regress/run_regression.py
```diff
@@ -370,33 +370,31 @@ def print_diff(actual, expected):
...
-            # The timer killed the process and is not active anymore.
-            if exit_status == -9 and not timer.is_alive():
-                exit_status = STATUS_TIMEOUT
-            timer.cancel()
+        res = subprocess.run(
+            cmd,
+            cwd=cwd,
+            input=s_input,
+            timeout=timeout,
+            capture_output=True,
```
nafur, line 390:
```suggestion
            stderr=subprocess.PIPE,
            stdout=subprocess.PIPE,
```

test/regress/regress0/parser/shadow_fun_symbol_all.smt2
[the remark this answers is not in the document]
abdoo8080: I see. Thanks for the explanation.

4tXJ7f:
Alright, it looks like I was able to find a workaround that also works for other platforms (other Python/shell versions? Not sure what exactly caused the issues).

nafur (approved):
LGTM. I assume you have checked that the latest version still supports pipes? :)

4tXJ7f:
> LGTM. I assume you have checked that the latest version still supports pipes? :)

I did and it looked good :)

4tXJ7f:
@abdoo8080 Is there anything else that needs to be addressed?

abdoo8080 (approved):
> @abdoo8080 Is there anything else that needs to be addressed?

No. The PR looks great.

PR #7820, opened 2021-12-16
Array-inspired Sequence Solver - Fixing several issues in the ArrayCoreSolver class and options

Ying1123:
This pull request fixed several issues in the ArrayCoreSolver class and the options:
1. Add range checking for an update rule.
2. Assign different inference ids to different rules.
3. small syntax optimizations.


src/theory/strings/array_core_solver.cpp
```diff
@@ -95,63 +95,68 @@ void ArrayCoreSolver::checkUpdate(const std::vector<Node>& updateTerms)
...
     Trace("seq-update") << "- " << termProxy << " = " << n << std::endl;
     std::vector<Node> exp;
     d_im.addToExplanation(termProxy, n, exp);
 
-    // optimization: add a short cut t == (seq.update n[0] n[1] n[2]) => t[i] ==
-    // n[2][0]
+    // optimization: add a short cut
+    // t = (seq.update n[0] n[1] n[2])
+    // ---------------------------------------------------------------------
+    // nth(t, n[1]) = (ITE, len(n[0]) > 0, nth(n[2], 0), nth(n[0], n[1]))
```
yoni206, line 113: Shouldn't we also check that `n[1] < len)n[0]`?

yoni206 (changes requested):
Thanks for this! I have one comment inline. In addition -- do we solve the benchmark you asked about after this fix? IF we do, can you add it to the regressions? If we don't, we might want to include some improvement in this PR already that solves. it.

src/theory/strings/array_core_solver.cpp
```diff
@@ -95,63 +95,68 @@ void ArrayCoreSolver::checkUpdate(const std::vector<Node>& updateTerms)
...
     Node left = nm->mkNode(SEQ_NTH, termProxy, n[1]);
-    Node right =
-        nm->mkNode(SEQ_NTH, n[2], nm->mkConstInt(Rational(0)));  // n[2][0]
+    Node cond = nm->mkNode(
+        GT, nm->mkNode(STRING_LENGTH, n[0]), nm->mkConstInt(Rational(0)));
+    Node body1 = nm->mkNode(SEQ_NTH, n[2], nm->mkConstInt(Rational(0)));
+    Node body2 = nm->mkNode(SEQ_NTH, n[0], nm->mkConstInt(Rational(0)));
+    Node right = nm->mkNode(ITE, cond, body1, body2);
```
yoni206, line 119: @4tXJ7f is there anything wrong with introducing ITEs in lemmas? Would it be preferable to introduce implications? Or maybe this is rewritten anyway before it is being sent?

src/theory/strings/array_core_solver.cpp
4tXJ7f, line 113: I don't think that this lemma is correct, because the length of `n[0]` should not matter. Rather, what matters is that `n[1]` is within `0 <= n[1] < len(n[0])`. Am I missing something?

src/theory/strings/array_core_solver.cpp
4tXJ7f: ITEs are fine, yes. They are removed as part of "term formula removal".

src/theory/inference_id.cpp
```diff
@@ -411,6 +411,10 @@ const char* toString(InferenceId i)
       return "STRINGS_ARRAY_NTH_EXTRACT";
     case InferenceId::STRINGS_ARRAY_NTH_UPDATE:
       return "STRINGS_ARRAY_NTH_UPDATE";
+    case InferenceId::STRINGS_ARRAY_NTH_UPDATE_SHORTCUT1:
+      return "STRINGS_ARRAY_NTH_UPDATE_SHORTCUT1";
+    case InferenceId::STRINGS_ARRAY_NTH_UPDATE_SHORTCUT2:
+      return "STRINGS_ARRAY_NTH_UPDATE_SHORTCUT2";
```
4tXJ7f, line 417: Rather than calling the inferences shortcut 1 and shortcut 2, can we give a name that is more informative?

src/theory/strings/array_core_solver.cpp
```diff
@@ -95,63 +95,68 @@ void ArrayCoreSolver::checkUpdate(const std::vector<Node>& updateTerms)
...
-    Trace("seq-array-debug") << "enter" << std::endl;
-    sendInference(exp, lem, InferenceId::STRINGS_ARRAY_NTH_UPDATE);
+    sendInference(exp, lem, InferenceId::STRINGS_ARRAY_NTH_UPDATE_SHORTCUT1);
 
     // enumerate possible index
-    for (auto nth : d_index_map)
+    for (const auto& nth : d_index_map)
     {
       Node seq = nth.first;
       if (d_state.areEqual(seq, n) || d_state.areEqual(seq, n[0]))
```
4tXJ7f, line 127: I am actually not sure if this condition is correct: It seems that this is implicitly requiring that if we have `t = update(s, i, n)` that `s = t`, i.e., that the update has no effect?

src/theory/strings/array_core_solver.cpp
```diff
@@ -95,63 +95,68 @@ void ArrayCoreSolver::checkUpdate(const std::vector<Node>& updateTerms)
...
       {
-        std::set<Node> indexes = nth.second;
+        const std::set<Node>& indexes = nth.second;
         for (Node j : indexes)
         {
-          // optimization: add a short cut for special case (seq.update n[0]
-          // n[1] (seq.unit e))
+          // optimization: add a short cut for special case
+          // n[2] is a unit term
+          // n[1] != j
```
4tXJ7f, line 134: Instead of referring to variables in the code (`n`, `j`, etc.), can we make this comment self contained?

src/theory/strings/array_core_solver.cpp
```diff
@@ -95,63 +95,68 @@ void ArrayCoreSolver::checkUpdate(const std::vector<Node>& updateTerms)
...
+          // ---------------------------------------------------------------
+          // nth(t, j) = nth(n[0], j)
```
4tXJ7f, line 136: The conclusion does not seem to match the lemma emitted by the code.

src/theory/strings/array_core_solver.cpp
```diff
@@ -95,63 +95,68 @@ void ArrayCoreSolver::checkUpdate(const std::vector<Node>& updateTerms)
...
           if (n[2].getKind() == SEQ_UNIT)
           {
             left = nm->mkNode(DISTINCT, n[1], j);
             Node nth1 = nm->mkNode(SEQ_NTH, termProxy, j);
             Node nth2 = nm->mkNode(SEQ_NTH, n[0], j);
             right = nm->mkNode(EQUAL, nth1, nth2);
             lem = nm->mkNode(IMPLIES, left, right);
```
4tXJ7f, line 143: We should think about whether we really want an implication here or whether we want to check the current state and only emit the lemma (with the left-hand side of the implication as the explanation) when the left-hand side is full-filled.

4tXJ7f (changes requested):
Some quick comments.

Ying1123:
> Thanks for this! I have one comment inline. In addition -- do we solve the benchmark you asked about after this fix? IF we do, can you add it to the regressions? If we don't, we might want to include some improvement in this PR already that solves. it.

It solves. We haven't finish the connection between core solver and sequence-array solver, so that we haven't start to add any regression test.

src/theory/strings/array_core_solver.cpp
Ying1123: You are right. Have been updated.

src/theory/strings/array_core_solver.cpp
Ying1123: It is an ``or'' rather than ``and''?

src/theory/strings/array_core_solver.cpp
Ying1123: Why does not match?

Ying1123: updated.

src/theory/strings/array_core_solver.cpp
Ying1123: updated.

src/theory/inference_id.cpp
Ying1123: updated.

src/theory/strings/array_core_solver.cpp
4tXJ7f: Oh shoot, right. I misread.

src/theory/strings/array_core_solver.cpp
```diff
@@ -95,63 +95,71 @@ void ArrayCoreSolver::checkUpdate(const std::vector<Node>& updateTerms)
...
     std::vector<Node> exp;
     d_im.addToExplanation(termProxy, n, exp);
 
-    // optimization: add a short cut t == (seq.update n[0] n[1] n[2]) => t[i] ==
-    // n[2][0]
+    // reasoning about nth(t, n[1]) even if it does not exist.
+    // t = (seq.update n[0] n[1] n[2])
+    // ---------------------------------------------------------------------
+    // nth(t, n[1]) = (ITE, n[1] in range(0, len(n[0])), nth(n[2], 0), nth(n[0],
+    // n[1]))
```
4tXJ7f, line 114:
```suggestion
    // x = update(s, n, t)
    // ---------------------------------------------------------------------
    // nth(x, n) = ite(n in range(0, len(s)), nth(t, 0), nth(s, n))
```

src/theory/strings/array_core_solver.cpp
```diff
@@ -95,63 +95,71 @@ void ArrayCoreSolver::checkUpdate(const std::vector<Node>& updateTerms)
...
+        const std::set<Node>& indexes = nth.second;
         for (Node j : indexes)
         {
-          // optimization: add a short cut for special case (seq.update n[0]
-          // n[1] (seq.unit e))
+          // optimization: add a short cut for special case
+          // t = (seq.update n[0] n[1] n[2])
+          // n[2] is a unit term
+          // -----------------------------------------
+          // n[1] != j -> nth(t, j) = nth(n[0], j)
```
4tXJ7f, line 139:
```suggestion
          // x = update(s, n, unit(t))
          // y = nth(s, m)
          // -----------------------------------------
          // n != m => nth(x, m) = nth(s, m)
```

4tXJ7f (approved):
Thanks for the updates. I just have some suggestions to make the comments easier to read and more consistent.

yoni206 (approved):

PR #7827, opened 2021-12-16
Remove Rewriter::rewrite from bags type enumerator

mudathirmahgoub:


ajreynol (approved):
LGTM, thanks for doing this.

PR #7828, opened 2021-12-16
Fix get-model when sort constructors are present

ajreynol:
Fixes a spurious error on kind benchmarks.

Note that we don't properly handle instantiated sort constructors currently, which I believe need to be handled as declared sorts when calling `getModel`.


abdoo8080 (approved):
LGTM!

PR #7787, opened 2021-12-10
Improve nonlinear solver

nafur:
This PR does two things:
- we remove splitting on shared values
- we add variable elimination for the cad-based solver, exploiting equalities present in the input.

PR #7829, opened 2021-12-16
Fix tracker in SubstitutionMap

nafur:
This PR fixes a subtle issue in #7787 where the rhs (instead of the lhs) of the applied substitution was inserted into the tracker.
Fixes cvc5/cvc5-projects#388


ajreynol:
The fix looks fine, @mpreiner can you confirm, is this the proper place for the added unit test?

mpreiner:
> The fix looks fine, @mpreiner can you confirm, is this the proper place for the added unit test?

Yep.

ajreynol (approved):

PR #7834, opened 2021-12-17
Refactoring initialization of proofs

ajreynol:
Eliminates one of the remaining calls to `Rewriter::callExtendedRewrite`, as well as making the initialization much clearer through use of Env.


HanielB (approved):
LGTM, really nice!

PR #7836, opened 2021-12-17
Add download link for examples in documentation

nafur:
This PR adds a download link to all examples in the documentation (that are included via the `examples` extension). I think we should do this, as we have text like "download the example here" at several places already...


aniemetz (approved):
Nice!

PR #7841, opened 2021-12-17
Fix proofs for datatype purify

ajreynol:
Previously we were failing to produce proofs for purification lemmas `(= t witness_form(t))`  from datatypes.


HanielB (commented):
Why was this not failing before? We do not have regressions that touch this? In this case maybe we should add one?

ajreynol:
This results in a pedantic level 0 failure before, which is not a regression failure. This makes our proof production for datatypes more complete, there are ~30-50 regressions that are impacted by this.

HanielB:
Ah, I see, the second change is what is leading to the first changed code to be executed. I was wondering what there was no proof checking failure. Thanks for clarifying.

HanielB (approved):

PR #7846, opened 2021-12-20
Eliminating some uses of const rational in arithmetic

ajreynol:
Note that there are several nested dependencies in arithmetic for constructing constants `Constant::mkConstant` ---> `mkRationalNode` ---> `mkConst(CONST_RATIONAL, r)`

This starts to disambiguate these calls.


src/theory/arith/arith_rewriter.cpp
```diff
@@ -83,9 +84,12 @@ RewriteResponse ArithRewriter::rewriteMinus(TNode t, bool pre){
 RewriteResponse ArithRewriter::rewriteUMinus(TNode t, bool pre){
   Assert(t.getKind() == kind::UMINUS);
 
-  if(t[0].getKind() == kind::CONST_RATIONAL){
+  if (t[0].isConst())
```
nafur, line 87: Technically this removes the check **which** type of constant this is. I assume the assertion above implicitly assures it is arithmetic?

ajreynol: Yes, the type rule for these kinds will ensure this is the case.

nafur (approved):
Otherwise LGTM.

PR #7839, opened 2021-12-17
Eliminate remaining calls to callExtendedRewrite

ajreynol:

PR #7848, opened 2021-12-21
Remove static extended rewrite from coverings solver

nafur:
This PR removes the last static rewrite from the arithmetic theory, which is a call to the extended rewrite within the coverings solver.


ajreynol:
I think this is subsumed by https://github.com/cvc5/cvc5/pull/7839, right?

nafur:
> I think this is subsumed by #7839, right?

Oh, you are right.

PR #7850, opened 2021-12-21
Support get-abduct-next

ajreynol:
Adds support for incremental abduction, adds regression for using push/pop in combination with get-abduct, and a use of get-abduct-next.

Adds this method to C++, java API.  

Interpolation/abduction not yet supported in Python API, I'm waiting for @yoni206 to add this; getAbductNext will be added to Python API in followup PR.

### DIFF
--- a/src/smt/abduction_solver.cpp
+++ b/src/smt/abduction_solver.cpp
@@ -82,7 +82,9 @@ bool AbductionSolver::getAbductNext(Node& abd)
   // Since we are using the subsolver's check-sat interface directly, we
   // simply call getAbductInternal again here. We assert that the subsolver
   // is already initialized, which must be the case or else we are not in the
-  // proper SMT mode to make this call.
+  // proper SMT mode to make this call. Due to the default behavior of
+  // subsolvers having synthesis conjectures, this is guaranteed to produce
+  // a new solution.
   Assert(d_subsolver != nullptr);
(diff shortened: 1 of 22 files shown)


src/smt/abduction_solver.cpp
```diff
@@ -73,19 +73,21 @@ bool AbductionSolver::getAbduct(const std::vector<Node>& axioms,
...
+bool AbductionSolver::getAbductNext(Node& abd)
 {
-  TypeNode grammarType;
-  return getAbduct(axioms, goal, grammarType, abd);
+  // Since we are using the subsolver's check-sat interface directly, we
+  // simply call getAbductInternal again here. We assert that the subsolver
+  // is already initialized, which must be the case or else we are not in the
+  // proper SMT mode to make this call.
+  Assert(d_subsolver != nullptr);
+  return getAbductInternal(abd);
```
yoni206, line 87: How come nothing is added to the call to `getAbductInternal` that would filter out the previous solution? Is it worth documenting this?

yoni206 (approved):
Just one question inline but otherwise LGTM.

src/smt/abduction_solver.cpp
ajreynol: Sure, I can add documentation.  The reason is this is the default behavior for sygus subsolvers with multiple check-sat.

PR #7851, opened 2021-12-21
Track input list for atoms in difficulty manager

ajreynol:
This makes our support for `get-difficulty` faster by tracking the set of input formulas that atoms occur in, which is highly important for benchmarks with many assertions and where many lemmas are added at standard effort.

This is work towards making `get-difficulty` scale on a large benchmark from Certora.


src/theory/difficulty_manager.cpp
```diff
@@ -48,8 +51,7 @@ void DifficultyManager::getDifficultyMap(std::map<Node, Node>& dmap)
   }
 }
 
-void DifficultyManager::notifyLemma(const context::CDHashMap<Node, Node>& rse,
-                                    Node n)
+void DifficultyManager::notifyLemma(Node n)
```
nafur, line 54: Make this a `TNode`?

src/theory/relevance_manager.cpp
```diff
@@ -40,7 +43,7 @@ RelevanceManager::RelevanceManager(Env& env, Valuation val)
 {
   if (options().smt.produceDifficulty)
   {
-    d_dman.reset(new DifficultyManager(userContext(), val));
+    d_dman.reset(new DifficultyManager(this, userContext(), val));
```
nafur, line 46:
```suggestion
    d_dman = std::make_unique<DifficultyManager>(this, userContext(), val);
```

src/theory/relevance_manager.cpp
```diff
@@ -156,7 +183,29 @@ void RelevanceManager::computeRelevance()
...
+    // if we are in full effort check and fail to justify, then we should
+    // give a failure and set success to false, or otherwise calls to
+    // isRelevant cannot be trusted.
+    if (d_inFullEffortCheck)
+    {
+      std::stringstream serr;
+      serr << "RelevanceManager::computeRelevance: WARNING: failed to justify "
+           << input;
+      Trace("rel-manager") << serr.str() << std::endl;
+      Assert(false) << serr.str();
```
nafur, line 203:
```suggestion
      Assert(false) << "RelevanceManager::computeRelevance: WARNING: failed to justify " << input;
```

src/theory/relevance_manager.cpp
```diff
@@ -389,6 +440,79 @@ bool RelevanceManager::isRelevant(Node lit)
...
+    TNode exp = getExplanationForRelevantInternal(lit);
+    if (!exp.isNull())
+    {
+      return exp;
+    }
+    // if the first time, we get the list of input formulas the atom occurs in
+    if (index==0)
+    {
+      ilist = getInputListFor(lit, false);
+      if (ilist!=nullptr)
```
nafur, line 466: format.

src/theory/relevance_manager.cpp
```diff
@@ -389,6 +440,79 @@ bool RelevanceManager::isRelevant(Node lit)
...
+{
+  NodeListMap::const_iterator it = d_atomMap.find(atom);
+  if (it == d_atomMap.end())
+  {
+    if (!doMake)
+    {
+      return nullptr;
+    }
+    d_atomMap[atom] = std::make_shared<NodeList>(userContext());
+    it = d_atomMap.find(atom);
```
nafur, line 511:
```suggestion
    it = d_atomMap.emplace(atom, std::make_shared<NodeList>(userContext()));
```

src/theory/relevance_manager.h
```diff
@@ -81,6 +81,7 @@ class RelevanceManager : protected EnvObj
   using RlvPairHashFunction = PairHashFunction<Node, uint32_t, std::hash<Node>>;
   using NodeList = context::CDList<Node>;
   using NodeMap = context::CDHashMap<Node, Node>;
+  using NodeListMap = context::CDHashMap<Node, std::shared_ptr<NodeList>>;
```
nafur, line 84: Why are we using a `std::shared_ptr` here?

src/theory/relevance_manager.cpp
ajreynol: This changes the intended behavior, I want a debug assertion failure or a Trace warning in production.

src/theory/relevance_manager.h
ajreynol: I'll change to unique.

ajreynol: Actually, this is currently a limitation of our context-dependent data structures, we can't store unique_ptr due to the requirements on the templated classes.

src/theory/relevance_manager.cpp
nafur: My concern is that this induces an overhead in a (non-tracing) production build.
How about this?
```
      Trace("rel-manager") << "RelevanceManager::computeRelevance: failed to justify " << input << std::endl
      Assert(false) << "RelevanceManager::computeRelevance: failed to justify " << input;
```

src/theory/relevance_manager.h
nafur: Ah I see...

nafur, line 84: Does it actually make sense to have a `CDList` inside a `CDHashMap`?

nafur (approved):
Just one question (possibly answered with a simple "yes"), otherwise LGTM.

src/theory/relevance_manager.cpp
ajreynol: This code really should never be run, so performance is not an issue here.

PR #7852, opened 2021-12-21
Guarding gmp `pow` function

yoni206:
The following test fails when using `gmp`, but not with `cln`.

```
TEST_F(TestUtilWhiteInteger, pow) {
  unsigned long int exp = std::numeric_limits<unsigned long int>::max();
  Integer ithree(3);
  Integer three_to_exp = ithree.pow(exp);
}
```

I suggest guarding the call to the gmp power function for numbers up to max `unsigned int`. 
I am not adding a variant of the above test because it fails only on `gmp`, and takes more than a minute to run on `cln`.

Co-authored-by: Gereon Kremer <nafur42@gmail.com>


nafur (commented):
I have a couple of suggestions:
- can we move to `uint64_t` and `uint32_t`?
- maybe we should only declare `pow(uint32_t)` and make this issue a parser error?
- I don't really like that we explicitly introduce this distinction between gmp and cln and even tell users about it.

yoni206:
Thanks for the review @nafur . I have pushed an update. Does it satisfy all your comments?

aniemetz (commented):
We have a potential issue here with this change:
https://github.com/cvc5/cvc5/blob/master/src/theory/bv/int_blaster.cpp#L45
(and related functions)

aniemetz (commented):
And here (passes in size_t):
https://github.com/cvc5/cvc5/blob/master/src/theory/arith/arith_rewriter.cpp#L447

aniemetz (commented):
And here (passes in unsigned long):
https://github.com/cvc5/cvc5/blob/master/src/util/cardinality.cpp#L172

aniemetz (commented):
And here:
https://github.com/cvc5/cvc5/blob/master/src/theory/arith/nl/iand_utils.cpp#L34
https://github.com/cvc5/cvc5/blob/master/src/theory/bv/int_blaster.cpp#L45

yoni206:
Thanks @aniemetz for these pointers!
Do you think my original commit would be better, given these problems?
https://github.com/cvc5/cvc5/pull/7852/files/7173435cf3d4962ea2e5956528d9fafe3c342362

Or do you have an idea for another solution?

Regardless of changing the signature of the function or throwing an exception from it — would you suggest I go to each call to this function and guard the call?

yoni206:
@nafur ping :)

nafur:
Hm, I'm somewhat split on this. On the one hand, computing the power of something with an exponent about `2^32` does not seem particularly useful. On the other hand, it may just make the difference to make some stuff work. `pow(2, 2^32)` takes 512MB, so it's in reach unless we do it all over the place...
Maybe we can simply do something like this?
```
uint32_t low = static_cast<uint32_t>(exp);
uint32_t high = static_cast<uint32_t>(exp << 32);
return pow(base, low) * pow(pow(base, static_cast<uint32_t>(-1)) * base, high);
```
(i.e. `a^(x + 2^n*y) = a^x * (a^(2^n))^y`)

If we implement this carefully, it should just add one simple check (`low == exp`) for the case that `exp < 2^32`...

yoni206:
This looks nice, though I would argue that that’s the job of the gmp people :)
But still, we could do that. Would you suggest we do this only for gmp or also for cln?

nafur:
> This looks nice, though I would argue that that’s the job of the gmp people :) But still, we could do that. Would you suggest we do this only for gmp or also for cln?

We only need that for gmp. cln supports arbitrary integer exponentiation via `cln::expt_pos(cl_I, cl_I)`.

yoni206:
@nafur the PR is not updated according to your suggestion. The test I mentioned when opening the PR now passes (but is not added to the PR because it takes a while to run).

src/util/integer_gmp_imp.cpp
```diff
@@ -340,6 +340,21 @@ bool Integer::isNegativeOne() const
 }
 
 Integer Integer::pow(unsigned long int exp) const
```
nafur, line 342: Please make it a `uint64_t`

src/util/integer_gmp_imp.cpp
```diff
@@ -340,6 +340,21 @@ bool Integer::isNegativeOne() const
...
+  // check if exp is within the uint32_t range
+  uint32_t low = static_cast<uint32_t>(exp); 
+  uint32_t high = static_cast<uint32_t>(exp << 32);
+  if (low == exp) {
+    // if it is, safely call the gmp pow function
+    return powHelper(exp);
+  } else {
+      // if exp is bigger than max uint32_t, use the following identity:
+      // a^(x + 2^32*y) = a^x * (a^(2^32))^y
+      return powHelper(low) * ((Integer(static_cast<uint32_t>(-1))) * (*this)).powHelper(high);
```
nafur, line 353: I don't think this is correct: it should have three calls to `powHelper()`, but only has two. Please also add one or two unit tests that actually exercise power-by-more-than-2^32 :)

src/util/integer_gmp_imp.cpp
yoni206: Also here? https://github.com/cvc5/cvc5/blob/85ab7cba7ebf1fa989777f074cb850949f861e93/src/util/integer_gmp_imp.h#L208
and here and here?
https://github.com/cvc5/cvc5/blob/master/src%2Futil%2Finteger_cln_imp.h#L210
https://github.com/cvc5/cvc5/blob/85ab7cba7ebf1fa989777f074cb850949f861e93/src/util/integer_cln_imp.cpp#L274

src/util/integer_gmp_imp.cpp
yoni206: fixed.
However, regarding the tests — they take too much time to be added as unit tests I believe. I tried also using max-unsigned-int + 1 and it was too slow.

src/util/integer_gmp_imp.cpp
nafur: yes. There, it should be purely cosmetic as it is directly converted to a `cl_I`.

src/util/integer_gmp_imp.cpp
```diff
@@ -340,6 +340,21 @@ bool Integer::isNegativeOne() const
...
+  // check if exp is within the uint32_t range
+  uint32_t low = static_cast<uint32_t>(exp);
+  uint32_t high = static_cast<uint32_t>(exp << 32);
+  if (low == exp) {
+    // if it is, safely call the gmp pow function
+    return powHelper(exp);
+  } else {
+      // if exp is bigger than max uint32_t, use the following identity:
+      // a^(x + 2^32*y) = a^x * (a^(2^32))^y
+      return powHelper(low) * (powHelper(static_cast<uint32_t>(-1))).powHelper(high);
```
nafur, line 353: I **think** it is still not right.
Note that the `-1` corresponds to `2^32 - 1`, which is why we need to have `(powHelper(-1) * (*this)).powHelper(high)`.

src/util/integer_gmp_imp.cpp
```diff
@@ -340,6 +341,38 @@ bool Integer::isNegativeOne() const
...
+      // (***) a^{b*c} = (a^b)^c
+      // Therefore,
+      // this^{exp} = this^{low + 2^{32}*high} = (**)
+      // this^{low}*this^{2^{32}*high} = (***)
+      // this^{low}*(this^{2^{32}})^{high} = (*)
+      // this^{low}*(this^{2^{31}}*2)^{high}
+      Integer powLow = powHelper(low);
+      uint32_t twoToThe31 = 1 << 31;
+      Integer twoToThe32 = Integer(2) * Integer(twoToThe31);
+      Integer twoToThe32ToTheHigh = twoToThe32.powHelper(high);
```
yoni206, line 369: This line yields the error: 
```
gmp: overflow in mpz type
```
even though `high` is a `uint32_t`.

yoni206:
@nafur ,
I have reformulated this in more detail in the code + comments, but am getting a gmp error. Please see here: 
https://github.com/cvc5/cvc5/pull/7852/files#diff-4d31e6a46dd3ae0c29a52423f649d053e4f84231da97da7390dbb2b3e4c10464R369

Do you have an idea what's the problem? All exponents are below 2^32.

nafur:
So apparently operations that produce very large results behave somewhat weird in some cases (https://stackoverflow.com/questions/13328409/overflow-handling-in-gmp-pow, https://gmplib.org/list-archives/gmp-discuss/2016-March/005951.html)
My suggestion would be to go back and only support `uint32_t` as exponents (also for cln, even if cln would support 64bit exponents properly).
@aniemetz

yoni206:
This PR is now updated according to the following solution:
1. not allowing uint64_t as an argument for `pow` neither in gmp nor in cln.
2. changing parts of the code that are affected by this change

In its current state, the PR does not introduce any new `conversion` warnings that are not already in `main`.

@aniemetz

src/theory/arith/nl/iand_solver.cpp
```diff
@@ -264,8 +264,9 @@ Node IAndSolver::sumBasedLemma(Node i)
   Assert(i.getKind() == IAND);
   Node x = i[0];
   Node y = i[1];
-  size_t bvsize = i.getOperator().getConst<IntAnd>().d_size;
-  uint64_t granularity = options().smt.BVAndIntegerGranularity;
+  uint32_t bvsize = i.getOperator().getConst<IntAnd>().d_size;
+  Assert(options().smt.BVAndIntegerGranularity <= 8);
+  uint32_t granularity = (uint32_t)options().smt.BVAndIntegerGranularity;
```
aniemetz, line 269: Use c++ casts:
```suggestion
  uint32_t granularity = static_cast<uint32_t>(options().smt.BVAndIntegerGranularity);
```
They are checked at compile time (c-style casts are not).

src/theory/arith/nl/iand_solver.cpp
```diff
@@ -279,7 +280,8 @@ Node IAndSolver::bitwiseLemma(Node i)
   Node y = i[1];
 
   unsigned bvsize = i.getOperator().getConst<IntAnd>().d_size;
-  uint64_t granularity = options().smt.BVAndIntegerGranularity;
+  Assert(options().smt.BVAndIntegerGranularity <= 8);
+  uint32_t granularity = (uint32_t)options().smt.BVAndIntegerGranularity;
```
aniemetz, line 284:
```suggestion
  uint32_t granularity = static_cast<uint32_t>(options().smt.BVAndIntegerGranularity);
```

aniemetz (commented):
I dislike that option `BVAndIntegerGranularity` has to be downcasted to uint32_t. You should define `UINT32_MAX` (or any other appropriate upper bound) as max value for this option.

Else, looks good except for some minor things.

aniemetz (approved):

PR #7855, opened 2021-12-22
Change default granularity of proofs to macro

ajreynol:
Also changes the name "off" to "macro" to be more consistent.


HanielB (approved):
LGTM

PR #7856, opened 2021-12-22
Fix int blaster

ajreynol:
The int blaster crashes on most examples I try, due to a mismatch on what is looked up vs. what is computed during the traversal.


yoni206 (commented):
Thanks for this (and also sorry for this…)

Would it be possible for you to add one of the crashing examples as a test?

ajreynol:
Done.

yoni206 (approved):

PR #7857, opened 2021-12-23
Add bag.member operator to theory of bags

mudathirmahgoub:
This PR adds the predicate bag.member to be analogous to predicate set.member. 
The PR is motivated by converting regressions for sets to bags, which avoids defining a predicate for each set type
```smt2
(define-fun bag.member ((e E) (B (Bag E))) Bool (>= (bag.count e B) 1))
```


ajreynol (approved):
LGTM

PR #7862, opened 2022-01-03
Execute `(reset)` command in parse-only mode

### LINKED ISSUE #7860: Assertion failure after reset
The symbol table in the parser is not reset by `(reset)`, which may cause assertion failures later on.

Here's an example SMT2 input that triggers this:
```
(set-logic ALL)
(reset)
(set-logic QF_UF)
(declare-sort U 0)
```

Then:
```
cvc5 --parse-only reset.smt2
```
will crash:
```
Fatal failure within void cvc5::context::CDHashMap<std::string, cvc5::api::Term>::insertAtContextLevelZero(const Key &, const Data &) [Key = std::string, Data = cvc5::api::Term, HashFcn = std::hash<std::string>] at Repos/cvc5/src/./context/cdhashmap.h:408
Check failure

 d_map.find(k) == d_map.end()

Abort
```

This happens with a regular cvc5 build (not debug) on the master branch: 
commit 2b441eab892151ad4588299175e2b82ce78c6fa1

4tXJ7f:
Fixes #7860.

This commit fixes an issue where `--parse-only` would not execute
`(reset)` commands, leading to issues with symbols being defined
multiple times.

### DIFF
--- a/src/main/command_executor.cpp
+++ b/src/main/command_executor.cpp
@@ -207,7 +207,8 @@ bool solverInvoke(api::Solver* solver,
   // commands. We invoke define-fun commands because they add function names
   // to the symbol table.
   if (solver->getOptionInfo("parse-only").boolValue()
-      && dynamic_cast<DefineFunctionCommand*>(cmd) == nullptr)
+      && dynamic_cast<DefineFunctionCommand*>(cmd) == nullptr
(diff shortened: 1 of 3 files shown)


abdoo8080 (approved):
LGTM! This PR also fixes a similar problem I was facing with dump tester 🙂

PR #7867, opened 2022-01-04
Properly set __file__ in python bindings

yoni206:
When running `make docs` after changing the documentation in `cvc5.pxi`, the `html` files are not updated. This PR fixes this issue.

Co-authored-by: Gereon Kremer gereon.kremer@cs.rwth-aachen.de


nafur:
I'm pretty sure that this is not the right place to solve the issue. Note that changing `cvc5.pxi` should rebuild the python module, which does not happen. The way `add_cython_target` is set up does not account for input files other than `pycvc5.pyx`. I'm not sure how to fix this right now, but I'll have a look as well...

nafur:
I've also looked once again at this issue, as I really don't like the route of this PR... it essentially removes the sphinx cache all the time. Meanwhile sphinx does proper dependency tracking and autodoc tries pretty hard to figure out where a particular piece of documentation (or rather the class, function, etc) lives. Unfortunately the way we set up our module, everything always points to the `__init__.py` while it actually lives in the `.so` file.
This simple fact also directly points to the proper solution (though it took me quite some time and digging through autodoc code to realize): if we simply add `__file__ = pycvc5.__file__` to the `__init__.py`, everything works as desired (at least for me...)

yoni206:
Thanks a lot for looking into this! Your suggested solution works for me as well. Pushed.

nafur (approved):
LGTM.

PR #7871, opened 2022-01-04
Remove spurious call to applySubs

ajreynol:
This call is preceded by another call to `apply-subs` at the beginning of preprocessing.

This is motivated by very large input benchmarks from Cetora where this preprocessing pass can take up to 5 seconds.


4tXJ7f (approved):
LGTM

PR #7876, opened 2022-01-04
Properly parse arithmetic values

ajreynol:
This makes our parser correct for arithmetic values e.g. `(- n)` is parsed as a value not UMINUS when `n` is a numeral, as indicated in SMT-LIB. Similarly for `(/ m n)` and `(/ (- m) n)` we parse as values and not DIVISION.

This subsumes a patch for constructing array constants where this led to issues.


src/parser/smt2/smt2.cpp
```diff
@@ -1138,10 +1103,31 @@ api::Term Smt2::applyParseOp(ParseOp& p, std::vector<api::Term>& args)
     }
     else if (kind == api::MINUS && args.size() == 1)
     {
+      if (isConstInt(args[0]) && args[0].getRealOrIntegerValueSign() > 0)
```
nafur, line 1106: So `(- 0)` would still yield `(UMINUS 0)`? Can the sign actually be negative here?

ajreynol: Yes and yes.
(- 0) would be (UMINUS CONST_RATIONAL{0}) and (- (- 6)) would be (UMINUS CONST_RATIONAL{-6}).  This is expected since these are not SMT-LIB syntax for values.

nafur (approved):
LGTM.

abdoo8080 (approved):

PR #7884, opened 2022-01-06
Enable dump tester.

abdoo8080:
This PR enables the dump tester, a `parse-print-parse` test for cvc5, on CI. It also includes some changes that reduce the number of regressions that fail with this tester.


test/regress/regress0/fp/issue-5524.smt2
```diff
@@ -2,5 +2,5 @@
 ; EXPECT: unsat
 (set-logic QF_FP)
 (declare-fun fpv5 () Float32)
-(assert (fp.eq (fp.mul RTP fpv5 fpv5) ((_ to_fp 8 24) RTN 0.04)))
+(assert (fp.eq (fp.mul RTP fpv5 fpv5) ((_ to_fp 8 24) #x3d23d70a)))
```
ajreynol, line 5: Is this change necessary? I'd rather disable the dump-tester for this benchmark instead of changing this.  Or I thought you mentioned that changing the logic to e.g. `QF_FPLRA` ?

abdoo8080: I thought we didn't want to modify the logic. Reverted the changes and added `LRA` to the 2 FP regressions.

ajreynol: Ok, yes, I think changing the logic to make it SMT-LIB compliant is the better fix here.

ajreynol:
This LGTM although I'd like @4tXJ7f to also have a look.

test/regress/run_regression.py
```diff
@@ -44,13 +44,13 @@ def __init__(self):
     def applies(self, benchmark_info):
         return True
 
-    def run(self, benchmark_info):
+    def run(self, benchmark_info, compare_outputs=True):
```
4tXJ7f, line 47: It would be more consistent with the rest of the code to make `compare_outputs` part of `benchmark_info` IMHO. What do you think?

test/regress/run_regression.py
```diff
@@ -245,12 +245,16 @@ def run(self, benchmark_info):
 
 class DumpTester(Tester):
     def applies(self, benchmark_info):
-        return benchmark_info.expected_exit_status == EXIT_OK
+        return (
+            benchmark_info.benchmark_ext != ".p"
+            and benchmark_info.expected_exit_status == EXIT_OK
+            and "Parse Error" not in benchmark_info.expected_output
```
4tXJ7f, line 251: Checking for `"Parse Error"` in the expected output seems rather fragile. Can we instead explicitly disable all of those tests?

You should be able to do something like:

```
 grep -lr "Parse Error" test/regress/regress[0-4] | xargs sed -i '1 i\; DISABLE-TESTER: dump'
```

4tXJ7f (changes requested):
Awesome, I just have some minor comments.

test/regress/run_regression.py
abdoo8080: I agree. Though, I wonder if we should rename `benchmark_info` to `benchamrk_config` (or something else) as `comapre_outputs` is specific to the testers and not part of the benchmark.

4tXJ7f (approved):
LGTM, thanks for the updates!

test/regress/run_regression.py
4tXJ7f: Yes, `benchmark_config` sounds pretty good. We can do that in an upcoming PR.

Issue #7870, opened 2022-01-04
Feature request: make get-qe results strictly typed


daniel-larraz:
### Discussed in https://github.com/cvc5/cvc5/discussions/7859

<div type='discussions-op-text'>

<sup>Originally posted by **daniel-larraz** December 23, 2021</sup>
Hello. Consider this dummy example:
```
(set-logic LRA)  
(declare-fun x () Real)
(get-qe
 (exists ((Y Bool)) (>= (* (- 1.0) x) (- 100.0))
))
```
Currently CVC5 returns:
`(>= (* (- 1) x) (- 100))`
But I would like CVC5 to return:
`(>= (* (- 1.0) x) (- 100.0))`
Or a term that uses rational numbers, but not numerals.
This would make the parsing and type inference of the expressions occurring in the resulting term easier for external tools that read the result, in particular, when other theories like integers are involved.
Thanks!</div>

PR #7885, opened 2022-01-06
Eliminate use of subtyping from results of quantifier elimination

ajreynol:
Fixes https://github.com/cvc5/cvc5/issues/7870.

Notice this node converter is potentially unnecessary if we ever get rid of arithmetic subtyping.


HanielB:
Since this is for QE the idea is to use only for pure arith, right?

Another thing: do you want to be robust to something like `(= (+ (* 2 2) 1) 5.0)`? To handle this case I had to do some more intense stuff (currently only in proof-new: https://github.com/cvc5/cvc5/blob/c6f21c4dc887219e7f1197d6ffa2b1061dd89e56/src/proof/alethe/alethe_nosubtype_node_converter.cpp#L116)

ajreynol:
QE can be used in any logic.  Indeed, the code is not robust to cases like the one you mention.  Although this is a relatively minor issue I think, you'd get `(= (to_real (+ (* 2 2) 1)) 5.0)` instead of `(= (+ (* 2.0 2.0) 1.0) 5.0)`.

HanielB:
Ah, agreed. I had missed that you have the different kind of treatment for const vs non-const.

HanielB (approved):

PR #7890, opened 2022-01-06
Remove CDDenseSet data structure

matthewsot:
Submitting per discussion with Andres. `CDDenseSet` was not used anywhere, and does not appear to build under, at least, GCC/9.1.0.


4tXJ7f (approved):
LGTM, nice catch :)

PR #7892, opened 2022-01-07
[Regressions] Add directive for disabling testers

4tXJ7f:
This commit adds a directive for regressions (`DISABLE-TESTER`) that
allows disabling a given tester for a regression. Currently, we disable
testers based on properties such as command line arguments, which can be
error prone.


abdoo8080 (approved):
LGTM!

PR #7899, opened 2022-01-07
Guard use of unsat core mode pp-only

ajreynol:
Fixes https://github.com/cvc5/cvc5-projects/issues/366. Now gives
`(error "Error in option parsing: Unsat core mode pp-only is for internal use only.")`


mpreiner (approved):

src/smt/set_defaults.cpp
```diff
@@ -62,6 +62,12 @@ void SetDefaults::setDefaults(LogicInfo& logic, Options& opts)
 
 void SetDefaults::setDefaultsPre(Options& opts)
 {
+  // internal-only options
+  if (opts.smt.unsatCoresMode == options::UnsatCoresMode::PP_ONLY)
+  {
+    throw OptionException(
+        std::string("Unsat core mode pp-only is for internal use only."));
```
mpreiner, line 69: What would probably be better is to use the string representation of `options::UnsatCoresMode::PP_ONLY` instead of the `pp-only` string.

ajreynol: Aren't they the same? `pp-only` is the what string is used in `set-option`.

PR #7900, opened 2022-01-07
Improve docs extension for examples

nafur:
We recently added automatic download links for examples in our documentation, however they assumed everything to live inside the main `cvc5` repo while, e.g., the examples for the z3py compat API do not. While #7896 provided a very simple fix, it mixes different concerns (rendering of tabs vs. locating the input files, though they should coincide right now).
This PR takes a more fundamental approach by extending the current pattern mechanism to allow for explicit url specification.

Closes #7896.


alex-ozdemir (approved):
Very nice!

PR #7901, opened 2022-01-08
Fix docs upload for tags

nafur:
This PR fixes a subtle issue with the docs upload for releases. While most of the context of the original action is passed into the docs upload action, the git ref changes from `refs/tags/...` to `refs/heads/...`. However, we still can check the current ref against the `head_branch` of the original workflow.


nafur:
Check https://github.com/nafur/cvc5-versioning/runs/4745614714?check_suite_focus=true for what is in the context...

mpreiner (approved):

PR #7904, opened 2022-01-10
Document substitute in API

ajreynol:
@mudathirmahgoub substitute doesnt appear to be in the Java API?


mudathirmahgoub:
> @mudathirmahgoub substitute doesnt appear to be in the Java API?
Perhaps you meant `Sort.substitute` method which is located in `Sort.java`
https://github.com/cvc5/cvc5/blob/master/src/api/java/io/github/cvc5/api/Sort.java#L470

ajreynol:
Ah yes, sorry, was looking in the wrong place.

src/api/cpp/cvc5.h
```diff
@@ -596,13 +596,21 @@ class CVC5_EXPORT Sort
    * Substitution of Sorts.
    * @param sort the subsort to be substituted within this sort.
    * @param replacement the sort replacing the substituted subsort.
+   *
+   * Note that this replacement is applied during a pre-order traversal and
+   * only once to the sort. It is not run until fix point.
```
aniemetz, line 601: Should we maybe add the example from our discussion?
```
(Array A B) .substitute ( A -> (Array C D), C -> (Array A B) ) will return (Array (Array C D) B)
```

aniemetz (approved):

PR #7905, opened 2022-01-10
[API] Add missing arity check

### LINKED ISSUE #7894: Fatal failure at src/expr/node_builder.cpp:433
Commit: [a2f099d](https://github.com/cvc5/cvc5/commit/a2f099d72f65bbfe4224bc96ed16c1b423774760)
OS: Ubuntu 20.04

```
$ cvc5 -q delta.smt2
Fatal failure within cvc5::expr::NodeValue* cvc5::NodeBuilder::constructNV() at /cvc5/src/expr/node_builder.cpp:433
Check failure

 getNumChildren() >= kind::metakind::getMinArityForKind(getKind())
Nodes with kind PLUS must have at least 2 children (the one under construction has 1)
Aborted (core dumped)
$ cat delta.smt2
(assert (= (+ 0)))
```

4tXJ7f:
Fixes #7894. The API was not performing an arity check for associative
operators with less than two children. This commit adds the missing
check.


aniemetz (changes requested):
I have an API trace that produces the same error but is not fixed with your changes. Will post and link to here.

aniemetz (commented):
OK, so the error I found is in `Sort::substitute` and unrelated: cvc5/cvc5-projects#399

aniemetz (approved):

PR #7909, opened 2022-01-10
Fix internal type error when printing lambdas with more than one variable in LFSC

ajreynol:


abdoo8080 (commented):
LGTM. This error is somewhat obscure. Can you add a regression?

abdoo8080 (approved):

ajreynol:
We don't have LFSC proofs in our regressions, but I suppose I can add a minimized version of the benchmark you sent.

ajreynol:
Actually, while the error is fairly obscure in how it occurs, it will trigger with any problem involving a define-fun with more than one argument, this is the minimized example:
```
(define-fun __ ((add_two.impl.usr.v@0 Real) (add_two.impl.usr.P@0 Bool)) Bool false)
(assert false)
(check-sat)
```
which is probably not worth adding :)

PR #7912, opened 2022-01-10
Add new idiomatic examples

alex-ozdemir:


docs/api/python/z3compat/z3compat.rst
```diff
@@ -7,6 +7,24 @@ z3py compatibility Python API
...
+
+This API is missing some features from cvc5 and Z3Py.
+
+It does not (currently) support these cvc5 features:
+* The theories of strings and sequences
+* unsatisfiable cores
+* syntax-guided synthesis (SyGuS)
+
+It does not support the following features of Z3Py:
+* Patterns for quantifier instantiation
```
nafur, line 19:
```suggestion
It does not (currently) support these cvc5 features:

* The theories of strings and sequences
* unsatisfiable cores
* syntax-guided synthesis (SyGuS)

It does not support the following features of Z3Py:

* Patterns for quantifier instantiation
```

docs/examples/floatingpoint.rst
```diff
@@ -4,4 +4,5 @@ Theory of Floating-Points
 
 .. api-examples::
     <examples>/api/java/FloatingPointArith.java
+    <z3pycompat>/test/pgms/example_floatingpoint.py
```
nafur, line 7: There is something wrong here...

alex-ozdemir: Good catch. Yes. It should be fixed in c60228f

PR #7914, opened 2022-01-11
api: Fix formatting of docs for Term::getSetValue().

aniemetz:


docs/ext/smtliblexer.py
```diff
@@ -100,6 +100,8 @@ class SmtLibLexer(RegexLexer):
             # parentheses
             (r'\(', token.Text),
             (r'\)', token.Text),
+            (r'\{', token.Text),
+            (r'\}', token.Text),
```
nafur, line 104: You wanted to fix the highlighting of the `c_{n-1}`, right? For some reason, it still doesn't work. Maybe we can simply include them in the regular symbols?
Still there is something weird, though: why is `-1` highlighted differently and not put together with the `n`?

aniemetz: No, I wanted that it doesn't surround `{` with a red box, which did work locally for me.

aniemetz: The generated docs on `docs-ci` look good, too. https://cvc5.github.io/docs-ci/docs-pr7914/api/cpp/term.html#_CPPv4NK4cvc53api4Term10isSetValueEv

nafur (approved):
LGTM.

Issue #7924, opened 2022-01-12
(arith-no-partial)Refutation soundness issue in NRA


merlinsun:
Hi, for this formula, cvc5 gives `sat`, while gives `unsat` with `--arith-no-partial-fun`.

```
$ cvc5 -q delta.smt2
sat
$ cvc5 -q --arith-no-partial-fun delta.smt2
unsat
$ cat delta.smt2
(assert (exists ((V Real)) (distinct (sqrt 1.0) (sqrt 0.0))))
(check-sat)
```

Commit: [69c88d4](https://github.com/cvc5/cvc5/commit/69c88d4cb47b00b807281771bf0e7637f3dda37b)
OS: Ubuntu 20.04

ajreynol:
`--arith-no-partial-fun` changes the semantics of cvc5's arithmetic theory.  This option is highly experimental (although indeed this case should not be unsat).

Please follow the guidelines here for fuzzing cvc5: https://github.com/cvc5/cvc5/wiki/Fuzzing-cvc5

In particular, do not report bugs for experimental options.

merlinsun:
Thanks, @ajreynol !
I will not report this kind of issue.

PR #7926, opened 2022-01-12
Always use partial function for sqrt

ajreynol:
Fixes the behavior reported in https://github.com/cvc5/cvc5/issues/7924.


nafur (commented):
Can you add a regression?

nafur (approved):
LGTM.

PR #7931, opened 2022-01-12
Refactor post rewriter for addition

nafur:
This makes the post rewriter for addition aware of real algebraic numbers. Also, it tries a bit harder to avoid the `normal_form` utilities if only constants are present.


src/theory/arith/arith_rewriter.cpp
```diff
@@ -346,27 +347,64 @@ RewriteResponse ArithRewriter::preRewritePlus(TNode t){
 RewriteResponse ArithRewriter::postRewritePlus(TNode t){
   Assert(t.getKind() == kind::PLUS);
 
+  if (t.getNumChildren() == 1)
+  {
+    return RewriteResponse(REWRITE_DONE, t[0]);
+  }
```
4tXJ7f, line 353: [It looks like this shouldn't be possible.](https://github.com/cvc5/cvc5/blob/c9daef3594dec48693ede0c13e7fc9804835ef14/src/theory/arith/kinds#L15)

src/theory/arith/arith_rewriter.cpp
```diff
@@ -346,27 +347,64 @@ RewriteResponse ArithRewriter::preRewritePlus(TNode t){
...
   }
 
   if(!monomials.empty()){
     Monomial::sort(monomials);
     Monomial::combineAdjacentMonomials(monomials);
-    polynomials.push_back(Polynomial::mkPolynomial(monomials));
+    polynomials.emplace_back(Polynomial::mkPolynomial(monomials));
   }
+  polynomials.emplace_back(
+      Polynomial::mkPolynomial(Constant::mkConstant(rational)));
```
4tXJ7f, line 390: Shouldn't this only be done if `rational` is non-zero? Probably doesn't matter too much, but still.

src/theory/arith/arith_rewriter.cpp
```diff
@@ -346,27 +347,64 @@ RewriteResponse ArithRewriter::preRewritePlus(TNode t){
...
+  auto* nm = NodeManager::currentNM();
+  if (poly.isConstant())
+  {
+    ran += RealAlgebraicNumber(poly.getHead().getConstant().getValue());
+    return RewriteResponse(REWRITE_DONE, nm->mkRealAlgebraicNumber(ran));
+  }
+  return RewriteResponse(
+      REWRITE_DONE,
+      nm->mkNode(
+          Kind::PLUS, nm->mkRealAlgebraicNumber(ran), poly.getNode()));
```
4tXJ7f, line 407: Is it safe to do `REWRITE_DONE` here? Couldn't it happen that this gets flattened later on?

4tXJ7f (changes requested):
Some questions.

src/theory/arith/arith_rewriter.cpp
nafur: Good catch.

src/theory/arith/arith_rewriter.cpp
nafur: true

src/theory/arith/arith_rewriter.cpp
nafur: At this point the poly is not constant and will thus not further rewrite. The only thing that one can further do is merge the constant from poly (if it has one) into the ran. I've added this.

4tXJ7f (approved):
LGTM

PR #7932, opened 2022-01-12
Refactor rewriteMinus

nafur:
Refactors `rewriteMinus` to be generally simpler and explicitly rely on rewriting addition.
Note that `rewriteMinus` would almost never be called in post rewrite anyway, as it would rewrite to addition in pre rewrite.

### DIFF
--- a/src/theory/arith/arith_rewriter.cpp
+++ b/src/theory/arith/arith_rewriter.cpp
@@ -136,23 +136,22 @@ RewriteResponse ArithRewriter::preRewriteTerm(TNode t){
           if (rat >= 0)
           {
             return RewriteResponse(REWRITE_DONE, t[0]);
-          } else {
-          return RewriteResponse(
-              REWRITE_DONE,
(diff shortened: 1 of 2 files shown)


src/theory/arith/arith_rewriter.cpp
```diff
@@ -106,44 +102,41 @@ RewriteResponse ArithRewriter::preRewriteTerm(TNode t){
...
+      case kind::INTS_DIVISION_TOTAL:
+      case kind::INTS_MODULUS_TOTAL: return rewriteIntsDivModTotal(t, true);
+      case kind::ABS:
+        if (t[0].isConst())
+        {
+          const Rational& rat = t[0].getConst<Rational>();
+          if (rat >= 0)
+          {
+            return RewriteResponse(REWRITE_DONE, t[0]);
+          } else {
```
4tXJ7f, line 139: Formatting seems off past this line

4tXJ7f (approved):
Just the formatting looks a bit off.

PR #7935, opened 2022-01-12
Refactor abs rewriting

nafur:
This PR refactors rewriting for `ABS` to also support real algebraic number. We also generalize the `ABS` operator to real arguments in general, instead of integer arguments.


ajreynol:
Don't we allow ABS for real as well?

nafur:
> Don't we allow ABS for real as well?

No, the type rule requires integers.

nafur:
> Hm I'm surprised by this, shouldn't we support ABS for real?

I'd argue that we should, in particular as we only need to remove the assertion: the code already uses `mkConstRealOrInt` properly, I think. One general question about `ABS`, though: is there **any** solving technique for this?

ajreynol:
ABS is just syntax sugar, it should just be eliminated during preprocessing/rewriting.

nafur:
> ABS is just syntax sugar, it should just be eliminated during preprocessing/rewriting.

The rewriter should now support ABS for both integer and real, rational and RAN numbers.

ajreynol (approved):

PR #7939, opened 2022-01-13
Use proper RAN nodes for nl model

nafur:
This PR finally uses proper `RealAlgebraicNumber` (wrapped in nodes) to represent nonlinear model values. This simplifies the implementation of `getEqualityStatus`, and should make theory combination much more robust.
Also, we reintroduce the possibility to have `getEqualityStatus` return unknown, though this should only happen in the presence of witness terms.

Requires the PRs concerning rewriting of RANs to be merged first.


ajreynol (approved):
LGTM, don't forget to format.

Issue #7917, opened 2022-01-11
(sygus-rr-synth-input) Fatal failure at /cvc5/src/theory/evaluator.cpp:179


merlinsun:
Commit: [8c18b39](8c18b3962679ead1c66cfa8118ec93176d8860b6)
OS: Ubuntu 20.04

```
$ cvc5 -q delta.smt2
sat
$ cvc5 -q --sygus-rr-synth-input delta.smt2
(define-fun f1 ((A Int) (B Int) (C Int) (D Real) (E Real) (F Real)) Bool true)
(define-fun f2 ((A Int) (B Int) (C Int) (D Real) (E Real) (F Real)) Int A)
(define-fun f3 ((A Int) (B Int) (C Int) (D Real) (E Real) (F Real)) Real D)
(define-fun f1 ((A Int) (B Int) (C Int) (D Real) (E Real) (F Real)) Bool false)
(define-fun f2 ((A Int) (B Int) (C Int) (D Real) (E Real) (F Real)) Int B)
(define-fun f3 ((A Int) (B Int) (C Int) (D Real) (E Real) (F Real)) Real E)
(define-fun f2 ((A Int) (B Int) (C Int) (D Real) (E Real) (F Real)) Int C)
(define-fun f3 ((A Int) (B Int) (C Int) (D Real) (E Real) (F Real)) Real F)
(define-fun f1 ((A Int) (B Int) (C Int) (D Real) (E Real) (F Real)) Bool (= D A))
(define-fun f2 ((A Int) (B Int) (C Int) (D Real) (E Real) (F Real)) Int 0)
(define-fun f1 ((A Int) (B Int) (C Int) (D Real) (E Real) (F Real)) Bool (= D B))
Fatal failure within cvc5::Node cvc5::theory::Evaluator::eval(cvc5::TNode, const std::vector<cvc5::NodeTemplate<true> >&, const std::vector<cvc5::NodeTemplate<true> >&, const std::unordered_map<cvc5::NodeTemplate<true>, cvc5::NodeTemplate<true> >&) const at /cvc5/src/theory/evaluator.cpp:179
Check failure

 itn != evalAsNode.end()

Aborted (core dumped)
$cat delta.smt2
(assert (exists ((R Real)) (= 0.0 (/ 0 0.0))))
(check-sat)
```

merlinsun:
[...]

PR #7942, opened 2022-01-13
Fix bug in evaluator for division by zero

ajreynol:
Fixes https://github.com/cvc5/cvc5/issues/7917.


nafur (approved):
LGTM. Can you add the regression?

ajreynol:
The regression involves an experimental option and has output that is not repeatable and is non-terminating.  I can't seem to craft another example that triggers this.

PR #7944, opened 2022-01-14
Implement -o subs to show learned top-level substitutions

ajreynol:
Also adds substitutions as part of the output of `-o learned-lits` for completeness.


src/preprocessing/preprocessing_pass_context.h
```diff
@@ -79,7 +79,11 @@ class PreprocessingPassContext : protected EnvObj
   /** Spend resource in the resource manager of the associated Env. */
   void spendResource(Resource r);
 
-  /** Get a reference to the top-level substitution map */
+  /**
+   * Get a reference to the top-level substitution map. Note that all
+   * substitutions added to this map should use the addSubstitution methods
+   * below for the purposes of proper debugging information.
+   */
```
nafur, line 86: Given this description, I would expect this method to return a `const &`. I'm aware that this is not trivial to do, but I think we should consider aiming for it...

src/theory/substitutions.cpp
```diff
@@ -31,6 +31,18 @@ SubstitutionMap::SubstitutionMap(context::Context* context)
...
 
+std::unordered_map<Node, Node> SubstitutionMap::getSubstitutions()
+{
+  std::unordered_map<Node, Node> subs;
+  for (NodeMap::const_iterator it = d_substitutions.begin(), it_end = d_substitutions.end();
+       it != it_end;
+       ++it)
+  {
+    subs[(*it).first] = (*it).second;
+  }
```
nafur, line 42:
```suggestion
  for (const auto& sub: d_substitutions)
  {
    subs.emplace(sub.first, sub.second);
  }
```

src/preprocessing/preprocessing_pass_context.cpp
```diff
@@ -85,15 +86,46 @@ void PreprocessingPassContext::addSubstitution(const Node& lhs,
...
                                                ProofGenerator* pg)
 {
-  getTopLevelSubstitutions().addSubstitution(lhs, rhs, pg);
+  if (isOutputOn(OutputTag::LEARNED_LITS) || isOutputOn(OutputTag::SUBS))
+  {
+    Node eq = lhs.eqNode(rhs);
+    output(OutputTag::LEARNED_LITS)
+        << "(learned-lit " << eq << " :preprocess-subs)" << std::endl;
+    output(OutputTag::SUBS)  << "(substitution " << eq << ")" << std::endl;
+  }
```
nafur, line 95: I'd prefer having this in a utility function.

src/preprocessing/preprocessing_pass_context.h
ajreynol: It is hard to do this (I abandoned an attempt on this branch).  This is because it keeps a cache.

nafur: Yes, I know. I think we should consider making the cache in `SubstitutionMap` mutable, and then make `SubstitutionMap::apply()` const. But this is not in scope for this PR...

nafur (approved):
LGTM.

PR #7947, opened 2022-01-14
Clean enumerative instantiation options

ajreynol:
`--full-saturate-quant` is kept as a generic option, and implies `--enum-inst`.


src/options/quantifiers_options.toml
```diff
@@ -505,44 +505,52 @@ name   = "Quantifiers"
   long       = "full-saturate-quant"
   type       = "bool"
   default    = "false"
+  help       = "resort to full effort techniques instead of answering unknown due to quantifiers. Currently enables enumerative instantiation"
```
HanielB, line 508:
```suggestion
  help       = "resort to full effort techniques instead of answering unknown due to limited quantifier reasoning. Currently enables enumerative instantiation"
```

HanielB (approved):
LGTM, just a minor suggestion

PR #7949, opened 2022-01-14
Refactor options related to rewriting and symmetry breaking in sygus

ajreynol:
Note this also corrects a few inconsistencies in how the rewriter is invoked in fast / variable agnostic enumeration.

Notably, fast enumeration now leverages the term datatype sygus for rewriting, meaning it can evaluate recursive functions for enumerated terms.


src/options/datatypes_options.toml
```diff
@@ -86,28 +86,40 @@ name   = "Datatypes Theory"
...
-  name       = "sygusSymBreakDynamic"
-  category   = "regular"
-  long       = "sygus-sym-break-dynamic"
-  type       = "bool"
-  default    = "true"
-  help       = "dynamic sygus symmetry breaking lemmas"
+  long       = "sygus-simple-sym-break=MODE"
+  type       = "SygusSimpleSymBreakMode"
+  default    = "AGG"
+  help       = "if and how to apply simple symmetry breaking for smart enumeration"
```
HanielB, line 94: Why is it that we call this "simple" again? It's somewhat cryptic. Maybe we could change the name at this point?

ajreynol: We called it simple symmetry breaking in our CAV draft although this didnt make it to the final version.  I think it is the right name for it, it is just simple custom symmetry breaking lemmas that are added statically.   Its also the name of the module: https://github.com/cvc5/cvc5/blob/master/src/theory/datatypes/sygus_simple_sym.h#L28.

HanielB: I see. Indeed I had looked into the CAV paper and not found it, so I was confused. Maybe we could then add a simple qualifier in the helper, like "(based on grammar/conjecture structure)"?

ajreynol: Its actually only based on the grammar.

HanielB (approved):

PR #7951, opened 2022-01-14
Preprare central model building for RANs

nafur:
This PR uses a new utility `isEvaluationResult()` instead of `isConst()` within the core model building. This prepares model building for model values that are not constants (in the sense of `isConst()`) but still constant-lilke values, like real algebraic numbers.


ajreynol:
Come to think of it, can we replace this method https://github.com/cvc5/cvc5/blob/master/src/theory/theory_model_builder.cpp#L128 by the one you introduce in this PR?  Note it would simplify that assertion in SolverEngine.

ajreynol (approved):

nafur:
> Come to think of it, can we replace this method https://github.com/cvc5/cvc5/blob/master/src/theory/theory_model_builder.cpp#L128 by the one you introduce in this PR? Note it would simplify that assertion in SolverEngine.

Right, they essentially do the same thing. I still moved it to `TheoryModel` so that `SolverEngine` can easily access it.

PR #7954, opened 2022-01-14
Add inverse inference for update-over-concat

ajreynol:
This adds an inference that applies when an atomic update term becomes equal to a concatenation term.

This fixes bogus models discovered when writing the model soundness proof for the sequences array solver.


4tXJ7f (approved):
LGTM

PR #7957, opened 2022-01-15
Fix CMake script for static, auto-download, cln configuration

matthewsot:
The script was previously creating a 'CLN_STATIC_LIBRARIES' variable that
wasn't known about elsewhere in the code and caused issues when the Makefiles
depended on a CLN_LIBRARIES variable. Instead, the static libraries should just
go into CLN_LIBRARIES as normal.

This only seems to have affected the case where cln is set to auto-download and
the build is static.

Signed-off-by: Matthew Sotoudeh <sotoudeh@stanford.edu>


matthewsot:
I should clarify my use of "should" above really means "I did this and it seemed to work"...I don't have a great grasp on how all the cmake files work together so feedback appreciated!

nafur (approved):
Good catch. This is a leftover from when we tried to build shared and static at once.

PR #7958, opened 2022-01-16
api: Add explicit guard for option produce-assertions in blockModelValues().

aniemetz:
Fixes cvc5/cvc5-projects#412.


ajreynol:
I don't think that blockModelValues always requires assertions, I think there are modes that don't require assertions, e.g. `block-models=values`.

aniemetz:
So which modes always require assertions then?

ajreynol:
I looked into this, it's possible to make the values option not require produce-assertions, although I don't think it is worthwhile to do so.  Thus, I'm fine with guarding this unconditionally.

ajreynol (approved):

PR #7961, opened 2022-01-18
Add utilities for flattening nodes

nafur:
This PR adds new utilities for flattening nodes and checking whether they can be flattened. They replace the custom implementation we used in the arithmetic rewriter.


src/theory/arith/arith_rewriter.cpp
```diff
@@ -56,6 +57,62 @@ bool evaluateRelation(Kind rel, const L& l, const R& r)
...
+  bool canFlatten = std::any_of(
+      t.begin(), t.end(), [k](TNode child) { return child.getKind() == k; });
+  if (!canFlatten)
+  {
+    return t;
+  }
+  std::vector<TNode> children;
+  flatten(t, children);
+  Assert(children.size() >= 2);
+  return NodeManager::currentNM()->mkNode(t.getKind(), std::move(children));
```
4tXJ7f, line 113: I don't think that the `std::move` is necessary here. `children` is passed by reference.

src/theory/arith/arith_rewriter.cpp
```diff
@@ -56,6 +57,62 @@ bool evaluateRelation(Kind rel, const L& l, const R& r)
...
 
+/** Flatten the given node (with child nodes of the same kind) into a vector */
+void flatten(TNode t, std::vector<TNode>& children)
+{
+  Kind k = t.getKind();
+  for (const auto& child : t)
+  {
+    if (child.getKind() == k)
+    {
+      flatten(child, children);
```
4tXJ7f, line 68: Let's implement this in an iterative fashion instead of recursively. Also, I think that it would be good to just implement this in `node_algorithm.h` as a generic method.

src/theory/arith/arith_rewriter.cpp
```diff
@@ -56,6 +57,62 @@ bool evaluateRelation(Kind rel, const L& l, const R& r)
...
+                              Kind k1,
+                              Kind k2,
+                              std::vector<TNode>& children)
+{
+  Assert(t.getKind() == k1 || t.getKind() == k2);
+  for (const auto& child : t)
+  {
+    if (child.getKind() == k1 || child.getKind() == k2)
+    {
+      flatten(child, k1, k2, children);
```
4tXJ7f, line 91: See comment above.

4tXJ7f (changes requested):
I think that it would be better if the flatten methods were generic utilities that we could reuse in other places and implemented iteratively.

src/theory/arith/arith_rewriter.cpp
nafur: Hm, indeed. We should consider adding an appropriate overload...

src/expr/algorithms/flatten.h
```diff
@@ -0,0 +1,121 @@
...
+#include "cvc5_private.h"
+
+#ifndef CVC5__EXPR__ALGORITHMS__FLATTEN_H
+#define CVC5__EXPR__ALGORITHMS__FLATTEN_H
+
+#include <algorithm>
+
+#include "expr/node.h"
+
+namespace cvc5::expr {
```
4tXJ7f, line 25: I think that it would be a bit more consistent to move this to the `cvc5::expr::algorithms` namespace. I also have a slight preference for `algorithm` over `algorithms`. What do you think?

```suggestion
namespace cvc5::expr::algorithms {
```

4tXJ7f (approved):
Minor suggestion to consider.

PR #7962, opened 2022-01-18
Refactor how arith rewriting checks for mult-by-zero

nafur:
This adds an explicit check for a zero factor in the post rewriter for multiplication.


src/theory/arith/arith_rewriter.cpp
```diff
@@ -563,10 +566,27 @@ RewriteResponse ArithRewriter::postRewritePlus(TNode t){
       nm->mkNode(Kind::PLUS, nm->mkRealAlgebraicNumber(ran), poly.getNode()));
 }
 
+RewriteResponse ArithRewriter::preRewriteMult(TNode node)
+{
+  Assert(node.getKind() == kind::MULT
+         || node.getKind() == kind::NONLINEAR_MULT);
+
+  if (auto res = getZeroChild(node); res)
```
4tXJ7f, line 574: In this case (and the case below), this syntax seems a bit overkill IMHO.

4tXJ7f (approved):
LGTM. Marking it as "do not merge" to give a chance to reconsider the syntax used.

PR #7964, opened 2022-01-19
Consider RANs in variable ordering

nafur:
The variable ordering in the arithmetic normal form should also consider real algebraic numbers. This PR also adds new comparators (essentially doing the same) in the rewriter that will replace the normal form utility classes.


src/theory/arith/arith_rewriter.cpp
```diff
@@ -42,6 +42,88 @@ namespace arith {
...
+    if (aIsVar != bIsVar) return aIsVar;
+
+    return a < b;
+  }
+};
+
+/**
+ * Implements an ordering on arithmetic nonlinear multiplications. As we assume
+ * rationals to be handled separately, we only consider Kind::NONLINEAR_MULT as
+ * multiplication term. For individual factors of the product, we rely on the
```
4tXJ7f, line 84:
```suggestion
 * multiplication terms. For individual factors of the product, we rely on the
```

src/theory/arith/arith_rewriter.cpp
```diff
@@ -42,6 +42,88 @@ namespace arith {
...
+    if (a == b) return false;
+
+    Assert(a.getKind() != Kind::MULT);
+    Assert(b.getKind() != Kind::MULT);
+
+    bool aIsMult = a.getKind() == Kind::NONLINEAR_MULT;
+    bool bIsMult = b.getKind() == Kind::NONLINEAR_MULT;
+    if (aIsMult != bIsMult) return !aIsMult;
+
+    if (!aIsMult)
```
4tXJ7f, line 105: Shouldn't this be:

```suggestion
    if (!aIsMult && !bIsMult)
```

4tXJ7f (changes requested):
Looks mostly good to me, but I have a question.

src/theory/arith/arith_rewriter.cpp
nafur: As this point, we know that `aIsMult == bIsMult`, so it's sufficient to only check one.

ajreynol (approved):

src/theory/arith/arith_rewriter.cpp
4tXJ7f: Ah, fair enough.

4tXJ7f (approved):
LGTM, thanks for the explanation.

PR #7965, opened 2022-01-19
Refactor multiplication in arithmetic rewriter

nafur:
This PR refactors the rewriting of multiplication. Most importantly, we explicitly deal with distributivity of addition and multiplication explicitly using the new utilities.


nafur:
While it shows no discernible impact on arithmetic logics on SMT-LIB, the unit tests is significantly faster. On my machine in debug mode, evaluating `(a+b)^40` goes from ~2.2s to ~275ms. (for n=10, it is still 61ms vs 42ms).

ajreynol (approved):

PR #7967, opened 2022-01-19
Fix a subtle issue with double negations in coverings proof

nafur:
When the coverings solver finds a conflict that consists only of a single negated assertion, the way we constructed the lemma would silently drop the double negation. While this is not a problem by itself, the generated proof is out of sync then.
This PR makes sure the double negation stays in place and is only removed by rewriting in the core lemma processing.
Fixes cvc5/cvc5-projects#430.


HanielB (approved):
LGTM

PR #7968, opened 2022-01-20
Fix `Nth-Update` rule, add `Update-Bound` rule

4tXJ7f:
This commit updates the inferences in the sequences array solver to more
closely resemble the current rules in the paper. Concretely, it adds an
implementation of the `Update-Bound` rule and it fixes the `Nth-Update`
rule to take into account all three branches in the conclusion of the
rule.


src/theory/strings/array_core_solver.cpp
```diff
@@ -81,6 +81,19 @@ void ArrayCoreSolver::checkNth(const std::vector<Node>& nthTerms)
...
   }
+  for (uint i=0; i < nthTerms.size(); i++) {
+    for (uint j=i+1; j < nthTerms.size(); j++) {
+       std::vector<Node> exp;
+       Node x = nthTerms[i][0];
+       Node y = nthTerms[j][0];
+       Node n = nthTerms[i][1];
+       Node m = nthTerms[j][1];
+       if (d_state.areEqual(n, m)) {
+         Node lem = nm->mkNode(OR, nm->mkNode(EQUAL, x, y), nm->mkNode(DISTINCT, x, y));
```
ajreynol, line 92: Don't use DISTINCT here, just do a split on the equality.

ajreynol (changes requested):
One minor request.

Also, is CI failing?

ajreynol (approved):

PR #7969, opened 2022-01-20
Fix proofs for trivial cases of datatypes tester merge

ajreynol:
Fixes https://github.com/cvc5/cvc5-projects/issues/433.

Also ensures we don't add (= t t) as part of a reported conflict.

### DIFF
--- a/src/theory/datatypes/theory_datatypes.cpp
+++ b/src/theory/datatypes/theory_datatypes.cpp
@@ -902,7 +902,10 @@ void TheoryDatatypes::addTester(
     std::vector<Node> conf;
     conf.push_back(j);
     conf.push_back(t);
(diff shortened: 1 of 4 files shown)


src/theory/datatypes/theory_datatypes.cpp
```diff
@@ -902,7 +902,10 @@ void TheoryDatatypes::addTester(
     std::vector<Node> conf;
     conf.push_back(j);
     conf.push_back(t);
-    conf.push_back(jt[0].eqNode(t_arg));
+    if (jt[0]!=t_arg)
```
HanielB, line 905: format

HanielB (approved):
LGTM, just need formatting

PR #7971, opened 2022-01-20
Add sls-reachability benchmarks to sets and bags

mudathirmahgoub:
This PR converts sls-reachability benchmarks to cvc5 syntax for sets/bags. 
Benchmarks from card/fol_0000001.smt2 to card/fol_0000044.smt2 are duplicates, so they are removed. 
This PR does not modify code (only tests).


ajreynol:
You should only add 3-4 benchmarks from this set, not all of them.

PR #7973, opened 2022-01-20
Fix trivial explantions in sequences array solver

ajreynol:
Also tightens the explanation slightly, we were previously adding a spurious equality to the representative, instead of the base of the normal form.


4tXJ7f (approved):
LGTM

PR #7976, opened 2022-01-24
Let RealAlgebraicNumber fall back to Rationals

nafur:
Right now, the `RealAlgebraicNumber` can not be used if libpoly is not available. This makes using it in the arithmetic rewriter incredibly awkward. This PR makes it fall back to a wrapper around `Rational` if libpoly is disabled, making its usage way easier.


4tXJ7f (approved):
LGTM

PR #7977, opened 2022-01-24
[FP] Fix unused variable warning

4tXJ7f:


nafur (approved):
LGTM.

src/theory/fp/theory_fp.cpp
```diff
@@ -866,7 +866,7 @@ bool TheoryFp::collectModelValues(TheoryModel* m,
 
       // At most one of the flags (NaN, inf, zero) can be set
       Node one = nm->mkConst(BitVector(1U, 1U));
-      size_t numFlags = 0;
+      size_t numFlags CVC5_UNUSED = 0;
```
martin-cs, line 869: If `ee->getRepresentative(compNaN)` et al. can be stored in a Node (which, really, it should be able to be) this could be simplified to:
```
Assert((cNaN == one ? 1 : 0) + (cInf == one ? 1: 0) + (cZero == one ? 1 : 0) <= 1);
```

so avoiding the need to compute any of this if it is not being checked.

martin-cs (approved):

src/theory/fp/theory_fp.cpp
4tXJ7f: Ah, good point. I have refactored things a bit.

src/theory/fp/theory_fp.cpp
```diff
@@ -855,22 +855,26 @@ bool TheoryFp::collectModelValues(TheoryModel* m,
...
-      Assert(ee->hasTerm(compNaN) && ee->getRepresentative(compNaN).isConst());
-      Assert(ee->hasTerm(compInf) && ee->getRepresentative(compInf).isConst());
-      Assert(ee->hasTerm(compZero)
-             && ee->getRepresentative(compZero).isConst());
+      Assert(ee->hasTerm(compNaN));
+      Assert(ee->hasTerm(compInf));
+      Assert(ee->hasTerm(compZero));
+      Node rCompNaN = ee->getRepresentative(compNaN);
+      Node rCompInf = ee->getRepresentative(compInf);
+      Node rCompZero = ee->getRepresentative(compZero);
```
nafur, line 863:
```suggestion
      TNode rCompNaN = ee->getRepresentative(compNaN);
      TNode rCompInf = ee->getRepresentative(compInf);
      TNode rCompZero = ee->getRepresentative(compZero);
```

nafur: As we are trying to get rid of the overhead of three uint64-additions, we should not add unnecessary reference counting here...

martin-cs: Fair.
( FWIW my goal wasn't about overhead as such, more about simplicity of code. )

4tXJ7f: Yeah, the whole block is only executed when assertions are enabled, so the overhead shouldn't really matter :)

PR #7979, opened 2022-01-24
Always compile RANs

nafur:
This PR makes sure that we always compile with real algebraic numbers, following #7976.


4tXJ7f (approved):
LGTM

PR #7980, opened 2022-01-24
Fixes and improvements to sygus satisfiable query generation

ajreynol:
Fixes issues related to the use of partial functions.  Also ensures we avoid duplicate queries.

### DIFF
--- a/src/theory/quantifiers/query_generator.cpp
+++ b/src/theory/quantifiers/query_generator.cpp
@@ -155,7 +155,7 @@ bool QueryGenerator::addTerm(Node n, std::ostream& out)
   return true;
 }
 
-void QueryGenerator::checkQuery(Node qy, unsigned spIndex)
+void QueryGenerator::checkQuery(Node qy, unsigned spIndex, std::ostream& out)
 {
   if (d_allQueries.find(qy) != d_allQueries.end())
   {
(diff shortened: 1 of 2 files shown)


src/theory/quantifiers/query_generator.cpp
```diff
@@ -135,9 +141,14 @@ bool QueryGenerator::addTerm(Node n, std::ostream& out)
...
     {
       // take two random queries
-      std::shuffle(qsi.begin(), qsi.end(), Random::getRandom());
-      Node qy = nm->mkNode(AND, qsi[0], qsi[1]);
-      checkQuery(qy, i);
+      size_t rindex = Random::getRandom().pick(0, qsi.size() - 1);
+      size_t rindex2 = Random::getRandom().pick(0, qsi.size() - 1);
+      if (rindex == rindex2)
+      {
+        rindex2 = rindex + 1 == qsi.size() ? 0 : rindex + 1;
```
4tXJ7f, line 148: I guess we don't care that the distribution of pairs is not uniform anymore, right? Can we maybe add a comment about that?

4tXJ7f (approved):
LGTM, it would just be good to add a comment regarding the non-uniformity of the query pairs we pick.

src/theory/quantifiers/query_generator.cpp
ajreynol: It's not a big deal, I'd prefer this than a while loop to avoid collisions.

ajreynol: Actually its simple to make it uniform, I've updated.

PR #7982, opened 2022-01-25
[Strings] Avoid trivial explanation

4tXJ7f:
The `STRINGS_N_UNIFY` inference could produce trivial explanations. For
example, if both `x` and `y` were of kind `seq.unit`, then the
explanation was `(= 1 1)`. This lead to issues when checking proofs.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -201,8 +201,7 @@ if (WIN32)
 endif ()
 
 #-----------------------------------------------------------------------------#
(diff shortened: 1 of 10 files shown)


ajreynol (approved):
Is there a regression we can add?

4tXJ7f:
I have added Yoni's test case that triggered the issue. We'll also have to fix that regression when using the array solver.

PR #7986, opened 2022-01-25
Add Card solver to bags

mudathirmahgoub:
This PR avoids reducing bag.card terms with quantifiers unless it is needed. 
It does so by building a cardinality graph for bags and adds appropriate constraints on internal nodes in the graph and min size constraints for the leaves.
This works for many of the benchmarks where each internal node has a single set of children. 
Reduction lemmas are only generated when there are multiple sets of children. 
Cardinality for bag.difference_remove and bag.duplicate_removal is not supported and would handled in future PR


src/theory/bags/solver_state.h
```diff
@@ -41,16 +41,35 @@ class SolverState : public TheoryState
...
+  bool hasCardinalityTerms() const;
+
   /** get all bag terms that are representatives in the equality engine.
    * This function is valid after the current solver is initialized during
    * postCheck. See SolverState::initialize and BagSolver::postCheck
    */
   const std::set<Node>& getBags();
+
+  /**
+   * get all cardinality terms
```
ajreynol, line 69: Can you say a bit more what is stored in the map?

src/theory/bags/solver_state.h
```diff
@@ -41,16 +41,35 @@ class SolverState : public TheoryState
...
    * @param n has the form (bag.count e A)
-   * @pre bag A needs is already registered using registerBag(A)
+   * @pre bag A is already registered using registerBag(A)
    * @return a lemma (= skolem (bag.count eRep ARep)) where
    * eRep, ARep are representatives of e, A respectively
    */
   Node registerCountTerm(TNode n);
+
+  /**
+   * @param n has the form (bag.card A)
```
ajreynol, line 51: Can you doc what this method does?

src/theory/bags/inference_generator.h
```diff
@@ -170,6 +176,37 @@ class InferenceGenerator
...
+   * @param cardTerm a term of the form (bag.card A) where A has type (Bag E)
+   * @param n is a node of the form (bag x c) of type (Bag E)
+   * @return an inference that represents the following implication
+   * (=>
+   *     (and (= A (bag x c)) (>= 0 c))
+   *     (= (bag.card A) c))
+   */
+  InferInfo cardBagMake(const std::pair<Node, Node>& pair, Node n);
+  /**
+   * @param premise a boolean node
```
ajreynol, line 197: This parameter seems vague, what guarantees do we have on premise?

src/theory/bags/inference_generator.h
```diff
@@ -46,6 +46,12 @@ class InferenceGenerator
    *   (>= (bag.count e A) 0)
    */
   InferInfo nonNegativeCount(Node n, Node e);
+  /**
+   * @param n a node of integer type
```
ajreynol, line 50: Is `n` a cardinality term?

ajreynol (changes requested):
Some initial comments and requests.

src/theory/bags/solver_state.h
mudathirmahgoub: Done.

src/theory/bags/solver_state.h
mudathirmahgoub: Done

src/theory/bags/inference_generator.h
mudathirmahgoub: I added a comment saying the premise explains why parent equals the disjoint union of its children. The premise can actually be literal true after rewriting when we have 
`(= (bag.union_disjoint A B) (bag.union_disjoint A B))`
But in general, we would have this equality `(= bagRep (bag.union_disjoint A B))` from the equality engine which we need as a premise to the implication in the generated inference.

src/theory/bags/inference_generator.h
mudathirmahgoub: Yes, but this function is called on a skolem variable equals to a card term

mudathirmahgoub: add this documentation
 ` @param n a node of integer type that equals to a card term`

src/theory/bags/card_solver.cpp
```diff
@@ -0,0 +1,455 @@
...
+CardSolver::~CardSolver() {}
+
+void CardSolver::reset() { d_cardGraph.clear(); }
+
+bool CardSolver::isLeaf(const Node& bag)
+{
+  Node rep = d_state.getRepresentative(bag);
+  Trace("bags-card") << "-----------------------------------------"
+                     << std::endl;
+  Trace("bags-card") << "d_cardGraph :" << std::endl;
```
ajreynol, line 55: You shouldn't for a for-loop in the code to print Trace information.  Guard this block with `Trace.isOn("bags-card")`.

ajreynol (approved):
One minor request, else LGTM

src/theory/bags/card_solver.cpp
mudathirmahgoub: Deleted the whole block which was for debugging, and should not be merged to master.
I only use auto keywords for debugging. I simply forgot to delete the code.

PR #7989, opened 2022-01-25
Properly recognize whether the current commit is a tag

nafur:
This actually fixes what #7901 tried: when uploading the documentation, we need to figure out whether it was a tag and we should upload as a release as well.
Furthermore it fixes how we setup the deploy keys.


nafur:
See https://github.com/nafur/cvc5-versioning/runs/4943086844 and https://github.com/nafur/cvc5-versioning/runs/4943087157.

mpreiner (approved):

PR #7990, opened 2022-01-26
Send all `nth` terms to the core array solver

4tXJ7f:
This updates our policy for the sequences array solver to send all
updates to the array core solver.


ajreynol:
Does this fix a wrong model issue? If so, can we add the regression?

ajreynol (approved):

PR #7991, opened 2022-01-26
Add additional check to avoid cyclic substitution

nafur:
The substitutions we extract from equalities in the nonlinear solver would sometimes generate cyclic substitutions.
This PR tries harder to avoid such cases.
Fixes cvc5/cvc5-projects#444.


ajreynol:
Is there a regression that can be added for this?

nafur:
> Is there a regression that can be added for this?

Not really. The example I used to fix it runs for more than 4 minutes and I haven't seen it terminate yet. Note that z3 solves it way quicker, though, so we might want to save it as a challenge problem.

src/theory/arith/nl/equality_substitution.cpp
```diff
@@ -22,6 +22,22 @@ namespace theory {
 namespace arith {
 namespace nl {
 
+namespace {
+struct ShouldTraverse : public SubstitutionMap::ShouldTraverseCallback
```
ajreynol, line 26: FYI, for the substitute utitlity I had used, it was also important not to traverse beneath certain arithmetic operators, see comment here: https://github.com/cvc5/cvc5/blob/master/src/theory/arith/arith_utilities.cpp#L238

This might lead to issues for transcendentals.  I think I'd recommend changing this method to only return true for "core" arithmetic operators.

ajreynol (changes requested):

src/theory/arith/nl/equality_substitution.cpp
nafur: Ah, right, I somehow forgot about this detail. I added this.

ajreynol (approved):

PR #7992, opened 2022-01-27
[sat] Add option to disable Minisat simplifications

HanielB:
Useful to better control behavior, specially in terms of how the solver performs
when these simplifications (mainly variable and clause elimination) are
disabled, e.g. with incremental and/or proofs.

Also renames the previously confusing "--minisat--elimination" option, which
only refers to variable elimination, a subset of the Minisat simplifications.

Moreover deletes an old comment which no longer applies.


src/smt/set_defaults.cpp
```diff
@@ -684,10 +684,9 @@ void SetDefaults::setDefaultsPost(const LogicInfo& logic, Options& opts) const
     }
   }
 
-  // until bugs 371,431 are fixed
-  if (!opts.prop.minisatUseElimWasSetByUser)
+  if (!opts.prop.minisatVarElimWasSetByUser)
```
aniemetz, line 687: Don't we also have to guard for `minisatSimp` here?

HanielB: No. The "varElim" option was added (it does not exist on original Minisat) specifically to allow all other simplifications except variable elimination, which is problematic for models and the other stuff in this block.

aniemetz: The description of `minisatSimp` says `use Minisat variable and clause elimination, plus other simplifications` though?

HanielB: Yes. So the behavior is that if `minisatSimp` *and* `minisatVarElim` are on, the solver will do variable and clause elimination and some other stuff. If `minisatSimp` is off, does nothing. If `minisatSimp` is on but `minisatVarElim` is off, does all the simplifications except for variable elimination.

Changing the helper of `minisatVarElim` to `"use Minisat variable elimination (requires --minisat-simplification)" would help? It'd not account for the disabling of var elim even when `minisatSimp` is on, though, but I think it'd get convoluted to explain this is the helper.

aniemetz: So I think this should be made more clear in both help messages. Right now it is implying a very different behavior (them being independent, and both enabling variable elimination).

HanielB: I gave it a shot, PTAL

src/options/prop_options.toml
```diff
@@ -64,12 +64,20 @@ name   = "SAT Layer"
   help       = "refine theory conflict clauses (default false)"
 
 [[option]]
-  name       = "minisatUseElim"
+  name       = "minisatVarElim"
```
ajreynol, line 67: Can we just make all of this a single, 3 mode option?  Having 2 options here seems more confusing.

ajreynol (changes requested):

src/options/prop_options.toml
HanielB: Good point. I'll do that.

src/prop/minisat/simp/SimpSolver.cc
```diff
@@ -62,27 +61,18 @@ SimpSolver::SimpSolver(Env& env,
...
-      use_elim(options().prop.minisatUseElim && !enableIncremental),
       merges(0),
       asymm_lits(0),
       eliminated_vars(0),
       elimorder(1),
-      use_simplification(!enableIncremental && !options().smt.unsatCores
-                         && !pnm)  // TODO: turn off simplifications if
-                                   // proofs are on initially
-      ,
+      use_simplification(
```
ajreynol, line 68: We should not be passing `pnm` to SimpSolver, nor checking whether it is non-null here. I think we need to change `!pnm` in this check to `!env.isSatProofProducing()`.

ajreynol (changes requested):

src/options/prop_options.toml
```diff
@@ -64,12 +64,22 @@ name   = "SAT Layer"
...
-  type       = "bool"
-  default    = "true"
-  help       = "use Minisat elimination"
+  long       = "minisat-simplification=MODE"
+  type       = "MinisatSimpMode"
+  default    = "ALL"
+  help       = "Simplifications to be performed by Minisat."
+  help_mode  = "Modes for Minisat simplifications."
+[[option.mode.ALL]]
+  name = "status"
```
ajreynol, line 75: Fix name.

src/options/prop_options.toml
```diff
@@ -64,12 +64,22 @@ name   = "SAT Layer"
...
+  help = "Variable and clause elimination, plus other simplifications."
+[[option.mode.NO_VAR_ELIM]]
+  name = "no-var-elim"
```
ajreynol, line 78: Maybe call this `clause-elim` and the mode `CLAUSE_ELIM`.

src/prop/minisat/simp/SimpSolver.cc
ajreynol: Ping.

Would also help to make this not check for incremental here.

HanielB: This requires changing several interfaces, I think it's better to be done in anather PR.

ajreynol: Ok

ajreynol (approved):

PR #7995, opened 2022-01-27
Add skolem lemmas for bags card terms

mudathirmahgoub:
This PR refactors the way skolem lemmas are generated for bags, count terms, and card terms. 
As a side effect, this refactoring fixed https://github.com/cvc5/cvc5-projects/issues/481


ajreynol:
LGTM, but does this fix an issue?  Can we add a regression?

mudathirmahgoub:
> LGTM, but does this fix an issue? Can we add a regression?

No, but it adds missing skolem lemmas for card terms

PR #7998, opened 2022-01-28
Try a bit harder on the EQ_NCTN rewrite rule

BrunoDutertre:
This makes a small change to the sequence rewriter to work a bit harder
on one of the rewrite rule. This fixes some performance issues we've
observed.

Signed-off-by: Bruno Dutertre <dutebrun@amazon.com>

### DIFF
--- a/src/theory/strings/sequences_rewriter.cpp
+++ b/src/theory/strings/sequences_rewriter.cpp
@@ -117,7 +117,8 @@ Node SequencesRewriter::rewriteStrEqualityExt(Node node)
     Node prev = ctn;
     ctn = rewriteContains(ctn);
     Assert(!ctn.isNull());
-    if (ctn != prev && ctn.getKind() == STRING_CONTAINS) {
+    if (ctn != prev && ctn.getKind() == STRING_CONTAINS)
+    {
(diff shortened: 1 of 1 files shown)


src/theory/strings/sequences_rewriter.cpp
```diff
@@ -114,8 +114,14 @@ Node SequencesRewriter::rewriteStrEqualityExt(Node node)
   {
     // must call rewrite contains directly to avoid infinite loop
     Node ctn = nm->mkNode(STRING_CONTAINS, node[r], node[1 - r]);
+    Node prev = ctn;
     ctn = rewriteContains(ctn);
     Assert(!ctn.isNull());
+    if (ctn != prev && ctn.getKind() == STRING_CONTAINS) {
+      prev = ctn;
```
4tXJ7f, line 121:
```suggestion
    if (ctn != prev && ctn.getKind() == STRING_CONTAINS)
    {
      prev = ctn;
```

4tXJ7f (changes requested):
LGTM

4tXJ7f (approved):
LGTM

PR #8000, opened 2022-01-29
Fix handling of `LogicException` during solving

### LINKED ISSUE #7974: Fatal failure at at /cvc5/src/prop/minisat/core/Solver.cc:1971
Hi, for this formula, cvc5https://github.com/cvc5/cvc5/commit/6d3ef3f8b8256f5e17f5a045e42b5dc1effb281f debug build throws a fatal failure.
If enable `--strings-exp`, it does not return results for a long time.
For cvc5 nightly build, it could not return anything for a long time with `--strings-exp` either.

```
$ cvc5-debug -i -q delta.smt2
(error "Strings Incomplete (due to Negative Membership) by default, try --strings-exp option.")
Fatal failure within void cvc5::Minisat::Solver::pop() at /home/sml/cvc5/src/prop/minisat/core/Solver.cc:1971
Check failure

 decisionLevel() == 0

Aborted (core dumped)
$ cat delta.smt2
(declare-fun v () String)
(declare-fun a () Int)
(push)
(assert (and (= 0 (mod 0 a)) (not (str.in_re v (re.* (re.comp (str.to_re "")))))))
(check-sat)

```

4tXJ7f:
Fixes #7974. This commit fixes our handling of `LogicException`s during solving. Instead of leaving MiniSat in a bad state, it catches the exception, resets the decision trail, prints the error message, and then returns `unknown`.


ajreynol:
I think we still should be allowed to throw a LogicException here.  

The thing I don't like about this fix is that it would still allow the same kind of problem to occur with other LogicExceptions that we recognize dynamically.  Is there a more general fix?

ajreynol (approved):
LGTM

PR #8006, opened 2022-01-31
Add bag.filter operator

mudathirmahgoub:


src/theory/bags/bags_utils.cpp
```diff
@@ -683,11 +708,42 @@ Node NormalForm::evaluateBagMap(TNode n)
...
+  Node ret = BagsUtils::constructConstantBagFromElements(t, mappedElements);
+  return ret;
+}
+
+Node BagsUtils::evaluateBagFilter(TNode n)
+{
+  Assert(n.getKind() == BAG_FILTER);
+
+  // (bag.filter p (as bag.empty (Bag T)) = (as bag.empty (Bag T))
+  // (bag.filter p (bag "a" 3) ((bag "b" 2))) =
```
ajreynol, line 720: This line doesnt parse, do you mean a disjoint union here?

ajreynol (changes requested):
It looks like the inferences involve only positive membership, is there something we have to do for negative membership into filter?

src/theory/bags/bags_utils.cpp
mudathirmahgoub: Yes, fixed.

mudathirmahgoub:
> It looks like the inferences involve only positive membership, is there something we have to do for negative membership into filter?

I guess not. Negative membership here means multiplicity is zero. 
From this table, I don't think we need inference rules for completeness. 
|(p x)| (bag.member x A) |(bag.member x (filter p A)) | Comment|
| --- | ----------- |------------|------------|
| true | 0 | 0 | - |
| true | 0 | 1 | unsat (down rule) |
| true | 1 | 0 | unsat (up rule) |
| true | 1 | 1 | sat (down + up) |
| false | 0 | 0 | - |
| false | 0 | 1 | unsat (down rule) |
| false | 1 | 0 | sat (up rule) |
| false | 1 | 1 | unsat (up rule) |

ajreynol (approved):

PR #8007, opened 2022-01-31
[Rewriter] Always rewrite again when kind changes

4tXJ7f:
Fixes https://github.com/cvc5/cvc5-projects/issues/438. The rewriter
returned  `REWRITE_DONE` prematurely when rewriting operators such as
`bvand` in certain cases. If only a single term was left after removing
duplicates from a `bvand`, the kind of the current node would change and
another pre-rewrite could apply, but the rewriter still returned
`REWRITE_DONE`. After adding an assertion  in the rewriter that checks
that `REWRITE_DONE` implies that no other pre-rewrite applies, it was
discovered that not only the bit-vector rewriter has this issue, but
other rewriters, e.g., the Boolean rewriter, as well. This commit
changes our semantics for `REWRITE_DONE` to mean that no other rewrite
applies provided that the kind did not change. This makes it easier for
developers to decide when to return `REWRITE_DONE`, because it only
requires reasoning about rewrites that apply to a single kind.


mpreiner (approved):
LGTM

src/theory/rewriter.cpp
```diff
@@ -210,16 +210,31 @@ Node Rewriter::rewriteTo(theory::TheoryId theoryId,
...
 
           // Put the rewritten node to the top of the stack
           rewriteStackTop.d_node = response.d_node;
           TheoryId newTheory = theoryOf(rewriteStackTop.d_node);
-          // In the pre-rewrite, if changing theories, we just call the other theories pre-rewrite
+          // In the pre-rewrite, if changing theories, we just call the other
+          // theories pre-rewrite. If the kind of the node was changed, then we
+          // pre-rewrite again.
           if (newTheory == rewriteStackTop.getTheoryId()
+              && originalKind == rewriteStackTop.d_node.getKind()
```
ajreynol, line 224: Is it the case that we can remove the `theoryOf` condition here?  Perhaps it is implied by the kind not changing now?

4tXJ7f: Yes, you are right. I am going to remove that condition.

PR #8008, opened 2022-01-31
Add new ordering functors for arithmetic rewriter

nafur:
This PR provides a clean implementation of the comparators from the normal form utility, also supporting real algebraic numbers. They will be used within the refactored arithmetic rewriter.


4tXJ7f:
Can't we somehow combine this with #7964 and have a single ordering?

nafur:
> Can't we somehow combine this with #7964 and have a single ordering?

This one is planned to replace the one from #7964. #7964 simply extends the old one to support RANs until it is no longer necessary.

ajreynol (approved):

PR #8010, opened 2022-01-31
[BV] Fix order of rewrites for `concat`

4tXJ7f:
This commit fixes the order of the rewrites for bit-vector
concatentation such that the rewrite that merges adjacents constants
comes last (the commit also updates the comments to be less confusing).
Before, it could happen that the rewrite step that removes unnecessary
extract operations introduces constants then then would not get merged
before returning `REWRITE_DONE`.


4tXJ7f:

mpreiner (approved):

PR #8012, opened 2022-02-01
Add node utils for the arithmetic rewriter

nafur:
This PR adds a bunch of small utilities to inspect or create nodes in the arithmetic rewriter.


src/theory/arith/rewriter/node_utils.h
```diff
@@ -0,0 +1,163 @@
...
+#ifndef CVC5__THEORY__ARITH__REWRITER__NODE_UTILS_H
+#define CVC5__THEORY__ARITH__REWRITER__NODE_UTILS_H
+
+#include "base/check.h"
+#include "expr/node.h"
+#include "util/integer.h"
+#include "util/rational.h"
+#include "util/real_algebraic_number.h"
+
+namespace cvc5::theory::arith::rewriter {
```
ajreynol, line 27: I'd prefer the nested syntax for namespaces here, for consistency.

src/theory/arith/rewriter/node_utils.h
```diff
@@ -0,0 +1,163 @@
...
+ * that monomial is either empty (implicitly one) or  a list of non-values
+ * (neither rational nor real algebraic numbers). If multiplicity is rational,
+ * sort the monomial, create a nonlinear mult term and defer to the appropriate
+ * overload. Otherwise return the nonlinear product of the two, i.e.
+ * `(NONLINEAR_MULT multiplicity *monomial)`. The monomial is taken as rvalue as
+ * it may be modified in the process.
+ *
+ */
+Node mkMultTerm(const RealAlgebraicNumber& multiplicity,
+                std::vector<Node>&& monomial);
```
ajreynol, line 159: Why is this a double reference?

src/theory/arith/rewriter/node_utils.h
```diff
@@ -0,0 +1,163 @@
+/******************************************************************************
```
ajreynol, line 1: Do any of these methods coincide with those in https://github.com/cvc5/cvc5/blob/master/src/theory/arith/arith_utilities.h?

I think I'd prefer modifying/extending that file instead of adding a new utility file.

ajreynol (changes requested):

src/theory/arith/rewriter/node_utils.h
nafur: The method possibly changes the monomial argument in a way that it does not make any sense to the caller. Thus, the caller should no longer use it afterwards. We therefore move it into the function (the `&&`), as noted in the comment.

src/theory/arith/rewriter/node_utils.h
nafur: `arith_utilities` has `mkRationalNode` and `mkBoolNode` (which is called `mkConst` in this PR). Also it has `mkMult`, which is different from `mkMultTerm`, though. Given that the semantics of `mkMultTerm` are somewhat special (it assumes stuff about the monomial), I'd rather keep this separate/internal to the rewriter and not "advertise" it to the rest of arithmetic.

src/theory/arith/rewriter/node_utils.cpp
```diff
@@ -0,0 +1,105 @@
...
+ *
+ * Utilities for nodes in the arithmetic rewriter.
+ */
+
+#include "theory/arith/rewriter/node_utils.h"
+
+#include "base/check.h"
+#include "theory/arith/rewriter/ordering.h"
+
+namespace cvc5::theory::arith::rewriter {
```
ajreynol, line 21: This namespace too.

ajreynol (approved):

PR #8013, opened 2022-02-01
Add addition utilities for the arithmetic rewriter

nafur:
This PR adds methods to deal with sums in the arithmetic rewriter. Sums are stored as `std::map<Node,RealAlgebraicNumber>` (see the code for some reasoning), and we implement adding terms to sums, collecting sums into nodes and distributing a multiplication over nested sums.


src/theory/arith/rewriter/addition.h
```diff
@@ -0,0 +1,85 @@
...
+#define CVC5__THEORY__ARITH__REWRITER__ADDITION_H
+
+#include <map>
+#include <iosfwd>
+
+#include "expr/node.h"
+#include "theory/arith/rewriter/ordering.h"
+#include "util/real_algebraic_number.h"
+
+namespace cvc5::theory::arith::rewriter {
```
ajreynol, line 28: Split the namespaces here.

src/theory/arith/rewriter/addition.cpp
```diff
@@ -0,0 +1,259 @@
...
+
+#include <iostream>
+
+#include "base/check.h"
+#include "expr/node.h"
+#include "theory/arith/rewriter/node_utils.h"
+#include "theory/arith/rewriter/ordering.h"
+#include "util/real_algebraic_number.h"
+
+namespace cvc5::theory::arith::rewriter {
```
ajreynol, line 26: Ditto.

ajreynol (changes requested):
Minor requests.

Also, as a longer term goal I'm wondering if it makes sense to use the new utilities in place of this one: https://github.com/cvc5/cvc5/blob/master/src/theory/arith/arith_msum.h.

Or do you consider this to be rewriter specific?

nafur:
> Also, as a longer term goal I'm wondering if it makes sense to use the new utilities in place of this one: https://github.com/cvc5/cvc5/blob/master/src/theory/arith/arith_msum.h.
> 
> Or do you consider this to be rewriter specific?

My goal is to split everything into one part that actually modifies expressions and one that merely inspects them (assuming they are rewritten). My goal would be to uses these utilities instead of `ArithMSum` eventually, when appropriate. My feeling was that it is used as a very heavy hammer in some places where a much simpler node wrapper would also do.

ajreynol (approved):

PR #8014, opened 2022-02-01
Add utilities to rewrite atoms for the arithmetic rewriter

nafur:
This PR adds utilities to rewrite atoms in the arithmetic rewriter. They retain compatibility with the current normal_form utility, but make the process more transparent so that changing the normal form in the future should be simpler.


ajreynol (changes requested):
Minor request, else LGTM

ajreynol (approved):

PR #8015, opened 2022-02-01
[Arrays] Fix response for `store` chains

4tXJ7f:
When rewriting applying the rule `store(store(a,i,v),i,w) --->
store(a,i,w)`, the arrays rewriter would return `REWRITE_DONE`. However,
this is incorrect if there are three or more nested stores of that kind
or the rewrite `store(a,i,select(a,i)) ---> a` applies to the result.
This commit fixes the response to always be `REWRITE_AGAIN`.


4tXJ7f:

ajreynol (approved):

PR #8017, opened 2022-02-01
Extend proof step buffer to optionally ensure unique conclusions

ajreynol:
Fixes https://github.com/cvc5/cvc5-projects/issues/439.

This extended (theory) proof step buffer with options for discarding duplicate conclusions.  This option is now used in the strings theory proof construction, which may have duplication of conclusions due to purification.  It simplifies some of the logic regarding popStep in this utility which is required due to the new policy.


HanielB (approved):
LGTM

PR #8018, opened 2022-02-01
Add variant of get-difficulty for full effort lemmas

ajreynol:
Makes it so that we don't consider lemmas at standard effort by default.  This ensures e.g. arith unate lemmas do not impact difficulty by default.

### DIFF
--- a/src/theory/difficulty_manager.cpp
+++ b/src/theory/difficulty_manager.cpp
@@ -51,11 +51,14 @@ void DifficultyManager::getDifficultyMap(std::map<Node, Node>& dmap)
   }
 }
 
-void DifficultyManager::notifyLemma(Node n)
+void DifficultyManager::notifyLemma(Node n, bool inFullEffortCheck)
 {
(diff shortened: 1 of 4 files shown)


src/theory/difficulty_manager.cpp
```diff
@@ -51,11 +51,15 @@ void DifficultyManager::getDifficultyMap(std::map<Node, Node>& dmap)
...
 {
-  if (options::difficultyMode() != options::DifficultyMode::LEMMA_LITERAL)
+  if (options::difficultyMode() != options::DifficultyMode::LEMMA_LITERAL_ALL)
   {
-    return;
+    if (!inFullEffortCheck
+        || options::difficultyMode() != options::DifficultyMode::LEMMA_LITERAL)
+    {
+      return;
+    }
```
HanielB, line 62: I think it be simpler to test that you return when `!inFullEffortCheck && options::difficultyMode() == options::DifficultyMode::LEMMA_LITERAL` holds.

PR #8019, opened 2022-02-01
Fix docs warnings

nafur:
This fixes a bunch of warnings when generating our sphinx documentation.
They are mostly related to incorrect indentation/spacing, line breaks where no line breaks should be, or missing code blocks.
Note that running `clang-format` causes some of these issues.


src/api/cpp/cvc5_kind.h
```diff
@@ -1939,12 +1939,16 @@ enum Kind : int32_t
   /**
    * Match expressions.
    * For example, the smt2 syntax match term
-   *   `(match l (((cons h t) h) (nil 0)))`
+   * `(match l (((cons h t) h) (nil 0)))`
    * is represented by the AST
    *
+   * \rst
+   * .. code:: smtlib
```
aniemetz, line 1946: This is technically not smt-lib (there's no MATCH_BIND_CASE and MATCH_CASE in smt-lib, it's denoted differently), but literally how it is represented internally. Not sure if this makes a difference for the docs, though.

src/api/cpp/cvc5_kind.h
```diff
@@ -1939,12 +1939,16 @@ enum Kind : int32_t
   /**
    * Match expressions.
    * For example, the smt2 syntax match term
-   *   `(match l (((cons h t) h) (nil 0)))`
+   * `(match l (((cons h t) h) (nil 0)))`
```
aniemetz, line 1942: This should be an smt-lib block.

src/api/cpp/cvc5_kind.h
nafur: I guess we could use lisp instead, or use no highlighting at all.

aniemetz (approved):

PR #8020, opened 2022-02-01
Add table.product operator

mudathirmahgoub:


src/theory/datatypes/datatypes_utils.h
```diff
@@ -0,0 +1,83 @@
+/******************************************************************************
```
ajreynol, line 1: Note there is already `theory_datatypes_utils.h`.

Although this file should perhaps be called `tuple_utils.h`.

src/theory/bags/bags_rewriter.cpp
```diff
@@ -654,6 +655,20 @@ BagsRewriteResponse BagsRewriter::postRewriteFold(const TNode& n) const
   }
   return BagsRewriteResponse(n, Rewrite::NONE);
 }
+
+BagsRewriteResponse BagsRewriter::postRewriteProduct(const TNode& n) const
```
ajreynol, line 659: I'd expect there to be a rewrite that does evaluation in the case that all arguments are constant.

ajreynol (changes requested):

src/theory/datatypes/datatypes_utils.h
mudathirmahgoub: File renamed to tuple_utils and class renamed to TupleUtils.

src/theory/bags/bags_rewriter.cpp
mudathirmahgoub: Yes, this function is only called if not all children are constants. 
When they are constants, this function
https://github.com/cvc5/cvc5/blob/master/src/theory/bags/bags_rewriter.cpp#L68
will call `evaluateProduct` which computes the product from constants. 
https://github.com/cvc5/cvc5/pull/8020/files#diff-a9026cb0a61c5f5cc41cb5b0753617a74d133dd35186dc08fa37d192df4f69a2R119

ajreynol: Ah cool, didnt realize this was handled more generally.

test/regress/regress1/bags/product3.smt2
```diff
@@ -0,0 +1,21 @@
...
+
+(declare-fun x () (Tuple Int Int Int))
+(declare-fun y () (Tuple Int Int Int))
+(declare-fun z () (Tuple Int Int Int Int Int Int))
+
+(assert (bag.member x A))
+(assert (bag.member y B))
+(assert (bag.member z C))
+
+(assert (distinct x y ((_ tuple_project 0 1 2) z ((_ tuple_project 3 4 5) z))))
```
ajreynol, line 19: Does this parse?  Seems like there is a missing paren after the first z.

mudathirmahgoub: The parse didn't complain :) Do we have a problem in the parser?

ajreynol: Yes, this is a problem (assuming you agree this is not what you meant to write).

mudathirmahgoub: Fixed. I just put the parentheses in the wrong places. Both versions return sat (with different models)

src/theory/sets/rels_utils.h
```diff
@@ -10,90 +10,60 @@
...
+ public:
+  /**
+   * compute the transitive closure of a binary relation
+   * @param members constant nodes of type (Tuple E E) that are known to in the
+   * relation rel
+   * @param rel a binary relation of type (Set (Tuple E E))
+   * @pre all members need to be constants
+   * @return the transitive closure of the relation
+   */
+  static std::set<Node> computeTC(std::set<Node> members, Node rel);
```
ajreynol, line 36: Thanks for updating this, let's update this to pass a reference. `std::set<Node>& members`.

ajreynol (changes requested):

src/theory/bags/bags_utils.h
```diff
@@ -94,6 +94,30 @@ class BagsUtils
...
   static Node evaluateBagFilter(TNode n);
 
+  /**
+   * @param n of the form (table.product A B) where A , B of types (Bag T1),
+   * (Bag T2) respectively.
+   * @param e1 a tuple of type T1 of the form (tuple a1 ... an)
+   * @param e2 a tuple of type T2 of the form (tuple b1 ... bn)
+   * @return  (tuple a1 ... an b1 ... bn)
+   */
+  static Node constructTableTuple(TNode n, std::vector<TNode> elements);
```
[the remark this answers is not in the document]
mudathirmahgoub: Actually this function is replaced with TupleUtils::constructTupleFromElements which does receive a constant reference to a vector. 
So this function is removed.

src/theory/sets/rels_utils.h
mudathirmahgoub: Done.

ajreynol (approved):

PR #8023, opened 2022-02-02
CMS can handle conflict resource limit

msoos:
This is how :)

Sorry, couldn't compile this due to various issues, I'll work on it later but it should e kinda OK, I hope!


mpreiner:
@msoos Thanks for the clarification! This method is currently not used at all and will be removed in the near future since we'll refactor how resource limits will be used with the SAT solvers.

msoos:
Ah ok. So no need for this. Let me close this then!

Issue #8016, opened 2022-02-01
(Default) invalid model issue in integer arithmetic


merlinsun:
Hi, for this formula, cvc5 https://github.com/cvc5/cvc5/commit/ac2a185f2895095f2c8e1ddb52aa33c40ab07e2e debug build gives an invalid model.
```
$ cvc5 -q --check-models delta.smt2
Fatal failure within void cvc5::smt::CheckModels::checkModel(cvc5::theory::TheoryModel*, const cvc5::context::CDList<cvc5::NodeTemplate<true> >&, bool) at /home/sml/cvc5/src/smt/check_models.cpp:134
Internal error detected SolverEngine::checkModel(): ERRORS SATISFYING ASSERTIONS WITH MODEL:
assertion:     (exists ((V Int)) (and (= 2 v) (= v ((_ iand 1) 1 v))))
simplifies to: false
expected `true'.
Run with `--check-models -v' for additional diagnostics.
Aborted (core dumped)
$ cat delta.smt2
(declare-fun v () Int)
(assert (exists ((V Int)) (and (= 2 v) (= v ((_ iand 1) 1 v)))))
(check-sat)
```

PR #8026, opened 2022-02-02
Fix invalid rewrite involving iand

ajreynol:
Fixes https://github.com/cvc5/cvc5/issues/8016.


src/theory/arith/arith_rewriter.cpp
```diff
@@ -749,8 +749,10 @@ RewriteResponse ArithRewriter::postRewriteIAnd(TNode t)
     }
     if (t[i].getConst<Rational>().getNumerator() == Integer(2).pow(bsize) - 1)
     {
-      // ((_ iand k) 111...1 y) ---> y
-      return RewriteResponse(REWRITE_DONE, t[i == 0 ? 1 : 0]);
+      // ((_ iand k) 111...1 y) ---> (mod y 2^k)
+      Node twok = nm->mkConstInt(Rational(Integer(2).pow(bsize)));
+      Node ret = nm->mkNode(kind::INTS_MODULUS, t[1-i],  twok);
```
yoni206, line 754:
```suggestion
      Assert(i==0 || i==1);
      Node ret = nm->mkNode(kind::INTS_MODULUS, t[1-i],  twok);
```

yoni206 (approved):
LGTM — one suggestion inline for your consideration.

src/theory/arith/arith_rewriter.cpp
ajreynol: This is a pretty obvious assertion, i is being looped 0, <2 in the for-loop above :)

PR #8027, opened 2022-02-02
Add missing null terminators for regexp

ajreynol:
Led to some LFSC proof checking failures on strings benchmarks.


4tXJ7f (approved):
LGTM

PR #8031, opened 2022-02-02
Fix rewrite for eliminating constant factors of PI from argument to sine

ajreynol:
Fixes https://github.com/cvc5/cvc5-projects/issues/441.
Fixes https://github.com/cvc5/cvc5-projects/issues/414.


nafur (approved):
LGTM.

PR #8040, opened 2022-02-03
Eliminate more static uses of rewrite

ajreynol:


src/theory/quantifiers/sygus/ce_guided_single_inv.cpp
```diff
@@ -431,7 +431,12 @@ void CegSingleInv::setSolution()
   if (!d_solvedf.empty())
   {
     // replace the final solution into the solved functions
-    finalSol.applyToRange(d_solvedf, true);
+    finalSol.applyToRange(d_solvedf);
+    // rewrite the solution
+    for (size_t i = 0, nslv = d_solvedf.d_subs.size(); i < nslv; i++)
+    {
+      d_solvedf.d_subs[i] = rewrite(d_solvedf.d_subs[i]);
+    }
```
nafur, line 439:
```suggestion
    for (Node& n : d_solvedf.d_subs)
    {
      n = rewrite(n);
    }
```

nafur (approved):
LGTM.

src/theory/quantifiers/sygus/ce_guided_single_inv.cpp
ajreynol: I don't think this updates the vector, right?

nafur: It does. The Node is taken as a non-const reference and is updated in the loop body.

ajreynol: I see, yeah, its implicitly a double pointer.

PR #8041, opened 2022-02-03
Always produce assertions

ajreynol:
There is no reason to disable `--produce-assertions`, this simplifies the code to assume that assertions are always available, and always warns if the user disables this option.


src/smt/assertions.cpp
```diff
@@ -37,13 +37,13 @@ namespace smt {
...
     : EnvObj(env),
       d_absValues(absv),
-      d_produceAssertions(false),
       d_assertionList(userContext()),
       d_assertionListDefs(userContext()),
       d_globalDefineFunLemmasIndex(userContext(), 0),
       d_globalNegation(false),
       d_assertions(env)
 {
+    d_globalDefineFunLemmas.reset(new std::vector<Node>());
```
4tXJ7f, line 46: I think that it would be a bit better to move this to the initializer list.

src/smt/proof_manager.cpp
```diff
@@ -304,9 +304,8 @@ std::shared_ptr<ProofNode> PfManager::getFinalProof(
 void PfManager::getAssertions(Assertions& as,
                               std::vector<Node>& assertions)
 {
+  // note that the assertion list is always available
```
4tXJ7f, line 307:
```suggestion
```

test/unit/api/cpp/solver_black.cpp
```diff
@@ -2102,7 +2102,7 @@ TEST_F(TestApiBlackSolver, blockModelValues2)
   Term x = d_solver.mkConst(d_solver.getBooleanSort(), "x");
   d_solver.assertFormula(x.eqTerm(x));
   d_solver.checkSat();
-  ASSERT_THROW(d_solver.blockModelValues({x}), CVC5ApiException);
+  ASSERT_NO_THROW(d_solver.blockModelValues({x}));
```
4tXJ7f, line 2105: We need to change the tests for the language bindings, too. They are currently failing.

src/options/smt_options.toml
```diff
@@ -245,6 +245,7 @@ name   = "SMT Layer"
   long       = "produce-assertions"
   type       = "bool"
   alias      = ["interactive-mode"]
+  default    = "true"
   help       = "keep an assertions list (enables get-assertions command)"
```
4tXJ7f, line 249: Can we change the help message to say that this is always on and changing it has no effect?

4tXJ7f (changes requested):
Some unit tests are failing and I have some suggestions.

4tXJ7f (approved):
LGTM, thanks for the updates.

PR #8043, opened 2022-02-03
Use Add instead of Plus in pythonic API docs

nafur:
This fixes the current documentation builds for the pythonic API

### DIFF
--- a/docs/api/python/pythonic/arith.rst
+++ b/docs/api/python/pythonic/arith.rst
@@ -62,7 +62,7 @@ equal (``==``)
 not equal (``!=``)
   :py:meth:`cvc5_z3py_compat.ExprRef.__ne__`
 
-.. autofunction:: cvc5_z3py_compat.Plus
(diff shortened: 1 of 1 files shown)


alex-ozdemir (approved):

docs/api/python/pythonic/arith.rst
```diff
@@ -62,7 +62,7 @@ equal (``==``)
 not equal (``!=``)
   :py:meth:`cvc5_z3py_compat.ExprRef.__ne__`
 
-.. autofunction:: cvc5_z3py_compat.Plus
+.. autofunction:: cvc5_z3py_compat.Add
 .. autofunction:: cvc5_z3py_compat.Mult
 .. autofunction:: cvc5_z3py_compat.Sub
 .. autofunction:: cvc5_z3py_compat.UMinus
```
aniemetz, line 68: This should actually be `.Neg`.

PR #8047, opened 2022-02-03
Simplify handling of disequalities in strings

ajreynol:
This simplifies how string disequalities are handled, and fixes a caching bug in our implementation of extensionality.

The simplifications are two-fold:
- We track disequalities via assertions, not via an equality engine callback
- We process disequalities by iterating on the disequality list, not via iterating on pairs of equivalence classes

Extensionality is fixed by not *explaining* disequalities, which leads to being out of sync with the cache (as the previous explanation for the disequality might not hold in another SAT context).

This fixes the last known issues with `--seq-array=X`.


4tXJ7f (approved):
LGTM, great find!

PR #8048, opened 2022-02-04
Remove unused `SEQ_NTH_TOTAL` kind

4tXJ7f:
Originally, the idea was to expand the definition of `SEQ_NTH` to use
`SEQ_NTH_TOTAL` (i.e., a total version of that operator), but we have
since then decided to just use `SEQ_NTH`.


ajreynol (approved):
Nice!

ajreynol:
You can update the regression for now, I think its fair to say that term does not evaluate.

Issue #5461, opened 2020-11-18
(check-unsat-cores) Fatal failure at smt/smt_engine.cpp:1542


wintered:
```
[577] % cvc4 --check-unsat-cores -q small.smt2 
Fatal failure within void CVC4::SmtEngine::checkUnsatCore() at smt/smt_engine.cpp:1542
Internal error detectedSmtEngine::checkUnsatCore(): produced core was satisfiable.
Aborted
[578] % 
[578] % cat small.smt2 
(declare-fun x () Int)
(declare-fun y () Int)
(assert (< x 0))
(assert (= (* x x) 16))
(assert (= x y))
(assert (> ((_ iand 4) x y) 0))
(check-sat)
[579] %
```

OS: Ubuntu 18.04
Commit: 83a502c

nafur:
```
(declare-fun x () Int)
(declare-fun y () Int)
(assert (< y 0))
(assert (> ((_ iand 1) x y) 0))
(check-sat)
```

rainoftime:
Maybe related
~~~~
(set-logic QF_UFLIA)
(declare-fun i1 () Int)
(declare-fun i4 () Int)
(declare-fun i7 () Int)
(declare-fun i9 () Int)
(declare-fun i10 () Int)
(assert (= 327 i7 ((_ iand 4) i4 i9) i10 i1))
(check-sat)

$./cvc4 --check-models xx.smt2
Fatal failure within void CVC4::smt::CheckModels::checkModel(CVC4::smt::Model*, CVC4::context::CDList<CVC4::NodeTemplate<true> >*, bool) at /CVC4/src/smt/check_models.cpp:259
Internal error detectedSmtEngine::checkModel(): ERRORS SATISFYING ASSERTIONS WITH MODEL:
assertion:     (let ((_let_0 ((_ iand 4) i4 i9))) (and (= 327 i7) (= i7 _let_0) (= _let_0 i10) (= i10 _substvar_11_)))
simplifies to: false
expected `true'.
Run with `--check-models -v' for additional diagnostics.
Aborted
~~~~

nafur:
Second one:
```
(set-logic QF_LIA)
(declare-fun i () Int)
(assert (= 0 ((_ iand 1) 1 i)))
(check-sat)
```

ajreynol:
The latter two examples are due to not recognizing that iand is non-linear and hence should be logic exceptions.

PR #8053, opened 2022-02-04
Fix unsoundness in IAND solver

ajreynol:
Fixes https://github.com/cvc5/cvc5/issues/5461.  The initial lemma schema was assuming arguments were in bounds.

This also modifies the solver so that we throw logic exceptions when encountering iand or transcendentals.


ajreynol:
Lets merge https://github.com/cvc5/cvc5/pull/8054 first.

yoni206 (approved):

PR #8056, opened 2022-02-04
Add well formed term check to solver engine

ajreynol:
Fixes https://github.com/cvc5/cvc5-projects/issues/413.


src/smt/solver_engine.h
```diff
@@ -1048,6 +1048,12 @@ class CVC5_EXPORT SolverEngine
    */
   const Options& options() const;
 
+  /**
+   * Check that the given term is a valid closed term, which can be used as an
+   * argument to e.g. assert, get-value, block-model-values, etc.
```
aniemetz, line 1053:
```suggestion
   * argument to, e.g., assert, get-value, block-model-values, etc.
```

src/smt/solver_engine.h
```diff
@@ -1048,6 +1048,12 @@ class CVC5_EXPORT SolverEngine
...
+   */
+  void ensureWellFormedTerm(const Node& n, const std::string& src) const;
```
aniemetz, line 1055: Document what `src` is.

PR #8059, opened 2022-02-04
Throw option exceptions when combining input conversion preprocessing with sygus

ajreynol:
Fixes https://github.com/cvc5/cvc5-projects/issues/436.


src/smt/set_defaults.cpp
```diff
@@ -870,6 +870,27 @@ bool SetDefaults::usesSygus(const Options& opts) const
...
 
+bool SetDefaults::usesInputConversion(const Options& opts,
+                                      std::ostream& reason) const
+{
+  if (opts.smt.solveBVAsInt != options::SolveBVAsIntMode::OFF)
+  {
+    reason << "solveBVAsInt";
+    return true;
+  }
+  if (opts.smt.solveIntAsBV > 0)
```
aniemetz, line 881: Why is this option a `uint64_t` rather than `bool` in the toml option configuration?

ajreynol: The value is the bitwidth to try.

aniemetz (approved):

PR #8061, opened 2022-02-04
[Rewriter] Do not attempt to rewrite constants

4tXJ7f:
This commit changes our rewriter policy to not attempt to rewrite nodes
with no children.


ajreynol (changes requested):
Actually as I recall, the better check is `getNumChildren()==0`, this is already an optimization here: https://github.com/cvc5/cvc5/blob/master/src/theory/rewriter.cpp#L92.

ajreynol:
Can you update to check `getNumChildren()==0` instead of `isConst()`?

4tXJ7f:
> Can you update to check `getNumChildren()==0` instead of `isConst()`?

I have finally addressed this.

PR #8062, opened 2022-02-04
Distinguish proof mode from unsat core mode

ajreynol:
Previously, we were making unsat cores mode the source of our configuration for unsat cores and proofs at the same time.

This led to a lot of unnecessary complication in the logic, and would allow bugs e.g. if produce-difficulty but not produce-unsat-cores was enabled.

Fixes https://github.com/cvc5/cvc5-projects/issues/406.


src/smt/set_defaults.cpp
```diff
@@ -109,37 +96,72 @@ void SetDefaults::setDefaultsPre(Options& opts)
...
+  // enabled is implied
+  if (opts.smt.checkProofs || opts.driver.dumpProofs
+      || opts.smt.proofMode == options::ProofMode::FULL)
   {
     opts.smt.produceProofs = true;
   }
 
-  if (opts.smt.produceProofs
-      && opts.smt.unsatCoresMode != options::UnsatCoresMode::FULL_PROOF)
+  // this check assumes the user has requested *full* proofs
```
HanielB, line 107: can't we add a check in this block that if the user requested proofs but set the proof mode to something other than full it's an error?

HanielB (commented):
Looks really nice. A question though:

src/smt/set_defaults.cpp
HanielB: Ah, indeed, that'd be nice.

HanielB (approved):

PR #8064, opened 2022-02-04
Option exception when incompatible with proofs

ajreynol:
This changes set defaults so that it doesn't silently disable proofs or unsat cores.

Fixes https://github.com/cvc5/cvc5-projects/issues/440.


aniemetz (approved):

ajreynol:
Waiting to merge this until after IJCAR deadline, since it may inadvertently impact the proof evaluation.

Issue #6654, opened 2021-06-01
(ext-rewrite-quant, full-saturate-quant) refutation unsoundness on string formula 


zhendongsu:
Commit: https://github.com/cvc5/cvc5/commit/172573dba45f7d231ec06a3a3992f41cf794b75e
OS: Ubuntu 18.04

```
[598] % cvc5 -q --strings-exp small.smt2
sat
[599] % cvc5 -q --strings-exp --ext-rewrite-quant --full-saturate-quant small.smt2
unsat
[600] % cat small.smt2
(declare-fun a () String)
(declare-fun b () String)
(assert (= b (str.replace_re a (str.to_re "A") "A")))
(assert (not (str.prefixof (str.++ b b) a)))
(check-sat)
[601] % 
```

PR #8065, opened 2022-02-04
Fix indexof_re reduction

ajreynol:
Fixes https://github.com/cvc5/cvc5/issues/6654.

Note that this regression now answers unknown with `--ext-rewrite-quant`; the regression without this option is added.


4tXJ7f (approved):
LGTM, nice find.

PR #8066, opened 2022-02-05
[Seq] Check types for split on indices

4tXJ7f:
We applied `STRINGS_ARRAY_EQ_SPLIT` to pairs of `seq.nth` terms of
different types. This lead to assertion (in debug) and performance
problems (in production). The commit adds a check.


ajreynol (approved):
Good catch

PR #8070, opened 2022-02-06
Correct search location for CLN libs

andrewvaughanj:
On (e.g.,) openSUSE, the default install path (via CMake) for libraries is `lib64` and not `lib`. However, `FindCLN.cmake` is hard-coded to only search in `lib`.

This PR corrects `FindCLN.cmake` such that the build looks in `${CMAKE_INSTALL_LIBDIR}` instead.


nafur (approved):
LGTM, thanks!

PR #8071, opened 2022-02-06
Remove superfluous 'lbsd' for Editline

aytey:
On some systems (e.g., openSUSE), you can link against `editline` *without* having the BSD compatibility libraries. For example, on my machine, I have `libedit-devel` installed without having `libbsd-devel`, which then leads to:

```
: && /usr/bin/c++ -fPIC -fno-inline -Og -Wall -Wsuggest-override -Wnon-virtual-dtor -Wimplicit-fallthrough -Wshadow -fno-extern-tls-init -Wno-class-memaccess -ggdb3 -g -flto -fno-fat-lto-objects   -shared -Wl,-soname,libmain-test.so -o src/main/libmain-test.so src/main/CMakeFiles/main.dir/command_executor.cpp.o src/main/CMakeFiles/main.dir/interactive_shell.cpp.o src/main/CMakeFiles/main.dir/signal_handlers.cpp.o src/main/CMakeFiles/main.dir/time_limit.cpp.o src/main/CMakeFiles/main.dir/options.cpp.o src/main/CMakeFiles/main-test.dir/driver_unified.cpp.o  -Wl,-rpath,/home/avj/clones/cvc5/master/build/src/parser:/home/avj/clones/cvc5/master/build/src:/home/avj/clones/cvc5/master/build/deps/lib64:/home/avj/clones/cvc5/master/build/deps/lib  src/parser/libcvc5parser.so.1  -ledit  -lbsd  -ltinfo  src/libcvc5.so  deps/lib64/libcln.so  deps/lib/libpolyxx.so  deps/lib/libpoly.so  /usr/lib64/libgmp.so && :
/usr/lib64/gcc/x86_64-suse-linux/11/../../../../x86_64-suse-linux/bin/ld: cannot find -lbsd
```

However, there is no requirement that you need `lbsd` when linking `ledit`.

This PR therefore makes it such that `bsd` is only linked when you're using `editline` if you have it.

Signed-off-by: Andrew V. Teylu <andrew@tey.lu>

### DIFF
--- a/cmake/FindEditline.cmake
+++ b/cmake/FindEditline.cmake
@@ -42,7 +42,11 @@ if(Editline_INCLUDE_DIRS)
   unset(CMAKE_REQUIRED_INCLUDES)
 
   if(NOT CMAKE_SYSTEM_NAME STREQUAL "Darwin")
-    set(Editline_LIBRARIES ${Editline_LIBRARIES} bsd tinfo)
+    find_library(BSD_LIBRARIES NAMES bsd)
+    if (BSD_LIBRARIES)
+      set(Editline_LIBRARIES ${Editline_LIBRARIES} bsd)
+    endif()
+    set(Editline_LIBRARIES ${Editline_LIBRARIES} tinfo)
   endif()
 endif()
 


andrewvaughanj:
Okay, so this doesn't work; I'll revise it to check if `bsd` can be found, and only use it if I can be found.

nafur (commented):
I have three comments:
- if we touch this code anyway, can we remove this case for MacOS?
- can we somehow check whether we actually need `libbsd`? (similar to the check above)
- this would change the linking order of tinfo and bsd, I think. Is this okay?

andrewvaughanj:
> * if we touch this code anyway, can we remove this case for MacOS?
>

I didn't write that bit (I did write most of the rest of the file):

- https://github.com/cvc5/cvc5/commit/65e5a909b10855d2e6b3844f7bc4efb7fbe4fe2f#diff-21bc1bd8b72fa3213c73b5f99d4f4c47699a62c8126447b6b7b85c0a66d93721R42

> * can we somehow check whether we actually need `libbsd`? (similar to the check above)
>

I think that's what the use of `PkgConfig` is supposed to do, but I guess it doesn't work (i.e., it should be returning `tinfo` and `bsd` on Ubuntu and doesn't).

Thought: I wonder if the problem is `tinfo`? Maybe `tinfo` needs `bsd`, but `libedit` doesn't actually *need* `tinfo`? That would explain why `PkgConfig` isn't returning `bsd`.

> * this would change the linking order of tinfo and bsd, I think. Is this okay?
>

I can easily just flip the order if that's a concern.

nafur:
Hah, so I added this noop MacOS branch :)
So yeah, if you could simplify it, that would be great. And I'd rather not change the linking order, I've been burnt by this in the past...

mpreiner:
@aytey is this PR still relevant?

aytey:
> @aytey is this PR still relevant?

Erm, wow, this is what you get for having too many notifications.

Yes, this still affects system such as OpenSUSE where you can have `libedit` but not `libbsd`.

I'll check Ubuntu this week.

aytey:
Just checked on OpenSUSE:

```
-- Found Editline: /usr/include;/usr/include/editline
-- Editline library: edit;bsd;tinfo
```

This is incorrect; you don't need to have `libbsd` to link `libedit`

aytey:
PR is updated; looks good now:

```
-- Checking for module 'libedit'
--   Found libedit, version 3.1
-- Found Editline: /usr/include;/usr/include/editline
-- Editline library: edit;tinfo
```

cmake/FindEditline.cmake
```diff
@@ -42,7 +42,11 @@ if(Editline_INCLUDE_DIRS)
   unset(CMAKE_REQUIRED_INCLUDES)
 
   if(NOT CMAKE_SYSTEM_NAME STREQUAL "Darwin")
-    set(Editline_LIBRARIES ${Editline_LIBRARIES} bsd tinfo)
+    set(Editline_LIBRARIES ${Editline_LIBRARIES} tinfo)
+    find_library(BSD_LIBRARIES NAMES bsd)
+    if (BSD_LIBRARIES)
+      set(Editline_LIBRARIES ${Editline_LIBRARIES} bsd)
+    endif()
```
mpreiner, line 49:
```suggestion
    find_library(BSD_LIBRARIES NAMES bsd)
    if (BSD_LIBRARIES)
      set(Editline_LIBRARIES ${Editline_LIBRARIES} bsd)
    endif()
    set(Editline_LIBRARIES ${Editline_LIBRARIES} tinfo)
```

mpreiner: Let's keep the original order, just to make sure not to break anything.

aytey: Good call; done!

mpreiner (approved):

PR #8073, opened 2022-02-07
Simplify formalism of introduced arithmetic skolems

ajreynol:
This ensures that all skolems introduced by arithmetic have concrete terms as witnesses. This avoids introducing witness terms.

This PR is important for ensuring that the "original form" of terms does not contain introduced free variables.  This means that certain features (quantifier elimination, learned literals) are useful to the user.  This was requested by Certora.

This also will improve proof reconstruction for extensions of non-linear arithmetic.


HanielB (commented):
Do I understand correctly that what this PR is doing is introducing defined functions to represent the skolems for these operations rather than using witness terms?

HanielB:
I see, thanks!

HanielB (approved):

PR #8077, opened 2022-02-07
Generalize LFSC concat unify for splitting constants

ajreynol:
Was causing LFSC proof failures on ~5 SMT-LIB strings benchmarks.


4tXJ7f (approved):
LGTM

PR #8080, opened 2022-02-07
Properly sanatize user names in LFSC

ajreynol:
Previously, we would get LFSC proof checking failures if we were using an identifier that was valid in SMT-LIB but not LFSC.

I didn't verify that this is least invasive fix, although it should be conservative.


src/proof/lfsc/lfsc_node_converter.cpp
```diff
@@ -604,11 +601,43 @@ TypeNode LfscNodeConverter::postConvertType(TypeNode tn)
...
+  // where "32" is the unicode value of " ", 32@1 indicates that space was
+  // removed from the user string at position 1.
+  std::string sname = name;
+  size_t sanCount = 0;
+  size_t found;
+  do
+  {
+    found = sname.find_first_not_of("ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz"
+                          "0123456789~!@$%^&*_-+=<>.?/");
+    if (found!=std::string::npos)
```
4tXJ7f, line 621: Apply ClangFormat.

src/proof/lfsc/lfsc_node_converter.cpp
```diff
@@ -604,11 +601,43 @@ TypeNode LfscNodeConverter::postConvertType(TypeNode tn)
...
 {
+  std::stringstream ssan;
   std::stringstream ss;
-  ss << "cvc." << name;
+  // We also must sanitize symbols that are not allowed in LFSC identifiers
+  // here. For the sake of generating unique symbols, we record the santization
+  // as a prefix before the "." separating the user name. Thus, e.g.
+  //   cvc.|a b|
+  // is converted to:
+  //   cvc_32@1.ab
```
4tXJ7f, line 611: I think that it would be good if the sanitized string started with an `@` or a `.` because the SMT-LIB standard specifies that those are reserved for solver-internal uses. This would hopefully ensure that the sanitized symbol names don't accidentally clash with symbols introduced by the user.

src/proof/lfsc/lfsc_node_converter.cpp
```diff
@@ -604,11 +601,43 @@ TypeNode LfscNodeConverter::postConvertType(TypeNode tn)
...
+  // where "32" is the unicode value of " ", 32@1 indicates that space was
+  // removed from the user string at position 1.
```
4tXJ7f, line 613: Why include an index instead of just replacing the character with its unicode, e.g., `cvc.|a32@b|`?

src/proof/lfsc/lfsc_node_converter.cpp
```diff
@@ -604,11 +601,43 @@ TypeNode LfscNodeConverter::postConvertType(TypeNode tn)
...
+  size_t sanCount = 0;
+  size_t found;
+  do
+  {
+    found = sname.find_first_not_of("ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz"
+                          "0123456789~!@$%^&*_-+=<>.?/");
+    if (found!=std::string::npos)
+    {
+      // erase the character, print that we eliminated that character
+      ssan << "_" << static_cast<size_t>(sname[found])+sanCount << "@" << found;
```
4tXJ7f, line 624: Why add `sanCount` here?

4tXJ7f (changes requested):
Just some minor comments about the details of the sanitized symbol.

src/proof/lfsc/lfsc_node_converter.cpp
ajreynol: So the main purpose of this sanitization is to ensure user symbols do not clash with *lfsc* definitions (e.g. proof rule names).

Previously for user symbols, we just did `cvc.X`, i.e. prefix the symbol with "cvc.".

Now we do `cvc_[c1]@[p1]..._[cn]@[pn].Y` where c1...cn are code points and p1...pn are positions and Y is the result of removing all bad characters from X.

This should be injective.  It is unique from LFSC definitions, which dont have `cvc` as a prefix.

I'm not sure why we'd need to special case symbols that start with `@`?

src/proof/lfsc/lfsc_node_converter.cpp
ajreynol: |a32@b| could be the name of the user symbol, that's why the information about santization is pushed before the `.`.

src/proof/lfsc/lfsc_node_converter.cpp
ajreynol: Since we are updating sname in this loop, sanCount ensures we print the position in the original string.

src/proof/lfsc/lfsc_node_converter.cpp
4tXJ7f: Ah I see. I worded my comment above poorly. I was suggesting to add `@` or `.` as a prefix to make sure that they don't clash with user definitions, but this takes care of it then.

src/proof/lfsc/lfsc_node_converter.cpp
4tXJ7f: See my reply to the comment below.

4tXJ7f (approved):
LGTM, sorry about the delay.

PR #8084, opened 2022-02-09
Fix pruning of covering intervals in proofs

nafur:
We need to mirror the pruning of intervals in the coverings solver in the corresponding proof. To match the proof nodes with the coverings-internal intervals, we use ids, where non-interval nodes have the id zero. In this process, we had three separate issues that never showed up because we don't have a proof checker for CAD proofs:
- we did not check for zeros, pruning all non-interval nodes
- we ran the pruning on the direct children of a CAD_RECURSIVE proof step, which are `SCOPE` nodes that don't have ids
- iterating over and then removing children was done by a custom but faulty reimplementation of `std::remove_if`

Thus, we now
- never prune zero nodes,
- prune based on the (single) child node, if we are checking a `SCOPE` node, and
- simply use `std::remove_if`.


HanielB:
How does this impact the non-linear proofs? Makes the solver generate less THEORY_LEMMA or something like that?

nafur:
> How does this impact the non-linear proofs? Makes the solver generate less THEORY_LEMMA or something like that?

It impacts non-linear proofs in the sense that it now generates meaningful proofs again. Right now, all proofs from the coverings solver simply infer `false` from nothing...

HanielB:
But the occurrences of `ARITH_NL_CAD_DIRECT` and `ARITH_NL_CAD_RECURSIVE` would be the same, just their content would change?

nafur:
> But the occurrences of `ARITH_NL_CAD_DIRECT` and `ARITH_NL_CAD_RECURSIVE` would be the same, just their content would change?

They now look like they were always supposed to look, and like they looked at some point in the past. Covering proofs are simply broken right now, most probably at least since #7470.

test/unit/theory/theory_arith_cad_white.cpp
```diff
@@ -359,6 +381,51 @@ void test_delta(const std::vector<Node>& a)
...
+
+  cac.computeVariableOrdering();
+
+  auto cover = cac.getUnsatCover();
+  EXPECT_FALSE(cover.empty());
+  
+  std::vector<Node> mis{dummy(1), dummy(3), dummy(4), dummy(5)};
+  LazyTreeProofGenerator* pg = dynamic_cast<LazyTreeProofGenerator*>(cac.closeProof(mis));
+  EXPECT_TRUE(pg != nullptr);
+  //std::cout << *pg << std::endl;
```
ajreynol, line 425: Remove this.

ajreynol (changes requested):
One minor request, else LGTM

PR #8086, opened 2022-02-09
bv: Add --tlimit-per support for CryptoMiniSat.

mpreiner:


src/util/resource_manager.cpp
```diff
@@ -188,6 +188,11 @@ uint64_t ResourceManager::getResourceUsage() const
 
 uint64_t ResourceManager::getTimeUsage() const { return d_cumulativeTimeUsed; }
 
+uint64_t ResourceManager::getRemainingTime() const
+{
+  return d_options.base.perCallMillisecondLimit - d_cumulativeTimeUsed;
```
nafur, line 193: `d_cumulativeTimeUsed` gives you the overall time spent on check calls so far. You'll have to inspect `d_perCallTimer` instead, I guess `d_limit - clock::now()` with some `duration_cast` like in `elapsed()`.

nafur (approved):
LGTM. Also some comments here? :-)

PR #8087, opened 2022-02-10
Build python wheels in our CI

nafur:
This PR integrates building and publishing cvc5 with its base and pythonic python APIs as a package to PyPi into our CI.
We build wheels for Linux and macOS for CPython 3.6 to 3.10 and PyPy 3.7 and 3.8.
The job is run nightly and for a release, and only published to PyPi for a release (as long as there is no reasonable way to automatically prune nightly builds from either PyPi or TestPyPi).


.github/actions/package-python-wheel/action.yml
```diff
@@ -0,0 +1,109 @@
...
+runs:
+  using: composite
+  steps:
+    - name: Build wheels for Linux
+      if: runner.os == 'Linux'
+      shell: bash
+      run: |
+        echo "::group::Create docker image"
+        res=$(docker image inspect pycvc5-manylinux2014 ; echo $? ; exit 0)
+        if [[ "$res" != 0 ]]; then
```
mpreiner, line 21:
```suggestion
        if ! docker image inspect pycvc5-manylinux2014 > /dev/null 2>&1; then
```

.github/actions/package-python-wheel/action.yml
```diff
@@ -0,0 +1,109 @@
...
+      if: runner.os == 'macOS'
+      uses: ./.github/actions/package-python-wheel-macos
+      with:
+        python-version: 'pypy-3.8'
+
+    - name: Upload wheels to pypi.org
+      if: inputs.upload-to-pypi == 'true'
+      shell: bash
+      env:
+        TWINE_USERNAME: __token__
```
mpreiner, line 86: Should `__token__` be a secret? If yes, then we should maybe use uppercase, e.g., `TWINE_TOKEN`?

.github/actions/package-python-wheel/action.yml
```diff
@@ -0,0 +1,109 @@
...
+        TWINE_PASSWORD: ${{ inputs.pypi-token }}
+      run: |
+        echo "::group::Upload to pypi.org"
+        for wheel in `ls *.whl`
```
mpreiner, line 90: Is it guaranteed that we always have all *.whl built or can it happen that maybe the 3.7 wheel won't be available due to errors? Should we add a check here to make sure that we upload all configured packages?

.github/workflows/package_pypi.yml
```diff
@@ -0,0 +1,52 @@
...
+    - uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
+
+    - name: Install dependencies
+      uses: ./.github/actions/install-dependencies
+      if: runner.os == 'Linux'
+      with:
+        with-documentation: false
+        with-python-bindings: false
```
mpreiner, line 29: Why is this false but for macOS it is true?

cmake/FindCVC5PythonicAPI.cmake
```diff
@@ -0,0 +1,44 @@
...
+if(NOT CVC5PythonicAPI_DOWNLOADED)
+  check_auto_download("CVC5PythonicAPI" "--no-python-bindings")
+endif()
+
+include(ExternalProject)
+
+ExternalProject_Add(
+  CVC5PythonicAPI
+  ${COMMON_EP_CONFIG}
+  GIT_REPOSITORY https://github.com/cvc5/cvc5_z3py_compat.git
```
mpreiner, line 30: Wasn't this repo renamed?

.github/actions/package-python-wheel/action.yml
nafur: No, the username is always `__token__` if an API token is used.

.github/actions/package-python-wheel/action.yml
nafur: I would expect that some of the steps would fail along the way and abort the whole build. Also, we can't really protect against an incomplete release as the Linux build may work while the macOS build may fail. (Unless, of course, we build the wheels, store them as an artifact, and then have a separate job that collects all of them and uploads them separately).

.github/workflows/package_pypi.yml
nafur: For Linux, the compilation takes place within Docker and the software is thus installed in the Docker container. For macOS, compilation is done directly on the host.

cmake/FindCVC5PythonicAPI.cmake
nafur: Yes, see #8131 :)

mpreiner (approved):

PR #8089, opened 2022-02-10
Use witness terms to represent values for large strings in models

ajreynol:
This makes it so that we don't prematurely fail on some Facebook benchmarks.


4tXJ7f (changes requested):
Waiting for additional changes

4tXJ7f (approved):
LGTM

PR #8090, opened 2022-02-10
Eliminate match from LFSC proofs

ajreynol:
The smt 2.6 term `match` is very hard to represent in LFSC, this eliminates it in favor of the ITE term that it is syntax sugar for. Like other aspects of the LFSC conversion, this is part of the trusted core.

This avoids internal type errors in the LFSC node converter.


HanielB (approved):
LGTM

Issue #8094, opened 2022-02-11
Fatal failure at /src/theory/theory_model_builder.cpp:1034


merlinsun:
Hi,
for this instance, cvc5 https://github.com/cvc5/cvc5/commit/50d93bfc5bcd160059522e22310934ea4ecb2c57 debug build
```
$ cvc5 -q delta.smt2
Fatal failure within bool cvc5::theory::TheoryEngineModelBuilder::buildModel(cvc5::theory::TheoryModel*) at /home/sml/cvc5/src/theory/theory_model_builder.cpp:1034
Check failure

 false

Aborted (core dumped)
$ cat delta.smt2
(declare-fun v () String)
(declare-fun r () Bool)
(assert (forall ((V String)) (or r (exists ((V String)) (str.in_re (str.++ "z" v) (re.* (str.to_re (str.from_code (str.len v)))))))))
(check-sat)
```

ajreynol:
This is caused by the new symbolic model values for strings, we require building a value`(str.++ "z" (witness ((s String)) (= (str.len s) 196608)))` which is not considered a legitimate value.

zhendongsu:
The following must be related, but it only triggers with `--check-models`:
```
[533] % cvc5 small.smt2 
sat
[534] % cvc5 --check-models small.smt2 
The model was computed to have strings of length 2400. Based on the current value of option --strings-model-max-len, we only allow strings up to length 1000
The model was computed to have strings of length 1200. Based on the current value of option --strings-model-max-len, we only allow strings up to length 1000
Fatal failure within bool cvc5::theory::TheoryEngineModelBuilder::buildModel(cvc5::theory::TheoryModel*) at /local/suz-local/software/CVC4/src/theory/theory_model_builder.cpp:1034
Check failure

 false

Aborted
[535] % cat small.smt2 
(set-logic QF_S)
(declare-fun x () String)
(declare-fun y () String)
(assert (distinct x (str.++ y y)))
(assert (= 1200 (str.len y)))
(check-sat)
```

PR #8095, opened 2022-02-11
Further relax what is considered a value in the model

ajreynol:
Fixes https://github.com/cvc5/cvc5/issues/8094.

This makes it so that "value-like" terms can appear as subterms in other terms that are then subsequently also considered values.  For example `(str.++ (witness ((x String)) (= (str.len x) 1000)) "A")` is now considered a value.

This also changes the default option for when to use witness terms for strings in models, based on some user feedback.


ajreynol:
FYI @barrettcw .

I'm not totally sure this is the right direction, but it seems to be necessary.

src/theory/theory_model.h
```diff
@@ -358,10 +358,16 @@ class TheoryModel : protected EnvObj
   std::string debugPrintModelEqc() const;
 
   /**
-   * Is the node n a "value"? This is true if n is constant, a constant-like
-   * value (e.g. a real algebraic number) or if n is a lambda.
+   * Is the node n a "value"? This is true if n is a "base value", where
+   * a base value is one where isConst() returns tru, a constant-like
```
nafur, line 362:
```suggestion
   * a base value is one where isConst() returns true, a constant-like
```

src/theory/theory_model.cpp
```diff
@@ -803,15 +804,57 @@ std::string TheoryModel::debugPrintModelEqc() const
...
+struct IsModelValueTag
 {
-  if (node.isConst())
+};
+struct IsModelValueComputedTag
+{
+};
+/** Attribute true for expressions that are quasi-values */
+typedef expr::Attribute<IsModelValueTag, bool> IsModelValueAttr;
+typedef expr::Attribute<IsModelValueComputedTag, bool> IsModelValueComputedAttr;
```
nafur, line 815:
```suggestion
using IsModelValueAttr = expr::Attribute<IsModelValueTag, bool>;
using IsModelValueComputedAttr = expr::Attribute<IsModelValueComputedTag, bool>;
```

nafur (approved):
Just a very minor style issue, otherwise LGTM.

src/theory/theory_model.cpp
```diff
@@ -803,15 +804,57 @@ std::string TheoryModel::debugPrintModelEqc() const
...
+      }
+    }
+    n.setAttribute(IsModelValueAttr(), isQv);
+    n.setAttribute(IsModelValueComputedAttr(), true);
+    return isQv;
   }
-  Kind k = node.getKind();
-  return k == kind::REAL_ALGEBRAIC_NUMBER || k == kind::LAMBDA
-         || k == kind::WITNESS;
+  return n.getAttribute(IsModelValueAttr());
```
4tXJ7f, line 857: Style-wise, it would be a bit easier to read if the method started with:
```
if (n.getAttribute(IsModelValueComputedAttr()))
{
  return n.getAttribute(IsModelValueAttr());
}
```

src/theory/theory_model.cpp
```diff
@@ -803,15 +804,57 @@ std::string TheoryModel::debugPrintModelEqc() const
...
+        // we are a value if we are one of the above kinds
+        isQv = true;
+      }
+      else if (n.getNumChildren() > 0 && rewrite(n) == n)
+      {
+        // note that we must be in rewritten form
+        isQv = true;
+        for (TNode nc : n)
+        {
+          if (!isValue(nc))
```
4tXJ7f, line 841: It would be nice if we could make this method iterative instead of recursive. The terms passed to this function are probably not deeply nested in most cases, but it'd be nice for consistency.

4tXJ7f (changes requested):

src/theory/theory_model.cpp
ajreynol: I've done this, although notice that the algorithm is really complex now to account for how the original computation worked.

4tXJ7f (approved):
LGTM, thanks for the update!

PR #8098, opened 2022-02-14
Support `try` and `all` reconstruction modes.

abdoo8080:
This PR adds a simple procedure to quickly reconstruct Sygus terms that are already in the grammar. This new procedure fails if the terms are not in the grammar.


ajreynol (changes requested):
Looks good, although can we change the default of `sygus-si-rcons` option to ALL?  Is it possible to also support ALL_LIMIT, e.g. by having it fail after `n` terms are enumerated (as determined by cegqiSingleInvReconstructLimit)?

ajreynol (approved):

PR #8100, opened 2022-02-14
Introduce all_simplify and replace old trust rules

Lachnitt:
Instead of printing "trust" in the case of a simplification rule that cannot be translated to an Alethe simplify rule print "all_simplify". This also replaces the heuristic in THEORY_REWRITE to determine which simplify rule is needed.


HanielB (commented):
The motivation here is that we'd not need this heuristic if we were to leverage the DSL rewrites? Until then won't this limit the integration with Sledghammer?

Lachnitt:
Yes, with DSL rewrites we can get rid of a lot the remaining THEORY_REWRITES. For the remaining ones what happened is that we tried to guess what simplify rule they could be the result of and if that fails applied auto. Now, they are translated to ALL_SIMPLIFY where we also apply auto. If a step can be traYes, with DSL rewrites we can get rid of a lot the remaining THEORY_REWRITES. For the remaining ones what happened is that we tried to guess what simplify rule they could be the result of and if that fails applied auto. Now, they are translated to ALL_SIMPLIFY where we also apply auto. If a step can be translated into one of the alethe simplify rules it is usually so easy that it is solvable by auto anyway. nslated into one of the alethe simplify rules it is usually so easy that it is solvable by auto anyway. 

One of the main advantages is that we get to make better use of the proof checker since also rules besides THEORY_REWRITE use simplify rules that should be correct and the alternative would be to print them all as trust/all_simplify rules.

HanielB:
I see, sounds good! You should resolve the conflict though.

HanielB (approved):

PR #8101, opened 2022-02-14
Enable lazy lambda lifting with higher order logic

mudathirmahgoub:


src/theory/uf/ho_extension.cpp
```diff
@@ -42,6 +42,12 @@ HoExtension::HoExtension(Env& env,
       d_uf_std_skolem(userContext())
 {
   d_true = NodeManager::currentNM()->mkConst(true);
+  if (d_env.getLogicInfo().isHigherOrder()
```
ajreynol, line 45: This is not the right place for this, options should never be modified after the SMT engine is initialized.

We should either move this code to `smt/set_defaults.cpp` or simply set the default value of the option to `true` in `options/uf_options`.

ajreynol (changes requested):

src/theory/uf/ho_extension.cpp
mudathirmahgoub: Modified options/uf_options

ajreynol:
Can you comment on why the 2 regressions fail?

ajreynol:
I will address this default option change in a separate PR.

PR #8102, opened 2022-02-15
[Strings] Register terms more lazily in EE

4tXJ7f:
Until now, we have been registering terms in the equality engine of the
string solver eagerly during pre-registration. At the same time, the
shared solver is only considering shared terms in facts that have been
asserted. As a result, it could happen that the string solver was asking
for a value of a shared term during model construction that the shared
solver did not get notified of, leading to an assertion failure. This
commit changes when terms are registered with the equality engine.


4tXJ7f:
Overall results on string benchmarks (300s timeout):

```
                                       status  total  solved    sat  unsat   best  timeout  memout  error  uniq  disagr  time_cpu   memory
config
2022-02-10-strings-refactorReg-fixed-4     ee  69959   69315  42695  26620  65416      596       0     46    40       0  217300.4 387187.4
2022-02-10-strings                         ee  69959   69286  42667  26619  59350      625       0     46    11       0  226681.4 394370.1
```

ajreynol:
So this PR makes it so that we don't call `d_equalityEngine->addTriggerPredicate` on e.g. equalities during preregistration?  It is surprising to me that this is better, as all theories generally do this.

ajreynol:
I think we can close this PR, right?

4tXJ7f:
Yes, let's close it for now.

PR #8104, opened 2022-02-15
Make cache (optionally) persistent in lazy CD proof

ajreynol:
Should address severe performance issues in some use cases of proofs.


HanielB (approved):
LGTM. I have tested it on SMT-LIB and there are no errors.

PR #8105, opened 2022-02-15
Only consider relevant terms in the array-inspired sequence solver

ajreynol:
Avoids segfault in model construction due to asking for the value of irrelevant (non-shared) term.


4tXJ7f (approved):
LGTM, except for the formatting (see Zulip message).

PR #8107, opened 2022-02-16
Enable DAG and tree printing with dot files

vinciusb:
Add a proof option that allows DAG and tree print format when using dot files. 

Signed-off-by: Vinícius Braga Freire <vinicius.braga.freire@gmail.com>


src/proof/dot/dot_printer.h
```diff
@@ -48,17 +48,18 @@ class DotPrinter
    * the child as argument.
    * @param out the output stream
    * @param pn the proof node to print
-   * @param ruleID the id of the rule to print
+   * @param pfLet the map of the variables that were already printed
    * @param scopeCounter counter of how many SCOPE were already depth-first
    * traversed in the proof up to this point
    * @param inPropositionalView flag used to mark the proof node being traversed
    * was generated by the SAT solver and thus belong to the propositional view
+   * @return the id of the current node being printed
```
HanielB, line 56:
```suggestion
   * @return the id of the proof node printed
```

src/proof/proof_node.h
```diff
@@ -133,6 +134,26 @@ inline size_t ProofNodeHashFunction::operator()(
...
 }
 
+inline size_t ProofNodeHashFunction::operator()(const ProofNode* pfn) const
+{
+  uint64_t ret = fnv1a::offsetBasis;
+
+  ret = fnv1a::fnv1a_64(ret, std::hash<Node>()(pfn->getResult()));
+  ret = fnv1a::fnv1a_64(ret, static_cast<size_t>(pfn->getRule()));
+
+  for (const cvc5::Pf& child : pfn->getChildren())
```
HanielB, line 144: Can't it be 
```suggestion
  for (const Pf& child : pfn->getChildren())
```
?

Also store `pfn->getChildren()` into a constant reference that you use in the loop. The same for `pfn->getArguments()` below.

HanielB (changes requested):
Looks nice, just some minor comments.

HanielB:
Thanks for the changes! One thing I forgot: could you move the implementation of the hash function to the `cpp` file? We have been avoiding having implementations in the headers.

HanielB (approved):
LGTM

PR #8109, opened 2022-02-16
Consolidate multiple substitutions to single AND child in trust substitution map

ajreynol:
This ensures that a common proof node is used for a substitution instead of copying its children. This optimization had been noted in a comment but had not yet been enabled.

This should speed up proof reconstruction for substitutions.


HanielB (approved):
LGTM. Testing on SMT-LIB shows no errors and further gains in reconstruction time.

PR #8110, opened 2022-02-16
Remove some irrelevant node kinds from the model

ajreynol:
There are several kinds of nodes that are currently sent to the model that don't need to be.  Furthermore, we require not sending some of these due to the new invariants in https://github.com/cvc5/cvc5/pull/8095.


HanielB (approved):
LGTM

PR #8113, opened 2022-02-17
Relax what is generated in the model from NL

ajreynol:
This makes it so that we don't assign concrete values to equivalence classes with transcendental functions.

It also removes some unused infrastructure.


nafur (approved):
LGTM.

Issue #8111, opened 2022-02-17
(sygus-rr-synth-input) Fatal failure at /cvc5/src/./theory/type_enumerator.h:172


merlinsun:
Hi, for this instance, cvc5 https://github.com/cvc5/cvc5/commit/87937a8e7410d2222d40c209a550abb3f62dc962 debug build
```
$ cvc5 --sygus-rr-synth-input -q delta.smt2
Fatal failure within cvc5::Node cvc5::theory::TypeEnumerator::operator*() at /root/cvc5/src/./theory/type_enumerator.h:172
Check failure

 n.isConst()
Term (str.to_re "") from type enumerator is not constant
Aborted (core dumped)
$ cat delta.smt2
(declare-fun v () String)
(assert (forall ((V String)) (str.in_re (str.++ "" v) (re.++ (str.to_re "b") (re.union (str.to_re "") (str.to_re "z"))))))
(check-sat)
```

PR #8114, opened 2022-02-17
Support some cases of isConst for regular expressions

ajreynol:
Fixes https://github.com/cvc5/cvc5/issues/8111.

Having `isConst` return true is required for some rare use cases of regular expressions, e.g. rewrite rule synthesis involving regular expression variables.


4tXJ7f (approved):
LGTM

PR #8115, opened 2022-02-17
Change inference scheme in transcendentals to rewrite rule

ajreynol:
Avoids introducing new `sine` terms.


nafur (commented):
I think there are cases where we can not recover this lemma.
Assume
`(not (= (sin x) (sin y)))` and somewhere we have `(= x (- y))`, I'm not sure how we would resolve this after that PR...

ajreynol:
Yes, I thought of this, although I think it'd be best to do this on demand, i.e. we only infer the equality if x becomes equal to -y, and those terms already exist.  Right now, we are needlessly introducing new sine terms, which is a bad idea.

src/theory/arith/nl/transcendental/sine_solver.cpp
```diff
@@ -134,28 +134,6 @@ void SineSolver::checkInitialRefine()
...
-            proof = d_data->getProof();
-            Node tmplem =
-                nm->mkNode(Kind::ADD,
-                           t,
-                           nm->mkNode(
-                               Kind::SINE,
-                               nm->mkNode(Kind::MULT, d_data->d_neg_one, t[0])))
-                    .eqNode(d_data->d_zero);
-            proof->addStep(
-                tmplem, PfRule::ARITH_TRANS_SINE_SYMMETRY, {}, {t[0]});
```
nafur, line 152: This is probably the only usage of this proof rule. Should we remove it?

ajreynol: Let's wait.  Once the refactoring of this issue is done, I will revisit.

nafur (approved):
Okay, then LGTM.

PR #8116, opened 2022-02-17
Introduce skolem function to make transcendental function purification unique

ajreynol:
Previously, we could introduce multiple purifications for the same sin term.


nafur (approved):
LGTM.

Issue #7948, opened 2022-01-14
Refutation unsoundness or Fatal failure at src/smt/solver_engine.cpp:1428 (nightly build)


zhendongsu:
```
[558] % cvc5-2022-01-14-x86_64-Linux-production -q --check-unsat-cores small.smt2
Fatal failure within void cvc5::SolverEngine::checkUnsatCore() at /buildbot/ubuntu20-amd64-production/build/src/smt/solver_engine.cpp:1428
Internal error detected SolverEngine::checkUnsatCore(): produced core was satisfiable.
Aborted
[559] % cat small.smt2 
(assert (<= (cos (* (cos (- 3.0)) 0.5)) 1.0))
(check-sat)
```

ajreynol:
An empty unsat core is returned, this suggests a soundness issue in the NL transcendental solver.

zhendongsu:
Another instance:
```
[524] % cvc5 -q --check-unsat-cores small.smt2 
Fatal failure within void cvc5::SolverEngine::checkUnsatCore() at /local/suz-local/software/CVC4/src/smt/solver_engine.cpp:1428
Internal error detected SolverEngine::checkUnsatCore(): produced core was satisfiable.
Aborted
[525] % cat small.smt2 
(declare-fun a () Real)
(declare-fun b () Real)
(assert (= a 0))
(assert (= b (cos a)))
(check-sat)
```

zhendongsu:
Combing the two asserts in the instance above leads to a refutation unsoundness issue:
```
[527] % cvc5 -q small.smt2 
sat
[528] % cvc5 -q --check-unsat-cores small.smt2 
unsat
[529] % cat small.smt2 
(declare-fun a () Real)
(declare-fun b () Real)
(assert (and (= a 0) (= b (cos a))))
(check-sat)
```

ajreynol:
The use of `check-unsat-cores` in these examples imply `simplification=none`, which is required for making the benchmarks not trivially solved.

The unsoundness is caused by incorrectly assigning a concavity region to the argument of `sin` terms.

ajreynol:
The first benchmark times out on master.
The latter 2 benchmarks can be solved with simplification, but time out on master.
Thus, closing the issue.

PR #8117, opened 2022-02-17
Make sine solver sound with respect to region boundaries

ajreynol:
Fixes the last benchmark on https://github.com/cvc5/cvc5/issues/7948.

Fixes a refutation soundness issue in the transcendental solver where a concavity region would be incorrectly assigned to a point if it was between the current model value of `c*PI` and its true value, where `c in {-1, -1/2, 1/2, 1}`.

We now only assign a concavity region if the model value of the argument lies within sound lower/upper bounds for the boundaries.

Notice that this means that points may be unassignable to a region if they lie inside the approximation interval for `c*PI`. An application of `sin` applied to an argument whose model value is that point cannot be refined.

A followup PR will address termination issues where the Taylor degree is incremented even when no function can be refined.


src/theory/arith/nl/transcendental/sine_solver.cpp
```diff
@@ -266,13 +266,27 @@ void SineSolver::checkMonotonic()
                                d_data->d_zero,
                                d_data->d_pi_neg_2,
                                d_data->d_pi_neg};
-  std::vector<Node> mpoints_vals;
+  // sound upper, lower bounds for the above points
+  std::vector<Node> mpointsBound[2];
 
   // get model values for points
-  for (const auto& point : mpoints)
+  TNode tpi = d_data->d_pi;
```
nafur, line 273: Can you add a comment on what exactly `mpointsBound` contains afterwards?
I think `mpointsBound[0/1][x]` is an upper/lower bound on `mpoints[x]`, but I'm really not sure.
Below there is a comment that indicates it is actually lower/upper...

src/theory/arith/nl/transcendental/sine_solver.cpp
```diff
@@ -283,6 +297,7 @@ void SineSolver::checkMonotonic()
   {
     Node sargval = d_data->d_model.computeAbstractModelValue(sarg);
     Assert(sargval.isConst());
+    Rational sargvalr = sargval.getConst<Rational>();
```
nafur, line 300:
```suggestion
    const Rational& sargvalr = sargval.getConst<Rational>();
```

src/theory/arith/nl/transcendental/sine_solver.cpp
```diff
@@ -244,13 +244,26 @@ void SineSolver::checkMonotonic()
...
+      if (j != 2)
+      {
+        // bounds are flipped for negative pi
+        TNode tb = d_data->d_pi_bound[j > 2 ? 1 - i : i];
+        mpointapprox = point.substitute(tpi, tb);
+      }
+      mpointsBound[i].emplace_back(
+          d_data->d_model.computeAbstractModelValue(mpointapprox));
+      Assert(mpointsBound[i].back().isConst());
+    }
```
nafur, line 266: I really don't like this loop... it seems super specific to `mpoints` and weirdly abstract at the same time.
My intuition is more along these lines:
```
for (const Node& point: mpoints)
{
  for (size_t i = 0; i < 2; ++i)
  {
    Node approx = point.substitute(d_data->d_pi_bound[i]);
    mpointsBound[i].emplace_back(d_data->d_model.computeConcreteModelValue(approx));
    Assert(mpointsBound[i].back().isConst());
  }
  if (mpointsBound[0].back().getConst<Rational>() > mpointsBound[1].back().getConst<Rational>())
  {
    std::swap(mpointsBound[0].back(), mpointsBound[1].back());
  }
}
```
This would be robust against changing the order of the evaluation points and is much more clear in why it sometimes flips the bounds (at least to me). Also note that I think we should use `computeConcreteModelValue()`, as they should do the very same thing here... right?

nafur (approved):
LGTM.

PR #8120, opened 2022-02-18
Fixes and additions for LFSC signatures

ajreynol:
Fixes a critical issue in the re intersection rule, where a null terminator was missed.

Also adds a sketch of the theory signature for floating points.


HanielB (approved):
LGTM

PR #8122, opened 2022-02-18
[Seq/Array] Ignore congruent and inactive terms

4tXJ7f:
This commit ensures that our array solver ignores congruent terms. For
example, if we have two update terms where all children are equal in the
current context, our array solver will now only process one of them.
Additionally, this commit removes terms from the relevant term set that
are inactive according to the extended theory infrastructure. This
ensures that the array solver does not process terms that have been
marked as reduced (e.g., through CD simplifications).


src/theory/strings/array_solver.cpp
```diff
@@ -125,14 +127,33 @@ void ArraySolver::checkTerms(const std::set<Node>& termSet)
...
       // for update terms, also check the inverse inference
-      checkTerm(t, true);
+      checkInv = true;
     }
     else if (k != SEQ_NTH)
     {
       continue;
     }
+
+    // Ignore congruent terms
```
ajreynol, line 137: Can we just use `BaseSolver::isCongruent` https://github.com/cvc5/cvc5/blob/master/src/theory/strings/base_solver.cpp#L707?  I think this should have the same effect.

src/theory/strings/term_registry.cpp
```diff
@@ -676,9 +681,20 @@ void TermRegistry::removeProxyEqs(Node n, std::vector<Node>& unproc) const
...
 void TermRegistry::getRelevantTermSet(std::set<Node>& termSet)
 {
-  d_theory.collectAssertedTerms(termSet);
+  std::set<Node> relevantTerms;
+  d_theory.collectAssertedTerms(relevantTerms);
   // also, get the additionally relevant terms
-  d_theory.computeRelevantTerms(termSet);
+  d_theory.computeRelevantTerms(relevantTerms);
+
+  // Remove inactive terms from the relevant term set
```
ajreynol, line 689: I'd rather not having this filtering builtin to this method, or make term registry depend on ExtTheory.  Actually, perhaps the combined method could be a method in ExtfSolver, which has access to both?

ajreynol (changes requested):

src/theory/strings/term_registry.cpp
4tXJ7f: Seems like a good idea. I've done that.

ajreynol:
I think I'm ok with the changes on this PR, do we have performance numbers for this?

src/theory/strings/array_solver.cpp
```diff
@@ -126,14 +128,23 @@ void ArraySolver::checkTerms(const std::set<Node>& termSet)
...
       // for update terms, also check the inverse inference
-      checkTerm(t, true);
+      checkInv = true;
     }
     else if (k != SEQ_NTH)
     {
       continue;
     }
+
+    if (d_bsolver.isCongruent(t)) {
```
ajreynol, line 138: Clang format.

ajreynol (approved):
Remember to clang-format.

Issue #8124, opened 2022-02-20
Invalid model on set and tuple formula


merlinsun:
Hi,
For this instance, cvc5https://github.com/cvc5/cvc5/commit/4683179f4aff1a1b3e797cea23f8f06ed121a406 (debug) gives a wrong model. 
```
$ cvc5 --check-models delta.smt2
Fatal failure within void cvc5::smt::CheckModels::checkModel(cvc5::theory::TheoryModel*, const cvc5::context::CDList<cvc5::NodeTemplate<true> >&, bool) at /root/cvc5/src/smt/check_models.cpp:134
Internal error detected SolverEngine::checkModel(): ERRORS SATISFYING ASSERTIONS WITH MODEL:
assertion:     (exists ((V Real)) (let ((_let_1 ((_ tuple_select 1) x))) (let ((_let_2 ((_ tuple_select 0) x))) (let ((_let_3 0.0)) (and (distinct x (tuple _let_3 _let_3)) (not (set.member (tuple (to_int _let_2) _let_1) e)) (not (set.member (tuple _let_2 (to_int _let_1)) t)))))))
simplifies to: false
expected `true'.
Run with `--check-models -v' for additional diagnostics.
Aborted (core dumped)
$ cat delta.smt2
(set-logic ALL)
(declare-fun x () (Tuple Real Real))
(declare-fun t () (Set (Tuple Real Int)))
(declare-fun e () (Set (Tuple Int Real)))
(assert (exists ((V Real)) (and (distinct x (tuple 0.0 0.0)) (not (set.member (tuple (to_int ((_ tuple_select 0) x)) ((_ tuple_select 1) x)) e)) (not (set.member (tuple ((_ tuple_select 0) x) (to_int ((_ tuple_select 1) x))) t)))))
(check-sat)
```

ajreynol:
This error is caused by a bug in the care graph computed by the datatypes theory.  It is improperly indexing terms in a way where subtyping can cause care pairs to be skipped.  I will commit a fix shortly.

PR #8125, opened 2022-02-21
Fix issue in datatypes care graph computation involving subtyping

ajreynol:
The datatypes care graph was incorrectly computed for constructors taking `Real` arguments, due to subtyping.

This fixes the term index to cache on the *return* type for constructors, not the argument types.  Caching based on argument types was incorrect even for non-parametric datatypes with arguments that have subtyping.

Fixes https://github.com/cvc5/cvc5/issues/8124.

This also does minor cleanup and additions done while debugging the issue.


4tXJ7f (approved):
LGTM

Issue #8127, opened 2022-02-22
Test case unit/api/solver_black: proj_issue434 fails with type checking failure


Ouyancheng:
**Describe the bug**
Test case unit/api/solver_black: proj_issue434 fails on macOS 12.3 and 11.4 as one of the bound var lists fails the type check... 
In line 3042: `ASSERT_NO_THROW(slv.checkSatAssuming({t1073, t510}));` fails. 
If we remove t510 in the assumption it will work. 
but if we solely include t510 as the assumption and print out the bound var lists: 
```
[ RUN      ] TestApiBlackSolver.proj_issue434
n=((_f29_1 Bool))
n[0]=_f29_1
n=((_f31_0 Bool))
n[0]=_f31_0
n=((_f37_1 Bool))
n[0]=_f37_1
n=((_f40_0 Bool))
n[0]=_f40_0
n=((_f45_0 Bool))
n[0]=_f45_0
n=((true Bool))
n[0]=true
/Volumes/Data/workspaces/CVC4/test/unit/api/cpp/solver_black.cpp:3043: Failure
Expected: slv.checkSatAssuming(std::vector<Term>{t510}) doesn't throw an exception.
  Actual: it throws.
[  FAILED  ] TestApiBlackSolver.proj_issue434 (12 ms)
```
One of the bound variable becomes `true`. 

**Command line arguments**: `make units`
**cvc5 version/commit**: 3dd30794611fa1e4344b0f7f3068211533b48f4f 
**Operating system**: macOS 12.3, macOS 11.4

**`configure.sh` options**  

`./configure.sh debug --auto-download`

**`configure.sh` output**  
```
-- The C compiler identification is AppleClang 13.0.0.13000029
-- The CXX compiler identification is AppleClang 13.0.0.13000029
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /Library/Developer/CommandLineTools/usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /Library/Developer/CommandLineTools/usr/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Found Git: /usr/bin/git (found version "2.32.0 (Apple Git-132)") 
-- Building Debug build
-- Performing Test HAVE_C_FLAG_fno_inline
[...]

PR #8130, opened 2022-02-22
Check for free variables in several SolverEngine calls

ajreynol:
Fixes the nightlies due to ensuring that a unit test fails in production (it was failing only in debug).

Fixes https://github.com/cvc5/cvc5/issues/8127.


aniemetz (commented):
As discussed offline, I suggest to check this on the API level, and have internal checks on the solver engine level only in debug mode (selectedly).

mpreiner (approved):
LGTM

PR #8131, opened 2022-02-22
Integrate pythonic api

nafur:
We decided we want to ship the pythonic API together with our base python API.
This PR adds a new target `cvc5_python_api` that first builds the base python API and then copies the pythonic API over. Furthermore we now use the `cvc5.pythonic` module to generate the corresponding documentation.


cmake/FindCVC5PythonicAPI.cmake
```diff
@@ -0,0 +1,44 @@
...
+  check_auto_download("CVC5PythonicAPI" "--no-python-bindings")
+endif()
+
+include(ExternalProject)
+
+ExternalProject_Add(
+  CVC5PythonicAPI
+  ${COMMON_EP_CONFIG}
+  GIT_REPOSITORY https://github.com/cvc5/cvc5_z3py_compat.git
+  GIT_TAG main
```
mpreiner, line 31: Do we really want to always pull main instead of a commit? I'd suggest to use a fixed commit as otherwise it will be hard to debug errors if the pythonic API can change anytime.

cmake/FindCVC5PythonicAPI.cmake
```diff
@@ -0,0 +1,44 @@
...
+if(NOT CVC5PythonicAPI_DOWNLOADED)
+  check_auto_download("CVC5PythonicAPI" "--no-python-bindings")
+endif()
+
+include(ExternalProject)
+
+ExternalProject_Add(
+  CVC5PythonicAPI
+  ${COMMON_EP_CONFIG}
+  GIT_REPOSITORY https://github.com/cvc5/cvc5_z3py_compat.git
```
mpreiner, line 30: We should also rename the repo to `cvc5_pythonic_api` (or similar).

src/api/python/CMakeLists.txt
```diff
@@ -113,6 +113,17 @@ set_target_properties(cvc5_python_base
 
 python_extension_module(cvc5_python_base)
 
+find_package(CVC5PythonicAPI)
+
+add_custom_target(
```
mpreiner, line 118: Add comment about why we have to copy the directory. Can't we directly install the API to `"${CMAKE_CURRENT_BINARY_DIR}/cvc5/pythonic"` in `ExternalProject_Add`?

cmake/FindCVC5PythonicAPI.cmake
nafur: @alex-ozdemir we did want to rename it at some point, right?

cmake/FindCVC5PythonicAPI.cmake
nafur: The pythonic API has no installation mechanism by itself (right now).

cmake/FindCVC5PythonicAPI.cmake
nafur: @mpreiner I'd rather not wait with this PR until it has been renamed, though...

alex-ozdemir: Ah, yes. I'm happy to rename before or after this PR. Your call, Gereon.

alex-ozdemir: Done

mpreiner (approved):

